PR #17, opened 2023-10-25
Clean Up Workflow and Fix Failing Tests

Umang01-hash:
- This PR is targeted towards solving the following existing issues:

## Separation of various data-stores in workflow for testing
In this new approach, we are trying to separate out tests for various datastores like pubsub, ycql ,cql etc . So it will run parallel with `unit test` step and will reduce overall time taken by workflow.

The separation of test is done on the basis of `regex` & `build tags`  for **go test** command using `-run` and `-tags` flag respectively.

## Renaming of Tests:

In order to make the naming uniform, we have renamed all the test related to:
1. YCQL with prefix `Test_YCQL_` 
2. CQL with prefix `Test_CQL_` 
3. EventHub, EventBridge, Avro, Kafka, PubSub with prefix `Test_PubSub_`


aryanmehrotra:
Closing this as workflow has been divided

Issue #89, opened 2023-11-27
Remove duplicate initialisation of metrics in pubsub package


VipulSharma1999:
The same set of metrics is initialized in multiple places in pubsub package, instead, it can be done in a specific file and can be used from there wherever needed in pubsub package.
Thoughts?

PR #90, opened 2023-11-27
Removed duplicate metrics and added in the common file

VipulSharma1999:
- Removed duplicate metrics from the pubsub package and added those in the common file(metrics.go).

Github issue: https://github.com/gofr-dev/gofr/issues/89

### DIFF
--- a/pkg/constants.go
+++ b/pkg/constants.go
@@ -14,5 +14,5 @@ const (
 	PathOpenAPI              = "/.well-known/openapi.json"
 	PathSwagger              = "/.well-known/swagger"
 	PathSwaggerWithPathParam = "/.well-known/swagger/{name}"
-	FrameworkMetricsPrefix   = "zs_"
+	FrameworkMetricsPrefix   = "gofr_"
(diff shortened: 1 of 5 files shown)


vipul-rawat:
@VipulSharma1999 please check the linters, the code quality check is failing

pkg/constants.go
```diff
@@ -14,4 +14,5 @@ const (
 	PathOpenAPI              = "/.well-known/openapi.json"
 	PathSwagger              = "/.well-known/swagger"
 	PathSwaggerWithPathParam = "/.well-known/swagger/{name}"
+	FrameworkMetricsPrefix   = "zs_"
```
vikash, line 17: Should we change all the metrics prefix to gofr instead of zs?

VipulSharma1999: Yes, we can do that.

vikash: Tracking the issue here: https://github.com/gofr-dev/gofr/issues/99

Issue #99, opened 2023-11-28
Change prefix from zs to gofr for all the metrics


VipulSharma1999:
As part of this issue, all metrics prefixes should be changed from zs to gofr.

vikash:
This will be a breaking change and should be part of major release.

PR #105, opened 2023-11-30
change metrics prefix from zs to gofr

Umang01-hash:
- Fixes issue #99.


pkg/datastore/db.go
```diff
@@ -90,23 +90,23 @@ type SQLClient struct {
 //nolint:gochecknoglobals // sqlStats has to be a global variable for prometheus
 var (
 	sqlStats = prometheus.NewHistogramVec(prometheus.HistogramOpts{
-		Name:    "zs_sql_stats",
+		Name:    "gofr_sql_stats",
```
srijan-27, line 93: instead of updating like this, we can use `FrameworkMetricsPrefix` constant, just like PubSub metrics.

Umang01-hash: Absolutely @srijan-27 was just testing if the tests and code are working, will update it in final commit.

srijan-27 (approved):

vipul-rawat (approved):

PR #106, opened 2023-12-03
Fix panic on nil SubscriptionDetails in Google PubSub

### LINKED ISSUE #34: Handle Panic when subscriptionDetails is passed nil in google PubSub
- I was trying to create publisher but got panic, if didn't passed subscriptionDetails

### How to regenerate the panic 
```
disputeStatusPublisher, pubErr := googlePubSub.New(&googlePubSub.Config{
			TopicName:           disputeStatusPublisherTopicName,
			ProjectID:           projectID,
			TimeoutDuration:     30,
			Subscription:        nil,
			Topic:               nil,
			SubscriptionDetails: nil,
		}, logger)`
```
currently it is giving panic due to  https://github.com/gofr-dev/gofr/blob/03794b763967541f0b8358abd392489c06325a83/pkg/datastore/pubsub/google/google.go#L136
		
### Expected :-
It should give an error stating subscriptionDetails is necessary

Onyx2406:
Fixes #34


pkg/datastore/pubsub/google/google.go
```diff
@@ -133,6 +133,12 @@ func createTopicsIfNotExist(ctx context.Context, g *GCPubSub) error {
 }
 
 func createSubscription(ctx context.Context, g *GCPubSub) error {
+
```
vipul-rawat, line 136: Please remove this new line, this would cause a linter error

pkg/datastore/pubsub/google/google.go
```diff
@@ -133,6 +133,12 @@ func createTopicsIfNotExist(ctx context.Context, g *GCPubSub) error {
...
+	// Check if SubscriptionDetails is nil before accessing its properties
+	if g.config.SubscriptionDetails == nil {
```
vipul-rawat, line 138: Now, this is more open to improvement!

The question is do we always require a subscription?

If the user comes looking to create a publish-only application does google pubsub support publishing without a subscription?

If yes, we should skip creating a subscription when no subscription details are provided, and a check should be put when a user is trying to subscribe with no subscriptions!

Onyx2406: Google Cloud PubSub is designed to ensure that messages are delivered to subscribers at least once. This means that even in scenarios where a subscriber is temporarily unavailable, PubSub retains unacknowledged messages in persistent storage. These messages are stored for up to 7 days from their publication date.

Notably, messages that are published before the creation of a subscription cannot be retrieved later. Additionally, there's a maximum message retention period of 7 days. After this period, any undelivered messages are permanently deleted.

Therefore, it is a standard behavior in Google Cloud PubSub for subscribers to receive messages that were published while they were inactive, as long as the messages are less than 7 days old. This feature is particularly useful for ensuring message delivery continuity, even during temporary subscriber downtime.

Onyx2406:
> @Onyx2406 linters are failing, please check

Done

Umang01-hash (approved):

Umang01-hash:
@Onyx2406 Linters failed for your PR. Please check it once.
And also once before pushing try to update your branch with `development`

vipul-rawat:
@Onyx2406 
Your commits are not verified on GitHub, could you please check the [GitHub commit signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)! For the PR to be merged all the commits in the PR needs to have verified signature.

Issue #2, opened 2023-10-24
Upgrade golangci-lint and fix the errors


vikash:
We should upgrade it to the latest stable version and fix issues. If any new linters are available, consider adding those as well.

xan678:
I would like work on issue.

xan678:
Please approve: https://github.com/gofr-dev/gofr/pull/37

PR #115, opened 2023-12-11
En/upgrade linter

Umang01-hash:
- Closes issue #2


cmd/gofr/entity/entity.go
```diff
@@ -213,7 +215,7 @@ func addConsumer(f fileSystem, projectDirectory, entity string) error {
 		return err
 	}
 
-	filePtr, err := f.OpenFile(entity+".go", os.O_APPEND|os.O_CREATE|os.O_WRONLY, migration.RWMode)
+	filePtr, err := f.OpenFile(entity+goExtension, os.O_APPEND|os.O_CREATE|os.O_WRONLY, migration.RWMode)
```
aryanmehrotra, line 218: Shouldn't we rename it to extensionGo, such that we have more such in future we can follow the same way, and IDE would give suggestions

pkg/datastore/solr.go
```diff
@@ -11,6 +11,8 @@ import (
 	"go.opencensus.io/plugin/ochttp"
 )
 
+const schema, update = "/schema", "/update"
```
aryanmehrotra, line 14: Shouldn't name be much more descriptive?

aryanmehrotra (changes requested):

pkg/datastore/solr.go
vipul-rawat: can be prefixed with `solr` ?

vipul-rawat (dismissed):

cmd/gofr/entity/entity.go
Umang01-hash: Sure will rename that

pkg/datastore/solr.go
Umang01-hash: renamed that to `updateSchema`

aryanmehrotra (dismissed):

aryanmehrotra (approved):

PR #123, opened 2023-12-13
add tests

Umang01-hash:


aryanmehrotra (approved):

pkg/file/file_test.go
```diff
@@ -47,6 +77,64 @@ func TestLocalFileOpen(t *testing.T) {
 	}
 }
 
+func TestOpen_Combined(t *testing.T) {
+	openTestCases := []struct {
+		desc          string
```
srijan-27, line 82: we have added desc for test cases, but we ain't using it anywhere. Should we update this?

pkg/file/file_test.go
```diff
@@ -47,6 +77,64 @@ func TestLocalFileOpen(t *testing.T) {
...
+			desc:          "Error: Unexpected EOF during download but fileMode is Append",
+			fileName:      "bigfile.zip",
+			fileMode:      fetchLocalFileMode(APPEND),
+			mockFetchFunc: func(fd *os.File) error { return io.EOF },
+			expectedError: false,
+			expectedMode:  1090,
+		},
+	}
+
+	for _, tc := range openTestCases {
```
srijan-27, line 115: Also, we are not using the iterator, it can help us know which test case failed. Please take a look, if that should be included.

Umang01-hash: Sure will include iterator too when logging errors

pkg/file/file_test.go
Umang01-hash: My bad. Will update the test method to use desc for logging in error case

aryanmehrotra (approved):

VipulSharma1999:
@Umang01-hash please update the description of the PR adding what's added/changed going forward. It will help others just to get some insight about the PR.

PR #130, opened 2023-12-17
Upgrade go to 1.21.5

### LINKED ISSUE #129: Update Go Version to Go 1.21.5
Current Go version is 1.21.
Has to be update to 1.21.5.

webstradev:
Closes #129


aryanmehrotra:
Hey, you would need to update the go version in code quality step as well.

webstradev:
> Hey, you would need to update the go version in code quality step as well.

I've been looking for that, but I don't see where it is specified

webstradev:
@aryanmehrotra Little bit at a loss with this one.
If I add an explicit step to setup go 1.21.5 than the linting fails with a variety of typecheck erros that don't occur on 1.21.3 however this doesn't happen if I lint locally with 1.21.5

Umang01-hash:
@webstradev can you please check if the environment on your local matches with the one in CI/CD like go-version , linter version etc. Many times these issues comes when there is a version mismatch in local and CI/CD environment

webstradev:
I ran it on my own fork now and I do see the same errors as here, however I'm unsure whether or not I should resolve them in the code, or that something else is going wrong.
I struggle to imagine that this code would break linting bumping from 1.21.3 to 1.21.5. Wouldn't expect that to happen on a patch release.

umang-zs:
@webstradev We definitely to resolve that linter errors in order to merge the PR.

webstradev:
> @webstradev We would need to resolve that linter errors to merge the PR.

I understand that, I'm just uncertain whether or not something else is going wrong because these erros were not here on 1.21.3 and they are here on 1.21.5 and I wouldnt expect a security path of go to suddenly cause this.

webstradev:
> > @webstradev We would need to resolve that linter errors to merge the PR.
> 
> I understand that, I'm just uncertain whether or not something else is going wrong because these erros were not here on 1.21.3 and they are here on 1.21.5 and I wouldnt expect a security path of go to suddenly cause this.

Never mind I have figured it out, the version of golangci-lint we were previously on does not support 1.21. Updating to latest seems to resolve it, will amend my PR

webstradev:
Seems to be running just fine for me locall, really not sure why it is still failing in CI, I am definetly on the right version.
image

Could it be because 1.55.2 doesn't even support go 1.21.5 yet?

Umang01-hash:
> Seems to be running just fine for me locall, really not sure why it is still failing in CI, I am definetly on the right version. image
> 
> Could it be because 1.55.2 doesn't even support go 1.21.5 yet?

Could be the reason , I am still struggling with resources to confirm that but seems this linter is having  some errors with updated Go version or may be vice versa is true.

Read this also: https://github.com/golang/go/issues/64592

PR #137, opened 2023-12-20
Implement MQTT as a publisher subscriber

### LINKED ISSUE #48: Add support for MQTT

vipul-rawat:
MQTT is a standard messaging protocol for the Internet of Things (IoT). It is designed as an extremely lightweight publish/subscribe messaging transport. Read more [here](https://mqtt.org/)

- Added support for MQTT for IoT messaging
- Added test cases
- Added a step in GitHub pipeline to run mosquitto server for MQTT

closes #48


pkg/datastore/pubsub/mqtt/mqtt.go
```diff
@@ -0,0 +1,193 @@
...
+		logger.Errorf("Connection lost: %v", err)
+	}
+
+	options.ConnectRetryInterval = time.Second * time.Duration(config.ConnectionRetryDuration)
+
+	// create the client using the options above
+	client := mqtt.NewClient(options)
+
+	if token := client.Connect(); token.Wait() && token.Error() != nil {
+		logger.Errorf("cannot connect to MQTT, HostName : %v, Port : %v, error : %v", config.Topic, config.Port, token.Error())
```
VipulSharma1999, line 69: In this error log, config.Topic is used for the value of HostName, isn't it should be Topic?

pkg/datastore/pubsub/mqtt/mqtt.go
```diff
@@ -0,0 +1,193 @@
...
+		return &MQTT{config: config, logger: logger}, token.Error()
+	}
+
+	logger.Debugf("connected to MQTT, HostName : %v, Port : %v", config.Topic, config.Port)
+
+	return &MQTT{config: config, logger: logger, Client: client}, nil
+}
+
+func (m *MQTT) PublishEvent(_ string, value interface{}, _ map[string]string) error {
+	if m.Client == nil {
```
VipulSharma1999, line 80: Metrics are not integrated in this method.

pkg/datastore/pubsub/mqtt/mqtt.go
```diff
@@ -0,0 +1,193 @@
...
+	token := m.Client.Subscribe(m.config.Topic, m.config.QoS, handler)
+
+	if token.Wait() && token.Error() != nil {
+		// increment failure count for failed subscribing
+		pubsub.SubscribeFailureCount(m.config.Topic, "")
+		return nil, token.Error()
+	}
+
+	// increment success counter for successful subscribing
+	pubsub.PublishSuccessCount(m.config.Topic, "")
```
VipulSharma1999, line 126: this metric should be SubscribeSuccessCount as it's for successful subscription.

VipulSharma1999 (changes requested):

pkg/datastore/pubsub/mqtt/mqtt.go
```diff
@@ -0,0 +1,193 @@
...
+	QoS                     byte
+	Order                   bool
+	ConnectionRetryDuration int
+}
+
+// New establishes connection to Kafka using the config provided in KafkaConfig
+func New(config *Config, logger log.Logger) (pubsub.PublisherSubscriber, error) {
+	options := mqtt.NewClientOptions()
+	options.AddBroker("tcp://" + config.Hostname + ":" + strconv.Itoa(config.Port))
+	options.SetClientID(config.ClientID)
```
VipulSharma1999, line 41: if this clientID is not passed in the configs, it should be considered something specific to the framework i.e `gofr_mqtt_client`

pkg/datastore/pubsub/mqtt/mqtt.go
```diff
@@ -0,0 +1,193 @@
...
+	"gofr.dev/pkg/datastore/pubsub"
+	"gofr.dev/pkg/errors"
+	"gofr.dev/pkg/gofr/types"
+	"gofr.dev/pkg/log"
+
+	mqtt "github.com/eclipse/paho.mqtt.golang"
+)
+
+type MQTT struct {
+	Client mqtt.Client
```
VipulSharma1999, line 19: why is the client exported?

VipulSharma1999 (changes requested):

pkg/datastore/pubsub/mqtt/mqtt.go
vipul-rawat: This would be a user-managed thing, as `client_id` is required to maintain the uniqueness of the subscribers and no two subscribers are to be using the same client_id, or else that would lead to unexpected behavior. 
Wouldn't this be at the user's discretion to maintain the `client_id`?
> The ClientId is a unique identifier that distinguishes each MQTT client connecting to a broker and enables the broker to keep track of the client's current state

VipulSharma1999: @vipul-rawat - I meant that for a single app(client) this will be unique and should work even after not passing this client_id via configs and can be added by default by the framework in that case.
Though, after going through some examples, I see different clientIDs are passed even for the publisher and subscriber clients connecting to the same broker essentially means we can't have same clientIDs even for the publisher/subscriber. 
Issue: https://github.com/eclipse/paho.mqtt.golang/issues/426

What we have to do here is:  
If no ClientID is passed then for 
publisher - create it with `topicName`+ "pub" +"mqtt_client"
consumer - create it with `topicName`+ "sub" +"mqtt_client".

Let me know your thoughts.

pkg/datastore/pubsub/mqtt/mqtt.go
```diff
@@ -0,0 +1,202 @@
...
+	QoS                     byte
+	Order                   bool
+	ConnectionRetryDuration int
+}
+
+// New establishes connection to Kafka using the config provided in KafkaConfig
+func New(config *Config, logger log.Logger) (pubsub.PublisherSubscriber, error) {
+	pubsub.RegisterMetrics()
+
+	options := mqtt.NewClientOptions()
```
VipulSharma1999, line 40: there are some configs related to the client connection:
`ConnectRetry`, `AutoReconnect`.
Can you kindly check the criticality of these and see if we need these?

vipul-rawat: We are doing manual retry connection like other databases so I think this might not be required

pkg/datastore/pubsub/mqtt/mqtt.go
vipul-rawat: There is no issue with using a same clientId fro publish and subscribe with the same client, but will add a default clientId in case it is not provided!

PR #141, opened 2023-12-22
Add migration support for clickhouse | Enhance example using migrations

### LINKED ISSUE #124: Add migrations support for clickhouse DB

stella551:
- Added migrations support for clickhouse.
- Enhanced using-clickhouse example using migrations.

closes #124


cmd/gofr/migration/dbMigration/clickhouse.go
```diff
@@ -0,0 +1,179 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (c *Clickhouse) preRun(app, method, name string) error {
+	const migrationTableSchema = `CREATE TABLE IF NOT EXISTS gofr_migrations (
+    app String,
+    version Int64,
```
vipul-rawat, line 58: can we use an Unsigned integer? Since we are storing unix timestamps as versions

PR #195, opened 2024-01-09
Add tests for Logger , Service pkg

Umang01-hash:
- Add tests for the `Logger` pkg


pkg/gofr/logging/level_test.go
```diff
@@ -0,0 +1,95 @@
...
+		{FATAL, 160},
+		{WARN, 220},
+		{NOTICE, 220},
+		{INFO, 6},
+		{DEBUG, 8},
+		{Level(99), 37}, // Test default case
+	}
+
+	for i, tc := range tests {
+		assert.Equal(t, tc.expectedColor, tc.level.color(), "TEST[%d], Failed.\n", i)
```
aryanmehrotra, line 43: \n is not needed

aryanmehrotra (changes requested):

pkg/gofr/logging/level_test.go
umang-zs: removed

PR #197, opened 2024-01-09
Add tests for db, gofr, grpc, handler and httpServer

vipul-rawat:


pkg/gofr/container/db.go
```diff
@@ -3,7 +3,7 @@ package container
 import (
 	"context"
 	"database/sql"
-	"fmt"
+	"gofr.dev/pkg/gofr/logging"
```
aryanmehrotra, line 6: should be at last

pkg/gofr/httpServer_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// Send a GET request to the server
+	resp, err := http.Get("http://localhost:8080")
+	if err != nil {
+		t.Errorf("Failed to send GET request: %v", err)
+	}
+
+	// Check if the server is listening
+	if resp.StatusCode != http.StatusOK {
+		t.Errorf("Expected status code %d, but got %d", http.StatusOK, resp.StatusCode)
```
aryanmehrotra, line 45: assert can be used

pkg/gofr/handler_test.go
```diff
@@ -0,0 +1,79 @@
+package gofr
```
aryanmehrotra, line 1: test desc should be added in all assert statements
"TEST[%d], Failed.\n%s", i, tc.desc

pkg/gofr/handler_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"context"
+	"errors"
+	"github.com/stretchr/testify/assert"
+	"gofr.dev/pkg/gofr/container"
+	"gofr.dev/pkg/gofr/http/response"
+	"gofr.dev/pkg/gofr/logging"
+	"net/http"
+	"net/http/httptest"
```
aryanmehrotra, line 11: sort imports

aryanmehrotra (changes requested):

pkg/gofr/gofr.go
```diff
@@ -39,12 +39,12 @@ type App struct {
 
 // RegisterService adds a grpc service to the gofr application.
 func (a *App) RegisterService(desc *grpc.ServiceDesc, impl interface{}) {
-	a.grpcRegistered = true
 	a.container.Logger.Infof("Registering GRPC Server: %s", desc.ServiceName)
 	a.grpcServer.server.RegisterService(desc, impl)
+	a.grpcRegistered = true
```
aryanmehrotra, line 44: why this change?

pkg/gofr/gofr_test.go
```diff
@@ -14,3 +19,79 @@ func TestNewCMD(t *testing.T) {
...
+
+		if res.Data != tc.response {
+			t.Errorf("Unexpected response for %s %s. \t expected: %s \t got: %s", tc.method, tc.target, tc.response, res.Data)
+		}
+
+		if ctype := w.Header().Get(tc.headerKey); ctype != tc.headerVal {
+			t.Errorf("Header mismatch for %s %s. \t expected: %s \t got: %s", tc.method, tc.target, tc.headerVal, ctype)
+		}
+	}
+}
```
aryanmehrotra, line 93: assert?

aryanmehrotra (changes requested):

pkg/gofr/gofr.go
vipul-rawat: was suggested, since we should set it to be true after it is actually registered!

aryanmehrotra (approved):

Issue #198, opened 2024-01-09
Pipeline Failing


aryanmehrotra:
One of the example or package is failing when package and example tests are running together.
Running either of them after one passes is passed succefuly.
One Proposed solution is to make them run after one another or merge into one.

JigarJoshi04:
I can work on this issue.
Checking if I can fix this.
Taking your proposal to separate the two.
Let me check and will keep this issue updated.
Thanks :)

JigarJoshi04:
Hi @aryanmehrotra ,
Please assign this issue to me.
I have raised the MR.
Please review the request. 
Thanks :)

JigarJoshi04:
Hi @aryanmehrotra ,
I am facing a broken test case issue https://github.com/JigarJoshi04/gofr/pull/2
Here if you see the workflows, I am just running the package unit test cases, nothing else
Still this is failing.

Can you please help me here?

aryanmehrotra:
> Hi @aryanmehrotra , I am facing a broken test case issue [JigarJoshi04#2](https://github.com/JigarJoshi04/gofr/pull/2) Here if you see the workflows, I am just running the package unit test cases, nothing else Still this is failing.
> 
> Can you please help me here?

Hey - it would require to fix the tests which are panicking due to nil pointers.

PR #200, opened 2024-01-09
(DRAFT) Fix: failing pipeline fixed

JigarJoshi04:
Fixes a pipeline issue where concurrent execution of package and example unit tests resulted in failures. The tests were passing when run independently. This commit addresses the concurrency problem by consolidating both test jobs into a single unit-testing job. The changes ensure a more stable and reliable CI pipeline.


JigarJoshi04:
The issue https://github.com/gofr-dev/gofr/issues/198
Should be addressed by this Merge request

aryanmehrotra:
> I have identified few issues. Please consider this as a draft. Let me fix the issues first

Also find a way if we can run it concurrently such that the overall time is reduced.
Moving forward there will be lot more test and examples running sequentially will take more time.

JigarJoshi04:
Sure let me check this once

PR #202, opened 2024-01-10
[FIX] go.mod and linters

srijan-27:
- Updated go.mod and go.sum
- Fixed linter errors


pkg/gofr/container/redis.go
```diff
@@ -3,8 +3,8 @@ package container
 import (
 	"context"
 	"fmt"
-	"github.com/redis/go-redis/extra/redisotel/v9"
 
+	"github.com/redis/go-redis/extra/redisotel/v9"
```
vipul-rawat, line 7: give imports name

vipul-rawat (approved):

PR #203, opened 2024-01-10
Add tests for gofr.go

vipul-rawat:
- ServerHealthHandlerAddCheck - tests if health-check at `/.well-known/health` is added by default
- ServerNotRunningForNoRoutes - tests if server is not started if no routes are added


pkg/gofr/gofr_test.go
```diff
@@ -130,3 +132,72 @@ func TestGofr_ServerRun(t *testing.T) {
...
+		return
+	}
+
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+	assert.Equal(t, response.Raw{Data: "OK"}, respn)
+
+	resp.Body.Close()
+}
+
+func TestGofr_ServerNotRunningForNoRoutes(t *testing.T) {
```
vikash, line 181: I think this does not really confirms if the server is running or not. Even if g.Run call is delayed - it will still pass, even though server will eventually run.

vipul-rawat: So we just need to add a delay to let the g.Run() call be processed and then, check if the HTTP request passes through!

vikash: How do we know for sure how much time? I do not have solution, just asking. Can we check how others choose to test this scenario?

pkg/gofr/gofr_test.go
```diff
@@ -130,3 +132,72 @@ func TestGofr_ServerRun(t *testing.T) {
...
+
+	// make response bytes slice with content-length
+	responseBytes := make([]byte, resp.ContentLength)
+
+	// read the response into responseBytes
+	_, err = resp.Body.Read(responseBytes)
+	if err != nil && !errors.Is(err, io.EOF) {
+		t.Errorf("TEST failed: %v", err)
+
+		return
```
srijan-27, line 165: we can use `t.Fatalf` instead of `t.Errorf` followed by a return statement.

vipul-rawat:
I could not find a feasible way to ensure that the server would run while running the test. Maybe I will get back to this in the future.

PR #207, opened 2024-01-11
[EN] Add tests for config, requests

srijan-27:
- Fixed tests for cmd/request and config
- Added tests for http/request


pkg/gofr/cmd/request_test.go
```diff
@@ -28,17 +28,25 @@ func TestRequest_Bind(t *testing.T) {
...
+	r := NewRequest([]string{"-name=gofr", "-valid=true", "-value=12", "-test"})
+
+	testCases := []struct {
+		param string
+		want  string
+	}{
+		{"name", "gofr"},
+		{"valid", "true"},
+		{"value", "12"},
+		{"test", "true"},
```
aryanmehrotra, line 40: case can also be added about behaviour when we have keys with special characters and empty keys, same for values as well.

pkg/gofr/config/godotenv_test.go
```diff
@@ -13,36 +14,72 @@ func TestNewEnvFile(t *testing.T) {
 }
 
 func TestNewGoDotEnvProvider(t *testing.T) {
-	f := new(EnvFile)
+	var (
+		configs = "TEST=test\nNAME=gofr\nKEY_123=value123"
```
aryanmehrotra, line 18: same can be tested here about special characters and different possible keys and values

aryanmehrotra (changes requested):

srijan-27:
Deleting this for now.

PR #210, opened 2024-01-11
Add HTTP Call methods in `new.go` for interservice calls with Logs

aryanmehrotra:
image


pkg/gofr/service/new.go
```diff
@@ -1,52 +1,96 @@
...
+	PutWithHeaders(ctx context.Context, api string, queryParams map[string]interface{}, body []byte, headers map[string]string) (*http.Response, error)
 
-type HTTP interface {
-	Get(ctx context.Context, api string, params map[string]interface{}) (*http.Response, error)
+	Delete(ctx context.Context, api string, body []byte) (*http.Response, error)
+	DeleteWithHeaders(ctx context.Context, api string, body []byte, headers map[string]string) (*http.Response, error)
 }
 
-func NewHTTPService(serviceAddress string) HTTP {
+func NewHTTPService(serviceAddress string, logger logging.Logger) HTTP {
```
vikash, line 37: This is breaking contract and the code example will not work with this. Also, because logger should be automatically taken care by the framework, one should not need to pass a new logger for using a service.

pkg/gofr/cmd/request.go
```diff
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"go.opentelemetry.io/otel/trace"
```
Umang01-hash, line 5: Sort the imports please

pkg/gofr/container/container.go
```diff
@@ -1,6 +1,7 @@
 package container
 
 import (
+	"gofr.dev/pkg/gofr/service"
```
Umang01-hash, line 4: Sort the imports please

pkg/gofr/http/request.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"gofr.dev/pkg/gofr/http/middleware"
```
Umang01-hash, line 8: Please sort the imports

pkg/gofr/http/middleware/logger.go
```diff
@@ -2,6 +2,8 @@ package middleware
 
 import (
 	"encoding/json"
+	"fmt"
```
Umang01-hash, line 5: Please sort the imports

pkg/gofr/gofr.go
```diff
@@ -2,6 +2,7 @@ package gofr
 
 import (
 	"fmt"
+	"gofr.dev/pkg/gofr/service"
```
Umang01-hash, line 5: Please sort the imports

pkg/gofr/http/middleware/logger.go
```diff
@@ -112,3 +114,38 @@ func panicRecovery(w http.ResponseWriter, logger logger) {
 		_ = json.NewEncoder(w).Encode(res)
 	}
 }
+
+func GetCorrelationID(r *http.Request) string {
```
Umang01-hash, line 118: Instead of modifying the request header directly, can we return (string,error) from GetCorrelationID and set the headers where it is being called.. ?
Or think of moving it to some other location where it can access the request easily.

pkg/gofr/http/middleware/logger.go
```diff
@@ -112,3 +114,38 @@ func panicRecovery(w http.ResponseWriter, logger logger) {
...
+	var correlationID string
+
+	cID, err := trace.TraceIDFromHex(getTraceID(r))
```
Umang01-hash, line 121: `cID` can be renamed here to more meaningful variable name..

pkg/gofr/http/middleware/logger.go
```diff
@@ -112,3 +114,38 @@ func panicRecovery(w http.ResponseWriter, logger logger) {
...
+	if err != nil {
+		correlationID = trace.SpanFromContext(r.Context()).SpanContext().TraceID().String()
+		// if tracing is not enabled, otel sets the trace-ID to "00000000000000000000000000000000" (nil type of [16]byte)
+
+		const correlationIDLength = 32
+
+		nullCorrelationID := fmt.Sprintf("%0*s", correlationIDLength, "")
+
+		if correlationID == nullCorrelationID {
```
Umang01-hash, line 130: Can we use below code for the logic from line **126-130**:
```go
if strings.HasPrefix(correlationID, "00000000000000000000000000000000")
```

Umang01-hash (changes requested):

pkg/gofr/service/new.go
```diff
@@ -1,52 +1,140 @@
...
+func (h *httpService) createAndSendRequest(ctx context.Context, method string, path string,
+	queryParams map[string]interface{}, body []byte, headers map[string]string) (*http.Response, error) {
+	uri := h.url + "/" + path
+	uri = strings.TrimRight(uri, "/")
+
+	spanContext, span := h.Tracer.Start(ctx, uri)
+	defer span.End()
+
+	spanContext = httptrace.WithClientTrace(spanContext, otelhttptrace.NewClientTrace(ctx))
+	req, _ := http.NewRequestWithContext(spanContext, method, uri, bytes.NewBuffer(body))
```
Umang01-hash, line 97: Please handle the error here

pkg/gofr/gofr.go
```diff
@@ -134,6 +135,19 @@ func (a *App) readConfig() {
 	a.Config = config.NewEnvFile(configLocation)
 }
 
+// GET adds a Handler for http GET method for a route pattern.
```
Umang01-hash, line 138: This comment doesn't seem suitable here. Please change it

Umang01-hash (changes requested):
Please add relavant comment wherever necessary

Umang01-hash (approved):

srijan-27 (approved):

PR #230, opened 2024-01-23
Fix the trace tree and add feature for inter service trace propagation

vipul-rawat:
image

### DIFF
--- a/pkg/gofr/http/middleware/tracer.go
+++ b/pkg/gofr/http/middleware/tracer.go
@@ -19,6 +19,7 @@ func Tracer(inner http.Handler) http.Handler {
 		// its parentID, thus connecting two spans
 		ctx = otel.GetTextMapPropagator().Extract(ctx, propagation.HeaderCarrier(r.Header))
 
+		// TODO - inject version
 		tr := otel.GetTracerProvider().Tracer("gofr")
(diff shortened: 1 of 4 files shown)


pkg/gofr/http/middleware/tracer.go
```diff
@@ -6,14 +6,19 @@ import (
...
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
 )
 
 func Tracer(inner http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		// Start context and Tracing
 		ctx := r.Context()
 
-		// TODO - version has to be injected
```
aryanmehrotra, line 16: This has to be added back

pkg/gofr/service/new.go
```diff
@@ -55,7 +55,7 @@ type HTTP interface {
 
 func NewHTTPService(serviceAddress string, logger Logger) HTTP {
 	return &httpService{
-		Client: &http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport)},
+		Client: &http.Client{},
```
aryanmehrotra, line 58: lets mention the idea behind not using otel transport layer

aryanmehrotra (changes requested):

PR #242, opened 2024-02-05
Remote Logging

Umang01-hash:


srijan-27 (approved):

pkg/gofr/logging/dynamicLevelLogger.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func fetchAndUpdateLogLevel(remoteService service.HTTP) (Level, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), requestTimeout) // Set timeout for 5 seconds
+	defer cancel()
+
+	resp, err := remoteService.Get(ctx, "", nil)
+	if err != nil {
+		return INFO, err
+	}
```
vikash, line 63: If remote service does not report anything because of error, we should not be returning "INFO". Because if the previous level was anything else (like WARN), we will end up changing to INFO which is not the intention - as the long level should not change in such case. @Umang01-hash

Umang01-hash: Sure @vikash. Understood your point. We can return the current log Level in such case. Will do the alteration

PR #403, opened 2024-03-25
En/add env support for diff environments.

Umang01-hash:
- Add support for reading env files from `GOFR_ENV` config's value if set.
- Add support for reading `.local.env` if `GOFR_ENV` is not set.


pkg/gofr/gofr.go
```diff
@@ -164,13 +164,19 @@ func (a *App) Run() {
 }
 
 // readConfig reads the configuration from the default location.
-func (a *App) readConfig() {
+func (a *App) readConfig(isAppCMD bool) {
```
Umang01-hash, line 167: Needed to have a new flag `isAppCMD` because we want a **FileLogger** in case of a CMD app and a **normal logger i**n case of a normal app.

pkg/gofr/logging/logger.go
```diff
@@ -241,7 +241,7 @@ func NewFileLogger(path string) Logger {
 		errorOut:  io.Discard,
 	}
 
-	if path != "" {
+	if path == "" {
```
Umang01-hash, line 244: If path is not provided we should return from here

pkg/gofr/config/mock_logger.go
```diff
@@ -0,0 +1,107 @@
+package config
```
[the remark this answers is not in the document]
vipul-rawat: Cyclic dependencies can be removed by declaring Config interface in mock_config.go since it is using it.

PR #419, opened 2024-04-03
Remove error for .local.env

### LINKED ISSUE #416: Missing configs/.local.env should not throw a warning
.local.env file is required only if we want to override something locally while not intending to commit it. This should not be a required if the overriding behaviour is not desired. So, the WARN level log is unnecessary in this case.

rokerzfirst101:
**Description:**

-   Moved `defaultFileName` and `defaultOverrideFileName` to const block
-  Refactored the logic to not throw an error for .local.env unless specified in `APP_ENV`

Closes #416

### DIFF
--- a/pkg/gofr/config/godotenv.go
+++ b/pkg/gofr/config/godotenv.go
@@ -44,9 +44,11 @@ func (e *EnvLoader) read(folder string) {
 	// If 'APP_ENV' is not set , GoFr will first try to read '.local.env', if failed then it
 	// will read default env file.
 	err := godotenv.Load(overrideFile)
-	if err != nil && env != "" {
-		e.logger.Warnf("Failed to load config from file: %v, Err: %v", overrideFile, err)
-	} else if err == nil {
+	if err != nil {
+		if env != "" {
+			e.logger.Warnf("Failed to load config from file: %v, Err: %v", overrideFile, err)
+		}
+	} else {
 		e.logger.Infof("Loaded config from file: %v", overrideFile)
 
 		return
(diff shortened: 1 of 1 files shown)


vipul-rawat:
Please mention the issue while creating the pull request.

rokerzfirst101:
@vipul-rawat I've added tagged the issue in the PR description. Let me know if anything else requires changing. 😄

pkg/gofr/config/godotenv.go
```diff
@@ -24,30 +29,30 @@ func NewEnvFile(configFolder string, logger logger) *EnvLoader {
...
 	// If 'APP_ENV' is set to x, then GoFr will try to read '.x.env' file from configs directory, if failed it will read
 	// the default file '.env'.
 	// If 'APP_ENV' is not set , GoFr will first try to read '.local.env', if failed then it
 	// will read default env file.
-	if err := godotenv.Load(overrideFile); err != nil {
+	err := godotenv.Load(overrideFile)
+	if err != nil && env != "" {
 		e.logger.Warnf("Failed to load config from file: %v, Err: %v", overrideFile, err)
-	} else {
+	} else if err == nil {
```
vipul-rawat, line 49: The `err == nil` can be simplified further if the preceding error check is broken down!
I believe that it would make these checks more readable and easy to understand.

rokerzfirst101: sure, will make the change

rokerzfirst101:
@vipul-rawat can you take a look at the changes

vipul-rawat (approved):

PR #445, opened 2024-04-08
Add support for sqlite

### LINKED ISSUE #423: Support sqlite as DB_DIALECT

rokerzfirst101:
**Description:**
- Added `sqlite3` as an acceptable value for `DB_DIALECT`
- Add import for `github.com/mattn/go-sqlite3` driver
- Updated the logic to initialize the connection string using only`DB_HOST` for SQLite
- The server/app needs to be built with `CGO_ENABLED=1` in order for the SQLite driver to work

**How it works:**

To use SQLite as the Database, the user would need to provide the following configs or environment variables:
```env
DB_DIALECT=sqlite3
DB_HOST={connection-string}
```

Connection String Ref: https://github.com/mattn/go-sqlite3?tab=readme-ov-file#connection-string

Closes #423


Umang01-hash:
@rokerzfirst101 Can you please also update the Test function `TestSQL_getDBConnectionString`and add the test case for newly added code.

rokerzfirst101:
sure

vipul-rawat:
@rokerzfirst101 - please add/extend documentation for using Sqlite in gofr - https://gofr.dev/docs/advanced-guide/dealing-with-datasources

rokerzfirst101:
@vipul-rawat I've updated the database documentation

vikash:
Adding a requirement for cgo reduces compatibility. Many people do prefer to use go without cgo. There is an implementation which does not require cgo, we should prefer that.

vikash (changes requested):
Adding a requirement for cgo reduces compatibility. Many people do prefer to use go without cgo. There is an implementation which does not require cgo, we should prefer that

pkg/gofr/migration/sql.go
```diff
@@ -21,8 +22,6 @@ const (
     constraint primary_key primary key (version, method)
 );`
 
-	checkSQLGoFrMigrationsTable = `SELECT EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = 'gofr_migrations');`
-
```
rokerzfirst101, line 25: this statement will not work when there are multiple databases with the `gofr_migrations` table

rokerzfirst101:
I've replaced the `github.com/mattn/go-sqlite3` pkg with `modernc.org/sqlite`

aryanmehrotra:
@rokerzfirst101 Rather than taking connection string from the user, we can take the configs in the following way and it keep consistent across for all SQL databases which we currently support
```
DB_HOST=localhost #Not possible to add? does it support connecting to a remote file in that case it can be supported
DB_USER=root # can be used
DB_PASSWORD=password # can be used
DB_NAME=test #db_name?
DB_DIALECT=mysql #sqlite will be used here
```

rokerzfirst101:
closing as stale, will add another one

Issue #446, opened 2024-04-08
wrong status codes for POST and DELETE responses


Umang01-hash:
GoFr gives status code `200` for all success responses, whereas we should be having appropriate status code according to diff request for ex **201** for `POST` and  **204** for `DELETE`

PR #452, opened 2024-04-09
Fix status code for response of Post and Delete requests

ssshekhu53:
- Added types `PostResponder` and `DeleteResponder` implementing Responder interface to write responses for post and delete requests respectively
- Updated `ServeHTTP` method to create instances of `Responder` on the basis of request methods
- For details refer https://github.com/gofr-dev/gofr/issues/446

### DIFF
--- a/pkg/gofr/handler_test.go
+++ b/pkg/gofr/handler_test.go
@@ -28,11 +28,16 @@ func TestHandler_ServeHTTP(t *testing.T) {
 		data       interface{}
 		err        error
 		statusCode int
+		body       string
 	}{
-		{"method is get, data is nil and error is nil", http.MethodGet, nil, nil, http.StatusOK},
-		{"method is get, data is mil, error is not nil", http.MethodGet, nil, errTest, http.StatusInternalServerError},
-		{"method is post, data is nil and error is nil", http.MethodPost, nil, nil, http.StatusCreated},
-		{"method is delete, data is nil and error is nil", http.MethodDelete, nil, nil, http.StatusNoContent},
+		{"method is get, data is nil and error is nil", http.MethodGet, nil, nil, http.StatusOK,
+			`{}`},
+		{"method is get, data is mil, error is not nil", http.MethodGet, nil, errTest, http.StatusInternalServerError,
+			`{"error":{"message":"some error"}}`},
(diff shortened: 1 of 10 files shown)


srijan-27:
@ssshekhu53 - It is not advisable to create new duplicate responders just for the sake of adding statusCodes in response.
We would need to look into a better approach to resolve this.

shashank-zopsmart:
> @ssshekhu53 - It is not advisable to create new duplicate responders just for the sake of adding statusCodes in response. We would need to look into a better approach to resolve this.

Have removed the new responders and added a field `method` in responder struct. Factory function for responder has been updated too to accept ResponseWriter and method in the parameter.  `HTTPStatusFromRequestMethod` will deduce the status code on the basis of http method for success cases.

pkg/gofr/http/responder.go
```diff
@@ -65,6 +70,22 @@ func (r Responder) HTTPStatusFromError(err error) (status int, errObj interface{
 	}
 }
 
+// HTTPStatusFromRequestMethod maps request method to HTTP status codes.
+func (r Responder) HTTPStatusFromRequestMethod() int {
```
srijan-27, line 74: This method is unnecessary addition. Could be included in `HTTPStatusFromErr` method itself.

srijan-27 (changes requested):

pkg/gofr/handler_test.go
```diff
@@ -24,17 +24,20 @@ var (
...
 		data       interface{}
 		err        error
 		statusCode int
 	}{
-		{"data is nil and error is nil", nil, nil, http.StatusOK},
-		{"data is mil, error is not nil", nil, errTest, http.StatusInternalServerError},
+		{"method is get, data is nil and error is nil", http.MethodGet, nil, nil, http.StatusOK},
+		{"method is get, data is mil, error is not nil", http.MethodGet, nil, errTest, http.StatusInternalServerError},
+		{"method is post, data is nil and error is nil", http.MethodPost, nil, nil, http.StatusCreated},
+		{"method is delete, data is nil and error is nil", http.MethodDelete, nil, nil, http.StatusNoContent},
```
srijan-27, line 35: Since, data is nil for every case, so either we can include tests with data not being nil or we can remove the field and assume it `nil` by default.

srijan-27 (approved):

vipul-rawat (approved):

Issue #459, opened 2024-04-13
Support  to filter and mask PII data from logs at the framework level


SUKANT007:
Due to privacy concerns, many times we have to mask PII data from logs, rather than each service writing their own filter and masking libraries, This could be supported at the framework level.

**Expected Behaviour:** 

 - Gofr logger supports filter interface which can be implemented.
 - Default filter interface implementation by the framework.
 - Support struct tags for to filter particular fields from the struct while logging.

Attaching references where this is implemented: 

 - https://go-kratos.dev/en/docs/component/log/#filter-log-filtering
 - https://github.com/m-mizutani/masq

srinivas365:
@srijan-27 Please assign this to me.

srijan-27:
Sure @srinivas365 - assigning this to you.

srijan-27:
@srinivas365 - Hi, just wanted to know if you are working on it? Pls let me know if you are stuck or need any help.

srinivas365:
@srijan-27 got little busy with projects. unassigning myself.

PR #465, opened 2024-04-15
Added Masking of PII data in logs

SuyeshBadge:
## Pull Request Template
Added feature to mask PII data into logs.

**Description:**

- This PR introduces a new feature to mask Personally Identifiable Information (PII) fields in log entries. The `MaskingFilter` in the `logging` package has been updated to support masking of various PII fields, such as name, email, phone number, Social Security number, credit card number, date of birth, biometric data, and IP address.
- The motivation behind this change is to enhance data privacy and security by preventing sensitive user information from being logged in plaintext.
- This feature will benefit users by ensuring compliance with privacy regulations and reducing the risk of data breaches or unauthorized access to sensitive information.
- Closes: https://github.com/gofr-dev/gofr/issues/459

- There are no breaking changes introduced by this PR.

**Additional Information:**

- The changes rely on the built-in `reflect` package in Go for inspecting and modifying struct fields.
- New test cases have been added to ensure comprehensive testing of the PII masking functionality.

**Checklist:**


SuyeshBadge:
@aryanmehrotra, @vipul-rawat, and @Umang01-hash,

I'm initiating a discussion thread to commence the implementation of the mentioned feature. Let's delve into the various potential changes and methodologies to effectively incorporate this feature.

Looking forward to your valuable insights and contributions.

pkg/gofr/logging/logger.go
```diff
@@ -223,24 +297,69 @@ func colorForGRPCCode(status int32) int {
 }
 
 // NewLogger creates a new logger instance with the specified logging level.
-func NewLogger(level Level) Logger {
+func NewLogger(level Level, args ...interface{}) Logger {
+	var filter Filter
+	if len(args) > 0 {
+		filter = args[0].(Filter)
```
vipul-rawat, line 303: This would panic if the args provided do not implement `Filter` interface.
We can do a boolean(ok) check instead.

pkg/gofr/logging/logger.go
```diff
@@ -223,24 +297,69 @@ func colorForGRPCCode(status int32) int {
...
 	return l
 }
 
 // NewFileLogger creates a new logger instance with logging to a file.
-func NewFileLogger(path string) Logger {
+func NewFileLogger(path string, args ...interface{}) Logger {
+
+	var filter Filter
+	if len(args) > 0 {
+		filter = args[0].(Filter)
```
vipul-rawat, line 339: This would panic if the args provided do not implement `Filter` interface.
We can do a boolean(ok) check instead.

pkg/gofr/logging/logger_test.go
```diff
@@ -387,3 +387,78 @@ func Test_clean(t *testing.T) {
...
+				PhoneNumber:          1234567890,
+				SocialSecurityNumber: 123456789,
+				CreditCardNumber:     1234567890123456,
+				DateOfBirth:          19800101,
+				BiometricData:        123.456,
+				IPAddress:            "192.168.1.1",
+			},
+			expected: `{"phoneNumber":0,"socialSecurityNumber":0,"creditCardNumber":0,"dateOfBirth":0,"biometricData":0,"ipAddress":"***********"}`,
+		},
+		// Add more test cases as needed
```
vipul-rawat, line 429: Empty comment

pkg/gofr/logging/logger.go
```diff
@@ -41,11 +42,81 @@ type Logger interface {
...
+		}
+	}
+
+	// Update the original message with the modified value
+	entry.Message = newVal.Interface()
+
+	return entry
+}
+
+func contains(slice []string, str string) bool {
```
vipul-rawat, line 100: naming convention could be better for `slice`, not informative of the usage

vipul-rawat (commented):
Please check the linters, pipeline is failing.

Also is there a way to not use reflect, if possible?

You might also need to look for a way for users to the masking fields at the time of application creation! Or if masking is required at all!

SuyeshBadge:
Hey @vipul-rawat, I've pushed the requested changes along with the formatting and linting fixes. Please have a look once you get a chance and let me know if anything is missing.

vipul-rawat (changes requested):
There are still 2 things - 
1. Can you add an example on how enabling and disabling the filtered logging would look from user's perspective
2. I am still not sure if this would handle the pointers

SuyeshBadge:
Certainly, @vipul-rawat will make the changes and update you accordingly.

SuyeshBadge:
> I am still not sure if this would handle the pointers

You're right to be concerned about handling pointers. The current implementation assumes that the fields are direct values and not pointers. Will update this too.

SuyeshBadge:
Hey @vipul-rawat ,
I've updated the filtered logging feature to support pointer values and added documentation on how to use it. However, one test check is failing, and I'm not sure why.
Could you please take a look and provide some guidance on fixing it?
Thanks for your help!

vipul-rawat:
@SuyeshBadge This implementation requires users to be aware of the logger and define it themselves, whereas for most use cases the framework is responsible for managing the logger itself. Can we use configs reading to achieve the same instead of the user defining the filter in the main file?

Also, how can a user-defined filter be called `DefaultFilter`? Should be CustomFilter or equivalent.

The pipeline was failing due to an MQTT test! Probably the public server was experiencing some downtime. Rerunning should fix the same. Also, update the local with development.

SuyeshBadge:
Hello @vipul-rawat,

Could you kindly provide further clarification regarding the utilization of configuration readings in this context? Specifically, would you prefer me to:

1. Pass the configuration explicitly to the `NewLogger` function?
2. Expose a `setConfig` method that will be invoked prior to declaring the logger?
3. Directly invoke `godotenv` from the `logger.go` file?

I require additional details to proceed effectively. Thank you.

vipul-rawat:
@SuyeshBadge What I mean is that currently according to your implementation, the NewLogger is being made by the user in main, but what we want is that the filter gets configured automatically during the app creation (during `gofr.New()`), so for example if user gives the config `MASKED_FIELDS=<comma separated field names>` then what framework does is that it adds the filter with those names, we don't want to have the filtering of fields by default. 

I think something like `SetFilter` similar to `setConfig` that is suggested in point 2 seems good.

> Point 1 will not be viable as it would be a breaking change and not worth it.
> Point 3 also not viable as logger's responsibility is not to read configurations or load them.

@srijan-27 FYA

SuyeshBadge:
Hey @vipul-rawat,

I've implemented the changes as per your instructions. Additionally, I addressed several linting issues that were present. Kindly review the updates at your convenience. With the latest implementation, the application now reads values from the configuration and configures the logger accordingly. Furthermore, I've revised the documentation for improved clarity and understanding.

SuyeshBadge:
Also if possible could you please re-run the checks, some of them are failing.

docs/advanced-guide/filtered-logging/page.md
```diff
@@ -0,0 +1,72 @@
+# Filtered Logging
+
+Filtered logging is a feature that allows you to mask sensitive data in your log messages. It helps protect personally identifiable information (PII) and other confidential data from being exposed in logs.
+
+## Usage
+
+To use filtered logging, you need to set the masking configuration in your `.env` file. The following configuration options are available:
+
+- `LOGGER_MASKING_ENABLED`: Set to `true` to enable masking of sensitive fields in log messages. Default is `false`.
+- `LOGGER_MASKING_FIELDS`: A comma-separated list of field names that should be masked in log messages.
```
srijan-27, line 10: @SuyeshBadge - This use-case can be fulfilled with a single config, not required to create a bool config as well. Please fix this.

docs/advanced-guide/filtered-logging/page.md
```diff
@@ -0,0 +1,72 @@
...
+import (
+    "gofr.dev/pkg/gofr"
+    "gofr.dev/pkg/gofr/logging"
+)
+
+func main() {
+    app := gofr.New()
+
+    // Use the logger
+    logger := logging.NewLogger(logging.INFO)
```
srijan-27, line 37: it is not required to create a new logger. We can use `app.Logger()`

pkg/gofr/container/container.go
```diff
@@ -54,80 +54,126 @@ func NewContainer(conf config.Config) *Container {
 }
 
 func (c *Container) Create(conf config.Config) {
+	c.setAppName(conf)
+	c.setAppVersion(conf)
+	c.createLogger(conf)
+	c.createMetricsManager()
+	c.createRedis(conf)
+	c.createSQL(conf)
+	c.createPubSub(conf)
```
srijan-27, line 63: This should not be included in this PR, as it is not directly related to masking logs. Please create a separate PR for this.

pkg/gofr/container/container.go
```diff
@@ -54,80 +54,126 @@ func NewContainer(conf config.Config) *Container {
...
 	c.Debug("Container is being created")
+}
 
+func (c *Container) createMetricsManager() {
 	c.metricsManager = metrics.NewMetricsManager(exporters.Prometheus(c.appName, c.appVersion), c.Logger)
-
-	// Register framework metrics
 	c.registerFrameworkMetrics()
-
-	// Populating an instance of app_info with the app details, the value is set as 1 to depict the no. of instances
```
srijan-27, line 77: deleting the comments is not required. Please revert

srijan-27 (changes requested):

srijan-27:
@SuyeshBadge - Are you still working on this?

SuyeshBadge:
> @SuyeshBadge - Are you still working on this?

Yes @srijan-27 will update soon

SuyeshBadge:
Hey @srijan-27 and @vipul-rawat,

The changes we discussed have been implemented. Please review them and let me know if anything else is needed.

Thanks!

pkg/gofr/logging/logger_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"reflect"
+
```
srijan-27, line 8: unwanted extra newline, it can be removed.

pkg/gofr/logging/logger_test.go
```diff
@@ -214,3 +216,284 @@ func TestPrettyPrint(t *testing.T) {
...
+	}
+
+	expected := `{"name":"********","email":"********************","password":"**********"}`
+	if string(maskedMessage) != expected {
+		t.Errorf("Expected %s, but got %s", expected, maskedMessage)
+	}
+
+	if !reflect.DeepEqual(filteredMessage, expectedOutput) {
+		t.Errorf("Expected output %v, but got %v", expectedOutput, filteredMessage)
+	}
```
srijan-27, line 260: same as above, assert.Equal can be used.

srijan-27 (changes requested):
@SuyeshBadge - please take a look

SuyeshBadge:
Hey @srijan-27 @vipul-rawat,

I see there have been some recent changes pushed on the PR. Is there anything pending on my end? If yes, please let me know. Otherwise, let's move forward with this.

Thanks!

srijan-27:
@SuyeshBadge - Have you tested these changes by running an example on local and verifying whether the fields of struct are getting masked?

SuyeshBadge:
Not yet @srijan-27 .

aryanmehrotra:
@SuyeshBadge can we connect and close this ASAP?
If you are stuck we can work together!

srijan-27:
> Certainly, @aryanmehrotra .
> 
> I identified the issue with the filter not working and have implemented a fix. However, we can still connect if you'd like to discuss the changes in detail.

@SuyeshBadge - If you have implemented the fix, please push the code and we can verify then.

SuyeshBadge:
Sure folks, will push it today.

SuyeshBadge:
Hey Folks, Pushed some latest changes please take a look and let me know if this works. 

FYI @srijan-27 @aryanmehrotra

srijan-27:
@SuyeshBadge - can you please attach the snapshots for the masked logs, if you have tested it locally on any example.

SuyeshBadge:
Sure, @srijan-27. Attached are the snapshots of the logs, the code, and the keys that need to be masked.

- **Snapshot of fields that need to be masked**  
  fields to be masked

- **Snapshot of the log after masking**  
  log of masking

- **Snapshot of logs without masking**  
  logs without masking

- **Snapshot of the code snippet**  
  code for masking

SuyeshBadge:
Sure, @srijan-27. Attached are the snapshots of the logs, the code, and the keys that need to be masked.

- **Snapshot of fields that need to be masked**  
  fields to be masked

- **Snapshot of the log after masking**  
  log of masking

- **Snapshot of logs without masking**  
  logs without masking

- **Snapshot of the code snippet**  
  code for masking

- **Snapshot of the log, if the data passed, is a pointer**
code

Pointer

srijan-27:
@SuyeshBadge - have you tried passing the struct as args in log?

Something like this:
```go
func helloHandler(ctx *gofr.Context) (interface{}, error) {
	user := struct {
		Name     string
		Email    string
		Password string
	}{
		"srijan", "srijan.rastogi@gofr.dev", "secret-123",
	}

	ctx.Logger.Infof("%v", user)

	return "hello world!", nil
}
```

SuyeshBadge:
Hey @srijan-27,
In this case, the output will not be masked because the message being passed to the filter method is a string that is formatted before being passed. I'm not entirely sure if this is the intended behavior. Let me know what you think and what we can do here.

srijan-27:
@SuyeshBadge - I don't think this should be intended behaviour. As if we consider this, then we are restricting user to only mask fields in unformatted logs. We should have a behaviour such that user can mask fields irrespective of the log formatting.

SuyeshBadge:
Hey @srijan-27, Can we discuss this over a call? Let me know if you are available.

SuyeshBadge:
Hi @srijan-27,

The issue is fixed. We are now masking the arguments, ensuring that all types of data passed are masked and not limited to any specific case. If you encounter any more edge cases, please feel free to raise them.

Also, let me know if this solution works for you. I am attaching a snapshot of the updated use cases.

---
**Snapshots**
Struct example
terminal

srijan-27:
@SuyeshBadge - Although this implementation is masking the fields as required. But it is also adding a log which is not helpful and not understandable for user. See below:
Screenshot 2024-06-20 at 12 15 11 PM

Also, the request/response log is missing in the new implementation:
Screenshot 2024-06-20 at 12 15 40 PM

Please check this.

SuyeshBadge:
> @SuyeshBadge - Although this implementation is masking the fields as required. But it is also adding a log which is not helpful and not understandable for user. See below: Screenshot 2024-06-20 at 12 15 11 PM
> 
> Also, the request/response log is missing in the new implementation: Screenshot 2024-06-20 at 12 15 40 PM
> 
> Please check this.

Ohh Let me check why this log is being added here.

srijan-27:
Spent time on fixing this implementation, but seems like a lot of issues with the current implementation like: not proper masking, breaking on existing logs. And also since this is using reflections a lot, so the performance is affected considerably. 
Hence, closing this PR. Would need to think of a better solution.
JFYI: @vipul-rawat, @aryanmehrotra, @Umang01-hash

srinivas365:
@srijan-27
I have created a package to redact the pii data. Can I implement this in gofr?
https://github.com/srinivas365/piishield/tree/v1.0.0

srijan-27:
@srinivas365 - Sure, you can work on this issue: #459.
Please drop a comment there so I can assign this to you.

Issue #535, opened 2024-04-28
Add support to connect to postgres with AddRESTHandlers.


aryanmehrotra:
Issue described here as potential Issue 4: https://medium.com/@kostiantynillienko/gofr-features-and-potential-improvements-review-configuration-http-server-client-redis-e4da36819b22

> After inspecting the source code I have found the reason in /pkg/gofr/crud_handlers.go , ? used as a placeholder, but PostgreSQL requires an ordered one like $1:

illenko:
I will try to fix it.

PR #549, opened 2024-05-01
Fix PostgreSQL support in AddRESTHandlers.

illenko:
## PostgreSQL support in AddRESTHandlers

**Description:**

-   Fixed [issue #535](https://github.com/gofr-dev/gofr/issues/535).
-   Introduced new exported method `sql.Rebind`, used to replace **bindvars** according to dialect provided.

**Checklist:**

### DIFF
--- a/pkg/gofr/crud_handlers.go
+++ b/pkg/gofr/crud_handlers.go
@@ -119,14 +119,14 @@ func (e *entity) Create(c *Context) (interface{}, error) {
 		fieldValues = append(fieldValues, reflect.ValueOf(newEntity).Elem().Field(i).Interface())
 	}
 
-	stmt := fmt.Sprintf("INSERT INTO %s (%s) VALUES (%s)",
+	rawStmt := fmt.Sprintf("INSERT INTO %s (%s) VALUES (%s)",
 		e.name,
 		strings.Join(fieldNames, ", "),
 		strings.Repeat("?, ", len(fieldNames)-1)+"?",
 	)
-	rStmt := sql.Rebind(c.SQL.Dialect(), stmt)
+	stmt := sql.Rebind(c.SQL.Dialect(), rawStmt)
 
-	_, err = c.SQL.ExecContext(c, rStmt, fieldValues...)
+	_, err = c.SQL.ExecContext(c, stmt, fieldValues...)
 	if err != nil {
 		return nil, err
 	}
@@ -177,10 +177,10 @@ func (e *entity) Get(c *Context) (interface{}, error) {
 	newEntity := reflect.New(e.entityType).Interface()
 	id := c.Request.PathParam("id")
 
-	query := fmt.Sprintf("SELECT * FROM %s WHERE %s = ?", e.name, e.primaryKey)
-	rQuery := sql.Rebind(c.SQL.Dialect(), query)
+	rawQuery := fmt.Sprintf("SELECT * FROM %s WHERE %s = ?", e.name, e.primaryKey)
+	query := sql.Rebind(c.SQL.Dialect(), rawQuery)
 
-	row := c.SQL.QueryRowContext(c, rQuery, id)
+	row := c.SQL.QueryRowContext(c, query, id)
 
 	dest := make([]interface{}, e.entityType.NumField())
 	val := reflect.ValueOf(newEntity).Elem()
@@ -224,15 +224,15 @@ func (e *entity) Update(c *Context) (interface{}, error) {
 
 	query := strings.Join(paramsList, ", ")
 
-	stmt := fmt.Sprintf("UPDATE %s SET %s WHERE %s = %s",
+	rawStmt := fmt.Sprintf("UPDATE %s SET %s WHERE %s = %s",
 		e.name,
 		query,
 		e.primaryKey,
 		id,
 	)
-	rStmt := sql.Rebind(c.SQL.Dialect(), stmt)
+	stmt := sql.Rebind(c.SQL.Dialect(), rawStmt)
 
-	_, err = c.SQL.ExecContext(c, rStmt, fieldValues[1:]...)
+	_, err = c.SQL.ExecContext(c, stmt, fieldValues[1:]...)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


pkg/gofr/datasource/sql/sql_test.go
```diff
@@ -156,6 +156,15 @@ func Test_NewSQLMock(t *testing.T) {
 	assert.NotNil(t, mockMetric)
 }
 
+func Test_NewSQLMockWithConfig(t *testing.T) {
+	dbConfig := DBConfig{Dialect: "dialect", HostName: "hostname", User: "user", Password: "password", Port: "port", Database: "database"}
+	db, mock, mockMetric := NewSQLMocksWithConfig(t, &dbConfig)
+	assert.NotNil(t, db)
```
aryanmehrotra, line 162: declaration can be separated from assert statements - need to give an empty line between 161 and 162

pkg/gofr/crud_handlers.go
```diff
@@ -117,10 +119,12 @@ func (e *entity) Create(c *Context) (interface{}, error) {
...
 
-	stmt := fmt.Sprintf("INSERT INTO %s (%s) VALUES (%s)",
-		e.name,
-		strings.Join(fieldNames, ", "),
-		strings.Repeat("?, ", len(fieldNames)-1)+"?",
+	stmt := sql.Rebind(c.SQL.Dialect(),
+		fmt.Sprintf("INSERT INTO %s (%s) VALUES (%s)",
+			e.name,
+			strings.Join(fieldNames, ", "),
+			strings.Repeat("?, ", len(fieldNames)-1)+"?",
```
aryanmehrotra, line 126: can be defined as a separate variable, would increase readability.

pkg/gofr/datasource/sql/bind.go
```diff
@@ -0,0 +1,47 @@
...
+		replacement := make([]interface{}, count)
+
+		for i := 0; i < count; i++ {
+			replacement[i] = fmt.Sprintf("$%v", i+1)
+		}
+
+		return fmt.Sprintf(queryFormat, replacement...)
+	}
+
+	return query
```
aryanmehrotra, line 35: This would remain unused.

Can we keep `case QUESTION, UNKNOWN:` as the dafault case?

pkg/gofr/crud_handlers.go
```diff
@@ -117,10 +119,12 @@ func (e *entity) Create(c *Context) (interface{}, error) {
 		fieldValues = append(fieldValues, reflect.ValueOf(newEntity).Elem().Field(i).Interface())
 	}
 
-	stmt := fmt.Sprintf("INSERT INTO %s (%s) VALUES (%s)",
-		e.name,
-		strings.Join(fieldNames, ", "),
-		strings.Repeat("?, ", len(fieldNames)-1)+"?",
+	stmt := sql.Rebind(c.SQL.Dialect(),
+		fmt.Sprintf("INSERT INTO %s (%s) VALUES (%s)",
```
aryanmehrotra, line 123: I am getting the following error when I am trying to test this feature - 
`{
    "error": {
        "message": "pq: syntax error at or near \"user\""
    }
}`

Similar error for getByID
`
{
    "error": {
        "message": "pq: column \"id\" does not exist"
    }
}
`

Need to check all of them, probably a slight syntax change for postgres and sql.

aryanmehrotra (changes requested):

pkg/gofr/crud_handlers.go
illenko: In Postgres, `user` is a reserved SQL keyword. You should avoid naming your tables using reserved keywords. Keyword usage was not expected to be fixed in the scope of this PR (just bindvars usage).

In tests, I have just updated queries for demonstrating `sql.Rebind` actually called and bindvars type changed.

illenko: I'm not sure, but potentially with a table name like `interval` with MySQL, you can get similar errors.

We need a new issue regarding keyword support (for both MySQL and PostgreSQL) in CRUD Handlers.

pkg/gofr/datasource/sql/bind.go
illenko: Yes, but we have to suppress:
`pkg/gofr/datasource/sql/bind.go:23:2: missing cases in switch of type sql.BindVarType: sql.UNKNOWN, sql.QUESTION (exhaustive)`

aryanmehrotra: In that case an If statement can also work.

We can move to switch once we have any new bindVarType.

aryanmehrotra (approved):

pkg/gofr/crud_handlers.go
illenko: Raised [issue](https://github.com/gofr-dev/gofr/issues/551)

pkg/gofr/crud_handlers.go
```diff
@@ -117,11 +119,12 @@ func (e *entity) Create(c *Context) (interface{}, error) {
 		fieldValues = append(fieldValues, reflect.ValueOf(newEntity).Elem().Field(i).Interface())
 	}
 
-	stmt := fmt.Sprintf("INSERT INTO %s (%s) VALUES (%s)",
+	rawStmt := fmt.Sprintf("INSERT INTO %s (%s) VALUES (%s)",
 		e.name,
 		strings.Join(fieldNames, ", "),
 		strings.Repeat("?, ", len(fieldNames)-1)+"?",
```
vipul-rawat, line 125: Instead of rebinding the `?` for Postgres after the formation of the Query, can we change this line(125) to get the values placeholder depending on the dialect?

illenko: Yeah, I believe we will do this, but in scope of next issue (with keywords), maybe we have to introduce functions like buildInsertQuery(dialect, table, fields) and use it in crud_handlers as well.

vipul-rawat: Yes, it would make more sense to build queries based on the configured dialect.

vipul-rawat (approved):

Umang01-hash (approved):

PR #564, opened 2024-05-03
[Fixed]: CodeQL go toolchain version warning 

surajitpore0:
- update mod file with go version  **toolchain go1.21.9** to avoid CodeQL warning.

from the CodeQL side, they have added some validation changes regarding this, so once this is released, this warning will be gone.

for now, we will get this warning even though we have fixed this.

> The CodeQL team mentioned if the Workflow run is successful then nothing to worry about
ref: https://github.com/github/codeql/issues/15647#issuecomment-2077160943

For further insights, refer to the issue [here](https://github.com/github/codeql/issues/15647).


vipul-rawat:
Stale PR, closing

Issue #551, opened 2024-05-02
SQL database keywords in AddRESTHandlers


illenko:
The issue has been found by @aryanmehrotra in the scope [PR](https://github.com/gofr-dev/gofr/pull/549#discussion_r1586785434).

> I am getting the following error when I am trying to test this feature -
> { "error": { "message": "pq: syntax error at or near \"user\"" } }
> 
> Similar error for getByID
> { "error": { "message": "pq: column \"id\" does not exist" } }

It is caused by using raw queries like `INSERT INTO user (id, name) VALUES ($1, $2)` with structs name or field is keyword:

Keyword examples:
* Postgres: `USER`, `TYPE`, `TOKEN` ([Full list](https://www.postgresql.org/docs/current/sql-keywords-appendix.html));
* MySQL: `user`, `view`, `year` ([Full list](https://dev.mysql.com/doc/refman/8.0/en/keywords.html)).

illenko:
I would like to fix it.

PR #566, opened 2024-05-04
Feature: SQL keywords support, snake-case for database table/column names in CRUD Handlers.

illenko:
Support SQL (database specific) keywords in SQL queries for CRUD Handlers.

-   Introducing a new query builder in `sql` package with util methods to build SQL queries according to dialect -> ([issue](https://github.com/gofr-dev/gofr/issues/551)).
-   Removed `Rebind` func, it is not needed for now.

**Breaking Changes (maybe):**

-   Now table/column names are in snake-case in CRUD handlers.

**Example for testing:**
```go
type user struct {
	ID    int    `json:"id"`
	Name  string `json:"name"`
	Limit int    `json:"limit"`
}

func main() {
	app := gofr.New()
	app.AddRESTHandlers(&user{})
	app.PUT("/upsert-user/{id}", func(c *gofr.Context) (interface{}, error) {
		var u user

		err := c.Bind(&u)

		id, _ := strconv.Atoi(c.PathParam("id"))

		rows, err := c.SQL.QueryContext(c, sql.SelectByQuery(c.SQL.Dialect(), "user", "id"), id)
		
		if rows.Next() {
			c.Info("Updating existing user")
			_, err := c.SQL.ExecContext(c, sql.UpdateByQuery(c.SQL.Dialect(), "user", []string{"name", "limit"}, "id"), u.Name, u.Limit, id)
		} else {
			c.Info("Inserting new user")
			sql.InsertQuery(c.SQL.Dialect(), "user", []string{"id", "name", "limit"})
			_, err := c.SQL.ExecContext(c, sql.InsertQuery(c.SQL.Dialect(), "user", []string{"id", "name", "limit"}), id, u.Name, u.Limit)
		}

		return user{
			ID:    id,
			Name:  u.Name,
			Limit: u.Limit,
		}, nil
	})

	app.Run()
}

```

**Checklist:**


illenko:
@aryanmehrotra I don't like how I modified `Test_GetHandler` in `crud_handlers_test.go` but unfortunately with the approach used in `Test_UpdateHandler` (when the test has 2 `for` loops) it fails with `sqlmock`. 
Maybe you can take a look?

illenko:
@aryanmehrotra I believe we have to think about `db` tags in structs in the scope of CRUD handlers to configure column names:

```go
type user struct {
	ID    int    `json:"id" db:"id"`
	Name  string `json:"name" db:"name"`
	Limit int    `json:"limit" db:"limit_value"`
}
```

Umang01-hash:
> @aryanmehrotra I don't like how I modified `Test_GetHandler` in `crud_handlers_test.go` but unfortunately with the approach used in `Test_UpdateHandler` (when the test has 2 `for` loops) it fails with `sqlmock`. Maybe you can take a look?

@illenko In order to remove the 2 loops from `Test_UpdateHandler` one way is we can break the tests in two more more tests so that we eliminate that looping over `dialects`. Having a separate test method for sql as well as postgres can make the tests more simpler.

illenko:
> > @aryanmehrotra I don't like how I modified `Test_GetHandler` in `crud_handlers_test.go` but unfortunately with the approach used in `Test_UpdateHandler` (when the test has 2 `for` loops) it fails with `sqlmock`. Maybe you can take a look?
> 
> 
> 
> @illenko In order to remove the 2 loops from `Test_UpdateHandler` one way is we can break the tests in two more more tests so that we eliminate that looping over `dialects`. Having a separate test method for sql as well as postgres can make the tests more simpler. 

@Umang01-hash , yes, but tests are different in query only, it will just make test file larger in 2 times.

Umang01-hash:
> > > @aryanmehrotra I don't like how I modified `Test_GetHandler` in `crud_handlers_test.go` but unfortunately with the approach used in `Test_UpdateHandler` (when the test has 2 `for` loops) it fails with `sqlmock`. Maybe you can take a look?
> > 
> > 
> > @illenko In order to remove the 2 loops from `Test_UpdateHandler` one way is we can break the tests in two more more tests so that we eliminate that looping over `dialects`. Having a separate test method for sql as well as postgres can make the tests more simpler.
> 
> @Umang01-hash , yes, but tests are different in query only, it will just make test file larger in 2 times.

Yups makes sense. Will review the code and see if we can improve on tests..

pkg/gofr/datasource/sql/query_builder_test.go
```diff
@@ -0,0 +1,154 @@
...
+		{
+			dialect:  "mysql",
+			expected: "INSERT INTO `user` (`id`, `name`) VALUES (?, ?)",
+		},
+		{
+			dialect:  "postgres",
+			expected: `INSERT INTO "user" ("id", "name") VALUES ($1, $2)`,
+		},
+		{
+			dialect:  "any-other-dialect",
```
Umang01-hash, line 26: This test case with dialect as `any-other-dialect` don't make a lot of sense here. As the quote function also works only on two dialects and we also don't have a default case here:

```golang
func quote(dialect string) string {
	if DialectPostgres == dialect {
		return QuoteDouble
	}

	return QuoteBack
}
```

It's better to remove this test case or if we really want to include it we should return an error in case of un-supported dialect but that i think should be handled much before code coming here.

pkg/gofr/crud_handlers_test.go
```diff
@@ -144,13 +147,8 @@ func Test_CreateHandler(t *testing.T) {
 	}
 }
 
+//nolint:funlen // test fails when used 2 loops in test cases approach
```
Umang01-hash, line 150: we shouldn't use `nolint:funlen`. As discussed you can break the test into `Test_GetHandler_MySQL` and `Test_GetHandler_Postgres`, i know there will be some code repetition (try extracting out the common code like making of context, db , metrics etc. if required ). we can prefer that over using nolints.

Umang01-hash (changes requested):

pkg/gofr/crud_handlers_test.go
illenko: I'm not sure why it was failing for me for some strange reason - that's why I avoided 2 loops approach (seems to be kind of local issue), but now I tried once again and it works.
So I propose to stay in this format:

```
for dialect : dialects {
     for testCase : testCases {
         // test body
     } 
}
```

It will help us avoid code duplication and reduce costs when we decide to add a new dialect.

Umang01-hash (approved):

Umang01-hash:
@illenko one more thing to address here is , the **Breaking change** that you mentioned:
`Now table/column names are lowercased in CRUD handlers.`

is good but while seeing through the [naming conventions of relational databases in go](https://gorm.io/docs/conventions.html#Column-Name), my thought is we should be having columns names in snakeCase instead of simple lower case.

For example: if a field name is `isEmployed`, corresponding column name shoudl be `is_employed`, but currently i think it will be `isemployed`.

We need to update it in the documentation too, so the user also gets to know what naming conventions we are following by default for this feature.

illenko:
> @illenko one more thing to address here is , the **Breaking change** that you mentioned:
> 
> `Now table/column names are lowercased in CRUD handlers.`
> 
> 
> 
> is good but while seeing through the [naming conventions of relational databases in go](https://gorm.io/docs/conventions.html#Column-Name), my thought is we should be having columns names in snakeCase instead of simple lower case.
> 
> 
> 
> For example: if a field name is `isEmployed`, corresponding column name shoudl be `is_employed`, but currently i think it will be `isemployed`.
> 
> 
> 
> We need to update it in the documentation too, so the user also gets to know what naming conventions we are following by default for this feature.

Yes, you are right, I may try to cover it in scope of this PR

aryanmehrotra (approved):

PR #582, opened 2024-05-08
Move mock_logger from testutil to logger

srijan-27:
- `mocklogger` was present as part of testutil package which isn't ideal.
- Hence, moved mocklogger in logging package itself.
- Although to fix the cyclic dependency between logging and service packages, have created a new sub package remotelogger.
**NOTE:** `NewRemoteLogger` is renamed to `New`, since now it is part of `remotelogger` package.
- Closes issue: #540

**Checklist:**


vikash:
I think it does not really solve the problem. For example, now there are different log levels in mocklogger and the logger.

vikash (changes requested):
Shifting the mockLogger to a sub package does not really change anything. Log level for examples are not the same.

aryanmehrotra (changes requested):
We can see separately if we even want to use MockLogger, as NewLogger should also be good to use!

aryanmehrotra (approved):

Umang01-hash (approved):

vikash (approved):

PR #602, opened 2024-05-13
add PATCH method to cors Allow methods

vipul-rawat:


srijan-27 (approved):

vikash:
Instead of hardcoding it like this, we can get the information from the router. When there is no handler attached, we should not be returning those methods in headers. That might make this a bit better.

srijan-27:
> Instead of hardcoding it like this, we can get the information from the router. When there is no handler attached, we should not be returning those methods in headers. That might make this a bit better.

Sure, will look into this.

Issue #621, opened 2024-05-15
Kafka Producer Batch Confguration


illenko:
Now GoFr uses Kafka Writer (https://github.com/segmentio/kafka-go) with default configuration.

### Problem
It causes latency issues when you implement a simple HTTP endpoint that reads the request body and produces it in Kafka Topic (it is just example).
Source code: https://github.com/illenko/gofr-issue-replication
* Latency with GoFr Producer [tested with autocannon](https://github.com/mcollina/autocannon):
```
autocannon  http://localhost:8080/transactions -m POST --header 'Content-Type: application/json' --body '[{"ID":"3b241101-e2bb-4255-8caf-4136c566a964","Datetime":"2022-04-01T12:34:56Z","Amount":123.45,"Category":"coffee","Country":"Ukraine"},{"ID":"1b241101-e2bb-4255-8caf-4136c566a964","Datetime":"2022-04-01T12:34:56Z","Amount":100.05,"Category":"taxi","Country":"Ukraine"}]' -a 100 -c 10
Running 100 requests test @ http://localhost:8080/transactions
10 connections


┌─────────┬─────────┬─────────┬─────────┬─────────┬────────────┬──────────┬─────────┐
│ Stat    │ 2.5%    │ 50%     │ 97.5%   │ 99%     │ Avg        │ Stdev    │ Max     │
├─────────┼─────────┼─────────┼─────────┼─────────┼────────────┼──────────┼─────────┤
│ Latency │ 2005 ms │ 2007 ms │ 2096 ms │ 2097 ms │ 2015.85 ms │ 26.49 ms │ 2097 ms │
└─────────┴─────────┴─────────┴─────────┴─────────┴────────────┴──────────┴─────────┘
┌───────────┬─────┬──────┬─────┬─────────┬─────────┬─────────┬─────────┐
│ Stat      │ 1%  │ 2.5% │ 50% │ 97.5%   │ Avg     │ Stdev   │ Min     │
├───────────┼─────┼──────┼─────┼─────────┼─────────┼─────────┼─────────┤
│ Req/Sec   │ 0   │ 0    │ 0   │ 10      │ 4.77    │ 5       │ 10      │
├───────────┼─────┼──────┼─────┼─────────┼─────────┼─────────┼─────────┤
│ Bytes/Sec │ 0 B │ 0 B  │ 0 B │ 3.32 kB │ 1.58 kB │ 1.66 kB │ 3.32 kB │
[...]

PR #631, opened 2024-05-17
Feature: Kafka Producer Batch Confguration.

illenko:
**Description:**
Introduced new properties: `KAFKA_BATCH_SIZE`, `KAFKA_BATCH_BYTES`, `KAFKA_BATCH_TIMEOUT` to configure Kafka Producer Batch behavior ([issue](https://github.com/gofr-dev/gofr/issues/621)).

**Checklist:**


docs/advanced-guide/using-publisher-subscriber/page.md
```diff
@@ -27,10 +27,23 @@ that are specific for the type of message broker user wants to use.
...
+| `PUBSUB_BROKER`       | Address to connect to kafka broker                                                                                     | +            |           | `localhost:9092` | Not empty string |
+| `CONSUMER_ID`         | Consumer group id to uniquely identify the consumer group                                                              | if consuming |           | `order-consumer` | Not empty string |
+| `PUBSUB_OFFSET`       | Determines from whence the consumer group should begin consuming when it finds a partition without a committed offset. | -            | `-1`      | `10`             | int              |
+| `KAFKA_BATCH_SIZE`    | Limit on how many messages will be buffered before being sent to a partition.                                          | -            | `100`     | `10`             | Positive int     |
+| `KAFKA_BATCH_BYTES`   | Limit the maximum size of a request in bytes before being sent to a partition.                                         | -            | `1048576` | `65536`          | Positive int     |
```
aryanmehrotra, line 37: Can you update the format based on this documentation
https://markdoc.dev/docs/tags#table

As it will not render in UI

aryanmehrotra, line 37: Any limit on the max batch_bytes if so can be added there

pkg/gofr/datasource/pubsub/kafka/kafka.go
```diff
@@ -92,6 +101,18 @@ func validateConfigs(conf Config) error {
...
+		return errBatchSize
+	}
+
+	if conf.BatchBytes <= 0 {
+		return errBatchBytes
+	}
+
+	if conf.BatchTimeout <= 0 {
+		return errBatchTimeout
+	}
```
aryanmehrotra, line 114: rather than returning an error we can give a warn log and use default value incase of any invalid value

illenko: I would keep errors in case of wrong value, it is better to follow the principle to fail application start rather then replace values set by user with defaults.

docs/advanced-guide/using-publisher-subscriber/page.md
illenko: I don't think it make sense, in fact all configs have max limit by value type:

* The maximum value for batch.size in Kafka is 2147483647 bytes, which is equivalent to 2^31 - 1 bytes. This is because the batch.size configuration is an integer, and the maximum value for an integer in Java (which Kafka is written in) is 2^31 - 1. 
* The linger.ms configuration in Kafka is a long type, and the maximum value for a long in Java (which Kafka is written in) is 9223372036854775807 or 2^63 - 1. However, this value is in milliseconds, which translates to a very long time (approximately 292 million years). In practical terms, you would typically set linger.ms to a much smaller value, such as a few milliseconds to a few seconds, depending on your specific use case and throughput-latency tradeoff.

aryanmehrotra (approved):

vipul-rawat (approved):

PR #604, opened 2024-05-13
En/custom errors

Umang01-hash:
## Pull Request Template

**Description:**

-   Add a new errors package containing some pre-defined errors and custom errors for GoFr.
-   Custom errors have `WithStack()` method in case they want to add a stack trace to the underlying error.
- Closes issue #541  

**Checklist:**

-   [x] I have formatted my code using  `goimport`  and  `golangci-lint`.
-   [x] All new code is covered by unit tests.
-   [x] This PR does not decrease the overall code coverage.
-   [x] I have reviewed the code comments and documentation for clarity.

**Thank you for your contribution!**

PR #643, opened 2024-05-21
chore: Update SQL query parameter in entity update function

reski-rukmantiyo:
**Description:**

Current status:
1. update function not respect param from QueryString based on [GoFr website](https://gofr.dev/docs/quick-start/add-rest-handlers). It only update based on primary ID that contain in rest object. 
2. if there's primary key, right now, create function output is not giving the correct ID (based on the newest generated primary key ID).
3. If the result of query for delete, get id and update is null due to no rows or no result, it will gives 404 not found error message instead of generic error of HTTP 500

Therefore this pull request is to patch the condition.


pkg/gofr/http/responder.go
```diff
@@ -23,6 +24,12 @@ type Responder struct {
 // status codes and formatting responses as JSON or raw data as needed.
 func (r Responder) Respond(data interface{}, err error) {
 	statusCode, errorObj := r.HTTPStatusFromError(err)
+	if errorObj != nil && len(errorObj.(map[string]interface{})) != 0 {
```
vipul-rawat, line 27: This is not required as we have the custom errors package in pipeline #604. Also we don't want to check strings to decide status codes.

reski-rukmantiyo: Hi @vipul-rawat, I don't notice it. Thanks for there reference. I'll remove this.

reski-rukmantiyo:
Hi @vipul-rawat 

I have update my pull request.

1. update function not respect param from QueryString based on [GoFr website](https://gofr.dev/docs/quick-start/add-rest-handlers). It only update based on primary ID that contain in rest object.
2. if there's primary key, right now, create function output is not giving the correct ID (based on the newest generated primary key ID).

Umang01-hash:
@reski-rukmantiyo Thankyou for your contribution.
Regarding the changes you proposed, they will work perfectly in case of `auto_increment` **primary key** in struct.
But in case of **manual id insertion** the `LastInsertedID` will not be populated and it will break the flow.


As of now `Add-Rest-Handlers` functionality works perfectly fine on scenario when **primary key** is not `auto-incremented`. In order to make it work for the other case, one solution is where we also create migration files for table creation along with handlers addition. In that file, we can make the primary key column autoincrement and refactor the handlers code to align with the same.


Just in case if you want to work on the above ,please go ahead! Else we will pick it up and make it support auto-increment for all cases.

reski-rukmantiyo:
> @reski-rukmantiyo Thankyou for your contribution. Regarding the changes you proposed, they will work perfectly in case of `auto_increment` **primary key** in struct. But in case of **manual id insertion** the `LastInsertedID` will not be populated and it will break the flow.
> 
> As of now `Add-Rest-Handlers` functionality works perfectly fine on scenario when **primary key** is not `auto-incremented`. In order to make it work for the other case, one solution is where we also create migration files for table creation along with handlers addition. In that file, we can make the primary key column autoincrement and refactor the handlers code to align with the same.
> 
> Just in case if you want to work on the above ,please go ahead! Else we will pick it up and make it support auto-increment for all cases.

Thanks @Umang01-hash for your feedback. So here's the thing based on your feedback
- it will return LastInsertId if the column is primary key & auto increment
- it will return whatever the value is if the column have some kind of metadata id in struct (ex: primary_key) 
- it will return fieldValues[0] if it's dont have anything

```
type ModelDetail struct {
	ID          int    `json:"id" sql:"primary_key"`
	ModelID     int    `json:"modelid"`
	Detail      string `json:"detail"`
	Description string `json:"description"`
	Version     string `json:"version"`
	Tags        string `json:"tags"`
	License     string `json:"license"`
}
```

it will have sql metadata id in struct to return the id value.
How about this?

Umang01-hash:
> > @reski-rukmantiyo Thankyou for your contribution. Regarding the changes you proposed, they will work perfectly in case of `auto_increment` **primary key** in struct. But in case of **manual id insertion** the `LastInsertedID` will not be populated and it will break the flow.
> > As of now `Add-Rest-Handlers` functionality works perfectly fine on scenario when **primary key** is not `auto-incremented`. In order to make it work for the other case, one solution is where we also create migration files for table creation along with handlers addition. In that file, we can make the primary key column autoincrement and refactor the handlers code to align with the same.
> > Just in case if you want to work on the above ,please go ahead! Else we will pick it up and make it support auto-increment for all cases.
> 
> Thanks @Umang01-hash for your feedback. So here's the thing based on your feedback
> 
> * it will return LastInsertId if the column is primary key & auto increment
> * it will return whatever the value is if the column have some kind of metadata id in struct (ex: primary_key)
> * it will return fieldValues[0] if it's dont have anything
> 
> ```
> type ModelDetail struct {
> 	ID          int    `json:"id" sql:"primary_key"`
> 	ModelID     int    `json:"modelid"`
> 	Detail      string `json:"detail"`
> 	Description string `json:"description"`
> 	Version     string `json:"version"`
> 	Tags        string `json:"tags"`
> 	License     string `json:"license"`
> }
> ```
> 
> it will have sql metadata id in struct to return the id value. How about this?

@reski-rukmantiyo So currently yes we assume the first field(fieldValues[0]) to be the primary key. Sure we will add a ticket to upgrade this functionality to get the primary key from the tag values, this would resolve our ambiguity on what to return when Create happens, i.e. 
Return last insertedId when no primary key is provided in the struct tags(assuming migrations are written in a way to have an id column to be primary key with auto increment)
Return the primary key value provided by the client in case of a successful create.

reski-rukmantiyo:
> > > @reski-rukmantiyo Thankyou for your contribution. Regarding the changes you proposed, they will work perfectly in case of `auto_increment` **primary key** in struct. But in case of **manual id insertion** the `LastInsertedID` will not be populated and it will break the flow.
> > > As of now `Add-Rest-Handlers` functionality works perfectly fine on scenario when **primary key** is not `auto-incremented`. In order to make it work for the other case, one solution is where we also create migration files for table creation along with handlers addition. In that file, we can make the primary key column autoincrement and refactor the handlers code to align with the same.
> > > Just in case if you want to work on the above ,please go ahead! Else we will pick it up and make it support auto-increment for all cases.
> > 
> > 
> > Thanks @Umang01-hash for your feedback. So here's the thing based on your feedback
> > 
> > * it will return LastInsertId if the column is primary key & auto increment
> > * it will return whatever the value is if the column have some kind of metadata id in struct (ex: primary_key)
> > * it will return fieldValues[0] if it's dont have anything
> > 
> > ```
> > type ModelDetail struct {
> > 	ID          int    `json:"id" sql:"primary_key"`
> > 	ModelID     int    `json:"modelid"`
> > 	Detail      string `json:"detail"`
> > 	Description string `json:"description"`
> > 	Version     string `json:"version"`
> > 	Tags        string `json:"tags"`
> > 	License     string `json:"license"`
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > it will have sql metadata id in struct to return the id value. How about this?
> 
> @reski-rukmantiyo So currently yes we assume the first field(fieldValues[0]) to be the primary key. Sure we will add a ticket to upgrade this functionality to get the primary key from the tag values, this would resolve our ambiguity on what to return when Create happens, i.e. Return last insertedId when no primary key is provided in the struct tags(assuming migrations are written in a way to have an id column to be primary key with auto increment) Return the primary key value provided by the client in case of a successful create.

Hi @Umang01-hash if you agree, I can do for this PR. I need this also for my internal project.

Umang01-hash:
> > > > @reski-rukmantiyo Thankyou for your contribution. Regarding the changes you proposed, they will work perfectly in case of `auto_increment` **primary key** in struct. But in case of **manual id insertion** the `LastInsertedID` will not be populated and it will break the flow.
> > > > As of now `Add-Rest-Handlers` functionality works perfectly fine on scenario when **primary key** is not `auto-incremented`. In order to make it work for the other case, one solution is where we also create migration files for table creation along with handlers addition. In that file, we can make the primary key column autoincrement and refactor the handlers code to align with the same.
> > > > Just in case if you want to work on the above ,please go ahead! Else we will pick it up and make it support auto-increment for all cases.
> > > 
> > > 
> > > Thanks @Umang01-hash for your feedback. So here's the thing based on your feedback
> > > 
> > > * it will return LastInsertId if the column is primary key & auto increment
> > > * it will return whatever the value is if the column have some kind of metadata id in struct (ex: primary_key)
> > > * it will return fieldValues[0] if it's dont have anything
> > > 
> > > ```
> > > type ModelDetail struct {
> > > 	ID          int    `json:"id" sql:"primary_key"`
> > > 	ModelID     int    `json:"modelid"`
> > > 	Detail      string `json:"detail"`
> > > 	Description string `json:"description"`
> > > 	Version     string `json:"version"`
> > > 	Tags        string `json:"tags"`
> > > 	License     string `json:"license"`
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > it will have sql metadata id in struct to return the id value. How about this?
> > 
> > 
> > @reski-rukmantiyo So currently yes we assume the first field(fieldValues[0]) to be the primary key. Sure we will add a ticket to upgrade this functionality to get the primary key from the tag values, this would resolve our ambiguity on what to return when Create happens, i.e. Return last insertedId when no primary key is provided in the struct tags(assuming migrations are written in a way to have an id column to be primary key with auto increment) Return the primary key value provided by the client in case of a successful create.
> 
> Hi @Umang01-hash if you agree, I can do for this PR. I need this also for my internal project.

Sure @reski-rukmantiyo take it ahead but please create a separate issue for that and start the work in. a new PR and keep this PR limited to `update function not respect param from QueryString based on [GoFr website](https://gofr.dev/docs/quick-start/add-rest-handlers). It only update based on primary ID that contain in rest object.`

reski-rukmantiyo:
Noted @Umang01-hash , let me create other PR for that. Thanks

aryanmehrotra:
Since this is a big change which would require how user's might write migrations when using this feature, it would be better if we can track this ticket from an issue.

Issue #507, opened 2024-04-18
Add websocket support


vikash:

SUKANT007:
@aryan-mehrotra-zs / @vipul-rawat  I would like to work on this issue.

aryanmehrotra:
> @aryan-mehrotra-zs / @vipul-rawat  I would like to work on this issue. Can you please assign it to me.

Sure, assigning to you!

Please discuss first on how user will be using this feature.

Umang01-hash:
@SUKANT007 just wanted to re-check you want to pick up this issue and work on it.... ?

SUKANT007:
Yes @Umang01-hash, I would like to pick up, We can have discussion on it how we want users to use this. If you could share the details that would be helpful. 

Mainly we need to implement webSocket Upgrader and provide webSocket connection to the user. I'm thinking of using 
https://github.com/gorilla/websocket to implement that.

Also I was thinking of giving them the option to get WebSocket connection from context, Please share your thoughts on it or are we thinking of exposing that in some other way.

Umang01-hash:
Sure @SUKANT007 I agree with your thoughts of implementing the webSocket Upgrader and provide connection to user.

Regarding `option to get WebSocket connection from context` ,it is not a bad idea but also we need to think of different use-cases which can ease out the process to check for open socket connection.

PR #645, opened 2024-05-23
Add websocket support

SUKANT007:
**Description:**
- Websocket upgrader with options
- WsConnectionCreate func to upgrade the http connection to websocket connection
- Closes: #507 

Capabilities: 
- Wanted to give the option to users to Override default websocket upgrader.

Limitations:
- Not able to implement it without introducing significant changes to Gofr structure

@Umang01-hash  will be continuing from my changes with internal team.

 Was thinking of giving the user the capability of something like `AddWebsocketHandler`, this can be also be something looked by internal team.


Umang01-hash:
The work of this PR has been updated and refactored to a new approach in PR #692. Closing here to avoid duplicacy

Issue #495, opened 2024-04-18
cassandra support


vikash:

shashank-zopsmart:
I will work on this issue

aryanmehrotra:
Sure, assigning to you!

PR #654, opened 2024-05-27
Fe/support for cassandra

ssshekhu53:
- Added Cassandra interface
- Issue https://github.com/gofr-dev/gofr/issues/495


pkg/gofr/gofr.go
```diff
@@ -398,3 +398,7 @@ func (a *App) AddCronJob(schedule, jobName string, job CronFunc) {
 		a.Logger().Errorf("error adding cron job, err : %v", err)
 	}
 }
+
+func (a *App) UseCassandra(db datasource.Cassandra) {
```
aryanmehrotra, line 402: seems like datasource is not imported here.
Also we can move this to externalDB file in the same pacakge

aryanmehrotra (commented):
Have we implemented the interface somewhere?
Would need to add more methods for the same.

pkg/gofr/gofr.go
ssshekhu53: I will make the changes

aryanmehrotra (changes requested):
1. Can we add cassandra implementations as a package as it is for mongo.
2. Comments are missing
3. Migration doesn't support transaction as of now, need to figure out the same

shashank-zopsmart:
> 1. Can we add cassandra implementations as a package as it is for mongo.
> 2. Comments are missing
> 3. Migration doesn't support transaction as of now, need to figure out the same

Removed migration support for now.
Added comments and cassandra implementation package.

aryanmehrotra (changes requested):
1. Tests are missing for cassandraSession and cassandra query methods.
2. GoDoc comments are missing, also add developer notes on why you have created the interfaces.
3. Add comments with example for usage.
4. In cassandra.go function `newClusterConfg` spelling is wrong

aryanmehrotra (approved):

aryanmehrotra (dismissed):

pkg/gofr/datasource/cassandra/interfaces.go
```diff
@@ -0,0 +1,36 @@
+package cassandra
+
+import "github.com/gocql/gocql"
+
+// clusterConfig defines methods for interacting with a Cassandra clusterConfig. This interface
```
Umang01-hash, line 5: Instead of repeating this line multiple times:
`This interface is designed to be mockable for unit testing purposes, allowing you to control the behavior`

We can say it commonly like `All the interfaces are designed .......` and place it at the top

Umang01-hash (changes requested):

Umang01-hash (approved):

aryanmehrotra (dismissed):

vipul-rawat (approved):

Issue #568, opened 2024-05-06
Add `-h` or `--help` for  `gofr.NewCMD()`


linux-china:
List sub commands if `./cli -h` invoked.

vipul-rawat:
we can also print their functionality along with the subcommands when `-h` or `--help` is invoked. Need to have a way to add helper/description strings to the sub-commands.

IvanGael:
Quite interesting! I want to work on it.

vipul-rawat:
@IvanGael any updates on this? Please let us know if you are facing any issues :)

IvanGael:
@vipul-rawat In fact, I try to understand the general structure of the code and the project to know exactly where I need to integrate this feature. If you could tell me the files that are affected by this feature, that would help me a lot. THANKS!

vipul-rawat:
So for gofr applications, we have tried to generalise the request/response structure for our applications. So command line applications also fall under the same. So it is important to understand the Request and Responder interface.
The files concerning the feature would pe `pkg/gofr/gofr.com`, `pkg/gofr/cmd.go`

vipul-rawat:
We would like to have a default command added to the cmd applications that would list out all the routes that are registered with the cmd applications, we can also print out them when an unknown sub-command is executed.

IvanGael:
Thank you for the clarification.
I'll try to look into that and get back to you shortly if I have any further questions.

KedarisettiSreeVamsi:
Can I try to work on it as well?

IvanGael:
@KedarisettiSreeVamsi I'm already working on this issue so I think it would be wiser to work on another issue. I think currently another issues are open! @vipul-rawat  what do you think about it ?

vipul-rawat:
@IvanGael any updates? Anywhere I could help you?

KedarisettiSreeVamsi:
@vipul-rawat and @aryanmehrotra Can I work on this issue?

IvanGael:
@vipul-rawat  I just created a pull request. Try to review it.

vipul-rawat:
> @vipul-rawat and @aryanmehrotra Can I work on this issue?

Hey @KedarisettiSreeVamsi IvanGael is already working on the same, you can provide feedbacks and suggestions on the work done in PR #661

IvanGael:
@vipul-rawat I just added new commit

KedarisettiSreeVamsi:
[...]

PR #661, opened 2024-05-28
add help options to display helper text for cmd applications

IvanGael:
Closes: #568

**Description:**

There are the changes have been made

// cmd.go

- Structure for Sub-Commands:route struct now includes a description field to store the command description.
- Handling -h or --help:he Run method checks for -h or --help flags. If found, it calls cmd.printHelp() to print the help message and exits.
- Listing Sub-Commands: printHelp method prints all registered sub-commands along with their descriptions.
- Default Command Handling: If no command is provided or an unknown command is encountered, the help message is printed along with an error message.

The files cmd_test.go, gofr.go, gofr_test.go and examples/sample-cmd/main.go are updated accordingly.

**Checklist:**


pkg/gofr/cmd.go
```diff
@@ -62,9 +76,17 @@ func (cmd *cmd) handler(path string) Handler {
 	return nil
 }
 
-func (cmd *cmd) addRoute(pattern string, handler Handler) {
+func (cmd *cmd) addRoute(pattern string, handler Handler, description string) {
```
ccoVeille, line 79: I would suggest a refactoring where you would simply pass a route struct to this function

pkg/gofr/gofr_test.go
```diff
@@ -218,18 +219,24 @@ func Test_otelErrorHandler(t *testing.T) {
 }
 
 func Test_addRoute(t *testing.T) {
+	// Mock the command-line arguments to simulate running the "log" sub-command.
+	os.Args = []string{"", "log"}
```
ccoVeille, line 223: This sounds a bit dirty. You might face race issues at some points with tests

ccoVeille (commented):
My feedback on your changes, you can of course ignore them

pkg/gofr/gofr.go
```diff
@@ -236,8 +236,8 @@ func (a *App) Logger() logging.Logger {
 
 // SubCommand adds a sub-command to the CLI application.
 // Can be used to create commands like "kubectl get" or "kubectl get ingress".
-func (a *App) SubCommand(pattern string, handler Handler) {
-	a.cmd.addRoute(pattern, handler)
+func (a *App) SubCommand(pattern string, handler Handler, description string) {
```
vipul-rawat, line 239: This is a breaking change as the SubCommand function takes description as a required argument.
I would recommend to take description as a variadic string(slice of string) and join them with "/n" at end of each string line.

pkg/gofr/gofr_test.go
vipul-rawat: I don't think that there is a problem with this since tests are not running in parallel, specially the ones changing the os.Args global variable. For a better practice I can recommend setting os.Args to their original state when tests are run with some command line arguments.

vipul-rawat:
@IvanGael This gives user to define the custom helper doc to be used printed for each sub command. We also wanted to have a default helper functionality that prints all available sub commands for the executed command in case the command is not present.

pkg/gofr/cmd.go
IvanGael: Okay @ccoVeille . I will try it!

pkg/gofr/gofr_test.go
IvanGael: I will try to fix that ! @vipul-rawat  @ccoVeille

IvanGael:
@vipul-rawat  After fixing the issues and implementing the various suggestions, should I close this pull request before creating another one?

vipul-rawat:
> @vipul-rawat After fixing the issues and implementing the various suggestions, should I close this pull request before creating another one?

@IvanGael Try to update this PR itself

pkg/gofr/gofr_test.go
ccoVeille: > The problem will occur when test are running in parallel and in the test that are changing the os.Args global variable. For a better practice I can recommend setting os.Args to their original state when tests are run with some command line arguments.

Yes, it's a common thing to do it with a t.Cleanup (cleaner than a defer)

I had to do it in some other go projects using cobra.

I'm a bit surprised there is no cleaner way to do it than overloading `os.Args` when using cobra.

Please note, I'm doing this commit from my phone. I didn't look at the code

pkg/gofr/gofr.go
ccoVeille: Your comment somehow matches mine

https://github.com/gofr-dev/gofr/pull/661#discussion_r1616710862

IvanGael:
@vipul-rawat I just added new commit to this PR. Try to review.

IvanGael:
@vipul-rawat  What causes merge conflict issues?

ccoVeille (approved):
Thanks for the changes. I way simpler to read and maintain now, I think.

I will let maintainers/owner share their point of view

srijan-27:
> @vipul-rawat What causes merge conflict issues?

Merge conflicts happen when 2 branches make different changes to the single line, it would be because of other PR (which had changes to the same line as this branch) getting merged to base branch. You can resolve them easily, but be careful as we would not want to remove the changes from other branch.

IvanGael:
Okay @srijan-27 I will try to resolve it

pkg/gofr/cmd_test.go
```diff
@@ -113,10 +118,10 @@ func Test_Run_ErrorRouteWithSpecialCharacters(t *testing.T) {
 		os.Args = tc.args
 		c := cmd{}
 
-		c.addRoute(tc.args[1], func(c *Context) (interface{}, error) {
-			c.Logger.Info("handler called")
-
-			return nil, nil
+		c.addRoute(route{
+			pattern:     tc.args[1],
+			handler:     func(c *Context) (interface{}, error) { c.Logger.Info("handler called"); return nil, nil },
```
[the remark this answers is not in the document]
IvanGael: @ccoVeille it's done!

IvanGael:
@vipul-rawat @srijan-27 @ccoVeille I don't understant why this line cause conflict

Screenshot 2024-05-30 123705

ccoVeille:
In such a case there is no real need to understand. Remove everything between <<<< and >>>> then save. Your IDE will fix the import

vipul-rawat:
> @vipul-rawat Normally there should be no more merge conflicts because I did what @ccoVeille suggested to me. I don't understand why this issue persists. Could you help me with this ?

You need to update your branch in the forked repo with the current development branch of this repo(in which this PR merges the code).
To do this: 
- Add a remote repo with this repo's link - `git remote add gofr https://github.com/gofr-dev/gofr`
- Merge the development onto your branch- `git pull gofr development`

These 2 steps should give you a conflict in your gofr_test file, resolve them and push it to your branch.

IvanGael:
@vipul-rawat Shouldn't I first merge the helpSubCommandsFeature branch to the development branch ?

pkg/gofr/gofr.go
```diff
@@ -250,8 +250,13 @@ func (a *App) Logger() logging.Logger {
...
 // SubCommand adds a sub-command to the CLI application.
 // Can be used to create commands like "kubectl get" or "kubectl get ingress".
-func (a *App) SubCommand(pattern string, handler Handler) {
-	a.cmd.addRoute(pattern, handler)
+func (a *App) SubCommand(pattern string, handler Handler, description ...string) {
+	desc := strings.Join(description, "\n")
+	a.cmd.addRoute(route{
+		pattern:     pattern,
+		handler:     handler,
+		description: desc,
```
vipul-rawat, line 258: Why are you not populating the help field, when that is the thing which stores the help documentation.

pkg/gofr/cmd.go
```diff
@@ -62,9 +77,18 @@ func (cmd *cmd) handler(path string) Handler {
...
+}
+
+func (cmd *cmd) printHelp() {
+	fmt.Println("Available commands:")
+	for _, route := range cmd.routes {
+		help := route.help
+		if help == "" {
+			help = route.description // Use description if custom helper documentation is not provided
+		}
+		fmt.Printf("  %s: %s\n", route.pattern, help)
```
vipul-rawat, line 91: Here we can try to print the description and help doc together in 2 separate lines. It would help if you could paste some screenshots of how the output would look like.

vipul-rawat (changes requested):

IvanGael:
Screenshot 2024-06-01 022509

pkg/gofr/cmd.go
```diff
@@ -28,22 +32,33 @@ func (cmd *cmd) Run(c *container.Container) {
...
+	if showHelp || command == "" {
+		cmd.printHelp()
+		return
+	}
+
+	h := cmd.handler(strings.TrimSpace(command))
 	ctx := newContext(&cmd2.Responder{}, cmd2.NewRequest(args), c)
 
 	if h == nil {
+		fmt.Println("Unknown command:", command)
```
vipul-rawat, line 60: This is not required, as responder is printing "No command found" on line 62

pkg/gofr/cmd.go
```diff
@@ -1,21 +1,25 @@
...
+	"strings"
 
 	cmd2 "gofr.dev/pkg/gofr/cmd"
+	"gofr.dev/pkg/gofr/container"
 )
 
 type cmd struct {
-	routes []route
+	routes      []route
+	defaultHelp string // Default helper documentation
```
vipul-rawat, line 15: What is defaultHelp, it is not being assigned any value anywhere!

pkg/gofr/gofr.go
```diff
@@ -250,8 +250,23 @@ func (a *App) Logger() logging.Logger {
 
 // SubCommand adds a sub-command to the CLI application.
 // Can be used to create commands like "kubectl get" or "kubectl get ingress".
-func (a *App) SubCommand(pattern string, handler Handler) {
-	a.cmd.addRoute(pattern, handler)
+func (a *App) SubCommand(pattern string, handler Handler, help string, description ...string) {
```
vipul-rawat, line 253: @IvanGael this is still a breaking change, we cannot have `help string` as parameter. Please stick to the variadic input `description`. Keep it simple, if descriptions are provided, just join them and make it as help doc for that sub command.

```
func (a *App) SubCommand(pattern string, handler Handler, description ...string) {
    if len(description) > 0 {
        desc := description[0]
    }

    help := strings.Join(description, "\n")
}
```

Something like this should suffice

ccoVeille: @IvanGael please note the code provided here is invalid

It was provided as is, but you will have to rework it `desc` must be defined before the `if description`. Otherwise, there is a scope issue


> ```go
> func (a *App) SubCommand(pattern string, handler Handler, description ...string) {
>     if len(description) > 0 {
>         desc := description[0]
>     }
> 
>     help := strings.Join(description, "\n")
> 
>     a.cmd.addRoute(route{
> 		pattern:     pattern,
> 		handler:     handler,
> 		description: desc,
>                  help:   help,
>     }
> }
> ```

IvanGael:
@vipul-rawat  @ccoVeille 
- The help field is populated based on if -h or --help flag is provided in variadic description
- The desc is extracted as well from variadic description
- I removed defaultHelp in cmd struct

KedarisettiSreeVamsi:
Hi @IvanGael, Please try to add usage in helper function and also try to use variadic parameters for help and description instead of mandatory parameters. You can refer to #675.

pkg/gofr/gofr.go
```diff
@@ -250,8 +250,28 @@ func (a *App) Logger() logging.Logger {
...
 // SubCommand adds a sub-command to the CLI application.
 // Can be used to create commands like "kubectl get" or "kubectl get ingress".
-func (a *App) SubCommand(pattern string, handler Handler) {
-	a.cmd.addRoute(pattern, handler)
+func (a *App) SubCommand(pattern string, handler Handler, description ...string) {
+	var descParts []string
+	help := ""
+	// Extract help information from the description and construct the new description
+	for _, d := range description {
+		if strings.HasPrefix(d, "-h ") || strings.HasPrefix(d, "--help ") {
```
vipul-rawat, line 258: I dont think that having prefix `-h` or `--help` is required to differentiate between description and help text. We can take the first string to be description and then the rest others combine to form the helper text. If only one string is provided only description is printed else both description and helper text are printed.

pkg/gofr/gofr.go
```diff
@@ -266,8 +255,19 @@ func (a *App) Logger() logging.Logger {
...
 // Can be used to create commands like "kubectl get" or "kubectl get ingress".
-func (a *App) SubCommand(pattern string, handler Handler) {
-	a.cmd.addRoute(pattern, handler)
+func (a *App) SubCommand(pattern string, handler Handler, description ...string) {
+	help := ""
+	// Extract help information from the description
+	if len(description) > 1 {
+		help = strings.Join(description[1:], " ")
+	}
+	desc := description[0]
```
vipul-rawat, line 264: This could panic if no description string is provided!

aryanmehrotra (changes requested):
1. Response doesn't seem right with unwanted square brackets
image
Would be better if we have something similar to how other help commands look like of -git or github (gh)

pkg/gofr/cmd.go
```diff
@@ -1,70 +1,174 @@
...
 
 func (cmd *cmd) handler(path string) Handler {
+	if len(path) > 1 && path[:2] == "--" {
+		path = path[2:]
+	} else if path[0] == '-' {
+		path = path[1:]
+	}
+
+	if strings.Contains(path, " ") {
+		path = strings.Split(path, " ")[0]
```
ccoVeille, line 107: The result will be the sane even there is no space in it, so maybe you could remove the `if`

```go
strings.Split("whatever")[0] == "whatever"
```

pkg/gofr/cmd.go
```diff
@@ -1,70 +1,174 @@
...
+		ctx.responder.Respond(h(ctx))
+
+		if it != len(command)-1 {
+			ctx.responder.Respond("\n", nil)
+		}
+	}
 }
 
 func (cmd *cmd) handler(path string) Handler {
+	if len(path) > 1 && path[:2] == "--" {
```
ccoVeille, line 100: Your code is equivalent to

```go
if strings.HasPrefix(path, "--") {
   path = strings.TrimPrefix("--")
} else if strings.HasPrefix(path, "-") {
   path = strings.TrimPrefix("-")
}
```

So your code could be written

```go
path = strings.Trim(path, "-")
```

pkg/gofr/gofr.go
```diff
@@ -266,8 +255,8 @@ func (a *App) Logger() logging.Logger {
 
 // SubCommand adds a sub-command to the CLI application.
 // Can be used to create commands like "kubectl get" or "kubectl get ingress".
-func (a *App) SubCommand(pattern string, handler Handler) {
-	a.cmd.addRoute(pattern, handler)
+func (a *App) SubCommand(pattern string, handler Handler, options ...Options) {
```
ccoVeille, line 258: This pattern change is great

pkg/gofr/gofr_test.go
```diff
@@ -225,18 +225,27 @@ func Test_otelErrorHandler(t *testing.T) {
 }
 
 func Test_addRoute(t *testing.T) {
+	originalArgs := os.Args // Save the original os.Args
+
+	// Modify os.Args for the duration of this test
+	os.Args = []string{"", "log"}
+	defer func() { os.Args = originalArgs }() // Restore os.Args after the test
```
ccoVeille, line 232: I would prefer a t.Cleanup here

pkg/gofr/cmd.go
ccoVeille: Unless, you have a need to allow "---whatever" to be transformed as "-path"

pkg/gofr/cmd.go
```diff
@@ -1,70 +1,173 @@
...
+
+		if it != len(command)-1 {
+			ctx.responder.Respond("\n", nil)
+		}
+	}
 }
 
 func (cmd *cmd) handler(path string) Handler {
+	// Trim leading dashes
+	path = strings.TrimPrefix(strings.TrimPrefix(path, "--"), "-")
```
ccoVeille, line 100: You didn't reply to my thread but fixed it somehow

https://github.com/gofr-dev/gofr/pull/661#discussion_r1632321861

I'm unsure why you are not using `strings.Trim(path, "-")`

Especially now you added a comment about removing trailing dashes

ccoVeille (approved):
Only one small question

pkg/gofr/cmd.go
```diff
@@ -1,70 +1,173 @@
...
 }
 
 type route struct {
-	pattern string
-	handler Handler
+	pattern     string
+	handler     Handler
+	description string
+	help        string
+	fullPattern string
```
vipul-rawat, line 22: what is the use case for `fullPattern` field?

vipul-rawat, line 22: What is the use case for `fullpatern` field?

vipul-rawat (changes requested):
Please remove the unused code - `fullpattern` field

pkg/gofr/cmd.go
KedarisettiSreeVamsi: Hi @vipul-rawat ,
The fullpattern also matches the same way as pattern but it will be like -h for pattern and --help for fullpattern.

vipul-rawat: But I believe this would be fixed in your PR @KedarisettiSreeVamsi, so not required as part of this PR and issue

aryanmehrotra (approved):

ccoVeille:
You did it @IvanGael

Issue #662, opened 2024-05-28
Bind Pub/Sub messages to basic data types.


illenko:
## Problem

For now, Gofr supports binding messages from Pub/Sub providers to:
* `struct`
* `map[string]any`

If user want to consume message as string (bool, int, float64) in this way:
```go
func  Process(c *gofr.Context) error {
	var message string
	err := c.Bind(&message)
	if err != nil {
		c.Logger.Error("Error unmarshalling message: ", err)
		return nil
	}

	c.Logger.Info("Consumed message: ", message)

	return nil
}
```
Message Example: `HELLO MESSAGE`.
Error `ERRO [08:27:47] [Error unmarshalling message:  invalid character 'H' looking for beginning of value]` occurred.

Reason (see `/datasource/pubsub/message.go`):
```go
func (m *Message) Bind(i interface{}) error {
	// TODO - implement other binding functionality
	err := json.Unmarshal(m.Value, i)
	return err
}
```

## Proposal
Add support for basic data types:
* `bool`
* `int`
* `float64`
* `string`

illenko:
@aryanmehrotra , if you are okay with this, I can take it.

PR #670, opened 2024-05-29
Bind consumed message to different types.

illenko:
**Description:**

-  Bind consumed message to different types: `struct`, `map[string]any`, `int`, `bool`, `float64` and `string`.
-  Check parameter passed to `Bind` is pointer.
- Closes: #662 


aryanmehrotra (approved):

Umang01-hash (dismissed):

illenko:
> @illenko should we not support all built in types like int64, int32, uint, etc. I think int64 is a much used type in applications.

@vipul-rawat  As for the difference between `int` and `int64`, this is architecture-dependent. `int` is simply an alias for either a 32-bit or 64-bit integer, depending on the architecture you are compiling for.

I can add this support, but I do not think it will have a real usage.

vipul-rawat:
@illenko sure we can have the current ones then, if requirement comes we can add other data types further.

vipul-rawat (approved):

Umang01-hash (approved):

Issue #502, opened 2024-04-18
Support for serving static files


vikash:

vipul-rawat:
#528

interrrp:
How should the API look like? I'm thinking of something like this:

```go
app.AddStaticFiles("/static", "./static")
```

vikash:
We can follow a convention also. If a public folder is present, files in that folder can be accessed by the server.

KedarisettiSreeVamsi:
Can I try to work on this?

Issue #156, opened 2023-12-29
Incorrect conversion between integer types in azure


aryanmehrotra:
Tracking issue for:
- [x] https://github.com/gofr-dev/gofr/security/code-scanning/27

PR #674, opened 2024-06-02
Added Static File Handler

KedarisettiSreeVamsi:
Added Static File Handling
If a public directory is found then it will create a public endpoint by default
If we want to added any other folders can use AddStaticFiles endpoint and then if endpoint and directory is provided then it will check whether the particular endpoint is available otherwise throws an error

Closes: #502.


pkg/gofr/gofr.go
```diff
@@ -93,9 +94,52 @@ func New() *App {
...
+	return fileSplitter
+}
+
+func (a *App) AddStaticFiles(endpoint, filePath string) {
+	a.httpRegistered = true
+	dupFilePath := ""
+	fileSplitter := getFilePathSplitter()
+	if filePath[:2] == "./" {
+		dupFilePath, _ = os.Getwd()
+		dupFilePath += fileSplitter + filePath + fileSplitter
```
ccoVeille, line 129: fileSpliter and getFilePathSplitter seems to me to reimplement filepath.Join

https://pkg.go.dev/path/filepath#Join

So I would discourage using the home made thing you add here

ccoVeille (changes requested):

pkg/gofr/gofr.go
KedarisettiSreeVamsi: Thanks @ccoVeille, I will make the changes and push the update in few minutes.

ccoVeille (approved):
Thanks for doing the changes I suggested

KedarisettiSreeVamsi:
Let me know if any changes can be done to accommodate more functionality.

pkg/gofr/gofr.go
```diff
@@ -93,9 +94,36 @@ func New() *App {
...
+	if filePath[:2] == "./" {
+		dupFilePath, _ = os.Getwd()
+		dupFilePath = filepath.Join(dupFilePath, filePath)
+	} else {
+		dupFilePath = filePath
+	}
+	if endpoint[0] != '/' {
+		endpoint = "/" + endpoint
+	}
+	if _, err := os.Stat(dupFilePath); err == nil {
```
ccoVeille, line 120: Invert the if and would be clearer

```go
        if _, err := os.Stat(dupFilePath); err != nil { 
                 a.container.Logger.Errorf("Couldn't register %s static endpoint", endpoint) 
                 return
         }


        a.httpServer.router.AddStaticFiles(endpoint, dupFilePath)
```

ccoVeille (changes requested):

Umang01-hash:
@KedarisettiSreeVamsi Thankyou for your contribution.
Please add consider adding tests for any new code added as not doing it will decrease the overall code coverage.
Also, in my opinion we don't actually need an example of how to serve static files. Your example is small and simple so a better way is to add a documentation for this.

You can add it in the `advanced-guide` directory under the `docs `directory of the repository.

pkg/gofr/gofr.go
```diff
@@ -93,9 +94,36 @@ func New() *App {
...
+func (a *App) AddStaticFiles(endpoint, filePath string) {
+	a.httpRegistered = true
+	dupFilePath := ""
+	if filePath[:2] == "./" {
+		dupFilePath, _ = os.Getwd()
+		dupFilePath = filepath.Join(dupFilePath, filePath)
+	} else {
+		dupFilePath = filePath
+	}
+	if endpoint[0] != '/' {
```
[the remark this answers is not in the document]
KedarisettiSreeVamsi: Made relevant changes

docs/advanced-guide/serving-static-files/page.md
```diff
@@ -0,0 +1,76 @@
+# Serving Static Files using GoFr
+
+Often times we require to serve static content be it a default profile image or a static website. We want to have a mechanism to serve those content without having a hassel of implementing it from scratch.
```
ccoVeille, line 3: `Often, we require` missing comma + often instead of often times

`static content be it a default` => `static content like a default` 

`Those contentS` plural

docs/advanced-guide/serving-static-files/page.md
```diff
@@ -0,0 +1,76 @@
...
+import "gofr.dev/pkg/gofr"
+
+func main(){
+    app := gofr.New()
+    app.Run()
+}
+
+```
+
+Additionally if we want to serve additional static endpoints, we have a open function called `AddStaticFiles()` which takes 2 parameters endpoint and the filepath of the static folder which we want to serve.
```
ccoVeille, line 36: `Additionally, if we` missing comma

`a open` => `an open`

Stylistic remarks:
Additionally followed by additional in same sentence seems odd
What is a open function? dedicated maybe?

ccoVeille (changes requested):

docs/advanced-guide/serving-static-files/page.md
KedarisettiSreeVamsi: Thanks for the remarks @ccoVeille , will update as per the recommendations.

docs/advanced-guide/serving-static-files/page.md
KedarisettiSreeVamsi: Yeah I should have rephrased it to a dedicated. Will make the changes.

pkg/gofr/http/router.go
```diff
@@ -13,6 +13,8 @@ type Router struct {
 	RegisteredRoutes *[]string
 }
 
+type Config func(r *Router)
```
vipul-rawat, line 16: this function seems to be unused!

KedarisettiSreeVamsi: I was thinking of some usecases but first we will push this static handler into the framework and later work on the enhancements. I will remove this line of code in next commit.

docs/advanced-guide/serving-static-files/page.md
```diff
@@ -0,0 +1,76 @@
...
+import "gofr.dev/pkg/gofr"
+
+func main(){
+    app := gofr.New()
+    app.Run()
+}
+
+```
+
+Additionally,if we want to serve more static endpoints, we have a dedicated function called `AddStaticFiles()` which takes 2 parameters endpoint and the filepath of the static folder which we want to serve.
```
ccoVeille, line 36: missing comma after " Additionally"

ccoVeille (changes requested):
Small typo

docs/advanced-guide/serving-static-files/page.md
KedarisettiSreeVamsi: comma is present after Additionally is it a space which you are mentioning after the comma @ccoVeille

ccoVeille: Indeed, missing space after comma

ccoVeille: Thanks for fixing it

vipul-rawat:
@KedarisettiSreeVamsi please add unit tests for the feature you've implemented!

KedarisettiSreeVamsi:
Hi @vipul-rawat,
Can you please mention how I can write unit tests without a static folder in the same folder as test file? Should I use a file within the folder for the testing purpose?

KedarisettiSreeVamsi:
I am pretty new to this so was the reason I am asking this question.

pkg/gofr/gofr_test.go
```diff
@@ -504,3 +505,114 @@ func Test_AddCronJob_Success(t *testing.T) {
...
...3,326.73c-2.77,0-5.8-.68-8.78-1.35-2.21-.5-4.16-.62-5.73-.72-2.37-.15-4.25-.27-5.29-1.66-1.03-1.37-1.03-3.7,0-8.3,1.37-6.11,4.42-15.86,8.16-23.88,4.82-10.35,9.46-15.11,13.78-14.14,5.79,1.3,8.1,10.96,8.96,16.61,1.2,7.88,1.1,17.24-.26,23.29h0c-1.8,8.02-5.93,10.14-10.84,10.14h0Zm1.06-48.37c-3.18,0-7.08,4.64-11.07,13.23-3.68,7.9-6.69,17.5-8.03,23.52-.87,3.86-.96,5.97-.31,6.83.55.73,1.98.82,3.97.94,1.63.1,3.65.23,6,.76,9.33,2.09,15.17,2.85,17.48-7.44,2.47-11.03.59-35.98-7.34-37.76-.23-.05-.46-.08-.69-.08h-.01Z\" fill=\"#dfdfe0\"/><polygon points=\"361.49 400.87 272.57 401.45 258.75 395.1 266.8 312.99 274.9 313.4 358.21 317.64 361.49 400.87\" fill=\"#e2e3e4\"/><polygon points=\"272.57 401.45 258.75 395.1 266.8 312.99 274.9 313.4 272.57 401.45\" fill=\"#272223\" isolation=\"isolate\" opacity=\".2\"/><path id=\"uuid-aa721d86-32e3-4ace-957f-0814f6d1eb89-48-46-89-48-101-33\" d=\"m329.89,281.37c1.49,7.32-1.24,14.01-6.08,14.94s-9.97-4.26-11.45-11.58c-.63-2.92-.53-5.94.29-8.82l-5.89-31.11,15.22-2.41,4.19,30.92c1.89,2.36,3.16,5.12,3.72,8.06h0Z\" fill=\"#f8a8ab\"/><path d=\"m269.54,97.92s20.33-.86,21.39,0c5.55,4.53,38.1,168.04,38.1,168.04h-20.62l-38.87-168.04s0,0,0,0Z\" fill=\"#6c63ff\"/></svg>")
```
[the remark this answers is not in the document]
KedarisettiSreeVamsi: I just want to check if a large file will result in some errors, so was the reason to use an svg, but if you want me to change it to txt file then I will make related updates.

pkg/gofr/gofr_test.go
```diff
@@ -504,3 +505,114 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+		t.Error("Couldn't create shopping.svg file")
+	}
+	file2, errFile := os.Create("./public/meeting.svg")
+	if errFile != nil {
+		t.Error("Couldn't create meeting.svg file")
+	}
+	svgFile1ContentSize, _ := file1.Write(svgFileContent1)
+	svgFile2ContentSize, _ := file2.Write(svgFileContent2)
+
+	defer func() {
```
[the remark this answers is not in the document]
KedarisettiSreeVamsi: Will make an update.

pkg/gofr/gofr_test.go
```diff
@@ -504,3 +505,111 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+		request.Header.Set("Content-Type", "application/json")
+		client := http.Client{}
+		resp, err := client.Do(request)
+		bodyBytes, _ := io.ReadAll(resp.Body)
+		defer resp.Body.Close()
+		body := string(bodyBytes)
+		assert.Nil(t, err, "TEST[%d], Failed.\n%s", it, tc.desc)
+		assert.Equal(t, tc.statusCode, resp.StatusCode, "TEST[%d], Failed with Status Body.\n%s", it, tc.desc)
+		if tc.expectedBody != "" {
+			assert.Contains(t, body, tc.expectedBody, "TEST [%d], Failed with Expected Body. \n%s", it, tc.desc)
```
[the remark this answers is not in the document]
KedarisettiSreeVamsi: will make an update.

pkg/gofr/gofr_test.go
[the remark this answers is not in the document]
ccoVeille: I'm just random guy

Maybe you could move the svg variables to another files

My problem is the fact it led to tests difficult to read

ccoVeille (approved):
LGTM 👍

Umang01-hash:
@KedarisettiSreeVamsi the pipeline for your PR is failing: https://github.com/gofr-dev/gofr/actions/runs/9414675135

Please have a look and fix

ccoVeille:
Poor @KedarisettiSreeVamsi this code review is a long journey

.gitignore
```diff
@@ -22,3 +22,6 @@ vendor/
 # IDE Cache
 .vscode
 .DS_Store
+
+# Public Folder
+pkg/gofr/public
```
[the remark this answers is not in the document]
KedarisettiSreeVamsi: Hi @vipul-rawat,
Some times the cleanup isn't happening as expected so that was the reason why I had to added it explicitly.

pkg/gofr/gofr_test.go
```diff
@@ -504,3 +505,111 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+		if errFolder != nil {
+			t.Error("Couldn't remove public folder")
+		}
+	})
+
+	// Run the Application and compare the fileType and fileSize
+	app.httpRegistered = true
+	app.httpServer.port = 8002
+
+	go app.Run()
```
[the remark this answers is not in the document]
KedarisettiSreeVamsi: For some reason test cases are failing when I am trying to use serveHTTP()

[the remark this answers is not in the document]
KedarisettiSreeVamsi: It is giving 301 status code

KedarisettiSreeVamsi:
Now the unit tests have been fixed, can someone please merge it.

pkg/gofr/gofr.go
```diff
@@ -94,9 +95,39 @@ func New() *App {
 
 	app.subscriptionManager = newSubscriptionManager(app.container)
 
+	// static fileserver
+	currentWd, _ := os.Getwd()
+	checkDirectory := filepath.Join(currentWd, "public")
+
+	if _, err := os.Stat(checkDirectory); !os.IsNotExist(err) {
```
ccoVeille, line 102: Strange logic

I think would have expected to see

```go
        if _, err := os.Stat(checkDirectory); err == nil {
```

I mean with current code, if you have any other error than not exist, you will try to serve file

pkg/gofr/gofr.go
```diff
@@ -94,9 +95,39 @@ func New() *App {
...
+		app.AddStaticFiles("public", checkDirectory)
```
ccoVeille, line 103: I'm a bit worried about the security concern here.

It's somehow logic to serve a public folder, but let's consider this feature is added to all existing repository.

So they are likely to serve files that were present in a public folder that weren't planned to be served.

I would expect people who does want to share public file to look for AddStaticFiles in documentation and add it in their code.

Maybe, it's OK  if the code is merged as this, it should come with a big warning or breaking change mention in the release.

pkg/gofr/gofr_test.go
```diff
@@ -504,3 +504,107 @@ func Test_AddCronJob_Success(t *testing.T) {
 
 	assert.Truef(t, pass, "unable to add cron job to cron table")
 }
+
+func TestStaticHandler(t *testing.T) {
+	// Generate a public directory endpoint
+	directory := "./public"
+
+	if _, err := os.Stat(directory); err != nil {
+		err := os.Mkdir("./public", 0755)
```
ccoVeille, line 513: Use "public" variable here and everywhere

pkg/gofr/gofr_test.go
```diff
@@ -504,3 +504,107 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+	file1, errFile := os.Create("./public/index.html")
+
+	if errFile != nil {
+		t.Error("Couldn't create index.html file")
+	}
+
+	htmlContentSize, _ := file1.Write(htmlContent)
+
+	t.Cleanup(func() {
+		err := os.RemoveAll("./public")
```
ccoVeille, line 533: Use a variable do not hard code public

pkg/gofr/gofr_test.go
```diff
@@ -504,3 +504,107 @@ func Test_AddCronJob_Success(t *testing.T) {
 
 	assert.Truef(t, pass, "unable to add cron job to cron table")
 }
+
+func TestStaticHandler(t *testing.T) {
+	// Generate a public directory endpoint
+	directory := "./public"
```
ccoVeille, line 510: You could use the constant I talked about

pkg/gofr/gofr_test.go
```diff
@@ -504,3 +504,107 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+	t.Cleanup(func() {
+		err := os.RemoveAll("./public")
+		if err != nil {
+			t.Log("Couldn't remove public folder")
+		}
+	})
+
+	// Run the Application and compare the fileType and fileSize
+	app.httpRegistered = true
+	app.httpServer.port = 8022
```
ccoVeille, line 541: This is a bad practice, something could be listening locally on 8022 and test will fall

pkg/gofr/gofr_test.go
```diff
@@ -504,3 +504,107 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+
+		assert.NoError(t, err, "TEST[%d], Failed.\n%s", it, tc.desc)
+		assert.Equal(t, tc.statusCode, resp.StatusCode, "TEST[%d], Failed with Status Body.\n%s", it, tc.desc)
+
+		if tc.expectedBody != "" {
+			assert.Contains(t, body, tc.expectedBody, "TEST [%d], Failed with Expected Body. \n%s", it, tc.desc)
+		}
+
+		if tc.expectedBodyLength != 0 {
+			contentLength, _ := strconv.Atoi(resp.Header.Get("Content-Length"))
```
ccoVeille, line 595: It's simpler to compare the Get value to a _expectedBodyLength_ defined as a string not an int

And it's error prone, while you are ignoring errors here

ccoVeille: Or you can nuke the variable and use len(expexeBody) instead of having a variable linked to another

ccoVeille (changes requested):

pkg/gofr/gofr.go
KedarisettiSreeVamsi: Hi @ccoVeille,
Someone has mentioned in the issue as well to serve public folder if it exists. So that was the reason behind me adding that logic.

ccoVeille: Then as I said, it should be mentioned in the changelog/release in a very clear way

Also, I would expect to see a log entry about either serving public folder (so here around this if) or better at the end of "AddStaticFiles" method

pkg/gofr/gofr_test.go
ccoVeille: In such case, either you rely on the server to be able to start on a random port, either you define a random port in the test.

pkg/gofr/gofr.go
ccoVeille: > Someone has mentioned in the issue as well to serve public folder if it exists. So that was the reason behind me adding that logic.

For reference:

>We can follow a convention also. If a public folder is present, files in that folder can be accessed by the server. 
 
_Originally posted by @vikash in https://github.com/gofr-dev/gofr/issues/502#issuecomment-2121842503_  

It's not because someone asks for something that a good idea for security.

I'm sure @vikash (hi :wave:) who is also the issue's author won't have any issue with having a public folder being served automatically. But other people wouldn't notice the change and might face security issue.

I'm raising my concern. As I said, I'm OK with having the feature of a public folder being automagically served.

pkg/gofr/gofr_test.go
Umang01-hash: used the constat

pkg/gofr/gofr_test.go
Umang01-hash: sure used the constant

pkg/gofr/gofr_test.go
Umang01-hash: sure

pkg/gofr/gofr.go
Umang01-hash: Replaced with your logic thanks

pkg/gofr/gofr_test.go
```diff
@@ -592,19 +582,51 @@ func TestStaticHandler(t *testing.T) {
...
-				it,
-				tc.desc)
+				"TEST [%d], Failed at Expected Content-Type.\n%s", it, tc.desc)
 		}
 
 		resp.Body.Close()
 	}
 }
+
+func createPublicDirectory(t *testing.T, htmlContent []byte) error {
```
ccoVeille, line 600: this is an helper, it should start with a `t.Helper()` call

https://pkg.go.dev/testing#T.Helper

pkg/gofr/gofr_test.go
```diff
@@ -592,19 +582,51 @@ func TestStaticHandler(t *testing.T) {
...
+	directory := "./" + publicDir
+	if _, err := os.Stat(directory); err != nil {
+		if err := os.Mkdir("./"+publicDir, 0755); err != nil {
+			t.Errorf("Couldn't create a "+publicDir+" directory, error: %s", err)
```
ccoVeille, line 604: you are using Errorf

so it should be 

			t.Errorf("Couldn't create %s directory, error: %s", publicDir, err)

pkg/gofr/gofr_test.go
```diff
@@ -506,35 +506,16 @@ func Test_AddCronJob_Success(t *testing.T) {
...
 	// Generate some files for testing
 	htmlContent := []byte("<html><head><title>Test Static File</title></head><body><p>Testing Static File</p></body></html>")
 
-	file1, errFile := os.Create("./public/index.html")
-
-	if errFile != nil {
-		t.Error("Couldn't create index.html file")
+	err := createPublicDirectory(t, htmlContent)
+	if err != nil {
+		return
```
ccoVeille, line 514: to respect the pattern you are using in your test, you should use
	
```go
err := createPublicDirectory(t, htmlContent)
if err != nil {
	t.Fatalf(err)
}
```

and do not log error in createPublicDirectory (except the cleanup you can do nothing except log)

pkg/gofr/gofr_test.go
```diff
@@ -592,19 +582,51 @@ func TestStaticHandler(t *testing.T) {
...
+
+	if errFile != nil {
+		t.Error("Couldn't create index.html file")
+		return errFile
+	}
+
+	_, err := file1.Write(htmlContent)
+	if err != nil {
+		t.Error("Couldn't write to index.html file")
+		return err
```
ccoVeille, line 619: you could use (here and everywhere in createPublicDirectory)

```go
_, err := file1.Write(htmlContent)
if err != nil {
        return fmt.Errorf("Couldn't write to index.html file: %w", err)
}
```

This way you could add a fatalf when calling createPublicDirectory

pkg/gofr/gofr_test.go
```diff
@@ -504,3 +504,129 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+}
+
+func createPublicDirectory(t *testing.T, htmlContent []byte) error {
+	t.Helper()
+
+	directory := "./" + publicDir
+	if _, err := os.Stat(directory); err != nil {
+		if err := os.Mkdir("./"+publicDir, 0755); err != nil {
+			t.Errorf("Couldn't create a "+publicDir+" directory, error: %s", err)
+			return err
```
ccoVeille, line 607: If you call t.Fatalf, you can remove the error you are returning in the function signature, then you no longer have to test err when calling it

srijan-27 (approved):

aryanmehrotra (changes requested):
Since, we support swagger from the static directory.
Let's support the same as default, user would anyway have access to read from any other specified directory.

pkg/gofr/gofr.go
```diff
@@ -92,9 +95,53 @@ func New() *App {
...
+
+	defaultConfig := a.httpServer.router.GetDefaultStaticFilesConfig()
+
+	updateConfigInformation(&defaultConfig, a.Config)
+
+	if strings.HasPrefix(filePath, "./") {
+		dupFilePath, _ = os.Getwd()
+		dupFilePath = filepath.Join(dupFilePath, filePath)
+	} else {
+		dupFilePath = filePath
```
ccoVeille, line 130: I would remove the else block and define dupfilePath := filepath before the if that check fie relative path

pkg/gofr/http/router.go
```diff
@@ -36,6 +47,109 @@ func (rou *Router) Add(method, pattern string, handler http.Handler) {
...
+	fileServer := http.FileServer(http.Dir(staticConfig.FileDirectory))
+	rou.Router.NewRoute().PathPrefix(endpoint + "/").Handler(http.StripPrefix(endpoint, staticHandler(fileServer, staticConfig)))
+}
+
+// Check all the static handling configs.
+func staticHandler(fileServer http.Handler, staticConfig StaticFileConfig) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		url := r.URL.Path
+
+		const forbiddenBody string = "403 forbidden"
```
ccoVeille, line 69: This const should be moved out of the function to the package level, so you don't have to pass it to checkDotFiles,                         checkExcludedExtensions, and                         checkExcludedFiles

pkg/gofr/http/router.go
```diff
@@ -36,6 +47,109 @@ func (rou *Router) Add(method, pattern string, handler http.Handler) {
...
+
+		if staticConfig.HideDotFiles {
+			checkDotFiles(w, r, staticConfig, fileName, url, forbiddenBody)
+		}
+
+		if len(staticConfig.ExcludeExtensions) > 1 {
+			checkExcludedExtensions(w, r, staticConfig, fileName, url, forbiddenBody)
+		}
+
+		if len(staticConfig.ExcludeFiles) > 1 {
```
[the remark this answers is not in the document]
ccoVeille: Same question for extension a few lines above

pkg/gofr/http/router.go
```diff
@@ -36,6 +47,109 @@ func (rou *Router) Add(method, pattern string, handler http.Handler) {
...
+}
+
+// Static File Handling.
+func (rou *Router) AddStaticFiles(endpoint string, staticConfig StaticFileConfig) {
+	fileServer := http.FileServer(http.Dir(staticConfig.FileDirectory))
+	rou.Router.NewRoute().PathPrefix(endpoint + "/").Handler(http.StripPrefix(endpoint, staticHandler(fileServer, staticConfig)))
+}
+
+// Check all the static handling configs.
+func staticHandler(fileServer http.Handler, staticConfig StaticFileConfig) http.Handler {
```
ccoVeille, line 65: Maybe this function should be a method of StaticFileConfig


```suggestion
func (staticConfig StaticFileConfig) (staticHandler(fileServer http.Handler) http.Handler {
```

ccoVeille: Same for checkDotFiles,                         checkExcludedExtensions, and                         checkExcludedFiles

pkg/gofr/http/router.go
```diff
@@ -36,6 +47,109 @@ func (rou *Router) Add(method, pattern string, handler http.Handler) {
...
+	}
+}
+
+func checkExcludedExtensions(w http.ResponseWriter, r *http.Request, staticConfig StaticFileConfig, fileName string, url string, forbiddenBody string) {
+	if _, err := os.Stat(filepath.Join(staticConfig.FileDirectory, url)); err != nil {
+		http.NotFound(w, r)
+		return
+	}
+
+	extensions := staticConfig.ExcludeExtensions[1:]
```
ccoVeille, line 123: Why excluding the first one?

ccoVeille: Are you sure there the list is not empty? It would panic

pkg/gofr/http/router.go
```diff
@@ -36,6 +47,109 @@ func (rou *Router) Add(method, pattern string, handler http.Handler) {
...
+		w.Header().Set("Content-Type", "text/plain;charset=utf-8")
+		w.Write([]byte(forbiddenBody))
+
+		return
+	}
+}
+
+func checkExcludedExtensions(w http.ResponseWriter, r *http.Request, staticConfig StaticFileConfig, fileName string, url string, forbiddenBody string) {
+	if _, err := os.Stat(filepath.Join(staticConfig.FileDirectory, url)); err != nil {
+		http.NotFound(w, r)
```
ccoVeille, line 119: I would say it's not the role of this method

ServeHTTP would do it in a better (relative path resolution) and shared way

I would only return here, but also I would simply not do this if check

First for performance reason, because all your helpers are checking for the file existence.

Then, because it's not a problem to refuse all URLs containing a .pdf extension if you excluded them, even if the file doesn't exist.

A better way to do this, would be to use http.middleware in fact

pkg/gofr/http/router.go
```diff
@@ -36,6 +47,109 @@ func (rou *Router) Add(method, pattern string, handler http.Handler) {
...
+		return
+	}
+
+	extensions := staticConfig.ExcludeExtensions[1:]
+	for _, ext := range extensions {
+		if strings.HasSuffix(fileName, ext) {
+			w.WriteHeader(http.StatusForbidden)
+			w.Header().Set("Content-Type", "text/plain;charset=utf-8")
```
ccoVeille, line 127: I'm doubtful about the fact the page have to be a text/plain

ccoVeille:
I hope you will my review useful even I'm doing a machine gun review 😅

pkg/gofr/http/router.go
KedarisettiSreeVamsi: Hi @ccoVeille,
When I was testing the function, I got an empty element in the array causing a lot of problem. So I had to remove the first element. But please let me know what can be done.

ccoVeille: If you have an empty element in the array, maybe a split… anyway, I would recommend to skip the empty entry in each for loop

pkg/gofr/gofr.go
```diff
@@ -92,9 +95,53 @@ func New() *App {
...
+	}
+
 	return app
 }
 
+func updateConfigInformation(config *gofrHTTP.StaticFileConfig, envConfig config.Config) {
+	config.DirectoryListing, _ = strconv.ParseBool(envConfig.GetOrDefault("STATIC_DIRECTORY_LISTING", "true"))
+	config.HideDotFiles, _ = strconv.ParseBool(envConfig.GetOrDefault("STATIC_HIDEDOTFILES", "true"))
+	config.ExcludeExtensions = strings.Split(envConfig.Get("STATIC_EXCLUDE_EXTENSIONS"), ",")
+	config.ExcludeFiles = strings.Split(envConfig.Get("STATIC_EXCLUDE_FILES"), ",")
```
ccoVeille, line 113: Here is your scapegoat @KedarisettiSreeVamsi ,you have to split only when the env variable is not empty

You should create a splitEnv (or whatever the name that
- will return an empty string when the variable name is empty
- split
- loop on them
- trim spaces on them

pkg/gofr/http/router.go
ccoVeille: https://github.com/gofr-dev/gofr/pull/674#discussion_r1636963565

ccoVeille:
Please resolve the comments one by one, so I can see what is fixed. Thanks

pkg/gofr/gofr.go
KedarisettiSreeVamsi: done

pkg/gofr/http/router.go
KedarisettiSreeVamsi: removed the content-type header for the response

pkg/gofr/http/router.go
KedarisettiSreeVamsi: Couldn't implement with http.middleware but made the fixes mentioned above

pkg/gofr/http/router.go
KedarisettiSreeVamsi: made appropriate changes

pkg/gofr/http/router.go
KedarisettiSreeVamsi: moved the const to outside of function

docs/advanced-guide/serving-static-files/page.md
```diff
@@ -0,0 +1,75 @@
...
+import "gofr.dev/pkg/gofr"
+
+func main(){
+    app := gofr.New()
+    app.Run()
+}
+```
+
+Additionally, if we want to serve more static endpoints, we have a dedicated function called `AddStaticFiles()`
+which takes 2 parameters endpoint and the filepath of the static folder which we want to serve.
```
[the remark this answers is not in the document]
KedarisettiSreeVamsi: enclosed them in backticks

docs/advanced-guide/serving-static-files/page.md
```diff
@@ -0,0 +1,75 @@
...
+
+```dotenv
+project_folder
+|
+|---configs
+|       .env
+|---static
+|       <img1>.jpeg
+|       <img2>.png
+|       <img3>.jpeg
```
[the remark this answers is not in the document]
KedarisettiSreeVamsi: Anyway added a static index.html in public endpoint so I think it won't be required

docs/advanced-guide/serving-static-files/page.md
```diff
@@ -0,0 +1,75 @@
...
+
+Example project structure:
+
+```dotenv
+project_folder
+|
+|---configs
+|       .env
+|---static
+|       <img1>.jpeg
```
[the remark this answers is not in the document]
KedarisettiSreeVamsi: removed the angular brackets

pkg/gofr/gofr.go
KedarisettiSreeVamsi: made dupFilePath as mentioned

pkg/gofr/http/router.go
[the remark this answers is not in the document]
KedarisettiSreeVamsi: changed to > 0

KedarisettiSreeVamsi:
Facing a weird issue when making the static folder as default as the workflow pipeline is failing for favicon handler and when public is made the default folder then I am not facing the issue. Can someone help in this case as to why it is failing? Is it due to embedding of static in `static/files.go` for favicon testing?

KedarisettiSreeVamsi:
Currently I made public as the default endpoint to check if tests are still failing.

pkg/gofr/gofr.go
```diff
@@ -92,9 +95,53 @@ func New() *App {
...
+	config.ExcludeFiles = append(config.ExcludeFiles, "openapi.json")
```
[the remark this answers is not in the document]
ccoVeille: You might get have people reporting about it later, but I would say I'm fine with this

pkg/gofr/http/router.go
ccoVeille: Yes, it's fine. It was just a suggestion

ccoVeille (approved):
Thanks for resolving and providing comment in each thread it helped me

srijan-27:
@KedarisettiSreeVamsi - thanks for creating a separate ticket for codeclimate fix. Although we would need to fix linter errors in this PR itself, so please check on that.

KedarisettiSreeVamsi:
Hi @srijan-27,
I do understand that I need to fix the linter issues but I need some help to fix those. Also as I mentioned above, if someone can help me why when keeping static as default folder the favicon test is failing and when public is made static the test is passing.

aryanmehrotra:
@KedarisettiSreeVamsi I have fixed the linters, you may see the commit to understand it more.

KedarisettiSreeVamsi:
Hi @aryanmehrotra,
Thanks for the help. Can we merge the pull request then?

docs/references/configs/page.md
```diff
@@ -78,6 +78,26 @@ This document lists all the configuration options supported by the Gofr framewor
...
 - Description: File to save the logs in case of a CMD application
 
+---
+
+- Name: STATIC_DIRECTORY_LISTING
+- Description: To enable directory listing of files in folder (default: true)
+
+---
+
+- Name: STATIC_HIDEDOTFILES
```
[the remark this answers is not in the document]
KedarisettiSreeVamsi: There is no need to serve a hidden file but when I was testing it it was serving by default when using fileServer so I had set it to true to not serve hidden files by default but if someone wants to serve it they can change the configuration to false to render it.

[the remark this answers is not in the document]
KedarisettiSreeVamsi: It is just an added option, I know that it is absurd to even provide such option but I added it just in case.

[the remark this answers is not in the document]
vipul-rawat: I don't think we should even give an option, just so we can avoid any accidental exposure of hidden files.

vipul-rawat:
> Currently I made public as the default endpoint to check if tests are still failing.

@KedarisettiSreeVamsi please make the default dir as `static` as we already have static files in the same dir. As for the failing tests, it is failing because of the `os.RemoveAll()` on line [614](https://github.com/gofr-dev/gofr/pull/674/files#diff-974ba3b5aa38878f1cdef653b92acb0cbc2a4df857867bc6a8b199b0921b24c7R614) clean up, instead of testing you can name the HTML file as test.html and remove the specific one! This is a small change.

docs/references/configs/page.md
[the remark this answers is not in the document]
ccoVeille: @vipul-rawat I agree with you, but I would like to spot something that is somehow standard now

https://en.wikipedia.org/wiki/Well-known_URI
https://github.com/moul/awesome-well-known
https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml

I'm unsure how people would be able to share such document

Here is a real one
https://accounts.google.com/.well-known/openid-configuration

But yes, it would be a breach to allow sharing "hidden" files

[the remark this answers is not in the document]
KedarisettiSreeVamsi: by default, I have set the dot files rendering to false

ccoVeille:
@vipul-rawat @srijan-27 @aryanmehrotra 

What are your thoughts about this ?

https://github.com/gofr-dev/gofr/pull/674#discussion_r1646130482

A side topic that I raised by reviewing the PR.

You could reply in the thread for keeping the conversation in one place

pkg/gofr/helpers.go
```diff
@@ -0,0 +1,25 @@
+package gofr
+
+import (
+	"strings"
+)
+
+func SplitEnv(envString, splitString string) []string {
```
ccoVeille, line 7: I would have passed the name of the env variable here, and call the os.GetEnv here


```suggestion
func SplitEnv(envName, splitString string) []string {

envString = os.GetEnv(envName)

// …
```

This way the code calling SplitEnv is simpler.

pkg/gofr/http/router.go
```diff
@@ -36,6 +47,95 @@ func (rou *Router) Add(method, pattern string, handler http.Handler) {
...
+			staticConfig.checkExcludedFiles(w, fileName, url)
+		}
+
+		fileServer.ServeHTTP(w, r)
+	})
+}
+
+func (staticConfig StaticFileConfig) checkDirectoryListing(w http.ResponseWriter, r *http.Request, url string) {
+	if _, err := os.Stat(filepath.Join(staticConfig.FileDirectory, "index.html")); err != nil && strings.HasSuffix(url, "/") {
+		http.NotFound(w, r)
```
ccoVeille, line 96: Any other error than NotFound should be logged I think

It's open for debate if it shouldn't, use a `http.InternalError(w, r)` (pseudo code here)

ccoVeille (approved):

docs/references/configs/page.md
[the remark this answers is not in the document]
vipul-rawat: @ccoVeille I think this is a matter of `can` and not `must`, this is what I can infer from the documents and I strongly believe that we should avoid cases where users can "accidentally" expose files that were not supposed to be.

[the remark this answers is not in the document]
ccoVeille: I do agree, no worry. It's just that maybe the well-known folder could/should be planned as a dedicated feature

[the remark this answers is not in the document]
vipul-rawat: Yes, we can have a discussion thread for the same -  if we want to support the well-known for all the standard implementations!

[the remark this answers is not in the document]
ccoVeille: I opened #856

pkg/gofr/helpers.go
ccoVeille: @KedarisettiSreeVamsi

pkg/gofr/http/router.go
ccoVeille: @KedarisettiSreeVamsi

pkg/gofr/http/router.go
```diff
@@ -14,6 +17,14 @@ type Router struct {
 	RegisteredRoutes *[]string
 }
 
+type StaticFileConfig struct {
+	DirectoryListing  bool
+	HideDotFiles      bool
```
ccoVeille, line 22: Is there still a need for this variable then?

pkg/gofr/helpers.go
KedarisettiSreeVamsi: Hi @ccoVeille,
I have already taken the required config details in gofr.go and then I am passing the data to SplitEnv, is it really necessary to make this change.

pkg/gofr/http/router.go
KedarisettiSreeVamsi: I will check it once.

pkg/gofr/http/router.go
```diff
@@ -36,6 +47,95 @@ func (rou *Router) Add(method, pattern string, handler http.Handler) {
...
+
+func (staticConfig StaticFileConfig) checkDirectoryListing(w http.ResponseWriter, r *http.Request, url string) {
+	if _, err := os.Stat(filepath.Join(staticConfig.FileDirectory, "index.html")); err != nil && strings.HasSuffix(url, "/") {
+		http.NotFound(w, r)
+		return
+	}
+}
+
+func (staticConfig StaticFileConfig) checkDotFiles(w http.ResponseWriter, fileName, url string) {
+	if _, err := os.Stat(filepath.Join(staticConfig.FileDirectory, url)); err == nil && strings.HasPrefix(fileName, ".") {
```
ccoVeille, line 102: I thought about it while writing #156

I don't think checkDotFiles check for the name of the directory to check if the file is not in a hidden folder

Think about `/foo/.bar/image.jpg`

KedarisettiSreeVamsi:
Hi @ccoVeille,
I think we already have api's with .well-known is it really necessary to expose .well-known folder?

pkg/gofr/http/router.go
KedarisettiSreeVamsi: Hmm, something to work upon then. I will split the request and check if any of them are a hidden folder or hidden file then.

pkg/gofr/http/router.go
KedarisettiSreeVamsi: Not required thanks for highlighting.

ccoVeille:
> Hi @ccoVeille,
> I think we already have api's with .well-known is it really necessary to expose .well-known folder?

.well-known file has to been publicly exposed

Lucky there is a bug I spotted here, so it works

https://github.com/gofr-dev/gofr/pull/674#discussion_r1652005236



I might be wrong, it needs further checks

KedarisettiSreeVamsi:
Can someone resolve the conflict of `docs/reference/configs/page.md` as it is blocking all the tests? I tried to fix it but unfortunately it Is not getting fixed.

aryanmehrotra (changes requested):
1. Remove helper and constant files.
2. Lot of lines are not covered (coverage is not 100% for the newly added code)
3. ExcludeFiles need to be unexported (we can move the code to add openapi.json in the router itself)
4. FileDirectory needs to be renamed, doesn't seem right
5. ExcludeExtensions, HideDotFiles can be removed for now, I dont find a valid usecase. (IMO user can just not keep that file if they dont want it to be rendered)
6. DirectoryListing can be renamed to ListDirectories or ListDirectoryFiles
7. A lot of functions/methods are being exported which can be unexported
8. .env Config STATIC_DIRECTORY_LISTING can be removed

ccoVeille:
> 5. ExcludeExtensions, HideDotFiles can be removed for now, I dont find a valid usecase. (IMO user can just not keep that file if they dont want it to be rendered) (fixed)

Seem safer and avoid potential breaking changes 👍

pkg/gofr/gofr.go
```diff
@@ -439,3 +450,27 @@ func contains(elems []string, v string) bool {
 
 	return false
 }
+
+func (a *App) AddStaticFiles(endpoint, filePath string) {
+	a.httpRegistered = true
+
+	dupFilePath := filePath
```
ccoVeille, line 457: I would move the declaration of this variable just before the if that follows

pkg/gofr/gofr_test.go
```diff
@@ -561,3 +561,123 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+		resp.Body.Close()
+	}
+}
+
+func createPublicDirectory(t *testing.T, htmlContent []byte) {
+	t.Helper()
+
+	directory := "./" + defaultPublicStaticDir
+	if _, err := os.Stat(directory); err != nil {
+		if err = os.Mkdir("./"+defaultPublicStaticDir, 0755); err != nil {
```
ccoVeille, line 659: The preferred notation is `0o755`, but I'm fine.

Added since go1.13 https://go.dev/ref/spec#Integer_literals

Because of multiple misunderstanding

https://stackoverflow.com/questions/63877081/why-does-go-programming-treat-any-number-starting-with-zero-as-an-octal-and-how

https://github.com/golang/go/issues/33363

I'm just reporting this as I already wrote some comment.

aryanmehrotra (approved):

pkg/gofr/helpers.go
ccoVeille: No, no worries

pkg/gofr/gofr_test.go
```diff
@@ -561,3 +561,141 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+		}
+
+		if tc.expectedResponseHeaderType != "" {
+			assert.Equal(t,
+				tc.expectedResponseHeaderType,
+				resp.Header.Get("Content-Type"),
+				"TEST[%d], Failed at Expected Content-Type.\n%s", i, tc.desc)
+		}
+
+		resp.Body.Close()
```
ccoVeille, line 661: I would have expected this to be moved in a t.Cleanup(func()) just before the `io.ReadAll`

pkg/gofr/gofr_test.go
```diff
@@ -561,3 +561,141 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+	directory := "./" + defaultPublicStaticDir
+	if _, err := os.Stat(directory); err != nil {
+		if err = os.Mkdir("./"+defaultPublicStaticDir, os.ModePerm); err != nil {
+			t.Fatalf("Couldn't create a "+defaultPublicStaticDir+" directory, error: %s", err)
+		}
+	}
+
+	file, errFile := os.Create(filepath.Join(directory, indexHTML))
+
+	if errFile != nil {
```
ccoVeille, line 691: Any reason to have an `errFile` variable, `err` would be enough, no?

pkg/gofr/gofr_test.go
```diff
@@ -561,3 +561,141 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+		t.Fatalf("Couldn't create %s file", indexHTML)
+	}
+
+	_, err := file.Write(htmlContent)
+	if err != nil {
+		t.Fatalf("Couldn't write to %s file", indexHTML)
+	}
+
+	file.Close()
```
ccoVeille, line 700: This one should be moved in a `t.Cleanup(func` just before `file.Write`

Issue #696, opened 2024-06-07
CMD response is followed by `%`


aryanmehrotra:
**Describe the bug**
Every cmd response is followed by %

<img width="389" alt="image" src="https://github.com/gofr-dev/gofr/assets/44036979/22a36ec7-d84d-4035-a201-57e3cb172f83">

aditya028:
@aryanmehrotra can i pick up this issue

PR #697, opened 2024-06-09
fix every cmd response is followed by %

aditya028:
The `%` symbol at the end of every response is not actually part of the output . It's a feature of some terminal emulators, including the default terminal in macOS, which display a % or other symbol when the output from the last command does not end with a newline `(\n)`.

Here is the output attached
Screenshot 2024-06-10 at 4 05 16 PM

Closes: https://github.com/gofr-dev/gofr/issues/696


vikash:
This just fixes the example and not the underlying problem.

aditya028:
okk i will look into it

aditya028:
please reivew it @vikash @srijan-27

srijan-27:
@aditya028 - Did you verified the changes by running the example? If yes, please attach the fixed terminal output.
Also, since now format is changed so some tests are failing, would need to fix them as well.

aditya028:
I have checked the examples and it works fine . Also, in the codebase it is mentioned there is some issue that needs to be resolved 
`TODO - provide proper exit codes here. Using os.Exit directly is a problem for tests.`
Screenshot 2024-06-10 at 4 05 16 PM

srijan-27:
@aditya028 - Yes, that needs to be done someway. If you want you can dive deep and take a look on that, if not, then you can just fix the current tests.

srijan-27:
@aditya028 - Forgot to mention previously, if you look into the TODO task, then create a separate issue and PR for that. As it has separate context than this PR.

aditya028:
ok sure @srijan-27 , for now i will fix the test case for this .

aditya028:
@srijan-27 can you check this

srijan-27:
> @srijan-27 can you check this

@aditya028 - Couple of tests are still failing.. you can check them from the workflows.

aryanmehrotra (approved):

srijan-27 (approved):

Issue #672, opened 2024-05-31
Fix Health-Check Response when no external service is attached


aryanmehrotra:
**Describe the bug**
`/.well-known/health` returns an empty response, when no db or external service is connected.

**To Reproduce**

Create service without any database or external service.

**Expected behavior**
It should return the status of service as `UP`.

**More description**
In, case if any of the dependent service is down the status should be degraded

TheMarvelFan:
Hi,

Is anyone working on this issue?

aryanmehrotra:
No, not yet

TheMarvelFan:
Yes I want to work on this issue.
I would appreciate it if you can assign it to me under the GSSOC '24 label.
Thanks.

srijan-27:
@TheMarvelFan - We are not part of GSSOC'24.

PR #706, opened 2024-06-11
Add apphealth in health response

srijan-27:
- Added `appHealth` details (appName, appVersion, status) in response of `.well-known/health` endpoint.
- Closes: #672


aryanmehrotra (dismissed):

pkg/gofr/container/health.go
```diff
@@ -6,25 +6,33 @@ import (
...
-		datasources[name] = svc.HealthCheck(ctx)
+		healthMap[name] = svc.HealthCheck(ctx)
 	}
 
-	return datasources
+	return healthMap
+}
+
+func (c *Container) appHealth(healthMap map[string]interface{}) {
+	healthMap["status"] = "UP"
```
aryanmehrotra, line 33: We should also handle, if any of the services is down status should be DEGRADED.

aryanmehrotra (changes requested):

aryanmehrotra (approved):

vipul-rawat (approved):

Issue #666, opened 2024-05-28
Fix Metrics missing for HTTPService


aryanmehrotra:
**Describe the bug**
If service A hits `http://google.comma` which does not exist we would get a 500 Error, for which metrics should be pushed, but it is not happening.

**To Reproduce**
Steps to reproduce the behavior, if applicable:

1. The code is

   ```go
   a.EnableOAuth("http://google.comma", 10)
   
   Add the following in any example's main.go
   ```

2. The error is

   ```
   Metrics are missing @ localhost:2121/metrics
   ```

**Expected behavior**
Even though the request failed metrics should come for the same.

**More description**
It is happening because it is being checked if resp is not nil @ https://github.com/gofr-dev/gofr/blob/89994c7774fd3c267d82441eaf5293dc57eb2d4f/pkg/gofr/service/new.go#L173

sipsdaoracle:
Hi @aryanmehrotra and @srijan-27 would you please assign this to me if no one is looking at it.

srijan-27:
@sipsdaoracle - Sure, assigning this to you.

srijan-27:
Hi @sipsdaoracle - just wanted to know if you are working on this? If any issues, do let me know.

PR #708, opened 2024-06-11
Add metric for httpService error

srijan-27:
- Populating `app_http_service_response` metric with 500 resp code, when http service returning an error.
- Closes: #666


pkg/gofr/container/mock_container.go
```diff
@@ -28,6 +30,14 @@ func NewMockContainer(t *testing.T) (*Container, Mocks) {
 
 	mocks := Mocks{Redis: redisMock, SQL: sqlMock}
 
+	ctrl := gomock.NewController(t)
```
Umang01-hash, line 33: Instead of creating a new controller `gomock.NewController(t)` we can create one at the starting and pass it everywhere

Umang01-hash (changes requested):

PR #710, opened 2024-06-12
Improve spans logs

srijan-27:


Umang01-hash (approved):

pkg/gofr/exporter.go
```diff
@@ -78,13 +78,13 @@ func (e *Exporter) processSpans(ctx context.Context, logger logging.Logger, span
 
 	resp, err := client.Do(req)
 	if err != nil {
-		logger.Error(err)
+		logger.Errorf("failed to create spans, error: %v", err)
```
ccoVeille, line 81: %w instead of %v seems better

ccoVeille (changes requested):

pkg/gofr/exporter.go
Umang01-hash: @ccoVeille we are not wrapping up the error here, just logging it. In my opinion, `%w` is specifically used for wrapping errors while `%v` is used for general formatting, including logging.

ccoVeille: Oh you are right, it's a logger line. I missed it

aryanmehrotra (approved):

PR #720, opened 2024-06-14
fix add recover call in ServeHTTP 

Sianao:
Add recovery logic in the serve-http function

**Description:**

-   recover in groutine of process Handler panic happed 
-   avoid a api crash avoid other api

**Checklist:**


Sianao:
such as this one
```go
package main

import "gofr.dev/pkg/gofr"

func main() {
	// initialise gofr object
	app := gofr.New()

	// register route greet
	app.GET("/greet", func(ctx *gofr.Context) (interface{}, error) {
		panic(nil)
		return "Hello World!", nil
	})

	// Runs the server, it will listen on the default port 8000.
	// it can be over-ridden through configs
	app.Run()
}
````

It will lead to the app crash。
But we don’t want other interfaces to be unavailable because of problems with one interface.
The groutine you create yourself should be responsible for its life cycle

**Comment by srijan-27** [comment]
@Sianao - have you tried running the above code locally?
As I believe it would not crash the app, but instead will return the error stack and app will continue to run.

**Comment by Sianao** [comment]
@srijan-27 
image

this  is the test result

**Comment by srijan-27** [comment]
@Sianao - Thanks! Will check this.

**Comment by Sianao** [comment]
I will rewrite the test code , because `panic(nil)`  can't be recover on older version with nil check

File: pkg/gofr/handler.go [hunk 1]
```diff
@@ -59,13 +61,24 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		defer func() {
+			re := recover()
+			if re != nil {
+				close(panicked)
+				h.container.Error(panicLog{
+					Error:      fmt.Sprintln(re),
+					StackTrace: string(debug.Stack()),
+				})
+			}
+		}()
```
srijan-27, line 81: we can also consider creating a `panicRecovery` function, similar to https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/http/middleware/logger.go#L127

srijan-27 (changes requested):

pkg/gofr/handler.go
Sianao: ok , i will do it now

pkg/gofr/http/errors.go
```diff
@@ -83,3 +83,13 @@ func (e ErrorRequestTimeout) Error() string {
...
+
+type ErrPanic struct{}
+
+func (e ErrPanic) Error() string {
+	return http.StatusText(http.StatusInternalServerError)
+}
+
+func (e ErrPanic) StatusCode() int {
+	return http.StatusInternalServerError
+}
```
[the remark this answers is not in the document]
Sianao: Done

pkg/gofr/handler.go
Sianao: I test it before do this , I also want to call a recover func  but only direct call recover work .

Sianao: All Done @srijan-27

srijan-27 (approved):

aryanmehrotra (approved):

PR #727, opened 2024-06-17
Add logging methods for migration

srijan-27:
- Currently we are providing logger to user with only `Info` and `Error` functionalities.
- Now, also added methods for `Debug`, `Notice`, `Warn`, `Fatal`

**Checklist:**

- [x] I have formatted my code using  `goimport`  and  `golangci-lint`.


vikash:
Why does migration need fatal logging? @srijan-27

srijan-27:
@vikash - Didn't had any specific use-case in mind. But since, these are the methods which we are providing user to log inside migrations, so just thought of having parity with the app logger.

Umang01-hash (approved):

PR #739, opened 2024-06-20
Add configs for idle and max open connections

aryanmehrotra:
**Checklist:**


pkg/gofr/datasource/sql/sql.go
```diff
@@ -72,6 +74,14 @@ func NewSQL(configs config.Config, logger datasource.Logger, metrics Metrics) *D
 		return database
 	}
 
+	// We are not setting idle connection timeout because we are checking for connection
```
ccoVeille, line 77: I might be tired but I don't get the negation "we are not", while the next line of code apparently does it.

Same remark for the other comment for the other variation sets in this PR

pkg/gofr/datasource/sql/sql.go
```diff
@@ -122,13 +132,30 @@ func retryConnection(database *DB) {
 }
 
 func getDBConfig(configs config.Config) *DBConfig {
+	// if the value of maxIdleConn is negative or 0, no idle connections are retained.
+	maxIdleConn, err := strconv.Atoi(configs.Get("DB_MAX_IDLE_CONNECTION"))
+	if err != nil {
+		// setting the max open connection as the default which is being provided by default package
+		maxIdleConn = 2
```
ccoVeille, line 139: maybe you could use

const defaultMaxIdleConn = 2, with the comment you provided.so here you could use this here


```suggestion
		maxIdleConn = defaultMaxIdleConn
```

Code would be clearer, and it would be clearer you are not using a magical or arbitrary value.

Same remark for max open variable

ccoVeille (approved):

ccoVeille:
also what about supporting SetConnMaxLifetime ?

pkg/gofr/datasource/sql/sql.go
ccoVeille: You could also document this as the default value, because even with an explanation, the 2 would sound to be magical number

https://github.com/golang/go/blob/4f77a83589baa5a3038cc6e35615084dc7e5f0c1/src/database/sql/sql.go#L952

pkg/gofr/datasource/sql/sql.go
aryanmehrotra: we are not setting the timeout we are setting the max idle connection

srijan-27 (approved):

pkg/gofr/datasource/sql/sql.go
ccoVeille: ok, clearer. Thanks. I'm tired then.

Umang01-hash (approved):

srijan-27 (approved):

vipul-rawat (approved):

PR #777, opened 2024-07-02
add request timeout for health check calls

### LINKED ISSUE #711: Request Time Out during Healthcheck
**Describe the bug**
The health check endpoint times out if any downstream service takes too long to respond to their health check.

**Expected behavior**
The health check should not time out.

vipul-rawat:
- Add request timeout on context while calling downstream service health check endpoint
- Add `Timeout` field to `service.HealthConfig` to set custom health-check timeout(default is 5 seconds)

closes #711


pkg/gofr/service/health.go
```diff
@@ -16,14 +18,20 @@ type Health struct {
...
+	return h.getHealthResponseForEndpoint(ctx, ".well-known/alive", defaultTimeout)
 }
 
-func (h *httpService) getHealthResponseForEndpoint(ctx context.Context, endpoint string) *Health {
+func (h *httpService) getHealthResponseForEndpoint(_ context.Context, endpoint string, timeout int) *Health {
 	var healthResponse = Health{
 		Details: make(map[string]interface{}),
 	}
 
+	// caeate a new context with 5 second timeout for healthCheck call.
```
ccoVeille, line 29:
```suggestion
	// create a new context with 5 second timeout for healthCheck call.
```

pkg/gofr/service/health.go
```diff
@@ -16,14 +18,20 @@ type Health struct {
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeout)*time.Second)
```
ccoVeille, line 30: That's the old way (before Go 1.21) to do it. I would discourage doing this.

Please consider the following code


```suggestion
	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), time.Duration(timeout)*time.Second)
```

https://pkg.go.dev/context#WithoutCancel

Thus way all data set in the context with WithValue are still in the context, you just create a context that won't be canceled. Then you set up a timeout

pkg/gofr/service/health_config.go
```diff
@@ -4,20 +4,28 @@ import "context"
 
 type HealthConfig struct {
 	HealthEndpoint string
+	Timeout        int
 }
 
 func (h *HealthConfig) AddOption(svc HTTP) HTTP {
+	// if timeout is not provided we set the timeout to 5 sec by default.
```
ccoVeille, line 11: Magic number in comment, the value may change and this comment will be outdated


```suggestion
	// if timeout is not provided we set the timeout to a convenient default value.
```

ccoVeille (changes requested):

pkg/gofr/service/health.go
vipul-rawat: Thanks for pointing this out! We were using new context for health check calls since we did not want to have the whole health check call getting canceled by just one timing out downstream service.

ccoVeille (approved):

Umang01-hash (approved):

PR #673, opened 2024-06-01
Add Shutdown and Close Functions for Graceful Connection Termination

maohieng:
## Pull Request Template

**Description:**

This pull request introduces essential shutdown and close functions to our web services and connections, ensuring graceful termination of connections and releasing associated resources. This addition empowers users to implement their own graceful shutdown procedures as needed.
Closes: #663 

**Breaking Changes (if applicable):**

-   No breaking changes to existing functionalities.

**Additional Information:**

- **Shutdown Functionality**: Implement shutdown functions for web services and connections to gracefully terminate connections and release associated resources.
- **Close Function**: Add a close function to properly close connections and release resources in a controlled manner.
- **User Flexibility**: Enable users to implement custom graceful shutdown procedures tailored to their specific requirements.

**Checklist:**

-   [ ] I have formatted my code using  `goimport`  and  `golangci-lint`.
-   [ ] All new code is covered by unit tests.
-   [ ] This PR does not decrease the overall code coverage.
-   [ ] I have reviewed the code comments and documentation for clarity.

**Thank you for your contribution!**

PR #781, opened 2024-07-03
Add Shutdown and Close Functions for Graceful Connection Termination 2

iamgoroot:
## Pull Request Template

**Description:**

this PR contains few fixes (including review comments) for this PR https://github.com/gofr-dev/gofr/pull/673 which seems to be abandoned but important for future improvements

Breaking Changes (if applicable):

    No breaking changes to existing functionalities.

**Additional Information:**

**Checklist:**


pkg/gofr/gofr.go
```diff
@@ -129,51 +132,35 @@ func (a *App) Run() {
 		a.cmd.Run(a.container)
 	}
 
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+	defer stop()
+
+	go func() {
+		<-ctx.Done()
+
+		err := a.Shutdown(context.Background())
```
ccoVeille, line 141: The shutdown should have a timeout, or you could wait for minutes

pkg/gofr/datasource/redis/redis.go
```diff
@@ -63,6 +63,15 @@ func NewClient(c config.Config, logger datasource.Logger, metrics Metrics) *Redi
 	return &Redis{Client: rc, config: redisConfig, logger: logger}
 }
 
+// Close shutdowns and save the state of the redis client.
+func (r *Redis) Close(ctx context.Context) error {
```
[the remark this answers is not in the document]
iamgoroot: did not fix that from previous contributors PR yet. Planning to. That's why it's still in draft

pkg/gofr/gofr.go
```diff
@@ -203,6 +190,62 @@ func (a *App) Run() {
...
+		return nil
+	})
+
+	*a.httpServer.router.RegisteredRoutes = registeredMethods
+}
+
+// Shutdown stops the service(s) and close the application.
+func (a *App) Shutdown(ctx context.Context) (err error) {
+	if a.httpServer != nil {
+		err = errors.Join(err, a.httpServer.Shutdown(ctx))
```
iamgoroot, line 229: those could probably be shutdown simultaneously

iamgoroot: been thinking something like that

```golang
func (a *App) Shutdown(ctx context.Context) error {
	wg := shutdownGroup{errs: make(chan error)}

	wg.shutdownWhen(ctx,a.httpServer != nil, a.httpServer.Shutdown)
	wg.shutdownWhen(ctx,a.grpcServer != nil, a.grpcServer.Shutdown)
	wg.shutdownWhen(ctx,a.metricServer != nil, a.metricServer.Shutdown)
	wg.shutdownWhen(ctx,a.container != nil, a.container.Close)
	err := wg.WaitAll()

	if err != nil {
		a.container.Logger.Errorf("error while shutting down: %v", err)
		return err
	}

	a.container.Logger.Infof("Application shutdown complete")

	return nil
}

type shutdownGroup struct {
	sync.WaitGroup
	errs chan error
}

func (wg *shutdownGroup) shutdownWhen(ctx context.Context, condition bool, shutdownFunc func(context context.Context) error) {
	if !condition {
		return
	}

	wg.Add(1)

	go func() {
		defer wg.Done()
		wg.errs <- shutdownFunc(ctx)
	}()
}

func (wg *shutdownGroup) WaitAll() (err error) {
	go func() {
		wg.Wait()

		close(wg.errs)
	}()

	for {
		newErr, ok := <-wg.errs
		if !ok {
			return err
		}

		err = errors.Join(err, newErr)
	}
}

```

pkg/gofr/datasource/redis/redis.go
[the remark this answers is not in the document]
ccoVeille: Resolved by

- d57f6163c40adfaba33161aedce8452cfaaffaa5

pkg/gofr/gofr.go
```diff
@@ -203,6 +190,62 @@ func (a *App) Run() {
...
+		err = errors.Join(err, a.grpcServer.Shutdown(ctx))
+	}
+
+	if a.metricServer != nil {
+		err = errors.Join(err, a.metricServer.Shutdown(ctx))
+	}
+
+	err = errors.Join(err, a.container.Close())
+	if err != nil {
+		a.container.Logger.Errorf("error while shutting down: %v", err)
```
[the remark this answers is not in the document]
ccoVeille: Resolved by 506e465be6a676d6288a25c791a663b45c5793a5

pkg/gofr/gofr.go
ccoVeille: Fixed by d57f6163c40adfaba33161aedce8452cfaaffaa5

pkg/gofr/gofr.go
ccoVeille: > been thinking something like that
> 
> ```golang
> func (a *App) Shutdown(ctx context.Context) error {
> 	wg := shutdownGroup{errs: make(chan error)}
> 
> 	wg.shutdownWhen(ctx,a.httpServer != nil, a.httpServer.Shutdown)
> 	wg.shutdownWhen(ctx,a.grpcServer != nil, a.grpcServer.Shutdown)
> 	wg.shutdownWhen(ctx,a.metricServer != nil, a.metricServer.Shutdown)
> 	wg.shutdownWhen(ctx,a.container != nil, a.container.Close)
> 	err := wg.WaitAll()
> 
> 	if err != nil {
> 		a.container.Logger.Errorf("error while shutting down: %v", err)
> 		return err
> 	}
> 
> 	a.container.Logger.Infof("Application shutdown complete")
> 
> 	return nil
> }
> 
> type shutdownGroup struct {
> 	sync.WaitGroup
> 	errs chan error
> }
> 
> func (wg *shutdownGroup) shutdownWhen(ctx context.Context, condition bool, shutdownFunc func(context context.Context) error) {
> 	if !condition {
> 		return
> 	}
> 
> 	wg.Add(1)
> 
> 	go func() {
> 		defer wg.Done()
> 		wg.errs <- shutdownFunc(ctx)
> 	}()
> }
> 
> func (wg *shutdownGroup) WaitAll() (err error) {
> 	go func() {
> 		wg.Wait()
> 
> 		close(wg.errs)
> 	}()
> 
> 	for {
> 		newErr, ok := <-wg.errs
> 		if !ok {
> 			return err
> 		}
> 
> 		err = errors.Join(err, newErr)
> 	}
> }
> 
> ```

Your idea is interesting and complete something I had in mind when I suggested adding the context on all closer

- https://github.com/gofr-dev/gofr/pull/781#discussion_r1664620968

What about doing this


```go
func (a *App) Shutdown(ctx context.Context) error {
	wg := shutdownGroup{errs: make(chan error)}

	wg.shutdownWhen(ctx, a.httpServer)
	wg.shutdownWhen(ctx,a.grpcServer)
	wg.shutdownWhen(ctx,a.metricServer) 
	wg.shutdownWhen(ctx,a.container)
	err := wg.WaitAll()

	if err != nil {
		a.container.Logger.Errorf("error while shutting down: %v", err)
		return err
	}

# 	a.container.Logger.Infof("Application shutdown complete")

	return nil
}

type shutdownGroup struct {
	sync.WaitGroup
	errs chan error
}

type Shutdowner interface {
     Shutdown (context.Context) error
}

func (wg *shutdownGroup) shutdownWhen(ctx context.Context, s Shutdowner) {
	if s == nil {
		return
	}

	wg.Add(1)

	go func() {
		defer wg.Done()
		wg.errs <- s.Shutdown(ctx)
	}()
}

func (wg *shutdownGroup) WaitAll() (err error) {
	go func() {
		wg.Wait()

		close(wg.errs)
	}()

	for {
		newErr, ok := <-wg.errs
		if !ok {
			return err
		}

		err = errors.Join(err, newErr)
	}
}
```

vipul-rawat:
@iamgoroot, we also need to close the subscription manager, stopping all ongoing subscriptions. Stopping only the pubsub client won't work as it will continuously call the subscriber.

https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/subscriber.go

Umang01-hash (changes requested):

pkg/gofr/gofr.go
iamgoroot: @ccoVeille Not all of them called `Shutdown` some are called `Close` and somewhere deeper there's even Close without params. 
For example Reader and Writer in Kafka which should be closed before shutdown as it described in kafka-go docs
Would be nice to move shutdowner somewhere else and reuse that in pubsub implementations to (I'm currently working on closing them as well)
image

iamgoroot: Also handling context cancellation in method WaitAll would be nice too for finishing when timeout is up

iamgoroot:
> @iamgoroot, we also need to close the subscription manager, stopping all ongoing subscriptions. Stopping only the pubsub client won't work as it will continuously call the subscriber.
> 
> https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/subscriber.go

yeah I know and already have some fixes locally. Will commit later

pkg/gofr/datasource/pubsub/mqtt/mqtt_test.go
```diff
@@ -113,6 +114,29 @@ func TestMQTT_Disconnect(t *testing.T) {
...
+	mockMetrics := NewMockMetrics(ctrl)
+
+	mockLogger := logging.NewMockLogger(logging.ERROR)
+	client := New(&Config{}, mockLogger, mockMetrics)
+
+	mockMetrics.EXPECT().
+		IncrementCounter(ctx, "app_pubsub_publish_total_count", "topic", "test")
+
+	// Disconnect the broker and then try to publish
+	client.Close(context.Background())
```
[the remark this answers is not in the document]
iamgoroot: @Umang01-hash No. You should never pass `nil` context anywhere in golang

pkg/gofr/container/datasources.go
```diff
@@ -10,6 +10,8 @@ import (
 	gofrSQL "gofr.dev/pkg/gofr/datasource/sql"
 )
 
+//go:generate go run go.uber.org/mock/mockgen -source=datasources.go -destination=mock_datasources.go -package=container
```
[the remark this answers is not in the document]
iamgoroot: This is golang way to generate code. Read up on this https://go.dev/blog/generate

pkg/gofr/gofr.go
ccoVeille: OK, it's clearer so yes, maybe you will need to find a name that fit your needs and won't conflict with Shutdown or Close

pkg/gofr/container/datasources.go
```diff
@@ -22,12 +24,14 @@ type DB interface {
...
 }
 
 type Redis interface {
 	redis.Cmdable
 	redis.HashCmdable
 	HealthCheck() datasource.Health
+	Close(context.Context) error
 }
 
 type Cassandra interface {
```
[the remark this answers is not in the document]
iamgoroot: https://gofr.dev/docs/advanced-guide/injecting-databases-drivers
As those are created, managed and injected by user, releasing resources should be users responsibility
Feel free to document shutdown procedure for those databases here
Using `defer Close() ` directly in main method should work nicely after merging this PR since it will release app.Run() instead of killing/hanging whole process

This is basically reason for this PR - ability to release resources properly inside gofr so user can release manually allocated resources

pkg/gofr/datasource/pubsub/kafka/kafka_test.go
```diff
@@ -298,7 +301,7 @@ func TestKafkaClient_CloseError(t *testing.T) {
 		logger := logging.NewMockLogger(logging.ERROR)
 		k.logger = logger
 
-		err = k.Close()
+		err = k.Close(context.Background())
```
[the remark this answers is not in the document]
iamgoroot: No. You should never pass nil context anywhere in golang

pkg/gofr/datasource/pubsub/kafka/kafka_test.go
```diff
@@ -271,11 +271,14 @@ func TestKafkaClient_Close(t *testing.T) {
...
-	k := kafkaClient{writer: mockWriter}
+	mockConnection := NewMockConnection(ctrl)
+	mockConnection.EXPECT().Close().Return(nil)
+
+	k := kafkaClient{writer: mockWriter, conn: mockConnection}
 
 	mockWriter.EXPECT().Close().Return(nil)
 
-	err := k.Close()
+	err := k.Close(context.Background())
```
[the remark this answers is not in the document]
iamgoroot: No. You should never pass nil context anywhere in golang

pkg/gofr/gofr.go
iamgoroot: This one is quite obvious

```
type Stopper interface {
   Stop(ctx context.Context) error
}
```
less obvious/easy to miss
```
type Halter interface {
   Halt(ctx context.Context) error
}
```

iamgoroot: I'll probably do that in separate smaller PR. This one is getting big

pkg/gofr/datasource/pubsub/kafka/interfaces.go
```diff
@@ -22,4 +24,5 @@ type Connection interface {
 	Controller() (broker kafka.Broker, err error)
 	CreateTopics(topics ...kafka.TopicConfig) error
 	DeleteTopics(topics ...string) error
+	Close() error
```
ccoVeille, line 27: I would have expected a context here, even if not used

Same for reader

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+func (m *MQTT) getSub(ctx context.Context, topic string) (*subscription, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	// get the message channel for the given topic
+	subs, ok := m.subscriptions[topic]
+	if !ok {
```
ccoVeille, line 38: please avoid this huge if block

```go
if!ok {
 	return &subs, nil
}

// ...
```

pkg/gofr/gofr.go
```diff
@@ -129,51 +135,35 @@ func (a *App) Run() {
 		a.cmd.Run(a.container)
 	}
 
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+	defer stop()
+
+	go func() {
+		<-ctx.Done()
+
+		shutdownCtx, done := context.WithTimeout(context.Background(), shutdownTimeout)
```
ccoVeille, line 144:
```suggestion
		shutdownCtx, done := context.WithTimeout(context.WithoutCancel(ctx), shutdownTimeout)
```

pkg/gofr/gofr_test.go
```diff
@@ -562,6 +562,20 @@ func Test_AddCronJob_Success(t *testing.T) {
...
+	g.GET("/hello", func(*Context) (interface{}, error) {
+		return helloWorld, nil
+	})
+
+	go g.Run()
+	time.Sleep(2 * time.Second)
+
+	err := g.Shutdown(context.Background())
+
+	assert.Nil(t, err, "Test_Shutdown Failed!")
```
ccoVeille, line 577:
```suggestion
	assert.NoError(t, err, "Test_Shutdown Failed!")
```

pkg/gofr/gofr_test.go
```diff
@@ -562,6 +562,20 @@ func Test_AddCronJob_Success(t *testing.T) {
...
 
+func Test_Shutdown(t *testing.T) {
+	g := New()
+
+	g.GET("/hello", func(*Context) (interface{}, error) {
+		return helloWorld, nil
+	})
+
+	go g.Run()
+	time.Sleep(2 * time.Second)
```
ccoVeille, line 573: this delay is huge for tests

100 ms would be enough

pkg/gofr/grpc_test.go
```diff
@@ -51,3 +53,23 @@ func TestGRPC_ServerRun(t *testing.T) {
...
+	go func() {
+		time.Sleep(2 * time.Second)
+		errChan <- g.Shutdown(context.Background())
+	}()
+
+	go g.Run(&c)
+
+	err := <-errChan
+
+	assert.Nil(t, err, "TEST Failed.\n")
```
ccoVeille, line 74:
```suggestion
	assert.NoError(t, err, "TEST Failed.\n")
```

pkg/gofr/httpServer_test.go
```diff
@@ -54,11 +54,42 @@ func TestRun_ServerStartsListening(t *testing.T) {
...
+
+	errChan := make(chan error, 1)
+	go func() {
+		time.Sleep(3 * time.Second)
+		errChan <- server.Shutdown(context.Background())
+	}()
+
+	err := <-errChan
+
+	assert.Nil(t, err, "TEST Failed.\n")
```
ccoVeille, line 86:
```suggestion
	assert.NoError(t, err, "TEST Failed.\n")
```

ccoVeille (approved):

pkg/gofr/datasource/pubsub/kafka/interfaces.go
iamgoroot: it is not gofr interface but reflection of kafka's Connection type. Not doable

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
iamgoroot: thanks

pkg/gofr/gofr_test.go
iamgoroot: There are a lot of leftovers like this. Probably could be done better after implementing readiness probe

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
```diff
@@ -0,0 +1,92 @@
...
+	select {
+	case <-token.Done():
+		if token.Error() != nil {
+			m.logger.Errorf("error getting a message from MQTT, error: %v", token.Error())
+			return &subs, token.Error()
+		}
+
+		m.subscriptions[topic] = subs
+	case <-ctx.Done():
+		return &subs, nil
```
ccoVeille, line 54:
```suggestion
-		return &subs, nil
```

The one at the end is enough

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	return &subs, nil
+}
+
+func (m *MQTT) createMqttHandler(_ context.Context, topic string, msgs chan *pubsub.Message) mqtt.MessageHandler {
```
ccoVeille, line 60:
```suggestion
func (m *MQTT) createMqttHandler(ctx context.Context, topic string, msgs chan *pubsub.Message) mqtt.MessageHandler {
```

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
```diff
@@ -0,0 +1,92 @@
...
+	return func(_ mqtt.Client, msg mqtt.Message) {
+		ctx := context.Background()
```
ccoVeille, line 62:
```suggestion
		ctx := context.WithoutCancel(ctx)
```

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -60,33 +59,41 @@ type subscription struct {
 // New establishes a connection to MQTT Broker using the configs and return pubsub.MqttPublisherSubscriber
 // with more MQTT focused functionalities related to subscribing(push), unsubscribing and disconnecting from broker.
 func New(config *Config, logger Logger, metrics Metrics) *MQTT {
+	var options *mqtt.ClientOptions
```
ccoVeille, line 62:
```diff
-        var options *mqtt.ClientOptions 
-          if config.Hostname == "" { 
-                 options = getDefaultClientOpts(config) 
-         } else { 
-                 options = getMQTTClientOptions(config) 
-          }
+options := getDefaultClientOpts(config)
+if config.Hostname != "" { 
+options = getMQTTClientOptions(config) 
+}
```

ccoVeille (approved):
LGTM 👍

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
```diff
@@ -50,16 +50,14 @@ func (m *MQTT) getSub(ctx context.Context, topic string) (*subscription, error)
 		}
 
 		m.subscriptions[topic] = subs
-	case <-ctx.Done():
```
iamgoroot: @Umang01-hash Any particular reason you removed this context expiry check? This doesn't seem right

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
```diff
@@ -0,0 +1,95 @@
+package mqtt
```
ccoVeille, line 1: It is me or this file as a space in its name, instead of an underscore?

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
Umang01-hash: @iamgoroot did the changes thinking about your review comment:
`The one at the end is enough`

I am sorry if it wasn't according to the way you wanted.
Please feel free to revert it or make further changes as required

pkg/gofr/grpc_test.go
```diff
@@ -51,3 +53,23 @@ func TestGRPC_ServerRun(t *testing.T) {
...
+func TestGRPC_ServerShutdown(t *testing.T) {
+	c := container.Container{
+		Logger: logging.NewLogger(logging.DEBUG),
+	}
+
+	g := newGRPCServer(&c, 9999)
+
+	var errChan = make(chan error, 1)
+	go func() {
+		time.Sleep(2 * time.Second)
```
ccoVeille, line 66: 2 seconds is too long for tests

pkg/gofr/httpServer.go
```diff
@@ -1,6 +1,7 @@
 package gofr
```
ccoVeille, line 1: Not related to this PR, but I don't think this filename should be in camelCase.

others are using snake case, no?

pkg/gofr/httpServer_test.go
```diff
@@ -54,11 +54,42 @@ func TestRun_ServerStartsListening(t *testing.T) {
...
+		router: router,
+		port:   8080,
+	}
+
+	// Start the server
+	go server.Run(c)
+
+	errChan := make(chan error, 1)
+	go func() {
+		time.Sleep(3 * time.Second)
```
ccoVeille, line 80: 3 seconds is like ages for tests

pkg/gofr/subscriber.go
```diff
@@ -24,35 +25,49 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
-		err = func(ctx *Context) error {
-			// TODO : Move panic recovery at central location which will manage for all the different cases.
-			defer panicRecovery(ctx.Logger)
-			return handler(ctx)
-		}(ctx)
+func (s *SubscriptionManager) handleSubscription(topic string, handler SubscribeFunc, msg *pubsub.Message, err error) {
+	if errors.Is(err, kafka.ErrConsumerGroupNotProvided) {
+		s.container.Logger.Errorf("cannot subscribe as consumer_id is not provided in configs")
+		return
+	} else if err != nil {
```
ccoVeille, line 47:
```suggestion
	}

if err != nil {
```
The return in the previous if allows to keep the code simple

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
ccoVeille: I think it was my comment 

My point was to remove the return after receiving the ctx.Done channel

Not to removing the ctx.Done channel in the select

Umang01-hash: Ohhh @ccoVeille got it.
Sorry for the confusion

pkg/gofr/subscriber.go
iamgoroot: I tried not to change most code unrelated to shutdown. But this PR is getting so big that I'm afraid it's not gonna be merge at all

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
iamgoroot: Thanks. Been having some weird rendering issues in vs code

pkg/gofr/grpc_test.go
iamgoroot: This is probably leftovers from previous contributor. Feel free to fix it up if you have spare time today

pkg/gofr/httpServer.go
iamgoroot: There are a lot of those it a current code base. Let's not fix everything in single PR

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
vipul-rawat: I was also wondering why the file is having a space 😅

pkg/gofr/grpc_test.go
ccoVeille: I'm opening an issue then

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
ccoVeille: I created 

- https://github.com/gofr-dev/gofr/issues/794#issue-2392254779

pkg/gofr/httpServer.go
ccoVeille: I created 

- https://github.com/gofr-dev/gofr/issues/794#issue-2392254779

pkg/gofr/grpc_test.go
ccoVeille: I created
- https://github.com/gofr-dev/gofr/issues/795

pkg/gofr/httpServer_test.go
iamgoroot: Thanks

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -59,10 +59,8 @@ type subscription struct {
 // New establishes a connection to MQTT Broker using the configs and return pubsub.MqttPublisherSubscriber
 // with more MQTT focused functionalities related to subscribing(push), unsubscribing and disconnecting from broker.
 func New(config *Config, logger Logger, metrics Metrics) *MQTT {
-	var options *mqtt.ClientOptions
```
iamgoroot: @Umang01-hash This one breaks a test

ccoVeille: Oups, it might be because of the changes I suggested

🤔😱

iamgoroot: @ccoVeille there's problem with getDefaultClientOpts that's mutable. Feel free to improve on that

iamgoroot: I with there was a linter like "Don't mutate anything in method caller getSomething(*Obj)"

pkg/gofr/datasource/pubsub/mqtt/mqtt sub.go
iamgoroot: fixed. Thanks for vigilance

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
ccoVeille: Indeed

ccoVeille (approved):

pkg/gofr/subscriber.go
```diff
@@ -24,38 +25,54 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
 	}
 }
 
-func (s *SubscriptionManager) startSubscriber(topic string, handler SubscribeFunc) {
+func (s *SubscriptionManager) startSubscriber(ctx context.Context, topic string, handler SubscribeFunc) {
```
vipul-rawat, line 28: In gofr, we try to have the same context for every request-response cycle. We tried to do the same in the subscriber where every subscription call is called as a part of the request-response.

Using the same context for every subscription call seems wrong in my opinion, we should be creating a new context maybe taking the parent context and not directly passing it to the `Subscribe()` function.

iamgoroot: This ctx is for managing cancellation. if we need new ctx for req/resp cycle - it should be done in the loop below.

Didn't do it before because mqtt implementation did create new context for every message but as I see kafka doesn't so I'll create new ctx inside the loop

iamgoroot: it is may cause memory leak when using defer in infinite loop. i'll move some logic into other method to handle that

pkg/gofr/container/container.go
```diff
@@ -45,6 +47,19 @@ type Container struct {
 	File datasource.FileSystem
 }
 
+func (c *Container) Close(ctx context.Context) error {
+	var err error
+	if c.PubSub != nil {
```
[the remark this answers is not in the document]
iamgoroot: done

pkg/gofr/subscriber.go
```diff
@@ -24,38 +25,61 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
+}
+
+func (s *SubscriptionManager) processSubscription(parentCtx context.Context, topic string, handler SubscribeFunc) {
+	ctx, done := context.WithCancel(parentCtx)
+	defer done()
+
+	msg, err := s.container.GetSubscriber().Subscribe(ctx, topic)
+
+	select {
+	case <-ctx.Done():
```
ccoVeille, line 42: I disagree with the changes you have just made.

Unless I missed something. You removed the code that was allowing to exit the for loop. Now you will infinitely logging this error.

Am I wrong?

ccoVeille (changes requested):

pkg/gofr/subscriber.go
iamgoroot: No you are not wrong. Thats why test failed. Been fixing before seeing your comment. Thanks for vigilance

iamgoroot: @ccoVeille it is fixed now. test passed

pkg/gofr/subscriber.go
```diff
@@ -24,38 +24,57 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
-			s.container.Logger.Errorf("error in handler for topic %s: %v", topic, err)
-		}
+func (s *SubscriptionManager) handleSubscription(parentCtx context.Context, topic string, handler SubscribeFunc) {
+	ctx, done := context.WithCancel(parentCtx)
+	defer done()
+
+	msg, err := s.container.GetSubscriber().Subscribe(ctx, topic)
+
+	if errors.Is(err, kafka.ErrConsumerGroupNotProvided) {
+		s.container.Logger.Errorf("cannot subscribe as consumer_id is not provided in configs")
```
ccoVeille, line 47: This one doesn't need an Errorf


```suggestion
		s.container.Logger.Error("cannot subscribe as consumer_id is not provided in configs")
```

pkg/gofr/subscriber.go
```diff
@@ -24,38 +24,57 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
-		} else {
-			s.container.Logger.Errorf("error in handler for topic %s: %v", topic, err)
-		}
+func (s *SubscriptionManager) handleSubscription(parentCtx context.Context, topic string, handler SubscribeFunc) {
+	ctx, done := context.WithCancel(parentCtx)
+	defer done()
+
+	msg, err := s.container.GetSubscriber().Subscribe(ctx, topic)
+
+	if errors.Is(err, kafka.ErrConsumerGroupNotProvided) {
```
ccoVeille, line 46: This one would cause an infinite loop.

I mean if the topic is not provided, on next loop it won't be there either no?

You should consider returning error in this function, and catching when one is returned

pkg/gofr/subscriber.go
```diff
@@ -24,38 +24,57 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
-		}(ctx)
-
-		// commit the message if the subscription function does not return error
-		if err == nil {
-			msg.Commit()
-		} else {
-			s.container.Logger.Errorf("error in handler for topic %s: %v", topic, err)
-		}
+func (s *SubscriptionManager) handleSubscription(parentCtx context.Context, topic string, handler SubscribeFunc) {
+	ctx, done := context.WithCancel(parentCtx)
```
ccoVeille, line 41: I don't get this. When would you need to cancel the context?

Thebobly moment is when you are leaving this function. But you have no go routine.

iamgoroot: this is correct. let me fix that

pkg/gofr/subscriber.go
iamgoroot: yep. there are probably a lot more errors like this that are critical and should finish consumer. 
Actually those should probably go way up and finish process altogether

pkg/gofr/subscriber.go
```diff
@@ -24,39 +24,62 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
-		}(ctx)
-
-		// commit the message if the subscription function does not return error
-		if err == nil {
-			msg.Commit()
-		} else {
-			s.container.Logger.Errorf("error in handler for topic %s: %v", topic, err)
-		}
+func (s *SubscriptionManager) handleSubscription(parentCtx context.Context, topic string, handler SubscribeFunc) error {
+	ctx := context.WithoutCancel(parentCtx)
```
ccoVeille, line 44: But then, you will never receive global context cancelation when app stops?

I'm unsure what you are trying to achieve with this change/PR

pkg/gofr/subscriber.go
ccoVeille: Thanks for doing the changes

pkg/gofr/subscriber.go
iamgoroot: global cancellation goes way up in call stack, this should be fine for handling single message. 
maybe i've been wrong thinking in terms of mqtt where context cancellation have no effect. but looking into kafka this may not be a good idea

pkg/gofr/subscriber.go
iamgoroot: this can put a user into situation where app is still running but consumer thread in down.
Should be improved upon even more.

pkg/gofr/subscriber.go
iamgoroot: or maybe you can provide clearer concern about previous change
`	ctx, done := context.WithCancel(parentCtx)`
seems correct if you lookup kafka implementation

ccoVeille: Let me look again at the code

ccoVeille: The more I look, the more awkward it looks like.

Some subscribers are using context without cancel in Subscriber

I would have suggested to cancel the subscriber context. But it sounds a but strange.

Someone who have a clearer idea of the whole code base will have to look at your PR and tell what should be done.

The implementation depends on how long a subscribe is expect to last. Also, I don't get why the code keep subscribing and do jot consume a subscribed resource

iamgoroot: I don't think there's clear understanding among contributors given that this PR exists
Some code seems to be thrown in just to "have support of smth" (that I derived from some tickets here like "NATS support")
If you're interested we can do discord call to review some related code some time later

pkg/gofr/subscriber.go
vipul-rawat: So since we should be creating a new context start in the for loop, we can later see if we can remove the `ctx := context.Background()` in the mqtt implementation, just the tracing context should be enough.

pkg/gofr/subscriber.go
vipul-rawat: We don't want to shut down the application when a subscriber cannot start. We kept in mind that a subscriber can be a part of a rest application where there are some endpoints registered, we did not want to make the server unreachable if there is a failure in case of a subscription. If only subscribers are registered, the server should not start and should shut down, but not otherwise. This was our design philosophy as we have seen such implementation frequently.

ccoVeille (approved):
I'm still wondering what you try to achieve with the cancelable context in startSubscriber. Because for me, the current implementation doesn't need a cancelable context. I would have understood a timeout but not a cancelable context while there is no go routine involved.

But, the code is way better now than it was.

So I would say go for it.

iamgoroot:
> I'm still wondering what you try to achieve with the cancelable context in startSubscriber. Because for me, the current implementation doesn't need a cancelable context. I would have understood a timeout but not a cancelable context while there is no go routine involved.
> 
> But, the code is way better now than it was.
> 
> So I would say go for it.

It may look that way. But there's dependencies and they have code too. Did you have a chance to look deeper into those? For example Kafka client implementation somewhere here

https://github.com/segmentio/kafka-go/blob/7b9c99dd6562f80bd3fc447305115029a7154382/reader.go#L793

Anyway I'm not interested in supporting this anymore. It has signs of merge/explain loop for feature users of opinionated framework have no use for

ccoVeille:
I'm sorry if my comments about your PRs led to closing it.

Your PR was great, and I think you were very close to have something production ready.

But I also understand that a PR with numerous comments and questions about the implementation on almost a week is a pain.

PR #793, opened 2024-07-05
fix error logic for SubscribeWithFunc calls

vipul-rawat:


ccoVeille (approved):

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -318,7 +319,15 @@ func getHandler(subscribeFunc SubscribeFunc) func(client mqtt.Client, msg mqtt.M
 		}
 
 		// call the user defined function
-		_ = subscribeFunc(pubsubMsg)
+		err := subscribeFunc(pubsubMsg)
```
iamgoroot, line 322: @vipul-rawat This seems to conflict with change I had to made due to code climate requirements (file was too long by 6 lines or so)
https://github.com/gofr-dev/gofr/pull/781/files#diff-01c8424bf0ca548fdc0070da34c818b40d70a65d64e08e10b7816bef1fea8f0e

Can we do this in single PR or so we don't lose this important fix in a merge?

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -318,7 +319,15 @@ func getHandler(subscribeFunc SubscribeFunc) func(client mqtt.Client, msg mqtt.M
...
+		if err != nil {
+			m.logger.Errorf("error subscribing to message, err : %v", err)
+
+			return
+		}
+
+		// acknowledge messge if subscription is successful
+		msg.Ack()
```
iamgoroot, line 330: There's another way to ack messages now. It's done one level above in subscription manager using Commiter interface (see mqtt/message.go)
Commit already made here
https://github.com/gofr-dev/gofr/blob/63ebe9e9a6dc9162f36eb5c5792cda0d71e873dd/pkg/gofr/subscriber.go#L52

vipul-rawat: The function is `SubscribeWithFunc`, an extended version of `Subscribe`. It is not called from the subscription manager, so the messages are not acknowledged, but this made me realise that the `pubsub.Message` created here was not populating the Commiter interface.

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
vipul-rawat: What do you suggest? In your PR #781 this function `SubscribeWithFunction` is already untouched. Shall we move this to `mqtt_sub.go`?

vipul-rawat: Or I can move the extended functionalities of the MQTT client to a new file

iamgoroot: @vipul-rawat That huge PR size becoming pain in a lower back. Can you wait for it or commit right into that branch?

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
iamgoroot: I'll try to debug later for better understanding what happens

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
vipul-rawat: I'll fix this after shutdown PR gets merged!

vipul-rawat:
A lot of new changes need to be addressed! This would be finalized after #837 is merged.

Issue #783, opened 2024-07-04
Enhance error handling


srijan-27:
- Error shouldn't be singular, instead it should be an array.
- Currently, error only have message, it should also have some details, and date time

PR #797, opened 2024-07-05
Enhance errors with multiple and custom errors

srijan-27:
- Add multiple errors support. Hence, error response is now an array.
- Add custom errors support, with statusCode, reason and details.
- Closes: #783


examples/http-server/main_test.go
```diff
@@ -108,9 +108,7 @@ func TestIntegration_SimpleAPIServer_Errors(t *testing.T) {
 
 		_ = json.Unmarshal(b, &data)
 
-		assert.Equal(t, tc.body, data.Error, "TEST[%d], Failed.\n%s", i, tc.desc)
-
-		assert.Nil(t, err, "TEST[%d], Failed.\n%s", i, tc.desc)
+		assert.Contains(t, fmt.Sprintf("%v", data.Errors), tc.body, "TEST[%d], Failed.\n%s", i, tc.desc)
```
ccoVeille, line 111: I'm curious if this doesn't work

```go
assert.Contains(t, data.Errors, tc.body, "TEST[%d], Failed.\n%s", i, tc.desc)
```

pkg/gofr/http/errors_test.go
```diff
@@ -111,3 +111,19 @@ func Test_ErrorErrorPanicRecovery(t *testing.T) {
...
+
+	assert.Equal(t, "route not registered", err.Error(), "TEST Failed.\n")
+
+	assert.Equal(t, http.StatusNotFound, err.StatusCode(), "TEST Failed.\n")
+}
+
+func Test_CustomError(t *testing.T) {
+	err := CustomError{Code: http.StatusUnauthorized, Reason: "request unauthorized"}
+
+	assert.Equal(t, "request unauthorized", err.Error(), "TEST Failed.\n")
```
ccoVeille, line 126:
```suggestion
assert.ErrorContains(t, err, "request unauthorized", , "TEST Failed.\n")
```

otherwise it would fail with a panic if err is nil

pkg/gofr/http/errors_test.go
```diff
@@ -111,3 +111,19 @@ func Test_ErrorErrorPanicRecovery(t *testing.T) {
 
 	assert.Equal(t, http.StatusInternalServerError, err.StatusCode(), "TEST Failed.\n")
 }
+
+func Test_MultipleErrors(t *testing.T) {
+	err := MultipleErrors{Code: http.StatusNotFound, Errors: []error{ErrorInvalidRoute{}}}
+
+	assert.Equal(t, "route not registered", err.Error(), "TEST Failed.\n")
```
ccoVeille, line 118:
```suggestion
assert.ErrorContains(t, err, "route not registered", , "TEST Failed.\n")
```

otherwise it would fail with a panic if err is nil

pkg/gofr/http/responder.go
```diff
@@ -49,40 +69,67 @@ func (r Responder) Respond(data interface{}, err error) {
...
-type statusCodeResponder interface {
-	StatusCode() int
+func getErrResponse(err error) []errResponse {
+	var (
+		errResp []errResponse
+		m       MultipleErrors
+		c       CustomError
+	)
+
+	if err != nil {
```
ccoVeille, line 113:
```suggestion
	if err == nil {
	    return nil
	}
```

pkg/gofr/http/errors_test.go
srijan-27: err is defined on line 116, so it can't be nil.

pkg/gofr/http/responder.go
srijan-27: please go through the logic

pkg/gofr/http/errors.go
```diff
@@ -7,6 +7,39 @@ import (
 	"strings"
 )
 
+type MultipleErrors struct {
+	Code   int
+	Errors []error
+}
+
+func (m MultipleErrors) Error() string {
```
ccoVeille, line 15: What about this?

```diff
func (m MultipleErrors) Error() string { 
-         var result string 
+        var result []string
          for _, v := range m.Errors { 
-                 result += fmt.Sprintf("%s\n", v) 
+               result = append(result, fmt.Sprint(v)
         } 
  
-         return strings.TrimSuffix(result, "\n") 
+        return strings.Join(result, "\n")
 }
```

pkg/gofr/http/errors_test.go
ccoVeille: Oh you're not calling a method but testing the Error() I get it now, I'm sorry

pkg/gofr/http/responder.go
ccoVeille: I dislike function wide if branch, but OK. Everyone has his style

pkg/gofr/http/errors.go
srijan-27: Although both are similar implementation wise. But when I ran the benchmark tests for both these implementations, I observed that the current one (result as `string`) is `much faster` than the suggested one (result as `[]string`). Attaching the results below:

Screenshot 2024-07-08 at 11 18 52 AM

Screenshot 2024-07-08 at 11 21 47 AM

aryanmehrotra (approved):

pkg/gofr/http/errors.go
ccoVeille: Thanks for doing the benchmark.

For me, it was more a question of code simplicity or clarification than performance.

Here we are in an error, so it occurs rarely. Performance is not the priority I think.

Anyway, I'm fine with both.

ccoVeille (approved):

Umang01-hash (approved):

vipul-rawat (approved):

aryanmehrotra (approved):

Umang01-hash (approved):

Issue #787, opened 2024-07-04
Help needed to enable more golangci-linter


ccoVeille:
Here is the list of errors I was able to detect via golangci-lint by enabling a lot of rules

```
951 issues:
* contextcheck: 12
* depguard: 2
* errorlint: 1
* gci: 22
* mirror: 1
* sqlclosecheck: 12
* testifylint: 271
* thelper: 16
* usestdlibvars: 19
* revive: 595
    425 use-any
     89 unused-receiver
     31 import-alias-naming
     24 unhandled-error
      5 unchecked-type-assertion
      5 import-shadowing
      3 early-return
      3 defer
      3 comment-spacings
      2 deep-exit
      2 confusing-naming
      1 modifies-value-receiver
      1 bool-literal-in-expr
      1 bare-return
```

Please let me know if you want me to help you to add them, I can work on a dedicated file and a dedicated GHA, so it won't cause any problem when merged

iamgoroot:
There are some really important lints that should be enabled. Especially after looking into the code...
Can this be enabled for new code for now? 
I think current issues can be taken care of later and it seems like gofr has "just throw in support of everything you can think of" in backlog so new code should definitely be checked

ccoVeille:
Yes, I can enable them for new code only, that's pretty easy I did it for clipse project

- https://github.com/savedra1/clipse/pull/84/files

aryanmehrotra:
Minor linters can be fixed in a single PR, but the larger one's needed to be added separately.
We would prefer adding with the fix/

ccoVeille:
#788 addresses a part of the issues, once merged more changes will be needed

ccoVeille:
I will create a PR about using testifylint with  updated code, not whole codebase

srijan-27:
@ccoVeille - should I create a PR to add the above specified 3 linters?

ccoVeille:
Yes, please.

I think they should be addressed in separate PR to ease the review

Some might be obvious and can be grouped, some might require large code rewrite.

You will see by working on them.

srijan-27:
Sure, yeah I am aware of changes.. just wanted to confirm if you worked on it.

ccoVeille:
Let's close this old issue.

There is still a need, but the values reported for each linter are outdated

PR #829, opened 2024-07-09
ci: fix testifylint errors and enable it

ccoVeille:
- **ci: enable testify and fix reported errors**

Related to #787


Umang01-hash (approved):

srijan-27:
@ccoVeille - Wanted to confirm few things.. I may be wrong in this, but what my understanding is, that `assert` waits for all the test statements to run before it fails, whereas `require` fails immediately.
If this is correct, then:
1. shouldn't we use `require.NoError` before `assert.Equal` in all the tests.
2. also, if NoError is the last statement in tests, then how would `require.NoError` would be beneficial from `assert.Equal`

ccoVeille:
> @ccoVeille - Wanted to confirm few things.. I may be wrong in this, but what my understanding is, that `assert` waits for all the test statements to run before it fails, whereas `require` fails immediately.
> If this is correct, then:
> 1. shouldn't we use `require.NoError` before `assert.Equal` in all the tests.

```go
         resp, err := client.SayHello(context.Background(), nil) 
          assert.Equal(t, "Hello World!", resp.Message) 
         require.NoError(t, err)
```

I assume you are talk it about code like this.

Here depending on the code, you may have a nil pointer for `resp`, so it would panic. So inverting would be safer

If we consider the go way, the first thing we should catch is the error.

While this is valid in code, it's open for debate in tests.

In other use case, it's open for debate.

Imagine a code that is not supposed to fail. So you use a NoError, you will have an error, but if you didn't check the returned value that might be meaningful like a JSON error message. It's important to check it before checking the NoError, so you have information about it.

cc @mmorel-35, who also contributes to testifylint, and we had a recent discussion about this

ccoVeille:
> also, if NoError is the last statement in tests, then how would require.NoError would be beneficial from assert.Equal

I'm unsure if testifylint is currently able to detect that.

I had to do a search and replace with sed to fix them, because of the lack of autofix. See:

- https://github.com/Antonboom/testifylint/discussions/147

We will have to check with @mmorel-35

ccoVeille:
@srijan-27 I don't have a computer with me at my holidays place. If you would like to review this PR and reorder the require.Error*/NoError* to be before the other when it makes sense, be my guest, and push it here.

I will review your changes

srijan-27:
@ccoVeille - Sure, let me check all the places.. will update accordingly.

ccoVeille:
Thanks @srijan-27

srijan-27:
@ccoVeille - Since, we are replacing `Assert` with `Require`, so why just restrict ourselves to error cases only, why not replace for response test as well? Or would you like to have those changes in separate PR as they would not be related to linter `testifylint`.

ccoVeille:
> @ccoVeille - Since, we are replacing `Assert` with `Require`, so why just restrict ourselves to error cases only, why not replace for response test as well? Or would you like to have those changes in separate PR as they would not be related to linter `testifylint`.

The idea in using require instead of assert for Errors asserters are fixing multiple cause of bugs.

Let's assume you are testing that you are facing no error when calling something.

So you have a `assert.Error(t, err)`, `err` comes from a call to a function.

Then you will check for the behavior you expect. So when you have no error.

There is usually no need to check further when an error occurs.

So using require will avoid useless checks. 

**This is the first reason.**

**But then comes the second reason.**

Code are often using asserter the wrong way (testifylint tries to sort some issue), especially when code are expecting an error.

Let's consider this piece of code

```go
assert.Error(t, err)
assert.Contains(t, err.Error(), "whatever")
```

Here the code will panic when there is no Error, as the call to err.Error() will be made on a nil error.

testifylint could suggest to use

```go
assert.Error(t, err)
assert.ErrorContains(t, err, "whatever")
```

But it's already an advanced test, and testifylint will never be able to detect code like this

```go
if condition {
     assert.Contains(t, err.Error(),  "whatever")
}
```

where `condition` could be something very complex, in a loop, sub-function or a helper.
I hope it's clearer

You can find more explanation [where the testifylint checker is described.](https://github.com/Antonboom/testifylint?tab=readme-ov-file#require-error)

cc @mmorel-35 who also maintains testifylint

srijan-27:
@ccoVeille - Yes, I understand your point. I agree that using require would be better than using assert.
But, again my point was shouldn't we replace `assert.Equal`to `require.Equal` as well?

ccoVeille:
It would work of course, but I wouldn't recommend.

Using assert allows to report multiple errors

Please consider something like this

Extracted from gofr code

```go
                statusCode, errObj := getStatusCode(tc.method, tc.data, tc.err)

                assert.Equal(t, tc.statusCode, statusCode, "TEST[%d], Failed.\n%s", i, tc.desc)

                assert.Equal(t, tc.errObj, errObj, "TEST[%d], Failed.\n%s", i, tc.desc)
```

Imagine this test is now failing because of code changes

The first asserter check the status code, the other the error available.

If the code changes are about a validation error that will return a 400, with associated error such as: invalid email, too long, not matching a new condition.

When using assert.Equal you will have a test error reporting the status code and the error are not the one reported.

So you now, that there is a 400, plus what you need to change/catch in the test

Now let's consider we are using a require.Equal for the status code and error.

You will only know you it's a validation error, the next require.Error won't be checked. So you will no idea what error it could be.

The only way to know would be to add fmt.Println everywhere, or remove temporarily the first require.Equal, or of course run the test with debugger.

These are boring thing that require extra works. It may cause issue as you might wrongly commit the things you added in the tests for debugging purpose.

Using assert allows you to get context when a test unexpectedly fails. If you use require, you won't have the context.

srijan-27:
Yeah, since require have fail fast mechanism we would be missing out on extra info that we are now getting from using assert. Same, would be lost for error cases as well.
So, you are suggesting that we should use require only for no error cases, since there we could be having unexpected behaviours.

srijan-27 (approved):

vipul-rawat (approved):

Issue #795, opened 2024-07-05
Fix too long time.Sleep in tests


ccoVeille:
> 2 seconds is too long for tests

_Originally posted by @ccoVeille in https://github.com/gofr-dev/gofr/pull/781#discussion_r1666561069_

> 3 seconds is like ages for tests

_Originally posted by @ccoVeille in https://github.com/gofr-dev/gofr/pull/781#discussion_r1666563213_

srijan-27:
@ccoVeille - I believe this is fixed now as long sleep timers have been removed.
Let me know if you think anything is missed, or if this can be closed.

ccoVeille:
I'm curious if we can see a difference in term of CI duration.

Before starting
After the time.Sleep fix
After the cronjob seconds patch

srijan-27:
As per my observations, previously package and example tests were taking more than 2 mins and around 2 mins respectively.
But after the fixes, both the pipelines are running in around 1 min, which is 50% more optimised.

PR #831, opened 2024-07-09
ci: reduce arbitrary long time.Sleep

ccoVeille:
Related to #795


ccoVeille:
there is so much oddities in the code ...

I will add TODO wherever I find them

Umang01-hash:
@ccoVeille I noticed that there’s a draft PR open, and I’m interested in contributing. Would you be comfortable with me taking it up while you’re away? I’d be happy to help.

Thank you!

ccoVeille:
Tes! Please !

You can start from where I was and push commits, or close this PR, and open a new one.

pkg/gofr/service/health_test.go
```diff
@@ -82,7 +82,7 @@ func TestHTTPService_HealthCheckTimeout(t *testing.T) {
 	server := httptest.NewServer(http.HandlerFunc(func(_ http.ResponseWriter, r *http.Request) {
 		assert.Equal(t, "/.well-known/alive", r.URL.Path)
 
-		time.Sleep(2 * time.Second)
+		time.Sleep(1000 * time.Millisecond)
```
srijan-27, line 85: instead of 1000 millisecond, its more realistic to use 1 sec

pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
```diff
@@ -23,15 +23,16 @@ func TestRemoteLogger_UpdateLevel(t *testing.T) {
 	}))
 
 	rl := remoteLogger{
-		remoteURL:          mockServer.URL,
-		levelFetchInterval: 1,
+		remoteURL: mockServer.URL,
+		// TODO: use a time.Duration for levelFetchInterval
```
srijan-27, line 27: should we fix this TODO in this PR itself?

pkg/gofr/handler_test.go
```diff
@@ -64,11 +64,12 @@ func TestHandler_ServeHTTP_Timeout(t *testing.T) {
 	w := httptest.NewRecorder()
 	r := httptest.NewRequest(http.MethodGet, "/", http.NoBody)
 
-	h := handler{requestTimeout: "1"}
+	// TODO: use a time.Duration for requestTimeout
```
srijan-27, line 67: should we fix this TODO in this PR itself?

srijan-27 (changes requested):

pkg/gofr/service/health_test.go
ccoVeille: I was about to say it yes

Umang01-hash: Sure changed it to second

pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
ccoVeille: I think so, because this int in seconds implies that the tests cannot be shorter than 1s

examples/using-cron-jobs/main_test.go
```diff
@@ -8,6 +8,7 @@ import (
 )
 
 func Test_UserPurgeCron(t *testing.T) {
+	// TODO find a way to do not have to wait for more than a minute
```
ccoVeille, line 11: This one may need a separate PR.

We have to use a wait to speed up the clock.

There are multiple lib for that, but it requires some research and tests

pkg/gofr/http/middleware/oauth.go
```diff
@@ -54,6 +54,7 @@ func NewOAuth(config OauthConfigs) PublicKeyProvider {
 	var publicKeys PublicKeys
 
 	go func() {
+		// TODO update the keys when starting
```
ccoVeille, line 57: This is about calling updateKeys without waiting the first ticker.

It will help to provide shorter delay for running the tests

But as the type if refresh interval is now a duration maybe it's not needed.

It's worth a double check

ccoVeille:
I added some context to the TODO I had added for myself, but as someone else is working on it. It may help

examples/using-subscriber/main_test.go
```diff
@@ -41,10 +41,12 @@ func initializeTest(t *testing.T) {
 func TestExampleSubscriber(t *testing.T) {
 	log := testutil.StdoutOutputForFunc(func() {
 		go main()
+
 		time.Sleep(time.Second * 1) // Giving some time to start the server
 
 		initializeTest(t)
 		time.Sleep(time.Second * 20) // Giving some time to publish events
```
ccoVeille, line 48: I had updated them

I assume they were rolled back as the tests were failing, I'm fine with it.

But then I would expect to see a TODO here

Maybe it could be fixed in this PR, maybe in another one

pkg/gofr/handler_test.go
```diff
@@ -65,11 +65,12 @@ func TestHandler_ServeHTTP_Timeout(t *testing.T) {
...
+	h := handler{requestTimeout: "1"} // the minimum value is 1s
 
 	h.container = &container.Container{Logger: logging.NewLogger(logging.FATAL)}
 	h.function = func(*Context) (interface{}, error) {
-		time.Sleep(2 * time.Second)
+		time.Sleep(1200 * time.Millisecond)
```
ccoVeille, line 73: Here you can see the link between a delay that cannot be shorter than 1s and the 1200ms we have to wait

ccoVeille (commented):
Except the cron that need to be refactored to help testing it in a delay shorter than 1min, the PR should look at the duration currently defined as a int to use time.Duration

It will allow to remove all the 1200ms, and use something way shorter

pkg/gofr/handler_test.go
Umang01-hash: done

pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
Umang01-hash: done

examples/using-cron-jobs/main_test.go
Umang01-hash: it needs to be handled in a separate PR

examples/using-subscriber/main_test.go
Umang01-hash: This also will be taken care in another PR

Umang01-hash:
> Except the cron that need to be refactored to help testing it in a delay shorter than 1min, the PR should look at the duration currently defined as a int to use time.Duration
> 
> It will allow to remove all the 1200ms, and use something way shorter

I changed int to `time.Duration` in this PR but still was not able to remove 1200 ms sleep. The test failed.

pkg/gofr/handler_test.go
Umang01-hash: i see this linking but didn't figured out how to reduce this sleep

examples/using-subscriber/main_test.go
```diff
@@ -56,7 +56,7 @@ func TestExampleSubscriber(t *testing.T) {
 			expectedLog: "Received order",
 		},
 		{
-			desc:        "valid product",
+			desc:        "valid  product",
```
ccoVeille, line 59: Strange

pkg/gofr/handler.go
```diff
@@ -46,10 +45,8 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if websocket.IsWebSocketUpgrade(r) {
 		// If the request is a WebSocket upgrade, do not apply the timeout
 		c.Context = r.Context()
-	} else if h.requestTimeout != "" {
-		reqTimeout := h.setContextTimeout(h.requestTimeout)
-
-		ctx, cancel := context.WithTimeout(r.Context(), time.Duration(reqTimeout)*time.Second)
+	} else if h.requestTimeout != 0 {
+		ctx, cancel := context.WithTimeout(r.Context(), h.requestTimeout*time.Second)
```
ccoVeille, line 49: Here is why you weren't able to reduce the time


```suggestion
		ctx, cancel := context.WithTimeout(r.Context(), h.requestTimeout)
```

Here you virtually created a timeout equal to the number on nanoseconds multiplied by 1 s

pkg/gofr/handler_test.go
```diff
@@ -65,11 +65,11 @@ func TestHandler_ServeHTTP_Timeout(t *testing.T) {
 	w := httptest.NewRecorder()
 	r := httptest.NewRequest(http.MethodGet, "/", http.NoBody)
 
-	h := handler{requestTimeout: "1"}
+	h := handler{requestTimeout: time.Duration(1)} // the minimum value is 1s
```
ccoVeille, line 68:
```suggestion
	h := handler{requestTimeout: 100* time.Millisecond} 
```

pkg/gofr/logging/remotelogger/dynamic_level_logger.go
```diff
@@ -30,7 +30,7 @@ func New(level logging.Level, remoteConfigURL, loggerFetchInterval string) loggi
 	l := remoteLogger{
 		remoteURL:          remoteConfigURL,
 		Logger:             logging.NewLogger(level),
-		levelFetchInterval: interval,
+		levelFetchInterval: time.Duration(interval),
```
ccoVeille, line 33:
```suggestion
		levelFetchInterval: interval * time.Second,
```

pkg/gofr/logging/remotelogger/dynamic_level_logger.go
```diff
@@ -43,15 +43,15 @@ func New(level logging.Level, remoteConfigURL, loggerFetchInterval string) loggi
...
+	levelFetchInterval time.Duration
 	currentLevel       logging.Level
 	logging.Logger
 }
 
 // UpdateLogLevel continuously fetches the log level from the remote configuration URL at the specified interval
 // and updates the underlying log level if it has changed.
 func (r *remoteLogger) UpdateLogLevel() {
-	interval := time.Duration(r.levelFetchInterval) * time.Second
+	interval := r.levelFetchInterval * time.Second
```
ccoVeille, line 54:
```suggestion
	interval := r.levelFetchInterval
```

pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
```diff
@@ -114,10 +114,10 @@ func TestDynamicLoggerSuccess(t *testing.T) {
 
 	log := testutil.StdoutOutputForFunc(func() {
 		// Create a new remote logger with the mock server URL
-		rl := New(logging.INFO, mockServer.URL, "1")
+		rl := New(logging.INFO, mockServer.URL, "1") // 1 second is the minimum value
```
ccoVeille, line 117: The signature of New should be refactored to pass a time.Duration

pkg/gofr/handler.go
Umang01-hash: We are asking the user to give the value only not units:

```go
reqTimeout, err := strconv.Atoi(a.Config.Get("REQUEST_TIMEOUT"))
	if err != nil && a.Config.Get("REQUEST_TIMEOUT") != "" {
		a.container.Error("invalid value of config REQUEST_TIMEOUT. setting default value to 5 seconds.")
	}

	a.httpServer.router.Add(method, pattern, handler{
		function:       h,
		container:      a.container,
		requestTimeout: time.Duration(reqTimeout),
	})
```


If i remove time.Second in `h.requestTimeout*time.Second` and use time.ParseDuration in above code user will need to give units in config

pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
Umang01-hash: This is how we populate the levelFetchInterval in `New` :

```go
if c.Logger == nil {
		c.Logger = remotelogger.New(logging.GetLevelFromString(conf.Get("LOG_LEVEL")), conf.Get("REMOTE_LOG_URL"),
			conf.GetOrDefault("REMOTE_LOG_FETCH_INTERVAL", "15"))
	}
```

if `New` will accpet time.Duartion then user will need to define the unit

pkg/gofr/handler.go
ccoVeille: reqTimeout is an interface

We can expect user to be OK to have seconds via the config. And it's what they had.

Our issue is to speed up the tests

So your code should be using

reqTimeOut * time.Second 

When passing the duration to the config

Umang01-hash: Sorry my bad got it

pkg/gofr/logging/remotelogger/dynamic_level_logger.go
Umang01-hash: yups reduced the duration

pkg/gofr/handler_test.go
Umang01-hash: yups reduced the duration

pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
Umang01-hash: changed the signature to accept `time.Duration`

pkg/gofr/logging/remotelogger/dynamic_level_logger.go
Umang01-hash: sure did the change

pkg/gofr/logging/remotelogger/dynamic_level_logger.go
Umang01-hash: sure did this change

examples/using-cron-jobs/main_test.go
ccoVeille: I opened

- https://github.com/gofr-dev/gofr/issues/916

pkg/gofr/container/container.go
```diff
@@ -73,8 +73,13 @@ func (c *Container) Create(conf config.Config) {
...
 
 	if c.Logger == nil {
+		levelFetchConfig, err := strconv.Atoi(conf.GetOrDefault("REMOTE_LOG_FETCH_INTERVAL", "15"))
+		if err != nil {
+			levelFetchConfig = 15
+		}
+
 		c.Logger = remotelogger.New(logging.GetLevelFromString(conf.Get("LOG_LEVEL")), conf.Get("REMOTE_LOG_URL"),
-			conf.GetOrDefault("REMOTE_LOG_FETCH_INTERVAL", "15"))
+			time.Duration(levelFetchConfig)*time.Second)
```
ccoVeille, line 82:
```suggestion
			levelFetchConfig*time.Second)
```

pkg/gofr/gofr.go
```diff
@@ -339,10 +339,15 @@ func (a *App) PATCH(pattern string, handler Handler) {
...
+	reqTimeout, err := strconv.Atoi(a.Config.Get("REQUEST_TIMEOUT"))
+	if err != nil && a.Config.Get("REQUEST_TIMEOUT") != "" {
+		a.container.Error("invalid value of config REQUEST_TIMEOUT. setting default value to 5 seconds.")
+	}
+
 	a.httpServer.router.Add(method, pattern, handler{
 		function:       h,
 		container:      a.container,
-		requestTimeout: a.Config.Get("REQUEST_TIMEOUT"),
+		requestTimeout: time.Duration(reqTimeout) * time.Second,
```
ccoVeille, line 350:
```suggestion
		requestTimeout: reqTimeout * time.Second,
```

pkg/gofr/handler_test.go
```diff
@@ -65,11 +65,11 @@ func TestHandler_ServeHTTP_Timeout(t *testing.T) {
 	w := httptest.NewRecorder()
 	r := httptest.NewRequest(http.MethodGet, "/", http.NoBody)
 
-	h := handler{requestTimeout: "1"}
+	h := handler{requestTimeout: 100 * time.Millisecond} // the minimum value is 1s
```
ccoVeille, line 68:
```suggestion
	h := handler{requestTimeout: 100 * time.Millisecond}
```

pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
```diff
@@ -32,7 +32,7 @@ func TestRemoteLogger_UpdateLevel(t *testing.T) {
 
 	go rl.UpdateLogLevel()
 
-	time.Sleep(2 * time.Second)
+	time.Sleep(1200 * time.Millisecond)
```
ccoVeille, line 35: I think this one can be reduced with the changes you made

pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
```diff
@@ -114,10 +114,10 @@ func TestDynamicLoggerSuccess(t *testing.T) {
 
 	log := testutil.StdoutOutputForFunc(func() {
 		// Create a new remote logger with the mock server URL
-		rl := New(logging.INFO, mockServer.URL, "1")
+		rl := New(logging.INFO, mockServer.URL, 100*time.Millisecond) // 1 second is the minimum value
```
ccoVeille, line 117:
```suggestion
		rl := New(logging.INFO, mockServer.URL, 100*time.Millisecond)
```

examples/using-subscriber/main_test.go
ccoVeille: Should we create a separate issue?

pkg/gofr/container/container.go
Umang01-hash: @ccoVeille  I didn't understand why are you eliminating these lines for fetching config values:
`conf.GetOrDefault("REMOTE_LOG_FETCH_INTERVAL", "15")`

Where and how will we fetch the config values then?

pkg/gofr/gofr.go
Umang01-hash: again how will we populate `reqTimeout ` without `a.Config.Get("REQUEST_TIMEOUT")` ?

pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
Umang01-hash: Sure reduced

pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
Umang01-hash: sure removed the comment

pkg/gofr/gofr.go
ccoVeille: I had simply removed the type conversion, not the value

pkg/gofr/container/container.go
ccoVeille: I had only removed the type conversion, not the value

But I was wrong, the only way to multiply a duration is to case the type ad you did.



So your code is fine

Umang01-hash: thankyou

pkg/gofr/gofr.go
```diff
@@ -339,10 +339,15 @@ func (a *App) PATCH(pattern string, handler Handler) {
 func (a *App) add(method, pattern string, h Handler) {
 	a.httpRegistered = true
 
+	reqTimeout, err := strconv.Atoi(a.Config.Get("REQUEST_TIMEOUT"))
+	if err != nil && a.Config.Get("REQUEST_TIMEOUT") != "" {
+		a.container.Error("invalid value of config REQUEST_TIMEOUT. setting default value to 5 seconds.")
```
ccoVeille, line 344: You aren't setting it to 5 here, am I wrong?

I see you copy pasted existing code, but I'm unsure it worked

pkg/gofr/http/middleware/oauth.go
ccoVeille: As you refactored and used a few milliseconds in the test, this refactoring is no longer needed maybe

pkg/gofr/gofr.go
```diff
@@ -339,10 +339,15 @@ func (a *App) PATCH(pattern string, handler Handler) {
 func (a *App) add(method, pattern string, h Handler) {
 	a.httpRegistered = true
 
+	reqTimeout, err := strconv.Atoi(a.Config.Get("REQUEST_TIMEOUT"))
```
ccoVeille, line 342: Why not using GetOrDefault?

pkg/gofr/container/container.go
```diff
@@ -75,8 +75,13 @@ func (c *Container) Create(conf config.Config) {
 	}
 
 	if c.Logger == nil {
+		levelFetchConfig, err := strconv.Atoi(conf.GetOrDefault("REMOTE_LOG_FETCH_INTERVAL", "15"))
+		if err != nil {
+			levelFetchConfig = 15
```
ccoVeille, line 80: No error log message here to report the value cannot be used

Umang01-hash: Gave an error message for incorrect value

pkg/gofr/gofr.go
Umang01-hash: Not using `GetOrDefault` because in that case even if the doesn't want to set the request timeout we will force set it.

The idea is if user has set the config `REQUEST_TIMEOUT` then only we will set it , and if the user has given incorrect value for the config we will set it default to 5 seconds.

pkg/gofr/gofr.go
Umang01-hash: I corrected the code here and now we are checking the value here but not logging here because `add` is being called multiple times to set the default routes and if we log here we will have multiple logs :

Screenshot 2024-08-08 at 11 37 50 AM

pkg/gofr/gofr.go
ccoVeille: Thanks it's clearer. Maybe the code could mention it in a comment

ccoVeille:
I'm unsure about the changes in last commit. The log somewhere, the default value somewhere else, and a defer that check on err.

I'm busy today, someone else will need to review it

Umang01-hash:
> I'm unsure about the changes in last commit. The log somewhere, the default value somewhere else, and a defer that check on err.
> 
> I'm busy today, someone else will need to review it

This is because in this below code:

```go
if c.Logger == nil {
		levelFetchConfig, err := strconv.Atoi(conf.GetOrDefault("REMOTE_LOG_FETCH_INTERVAL", "15"))
		if err != nil {
			levelFetchConfig = 15
		}

		c.Logger = remotelogger.New(logging.GetLevelFromString(conf.Get("LOG_LEVEL")), conf.Get("REMOTE_LOG_URL"),
			time.Duration(levelFetchConfig)*time.Second)

		defer func() {
			if err != nil {
				c.Logger.Error("Incorrect Value for REMOTE_LOG_FETCH_INTERVAL. Setting Default to 15 sec.")
			}
		}()
	}
```

when we are checking for the value to be a valid one, the logger is still not initiated and if we log there we will get a panic.
After we create a logger using **remotelogger.New** then only we can use it to log the error message so using defer inside a func to check on error.

srijan-27 (changes requested):

pkg/gofr/gofr.go
```diff
@@ -40,6 +40,7 @@ const (
 	shutDownTimeout        = 30 * time.Second
 	gofrTraceExporter      = "gofr"
 	gofrTracerURL          = "https://tracer.gofr.dev"
+	defaultRequestTimeout  = 5
```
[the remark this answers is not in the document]
ccoVeille: I agree

pkg/gofr/gofr.go
```diff
@@ -243,6 +244,11 @@ func (a *App) httpServerSetup() {
 		a.add(http.MethodGet, "/.well-known/{name}", SwaggerUIHandler)
 	}
 
+	_, err := strconv.Atoi(a.Config.Get("REQUEST_TIMEOUT"))
+	if err != nil && a.Config.Get("REQUEST_TIMEOUT") != "" {
+		a.container.Error("invalid value of config REQUEST_TIMEOUT. setting default value to 5 seconds.")
```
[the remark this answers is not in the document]
ccoVeille: Unless the fallback value is removed, see

https://github.com/gofr-dev/gofr/pull/831#discussion_r1714758722

pkg/gofr/container/container.go
```diff
@@ -75,8 +75,19 @@ func (c *Container) Create(conf config.Config) {
...
+		}
+
 		c.Logger = remotelogger.New(logging.GetLevelFromString(conf.Get("LOG_LEVEL")), conf.Get("REMOTE_LOG_URL"),
-			conf.GetOrDefault("REMOTE_LOG_FETCH_INTERVAL", "15"))
+			time.Duration(levelFetchConfig)*time.Second)
+
+		defer func() {
+			if err != nil {
+				c.Logger.Error("Incorrect Value for REMOTE_LOG_FETCH_INTERVAL. Setting Default to 15 sec.")
```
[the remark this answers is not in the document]
Umang01-hash: done

pkg/gofr/gofr.go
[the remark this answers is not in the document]
Umang01-hash: removed the fall back value and updated the log message

pkg/gofr/gofr.go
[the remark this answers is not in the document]
Umang01-hash: chose this option: simply log an error, and have no timeout

pkg/gofr/gofr.go
```diff
@@ -243,6 +244,11 @@ func (a *App) httpServerSetup() {
 		a.add(http.MethodGet, "/.well-known/{name}", SwaggerUIHandler)
 	}
 
+	_, err := strconv.Atoi(a.Config.Get("REQUEST_TIMEOUT"))
+	if err != nil && a.Config.Get("REQUEST_TIMEOUT") != "" {
```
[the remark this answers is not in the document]
Umang01-hash: refactored

pkg/gofr/handler.go
```diff
@@ -118,16 +115,6 @@ func catchAllHandler(*Context) (interface{}, error) {
 	return nil, gofrHTTP.ErrorInvalidRoute{}
 }
 
-// Helper function to parse and validate request timeout.
-func (h handler) setContextTimeout(timeout string) int {
-	reqTimeout, err := strconv.Atoi(timeout)
-	if err != nil || reqTimeout < 0 {
```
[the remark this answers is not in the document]
Umang01-hash: i have added that check at both places.

pkg/gofr/container/container.go
```diff
@@ -75,8 +75,19 @@ func (c *Container) Create(conf config.Config) {
...
+
 		c.Logger = remotelogger.New(logging.GetLevelFromString(conf.Get("LOG_LEVEL")), conf.Get("REMOTE_LOG_URL"),
-			conf.GetOrDefault("REMOTE_LOG_FETCH_INTERVAL", "15"))
+			time.Duration(levelFetchConfig)*time.Second)
+
+		defer func() {
+			if err != nil {
+				c.Logger.Error("incorrect value for REMOTE_LOG_FETCH_INTERVAL. setting default of 15 sec.")
+			}
+		}()
```
ccoVeille, line 90: I think I already asked, but here, the logger is available, you just created it.

Can't we just do this ?

```suggestion
			if err != nil {
				c.Logger.Error("incorrect value for REMOTE_LOG_FETCH_INTERVAL. setting default of 15 sec.")
			}
```

I'm afraid `err` could be reassigned to something else

Or the code could be something like

```suggestion
		if err != nil {
			defer func() {
				c.Logger.Error("incorrect value for REMOTE_LOG_FETCH_INTERVAL. setting default of 15 sec.")
			}()
		}
```

But I'm unsure about the scope of the defer here

pkg/gofr/gofr.go
```diff
@@ -235,6 +235,13 @@ func (a *App) httpServerSetup() {
 		a.add(http.MethodGet, "/.well-known/{name}", SwaggerUIHandler)
 	}
 
+	requestTimeout := a.Config.Get("REQUEST_TIMEOUT")
+	timeoutVal, err := strconv.Atoi(requestTimeout)
+
+	if (err != nil && requestTimeout != "") || timeoutVal < 0 {
+		a.container.Error("invalid value of config REQUEST_TIMEOUT.")
+	}
```
ccoVeille, line 243: I would like to suggest this.

```suggestion
	requestTimeout := a.Config.Get("REQUEST_TIMEOUT")
	if requestTimeout != "" {
		timeoutVal, err := strconv.Atoi(requestTimeout)
 		if err != nil || timeoutVal < 0 {
			a.container.Error("invalid value of config REQUEST_TIMEOUT.")
		}
	}
```

I think it's clearer

pkg/gofr/gofr.go
```diff
@@ -331,10 +338,15 @@ func (a *App) PATCH(pattern string, handler Handler) {
 func (a *App) add(method, pattern string, h Handler) {
 	a.httpRegistered = true
 
+	reqTimeout, err := strconv.Atoi(a.Config.Get("REQUEST_TIMEOUT"))
+	if (err != nil && a.Config.Get("REQUEST_TIMEOUT") != "") || reqTimeout < 0 {
+		reqTimeout = 0
+	}
```
ccoVeille, line 344: I agree with @srijan-27 

The code is duplicated

Can't we move the log about the error here?

Umang01-hash: Moving the logger here will have this impact as told earlier:
https://github.com/gofr-dev/gofr/pull/831#discussion_r1708752604

pkg/gofr/gofr.go
Umang01-hash: Sure will refactor it

pkg/gofr/container/container.go
Umang01-hash: I have tested this we can do it like :

```go
if c.Logger == nil {
		levelFetchConfig, err := strconv.Atoi(conf.GetOrDefault("REMOTE_LOG_FETCH_INTERVAL", "15"))
		if err != nil {
			levelFetchConfig = 15
		}

		c.Logger = remotelogger.New(logging.GetLevelFromString(conf.Get("LOG_LEVEL")), conf.Get("REMOTE_LOG_URL"),
			time.Duration(levelFetchConfig)*time.Second)

		if err != nil {
			c.Logger.Error("incorrect value for REMOTE_LOG_FETCH_INTERVAL. setting default of 15 sec.")
		}
	}
```

ccoVeille: Yes, it's what I was thinking of

pkg/gofr/gofr.go
srijan-27: yes, `add` is called for all routes so the log can't be moved here.. same is the case for config (why do we need to validate config for every route) as it is defined throughout the application.

ccoVeille: indeed. Thanks for reminding me.

Can't add a requestTimeout into the App struct, define it when reading the conf, and simply use it here?

Reading the env var, validating it, on each call to `add` seems a problem to me, even more when the code is duplicated, so it means code may divert, and could be a pain to maintain

pkg/gofr/container/container.go
```diff
@@ -75,8 +75,17 @@ func (c *Container) Create(conf config.Config) {
...
+		if err != nil {
+			levelFetchConfig = 15
+		}
+
 		c.Logger = remotelogger.New(logging.GetLevelFromString(conf.Get("LOG_LEVEL")), conf.Get("REMOTE_LOG_URL"),
-			conf.GetOrDefault("REMOTE_LOG_FETCH_INTERVAL", "15"))
+			time.Duration(levelFetchConfig)*time.Second)
+
+		if err != nil {
+			c.Logger.Error("incorrect value for REMOTE_LOG_FETCH_INTERVAL. setting default of 15 sec.")
```
srijan-27, line 87: we are using `incorrect` here and `invalid` in the request_timeout log.. pls use the same for consistency in logs.

pkg/gofr/gofr.go
Umang01-hash: IMO, I don't think so putting `request-timeout` field in app struct is a better option. The GoFr app struct hold the servers, conatiner, subscriptionManager etc. , putting an config field there may pollute the struct

ccoVeille: OK, I can agree that. 

One last suggestion: what about having a method to get the value from the env variable then?

Umang01-hash: We already have method to get value from Config:

```go

func (*EnvLoader) Get(key string) string {
	return os.Getenv(key)
}

func (*EnvLoader) GetOrDefault(key, defaultValue string) string {
	if val := os.Getenv(key); val != "" {
		return val
	}

	return defaultValue
}
```

Are you suggesting we should have a method on app to Get env variables ?

ccoVeille: Yes, I'm thinking about something like this

```go
func (*App) getRequestTimeout (vale int, valid bool) {
	
      requestTimeout := a.Config.Get("REQUEST_TIMEOUT") 
      if requestTimeout == "" {
            return 0, true
      }
      timeoutVal, err := strconv.Atoi(requestTimeout) 
      if err != nil || timeoutVal < 0 { 
             return 0, false // can be checked to log an error
      }

      return timeoutVal, true
}
```

Umang01-hash: @srijan-27 @vipul-rawat what do you guys think? Should we make a method on app for this use-case?

srijan-27: I guess creating a method also will not rectify our main concern.. (reading config multiple times) as we would anyways be calling this method in `add`.
Not a good thing to do, but should we create a local variable and update the value for this while reading the config in `httpServerSetup`, and using this variable in `add`?

ccoVeille: That was my suggestion 😅😁

Umang01-hash: if we will create a variable in `gofr.go` file then it will be a global variable. even if you keep it unexported but still it will be global on scope. And we should not create global variables untill very necessary. 
@srijan-27 do you agree?

srijan-27: couldn't think of any viable solution.. lets add a TODO for this.. and take it later, let's close this PR first since it is pending for long time.

pkg/gofr/container/container.go
Umang01-hash: sure corrected

srijan-27 (approved):

ccoVeille:
Thanks 🎉 @Umang01-hash for working on this.

I started and you had to cope with all the non-fun changes.

Thanks.

I'm curious, do we see a change in the CI duration?

Issue #499, opened 2024-04-18
badgerDB support for key value


vikash:

vishaljangid1729:
Hey @vikash 

I've got quite a bit of experience with badgerDB, having used it extensively in various projects.

Could you provide more details about the specific requirements or use cases for this feature? I'd love to dive deeper into it and see how I can contribute to its implementation.

Looking forward to hearing from you!

aryan-mehrotra-zs:
Hi @vishaljangid1729 we would want it to be implemented in the similar way Mongo is being implemented.
we have defined it interface and methods in gofr : https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/datasource/mongo.go
and its implementation exists in a different repo : https://github.com/vipul-rawat/gofr-mongo

aryanmehrotra:
Hi @vishaljangid1729 we would want it to be implemented in the similar way Mongo is being implemented.
we have defined it interface and methods in gofr : https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/datasource/mongo.go
and its implementation exists in a different repo : https://github.com/vipul-rawat/gofr-mongo

Moreover, you can join discord server where we can connect as well to discuss more on this : https://discord.gg/Kb3cnx9r

PR #835, opened 2024-07-11
Support for BadgerDB as KV Store

aryanmehrotra:
- Added support for badgerDB as key value store.
- Closes: #499

**Checklist:**


pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,103 @@
...
+	logger  Logger
+	metrics Metrics
+}
+
+func New(configs Configs) *client {
+	return &client{configs: configs}
+}
+
+// UseLogger sets the logger for the MongoDB client which asserts the Logger interface.
+func (c *client) UseLogger(logger interface{}) {
```
ccoVeille, line 24: We are far from Go 1.18, you can use any


```suggestion
func (c *client) UseLogger(logger any) {
```

ccoVeille, line 24: Why not using this?


```suggestion
func (c *client) UseLogger(logger Logger) {
```

Is it for mock consideration?

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,103 @@
...
+	if l, ok := logger.(Logger); ok {
+		c.logger = l
+	}
+}
+
+// UseMetrics sets the metrics for the MongoDB client which asserts the Metrics interface.
+func (c *client) UseMetrics(metrics interface{}) {
```
ccoVeille, line 31: Same question about using


```suggestion
func (c *client) UseMetrics(metrics Metrics) {
```

pkg/gofr/container/datasources.go
```diff
@@ -199,3 +199,15 @@ type HealthChecker interface {
...
+type KVStore interface {
+	Get(ctx context.Context, key string) (string, error)
+	Set(ctx context.Context, key string, value string) error
+	Delete(ctx context.Context, key string) error
+}
+
+type KVStoreProvider interface {
+	KVStore
+
+	provider
```
ccoVeille, line 212: Maybe you could add an HealthChecker, by using any random badger function

maybe

https://pkg.go.dev/github.com/dgraph-io/badger/v4#DB.IsClosed

I might be commenting at the wrong place as this is the KVStore interface

pkg/gofr/datasource/kv-store/badger/badger_test.go
```diff
@@ -0,0 +1,65 @@
...
+	cl.UseLogger(NewMockLogger(DEBUG, bytes.NewBuffer(logs)))
+	cl.UseMetrics(NewMockMetrics(ctrl))
+	cl.Connect()
+
+	return cl
+}
+
+func Test_ClientSet(t *testing.T) {
+	cl := setupDB(t)
+	defer os.RemoveAll("test_badger")
```
ccoVeille, line 29: Move this to a t.Cleanup(func () {}) inside setupDB this way you won't need this defer in each tests

https://pkg.go.dev/testing#T.Cleanup

It's the purpose and the magic that t.Cleanup brought

pkg/gofr/datasource/kv-store/badger/badger_test.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func Test_ClientGetError(t *testing.T) {
+	cl := setupDB(t)
+	defer os.RemoveAll("test_badger")
+
+	val, err := cl.Get(context.Background(), "lkey")
+
+	assert.EqualError(t, err, "Key not found")
+	assert.Equal(t, "", val)
```
ccoVeille, line 55: You could use 


```suggestion
	assert.Emptt(t, val)
```

pkg/gofr/datasource/kv-store/badger/logger.go
```diff
@@ -0,0 +1,42 @@
...
+		clean(ql.Query), "CASS", ql.Duration, clean(ql.Keyspace))
+}
+
+// clean takes a string query as input and performs two operations to clean it up:
+// 1. It replaces multiple consecutive whitespace characters with a single space.
+// 2. It trims leading and trailing whitespace from the string.
+// The cleaned-up query string is then returned.
+func clean(query string) string {
+	// Replace multiple consecutive whitespace characters with a single space
+	query = regexp.MustCompile(`\s+`).ReplaceAllString(query, " ")
```
ccoVeille, line 36: Regexp should be moved outside the function in a var, this way it's computed only once and not each time the function is called

pkg/gofr/datasource/kv-store/badger/logger.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"fmt"
+	"io"
+	"regexp"
+	"strings"
+)
+
+type Logger interface {
+	Debug(args ...interface{})
```
ccoVeille, line 11:
```suggestion
	Debug(args ...any)
```

pkg/gofr/datasource/kv-store/badger/mock_logger.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (m *MockLogger) logf(level Level, format string, args ...interface{}) {
+	out := m.out
+	if level == ERROR {
+		out = m.errOut
+	}
+
+	message := fmt.Sprintf(format, args...)
+
+	fmt.Fprintf(out, "%v\n", message)
```
ccoVeille, line 64: What about this?

```suggestion
	fmt.Fprintf(out, format+"\n", args...)
```
Of simply adding the "\n" to format before calling Fprintf

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,103 @@
...
+	err = txn.Commit()
+	if err != nil {
+		return "", err
+	}
+
+	return string(value), nil
+}
+
+func (c *client) Set(_ context.Context, key string, value string) error {
+	txn := c.db.NewTransaction(true)
```
ccoVeille, line 72: When I see a pattern repeated over and over, I would refacfor into something like this

```go
func (c *Client) useTransaction(f func(txn) error) error {
          txn := c.db.NewTransaction(true) 
         defer txn.Discard() 
  
         err := f(txn)
         if err != nil { 
                 return err 
         } 
  
         err = txn.Commit() 
         if err != nil { 
                 return err 
         } 
  
         return nil
}
```

ccoVeille: It might need to be adapted for the getter

ccoVeille (commented):
Hello, I looked at your PR.

Please bare with me and my comments

pkg/gofr/datasource/kv-store/badger/logger.go
aryanmehrotra: I am refraining from defining a variable globally, need to think of something else

pkg/gofr/datasource/kv-store/badger/badger.go
aryanmehrotra: provider interface has interface, that is why we are accepting interfaces.
Else we would need to have a common source for the Logger and Metrics interface which would bring dependency, which is we are refraining from

pkg/gofr/datasource/kv-store/badger/badger.go
aryanmehrotra: provider interface has interface, that is why we are accepting interfaces.
Else we would need to have a common source for the Logger and Metrics interface which would bring dependency, which is we are refraining from

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,125 @@
...
+
+// UseMetrics sets the metrics for the MongoDB client which asserts the Metrics interface.
+func (c *client) UseMetrics(metrics any) {
+	if m, ok := metrics.(Metrics); ok {
+		c.metrics = m
+	}
+}
+
+// Connect establishes a connection to MongoDB and registers metrics using the provided configuration when the client was Created.
+func (c *client) Connect() {
```
ccoVeille, line 39: This one should return an issue error instead of logging it

```suggestion
func (c *client) Connect() error {
```

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,125 @@
...
+	db, err := badger.Open(badger.DefaultOptions(c.configs.DirPath))
+	if err != nil {
+		c.logger.Errorf("failed to initialise badgerDB : %v", err)
```
ccoVeille, line 42:
```suggestion
		return fmt.Errorf("failed to initialize badgerDB : %w", err)
```

ccoVeille, line 42:
```suggestion
		c.logger.Errorf("failed to initialize badgerDB : %v", err)
```

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	c.db = db
+}
+
+func (c *client) Get(_ context.Context, key string) (string, error) {
+	var value []byte
+
+	txn := c.db.NewTransaction(false)
```
ccoVeille, line 51: Does a getter really require a transaction?

I may understand to use the transaction (storing it in the context could help here), but not a new one.

And not commit one

pkg/gofr/datasource/kv-store/badger/badger_test.go
```diff
@@ -0,0 +1,77 @@
...
+	var logs []byte
+
+	ctrl := gomock.NewController(t)
+
+	cl.UseLogger(NewMockLogger(DEBUG, bytes.NewBuffer(logs)))
+	cl.UseMetrics(NewMockMetrics(ctrl))
+	cl.Connect()
+
+	t.Cleanup(func() {
+		os.RemoveAll("test_badger")
```
ccoVeille, line 25: This one could be removed if you do that

https://github.com/gofr-dev/gofr/pull/835#discussion_r1675505360

pkg/gofr/datasource/kv-store/badger/badger.go
ccoVeille: OK. So maybe here you could define an interface with one method that will be implemented by each Logger (and do the same for Metrics)

There is usually one thing that can be shared.

This way the code is clearer. IDE are able to autocomplete.

But I concede, if it won't change the type assertion you are doing there. But it would be better I think.

pkg/gofr/datasource/kv-store/badger/logger.go
ccoVeille: OK

pkg/gofr/datasource/kv-store/badger/badger.go
aryanmehrotra: It can be part of a bigger refactor, as all external DBs need to be refactored

ccoVeille: Of course, it has to be a separate PR. You would have to rewrite a lot of files. It's out of the scope of this PR

ccoVeille: I opened 

- https://github.com/gofr-dev/gofr/issues/840

pkg/gofr/datasource/kv-store/badger/badger.go
aryanmehrotra: Its required as per the badgerDB documentation, to keep the data consistent.

pkg/gofr/datasource/kv-store/badger/badger_test.go
```diff
@@ -0,0 +1,77 @@
...
+	t.Helper()
+	cl := New(Configs{DirPath: "test_badger"})
+
+	var logs []byte
+
+	ctrl := gomock.NewController(t)
+
+	cl.UseLogger(NewMockLogger(DEBUG, bytes.NewBuffer(logs)))
+	cl.UseMetrics(NewMockMetrics(ctrl))
+	cl.Connect()
```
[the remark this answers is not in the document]
aryanmehrotra: This needs to be done later at every place if we decide to do it

pkg/gofr/datasource/kv-store/badger/badger.go
ccoVeille: Maybe it requires a comment to explain it

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,126 @@
...
+		c.logger.Errorf("failed to initialize badgerDB : %v", err)
+	}
+
+	c.db = db
+}
+
+func (c *client) Get(_ context.Context, key string) (string, error) {
+	var value []byte
+
+	// It is being set to false as we dont want to make any changes to data.
```
ccoVeille, line 51:
```suggestion
	//  transaction is set to false as we don't want to make any changes to data.
```

pkg/gofr/datasource/kv-store/badger/logger.go
vipul-rawat: These regexp.MustCompile() should be outside the function, this is a pattern in most external datasources and needs to be fixed in a separate PR. I will be creating an issue

vipul-rawat (approved):

Umang01-hash (dismissed):

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,125 @@
...
+	configs Configs
+	logger  Logger
+	metrics Metrics
+}
+
+func New(configs Configs) *client {
+	return &client{configs: configs}
+}
+
+// UseLogger sets the logger for the MongoDB client which asserts the Logger interface.
```
ccoVeille, line 25: I think there is a copy pasta 

```suggestion
// UseLogger sets the logger for the badgerDB client which asserts the Logger interface.
```

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,125 @@
...
+func (c *client) UseLogger(logger any) {
+	if l, ok := logger.(Logger); ok {
+		c.logger = l
+	}
+}
+
+// UseMetrics sets the metrics for the MongoDB client which asserts the Metrics interface.
```
ccoVeille, line 32:
```suggestion
// UseMetrics sets the metrics for the badgerDB client which asserts the Metrics interface.
```

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,125 @@
...
+func (c *client) UseMetrics(metrics any) {
+	if m, ok := metrics.(Metrics); ok {
+		c.metrics = m
+	}
+}
+
+// Connect establishes a connection to MongoDB and registers metrics using the provided configuration when the client was Created.
```
ccoVeille, line 39:
```suggestion
// Connect establishes a connection to badgerDB and registers metrics using the provided configuration when the client was Created.
```

ccoVeille (approved):

docs/advanced-guide/key-value-store/page.md
```diff
@@ -0,0 +1,80 @@
...
+    Set(ctx context.Context, key, value string) error
+    Delete(ctx context.Context, key string) error
+}
+```
+
+## BadgerDB
+GoFr supports injecting BadgerDB that supports the following interface. Any driver that implements the interface can be added
+using `app.AddKVStore()` method, and user's can use BadgerDB across application with `gofr.Context`.
+
+User's can easily inject a driver that supports this interface, this provides usability without
```
ccoVeille, line 23:
```suggestion
Users can easily inject a driver that supports this interface, this provides usability without
```

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,159 @@
...
+	configs Configs
+	logger  Logger
+	metrics Metrics
+}
+
+func New(configs Configs) *client {
+	return &client{configs: configs}
+}
+
+// UseLogger sets the logger for the badgerDB client which asserts the Logger interface.
```
ccoVeille, line 26:
```suggestion
// UseLogger sets the logger for the BadgerDB client which asserts the Logger interface.
```

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,159 @@
...
+// UseMetrics sets the metrics for the badgerDB client which asserts the Metrics interface.
+func (c *client) UseMetrics(metrics any) {
+	if m, ok := metrics.(Metrics); ok {
+		c.metrics = m
+	}
+}
+
+// Connect establishes a connection to badgerDB and registers metrics using the provided configuration when the client was Created.
+func (c *client) Connect() {
+	c.logger.Infof("connecting to badgerDB at %v", c.configs.DirPath)
```
ccoVeille, line 42:
```suggestion
	c.logger.Infof("connecting to BadgerDB at %v", c.configs.DirPath)
```

pkg/gofr/datasource/kv-store/badger/badger.go
```diff
@@ -0,0 +1,159 @@
...
+
+	badgerBuckets := []float64{.05, .075, .1, .125, .15, .2, .3, .5, .75, 1, 2, 3, 4, 5, 7.5, 10}
+	c.metrics.NewHistogram("app_badger_stats", "Response time of Badger queries in milliseconds.", badgerBuckets...)
+
+	db, err := badger.Open(badger.DefaultOptions(c.configs.DirPath))
+	if err != nil {
+		c.logger.Errorf("error while connecting to badgerDB: %v", err)
```
ccoVeille, line 49:
```suggestion
		c.logger.Errorf("error while connecting to BadgerDB: %v", err)
```

docs/advanced-guide/key-value-store/page.md
```diff
@@ -0,0 +1,80 @@
+# Key Value Store
```
ccoVeille, line 1: I'm thinking about it

Shouldn't Redis storage be considered as a KVStore

docs/advanced-guide/key-value-store/page.md
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"fmt"
+
+	"gofr.dev/pkg/gofr"
+	"gofr.dev/pkg/gofr/datasource/kv-store/badger"
+)
+
+type User struct {
+	Id   string
```
ccoVeille, line 37:
```suggestion
	ID   string
```

pkg/gofr/datasource/clickhouse/logger.go
```diff
@@ -23,7 +23,7 @@ type Log struct {
 
 func (l *Log) PrettyPrint(writer io.Writer) {
 	fmt.Fprintf(writer, "\u001B[38;5;8m%-32s \u001B[38;5;24m%-6s\u001B[0m %8d\u001B[38;5;8mµs\u001B[0m %s\n",
-		l.Type, "CHDB", l.Duration, clean(l.Query))
+		l.Type, "BADGR", l.Duration, clean(l.Query))
```
ccoVeille, line 26: So the clickhouse logger report badger?

pkg/gofr/datasource/kv-store/badger/logger.go
```diff
@@ -0,0 +1,28 @@
...
+type Log struct {
+	Type     string `json:"type"`
+	Duration int64  `json:"duration"`
+	Key      string `json:"key"`
+	Value    string `json:"value,omitempty"`
+}
+
+func (l *Log) PrettyPrint(writer io.Writer) {
+	fmt.Fprintf(writer, "\u001B[38;5;8m%-32s \u001B[38;5;162m%-6s\u001B[0m %8d\u001B[38;5;8mµs\u001B[0m %s \n",
+		l.Type, "BADG", l.Duration, strings.Join([]string{l.Key, l.Value}, " "))
```
ccoVeille, line 27: Badger or badgr ?

I think you updated clickhouse by error

docs/advanced-guide/key-value-store/page.md
aryanmehrotra: It would also support it, it would be configurable, user can use any kvstore of their choice

ccoVeille: My point is more the fact, gofr already supports Redis


https://pkg.go.dev/gofr.dev/pkg/gofr/datasource/redis

So my point was about the fact Redis code could be refactored to rely on KVStore interface maybe

Plus the fact Redis, could be documented here in this Markdown section.

Because right now it says somehow "badger only for now, other maybe later", while Redis is already there.

Please note, I might have misunderstood something.

srijan-27 (approved):

docs/advanced-guide/key-value-store/page.md
ccoVeille: Do @vipul-rawat @srijan-27 @aryanmehrotra

What are your thoughts?

Issue #663, opened 2024-05-28
Support For Graceful Shutdown


aryanmehrotra:
Application should wait before all the goroutines are completed before shutting down application

ccoVeille:
I second that, but I would say the app should wait a short delay and not forever. Something like 5s-1m depending on the app. So it could be an app parameter

vikash:
Agreed. 5 second is a respectable default - with an option to override it, if user so wishes.

PR #837, opened 2024-07-12
En/graceful shutdown

Umang01-hash:
## Pull Request Template

**Description:**

-   Closes issue #663 
- Taking reference from PR #781 

**Checklist:**


pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -338,19 +341,25 @@ func (m *MQTT) Unsubscribe(topic string) error {
...
 
 func (m *MQTT) Disconnect(waitTime uint) {
 	m.mu.RLock()
+	defer m.mu.RUnlock()
+
 	for topic := range m.subscriptions {
 		_ = m.Unsubscribe(topic)
 	}
 
 	m.Client.Disconnect(waitTime)
```
ccoVeille, line 362: Doesn't it return an error?

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -366,10 +375,11 @@ func (m *MQTT) Ping() error {
 func createReconnectHandler(mu *sync.RWMutex, config *Config, subs map[string]subscription) func(c mqtt.Client) {
 	return func(c mqtt.Client) {
 		mu.RLock()
+		defer mu.RUnlock()
+
 		for k, v := range subs {
 			c.Subscribe(k, config.QoS, v.handler)
```
ccoVeille, line 381: No error returned?

pkg/gofr/metrics_server.go
```diff
@@ -11,24 +12,34 @@ import (
 
 type metricServer struct {
 	port int
+	srvr *http.Server
```
ccoVeille, line 15: `srv` is used in other file, not srvr

```suggestion
	srv *http.Server
```

pkg/gofr/subscriber.go
```diff
@@ -24,39 +24,63 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
-		// commit the message if the subscription function does not return error
-		if err == nil {
-			msg.Commit()
-		} else {
-			s.container.Logger.Errorf("error in handler for topic %s: %v", topic, err)
-		}
+	msg, err := s.container.GetSubscriber().Subscribe(ctx, topic)
+
+	if errors.Is(err, kafka.ErrConsumerGroupNotProvided) {
+		s.container.Logger.Error("cannot subscribe as consumer_id is not provided in configs")
```
ccoVeille, line 50: I would expect the caller to log, so to remove all error logs

pkg/gofr/subscriber.go
```diff
@@ -24,39 +24,63 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
+		return err
+	}
+
+	if err != nil {
+		s.container.Logger.Errorf("error while reading from topic %v, err: %v", topic, err.Error())
+		return nil
```
ccoVeille, line 56: No error returned here?

pkg/gofr/subscriber.go
```diff
@@ -24,39 +24,63 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
-		if err == nil {
-			msg.Commit()
-		} else {
-			s.container.Logger.Errorf("error in handler for topic %s: %v", topic, err)
-		}
+	msg, err := s.container.GetSubscriber().Subscribe(ctx, topic)
+
+	if errors.Is(err, kafka.ErrConsumerGroupNotProvided) {
+		s.container.Logger.Error("cannot subscribe as consumer_id is not provided in configs")
+		return err
```
ccoVeille, line 51: I would wrap the information currently added in the logger in the error

fmt.Errorf("%w whatever")

ccoVeille (commented):
Thanks for taking back the PR

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -19,6 +19,7 @@ import (
 const (
 	publicBroker  = "broker.hivemq.com"
 	messageBuffer = 10
+	closeTimeout  = 250
```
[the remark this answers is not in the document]
Umang01-hash: The Disconnect function accepts only `uint` as valid duration. So can't convert that closeTimeout to time.Duration. But will include it in name.

```go
	// Disconnect will end the connection with the server, but not before waiting
	// the specified number of milliseconds to wait for existing work to be
	// completed.
	Disconnect(quiesce uint)
```

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -338,19 +341,25 @@ func (m *MQTT) Unsubscribe(topic string) error {
 		close(sub.msgs)
 		delete(m.subscriptions, topic)
 	}
-	m.mu.RUnlock()
+
+	return nil
+}
+
+func (m *MQTT) Close(_ context.Context) error {
+	m.Client.Disconnect(closeTimeout)
```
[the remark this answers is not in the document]
Umang01-hash: Sure I have included it in MQTT Config, so the user can define it while passing configs to New method of mqtt package, but when initializing it using GoFr env variables (using container) we have set the default value to 250 milliseconds so that user don't have to worry when using default pub-sub configurations.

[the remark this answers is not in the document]
Umang01-hash: No it doesnot.

```go
	// Disconnect will end the connection with the server, but not before waiting
	// the specified number of milliseconds to wait for existing work to be
	// completed.
	Disconnect(quiesce uint)
```

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -338,19 +341,25 @@ func (m *MQTT) Unsubscribe(topic string) error {
...
 
 	return nil
 }
 
 func (m *MQTT) Disconnect(waitTime uint) {
 	m.mu.RLock()
+	defer m.mu.RUnlock()
+
 	for topic := range m.subscriptions {
 		_ = m.Unsubscribe(topic)
```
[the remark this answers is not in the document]
Umang01-hash: I have caught each error and used errors.Join to append them and returned at the end.
Also i am logging the errors as soon as we catch them.

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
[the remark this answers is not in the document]
ccoVeille: Indeed,it's not funny then.

What about using a time.Duration, then use .Milliseconds

https://pkg.go.dev/time#Duration.Milliseconds

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
Umang01-hash: No it doesn't:

```go
// RUnlock undoes a single RLock call;
// it does not affect other simultaneous readers.
// It is a run-time error if rw is not locked for reading
// on entry to RUnlock.
func (rw *RWMutex) RUnlock() {
	if race.Enabled {
		_ = rw.w.state
		race.ReleaseMerge(unsafe.Pointer(&rw.writerSem))
		race.Disable()
	}
	if r := rw.readerCount.Add(-1); r < 0 {
		// Outlined slow-path to allow the fast-path to be inlined
		rw.rUnlockSlow(r)
	}
	if race.Enabled {
		race.Enable()
	}
}
```

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
Umang01-hash: Because RUnlock method doesn't return any error

pkg/gofr/metrics_server.go
Umang01-hash: Sure changed back to srv

pkg/gofr/datasource/pubsub/mqtt/mqtt_test.go
```diff
@@ -561,3 +562,28 @@ func getMockMQTT(t *testing.T, conf *Config) (*gomock.Controller, *MQTT, *MockCl
 
 	return ctrl, mq, mockClient, mockMetrics, mockToken
 }
+
+func TestMQTT_Close(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	ctx := context.TODO()
```
ccoVeille, line 570: It's a test. Todo is for helping people to add context in code

Here you could use


```suggestion
	ctx := context.Background()
```

pkg/gofr/subscriber.go
Umang01-hash: Corrected and returned `err`

pkg/gofr/subscriber.go
Umang01-hash: Yups sure moved the logging statement to the caller

pkg/gofr/subscriber.go
Umang01-hash: I prefer using **logger.ErrorF*()* directly if i need a formatted log error message.

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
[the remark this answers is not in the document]
Umang01-hash: Sure done

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -338,19 +341,25 @@ func (m *MQTT) Unsubscribe(topic string) error {
...
 	}
```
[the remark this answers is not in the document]
Umang01-hash: Do you want me to include `mutex` as a filed in config?

pkg/gofr/subscriber.go
```diff
@@ -24,39 +24,63 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
 
-		if errors.Is(err, kafka.ErrConsumerGroupNotProvided) {
-			s.container.Logger.Errorf("cannot subscribe as consumer_id is not provided in configs")
-			return
-		} else if err != nil {
-			s.container.Logger.Errorf("error while reading from topic %v, err: %v", topic, err.Error())
-			continue
-		}
+func (s *SubscriptionManager) handleSubscription(parentCtx context.Context, topic string, handler SubscribeFunc) error {
+	ctx, done := context.WithCancel(parentCtx)
```
ccoVeille, line 44: Like in previous PR, I font understand this context cancelation here. If it was something made in the caller of this function, maybe. But here, I don't understand. A comment would help if it's OK to do it. But, I really don't get it

pkg/gofr/datasource/pubsub/kafka/kafka.go
```diff
@@ -219,15 +219,20 @@ func (k *kafkaClient) Subscribe(ctx context.Context, topic string) (*pubsub.Mess
 	return m, err
 }
 
-func (k *kafkaClient) Close() error {
-	err := k.writer.Close()
-	if err != nil {
-		k.logger.Errorf("failed to close kafka writer, error: %v", err)
+func (k *kafkaClient) Close(_ context.Context) (err error) {
```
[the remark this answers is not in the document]
Umang01-hash: Sure i will remove context from close method in Client interface also just like Reader and Writer interface

ccoVeille (commented):
Thanks for taking back the PR

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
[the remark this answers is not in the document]
Umang01-hash: But i don't feel like we should add it as a part of config,

pkg/gofr/datasource/pubsub/mqtt/mqtt_test.go
```diff
@@ -561,3 +562,28 @@ func getMockMQTT(t *testing.T, conf *Config) (*gomock.Controller, *MQTT, *MockCl
...
+	// Close the client
+	client.Close(context.Background())
+
+	// Allow some time for the client to disconnects
+	time.Sleep(closeTimeout * time.Millisecond)
+
+	err := client.Publish(ctx, "test", []byte("hello"))
+
+	// Assertions
+	assert.NotNil(t, err)
```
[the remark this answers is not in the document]
Umang01-hash: Changed

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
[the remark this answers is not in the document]
Umang01-hash: We don't want to create a env variable for closeTimeout, as firstly we only can't create it for MQTT we shall also then provide support of this env for other pubsub brokers or maybe other DB's wherever applicable.

Currently when creating a **new mqtt client through container** we are using default close timeout value else what is the value set when using `New` to create a mqtt client, we are using that.

pkg/gofr/subscriber.go
Umang01-hash: I think we can remove this ctx formation with cancellation and directly use the parent context.

pkg/gofr/datasource/pubsub/google/google_test.go
```diff
@@ -173,3 +173,21 @@ func Test_validateConfigs(t *testing.T) {
...
+	err := g.Close()
+
+	assert.NoError(t, err)
+
+	// client empty
+	g = &googleClient{}
+
+	err = g.Close()
+
+	assert.Nil(t, err)
```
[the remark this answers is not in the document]
Umang01-hash: Changed

pkg/gofr/datasource/pubsub/kafka/kafka_test.go
```diff
@@ -294,16 +300,13 @@ func TestKafkaClient_CloseError(t *testing.T) {
...
-		k.logger = logger
+	logger := logging.NewMockLogger(logging.ERROR)
+	k.logger = logger
 
-		err = k.Close()
-	})
+	err = k.Close()
 
 	assert.Error(t, err)
 	assert.Equal(t, errClose, err)
```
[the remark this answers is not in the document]
Umang01-hash: Changed

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -19,6 +19,7 @@ import (
 const (
 	publicBroker  = "broker.hivemq.com"
 	messageBuffer = 10
+	CloseTimeout  = time.Duration(250)
```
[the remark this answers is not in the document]
Umang01-hash: Sure will change the `CloseTimeout` type to time.Duration
And change the const value as suggested.

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -338,19 +342,39 @@ func (m *MQTT) Unsubscribe(topic string) error {
...
 }
 
-func (m *MQTT) Disconnect(waitTime uint) {
+func (m *MQTT) Close() error {
+	if m.config.CloseTimeout != 0 {
+		m.Client.Disconnect(m.config.CloseTimeout)
+	}
+
+	// setting default value if value is not populated
+	m.Client.Disconnect(uint(CloseTimeout.Milliseconds()))
```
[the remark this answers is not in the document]
Umang01-hash: This is for the scenario when the user has not populated the `CloseTimeout` value in Configs while creating new client using `New` method, in that case if we go to default value it will be zero, so it will skip the condition
**if m.config.CloseTimeout != 0 {**


therefore setting our constant value here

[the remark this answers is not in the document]
Umang01-hash: If the user has defined the value we are picking it from config

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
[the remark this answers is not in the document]
Umang01-hash: Done

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -338,19 +342,41 @@ func (m *MQTT) Unsubscribe(topic string) error {
...
+		m.Client.Disconnect(uint(m.config.CloseTimeout.Milliseconds()))
+
+		return nil
+	}
+
+	// setting default value if value is not populated
+	m.Client.Disconnect(uint(m.config.CloseTimeout.Milliseconds()))
+
+	return nil
+}
```
ccoVeille, line 360: Here is what I would suggest


```suggestion
func (m *MQTT) Close() error {

	timeout := m.config.CloseTimeout

	if timeout == 0 {
	    // use default if unset
		  timeout = CloseTimeout
	}
	m.Client.Disconnect(uint(timeout.Milliseconds()))

	return nil
}
```

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -338,19 +342,41 @@ func (m *MQTT) Unsubscribe(topic string) error {
...
-func (m *MQTT) Disconnect(waitTime uint) {
+func (m *MQTT) Close() error {
+	if m.config.CloseTimeout != 0 {
+		m.Client.Disconnect(uint(m.config.CloseTimeout.Milliseconds()))
+
+		return nil
+	}
+
+	// setting default value if value is not populated
+	m.Client.Disconnect(uint(m.config.CloseTimeout.Milliseconds()))
```
ccoVeille, line 357: Both branch are using the same code

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
Umang01-hash: Ok seems better. Will implement in this way

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
Umang01-hash: Updated the code

pkg/gofr/subscriber.go
```diff
@@ -24,39 +22,56 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
+	err = func(ctx *Context) error {
+		// TODO : Move panic recovery at central location which will manage for all the different cases.
+		defer func() {
+			panicRecovery(recover(), ctx.Logger)
+		}()
+
+		return handler(ctx)
+	}(msgCtx)
+
+	// commit the message if the subscription function does not return error
```
ccoVeille, line 64: Shouldn't this comment be a few line further?

I mean isn't this line about the commit

pkg/gofr/subscriber.go
```diff
@@ -24,39 +22,56 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
+	if err != nil {
+		s.container.Logger.Errorf("error in handler for topic %s: %v", topic, err)
+		return nil
```
ccoVeille, line 67: Shouldn't you return the error here?

I mean you are returning it when reading the topic, so here it would make sense, no?

pkg/gofr/subscriber_test.go
```diff
@@ -102,7 +111,7 @@ func TestSubscriptionManager_SubscribeError(t *testing.T) {
 
 		// Run the subscriber in a goroutine
 		go func() {
-			subscriptionManager.startSubscriber("abc",
+			_ = subscriptionManager.startSubscriber(context.Background(), "abc",
```
ccoVeille, line 114: Shouldn't you return the error and test it here with assert.NoError?

pkg/gofr/subscriber.go
```diff
@@ -24,39 +22,56 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
 
-		if errors.Is(err, kafka.ErrConsumerGroupNotProvided) {
-			s.container.Logger.Errorf("cannot subscribe as consumer_id is not provided in configs")
-			return
-		} else if err != nil {
-			s.container.Logger.Errorf("error while reading from topic %v, err: %v", topic, err.Error())
-			continue
-		}
+func (s *SubscriptionManager) handleSubscription(parentCtx context.Context, topic string, handler SubscribeFunc) error {
+	msg, err := s.container.GetSubscriber().Subscribe(parentCtx, topic)
```
ccoVeille, line 42:
```suggestion
func (s *SubscriptionManager) handleSubscription(ctx context.Context, topic string, handler SubscribeFunc) error {
	msg, err := s.container.GetSubscriber().Subscribe(ctx, topic)
```

Let's go back to classic naming for a context

pkg/gofr/subscriber.go
```diff
@@ -24,39 +22,56 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
+		s.container.Logger.Errorf("error while reading from topic %v, err: %v", topic, err.Error())
+
+		return err
+	}
+
+	if msg == nil {
+		return nil
 	}
+
+	msgCtx := newContext(nil, msg, s.container)
```
ccoVeille, line 54: I'm a bit annoyed by the fact you are creating a context, while you already have one.

My main issue is not about having a new context. But the fact, the ctx cancelation won't cancel the message one

ccoVeille: I read the newContext code, and it uses the context from the msg one via using `msg.Context()` it should be clearer with a comment or renaming the method `newContext` to something like `createContextFromMessage` or something like that

ccoVeille:
My review may let you the j I'm defensive about this PR, but my expectation for getting this feature are high

pkg/gofr/subscriber.go
Umang01-hash: can't rename it to `createContextFromMessage` specifically as it is being used by crud handlers,cmd , subscriber etc. but added a comment from clarity

pkg/gofr/subscriber.go
Umang01-hash: Sure

pkg/gofr/subscriber_test.go
Umang01-hash: Since the test method is `TestSubscriptionManager_SubscribeError `, i have added assert.Contains() for error checking as `assert.ErrorIs` was resulting in failure:

```go
	expected: "subscription error: subscription error"
        	            	in chain: "subscription error: subscription error"
        	            		"subscription error"
```

pkg/gofr/subscriber.go
Umang01-hash: yes corrected that

pkg/gofr/subscriber.go
Umang01-hash: Yups moved it downwards

Umang01-hash:
> My review may let you the j I'm defensive about this PR, but my expectation for getting this feature are high

@ccoVeille we always appreciate the review comments given by our reviewers and i understand you have been working on reviewing this feature from long. So it all comes together for making it better only🙂

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -19,6 +19,7 @@ import (
 const (
 	publicBroker  = "broker.hivemq.com"
 	messageBuffer = 10
+	CloseTimeout  = 250 * time.Millisecond
```
ccoVeille, line 22: This would be clearer, maybe and I don't think we need an exported constant

```suggestion
	defaultCloseTimeout  = 250 * time.Millisecond
```

pkg/gofr/datasource/pubsub/mqtt/mqtt_test.go
```diff
@@ -561,3 +563,28 @@ func getMockMQTT(t *testing.T, conf *Config) (*gomock.Controller, *MQTT, *MockCl
...
+	_ = client.Close()
+
+	// Allow some time for the client to disconnects
+	time.Sleep(CloseTimeout)
+
+	err := client.Publish(ctx, "test", []byte("hello"))
+
+	// Assertions
+	assert.Error(t, err)
+	assert.Equal(t, "not Connected", err.Error())
```
ccoVeille, line 589:
```suggestion
	assert.ErrorContains(t, err, "not Connected")
```

pkg/gofr/gofr.go
```diff
@@ -2,15 +2,20 @@ package gofr
...
+	"os/signal"
 	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
+	"golang.org/x/sync/errgroup"
+
```
ccoVeille, line 18:
```suggestion
```

Remove the line and sort the import

pkg/gofr/subscriber_test.go
```diff
@@ -72,10 +79,12 @@ func TestSubscriptionManager_HandlerError(t *testing.T) {
 
 		// Run the subscriber in a goroutine
 		go func() {
-			subscriptionManager.startSubscriber("test-topic",
+			err := subscriptionManager.startSubscriber(context.Background(), "test-topic",
 				func(*Context) error {
 					return handleError("error in test-topic")
 				})
+
+			assert.Contains(t, handleError("error in test-topic").Error(), err.Error())
```
ccoVeille, line 87:
```suggestion
			assert.ErrorContains(t, err,  handleError("error in test-topic").Error())
```

or 

```suggestion
			assert.ErrorContains(t, err, "error in test-topic")
```

or 


```suggestion
			assert.ErrorIs(t, err, handleError("error in test-topic"))
```

Maybe you should return a sentinel and use

```suggestion
      assert.ErrorIs(t, err, ErrSentinel)
			assert.ErrorContains(t, err, "error in test-topic")
```

pkg/gofr/subscriber.go
ccoVeille: Thanks

pkg/gofr/subscriber.go
ccoVeille: My point was to use

```go
return fmt.Errorf("%w whatever")
```

So the caller would have context

pkg/gofr/subscriber_test.go
Umang01-hash: Changed

pkg/gofr/gofr.go
Umang01-hash: sure

pkg/gofr/datasource/pubsub/mqtt/mqtt_test.go
Umang01-hash: Changed

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
Umang01-hash: We are using this constant inside container's **createMqttPubSub** method, so it needs to be exported.
Will change name to `DefaultCloseTimeout`

ccoVeille (approved):

ccoVeille (changes requested):

pkg/gofr/gofr.go
```diff
@@ -195,17 +184,99 @@ func (a *App) Run() {
...
+		err = errors.Join(err, a.metricServer.Shutdown(ctx))
+	}
+
+	err = errors.Join(err, a.container.Close(ctx))
+	if err != nil {
+		a.container.Logger.Errorf("error while shutting down: %v", err)
+		return err
+	}
+
+	a.container.Logger.Infof("Application shutdown complete")
```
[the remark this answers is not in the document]
Umang01-hash: Changed

pkg/gofr/grpc.go
```diff
@@ -44,3 +45,9 @@ func (g *grpcServer) Run(c *container.Container) {
 		return
 	}
 }
+
+func (g *grpcServer) Shutdown(_ context.Context) error {
+	g.server.GracefulStop()
```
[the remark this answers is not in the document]
Umang01-hash: @ccoVeille yups , i have improved the usage of context and took care of context expiration in the updated code please review and let me know if it matches your expectation

pkg/gofr/grpc.go
```diff
@@ -39,8 +43,38 @@ func (g *grpcServer) Run(c *container.Container) {
...
+
+	go func() {
+		s := <-sigCh
+		c.Logger.Infof("received signal %v, shutting down gracefully", s)
+
+		if err := g.Shutdown(context.Background()); err != nil {
+			c.Logger.Errorf("error during grpc server shutdown: %v", err)
+		}
+	}()
+}
```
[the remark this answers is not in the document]
Umang01-hash: Oh my bad! Yups right only main should listen signals, i missed that. Will restore the previous code

pkg/gofr/grpc.go
```diff
@@ -39,8 +43,38 @@ func (g *grpcServer) Run(c *container.Container) {
...
+func (g *grpcServer) Shutdown(ctx context.Context) error {
+	stopCh := make(chan struct{})
+	go func() {
+		g.server.GracefulStop()
+		close(stopCh)
+	}()
+
+	select {
+	case <-ctx.Done():
+		g.server.Stop() // Force stop if context is done
```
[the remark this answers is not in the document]
Umang01-hash: Yups, i read it's godoc description:

// Stop stops the gRPC server. It immediately closes all open
// connections and listeners.
// It cancels all active RPCs on the server side and the corresponding
// pending RPCs on the client side will get notified by connection
// errors.

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -338,19 +342,41 @@ func (m *MQTT) Unsubscribe(topic string) error {
...
-func (m *MQTT) Disconnect(waitTime uint) {
+func (m *MQTT) Close() error {
+	timeout := m.config.CloseTimeout
+
+	if timeout == 0 {
+		// setting default value if value is not populated
+		timeout = DefaultCloseTimeout
+	}
+
+	m.Client.Disconnect(uint(timeout.Milliseconds()))
```
vipul-rawat, line 357: This should return the unsubscribe error.
```suggestion
	return m.Client.Disconnect(uint(timeout.Milliseconds()))
```

pkg/gofr/datasource/pubsub/mqtt/mqtt_test.go
```diff
@@ -561,3 +563,28 @@ func getMockMQTT(t *testing.T, conf *Config) (*gomock.Controller, *MQTT, *MockCl
...
 }
+
+func TestMQTT_Close(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	ctx := context.Background()
+	mockMetrics := NewMockMetrics(ctrl)
+	mockLogger := logging.NewMockLogger(logging.ERROR)
+	client := New(&Config{}, mockLogger, mockMetrics)
```
vipul-rawat, line 574: Use mockClient from `getMockMQTT`

pkg/gofr/grpc.go
```diff
@@ -39,8 +43,38 @@ func (g *grpcServer) Run(c *container.Container) {
...
+
+func (g *grpcServer) Shutdown(ctx context.Context) error {
+	stopCh := make(chan struct{})
+	go func() {
+		g.server.GracefulStop()
+		close(stopCh)
+	}()
+
+	select {
+	case <-ctx.Done():
```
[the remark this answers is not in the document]
vipul-rawat: So basically, for the Closers that do not require a context, should we wait using the `stopChan`, only to return when the close process is finished?

Do we need to make sure that the close does not close the client connection before any ongoing request/process is finished?

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -17,8 +17,9 @@ import (
 )
 
 const (
-	publicBroker  = "broker.hivemq.com"
-	messageBuffer = 10
+	publicBroker        = "broker.hivemq.com"
+	messageBuffer       = 10
+	DefaultCloseTimeout = 250 * time.Millisecond
```
aryanmehrotra, line 22: Why not keep `DefaultCloseTimeout` as 0.
As we are not keeping any wait time at other places.
Moreover, why not to stop subscribing first and the disconnect?

aryanmehrotra (changes requested):

vipul-rawat (changes requested):
The HTTP Server is not actually waiting for the request to complete before closing.

pkg/gofr/grpc.go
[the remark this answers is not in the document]
ccoVeille: Thanks for checking

pkg/gofr/grpc.go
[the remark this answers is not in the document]
ccoVeille: I'm unsure what you meant. So, please let me rephrase.

We have 

When a signal is caught,
We create a context that expires with a timeout
Then we call, App.Shutdown with this context.

So the code in the Shutdown is supposed to stop all services, before the context expiration.

So every single closer we call should consider the context, even the one that don't have a context.

Which mean we need to exit the closer when the context is expired, otherwise we don't respect the timeout.

All this implies that we do are working with an app that is shutting down.

We are trying to gracefully stop the open connections before the timeout

It implies that yes, we might close opened connection if the clients/closer doesn't reply. But the app is stopped, so any new connection is refused.

So for me, we should get out each closer as soon the timeout is reached.

Is that clearer ?

That said, right now we are shutting down resources sequentially, we could shut them down in parallel, with a small refactoring. Of course, we might have to consider the consequence of doing that.

ccoVeille (approved):

pkg/gofr/grpc.go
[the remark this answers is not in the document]
ccoVeille: > So maybe, we could copy this code in a function that accepts a function as argument in App.Shutdown\n\nSo you could use it for the following clients Kafka, Google, SQL …\n\nOr you copy paste this code in each Shutdowner/Closer called by App.Shutdown

This is the point where I would like to get feedbacks from you guys

pkg/gofr/datasource/pubsub/mqtt/mqtt_test.go
Umang01-hash: Used

pkg/gofr/datasource/pubsub/interface.go
```diff
@@ -8,6 +8,8 @@ import (
 	"gofr.dev/pkg/gofr/datasource"
 )
 
+//go:generate go run go.uber.org/mock/mockgen -source=interface.go -destination=mock_interfaces.go -package=mqtt
```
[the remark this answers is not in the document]
Umang01-hash: Sure

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
Umang01-hash: sure

pkg/gofr/datasource/pubsub/mqtt/mqtt_test.go
```diff
@@ -561,3 +562,18 @@ func getMockMQTT(t *testing.T, conf *Config) (*gomock.Controller, *MQTT, *MockCl
 
 	return ctrl, mq, mockClient, mockMetrics, mockToken
 }
+
+func TestMQTT_Close(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	ctrl, client, mockMQTT, _, _ := getMockMQTT(t, mockConfigs)
+	defer ctrl.Finish()
```
ccoVeille, line 571: This comment is out of the PR scope

But, I'm seeing a lot of defer like this one

Why not, adding a t.Cleanup to getMockMQTT and gomock.NewController

And remove all this.

I'm just starting the discussion about it,to see if there could be an agreement on doing so.

I would have to be done in another PR, and we should consider creating an issue referencing this thread

Umang01-hash: Nice thought we can do it in a separate issue

ccoVeille: I opened

- https://github.com/gofr-dev/gofr/issues/855

pkg/gofr/grpc.go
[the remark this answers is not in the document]
vipul-rawat: > It implies that yes, we might close opened connection if the clients/closer doesn't reply. But the app is stopped, so any new connection is refused.
> So for me, we should get out each closer as soon the timeout is reached.

Yes, we should not accept any new connection, also we need to finish the ongoing requests first.

> This is the point where I would like to get feedbacks from you guys

Yes we should have all the closers to use context for their closing, I think we can have a central closer that takes

Umang01-hash:
> The HTTP Server is not actually waiting for the request to complete before closing.

I have corrected that implementation. Also used a common closer function to gracefully shutdown the existing servre.
Please re-review the PR.

pkg/gofr/shutdown_test.go
```diff
@@ -0,0 +1,41 @@
...
+	mockShutdownFunc := func(_ context.Context) error {
+		// Simulate a quick shutdown
+		time.Sleep(50 * time.Millisecond)
+		return nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+	defer cancel()
+
+	err := shutdownWithTimeout(ctx, mockShutdownFunc, func() {})
```
ccoVeille, line 38:
```suggestion
	err := shutdownWithTimeout(ctx, mockShutdownFunc, nil)
```

pkg/gofr/http_server_test.go
```diff
@@ -85,3 +85,76 @@ func TestRegisterProfillingRoutes(t *testing.T) {
...
+	// Simulate a delay in the shutdown process to trigger context timeout
+	shutdownCh := make(chan error, 1)
+	go func() {
+		time.Sleep(300 * time.Millisecond) // Delay longer than the context timeout
+		shutdownCh <- server.Shutdown(ctx)
+	}()
+
+	err := <-shutdownCh
+
+	assert.Error(t, err, "context deadline exceeded", "Expected context deadline exceeded error")
```
ccoVeille, line 159:
```suggestion
	assert.ErrorIs(t, err, context.DeadlineExceeded, "Expected context deadline exceeded error")
```

pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -338,19 +342,39 @@ func (m *MQTT) Unsubscribe(topic string) error {
...
 }
 
-func (m *MQTT) Disconnect(waitTime uint) {
+func (m *MQTT) Close() error {
+	timeout := m.config.CloseTimeout
+
+	if timeout == 0 {
+		// setting default value if value is not populated
+		timeout = DefaultCloseTimeout
+	}
```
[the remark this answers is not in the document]
Umang01-hash: Sure will remove the constant and code

pkg/gofr/http_server.go
```diff
@@ -47,15 +49,28 @@ func (s *httpServer) RegisterProfilingRoutes() {
...
+}
+
+func (s *httpServer) Shutdown(ctx context.Context) error {
+	if s.srv == nil {
+		return nil
+	}
+
+	return shutdownWithTimeout(ctx, func(ctx context.Context) error {
+		return s.srv.Shutdown(ctx)
+	}, func() { s.srv.Close() })
```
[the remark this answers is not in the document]
Umang01-hash: Doing this will result in following error: `Cannot use 's.srv.Close()' (type error) as the type func()`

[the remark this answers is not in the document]
ccoVeille: Then it mean you should handle the error here, with a log or something else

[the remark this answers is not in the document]
Umang01-hash: done logged the error

pkg/gofr/http_server.go
```diff
@@ -47,15 +50,32 @@ func (s *httpServer) RegisterProfilingRoutes() {
...
+		return nil
+	}
+
+	return shutdownWithTimeout(ctx, func(ctx context.Context) error {
+		return s.srv.Shutdown(ctx)
+	}, func() {
+		if err := s.srv.Close(); err != nil {
+			logger.Errorf("Error closing http server: %v", err)
+		}
+	})
```
ccoVeille, line 80: I would have modified shutdownWithTimeout signature to be a func() error for second function

Then handling the error in this function, avoiding to deal with  the logger you have to import here

So something like this
https://github.com/gofr-dev/gofr/pull/837#discussion_r1687479094

pkg/gofr/shutdown.go
```diff
@@ -0,0 +1,24 @@
...
+	case <-ctx.Done():
+		if forceCloseFunc != nil {
+			forceCloseFunc()
+		}
+
+		return ctx.Err()
+	case err := <-errCh:
+		return err
+	}
+}
```
ccoVeille, line 24:
```suggestion
func shutdownWithTimeout(ctx context.Context, shutdownFunc func(ctx context.Context) error, forceCloseFunc func() error) error {
	errCh := make(chan error, 1)
	go func() {
		errCh <- shutdownFunc(ctx)
	}()

	select {
	case <-ctx.Done():
    err := ctx.Err()
		if forceCloseFunc != nil {
			err = errors.Join(err, forceCloseFunc())
		}

		return ctx.Err()
	case err := <-errCh:
		return err
	}
}
```

pkg/gofr/http_server.go
Umang01-hash: yups that can also be done, refactored the signature to that

ccoVeille (approved):
LGTM 🤞

pkg/gofr/grpc.go
```diff
@@ -44,3 +45,21 @@ func (g *grpcServer) Run(c *container.Container) {
...
+		}()
+
+		select {
+		case <-ctx.Done():
+			g.server.Stop() // Force stop if context is done
+			return ctx.Err()
+		case <-stopCh:
+			return nil
+		}
+	}, nil)
```
ccoVeille, line 64: I'm surprised here

We are using the shutdownWithTimeout but we have somehow a replicated code using what is in shutdownWithTimeout 

Am I missing something?

pkg/gofr/shutdown.go
```diff
@@ -0,0 +1,26 @@
+package gofr
+
+import (
+	"context"
+	"errors"
+)
+
+// shutdownWithTimeout handles the shutdown process with context timeout.
+func shutdownWithTimeout(ctx context.Context, shutdownFunc func(ctx context.Context) error, forceCloseFunc func() error) error {
```
ccoVeille, line 9:
```suggestion
// shutdownWithContext handles the shutdown process with context timeout.
func shutdownWithContext(ctx context.Context, shutdownFunc func(ctx context.Context) error, forceCloseFunc func() error) error {
```

There is no timeout in the method

Or maybe


```suggestion
// shutdownRespectContextExpiration handles the shutdown process with context timeout.
func shutdownRespectContextExpiration(ctx context.Context, shutdownFunc func(ctx context.Context) error, forceCloseFunc func() error) error {
```

Wording is a pain

ccoVeille: I'm fine with it

pkg/gofr/grpc.go
vipul-rawat: Yes, @Umang01-hash please use the `shutdownWithTimeout` function, as we can pass `GracefulStop` in normal server stop and `Stop` for force stopping the gRPC server.

As for the time.After is concerned, I don't think that would be necessary as we are cancelling the context after the timeout, and we have `<- ctx.Done()` to catch those scenarios.

@Umang01-hash please also add comments on the specific functionality to better document the same!

Umang01-hash: Passed `GracefulStop` and `Stop` for force stopping the gRPC server.

Also added the comments

pkg/gofr/http_server_test.go
```diff
@@ -85,3 +85,76 @@ func TestRegisterProfillingRoutes(t *testing.T) {
...
+	server := &httpServer{
+		router: router,
+		port:   8080,
+	}
+
+	// Start the server
+	go server.Run(c)
+
+	// Create a context with a timeout to test the shutdown
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
```
vipul-rawat, line 147: Can we reduce the timeouts in tests? I don't think even 200 ms is required.

pkg/gofr/subscriber.go
```diff
@@ -24,39 +22,58 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
-		if msg == nil {
-			continue
+		select {
+		case <-ctx.Done():
+			s.container.Logger.Infof("shutting down subscriber for topic %s", topic)
+			return nil
+		default:
+			err := s.handleSubscription(ctx, topic, handler)
+			if err != nil {
+				return err
```
vipul-rawat, line 35: Why are we stopping the subscription when there is an error, this is breaking the previous implementation, we tried not to commit the message so that it can be rerereadn the next pass.

vipul-rawat: We can log the error and continue subscribing to the same message!

PS: The error was provided to the handler for the very same purpose, to enable users to commit the message that they consider consumed.

ccoVeille (approved):
LGTM 👍

pkg/gofr/http_server_test.go
Umang01-hash: done reduced

pkg/gofr/gofr.go
```diff
@@ -136,51 +142,38 @@ func (a *App) Run() {
 		a.cmd.Run(a.container)
 	}
 
+	// Create a context that is canceled on receiving termination signals
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+	defer stop()
+
+	// Goroutine to handle shutdown when context is cancele
```
aryanmehrotra, line 149: type : cancele

aryanmehrotra (changes requested):

pkg/gofr/gofr.go
Umang01-hash: sure corrected

aryanmehrotra (dismissed):

pkg/gofr/datasource/redis/redis.go
```diff
@@ -63,6 +63,15 @@ func NewClient(c config.Config, logger datasource.Logger, metrics Metrics) *Redi
 	return &Redis{Client: rc, config: redisConfig, logger: logger}
 }
 
+// Close shuts down the Redis client, ensuring the current dataset is saved before exiting.
```
ccoVeille, line 66: Is this still a shutdown then?

pkg/gofr/datasource/redis/redis.go
```diff
@@ -63,6 +63,15 @@ func NewClient(c config.Config, logger datasource.Logger, metrics Metrics) *Redi
...
+func (r *Redis) Close(_ context.Context) error {
+	if r.Client != nil {
+		return r.Client.Close()
```
ccoVeille, line 69: Could you please explain why you changed from shutdown to close?

I'm OK with it,if it makes sense. I'm just asking.

Please rewrite the code to remove the context from the signature, and use the shutdowner method you already used for other closer without context then

vipul-rawat: The Close method is to close the client, the shutdown was closing and shutting down the actual instance of redis.

Umang01-hash: I have refactored the code and removed the context from signature

vipul-rawat (approved):

aryanmehrotra (approved):

ccoVeille:
Thanks everyone.

Issue #844, opened 2024-07-12
Add test about logger interface, especially Fatal


ccoVeille:
Fatal and Fatalf were removed from the logger and nothing caught they shouldn't be removed.

While this method are an anti pattern. They exists and should be tested.

The fact they are calling os.Exit(1) doesn't help to test them. But it's possible with a small refactoring maybe.

> Apparently, nothing was using it, and one of them implementation was doing a deep exit:
>
> Calling a os.Exit(1) in the code, slaughtering all the stacks and the possible defer, context cancelation.
>
> I missed people may use it by calling the framework

_Originally posted by @ccoVeille in https://github.com/gofr-dev/gofr/issues/841#issuecomment-2225437051_

ccoVeille:
One solution could be to add an exiter, that can be overloaded for test 

```go
var exiter func() = func() { os.Exit(1)}

func(l *Logger) Fatal(args ...any) {
      l.logger.Log("fatal", args...)
      exiter()
}

func overloadExiter(t *testing.T, f func()) {
      exiter = f
      t.Cleanup(func(){
         exiter =func() { os.Exit(1) }
      })
}
```

This is pseudo code, just an idea, it can be improved

srijan-27:
@ccoVeille - Yes, there should be tests for these methods.. although I was trying to add tests, but in the need of urgency to fix the broken release, I just went ahead with the fix. I'm looking into it, and will add the tests.

ccoVeille:
Yes, of course.

BTW, I found a photo of me when I removed the methods 😱😬😁😅

![ikg6tlu1slcd1.jpeg](https://github.com/user-attachments/assets/6dbc95e0-d35f-4cd4-ba3c-872c123ec3e3)

vipul-rawat:
> One solution could be to add an exiter, that can be overloaded for test
> 
> ```go
> var exiter func() = func() { os.Exit(1)}
> 
> func(l *Logger) Fatal(args ...any) {
>       l.logger.Log("fatal", args...)
>       exiter()
> }
> 
> func overloadExiter(t *testing.T, f func()) {
>       exiter = f
>       t.Cleanup(func(){
>          exiter =func() { os.Exit(1) }
>       })
> }
> ```
> 
> This is pseudo code, just an idea, it can be improved

Yes, we have to encapsulate the os.Exit() function! It is the only way to test for standard library functions.
```go
var exit = os.Exit

func (l *logger) Fatalf(format string, args ...interface{}) {
[...]

PR #848, opened 2024-07-16
Fix logger tests

srijan-27:
- Added test for fatal log method (os.Exit issue)
- Fixed validations for warn log tests
- Closes: #844

**Checklist:**

### DIFF
--- a/pkg/gofr/logging/logger_test.go
+++ b/pkg/gofr/logging/logger_test.go
@@ -3,7 +3,6 @@ package logging
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -155,9 +154,9 @@ func TestLogger_LevelFatal(t *testing.T) {
 	err := cmd.Wait()
 
 	var e *exec.ExitError
-	if !errors.As(err, &e) || e.Success() {
-		t.Fatalf("Process ran with err %v, want exit status 1", err)
-	}
+
+	require.ErrorAs(t, err, &e)
+	assert.False(t, e.Success())
 }
 
 func assertMessageInJSONLog(t *testing.T, logLine, expectation string) {
--- a/pkg/gofr/logging/logger_test.go
+++ b/pkg/gofr/logging/logger_test.go
@@ -106,9 +106,11 @@ func TestLogger_LevelWarn(t *testing.T) {
 	infoLog := testutil.StdoutOutputForFunc(printLog)
 	errLog := testutil.StderrOutputForFunc(printLog)
 
-	if strings.ContainsAny(infoLog, "NOTICE|INFO|DEBUG") && !strings.Contains(errLog, "ERROR") {
-		// Warn Log Level will not contain  DEBUG,INFO, NOTICE logs
-		t.Errorf("TestLogger_LevelDebug Failed!")
+	levels := []Level{DEBUG, INFO, NOTICE}
+
+	for i, l := range levels {
+		assert.Equal(t, false, strings.Contains(infoLog, l.String()), "TEST[%d], Failed.\n%s", i,
+			fmt.Sprintf("unexpected %s log", l))
 	}
 
 	assertMessageInJSONLog(t, errLog, "Test Error Log")
@@ -121,6 +123,7 @@ func TestLogger_LevelFatal(t *testing.T) {
 
 		logger.Debugf("%s", "Test Debug Log")
 		logger.Infof("%s", "Test Info Log")
+		logger.Logf("%s", "Test Log")
 		logger.Noticef("%s", "Test Notice Log")
 		logger.Warnf("%s", "Test Warn Log")
 		logger.Errorf("%s", "Test Error Log")
--- a/pkg/gofr/logging/logger_test.go
+++ b/pkg/gofr/logging/logger_test.go
@@ -3,9 +3,11 @@ package logging
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"strings"
 	"testing"
 
@@ -113,20 +115,48 @@ func TestLogger_LevelWarn(t *testing.T) {
 }
 
 func TestLogger_LevelFatal(t *testing.T) {
-	printLog := func() {
+	// running the failing part only when a specific env variable is set
+	if os.Getenv("EXITER") == "1" {
 		logger := NewLogger(FATAL)
+
 		logger.Debugf("%s", "Test Debug Log")
 		logger.Infof("%s", "Test Info Log")
 		logger.Noticef("%s", "Test Notice Log")
 		logger.Warnf("%s", "Test Warn Log")
 		logger.Errorf("%s", "Test Error Log")
+		logger.Fatalf("%s", "Test Fatal Log")
+
+		return
 	}
 
-	infoLog := testutil.StdoutOutputForFunc(printLog)
-	errLog := testutil.StderrOutputForFunc(printLog)
+	//nolint:gosec // starting the actual test in a different subprocess
+	cmd := exec.Command(os.Args[0], "-test.run=TestLogger_LevelFatal")
+	cmd.Env = append(os.Environ(), "EXITER=1")
+	stdout, _ := cmd.StderrPipe()
 
-	assert.Equal(t, "", infoLog, "TestLogger_LevelFatal Failed!")
-	assert.Equal(t, "", errLog, "TestLogger_LevelFatal Failed")
+	if err := cmd.Start(); err != nil {
+		t.Fatal(err)
+	}
+
+	logBytes, _ := io.ReadAll(stdout)
+	log := string(logBytes)
+
+	levels := []Level{DEBUG, INFO, NOTICE, WARN, ERROR} // levels which should not be present in case of FATAL log_level
+
+	for i, l := range levels {
+		assert.Equal(t, false, strings.Contains(log, l.String()), "TEST[%d], Failed.\n%s", i,
+			fmt.Sprintf("unexpected %s log", l))
+	}
+
+	assertMessageInJSONLog(t, log, "Test Fatal Log")
+
+	// Check that the program exited
+	err := cmd.Wait()
+
+	var e *exec.ExitError
+	if !errors.As(err, &e) || e.Success() {
+		t.Fatalf("Process ran with err %v, want exit status 1", err)
+	}
 }
 
 func assertMessageInJSONLog(t *testing.T, logLine, expectation string) {
--- a/pkg/gofr/logging/logger_test.go
+++ b/pkg/gofr/logging/logger_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"golang.org/x/term"
 
 	"gofr.dev/pkg/gofr/testutil"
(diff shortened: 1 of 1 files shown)


pkg/gofr/logging/logger_test.go
```diff
@@ -104,29 +106,60 @@ func TestLogger_LevelWarn(t *testing.T) {
...
+	//nolint:gosec // starting the actual test in a different subprocess
+	cmd := exec.Command(os.Args[0], "-test.run=TestLogger_LevelFatal")
+	cmd.Env = append(os.Environ(), "EXITER=1")
+	stdout, _ := cmd.StderrPipe()
 
-	assert.Equal(t, "", infoLog, "TestLogger_LevelFatal Failed!")
-	assert.Equal(t, "", errLog, "TestLogger_LevelFatal Failed")
+	if err := cmd.Start(); err != nil {
+		t.Fatal(err)
+	}
```
ccoVeille, line 142:
```suggestion
	require.NoError(cmd.Start())
```

pkg/gofr/logging/logger_test.go
```diff
@@ -104,29 +106,60 @@ func TestLogger_LevelWarn(t *testing.T) {
...
 		logger.Errorf("%s", "Test Error Log")
+		logger.Fatalf("%s", "Test Fatal Log")
+
+		return
 	}
 
-	infoLog := testutil.StdoutOutputForFunc(printLog)
-	errLog := testutil.StderrOutputForFunc(printLog)
+	//nolint:gosec // starting the actual test in a different subprocess
+	cmd := exec.Command(os.Args[0], "-test.run=TestLogger_LevelFatal")
```
ccoVeille, line 136: This ia dirty hack… but why not

pkg/gofr/logging/logger_test.go
```diff
@@ -104,29 +106,60 @@ func TestLogger_LevelWarn(t *testing.T) {
...
+	}
+
+	logBytes, _ := io.ReadAll(stdout)
+	log := string(logBytes)
+
+	levels := []Level{DEBUG, INFO, NOTICE, WARN, ERROR} // levels which should not be present in case of FATAL log_level
+
+	for i, l := range levels {
+		assert.Equal(t, false, strings.Contains(log, l.String()), "TEST[%d], Failed.\n%s", i,
+			fmt.Sprintf("unexpected %s log", l))
```
ccoVeille, line 151: Please simplify

```suggestion
		assert.NotContainsf(t, log, l.String(), "TEST[%d], Failed.\nunexpected %s log", i, log)
```

pkg/gofr/logging/logger_test.go
```diff
@@ -104,29 +106,60 @@ func TestLogger_LevelWarn(t *testing.T) {
...
 	}
 
 	assertMessageInJSONLog(t, errLog, "Test Error Log")
 }
 
 func TestLogger_LevelFatal(t *testing.T) {
-	printLog := func() {
+	// running the failing part only when a specific env variable is set
+	if os.Getenv("EXITER") == "1" {
```
ccoVeille, line 121: You should always use a constant prefixed with your project and to make sure it cannot conflict with existing env variable populated by various CI tools

Something like GOFR_LOGGER_EXITER

pkg/gofr/logging/logger_test.go
ccoVeille: I would have preferred something like this

https://github.com/gofr-dev/gofr/issues/844#issuecomment-2230145018

pkg/gofr/logging/logger_test.go
```diff
@@ -3,9 +3,12 @@ package logging
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"github.com/stretchr/testify/require"
```
Umang01-hash, line 8: Please sort the imports

Umang01-hash (changes requested):

pkg/gofr/logging/logger_test.go
```diff
@@ -104,29 +107,57 @@ func TestLogger_LevelWarn(t *testing.T) {
...
+		assert.Equal(t, false, strings.Contains(infoLog, l.String()), "TEST[%d], Failed.\n%s", i,
+			fmt.Sprintf("unexpected %s log", l))
 	}
 
 	assertMessageInJSONLog(t, errLog, "Test Error Log")
 }
 
 func TestLogger_LevelFatal(t *testing.T) {
-	printLog := func() {
+	// running the failing part only when a specific env variable is set
```
ccoVeille, line 121: This comment is explaining the what, not the why

maybe  something like this would be better


```suggestion
	// running the failing part only when a specific env variable is set
  // we are using this because it's one of the only way we found to test the Fatal
```

pkg/gofr/logging/logger_test.go
```diff
@@ -104,29 +107,57 @@ func TestLogger_LevelWarn(t *testing.T) {
...
+
+	assertMessageInJSONLog(t, log, "Test Fatal Log")
+
+	// Check that the program exited
+	err := cmd.Wait()
+
+	var e *exec.ExitError
+	if !errors.As(err, &e) || e.Success() {
+		t.Fatalf("Process ran with err %v, want exit status 1", err)
+	}
```
ccoVeille, line 160:
```suggestion
	require.ErrorAs(t, err, &e)
  assert.False(t, e.Success())
```

pkg/gofr/logging/logger_test.go
srijan-27: correct me if I am wrong.. but what I understood from your point is: you want `exitor` as a field in `logger` struct, and in `Fatal` methods, we can provide value for it as `true`, and in `logf` method we will check if `exitor` is true, then os.Exit(). Right?

If yes, I believe that will also not resolve our problem of testing, and we would need to write tests the same way as we are doing now (using subprocess).
Or, are you suggesting that instead of using NewLogger in test, we would create a logger struct manually?

ccoVeille: Let me write something. I'll be back soon to explain

pkg/gofr/logging/logger_test.go
```diff
@@ -3,13 +3,16 @@ package logging
 import (
```
ccoVeille, line 3: Commenting a random line

ccoVeille: Here is the current logger implementation

```go
type logger struct {
        level      Level
        normalOut  io.Writer
        errorOut   io.Writer
        isTerminal bool
        lock       chan struct{}
}

// … redacted

func (l *logger) Fatal(args ...interface{}) {
        l.logf(FATAL, "", args...)

        //nolint:revive // exit status is 1 as it denotes failure as signified by Fatal log
        os.Exit(1)
}

func (l *logger) Fatalf(format string, args ...interface{}) {
        l.logf(FATAL, format, args...)

        //nolint:revive // exit status is 1 as it denotes failure as signified by Fatal log
        os.Exit(1)
}
```

ccoVeille: Here is what I'm suggesting to consider

```go
type logger struct {
        level      Level
        normalOut  io.Writer
        errorOut   io.Writer
        isTerminal bool
        lock       chan struct{}
        exiter func()
}

// … redacted

func (l *logger) Fatal(args ...interface{}) {
        l.logf(FATAL, "", args...)

         l.exitFatal()
}

func (l *logger) Fatalf(format string, args ...interface{}) {
        l.logf(FATAL, format, args...)

        l.exitFatal()
}

func (l *logger) exitFatal() {
        if l.exiter != nil {
            // this is used for test purpose
            l.exiter()
            return
        }

        //nolint:revive // exit status is 1 as it denotes failure as signified by Fatal log
        os.Exit(1)
}

func SetExiterT(t *testing.T, l *logger, f func()) {
     t.Helper()

     l.exiter = f
}
```

Then in tests, using

```go
var called bool
logging.SetExiter(t, logger, func() {
     called = true
})

Then check for called value

**Review by ccoVeille — CHANGES_REQUESTED** [review]

File: pkg/gofr/logging/logger_test.go (hunk 8 above)
Comment (srijan-27) [reply pkg/gofr/logging/logger_test.go:3]: I think just for the testing purpose, we should not change the struct and actual implementation.. here for testing purpose we are creating a new function and toggling on if condition for actual and testing case. IMO, this should not be done.

**Review by Umang01-hash — DISMISSED** [review]

File: pkg/gofr/logging/logger_test.go (hunk 8 above)
Comment (ccoVeille) [reply pkg/gofr/logging/logger_test.go]: I know

What about using what xhd2015 suggested here then?

https://github.com/gofr-dev/gofr/issues/844#issuecomment-2230608103

File: pkg/gofr/logging/logger_test.go [hunk 9]
```diff
@@ -104,29 +106,57 @@ func TestLogger_LevelWarn(t *testing.T) {
...
 	errLog := testutil.StderrOutputForFunc(printLog)
 
-	if strings.ContainsAny(infoLog, "NOTICE|INFO|DEBUG") && !strings.Contains(errLog, "ERROR") {
-		// Warn Log Level will not contain  DEBUG,INFO, NOTICE logs
-		t.Errorf("TestLogger_LevelDebug Failed!")
+	levels := []Level{DEBUG, INFO, NOTICE}
+
+	for i, l := range levels {
+		assert.Equal(t, false, strings.Contains(infoLog, l.String()), "TEST[%d], Failed.\n%s", i,
+			fmt.Sprintf("unexpected %s log", l))
```
ccoVeille, line 113:
```suggestion
		assert.NotContainsf(t. infoLog, l.String(), "TEST[%d], Failed.\nunexpected %s log", i, l)
```

pkg/gofr/logging/logger_test.go
```diff
@@ -104,29 +106,57 @@ func TestLogger_LevelWarn(t *testing.T) {
...
-	infoLog := testutil.StdoutOutputForFunc(printLog)
-	errLog := testutil.StderrOutputForFunc(printLog)
+	//nolint:gosec // starting the actual test in a different subprocess
+	cmd := exec.Command(os.Args[0], "-test.run=TestLogger_LevelFatal")
+	cmd.Env = append(os.Environ(), "GOFR_EXITER=1")
+	stdout, _ := cmd.StderrPipe()
+
+	require.NoError(t, cmd.Start())
+
+	logBytes, _ := io.ReadAll(stdout)
```
ccoVeille, line 142:
```suggestion
	logBytes, err := io.ReadAll(stdout)
  require.NoError(t, err)
```

pkg/gofr/logging/logger_test.go
```diff
@@ -104,29 +106,57 @@ func TestLogger_LevelWarn(t *testing.T) {
...
+
+		return
 	}
 
-	infoLog := testutil.StdoutOutputForFunc(printLog)
-	errLog := testutil.StderrOutputForFunc(printLog)
+	//nolint:gosec // starting the actual test in a different subprocess
+	cmd := exec.Command(os.Args[0], "-test.run=TestLogger_LevelFatal")
+	cmd.Env = append(os.Environ(), "GOFR_EXITER=1")
+	stdout, _ := cmd.StderrPipe()
```
ccoVeille, line 138:
```suggestion
	stdout, err := cmd.StderrPipe()
  require.NoError(t, err)
```

pkg/gofr/logging/logger_test.go
ccoVeille: But I'm fine with current implementation, even if it's a dirty hack for me

ccoVeille (approved):

Umang01-hash (approved):

vipul-rawat (approved):

Issue #852, opened 2024-07-18
Issue: Failed to load config from file: ./configs/.env, Err: unexpected character


surajit-zs:
Currently, the framework gives a warning error when the .env file contains some invalid characters and proceeds to start the servers without taking any configuration from the .env file. This behavior can lead to significant issues, especially in production.

If a new build inadvertently includes an invalid env, the server will start, leading to various problems due to missing or incorrect configurations. To mitigate this risk, the application should not start if the .env file is present and contains errors. This will ensure that new pods do not come up with incorrect configurations.

ref:<img width="1347" alt="image" src="https://github.com/user-attachments/assets/2b1d8924-e61f-4265-97ae-49986a5bc76f">

srijan-27:
@surajit-zs - Yes, this should be the expected behaviour. Will update it to not start application in case of any issues.

PR #854, opened 2024-07-19
Fail app if error in config files

srijan-27:
- Fail app start in case of any error in `.env` or `.<APP_ENV>.env` files.
- Closes: #852

**Checklist:**

### DIFF
--- a/pkg/gofr/config/godotenv.go
+++ b/pkg/gofr/config/godotenv.go
@@ -1,7 +1,9 @@
 package config
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 
 	"github.com/joho/godotenv"
@@ -20,6 +22,7 @@ type logger interface {
 	Warnf(format string, a ...interface{})
 	Infof(format string, a ...interface{})
 	Debugf(format string, a ...interface{})
+	Fatalf(format string, a ...interface{})
 }
 
 func NewEnvFile(configFolder string, logger logger) Config {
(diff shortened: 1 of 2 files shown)


pkg/gofr/config/godotenv_test.go
```diff
@@ -99,20 +99,23 @@ func Test_EnvFailureWithHyphen(t *testing.T) {
...
-	// Call the function to create the .env file
-	createEnvFile(t, ".env", envData)
+	for _, file := range configFiles {
+		err := createConfigsDirectory()
+		if err != nil {
+			t.Error(err)
+		}
 
-	defer os.RemoveAll("configs")
+		createEnvFile(t, file, envData)
```
ccoVeille, line 110: You removed the defer cleanup.

You could add a t.Cleanup of the file into createEnvFile

Or maybe the configs folder is already cleaned?

I would add the t.Cleanup to createEnvFile as it could be reused

pkg/gofr/config/godotenv_test.go
```diff
@@ -99,20 +99,23 @@ func Test_EnvFailureWithHyphen(t *testing.T) {
...
 
-	env := NewEnvFile("configs", logger)
+		env := NewEnvFile("configs", logger)
+
+		assert.Equal(t, "test", env.GetOrDefault("KEY-WITH-HYPHEN", "test"), "TEST Failed.\n godotenv failure with hyphen")
+		assert.Equal(t, "", env.Get("UNABLE_TO_LOAD"), "TEST Failed.\n godotenv failure with hyphen")
 
-	assert.Equal(t, "test", env.GetOrDefault("KEY-WITH-HYPHEN", "test"), "TEST Failed.\n godotenv failure with hyphen")
-	assert.Equal(t, "", env.Get("UNABLE_TO_LOAD"), "TEST Failed.\n godotenv failure with hyphen")
+		os.RemoveAll("configs")
```
ccoVeille, line 117: This look like an error to that to me.

https://github.com/gofr-dev/gofr/pull/854#discussion_r1684278442

pkg/gofr/config/godotenv_test.go
ccoVeille: Another solution is to refactor the code of createEnvFile to path a folder as argument instead of hardcoded "configs" twice.

then to use t.TempDir to generate a folder, then pass it to createEnvFile and

ccoVeille: This would be the best solution according to me

pkg/gofr/config/godotenv.go
```diff
@@ -38,33 +41,28 @@ func (e *EnvLoader) read(folder string) {
...
-		if err != nil {
-			e.logger.Warnf("Failed to load config from file: %v, Err: %v", overrideFile, err)
-		} else {
-			e.logger.Infof("Loaded config from file: %v", overrideFile)
+	err = godotenv.Overload(overrideFile)
+	if err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
+			e.logger.Fatalf("Failed to load config from file: %v, Err: %v", overrideFile, err)
 		}
+	} else {
```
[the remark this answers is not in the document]
srijan-27: yes, this is `intentional`.. as if you see in line: 61 we are checking if err is not `fs.ErrNotExists` then we are logging err at fatal level.. so, we are skipping the file not exists error and since these files are optional.. hence the removed log.

pkg/gofr/config/godotenv.go
```diff
@@ -38,33 +41,28 @@ func (e *EnvLoader) read(folder string) {
 
 	err := godotenv.Load(defaultFile)
 	if err != nil {
-		e.logger.Warnf("Failed to load config from file: %v, Err: %v", defaultFile, err)
+		if errors.Is(err, fs.ErrNotExist) {
+			e.logger.Warnf("Failed to load config from file: %v, Err: %v", defaultFile, err)
+		} else {
+			e.logger.Fatalf("Failed to load config from file: %v, Err: %v", defaultFile, err)
+		}
```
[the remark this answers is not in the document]
srijan-27: this way we are checking the fatal condition first and hence, removing the need to `else`.. right? or does it have any other benefit which I am missing..

pkg/gofr/config/godotenv_test.go
srijan-27: error as in using RemoveAll at end manually instead of defer?

pkg/gofr/config/godotenv.go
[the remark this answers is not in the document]
ccoVeille: Yes, the code is clearer then.

Also, this pattern is already used by you in the same PR

pkg/gofr/config/godotenv.go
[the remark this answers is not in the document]
ccoVeille: 👍

Thanks for confirming you did it on purpose

pkg/gofr/config/godotenv_test.go
ccoVeille: Yes, somehow

If the code if refactored later,  and a return is added, then the configs folder and its env file won't be cleaned.

Also, it's a common pattern to create temporary file then to immediately plan their removal, either with a defer, or t.Cleanup when in file

pkg/gofr/config/godotenv_test.go
ccoVeille: Please note this could be addressed in another PR by creating a new backlog issue

ccoVeille: I created

- https://github.com/gofr-dev/gofr/issues/857

aryanmehrotra (approved):

vipul-rawat (approved):

Issue #851, opened 2024-07-17
Unsupported trace exporter


surajit-zs:
When the environment variable TRACER_URL is set but TRACE_EXPORTER I forget to add, the following error message is generated:
<img width="791" alt="image" src="https://github.com/user-attachments/assets/381964d0-f250-4a23-ab6d-55cea3fe77ec">

I would suggest that you add the error message to indicate that the TRACE_EXPORTER environment variable is missing.

Sianao:
I'm working on  it  [#856](https://github.com/gofr-dev/gofr/pull/856)

srijan-27:
@Sianao - Thanks for your contribution, but since this was already assigned so the work is in progress for the ticket.
You can refer branch: https://github.com/gofr-dev/gofr/tree/fix/tracer-config-validation

PR #856, opened 2024-07-19
feat: add  OpenTelemetry  trace_exporter

Sianao:
## Update TRACE_EXPORTER miss err and add OpenTelemetry Protocol (OTLP) trace_exporter

**Description:**

- Update TRACE_EXPORTER missing err log info
-  Refacter getExporter
-  Add otlp exporter 
-  Add more test code

**Checklist:**


pkg/gofr/gofr.go
```diff
@@ -3,6 +3,7 @@ package gofr
 import (
 	"context"
 	"fmt"
+
```
ccoVeille, line 6:
```suggestion
```

pkg/gofr/gofr.go
```diff
@@ -374,11 +351,50 @@ func (a *App) getExporter(name, host, port, url string) (sdktrace.SpanExporter,
...
+		if name == "" {
+			a.container.Errorf("TRACE_EXPORTER env is missing")
+		} else {
+			a.container.Errorf("unsupported trace exporter: %s", name)
+		}
 	}
 
 	return exporter, err
 }
+func (a *App) buildOtlp(url, host, port, exporter string) (sdktrace.SpanExporter, error) {
```
ccoVeille, line 363: OTLP is an acronym, and most people are not aware of what this acronym could mean

Add comments even if not exported, and respect acronym or Go initialism

```suggestion
// buildOTLP builds OpenTelemetry Protocol 
func (a *App) buildOTLP(url, host, port, exporter string) (sdktrace.SpanExporter, error) {
```

Or do that


```suggestion
func (a *App) buildOpenTelemetryProtocol(url, host, port, exporter string) (sdktrace.SpanExporter, error) {
```

pkg/gofr/gofr.go
```diff
@@ -335,36 +335,13 @@ func (a *App) getExporter(name, host, port, url string) (sdktrace.SpanExporter,
...
-		a.container.Logf("Exporting traces to jaeger at %s", url)
-
-		opts := []otlptracegrpc.Option{otlptracegrpc.WithInsecure(), otlptracegrpc.WithEndpoint(url)}
-
-		if authHeader != "" {
-			opts = append(opts, otlptracegrpc.WithHeaders(map[string]string{"Authorization": authHeader}))
-		}
-
-		exporter, err = otlptracegrpc.New(context.Background(), opts...)
+		return a.buildOtlp(url, host, port, strings.ToLower(name))
```
ccoVeille, line 342: Here you have to add a comment to explain why jaeger uses open telemetry one, because it looks like a bug.

Many have no idea if/how they are related

pkg/gofr/gofr.go
```diff
@@ -374,11 +351,50 @@ func (a *App) getExporter(name, host, port, url string) (sdktrace.SpanExporter,
 
 		exporter = NewExporter(url, logging.NewLogger(logging.INFO))
 	default:
-		a.container.Errorf("unsupported trace exporter: %s", name)
+		if name == "" {
+			a.container.Errorf("TRACE_EXPORTER env is missing")
+		} else {
+			a.container.Errorf("unsupported trace exporter: %s", name)
```
ccoVeille, line 357:
```suggestion
			a.container.Errorf("unsupported TRACE_EXORTER: %s", name)
```

It would be logic because of the change you are doing when empty

pkg/gofr/gofr.go
```diff
@@ -374,11 +351,50 @@ func (a *App) getExporter(name, host, port, url string) (sdktrace.SpanExporter,
...
+
+	opts := []otlptracegrpc.Option{otlptracegrpc.WithInsecure(), otlptracegrpc.WithEndpoint(url)}
+
+	authHeader := a.Config.Get("TRACER_AUTH_KEY")
+
+	if authHeader != "" {
+		opts = append(opts, otlptracegrpc.WithHeaders(map[string]string{"Authorization": authHeader}))
+	}
+
+	return otlptracegrpc.New(context.Background(), opts...)
```
ccoVeille, line 378: I would expect the current `buildOtlp` (whatever its future would be)  to receive the context as argument.

It's always a bad thing when a call to `context.Background()` is made in the depth of the stack.

I know, you are refactoring by moving code, but here you moved the context creation to a lower level, it should be avoided

pkg/gofr/gofr_test.go
```diff
@@ -386,11 +432,9 @@ func Test_initTracer(t *testing.T) {
 
 			a.initTracer()
 		})
-
 		assert.Contains(t, logMessage, tc.expectedLogMessage, "TEST[%d], Failed.\n%s", i, tc.desc)
 	}
 }
-
```
ccoVeille, line 393: Don't suppress lines between methods, there is no reason to glue them

ccoVeille (changes requested):

pkg/gofr/gofr.go
```diff
@@ -335,36 +335,13 @@ func (a *App) getExporter(name, host, port, url string) (sdktrace.SpanExporter,
 		err      error
 	)
 
-	authHeader := a.Config.Get("TRACER_AUTH_KEY")
-
 	switch strings.ToLower(name) {
+	case "otlp":
+		return a.buildOtlp(url, host, port, strings.ToLower(name))
```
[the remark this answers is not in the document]
Sianao: No,  otlp is not the same as jaeger. it  just said that jaeger is an implementation of otlp.  For ops, using otlp protocol can have more processing methods for trace. There are not only jaeger and zipkin for otlp. Data can be exported to es and other places   
image
image

pkg/gofr/gofr.go
```diff
@@ -374,11 +351,50 @@ func (a *App) getExporter(name, host, port, url string) (sdktrace.SpanExporter,
 
 		exporter = NewExporter(url, logging.NewLogger(logging.INFO))
 	default:
-		a.container.Errorf("unsupported trace exporter: %s", name)
+		if name == "" {
+			a.container.Errorf("TRACE_EXPORTER env is missing")
+		} else {
+			a.container.Errorf("unsupported trace exporter: %s", name)
+		}
```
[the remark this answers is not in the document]
Sianao: i will do it

Umang01-hash:
@Sianao it appears there are some merge conflicts in your PR. Kindly resolve them.

pkg/gofr/gofr_test.go
```diff
@@ -361,6 +361,19 @@ func Test_initTracer(t *testing.T) {
 		"TRACER_URL":     "https://tracer-service.dev",
 	})
 
+	mockConfig8 := config.NewMockConfig(map[string]string{
```
Umang01-hash, line 364: Instead of adding more configs like these, we can have a helper function:

```go
createMockConfig := func(traceExporter, host, port, url, authKey string) config.Config {
        return config.NewMockConfig(map[string]string{
            "TRACE_EXPORTER": traceExporter,
            "TRACER_HOST":    host,
            "TRACER_PORT":    port,
            "TRACER_URL":     url,
            "TRACER_AUTH_KEY": authKey,
        })
    } 
```    
    
and then call it like this: **createMockConfig("otlp", "localhost", "2005", "", "valid-token")**

This avoids unnecessary code duplication

pkg/gofr/gofr_test.go
```diff
@@ -386,30 +401,55 @@ func Test_initTracer(t *testing.T) {
...
 			a.initTracer()
 		})
-
 		assert.Contains(t, logMessage, tc.expectedLogMessage, "TEST[%d], Failed.\n%s", i, tc.desc)
 	}
 }
 
 func Test_initTracer_invalidConfig(t *testing.T) {
-	mockConfig := config.NewMockConfig(map[string]string{
+	mockConfig1 := config.NewMockConfig(map[string]string{
```
Umang01-hash, line 409: Same here we can utilize that `createMockConfig ` function here too

pkg/gofr/gofr.go
```diff
@@ -317,54 +317,36 @@ func (a *App) initTracer() {
...
-	if (traceExporter != "" && tracerHost != "") || tracerURL != "" || traceExporter == gofrTraceExporter {
-		exporter, err := a.getExporter(traceExporter, tracerHost, tracerPort, tracerURL)
-
-		if err != nil {
-			a.container.Error(err)
-		}
-
-		batcher := sdktrace.NewBatchSpanProcessor(exporter)
-		tp.RegisterSpanProcessor(batcher)
+	exporter, err := a.getExporter(traceExporter, tracerHost, tracerPort, tracerURL)
```
Umang01-hash, line 320: Resolve merge conflicts regarding issue #851.

Umang01-hash (changes requested):

pkg/gofr/gofr.go
```diff
@@ -310,86 +309,97 @@ func (a *App) initTracer() {
...
-
-	//nolint:revive // early-return is not possible here, as below is the intentional logging flow
-	if tracerURL == "" && traceExporter != "" && !strings.EqualFold(traceExporter, "gofr") {
-		if tracerHost != "" && tracerPort != "" {
-			a.Logger().Warn("TRACER_HOST and TRACER_PORT are deprecated, use TRACER_URL instead")
-		} else {
-			a.Logger().Error("missing TRACER_URL config, should be provided with TRACE_EXPORTER to enable tracing")
-			return
-		}
+	if tracerURL == "" && tracerHost != "" && tracerPort != "" {
```
ccoVeille, line 317: Shouldn't be this?

```suggestion
	if tracerURL == "" && (tracerHost != "" || tracerPort != "") {
```

pkg/gofr/gofr.go
```diff
@@ -310,86 +309,97 @@ func (a *App) initTracer() {
 	traceExporter := a.Config.Get("TRACE_EXPORTER")
 	tracerURL := a.Config.Get("TRACER_URL")
 
-	// deprecated : tracer_host and tracer_port is deprecated and will be removed in upcoming versions
+	// deprecated : tracer_host and tracer_port is deprecated and will be removed in upcoming versions.
 	tracerHost := a.Config.Get("TRACER_HOST")
 	tracerPort := a.Config.GetOrDefault("TRACER_PORT", "9411")
+	authHeader := a.Config.Get("TRACER_AUTH_KEY")
```
ccoVeille, line 315: TRACER_AUTH_KEY is not deprecated, it should be clearer, by inverting and separating

```suggestion
	authHeader := a.Config.Get("TRACER_AUTH_KEY")

	// deprecated : tracer_host and tracer_port are deprecated and will be removed in upcoming versions.
 	tracerHost := a.Config.Get("TRACER_HOST")
 	tracerPort := a.Config.GetOrDefault("TRACER_PORT", "9411")
```

pkg/gofr/gofr_test.go
```diff
@@ -321,41 +321,24 @@ func Test_AddRESTHandlers(t *testing.T) {
...
+		})
+	}
+	mockConfig1 := createMockConfig("zipkin", "localhost", "2005", "", "")
+	mockConfig2 := createMockConfig("jaeger", "localhost", "2005", "", "")
+	mockConfig3 := createMockConfig("gofr", "", "", "", "")
+	mockConfig4 := createMockConfig("zipkin", "localhost", "2005", "", "valid-token")
+	mockConfig5 := createMockConfig("jaeger", "localhost", "2005", "", "valid-token")
+	mockConfig6 := createMockConfig("zipkin", "", "", "https://tracer-service.dev", "")
+	mockConfig7 := createMockConfig("jaeger", "", "", "https://tracer-service.dev", "")
+	mockConfig8 := createMockConfig("otlp", "localhost", "2005", "", "")
```
ccoVeille, line 340: I'm unsure why old host and port are still present in the test.

I would have removed them completely

pkg/gofr/gofr_test.go
```diff
@@ -382,29 +367,51 @@ func Test_initTracer(t *testing.T) {
...
 func Test_initTracer_invalidConfig(t *testing.T) {
-	mockConfig := config.NewMockConfig(map[string]string{
-		"TRACE_EXPORTER": "abc",
-		"TRACER_URL":     "localhost:2005",
-	})
+	createMockConfig := func(traceExporter, host, port, url, authKey string) config.Config {
+		return config.NewMockConfig(map[string]string{
+			"TRACE_EXPORTER":  traceExporter,
+			"TRACER_HOST":     host,
+			"TRACER_PORT":     port,
```
ccoVeille, line 379: Same question

pkg/gofr/gofr.go
ccoVeille: Right, I would have removed it as we no longer need the default value

pkg/gofr/gofr_test.go
Sianao: I'm sorry, i forget it

pkg/gofr/gofr.go
Sianao: I can do this easily. Do you need me to do it?

ccoVeille: My point was to either remove the default value,vor the test on a field that have a default value, so that cannot be empty

ccoVeille: > I can do this easily. Do you need me to do it?

I only have a phone, so yes, I would o me if you could

Sianao: All done 🌞

Sianao: All done 🌞

pkg/gofr/gofr.go
```diff
@@ -34,6 +34,9 @@ import (
 const (
 	defaultPublicStaticDir = "static"
 	gofrTraceExporter      = "gofr"
+	zipkinTraceExporter    = "zipkin"
+	jaegerTraceExporter    = "jaeger"
+	otlpTraceExporter      = "otlp"
```
ccoVeille, line 39:
```suggestion
 	traceExporterGoFr      = "gofr"
	traceExporterZipkin    = "zipkin"
	traceExporterJaeger    = "jaeger"
	traceExporterOTLP      = "otlp"
```

Usually, constant are sharing a common prefix. It helps for IDE completion

pkg/gofr/gofr.go
```diff
@@ -309,87 +311,101 @@ func (a *App) initTracer() {
...
+	batcher := sdktrace.NewBatchSpanProcessor(exporter)
+	tp.RegisterSpanProcessor(batcher)
 }
 
-func (a *App) getExporter(name, host, port, url string) (sdktrace.SpanExporter, error) {
+func (a *App) getExporter(name, host, port, url, authHeader string) (sdktrace.SpanExporter, error) {
 	var (
 		exporter sdktrace.SpanExporter
-		err      error
+		ctx      context.Context
```
ccoVeille, line 336: I'm a bit worried this context is defined by it's type, but never gets a value, so right now it's equal to nil unless I'm wrong

But this one is used by open telemetry

So maybe instead of setting here to context.Background (one solution)

The ctx could be passed as an argument of getExporter

Sianao: This was finally called at

```go
// Start establishes a gRPC connection to the collector.
func (c *client) Start(context.Context) error {
	if c.conn == nil {
		// If the caller did not provide a ClientConn when the client was
		// created, create one using the configuration they did provide.
		conn, err := grpc.NewClient(c.endpoint, c.dialOpts...)
		if err != nil {
			return err
		}
		// Keep track that we own the lifecycle of this conn and need to close
		// it on Shutdown.
		c.ourConn = true
		c.conn = conn
	}

	// The otlptrace.Client interface states this method is called just once,
	// so no need to check if already started.
	c.tscMu.Lock()
	c.tsc = coltracepb.NewTraceServiceClient(c.conn)
	c.tscMu.Unlock()

	return nil
}
```

Sianao: I  add the context.Background() as the argument of getExporter

pkg/gofr/gofr.go
```diff
@@ -309,87 +311,102 @@ func (a *App) initTracer() {
 
 	traceExporter := a.Config.Get("TRACE_EXPORTER")
 	tracerURL := a.Config.Get("TRACER_URL")
+	authHeader := a.Config.Get("TRACER_AUTH_KEY")
 
-	// deprecated : tracer_host and tracer_port is deprecated and will be removed in upcoming versions
+	// deprecated : tracer_host and tracer_port are deprecated and will be removed in upcoming versions.
 	tracerHost := a.Config.Get("TRACER_HOST")
-	tracerPort := a.Config.GetOrDefault("TRACER_PORT", "9411")
+	tracerPort := a.Config.Get("TRACER_PORT")
```
srijan-27, line 318: why have we removed the default value?

srijan-27:
@Sianao - while testing the changes I found some unexpected logging behaviours, pls fix them:
- If both `TRACER_EXPORTER` and `TRACER_URL` is missing then there is a log:
`missing TRACE_EXPORTER config, should be provided with TRACER_URL to enable tracing` which should not be there.
- If anyone of `TRACER_HOST` or `TRACER_PORT` is missing then we are not giving any error, which is not correct.

pkg/gofr/gofr.go
ccoVeille: These env variables are no longer used. 

The idea is to report them as deprecated.

The default value implies we don't know if the port env variable was present or not.

Removing it allow to detect in tracer_host or tracer_port are present and ask user to remove them

ccoVeille:
> If anyone of TRACER_HOST or TRACER_PORT is missing then we are not giving any error, which is not correct.

I'm surprised because the code is this

        if tracerURL == "" && (tracerHost != "" || tracerPort != "") {

srijan-27:
> > If anyone of TRACER_HOST or TRACER_PORT is missing then we are not giving any error, which is not correct.
> 
> I'm surprised because the code is this
> 
> ```
>     if tracerURL == "" && (tracerHost != "" || tracerPort != "") {
> ```

These are the configs given by me:
Screenshot 2024-07-23 at 1 53 53 PM

These are the logs:
Screenshot 2024-07-23 at 1 53 18 PM

@ccoVeille - Have you tried running the same situation?

pkg/gofr/gofr.go
srijan-27: But wouldn't that be a breaking change?

srijan-27: We are already logging a warning in case user is using host and port instead of url..
I believe that is reporting them

ccoVeille:
No, I'm in holidays without a computer. I can only review with my phone

ccoVeille:
> These are the configs given by me:\n\n\nThese are the logs:\n

But wait, this is normal, you are not using TRACE_URL (the new format) so you have an error message

srijan-27:
> > These are the configs given by me:\n\n\nThese are the logs:\n
> 
> But wait, this is normal, you are not using TRACE_URL (the new format) so you have an error message

Correct, I am not using `TRACER_URL`.. but still there is no error log.. please take a look at logs again, you will see that there is just a warn log to report deprecated configs.. but the exporter gets initialised without any port.. hence, the request failed logs later

ccoVeille:
Ah indeed, I missed that.

Maybe @Sianao could find the root cause and fix it

pkg/gofr/gofr.go
Sianao: the default value is for  Zipkin's port , we add otel collecter and  jaeger . the latter using 4317 as the default port , so i think remove it should be better .

Sianao:
> Ah indeed, I missed that.
> 
> Maybe @Sianao could find the root cause and fix it
the previous code only consider both host and port null , I change it  to ` || `
image

ccoVeille (approved):

examples/using-cron-jobs/configs/.env
```diff
@@ -1,3 +1,5 @@
 APP_NAME=sample-cron
 
-TRACE_EXPORTER=zipkin
\ No newline at end of file
+TRACE_EXPORTER=zipkin
+TRACER_HOST=127.0.0.1
+TRACER_PORT=9411
```
srijan-27, line 5: why are we adding host and port instead of url?

pkg/gofr/gofr_test.go
```diff
@@ -321,54 +321,45 @@ func Test_AddRESTHandlers(t *testing.T) {
...
+		{"jaeger exporter", mockConfig2, "Exporting traces to jaeger at localhost:2005"},
+		{"gofr exporter", mockConfig3, "Exporting traces to GoFr at https://tracer.gofr.dev"},
 		{"zipkin exporter with auth", mockConfig4, "Exporting traces to zipkin at http://localhost:2005/api/v2/spans"},
 		{"zipkin exporter with custom tracer url", mockConfig6, "Exporting traces to zipkin at https://tracer-service.dev"},
-		{"jaeger exporter", mockConfig2, "Exporting traces to jaeger at localhost:2005"},
 		{"jaeger exporter with auth", mockConfig5, "Exporting traces to jaeger at localhost:2005"},
 		{"jaeger exporter custom tracer url", mockConfig7, "Exporting traces to jaeger at https://tracer-service.dev"},
-		{"gofr exporter", mockConfig3, "Exporting traces to GoFr at https://tracer.gofr.dev"},
+		{"jaeger exporter custom tracer url", mockConfig8, "Exporting traces to otlp at localhost:4317"},
+		{"jaeger exporter custom tracer url", mockConfig9, "Exporting traces to otlp at localhost:4317"},
```
srijan-27, line 362: both the tests have same description, pls update this

pkg/gofr/gofr_test.go
```diff
@@ -382,29 +373,47 @@ func Test_initTracer(t *testing.T) {
 
 			a.initTracer()
 		})
-
```
srijan-27, line 385: unnecessary removal of newline.. decreases readability

pkg/gofr/gofr_test.go
```diff
@@ -382,29 +373,47 @@ func Test_initTracer(t *testing.T) {
...
-	errLogMessage := testutil.StderrOutputForFunc(func() {
-		mockContainer, _ := container.NewMockContainer(t)
+	testErr := []struct {
+		desc               string
+		config             config.Config
+		expectedLogMessage string
+	}{
+		{"unsupported trace exporter", mockConfig1, "unsupported TRACE_EXPORTER: abc"},
+		{"missing exporter", mockConfig2, "missing TRACE_EXPORTER config, should be provided with TRACER_URL to enable tracing"},
+		{"set exporter but not provide address", mockConfig3,
```
srijan-27, line 399: maybe we can update the description to missing tracer url

srijan-27 (changes requested):

pkg/gofr/gofr.go
srijan-27: this would still be a breaking change. right?

srijan-27:
@Sianao - why are we removing go.mod files for examples?

pkg/gofr/gofr.go
Sianao: yeah ,  if some one using zipkin but not set  with a port

examples/using-cron-jobs/configs/.env
Sianao: the former only add a host without a port , so i add it  or we can replace to url

Sianao:
> @Sianao - why are we removing go.mod files for examples? @ccoVeille - since you approved, so just wanted to check with you also..

@srijan-27  the example useing with a gofr version
```go
module github.com/gofr-dev/gofr/examples/using-cron-jobs

go 1.22

replace gofr.dev v1.5.0 => ../../../gofr

require (
	github.com/stretchr/testify v1.9.0
	gofr.dev v1.13.0
)

```
but gofr release v1.15.0 

run with example lead 
```bash
gofr/examples/using-cron-jobs on  development [+] via 🐹 v1.22.5 
❯ go run main.go 
# gofr.dev/pkg/gofr/container
/home/sia/go/pkg/mod/gofr.dev@v1.13.0/pkg/gofr/container/mock_container.go:36:20: cannot use redisMock (variable of type *MockRedis) as Redis value in assignment: *MockRedis does not implement Redis (missing method BitFieldRO)
```

pkg/gofr/gofr.go
srijan-27: correct, and I believe we should not break something for someone using GoFr.
So, can we pls revert?

examples/using-cron-jobs/configs/.env
srijan-27: yes, we can add URL instead.. since it is new format

pkg/gofr/gofr.go
Sianao: either remove or add i have no problem

srijan-27:
@Sianao - yes, I got the error with versioning of GoFr and hence the issue in running example.. but this should not be fixed by removing the go.mod files :) and even if we are removing we should remove completely (you missed go.sum file in using-web-socket example)
Also, the fix for this issue should be included in a separate PR since that is unrelated to tracer changes.

I have done the changes to fix this.. maybe I can create a PR or let me know if you would like to work on it #868

ccoVeille:
> why are we removing go.mod files for examples?\n@ccoVeille - since you approved, so just wanted to check with you also..

I saw it. But as I had no experience or point of view I assumed someone else would point it out. Then I forgot it, as I had already marked the files as read.

Sianao:
> @Sianao - yes, I got the error with versioning of GoFr and hence the issue in running example.. but this should not be fixed by removing the go.mod files :) and even if we are removing we should remove completely (you missed go.sum file in using-web-socket example) Also, the fix for this issue should be included in a separate PR since that is unrelated to tracer changes.
> 
> I have done the changes to fix this.. maybe I can create a PR or let me know if you would like to work on it #868


This pr contain to many things , I hope you can do it .

pkg/gofr/gofr.go
```diff
@@ -309,87 +311,102 @@ func (a *App) initTracer() {
...
 
-	// deprecated : tracer_host and tracer_port is deprecated and will be removed in upcoming versions
+	// deprecated : tracer_host and tracer_port are deprecated and will be removed in upcoming versions.
 	tracerHost := a.Config.Get("TRACER_HOST")
 	tracerPort := a.Config.GetOrDefault("TRACER_PORT", "9411")
 
-	if tracerURL != "" && traceExporter == "" {
-		a.Logger().Error("missing TRACE_EXPORTER config, should be provided with TRACER_URL to enable tracing")
-		return
+	if tracerURL == "" && (tracerHost != "" || tracerPort != "") {
```
ccoVeille, line 320: I'm unsure why you restored tge default value for the port, but I'm fine with it.

But then, the tracer port is always non empty.

So you would display an error message if tracerURL is empty. So here you can face an error message even if tracer is not used at all.

So please use this

```suggestion
	if tracerURL == "" && tracerHost != "" {
```

pkg/gofr/gofr.go
```diff
@@ -309,87 +311,102 @@ func (a *App) initTracer() {
...
 
-	//nolint:revive // early-return is not possible here, as below is the intentional logging flow
-	if tracerURL == "" && traceExporter != "" && !strings.EqualFold(traceExporter, "gofr") {
-		if tracerHost != "" && tracerPort != "" {
-			a.Logger().Warn("TRACER_HOST and TRACER_PORT are deprecated, use TRACER_URL instead")
-		} else {
-			a.Logger().Error("missing TRACER_URL config, should be provided with TRACE_EXPORTER to enable tracing")
-			return
-		}
+	exporter, err := a.getExporter(context.Background(), traceExporter, tracerHost, tracerPort, tracerURL, authHeader)
```
ccoVeille, line 324: What about this?

```suggestion
	exporter, err := a.getExporter(context.Background(), traceExporter, tracerURL, authHeader)
```

Plus the two next suggestions

pkg/gofr/gofr.go
```diff
@@ -309,87 +311,102 @@ func (a *App) initTracer() {
...
-	authHeader := a.Config.Get("TRACER_AUTH_KEY")
+	if name == "" {
+		a.Logger().Errorf("missing TRACE_EXPORTER config, should be provided with TRACER_URL to enable tracing")
+		return exporter, nil
+	}
+
+	if (host == "" || port == "") && url == "" && name != traceExporterGoFr {
+		a.Logger().Errorf("missing TRACER_URL config, should be provided with TRACE_EXPORTER to enable tracing")
+		return exporter, nil
+	}
```
ccoVeille, line 348:
```suggestion
	if url == "" && name != traceExporterGoFr {
		a.Logger().Errorf("missing TRACER_URL config, should be provided with TRACE_EXPORTER to enable tracing")
		return exporter, nil
	}
```

pkg/gofr/gofr.go
```diff
@@ -309,87 +311,102 @@ func (a *App) initTracer() {
...
+	batcher := sdktrace.NewBatchSpanProcessor(exporter)
+	tp.RegisterSpanProcessor(batcher)
 }
 
-func (a *App) getExporter(name, host, port, url string) (sdktrace.SpanExporter, error) {
-	var (
-		exporter sdktrace.SpanExporter
-		err      error
-	)
+func (a *App) getExporter(ctx context.Context, name, host, port, url, authHeader string) (sdktrace.SpanExporter, error) {
```
ccoVeille, line 337:
```suggestion
func (a *App) getExporter(ctx context.Context, name, url, authHeader string) (sdktrace.SpanExporter, error) {
```

pkg/gofr/gofr.go
Sianao: @srijan-27  think this is a break change  so i revert it

ccoVeille (approved):
LGTM 🤞

Umang01-hash (dismissed):

pkg/gofr/gofr.go
```diff
@@ -33,7 +33,10 @@ import (
 
 const (
 	defaultPublicStaticDir = "static"
-	gofrTraceExporter      = "gofr"
+	traceExporterGoFr      = "gofr"
+	traceExporterZipkin    = "zipkin"
+	traceExporterJaeger    = "jaeger"
+	traceExporterOTLP      = "otlp"
```
srijan-27, line 39: @Sianao - if we are adding support for a new exporter name `otlp`, this should also be updated in docs

Sianao: All done

Umang01-hash (dismissed):

pkg/gofr/gofr.go
```diff
@@ -309,87 +311,102 @@ func (a *App) initTracer() {
...
 
-	if (traceExporter != "" && tracerHost != "") || tracerURL != "" || traceExporter == gofrTraceExporter {
-		exporter, err := a.getExporter(traceExporter, tracerHost, tracerPort, tracerURL)
+	if tracerURL == "" && tracerHost != "" {
+		a.Logger().Warn("TRACER_HOST and TRACER_PORT are deprecated, use TRACER_URL instead")
 
-		if err != nil {
-			a.container.Error(err)
-		}
+		tracerURL = fmt.Sprintf("%s:%s", tracerHost, tracerPort)
```
srijan-27, line 334: URL should not be formed here.. as this URL is only specific for otlp.. we are updating the URL again for zipkin and gofr, lines: 384 and 400 respectively.
I believe this needs to be fixed

srijan-27:
@Sianao - While testing the changes, I found some unexpected behaviours that were not happening before this PR, pls check:
- If `tracer_exporter=gofr`, and tracer_host and tracer_port are given, then:
Screenshot 2024-07-25 at 11 35 39 AM
Earlier these cases were handled and traces were successfully sent to tracer.gofr.dev


- If no configs are given, then:
Screenshot 2024-07-25 at 11 44 38 AM
This should not be the case, as tracing is optional and we should only try to enable it when configs are provided.

pkg/gofr/gofr.go
Sianao: host and port will be removed in later versions. For GofrTraceExporter, in my  latter commit  changes, gofr will only take effect on trace_url set or default , and the changes here to Zipkin will have no impact and even support  zipkin for url env set no only host and port .

docs/quick-start/observability/page.md
```diff
@@ -222,10 +222,23 @@ TRACE_EXPORTER=jaeger
...
+```dotenv
+# ... no change in other env variables
+
+# tracing configs 
+TRACE_EXPORTER=otlp
+TRACER_URL=localhost:4317
+```
+
 Open {% new-tab-link title="zipkin" href="http://localhost:16686/trace/" /%} and search by TraceID (correlationID) to see the trace.
 {% figure src="/jaeger-traces.png" alt="Jaeger traces" /%}
```
srijan-27, line 239: these lines are reference for `jaeger`, shouldn't they be before otlp point?

pkg/gofr/gofr.go
srijan-27: @Sianao - can we connect on discord? I have shared some idea, just wanted to check with you.

docs/quick-start/observability/page.md
```diff
@@ -225,7 +225,22 @@ TRACER_URL=localhost:14317
 Open {% new-tab-link title="zipkin" href="http://localhost:16686/trace/" /%} and search by TraceID (correlationID) to see the trace.
```
ccoVeille, line 225: Notre related to this PR, but I think it should be this

```suggestion
Open {% new-tab-link title="jaeger" href="http://localhost:16686/trace/" /%} and search by TraceID (correlationID) to see the trace.
```

pkg/gofr/gofr.go
```diff
@@ -309,87 +311,108 @@ func (a *App) initTracer() {
...
 	tracerHost := a.Config.Get("TRACER_HOST")
 	tracerPort := a.Config.GetOrDefault("TRACER_PORT", "9411")
 
-	if tracerURL != "" && traceExporter == "" {
-		a.Logger().Error("missing TRACE_EXPORTER config, should be provided with TRACER_URL to enable tracing")
-		return
+	if tracerURL == "" && tracerHost != "" && traceExporter != traceExporterGoFr {
+		a.Logger().Warn("TRACER_HOST and TRACER_PORT are deprecated, use TRACER_URL instead")
+
+		tracerURL = fmt.Sprintf("%s:%s", tracerHost, tracerPort)
```
ccoVeille, line 323: Nothing to add as an URI?

ccoVeille (approved):

pkg/gofr/gofr.go
Sianao: yeah this is for zipkin  the main brach only recive host and port build a http url to export to zipkin  ,  this enable tracer_url with https and also keep with host and port

Sianao: maybe call endpoint could be better or not

srijan-27 (approved):

Issue #765, opened 2024-06-26
Support for Batch in Cassandra


aryanmehrotra:
Currently, we don't support batch request processing in Cassandra.

To support migrations for Cassandra we would need add support for the following :

1. NewBatch()
2. ExecuteBatch()
3. ExecuteBatchCAS()

PR #860, opened 2024-07-22
Support for batch in cassandra

ssshekhu53:
- Added `NewBatch(batchType int) error`, `BatchQuery(stmt string, values ...interface{})` and `ExecuteBatch() error` methods in Cassandra interface to to create batch, add queries in batch and execute batch respectively.
- Added implementation for same in the driver
- Added and refactored tests
- Updated documentation
- Issue #765


pkg/gofr/datasource/cassandra/cassandra.go
```diff
@@ -177,6 +176,40 @@ func (c *Client) ExecCAS(dest interface{}, stmt string, values ...interface{}) (
 	return applied, err
 }
 
+func (c *Client) NewBatch(batchType int) error {
+	const (
+		loggedBatch = iota
+		unloggedBatch
+		counterBatch
+	)
```
aryanmehrotra, line 184: can we mention in the comment on why we have not created Exported Enum type

pkg/gofr/datasource/cassandra/cassandra.go
```diff
@@ -177,6 +176,40 @@ func (c *Client) ExecCAS(dest interface{}, stmt string, values ...interface{}) (
...
+
```
aryanmehrotra, line 185: can we mention in the comment on why we have not created Exported Enum type

pkg/gofr/datasource/constants.go
```diff
@@ -0,0 +1,8 @@
+package datasource
+
+// Cassandra batch types.
+const (
+	CassandraLoggedBatch = iota
+	CassandraUnloggedBatch
+	CassandraCounterBatch
+)
```
aryanmehrotra, line 8: Rather than creating a separate file, we can put it in cassandra.go file, where it is being used.
Moreover, these can be used in cassandra.go file where we have created similar constants for same purpose.

aryanmehrotra (changes requested):
1. With Respect to issue ticket ExecuteBatchCAS() is missing.

pkg/gofr/datasource/cassandra/cassandra.go
```diff
@@ -177,6 +176,40 @@ func (c *Client) ExecCAS(dest interface{}, stmt string, values ...interface{}) (
...
+		c.cassandra.batch = c.cassandra.session.newBatch(gocql.BatchType(batchType))
+
+	default:
+		err = errUnsupportedBatchType
+	}
+
+	return err
+}
+
+func (c *Client) BatchQuery(stmt string, values ...interface{}) {
```
ccoVeille, line 199:
```suggestion
func (c *Client) BatchQuery(stmt string, values ...any) {
```

pkg/gofr/datasource/constants.go
ccoVeille: I agree

pkg/gofr/datasource/cassandra/cassandra.go
```diff
@@ -177,6 +176,40 @@ func (c *Client) ExecCAS(dest interface{}, stmt string, values ...interface{}) (
...
+	c.cassandra.batch.Query(stmt, values...)
+}
+
+func (c *Client) ExecuteBatch() error {
+	defer c.postProcess(&QueryLog{Query: "", Keyspace: c.config.Keyspace}, time.Now())
```
ccoVeille, line 204: Is there a need to define query to its default value?

```suggestion
	defer c.postProcess(&QueryLog{Keyspace: c.config.Keyspace}, time.Now())
```

Except if you enforcing via exhaust struct linter or equivalent, I don't see why.

pkg/gofr/datasource/cassandra/errors.go
```diff
@@ -1,17 +1,23 @@
...
-	return msgDestinationIsNotPointer
-}
+var (
+	errDestinationIsNotPointer = errors.New(msgDestinationIsNotPointer)
+	errUnexpectedMap           = errors.New(msgUnexpectedMap)
+	errUnsupportedBatchType    = errors.New(msgUnsupportedBatchType)
+	errBatchNotInitialised     = errors.New(msgBatchNotInitialised)
+)
 
 type unexpectedPointer struct {
```
ccoVeille, line 22: Same remark for this one and unexpectedSlice

I would expect them to be exported

pkg/gofr/datasource/cassandra/cassandra.go
ccoVeille: `any` was introduced to provide a simpler way to write `interface{}` in Go 1.18

You can provide some information here

https://stackoverflow.com/questions/71628061/difference-between-any-interface-as-constraint-vs-type-of-argument

I tend to see `interface{}` added to new code to be an archaism, even it's still valid.

That's why I suggested it.

But you can ignore

pkg/gofr/datasource/cassandra/cassandra_batch.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (c *Client) ExecuteBatchCAS(name string, dest ...any) (bool, error) {
+	defer c.postProcess(&QueryLog{Query: "batch", Keyspace: c.config.Keyspace}, time.Now())
+
+	if b, ok := c.cassandra.batches[name]; ok {
+		return c.cassandra.session.executeBatchCAS(b, dest...)
+	}
+
+	return false, ErrBatchNotInitialised
+}
```
ccoVeille, line 35: The canonical way to write it would be this I think

```suggestion
func (c *Client) BatchQuery(name, stmt string, values ...any) error {
	b, ok := c.cassandra.batches[name]
	if !ok {
		return ErrBatchNotInitialised
	}
	b.Query(stmt, values...)
	return nil
}

func (c *Client) ExecuteBatch(name string) error {
	defer c.postProcess(&QueryLog{Query: "batch", Keyspace: c.config.Keyspace}, time.Now())

	b, ok := c.cassandra.batches[name]
	if !ok {
		return ErrBatchNotInitialised
	}
	return c.cassandra.session.executeBatch(b)
}

func (c *Client) ExecuteBatchCAS(name string, dest ...any) (bool, error) {
	defer c.postProcess(&QueryLog{Query: "batch", Keyspace: c.config.Keyspace}, time.Now())

	b, ok := c.cassandra.batches[name]
	if !ok {
		return false, ErrBatchNotInitialised
	}
	return c.cassandra.session.executeBatchCAS(b, dest...)
}
```

pkg/gofr/datasource/cassandra/errors.go
```diff
@@ -1,36 +1,29 @@
...
-}
+var (
+	ErrDestinationIsNotPointer = errors.New("destination is not pointer")
+	ErrUnexpectedMap           = errors.New("a map was not expected")
+	ErrUnsupportedBatchType    = errors.New("batch type not supported")
+	ErrBatchNotInitialised     = errors.New("batch not initialized")
+)
 
-type unexpectedPointer struct {
+type UnexpectedPointer struct {
```
ccoVeille, line 15:
```suggestion
type ErrUnexpectedPointer struct {
```

pkg/gofr/datasource/cassandra/internal.go
```diff
@@ -0,0 +1,157 @@
...
+type cassandraSession struct {
+	session *gocql.Session
+}
+
+// query creates a Cassandra query.
+// This method wraps the `Query` method of the underlying `session` object.
+func (c *cassandraSession) query(stmt string, values ...any) query {
+	q := &cassandraQuery{query: c.session.Query(stmt, values...)}
+
+	return q
```
ccoVeille, line 108:
```suggestion
	return &cassandraQuery{query: c.session.Query(stmt, values...)}
```

pkg/gofr/datasource/cassandra/cassandra.go
```diff
@@ -125,20 +130,20 @@ func (c *Client) Query(dest interface{}, stmt string, values ...interface{}) err
 	default:
 		c.logger.Debugf("a pointer to %v was not expected.", rv.Kind().String())
 
-		return unexpectedPointer{target: rv.Kind().String()}
+		return ErrUnexpectedPointer{target: rv.Kind().String()}
```
ccoVeille, line 133: Now, the code is clearer to me. Thanks for the changes by the way.

I would like to suggest something that would make more sense according to me

Here the code relies (and already relied) on something that single purpose could be defined by these:
- we need to be able to catch this error
- we want to provide a meaningful error message that will contain the kind.

The current way is OK, but somehow overkill

Here is what I would code if I had to.

We don't define a struct, we define a simple sentinel error. This will ensure we can catch the error with a simple `errors.Is`

```go
var ErrUnexpectedPointer = errors.New("unexpected pointer")
```

Then we use `fmt.Errorf("%w to %s", ErrUnexpectedPointer, rv.Kind.String())` here

I know code was there, I know I'm asking to change something you are not responsible for. But, I report what I would have expected here.

As the errors are now exported, I would expect them to be clean.

Please let me know what you think about it

ccoVeille (changes requested):

pkg/gofr/datasource/cassandra/cassandra.go
ccoVeille: I think so, the other changes that could also needed is to rewrite the possible existing errors.As using this struct to an errors.Is

Anyway, I will let maintainers tell us what they thi k about my suggestion

@vipul-rawat @srijan-27 @Umang01-hash maybe?

srijan-27: Yes @ccoVeille - I agree with you, if the errors are now being exported then it needs to be handled neatly.
But I think, we can separate out these changes from this PR. Let's create a different issue for that, and follow up there.

ccoVeille: Of course, I raise my points, but I didn't expect all to be fixed in this PR

Could you open an issue about it?

ccoVeille: So maybe for now, in this PR we could avoid to change error variable to exported. This way they won't be available, then changed later, leading to breaking changes

srijan-27: @ssshekhu53 - Please revert the changes as required. Will take these up in separate PR.

ccoVeille: For records #893 has been opened

ccoVeille (approved):

aryanmehrotra (approved):

Umang01-hash (approved):

Issue #858, opened 2024-07-22
App panics during start for AddRESTHandlers


aryanmehrotra:
**Describe the bug**
When using the functionality `AddRESTHandlers` and passing the struct by value i.e
```
err := app.AddRESTHandlers(Level{})
```

application panics and gives the following logs:

```
INFO [12:12:39] Loaded config from file: ./configs/.env
INFO [12:12:39] connected to 'log-level' database at 'localhost:2001'
panic: reflect: Elem of invalid type main.Level

goroutine 1 [running]:
reflect.elem(0x140005cfc48?)
        /opt/homebrew/opt/go/libexec/src/reflect/type.go:712 +0xc4
reflect.(*rtype).Elem(0x28?)
        /opt/homebrew/opt/go/libexec/src/reflect/type.go:716 +0x20
gofr.dev/pkg/gofr.scanEntity({0x1014fa760, 0x140000c2420})
        /Users/raramuri/go/pkg/mod/gofr.dev@v1.14.1/pkg/gofr/crud_handlers.go:64 +0x58
gofr.dev/pkg/gofr.(*App).AddRESTHandlers(0x14000117440, {0x1014fa760, 0x140000c2420})
        /Users/raramuri/go/pkg/mod/gofr.dev@v1.14.1/pkg/gofr/gofr.go:457 +0x30
main.main()
        /Users/raramuri/Desktop/gofr-app/main.go:39 +0xd4
exit status 2
```

**Expected behavior**
It should not panic rather startup should fail with the error that, failed to register routes for \<struct-name>, as it is passed by value should be pointer

srinivas365:
https://github.com/gofr-dev/gofr/pull/864

srinivas365:
https://github.com/gofr-dev/gofr/pull/864

srijan-27:
@srinivas365 - Since, you are working on this so, assigning this to you

PR #864, opened 2024-07-22
fix app panics during start for AddRESTHandlers

srinivas365:
Closes: https://github.com/gofr-dev/gofr/issues/858


pkg/gofr/crud_handlers.go
```diff
@@ -61,7 +61,16 @@ type entity struct {
 
 // scanEntity extracts entity information for CRUD operations.
 func scanEntity(object interface{}) (*entity, error) {
-	entityType := reflect.TypeOf(object).Elem()
+	if object == nil {
+		return nil, errInvalidObject
```
Umang01-hash, line 65: Can we find some way to include the struct details ,
So thinking of a scenario where user has used method `AddRESTHandlers` for 2 different structs, one he has passed by value and other by pointer.

So when we will run the code, w err message : ` invalid object for AddRESTHandlers` will not give clear description of which object is invalid....

Umang01-hash (changes requested):

pkg/gofr/crud_handlers.go
```diff
@@ -61,7 +63,16 @@ type entity struct {
...
 // scanEntity extracts entity information for CRUD operations.
 func scanEntity(object interface{}) (*entity, error) {
-	entityType := reflect.TypeOf(object).Elem()
+	if object == nil {
+		return nil, errObjectIsNil
+	}
+
+	objType := reflect.TypeOf(object)
+	if objType.Kind() != reflect.Ptr {
+		return nil, errObjectNotPointer
```
Umang01-hash, line 72: Can't we use `objType.Name()` here for a specialized error message in **errObjectNotPointer**

A error message like `failed to register routes for user struct, as it is passed by value should be pointer`

will be more informative from user's point of view

Umang01-hash (changes requested):

pkg/gofr/crud_handlers.go
srinivas365: Updated the error message and included the struct name as well

pkg/gofr/crud_handlers.go
```diff
@@ -10,8 +10,10 @@ import (
 )
 
 var (
-	errInvalidObject  = errors.New("unexpected object given for AddRESTHandlers")
-	errEntityNotFound = errors.New("entity not found")
+	errInvalidObject    = errors.New("unexpected object given for AddRESTHandlers")
+	errEntityNotFound   = errors.New("entity not found")
+	errObjectNotPointer = errors.New("pointer object is not given for AddRESTHandlers")
```
srijan-27, line 15: this is unused now, since we have changed the error for not pointer case.

pkg/gofr/crud_handlers.go
```diff
@@ -61,7 +63,16 @@ type entity struct {
...
 // scanEntity extracts entity information for CRUD operations.
 func scanEntity(object interface{}) (*entity, error) {
-	entityType := reflect.TypeOf(object).Elem()
+	if object == nil {
+		return nil, errObjectIsNil
+	}
+
+	objType := reflect.TypeOf(object)
+	if objType.Kind() != reflect.Ptr {
+		return nil, fmt.Errorf("failed to register routes for %s struct, as it is passed by value should be pointer", objType.Name())
```
srijan-27, line 72: we can just mention that passed object is not pointer, no need to specify `passed by value`
also, we can present struct in quotes like: `'%s' struct`

srijan-27 (changes requested):

pkg/gofr/crud_handlers.go
srijan-27: @srinivas365 - The linter error is yet to be fixed.

pkg/gofr/crud_handlers.go
```diff
@@ -61,7 +63,16 @@ type entity struct {
...
 // scanEntity extracts entity information for CRUD operations.
 func scanEntity(object interface{}) (*entity, error) {
-	entityType := reflect.TypeOf(object).Elem()
+	if object == nil {
+		return nil, errObjectIsNil
+	}
+
+	objType := reflect.TypeOf(object)
+	if objType.Kind() != reflect.Ptr {
+		return nil, fmt.Errorf("failed to register routes for '%s' struct, %w", objType.Name(), errNonPointerObject)
```
puellanivis, line 72: Use the formatting verb `%q` not `'%s'`.

vipul-rawat (approved):
`

Umang01-hash (approved):

srijan-27 (approved):

Issue #868, opened 2024-07-23
Fix go.mod for examples


srijan-27:
While running the examples `using-cron-handler` and `using-web-socket` the following error occurs:
<img width="1364" alt="Screenshot 2024-07-23 at 4 16 26 PM" src="https://github.com/user-attachments/assets/5a84cf36-e5a9-4389-b1b1-0cc47753c8b6">

This is because of gofr version mismatch in require and replace statement.

srijan-27:
Since, there is no difference in build size even after making every example a separate module. Removing go.mod for all examples (wherever present)

PR #869, opened 2024-07-23
Fix go.mod for examples

srijan-27:
- Removed go.mod for all examples
- Moved migrations for pubsub to using pubsub examples from using-migrations
- Closes: #868


Sianao:
The other example without a go.mod and go.sum

srijan-27:
> The other example without a go.mod and go.sum

@Sianao - Yes, I am working to add go.mod for all examples.

Sianao:
> > The other example without a go.mod and go.sum
> 
> @Sianao - Yes, I am working to add go.mod for all examples.

But if you add it , when new version release , we should update all example ?

.github/workflows/go.yml
```diff
@@ -208,4 +208,4 @@ jobs:
       - name: Check for file names errors
         uses: ls-lint/action@v2.2.3
         with:
-          config: .ls-lint.yml
\ No newline at end of file
+          config: .ls-lint.yml
```
Umang01-hash, line 211: Can we revert this change. Seems unwanted..

Umang01-hash (changes requested):

Umang01-hash (approved):

vipul-rawat (approved):

Issue #505, opened 2024-04-18
Support FTP/SFTP files abstraction


vikash:

Sarthak027:
Hi @vikash I want to work on this. Can u please explain this issue to me??

aryanmehrotra:
Hi @Sarthak027 we will be merging the interface which will has to be implemented for SFTP and FTP similar to how we are supporting mongoDB.

coolwednesday:
Hi @aryanmehrotra . Please Assign this issue to me.

PR #871, opened 2024-07-23
Ftp Server Abstraction

coolwednesday:
**Description:**

-   This pull request introduces abstraction of ftp commands to implement the generalised file interface of gofr.
-   Resolves issue #505 
-   Abstraction of ftp  commands for read/write and other file operations.

**Breaking Changes :**

-   Different transfer protocols like TLS are not handled in the code yet, might become an issue while dealing with secure      communication and firewalls.

**Additional Information:**

-   external library used : github.com/jlaffaye/ftp
-   Breaking Changes might occur if there are any changes in the code of the above library.

**Checklist:**

-   [x] All new code is covered by unit tests. ( Few test do not have mocking and would require an actual connection to test. )
-   [x] This PR does not decrease the overall code coverage 


go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/gorilla/mux v1.8.1
 	github.com/gorilla/websocket v1.5.3
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
+	github.com/jlaffaye/ftp v0.2.0
```
aryanmehrotra, line 19: this import is not used in gofr, should be removed from here, as it is already present in ftp package

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+
+	"github.com/jlaffaye/ftp"
+)
+
+// ftpFileSystem represents a file system interface over FTP.
+type ftpFileSystem struct {
+	file   *ftpFile          // Pointer to a single file object (not used in the provided methods)
+	conn   ServerConn        // FTP server connection
+	config *Config           // FTP configuration
+	logger datasource.Logger // Logger interface for logging
```
aryanmehrotra, line 23: Logger interface should be defined in ftp package

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"gofr.dev/pkg/gofr/datasource"
+
+	"github.com/jlaffaye/ftp"
```
aryanmehrotra, line 15: sort imports - datasource import will be removed

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+	logger datasource.Logger // Logger interface for logging
+}
+
+// Config represents the FTP configuration.
+type Config struct {
+	Host      string // FTP server hostname
+	User      string // FTP username
+	Password  string // FTP password
+	Port      string // FTP port
+	RemoteDir string // Remote directory path
```
aryanmehrotra, line 32: This will also be the base path for all the operation, should be mentioned in the comments

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+	config *Config           // FTP configuration
+	logger datasource.Logger // Logger interface for logging
+}
+
+// Config represents the FTP configuration.
+type Config struct {
+	Host      string // FTP server hostname
+	User      string // FTP username
+	Password  string // FTP password
+	Port      string // FTP port
```
aryanmehrotra, line 31: port should be integer

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+	RemoteDir string // Remote directory path
+}
+
+// New initializes a new instance of ftpFileSystem with provided configuration.
+func New(config *Config) datasource.FileSystem {
```
aryanmehrotra, line 36: Config can be passed by value as it is not a heavy struct.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+	return &ftpFileSystem{config: config}
+}
+
+// UseLogger sets the logger interface for the FTP file system.
+func (f *ftpFileSystem) UseLogger(logger interface{}) {
+	f.logger = logger.(datasource.Logger)
```
aryanmehrotra, line 42: this might panic.
```suggestion
	if l, ok := logger.(datasource.Logger); ok {
		c.logger = l
	}
```

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
+package ftp
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"log"
```
aryanmehrotra, line 7: it is not required as we have a gofr's logger

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+func (f *ftpFileSystem) Connect() {
+	// Construct FTP server address
+	ftpServer := fmt.Sprintf("%v:%v", f.config.Host, f.config.Port)
+
+	// Connect to FTP server
+	const dialTimeoutSeconds = 5
+
+	conn, err := ftp.Dial(ftpServer, ftp.DialWithTimeout(dialTimeoutSeconds*time.Second))
+	if err != nil {
+		log.Printf("Failed to connect to FTP server: %v", err)
```
aryanmehrotra, line 61: gofr logger should be used

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+// Create creates an empty file on the FTP server.
+func (f *ftpFileSystem) Create(name string) (datasource.File, error) {
+	// empty io.Reader
+	emptyReader := new(bytes.Buffer)
+
+	if name == "" {
+		return nil, errors.New("empty filename")
+	}
```
aryanmehrotra, line 83: variable should be declared after mandatory checks

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+	return &ftpFile{
+		response: res,
+		name:     s[len(s)-1],
+		path:     name,
+		conn:     f.conn,
+	}, nil
+}
+
+// Mkdir creates a directory on the FTP server.
+func (f *ftpFileSystem) Mkdir(name string, _ os.FileMode) error {
```
aryanmehrotra, line 115: lets add a comment that os.FileMode will be unused.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+			continue
+		}
+
+		// Build the full path up to this directory
+		currentDir = filepath.Join(currentDir, dir)
+
+		// Attempt to create the directory
+		err := f.conn.MakeDir(currentDir)
+		if err != nil {
+			continue // Continue creating next directory on failure
```
aryanmehrotra, line 145: If I want to create the following directory 
"pkg/gofr/ftp"

and it fails while creating, I should not be creating either gofr or ftp.

Should return an error.

Moreover if pkg is created and it fails at creating gofr - we should be deleting pkg as well.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+	// Return the file handle
+	return &ftpFile{
+		response: res,
+		name:     s[len(s)-1],
+		path:     name,
+		conn:     f.conn,
+	}, nil
+}
+
+// OpenFile retrieves a file from the FTP server with specified flags and permissions, and returns a file handle.
```
aryanmehrotra, line 177: Comment is misleading as we are not account the flags or permissions

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+// permissions are not clear for Ftp as file commands do not accept an argument and don't store their file permissions.
+// currently, this function just calls the Open function.
+func (f *ftpFileSystem) OpenFile(name string, _ int, _ os.FileMode) (datasource.File, error) {
```
aryanmehrotra, line 180: lets Define the File interface in the ftp package itself.
datasource and ftp package will have their different declarations

As the complete point of keeping them separate is - keeping only the different features used in the build.
This increases the build size. and we have gofr imported in ftp, which should not be the case.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	path = fmt.Sprintf("%s/%s", f.config.RemoteDir, path)
+
+	return f.conn.RemoveDirRecur(path)
+}
+
+// Rename renames a file or directory on the FTP server.
+func (f *ftpFileSystem) Rename(oldname, newname string) error {
+	if oldname == "" || newname == "" {
```
aryanmehrotra, line 208: we can have condition if both of them are same, we don't need to rename

pkg/gofr/datasource/file/ftp/go.mod
```diff
@@ -0,0 +1,22 @@
+module gofr.dev/pkg/gofr/datasource/file/ftp
+
+go 1.22.4
+
+require (
+	github.com/jlaffaye/ftp v0.2.0
+	github.com/joho/godotenv v1.5.1
+	github.com/stretchr/testify v1.9.0
+	go.uber.org/mock v0.4.0
+	gofr.dev v1.15.0
```
aryanmehrotra, line 10: gofr need not to be imported

aryanmehrotra (changes requested):

pkg/gofr/datasource/file/ftp/file_test.go
```diff
@@ -0,0 +1,632 @@
...
+				t.Errorf("Expected error: %v, but got: %v", tt.expectError, err)
+			}
+
+			// Log successful read
+			if err == nil {
+				t.Logf("Read successfully from %s", tt.filePath)
+			}
+		})
+	}
+}
```
aryanmehrotra, line 95: assert/require package can be used.

pkg/gofr/datasource/file/ftp/file_test.go
```diff
@@ -0,0 +1,632 @@
...
+	config := Config{
+		Host:      os.Getenv("FtpHost"),
+		User:      os.Getenv("FtpUser"),
+		Password:  os.Getenv("FtpPassword"),
+		Port:      os.Getenv("FtpPort"),
+		RemoteDir: os.Getenv("FtpRemoteDir"),
+	}
+
+	return &config
+}
```
aryanmehrotra, line 613: this is not needed

pkg/gofr/datasource/file/ftp/file_test.go
```diff
@@ -0,0 +1,632 @@
...
+
+	ftpClient := New(config)
+
+	val, ok := ftpClient.(*ftpFileSystem)
+	if ok {
+		logger := logging.NewMockLogger(logging.DEBUG)
+
+		val.UseLogger(logger)
+		val.Connect()
+	}
```
aryanmehrotra, line 628: I doubt if it needs to be checked if ftpFileSystem needs to be implemented

aryanmehrotra (changes requested):

pkg/gofr/datasource/file/ftp/file_test.go
coolwednesday: Connect is a method of filesystem provider interface that embeds the filesystem interface created....New returns a filesystem interface, so I convert it into its concrete type and then call the method connect.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,324 @@
...
+// peekJSONToken peeks the first JSON token without advancing the decoder.
+func (*ftpFile) peekJSONToken(decoder *json.Decoder) (json.Token, error) {
+	newDecoder := *decoder
+
+	token, err := newDecoder.Token()
+	if err != nil {
+		return nil, err
+	}
+
+	return token, nil
```
puellanivis, line 98: Consider replacing this split return into a simple `return newDecoder.Token()`

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,324 @@
...
+
+// Close closes the FTP file connection.
+func (f *ftpFile) Close() error {
+	// Close the FTP response.
+	err := f.response.Close()
+	if err != nil {
+		return err
+	}
+
+	return nil
```
puellanivis, line 188: This is just `return f.response.Close()`

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	b, err := io.ReadAll(r)
+	if err != nil {
+		return 0, err
+	}
+
+	for i := range b {
+		p[i] = b[i]
+	}
```
puellanivis, line 211: This could panic from an out-of-bounds access to `p[i]`.

But then, this should probably just be `n := copy(p, b)` anyways.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,324 @@
...
+		return 0, err
+	}
+
+	// Deep Copy
+	for i := range b {
+		p[i] = b[i]
+	}
+
+	// setting current offset of the file
+	f.offset = off + int64(len(b))
```
puellanivis, line 237: Please read the contract for `io.ReaderAt` and strictly comply with its provisions: https://pkg.go.dev/io@go1.22.5#ReaderAt

Specifically:
> If ReadAt is reading from an input source with a seek offset, ReadAt should not affect nor be affected by the underlying seek offset.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,324 @@
...
+	return 0, datasource.ErrOutOfRange
+}
+
+// Write writes data to the FTP file.
+func (f *ftpFile) Write(p []byte) (n int, err error) {
+	// Delete the existing file
+	err = f.conn.Delete(f.path)
+	if err != nil {
+		return 0, err
+	}
```
puellanivis, line 297: 😮 This is a very unexpected action to be taken!

Note that the semantics of “delete and replace” is very much different from “truncate and rewrite”.

Not to mention the race condition of some other client or task recreating this file between this delete and the `Stor` below.

Also, what if the `f.path` does not already exist? Would this error out with a message that references deleting the file? Because that would be a very surprising error message to receive as the caller.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,324 @@
...
+func (f *ftpFile) WriteAt(p []byte, off int64) (n int, err error) {
+	reader := bytes.NewReader(p)
+
+	// Write data to the FTP file from the specified offset
+	err = f.conn.StorFrom(f.path, reader, uint64(off))
+	if err != nil {
+		return 0, err
+	}
+
+	f.offset = off + int64(len(p))
```
puellanivis, line 321: As with the `ReaderAt` above, `WriteAt` is also not supposed to update the underlying offset.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+
+// Connect establishes a connection to the FTP server and logs in.
+func (f *ftpFileSystem) Connect() {
+	// Construct FTP server address
+	ftpServer := fmt.Sprintf("%v:%v", f.config.Host, f.config.Port)
+
+	// Connect to FTP server
+	const dialTimeoutSeconds = 5
+
+	conn, err := ftp.Dial(ftpServer, ftp.DialWithTimeout(dialTimeoutSeconds*time.Second))
```
puellanivis, line 59: `dialTimeoutSeconds` should be specified as `const dialTimeout = 5 * time.Second` instead of scaling it up separately.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+func (f *ftpFileSystem) Create(name string) (datasource.File, error) {
+	// empty io.Reader
+	emptyReader := new(bytes.Buffer)
+
+	if name == "" {
+		return nil, errors.New("empty filename")
+	}
+
+	// construct the path
+	name = fmt.Sprintf("%s/%s", f.config.RemoteDir, name)
```
puellanivis, line 86: Consider using `path.Join` instead.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+
+	// Issue STOR command to create the empty file on FTP server
+	err := f.conn.Stor(name, emptyReader)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Printf("Created file %s", name)
+
+	s := strings.Split(name, "/")
```
puellanivis, line 96: `_, filename := path.Split(name)`

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,217 @@
...
+	currentDir := ""
+
+	for _, dir := range dirs {
+		// Ignore empty directory names (can happen if there are double slashes)
+		if dir == "" {
+			continue
+		}
+
+		// Build the full path up to this directory
+		currentDir = filepath.Join(currentDir, dir)
```
puellanivis, line 140: Do not use `filepath` for building remote paths. A windows machine running this code would generate the path: `"C:\\Windows"` which the remote service would not understand.

pkg/gofr/datasource/file/ftp/fs.go
puellanivis: The Go `os.MkdirAll` does not cleanup created directories if there is an error: https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/os/path.go;l=18

aryanmehrotra: then we can follow the same logic, but `continue` needs to be removed, as it will result in unexpected results

puellanivis: Indeed, the `continue` still needs to be removed.

pkg/gofr/datasource/file/ftp/file.go
coolwednesday: I guess I will use `StorFrom` instead.

pkg/gofr/datasource/file/ftp/file.go
coolwednesday: Thanks for clarification.

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: Noted.

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: Noted.

pkg/gofr/datasource/file/ftp/mock_logger.go
```diff
@@ -0,0 +1,49 @@
...
+// Level represents different logging levels.
+type Level int
+
+const (
+	DEBUG Level = iota + 1
+	INFO
+	ERROR
+)
+
+type MockLogger struct {
```
puellanivis, line 18: This is now an exported type that is not limited to testing… not sure we want to pollute the public API with a mocked implementation?

But not only that, this isn’t a mock logger, because this is actually printing out log messages?

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// Seek sets the offset for the next Read or ReadAt operation.
+func (f *ftpFile) Seek(offset int64, whence int) (int64, error) {
+	var (
+		err error
+		r   ftpResponse
+	)
+
+	r, err = f.conn.Retr(f.path)
```
puellanivis, line 252: You can use short declaration syntax for `err` and only need to declare `r`:
```go
var r ftpResponse
r, err := f.conn.Retr(f.path)
```

Actually, it looks like `Retr` will return a `ftpResponse` already, because of how `ServerConn` is defined. So, you don’t even need to pre-declare `r` to override the type.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,302 @@
...
+	return res, err
+}
+
+// Write writes data to the FTP file.
+func (f *ftpFile) Write(p []byte) (n int, err error) {
+	reader := bytes.NewReader(p)
+
+	err = f.conn.StorFrom(f.path, reader, uint64(f.offset))
+	if err != nil {
+		return 0, errors.New("failed to write file")
```
puellanivis, line 280: You might want to wrap `err` so you know what the underlying error was?

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,302 @@
...
+	return len(p), nil
+}
+
+// WriteAt writes data to the FTP file starting at the specified offset.
+func (f *ftpFile) WriteAt(p []byte, off int64) (n int, err error) {
+	reader := bytes.NewReader(p)
+
+	err = f.conn.StorFrom(f.path, reader, uint64(off))
+	if err != nil {
+		return 0, fmt.Errorf("failed to write at offset: %v", off)
```
puellanivis, line 296: Same as above, you might want to wrap `err` so you know what the underlying error was?

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,276 @@
...
+		f.logger.Errorf("Failed to connect to FTP server: %v", err)
+		return
+	}
+
+	f.conn = &Conn{conn}
+
+	err = conn.Login(f.config.User, f.config.Password)
+	if err != nil {
+		f.logger.Errorf("Failed to login: %v", err)
+	}
```
puellanivis, line 89: Missing `return`? All it does is log a line after this, but it would still be confusing to see in the logs:
```
Failed to login: foo error
Login Successful
```

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,276 @@
...
+		return errEmptyDirectory
+	}
+
+	filePath := fmt.Sprintf("%s/%s", f.config.RemoteDir, name)
+
+	err := f.conn.MakeDir(filePath)
+	if err == nil {
+		f.logger.Logf("Created %s directory", name)
+		return nil
+	}
```
puellanivis, line 141: [unusual] Normally one does `if err != nil { … }` to keep the non-error path unindented.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,276 @@
...
+		if i == 0 {
+			currentDir = dir
+		} else {
+			currentDir = fmt.Sprintf("%s/%s", currentDir, dir)
+		}
+
+		err := f.conn.MakeDir(currentDir)
+		if err != nil {
+			// if error indicates that directory exists continue, else return.
+			if fmt.Sprint(err) != directoryAlreadyExistError {
```
puellanivis, line 169: [bug] Comment and log disagree with if conditional. In this case we would continue when the error is _not_ `directoryAlreadyExistError`

[fragile test] What happens if the server returns `"550 Fehler bei der Erstellung des Verzeichnis."` because the server is running with `LANG=de`?

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,276 @@
...
+		return errEmptyFilename
+	}
+
+	filePath := fmt.Sprintf("%s/%s", f.config.RemoteDir, name)
+
+	err := f.conn.Delete(filePath)
+	if err == nil || fmt.Sprint(err) == transferComplete {
+		f.logger.Logf("%s successfully removed", name)
+
+		err = nil
```
puellanivis, line 234: There’s not really any good reason to assign `nil` into `err` so someone else can return it, when we can just `return nil` here ourselves.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,302 @@
...
+
+	r, err = f.conn.Retr(f.path)
+	if err != nil {
+		return 0, err
+	}
+
+	p, err := io.ReadAll(r)
+	if err != nil {
+		return 0, err
+	}
```
puellanivis, line 260: I’m a bit confused why we would want to `ReadAll` into a buffer, just to get the length, and then drop it all on the ground for the garbage collector.

Couldn’t we use https://pkg.go.dev/github.com/jlaffaye/ftp#ServerConn.FileSize instead?

puellanivis (commented):
I’m concerned about the number of times this code reads in the whole remote file just to then drop it on the ground for the garbage collector to pick up.

This likely also would cause significant churn if the file size was particularly large compared to the individual reads…

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -43,92 +48,79 @@ func (f *ftpFile) ReadAll() (RowReader, error) {
...
+		f.logger.Errorf("failed to read json file: %v", readerError)
+		return nil, readerError
 	}
 
-	// Check if the JSON is an array
-	if d, ok := token.(json.Delim); ok && d == '[' {
-		// JSON array
-		return &jsonReader{decoder: decoder, token: token}, nil
+	// ASCII value of '[': 91 , '{': 123
+	if buffer[0] != 123 && buffer[0] != 91 {
```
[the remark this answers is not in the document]
coolwednesday: Noted

pkg/gofr/datasource/file/ftp/file.go
coolwednesday: I previously encountered an error with the function, but now that I’m trying it again, it’s working correctly. I will go ahead and replace the old implementation with this updated version.

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: Noted

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -43,92 +48,79 @@ func (f *ftpFile) ReadAll() (RowReader, error) {
...
-	reader := bytes.NewReader(emptyBuffer)
-
-	decoder := json.NewDecoder(reader)
-
-	// Peek the first JSON token to determine the type
-	token, err := f.peekJSONToken(decoder)
-	if err != nil {
-		return nil, err
+	n, readerError := f.Read(buffer)
+	if !errors.Is(readerError, nil) && !errors.Is(readerError, io.EOF) {
```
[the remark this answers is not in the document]
coolwednesday: Noted.

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: Noted.

pkg/gofr/datasource/file/ftp/file.go
coolwednesday: Noted.

pkg/gofr/datasource/file/ftp/file.go
coolwednesday: Noted.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// Read reads data from the FTP file into the provided byte slice and updates the file offset.
+func (f *ftpFile) Read(p []byte) (n int, err error) {
+	r, err := f.conn.RetrFrom(f.path, uint64(f.offset))
+	if err != nil {
+		return 0, err
+	}
+
+	b, err := io.ReadAll(r)
```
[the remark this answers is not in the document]
coolwednesday: Would refactor this.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	r.Close()
+
+	copy(p, b)
+
+	readbytes := len(p)
```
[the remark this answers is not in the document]
coolwednesday: noted.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -43,92 +48,79 @@ func (f *ftpFile) ReadAll() (RowReader, error) {
...
 
-	reader := bytes.NewReader(emptyBuffer)
-
-	decoder := json.NewDecoder(reader)
-
-	// Peek the first JSON token to determine the type
-	token, err := f.peekJSONToken(decoder)
-	if err != nil {
-		return nil, err
+	n, readerError := f.Read(buffer)
```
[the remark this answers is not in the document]
coolwednesday: It was for readability purpose.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	// ASCII value of '[': 91 , '{': 123
+	if buffer[0] != 123 && buffer[0] != 91 {
+		return nil, &json.SyntaxError{}
+	}
+
+	// ASCII value of '/n' : 10 , ']' : 93
+	if n == JSONBufferLength && buffer[n-1] != 10 && buffer[n-1] != 93 {
+		m := bytes.LastIndexByte(buffer, byte('\n'))
```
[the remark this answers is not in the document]
coolwednesday: Noted.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,302 @@
...
+
+		buffer = append(buffer[:m-1], byte(']'))
```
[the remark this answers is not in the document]
coolwednesday: Noted.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -163,58 +162,55 @@ func (f *ftpFile) Name() string {
...
-
-	return len(p), nil
+	return n, nil
 }
 
 // ReadAt reads data from the FTP file starting at the specified offset.
 func (f *ftpFile) ReadAt(p []byte, off int64) (n int, err error) {
 	resp, err := f.conn.RetrFrom(f.path, uint64(off))
 	if err != nil {
+		f.logger.Errorf("ReadAt", "Failed", `Error opening file with path "%v" at %v offset : %v`, f.path, off, err)
```
puellanivis, line 192: Use `%q` instead of `"%s"`

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -244,29 +240,18 @@ func (f *ftpFile) Check(whence int, offset, length int64) (int64, error) {
...
 	if err != nil {
+		f.logger.Errorf("Seek", "Failed", "Error : %v", err)
 		return 0, err
 	}
 
-	p, err := io.ReadAll(r)
+	res, err := f.check(whence, offset, n)
 	if err != nil {
-		return 0, err
+		f.logger.Errorf("Seek", "Failed", "Error : %v", err)
```
puellanivis, line 251: You have removed the `return` here.

This means, we will get a `Seek Failed Error:` message followed by a `Seek Success Offset at whence` message.

pkg/gofr/datasource/file/ftp/interface.go
```diff
@@ -1,12 +1,15 @@
 package ftp
 
 import (
-	"errors"
 	"io"
 	"os"
 	"time"
 )
 
+type Ftp interface {
```
puellanivis, line 9: FTP is an initialism and should be all caps.

pkg/gofr/datasource/file/ftp/interface.go
```diff
@@ -102,12 +97,8 @@ type FileSystemProvider interface {
...
-// ensure that ftp package is not dependent on logging package. That way logging package should be easily able
-// to import ftp package and provide a different "pretty" version for different log types defined here while
-// avoiding the cyclical import issue. Idiomatically, interfaces should be defined by packages who are using it; unlike
-// other languages. Also - accept interfaces, return concrete types.
 type Logger interface {
-	Logf(format string, args ...interface{})
-	Errorf(format string, args ...interface{})
+	Debugf(string, string, string, ...interface{})
+	Logf(string, string, string, ...interface{})
+	Errorf(string, string, string, ...interface{})
```
puellanivis, line 103: If you’re going to have a number of sequential `string` arguments, it can help to name them: `operation, status, msg string` for example.

pkg/gofr/datasource/file/ftp/logger.go
```diff
@@ -0,0 +1,114 @@
...
+	return &FileLog{
+		Level:  level,
+		Out:    os.Stdout,
+		ErrOut: os.Stderr,
+	}
+}
+
+// clean cleans up the query string.
+func clean(query string) string {
+	query = regexp.MustCompile(`\s+`).ReplaceAllString(query, " ")
```
puellanivis, line 42: `MustCompile` should be only used during initialization ( https://go.dev/wiki/CodeReviewComments#dont-panic ). Since this regexp is call invariant anyways, it should be compiled at init time, to avoid recompiling the regexp every time:
```
var regexpSpaces = regexp.MustCompile(`\s+`)
func clean(query string) string {
	return strings.TrimSpace(regexpSpaces.ReplaceAllString(query, " "))
}
```

pkg/gofr/datasource/file/ftp/logger.go
```diff
@@ -0,0 +1,114 @@
...
+func levelString(level Level) string {
+	switch level {
+	case DEBUG:
+		return "DEBUG"
+	case INFO:
+		return "INFO"
+	case ERROR:
+		return "ERROR"
+	default:
+		return ""
```
puellanivis, line 88: Return something instructive about the value, rather than an empty string. `fmt.Sprintf("LEVEL(%d)", int(level))` is common.

pkg/gofr/datasource/file/ftp/logger.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+// Debugf logs a debug message with a format string and arguments.
+func (fl *FileLog) Debugf(operation, status, format string, args ...interface{}) {
+	fl.Operation = operation
+	fl.Status = status
+	fl.Message = fmt.Sprintf(format, args...)
+	fl.logf(DEBUG)
```
puellanivis, line 97: There is no good reason why we should mutate the `FileLog` here, when we can pass these values instead through function arguments.

pkg/gofr/datasource/file/ftp/fs_test.go
```diff
@@ -421,6 +442,15 @@ func TestMkDir(t *testing.T) {
 	}
 }
 
+type directoryAlreadyExistsError struct {
+	code    int
```
puellanivis, line 446: This error seems to cover more errors than just `directory already exists`.

It should also probably be well marked that this is a test error, and is not expected to be coming from anything other than test code.

pkg/gofr/datasource/file/ftp/logger.go
coolwednesday: We would be using another package to print the logs instead of the implementation here. Hence it is just for a mocking the flow.

puellanivis: If it’s strictly just for testing, then I would recommend putting it in a `*_test.go` file.

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: This issue will be handled in the next iterations of improvement of this feature.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,277 @@
...
+// Name returns the name of the file.
+func (f *ftpFile) Name() string {
+	return f.name
+}
+
+// Read reads data from the FTP file into the provided byte slice and updates the file offset.
+func (f *ftpFile) Read(p []byte) (n int, err error) {
+	r, err := f.conn.RetrFrom(f.path, uint64(f.offset))
+	if err != nil {
+		f.logger.Errorf("Read Failed Failed to open file with path %q : %v", f.path, err)
```
ccoVeille, line 152: The "failed failed" seems strange. It could be normal or not. I'm unsure.

Maybe the issue could resolved by adding an hypen `-` or a colon `:` to separate text.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,277 @@
...
+		return 0, err
+	}
+
+	n, err = r.Read(p)
+	if err != nil && !errors.Is(err, io.EOF) {
+		f.logger.Errorf("Read Failed Failed to read from %q : %v", f.path, err)
```
ccoVeille, line 158: Same "failed failed"

pkg/gofr/datasource/file/ftp/fs_test.go
```diff
@@ -0,0 +1,596 @@
+package ftp
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+
```
ccoVeille, line 7:
```suggestion
```

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,290 @@
...
+// Create creates an empty file on the FTP server.
+func (f *ftpFileSystem) Create(name string) (File, error) {
+	if name == "" {
+		f.logger.Errorf("Create_File Failed Provide a valid filename: %v", errEmptyFilename)
+		return nil, errEmptyFilename
+	}
+
+	emptyReader := new(bytes.Buffer)
+
+	filePath := fmt.Sprintf("%s/%s", f.config.RemoteDir, name)
```
ccoVeille, line 105: Please use `filepath.Join`

https://pkg.go.dev/path/filepath#Join

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,290 @@
...
+
+	err := f.conn.Stor(filePath, emptyReader)
+	if err != nil {
+		f.logger.Errorf("Create_File Failed Error creating file  with path %q : %v", filePath, err)
+		return nil, err
+	}
+
+	s := strings.Split(filePath, "/")
```
ccoVeille, line 113: Use filepath.Split

https://pkg.go.dev/path/filepath#Split

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+// Mkdir creates a directory on the FTP server. Here, os.FileMode is unused.
+func (f *ftpFileSystem) Mkdir(name string, _ os.FileMode) error {
+	if name == "" {
+		f.logger.Errorf("Mkdir Failed Provide a valid directory : %v", errEmptyDirectory)
+		return errEmptyDirectory
+	}
+
+	filePath := fmt.Sprintf("%s/%s", f.config.RemoteDir, name)
```
ccoVeille, line 141: filepath.Join

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,290 @@
...
+
+// MkdirAll creates directories recursively on the FTP server. Here, os.FileMode is unused.
+// The directories are not created if even one directory exist.
+func (f *ftpFileSystem) MkdirAll(path string, _ os.FileMode) error {
+	if path == "" {
+		f.logger.Errorf("MkdirAll Failed Provide a valid path : %v", errEmptyPath)
+		return errEmptyPath
+	}
+
+	dirs := strings.Split(path, "/")
```
ccoVeille, line 162: filepath.Split

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+// Open retrieves a file from the FTP server and returns a file handle.
+func (f *ftpFileSystem) Open(name string) (File, error) {
+	if name == "" {
+		f.logger.Errorf("Open_file Failed Provide a valid filename : %v", errEmptyFilename)
+		return nil, errEmptyFilename
+	}
+
+	filePath := fmt.Sprintf("%s/%s", f.config.RemoteDir, name)
```
ccoVeille, line 197: Same

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,290 @@
...
+
+	res, err := f.conn.Retr(filePath)
+	if err != nil {
+		f.logger.Errorf("Open_file Failed Error opening file : %v", err)
+		return nil, err
+	}
+
+	s := strings.Split(filePath, "/")
```
ccoVeille, line 205: Same

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+// Remove deletes a file from the FTP server.
+func (f *ftpFileSystem) Remove(name string) error {
+	if name == "" {
+		f.logger.Errorf("Remove_file Failed Provide a valid filename : %v", errEmptyFilename)
+		return errEmptyFilename
+	}
+
+	filePath := fmt.Sprintf("%s/%s", f.config.RemoteDir, name)
```
ccoVeille, line 233: Same

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,290 @@
...
+
+	err := f.conn.Delete(filePath)
+	if err != nil && fmt.Sprint(err) != transferComplete {
```
ccoVeille, line 236:
```suggestion
	if errors.Is(err, transferComplete) {
```

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,277 @@
...
+
+	res, err := f.check(whence, offset, n)
+	if err != nil {
+		f.logger.Errorf("Seek Failed Error : %v", err)
+		return 0, err
+	}
+
+	f.logger.Logf("Seek Success Offset at whence %v : %v", whence, res)
+
+	return res, err
```
ccoVeille, line 244: The canonical way would be this, even if it's the same

```suggestion
	return res, nil
```

pkg/gofr/datasource/file/ftp/fs.go
ccoVeille, line 236: Nope

Please assert the error is a textproto.Error by using errors.As

https://pkg.go.dev/net/textproto#Error

Then check the code is equal to StatusClosingDataConnection via

https://github.com/jlaffaye/ftp/blob/v0.2.0/status.go#L23

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,290 @@
...
+		f.logger.Errorf("Rename Failed Provide valid arguments : %v", errInvalidArg)
+		return errInvalidArg
+	}
+
+	if oldname == newname {
+		f.logger.Logf("Rename No Action File has the same name")
+		return nil
+	}
+
+	oldFilePath := fmt.Sprintf("%s/%s", f.config.RemoteDir, oldname)
```
ccoVeille, line 278: filepath.Join

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,290 @@
...
+	newFilePath := fmt.Sprintf("%s/%s", f.config.RemoteDir, newname)
```
ccoVeille, line 279: filepath.Join

ccoVeille (changes requested):

pkg/gofr/datasource/file/ftp/fs.go
puellanivis: Please, _do not use `filepath.Join` for remote file paths. This causes an incompatibility when a Windows client talks to a *nix server.

`path.Join` should be used instead.

pkg/gofr/datasource/file/ftp/fs.go
puellanivis: Same, `path.Split`

pkg/gofr/datasource/file/ftp/fs.go
puellanivis: Same, `path.Split`

pkg/gofr/datasource/file/ftp/fs.go
ccoVeille: Oh that's right. I missed that

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: path.Split returns the file and directory ..... here this function creates all the directories iteratively .....so I do need all the directories to check whether they exist and if they don't .. then to create them .... I don't think I can use path.Split. If  I use path.Split I would have to put it inside the iteration to get the directory each time .

puellanivis: https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/os/path.go;l=18 read up on how the built in `MkdirAll` is done, it starts with the equivalent of a `path.Split`

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: thanks ....was stuck here.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,353 @@
...
+	return c.ServerConn.RetrFrom(filepath, offset)
+}
+
+var (
+	errEmptyFilename            = errors.New("filename cannot be empty")
+	errEmptyPath                = errors.New("file/directory path cannot be empty")
+	errEmptyDirectory           = errors.New("directory name cannot be empty")
+	errInvalidArg               = errors.New("invalid filename/directory")
+	transferCompleteError       = &textproto.Error{Code: 226, Msg: "Transfer complete."}
+	directoryAlreadyExistsError = &textproto.Error{Code: 550, Msg: "Create directory operation failed."}
```
ccoVeille, line 36:
```suggestion
	errTransferComplete       = &textproto.Error{Code: textproto.StatusClosingDataConnection, Msg: "Transfer complete."}
	errDirectoryAlreadyExists = &textproto.Error{Code: 550, Msg: "Create directory operation failed."}
```

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+}
+
+// Connect establishes a connection to the FTP server and logs in.
+func (f *ftpFileSystem) Connect() {
+	ftpServer := fmt.Sprintf("%v:%v", f.config.Host, f.config.Port)
+
+	defer f.processLog(&FileLog{Operation: "Connect", Location: ftpServer}, time.Now())
+
+	const dialTimeout = 5 * time.Second
```
ccoVeille, line 82: This should be a configuration variable

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,353 @@
...
+		name:     s,
+		path:     filePath,
+		conn:     f.conn,
+		logger:   f.logger,
+		metrics:  f.metrics,
+	}, nil
+}
+
+// Mkdir creates a directory on the FTP server. Here, os.FileMode is unused.
+func (f *ftpFileSystem) Mkdir(name string, _ os.FileMode) error {
```
ccoVeille, line 145: Why does this method had os.FileMode in its signature then.

I saw the comment, and I don't get it.

The only explanation I can think about is the fact this method could comply an interface or a func(string, os.FileMode)

But then I would expect a comment about it here.

The current comment explains the code, yes, the file mode param is unused. So the comment only confirm it was done on purpose. But there is no explanation about the why it's like that

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,353 @@
...
+		}
+
+	}
+
+	return dirs
+}
+
+// MkdirAll creates directories recursively on the FTP server. Here, os.FileMode is unused.
+// The directories are not created if even one directory exist.
+func (f *ftpFileSystem) MkdirAll(name string, _ os.FileMode) error {
```
ccoVeille, line 193: Same as https://github.com/gofr-dev/gofr/pull/871#discussion_r1699062703

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,353 @@
...
+	dirs := f.mkdirAllHelper(name)
+
+	currentDir := dirs[0]
+
+	for i, dir := range dirs {
+		if i == 0 {
+			continue
+		} else {
+			currentDir = path.Join(currentDir, dir)
+		}
```
ccoVeille, line 210:
```suggestion
		if i == 0 {
			continue
		}
		currentDir = path.Join(currentDir, dir)
```

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,353 @@
...
+		name:     s,
+		path:     filePath,
+		conn:     f.conn,
+		logger:   f.logger,
+		metrics:  f.metrics,
+	}, nil
+}
+
+// permissions are not clear for Ftp as file commands do not accept an argument and don't store their file permissions.
+// currently, this function just calls the Open function. Here, os.FileMode is unused.
```
ccoVeille, line 259: Same as https://github.com/gofr-dev/gofr/pull/871#discussion_r1699062703

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,353 @@
...
+		f.logger.Errorf("Remove_file failed. Provide a valid filename : %v", errEmptyFilename)
+		return errEmptyFilename
+	}
+
+	err := f.conn.Delete(filePath)
+
+	if err != nil {
+		var textprotoError *textproto.Error
+
+		if errors.As(err, &textprotoError) && textprotoError.Msg == transferCompleteError.Msg {
```
ccoVeille, line 280:
```suggestion
		if errors.As(err, &textprotoError) && textprotoError.Code == errTransferComplete.Code {
```

I would have checked only the code

pkg/gofr/datasource/file/ftp/fs.go
ccoVeille, line 36: I'm not even sure we need these in fact

See https://github.com/gofr-dev/gofr/pull/871#discussion_r1699068172

pkg/gofr/datasource/file/ftp/fs.go
ccoVeille: But then

I would have used this

```suggestion
		if errors.As(err, &textprotoError) && textprotoError.Code == textproto.StatusClosingDataConnection {
```

pkg/gofr/datasource/file/ftp/interface.go
```diff
@@ -0,0 +1,106 @@
+package ftp
+
+import (
+	"context"
+	"io"
+	"os"
+	"time"
+)
+
+// FileSystem : Any simulated or real filesystem should implement this interface.
```
ccoVeille, line 10: You should explain this method has this signature because you have to comply ftp.FileSystem

pkg/gofr/datasource/file/ftp/fs.go
ccoVeille, line 36: You are no longer using this one. You refactored mkdir

ccoVeille (changes requested):

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: Instead of Code, I need to check the message because a similar code is returned in various cases. 

for e.g : 
226 is Closing data connection. Requested file action successful (for example, file transfer or file abort).
550 is Requested action not taken. File unavailable (e.g., file not found, no access).

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: Added the comment.

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: We will be not providing it as a part of configs for now.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,355 @@
...
+// Here, os.FileMode is unused, but is added to comply with FileSystem interface.
+func (f *ftpFileSystem) MkdirAll(name string, _ os.FileMode) error {
+	defer f.processLog(&FileLog{Operation: "MkdirAll", Location: path.Join(f.config.RemoteDir, name)}, time.Now())
+
+	if name == "" {
+		f.logger.Errorf("MkdirAll failed. Provide a valid path : %v", errEmptyPath)
+		return errEmptyPath
+	}
+	// returns a slice of those directories that do not exist with the first index being the latest existing parent directory path.
+	dirs := f.mkdirAllHelper(name)
```
puellanivis, line 202: I feel like you’re making your life harder than just using a recursive algorithm…

I will again recommend https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/os/path.go;l=18 as a model.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,355 @@
...
+		return nil, errEmptyFilename
+	}
+
+	res, err := f.conn.Retr(filePath)
+	if err != nil {
+		f.logger.Errorf("Open_file failed. Error opening file : %v", err)
+		return nil, err
+	}
+
+	_, s := path.Split(filePath)
```
puellanivis, line 243: Now that this is simplified, this is just `s := path.Base(filePath)`.

I would pick a more descriptive name than `s` though. Something like `name` or at least `n` or `p`.

Actually, with `path.Base(filePath)` we can just throw this directly into the complex struct literal below.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,355 @@
...
+
+	f.logger.Logf("Open_file successful. Filepath : %q", filePath)
+
+	defer res.Close()
```
puellanivis, line 247: I recommend putting the `defer` as close as possible to the relevant code it is cleaning up after.

pkg/gofr/datasource/file/ftp/fs.go
ccoVeille: It's a pity then, maybe something you cannot fix.

I'm worried about the fact each FTP server may reply different message based on their implementation.

Maybe, @jlaffaye (hi :wave: ) who maintain the ftp lib might help, or suggest something here.

ccoVeille:
I'm glad so see such valuable code review made by you on this PR @puellanivis 

I consider myself as an advanced Gopher, and I'm discovering things with your review.

pkg/gofr/datasource/file/ftp/fs.go
ccoVeille: any suggestion here @puellanivis ? thanks

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,311 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if d, ok := token.(json.Delim); ok && d == '[' {
+		return &jsonReader{decoder: decoder, token: token}, readerError
+	}
+
+	// JSON object
+	return f.createJSONObjectReader(reader)
```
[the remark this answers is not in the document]
coolwednesday: I am logging a different file log for it.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,263 @@
...
+// UseLogger sets the logger interface for the FTP file system.
+func (f *ftpFileSystem) UseLogger(logger interface{}) {
+	if l, ok := logger.(Logger); ok {
+		f.logger = l
+	}
+}
+
+// UseMetrics sets the metrics for the ftpFileSystem client which asserts the Metrics interface.
+// Currently not implemented.
+func (*ftpFileSystem) UseMetrics(_ interface{}) {
```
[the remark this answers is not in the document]
ccoVeille: Good point

This one would have been reported by a revive checker I plan to add to golangci-lint 😁

pkg/gofr/container/container.go
```diff
@@ -121,7 +119,7 @@ func (c *Container) Create(conf config.Config) {
 		c.PubSub = c.createMqttPubSub(conf)
 	}
 
-	c.File = file.New(c.Logger)
+	//c.File = file.New(c.Logger)
```
ccoVeille, line 122: This change look awkward, like a left behind debug or a code that was forgotten to be cleaned

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,354 @@
...
+
+func (f *ftpFileSystem) processLog(fl *FileLog, startTime time.Time) {
+	duration := time.Since(startTime).Milliseconds()
+
+	fl.Duration = duration
+
+	f.logger.Debugf("%v", fl)
+
+	//f.metrics.RecordHistogram(context.Background(), "app_ftp_stats", float64(duration), "hostname", fmt.Sprintf("%v:%v", f.config.Host, f.config.Port),
+	//	"remote directory", f.config.RemoteDir, "type", fl.Operation)
```
ccoVeille, line 353: Same here it's strange

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: I am just more comfortable with iterative algorithm. However it implements the same logic.

go.mod
```diff
@@ -68,6 +68,8 @@ require (
 	github.com/googleapis/enterprise-certificate-proxy v0.3.2 // indirect
 	github.com/googleapis/gax-go/v2 v2.12.5 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.20.0 // indirect
+	github.com/hashicorp/errwrap v1.1.0 // indirect
```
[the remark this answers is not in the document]
coolwednesday: this file has been removed in the recent commits.

pkg/gofr/datasource/file/ftp/fs.go
puellanivis: It’s a sticky spot… I think a way to work around it might be to work backwards, and try and stat the file with `GetEntry` if this errors, move up the chain of dirs. (Eventually, at worst you should hit `/` and that should work. If it doesn’t then err permission out.)

At that point, you can then proceed down the directory tree making the directories as you come back up the call stack. Eventually, it either errors and you bubble that up, or you’re on the last, and you return success.

pkg/gofr/datasource/file/ftp/file.go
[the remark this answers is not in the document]
puellanivis: I mean, right now, you’re logging _both_.

ccoVeille:
One question, you said it resolves 

- #505

But, while it  fixes the FTP part, I don't think it fixes the SFTP one.

So, I won't say it would resolve this issue.

What do you think?

@aryanmehrotra @vipul-rawat @srijan-27 @vipul-rawat 

Should the issue be renamed and a new one will be opened for the SFTP

vipul-rawat:
> One question, you said it resolves 
> 
> - #505
> 
> But, while it  fixes the FTP part, I don't think it fixes the SFTP one.
> 
> So, I won't say it would resolve this issue.
> 
> What do you think?
> 
> @aryanmehrotra @vipul-rawat @srijan-27 @vipul-rawat 
> 
> Should the issue be renamed and a new one will be opened for the SFTP

I think a new child issue for SFTP can be created! Let's break it up for easier reviews.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,299 @@
...
+
+	return len(p), nil
+}
+
+func (f *file) postProcess(fl *FileLog, startTime time.Time) {
+	duration := time.Since(startTime).Milliseconds()
+
+	fl.Duration = duration
+
+	f.logger.Debugf("%v", fl)
```
ccoVeille, line 298: This debug has no context such as text, it seems a left behind debug made for Dev purpose

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,358 @@
...
+func (c *Conn) RetrFrom(filepath string, offset uint64) (ftpResponse, error) {
+	return c.ServerConn.RetrFrom(filepath, offset)
+}
+
+var (
+	errEmptyFilename      = errors.New("filename cannot be empty")
+	errEmptyPath          = errors.New("file/directory path cannot be empty")
+	errEmptyDirectory     = errors.New("directory name cannot be empty")
+	errInvalidArg         = errors.New("invalid filename/directory")
+	transferCompleteError = &textproto.Error{Code: 226, Msg: "Transfer complete."}
```
ccoVeille, line 37:
```suggestion
	transferCompleteError = &textproto.Error{Code: textproto.StatusClosingDataConnection, Msg: "Transfer complete."}
```

But would simply remove this line

```suggestion
```

Because I'm suggesting this https://github.com/gofr-dev/gofr/pull/871#discussion_r1700719709

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,358 @@
...
+	"fmt"
+	"net/textproto"
+	"os"
+	"path"
+	"slices"
+	"time"
+
+	"gofr.dev/pkg/gofr/container"
+
+	"github.com/jlaffaye/ftp"
```
ccoVeille, line 15:
```suggestion
	"time"

	"github.com/jlaffaye/ftp"

	"gofr.dev/pkg/gofr/container"
```

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,358 @@
...
+
+var (
+	errEmptyFilename      = errors.New("filename cannot be empty")
+	errEmptyPath          = errors.New("file/directory path cannot be empty")
+	errEmptyDirectory     = errors.New("directory name cannot be empty")
+	errInvalidArg         = errors.New("invalid filename/directory")
+	transferCompleteError = &textproto.Error{Code: 226, Msg: "Transfer complete."}
+)
+
+// ftpFileSystem represents a file system interface over FTP.
```
ccoVeille, line 40:
```suggestion
// fileSystem represents a file system interface over FTP.
```

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,358 @@
...
+
+	emptyReader := new(bytes.Buffer)
+
+	err := f.conn.Stor(filePath, emptyReader)
+	if err != nil {
+		f.logger.Errorf("Create_File failed. Error creating file with path %q : %v", filePath, err)
+		return nil, err
+	}
+
+	_, s := path.Split(filePath)
```
ccoVeille, line 123: @puellanivis suggested you using path.Base here

https://github.com/gofr-dev/gofr/pull/871#discussion_r1699879257

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,358 @@
...
+		f.logger.Errorf("Remove_file failed. Provide a valid filename : %v", errEmptyFilename)
+		return errEmptyFilename
+	}
+
+	err := f.conn.Delete(filePath)
+
+	if err != nil {
+		var textprotoError *textproto.Error
+
+		if errors.As(err, &textprotoError) && textprotoError.Msg == transferCompleteError.Msg {
```
ccoVeille, line 286: You are not checking the code here

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,299 @@
...
+
+	if err != nil && !errors.Is(err, io.EOF) {
+		f.logger.Errorf("Read failed : Failed to read from %q : %v", f.path, err)
+		return n, err
+	}
+
+	f.logger.Logf("Read successful : Read %v bytes from %q", n, f.path)
+
+	if errors.Is(err, io.EOF) {
+		return n, io.EOF
```
ccoVeille, line 183: Shouldn't it be?

```suggestion
		return n, err
```

Or better 

```suggestion
		return n, nil
```

Or simply remove the if block ?

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,299 @@
...
+	}
```
ccoVeille, line 184: Maybe this would be better no?

```suggestion
```

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,299 @@
...
+
+	return n, nil
```
ccoVeille, line 186: I see it in ReadAt but not here, is this missing?

```suggestion
	 if n < len(p) { 
      return n, io.EOF 
   }
   return nil
```

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,358 @@
...
+
+		f.logger.Errorf("Remove_file failed. Error while deleting the file: %v", err)
+
+		return err
+	}
+
+	f.logger.Logf("Remove_file success. File with path %q successfully removed", filePath)
+
+	return nil
+}
```
ccoVeille, line 299: I would suggest this to avoid having two branches doing the exact sane thing

```suggestion
  var textprotoError *textproto.Error
  switch {
    // expected error
   case errors.As(err, &textprotoError) && textprotoError.Code == textproto.StatusClosingDataConnection && whatever check on the message:
    // nothing to do

    case err != nil:
		  f.logger.Errorf("Remove_file failed. Error while deleting the file: %v", err)
		  return err
	}

	f.logger.Logf("Remove_file success. File with path %q successfully removed", filePath)

	return nil
}
```

pkg/gofr/datasource/file/ftp/.golangci.yml
```diff
@@ -0,0 +1,162 @@
+---
```
ccoVeille, line 1: Why?

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	if d, ok := token.(json.Delim); ok && d == '[' {
+		return &jsonReader{decoder: decoder, token: token}, err
+	}
+
+	// Reading JSON object
+	decoder = json.NewDecoder(reader)
+
+	return &jsonReader{decoder: decoder}, nil
```
ccoVeille, line 91: err is nil when checking the token

Please consider doing this


```suggestion
	token, err := decoder.Token()
	if err != nil {
		return nil, err
	}

	jsonReader := jsonReader{decoder: decoder}
	if d, ok := token.(json.Delim); ok && d == '[' {
		jsonReader.token = token
	}

	return jsonReader, nil
```

ccoVeille (changes requested):

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,358 @@
...
+		metrics:  f.metrics,
+	}, nil
+}
+
+// permissions are not clear for Ftp as file commands do not accept an argument and don't store their file permissions.
+// currently, this function just calls the Open function.
+// Here, os.FileMode is unused, but is added to comply with FileSystem interface.
+func (f *fileSystem) OpenFile(name string, _ int, _ os.FileMode) (container.File, error) {
+	return f.Open(name)
+}
```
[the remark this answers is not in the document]
coolwednesday: Open & OpenFile both methods are necessary for the ftp filesystem to assert the filesystem interface ....Hence both need to be exported.

pkg/gofr/datasource/file/ftp/file.go
coolwednesday: I have to again decode the bytes in the reader into the decoder... because while checking the token the offset moves from 0 to 1.  Then for a normal json object it results in error while parsing.

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: I have already sorted the imports.

pkg/gofr/datasource/file/ftp/file.go
coolwednesday: I think I won't be needing this, I can simply `return n, err` instead at the very last.

pkg/gofr/datasource/file/ftp/.golangci.yml
coolwednesday: might have pushed by mistake.

pkg/gofr/datasource/file/ftp/file.go
ccoVeille: What do you think about it ? @puellanivis

pkg/gofr/datasource/file/ftp/file.go
ccoVeille: Oh, OK. It's indeed something unexpected.

Then maybe adding a comment about the fact the new decoder need to be created because of that

pkg/gofr/datasource/file/ftp/fs.go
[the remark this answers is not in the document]
ccoVeille: Oh, OK I thought you added it when I suggested to add a comment about the unused file mode

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,353 @@
...
+			return err
+		}
+	}
+
+	f.logger.Logf("Directories creation completed successfully.")
+
+	return nil
+}
+
+// Open retrieves a file from the FTP server and returns a file handle.
```
ccoVeille, line 229: Please a comment about this method is needed to comply the interface

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,353 @@
...
+	if name == "" {
+		f.logger.Errorf("Remove_file failed. Provide a valid filename : %v", errEmptyFilename)
+		return errEmptyFilename
+	}
+
+	err := f.conn.Delete(filePath)
+
+	var textprotoError *textproto.Error
+	switch {
+	case errors.As(err, &textprotoError) && textprotoError.Code == ftp.StatusClosingDataConnection && textprotoError.Msg == "Transfer complete.":
```
ccoVeille, line 284: Thanks about doing the change.

I'm now thibk about something that could help I think.

I let you check.

If I understand well the issue this condition is trying to resolve.

- An error can be returned when deleting a file, even if the file was deleted anyway. It might depend on server implementation.

- when this occurs we receive the code

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+}
+
+// Connect establishes a connection to the FTP server and logs in.
+func (f *fileSystem) Connect() {
+	ftpServer := fmt.Sprintf("%v:%v", f.config.Host, f.config.Port)
+
+	defer f.processLog(&FileLog{Operation: "Connect", Location: ftpServer}, time.Now())
+
+	const dialTimeout = 5 * time.Second
```
ccoVeille, line 82: I'm still surprised this to be hard coded here, and not available as a configuration parameter, even if the default could be 5s

pkg/gofr/datasource/file/ftp/fs.go
ccoVeille: But, I'm guessing the text message may vary with FTP server.

So I'm thinking about this.

We are lucky because this strange behavior is only happening when we are deleting a file.

So we could check for the file presence after we receive the code.

This way, no need to check the error message

We receive the code, OK, but does the file exist? No, then it worked. Otherwise something unexpected occurred

pkg/gofr/datasource/file/ftp/fs.go
ccoVeille: I noticed something and suggested it here

https://github.com/gofr-dev/gofr/pull/871#discussion_r1701275839

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,291 @@
...
+	f.offset += int64(n)
+
+	if err != nil && !errors.Is(err, io.EOF) {
+		f.logger.Errorf("Read failed : Failed to read from %q : %v", f.path, err)
+		return n, err
+	}
+
+	f.logger.Logf("Read successful : Read %v bytes from %q", n, f.path)
+
+	return n, err
```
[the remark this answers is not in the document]
coolwednesday: Read returns an EOF error, it reaches the end of the file .... and so does ReadAt

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: Ok, will be moving it to the configs.

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -0,0 +1,311 @@
...
+		return 0, err
+	}
+
+	res, err := f.check(whence, offset, n)
+	if err != nil {
+		f.logger.Errorf("Seek failed. Error : %v", err)
+		return 0, err
+	}
+
+	f.logger.Logf("Seek successful. Offset at whence %v : %v", whence, res)
```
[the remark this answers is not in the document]
coolwednesday: handled this.

pkg/gofr/datasource/file/ftp/file.go
coolwednesday: This is for logging the operation that has been carried out, in case we need it in a file etc. Although I guess Debugf is not required ...debug is suffice for this case.

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: I noticed this behaviour sometimes...I guess the behaviour was observed because of response not being closed somewhere before calling the remove function. Because I am not able to replicate the behaviour anymore. I feel we can do without this check.

ccoVeille: That would be great in fact.

Also, if someone faces it. Then they will receive the error linked to their FTP server

So, maybe we could remove and add a comment to the method saying we noticed some server may return an error even if the delete is successful

ccoVeille (approved):

aryanmehrotra (approved):

docs/advanced-guide/handling-file/page.md
```diff
@@ -15,13 +43,13 @@ err := ctx.File.Mkdir("my_dir",os.ModePerm)
 
 To create subdirectories as well
 ```go
-err := c.File.MkdirAll("my_dir/sub_dir", os.ModePerm)
+err := ctx.File.MkdirAll("my_dir/sub_dir", os.ModePerm)
```
ccoVeille, line 46: I wouldn't use ctx as it's connoted for context.Context variable

ccoVeille (approved):

docs/advanced-guide/handling-file/page.md
aryanmehrotra: But if someone is using gofr, we have actually wrapped the context.Context in gofr's context.
Won't that be better for users to understand? as they won't be using context.Context

PR #885, opened 2024-08-02
Add populating container for gRPC servers at the time of registering service

### LINKED ISSUE #884: Accessing datasources in a gRPC server implementation
Currently, there is no workaround on how we can access the datasources that are initialized during `gofr.New()` in gRPC implementation.

**Describe alternatives you've considered**
One alternative was to define the features required in the server and then manually assign them by using factory functions of those datasources, etc. But we already have them in the `gofr.New()` step, so this kinda seems redundant.

vipul-rawat:
**Description:**
Currently, there is no way to inject the whole container struct and its functionalities through context in the gRPC calls. The only time the control comes to GoFr is when we are creating a gRPC server(through UnaryInterceptor) or while registering a service(app.RegisterService). Added a function that takes the service implementation and sets the container field in the service object using reflection. This feature can set the container field in the service struct.

Closes #884 

**Checklist:**


pkg/gofr/grpc.go
```diff
@@ -57,3 +58,49 @@ func (g *grpcServer) Shutdown(ctx context.Context) error {
...
+	tVal := val.Type()
+
+	for i := 0; i < val.NumField(); i++ {
+		f := tVal.Field(i)
+		v := val.Field(i)
+
+		if f.Type == reflect.TypeOf(c) {
+			if !v.CanSet() {
+				c.Logger.Errorf("cannot inject container as it is not addressable or is fail")
+				continue
```
ccoVeille, line 91: You log an error, an error means you cannot continue usually.

I would have expected a break or return here then

Otherwise it means you can have more than one error when injecting the container.

Or it means you can have an error, but then a success. But there I wouldn't expect to log an error before a success, only when at the end of the loop, no solutions were found.

That's why I think I would expect break or return here instead of continue

pkg/gofr/grpc.go
```diff
@@ -57,3 +58,49 @@ func (g *grpcServer) Shutdown(ctx context.Context) error {
 		return nil
 	})
 }
+
+// RegisterService adds a gRPC service to the GoFr application.
+func (a *App) RegisterService(desc *grpc.ServiceDesc, impl interface{}) {
```
ccoVeille, line 63:
```suggestion
func (a *App) RegisterService(desc *grpc.ServiceDesc, impl any) {
```

pkg/gofr/grpc.go
```diff
@@ -57,3 +58,49 @@ func (g *grpcServer) Shutdown(ctx context.Context) error {
...
+			v.Set(reflect.ValueOf(c))
+		}
+
+		if f.Type == reflect.TypeOf(*c) {
+			if !v.CanSet() {
+				c.Logger.Errorf("cannot inject container as it is not addressable or is fail")
+				continue
+			}
+
+			v.Set(reflect.ValueOf(*c))
```
ccoVeille, line 103: No break or return after this?

pkg/gofr/grpc.go
```diff
@@ -57,3 +58,49 @@ func (g *grpcServer) Shutdown(ctx context.Context) error {
...
+func (a *App) RegisterService(desc *grpc.ServiceDesc, impl interface{}) {
+	a.container.Logger.Infof("registering GRPC Server: %s", desc.ServiceName)
+	a.grpcServer.server.RegisterService(desc, impl)
+
+	injectContainer(impl, a.container)
+
+	a.grpcRegistered = true
+}
+
+func injectContainer(impl any, c *container.Container) {
```
ccoVeille, line 72: I would have added an interface like this

type containerInjecter interface {
       SetContainer(*container.Container)
}

Then in this code check if impl implements containerInjecter

Then add it to grpc example

vipul-rawat: I did try that approach but then thought that it had to be a lot user-dependent. @vikash thoughts on this suggested approach?

pkg/gofr/grpc.go
vipul-rawat: Yes, I think we can do a `break` / `return error` here and not start the GRPC service altogether, earlier it was supposed to be in DEBUG mode but then changed it to ERROR since the user expects the container to be populated.

pkg/gofr/grpc.go
```diff
@@ -57,3 +58,49 @@ func (g *grpcServer) Shutdown(ctx context.Context) error {
 		return nil
 	})
 }
+
+// RegisterService adds a gRPC service to the GoFr application.
+func (a *App) RegisterService(desc *grpc.ServiceDesc, impl interface{}) {
+	a.container.Logger.Infof("registering GRPC Server: %s", desc.ServiceName)
```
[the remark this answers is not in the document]
vipul-rawat: is `gRPC` a brand thing that google does, I see this format everywhere 😅

pkg/gofr/grpc.go
ccoVeille: One way to do it, without reinventing the wheel, is to use an **existing** dependency injection lib, I think.

I'm afraid about the consequence. Something like, you add a piece of that require a container, maybe for test purpose, and then boom, you get an unexpected behavior

ccoVeille:
Why reverting the gRPC change in the logs?

vipul-rawat:
I made `opposite` changes to the GRPC in grpc_test.go 😅

ccoVeille:
But it's [gRPC](https://en.m.wikipedia.org/wiki/GRPC)

vipul-rawat:
No I meant, I earlier changed `gRPC` to `GRPC`, so the reverted changes were changing it back to `gRPC`

pkg/gofr/grpc.go
```diff
@@ -57,3 +59,69 @@ func (g *grpcServer) Shutdown(ctx context.Context) error {
...
+
+		if f.Type == reflect.TypeOf(c) {
+			if !v.CanSet() {
+				c.Logger.Error(errNonAddressable)
+				return errNonAddressable
+			}
+
+			v.Set(reflect.ValueOf(c))
+
+			// early return expecting only one container field necessary for one GRPC implementation
```
ccoVeille, line 109:
```suggestion
			// early return expecting only one container field necessary for one gRPC implementation
```

pkg/gofr/grpc.go
```diff
@@ -57,3 +59,69 @@ func (g *grpcServer) Shutdown(ctx context.Context) error {
...
+
+		if f.Type == reflect.TypeOf(*c) {
+			if !v.CanSet() {
+				c.Logger.Error(errNonAddressable)
+				return errNonAddressable
+			}
+
+			v.Set(reflect.ValueOf(*c))
+
+			// early return expecting only one container field necessary for one GRPC implementation
```
ccoVeille, line 121:
```suggestion
			// early return expecting only one container field necessary for one gRPC implementation
```

pkg/gofr/grpc.go
vipul-rawat: @ccoVeille I am afraid that I don't get the consequences exactly. Will testing the gRPC server implementation cause unexpected behaviour?

ccoVeille (approved):

Umang01-hash (approved):

srijan-27 (approved):

PR #886, opened 2024-08-02
Fix intermittent fail for container and MQTT tests

### LINKED ISSUE #790: Mqtt tests in pipeline fails ofter due to disconnection with public mqtt broker
**Describe the bug**
As in tests the client connection is expected to be up and running, the tests fail if the client fails to disconnect. It is better to use a mock for the MQTT client.

**To Reproduce**
Run the pipeline or run the test for MQTT package locally

vipul-rawat:
Closes #790 

- Used mock Pubsub for container tests as an actual connection to MQTT was not required
- Changed the public broker from HiveMQ to EQMX for more stable connections

### DIFF
--- a/pkg/gofr/container/container_test.go
+++ b/pkg/gofr/container/container_test.go
@@ -164,7 +164,7 @@ func TestContainer_Close(t *testing.T) {
 	mockDB.EXPECT().Close().Return(nil)
 	mockRedis.EXPECT().Close().Return(nil)
 
-	c := NewContainer(config.NewMockConfig(map[string]string{}))
+	c := NewContainer(config.NewMockConfig(nil))
 	c.SQL = mockDB
 	c.Redis = mockRedis
 	c.PubSub = mockPubSub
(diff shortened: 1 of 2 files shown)


pkg/gofr/container/container_test.go
```diff
@@ -159,16 +159,15 @@ func TestContainer_Close(t *testing.T) {
...
+	mockPubSub := &MockPubSub{}
 
 	mockDB.EXPECT().Close().Return(nil)
 	mockRedis.EXPECT().Close().Return(nil)
 
-	configs := map[string]string{
-		"PUBSUB_BACKEND": "MQTT",
-	}
-	c := NewContainer(config.NewMockConfig(configs))
+	c := NewContainer(config.NewMockConfig(map[string]string{}))
```
srijan-27, line 167: I think instead of empty config, we can use `nil`

srijan-27 (changes requested):

srijan-27 (approved):

PR #914, opened 2024-08-07
Adding Tests for external_db.go to increase coverage

coolwednesday:
Description:
- Aim is to increase coverage of the overall package.
- This PR has test functions for external_db.go. We are using pre-existing mock of the datasources present in the container folder. FileProvider interface did not have their mock interfaces, hence generated it using mock gen so that we can mock all the Add database functions.

**Checklist:**


aryanmehrotra (dismissed):

pkg/gofr/datasource/file/mock_interface.go
```diff
@@ -0,0 +1,191 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: provider_interface.go
+//
+// Generated by this command:
+//
+//	mockgen -source=provider_interface.go -destination=mock_interface.go -package=file
```
srijan-27, line 6: I can't find any file with name `provide_interface.go`, can you please check from which file is this mock generated.

pkg/gofr/external_db_test.go
```diff
@@ -0,0 +1,107 @@
+package gofr
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"go.uber.org/mock/gomock"
```
srijan-27, line 8: both are external imports, can be clubbed together.

pkg/gofr/external_db_test.go
```diff
@@ -0,0 +1,107 @@
...
+	t.Run("Adding Clickhouse", func(t *testing.T) {
+		app := New()
+
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		mock := container.NewMockClickhouseProvider(ctrl)
+
+		mock.EXPECT().UseLogger(gomock.Any())
+		mock.EXPECT().UseMetrics(gomock.Any())
```
srijan-27, line 81: we can use `app.Logger()` and `app.Metrics()` instead of `gomock.Any()`, just like other functions.

srijan-27 (changes requested):

pkg/gofr/external_db_test.go
ccoVeille: I'm surprised gci didn't report via golangci-lint 🤔

pkg/gofr/datasource/file/mock_interface.go
coolwednesday: I added the interfaces in that file inside the interface.go file , itself. So that we have one file for interfaces . Updating the mock-gen command.

aryanmehrotra (approved):

srijan-27 (approved):

Issue #912, opened 2024-08-06
Empty data Key in Error Responses


surajit-zs:
Description:
Currently, error responses include a data key with a null value when no data is present. For example:

```json
{
    "error": {
        "message": "missing params: name, userName, password"
    },
    "data": null
}
```
Observation:
It may be more streamlined to omit the data key entirely if there is no data to return. This would make the error responses more concise and potentially easier to handle.

Suggested Action:
Consider modifying the response format to omit the data key when it is null, resulting in a cleaner and more straightforward error response structure.

Additional Context:
If there are specific reasons for including the data key even when null, please provide details. Otherwise, this change could improve response clarity.

surajit-zs:
I will raise a PR for the same

surajit-zs:
- https://github.com/gofr-dev/gofr/pull/915

Could you please review it and let me know if any changes are needed? 
Looking forward to your feedback. 

Thanks!

PR #915, opened 2024-08-07
Omit data Key in Error Responses When No Data is Present

surajit-zs:
This PR updates the response format for error cases, ensuring that the data key is omitted from the response body when there is no data to return. Previously, responses included a data key with a null value, which added unnecessary clutter to the error messages. The changes aim to produce cleaner and more concise error responses.

Added an `isNil` utility function to determine if an interface contains a nullable type with a nil value. This helps in accurately deciding whether to include the data key.

- Closes: #912


aryanmehrotra (dismissed):

pkg/gofr/http/responder.go
```diff
@@ -35,9 +36,11 @@ func (r Responder) Respond(data interface{}, err error) {
...
-		resp = response{
-			Data:  v,
-			Error: errorObj,
+		resp = response{Data: data, Error: errorObj}
+
+		// handling where an interface contains a nullable type with a nil value.
+		if isNil(data) {
+			resp = response{Error: errorObj}
 		}
 	}
```
ccoVeille, line 45: I'm unsure to understand why the omitempty is not enough?

pkg/gofr/http/responder.go
```diff
@@ -35,9 +36,11 @@ func (r Responder) Respond(data interface{}, err error) {
...
 	default:
-		resp = response{
-			Data:  v,
-			Error: errorObj,
+		resp = response{Data: data, Error: errorObj}
+
+		// handling where an interface contains a nullable type with a nil value.
+		if isNil(data) {
+			resp = response{Error: errorObj}
 		}
```
ccoVeille, line 44: The code is a bit strange to read

I would expect something like


```suggestion
		resp = response{Data: data, Error: errorObj}

		// handling where an interface contains a nullable type with a nil value.
		if isNil(data) {
			resp.Data = nil
 		}
```

or


```suggestion
		resp = response{Error: errorObj}

		if !isNil(data) {
			resp.Data = data
 		}
```

surajit-zs: @ccoVeille  sure we can go with 			
```go
resp.Data = nil
```

pkg/gofr/http/responder.go
surajit-zs: image

In this case, it is not working. 

I'm not entirely sure why this is happening and might need to read more about it.

pkg/gofr/http/responder.go
ccoVeille: Maybe I wasn't able to explain correctly.

I'll suggest something else

ccoVeille, line 44: This code would be simpler I think
	
```suggestion
		// handling where an interface contains a nullable type with a nil value.
		if isNil(data) {
			data = nil
 		}
		resp = response{Data: data, Error: errorObj}
```

ccoVeille (approved):

srijan-27:
This PR handles case where the response is `nil` (which is working fine). But what about these cases:
- If the response is empty and error is also present, then the body contains both error and data (empty data)
- If the response is valid struct with error, then the body contains both error and data. In this case I believe the statusCode should be `206`, but its `500`.

FYI: @surajit-zs, @vipul-rawat, @Umang01-hash, @aryanmehrotra

vipul-rawat:
> This PR handles cases where the response is `nil` (which is working fine). But what about these cases:
> 
> * If the response is empty and error is also present, then the body contains both error and data (empty data)
> * If the response is valid struct with error, then the body includes both error and data. In this case I believe the statusCode should be `206`, but its `500`.
> 
> FYI: @surajit-zs, @vipul-rawat, @Umang01-hash, @aryanmehrotra

I think the data field should be present for empty responses (zero values) and we should not modify the user-sent response. For `206` we can create another issue to solve this status code problem.

srijan-27:
created a ticket: #938

srijan-27 (approved):

aryanmehrotra (approved):

PR #929, opened 2024-08-13
Support Dir Operations in FileSystem 

### LINKED ISSUE #919: Support ReadDir, ChDir and Stat to perform Dir related operations
Currently, the filesystem does not have the reading the directory functionality to find information about the directory contents, similar to os.ReadDir()

Edit: Found some more missing functions that could be useful in a fileSystem - ChDir(to change the dir), Stat(to retrieve information about a file or dir)

coolwednesday:
## Pull Request Template

**Description:**

-   Resolves #919 
-   This PR adds support for ReadDir, ChangeDir, CurrentDir, Stat for local filesystem & FTP Server, two filesystems supported right now in GoFr.

**Checklist:**


pkg/gofr/datasource/file/fs.go
```diff
@@ -61,3 +61,26 @@ func (fileSystem) RemoveAll(path string) error {
 func (fileSystem) Rename(oldname, newname string) error {
 	return os.Rename(oldname, newname)
 }
+
+func (fileSystem) Stat(name string) (FileInfo, error) {
```
ccoVeille, line 65: How can it return a FileInfo ?

I would have understand os.FileInfo

If a local fileinfo defined as an alias of os.FileInfo, and was available in the package and here there was a type juggling

Does it even compile?

What am I missing?

pkg/gofr/datasource/file/fs.go
```diff
@@ -61,3 +61,26 @@ func (fileSystem) RemoveAll(path string) error {
...
+	return os.Stat(name)
+}
+
+func (fileSystem) CurrentDir() (string, error) {
+	return os.Getwd()
+}
+
+func (fileSystem) ChangeDir(dir string) error {
+	return os.Chdir(dir)
+}
```
ccoVeille, line 75: Overloading the methods name seems strange. I assume you are reusing the one defined in FTP code

I know it's because of an interface, but I would have added a comment about it

pkg/gofr/datasource/file/fs.go
```diff
@@ -61,3 +61,26 @@ func (fileSystem) RemoveAll(path string) error {
...
+
+func (fileSystem) ReadDir(dir string) ([]FileInfo, error) {
+	entries, err := os.ReadDir(dir)
+
+	fileInfo := make([]FileInfo, len(entries))
+	for i := range entries {
+		fileInfo[i], _ = entries[i].Info()
```
ccoVeille, line 82: I wouldn't ignore the error here

pkg/gofr/datasource/file/fs_test.go
```diff
@@ -321,3 +322,43 @@ func Test_ReadFromJSONArrayInvalidDelimiter(t *testing.T) {
...
+
+	// create a file in the directory.
+	_, err = fileStore.Create("Hello.txt")
+	require.NoError(t, err)
+
+	// ReadDir Operations
+	v, err := fileStore.ReadDir(".")
+	require.NoError(t, err)
+	assert.False(t, v[0].IsDir())
+	assert.Equal(t, "Hello.txt", v[0].Name())
```
ccoVeille, line 358: This would avoid panic

```suggestion
	v, err := fileStore.ReadDir(".")
	require.NoError(t, err)
	require.NotEmpty(t, v)
	assert.False(t, v[0].IsDir())
	assert.Equal(t, "Hello.txt", v[0].Name())
```

pkg/gofr/datasource/file/fs_test.go
```diff
@@ -321,3 +322,43 @@ func Test_ReadFromJSONArrayInvalidDelimiter(t *testing.T) {
...
+
+	v, err = fileStore.ReadDir("../Hello_world")
+	require.NoError(t, err)
+	assert.False(t, v[0].IsDir())
+	assert.Equal(t, "Hello.txt", v[0].Name())
```
ccoVeille, line 363:
```suggestion
	v, err = fileStore.ReadDir("../Hello_world")
	require.NoError(t, err)
	require.NotEmpty(t, v)
	assert.False(t, v[0].IsDir())
	assert.Equal(t, "Hello.txt", v[0].Name())
```

pkg/gofr/datasource/file/ftp/file_test.go
```diff
@@ -688,6 +689,62 @@ func Test_ReadFromJSONArrayInvalidDelimiter(t *testing.T) {
...
+		currentdir, err := fs.CurrentDir()
+		require.NoError(t, err)
+		assert.Equal(t, "/ftp/user/temp2", currentdir)
+
+		_, _ = fs.Create("temp.csv")
+
+		v, err := fs.ReadDir(".")
+		require.NoError(t, err)
+		assert.Equal(t, "temp.csv", v[0].Name())
+		assert.False(t, v[0].IsDir())
```
ccoVeille, line 735:
```suggestion
		v, err := fs.ReadDir(".")
		require.NoError(t, err)
		require.NotEmpty(t, v)
		assert.Equal(t, "temp.csv", v[0].Name())
		assert.False(t, v[0].IsDir())
```

pkg/gofr/datasource/file/ftp/interface.go
```diff
@@ -32,6 +34,9 @@ type ServerConn interface {
 	RemoveDir(path string) error
 	Quit() error
 	FileSize(name string) (int64, error)
+	CurrentDir() (string, error)
+	ChangeDir(path string) error
+	List(path string) ([]*ftp.Entry, error)
```
ccoVeille, line 39: I would defined a type alias to do not expose the ftp lib outside the project.

pkg/gofr/datasource/file/fs.go
coolwednesday: I have added an interface FileInfo that has all the methods of the os.FileInfo. This is because I need to define FileInfo for the ftp Filesystem also .... so that it works for both the filesystems.

pkg/gofr/datasource/file/fs.go
coolwednesday: I assumed that if there is any error getting the file info... then fileInfo[i] would be nil anyway....But I guess calling methods on nil can cause panic .....So I will add continue statement on error ..so that it ignores the file that gave error and appends the rest. 

Does this sound reasonable or should I just return ?

pkg/gofr/datasource/file/ftp/interface.go
coolwednesday: Are you referring to defining a local interface similar to how I did for FileInfo ?

pkg/gofr/datasource/file/fs.go
ccoVeille: Oh indeed, you return an interface that os.FileInfo satisfies 🤦‍♂️

Great then, thanks for the explanation

pkg/gofr/datasource/file/fs.go
ccoVeille: I would say return an error on first error.

It may never happen, but at least it would be covered

pkg/gofr/datasource/file/ftp/interface.go
ccoVeille: It could be ether:
- an interface (if it's possible)
- an alias

An alias would be this:

`type Entry *ftp.Entry`

This way in the List method you loop on the ftp.Entry to create a list of Entry

This way the caller only knows about the gofr ftp package, and not about the external ftp library

ccoVeille:
@coolwednesday you marked things as resolved, but you didn't push your changes for now, right?

coolwednesday:
> @coolwednesday you marked things as resolved, but you didn't push your changes for now, right?

yes, that is right, I have not pushed my changes. @ccoVeille, Sorry for the confusion. I have to also refactor some function's logic. I will push them all together.

pkg/gofr/datasource/file/fs.go
```diff
@@ -61,3 +61,36 @@ func (fileSystem) RemoveAll(path string) error {
...
+// If an error occurs reading the directory, ReadDir returns the entries it was able to read before the error, along with the error.
+// It returns the list of files/directories present in the current directory when "." is passed.
+func (fileSystem) ReadDir(dir string) ([]FileInfo, error) {
+	entries, err := os.ReadDir(dir)
+
+	fileInfo := make([]FileInfo, len(entries))
+	for i := range entries {
+		fileInfo[i], err = entries[i].Info()
+		if err != nil {
+			return fileInfo, err
```
ccoVeille, line 91: I would return nil, but it's open for debate

```suggestion
			return nil, err
```

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -301,6 +348,13 @@ func (f *file) Write(p []byte) (n int, err error) {
 
 	f.offset += int64(len(p))
 
+	t := time.Time{}
+
+	mt := f.ModTime()
+	if mt != t {
+		f.modTime = mt
+	}
```
ccoVeille, line 356: Is this what you were looking for?

```suggestion
	mt := f.ModTime()
	if !mt.IsZero() {
		f.modTime = mt
	}
```

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -323,6 +377,13 @@ func (f *file) WriteAt(p []byte, off int64) (n int, err error) {
 		return 0, err
 	}
 
+	t := time.Time{}
+
+	mt := f.ModTime()
+	if mt != t {
+		f.modTime = mt
+	}
```
ccoVeille, line 385:
```suggestion
	mt := f.ModTime()
	if !mt.IsZero() {
		f.modTime = mt
	}
```

pkg/gofr/datasource/file/ftp/file_test.go
```diff
@@ -688,6 +691,122 @@ func Test_ReadFromJSONArrayInvalidDelimiter(t *testing.T) {
...
+		defer func(fs file_interface.FileSystem) {
+			removeErr := fs.Remove("temp.json")
+			if removeErr != nil {
+				t.Error(removeErr)
+			}
+		}(fs)
+
+		p, err := fs.Stat("temp.json")
+		require.NoError(t, err)
+		assert.NotEqual(t, time.Time{}, p.ModTime())
```
ccoVeille, line 806:
```suggestion
		assert.False(t, p.ModTime().IsZero())
```

pkg/gofr/datasource/file/ftp/file_test.go
```diff
@@ -688,6 +691,122 @@ func Test_ReadFromJSONArrayInvalidDelimiter(t *testing.T) {
...
+
+		mockLogger.EXPECT().Logf(gomock.Any(), gomock.Any()).AnyTimes()
+		mockLogger.EXPECT().Debug(gomock.Any()).AnyTimes()
+
+		_, _ = fs.Create("temp.json")
+		defer func(fs file_interface.FileSystem) {
+			removeErr := fs.Remove("temp.json")
+			if removeErr != nil {
+				t.Error(removeErr)
+			}
```
ccoVeille, line 801:
```suggestion
			removeErr := fs.Remove("temp.json")
			require.NoError(t, removeErr)
```

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -28,6 +29,18 @@ func (c *Conn) RetrFrom(filepath string, offset uint64) (ftpResponse, error) {
 	return c.ServerConn.RetrFrom(filepath, offset)
 }
 
+// List wraps the ftp list method to return a locally defined Entry type.
+func (c *Conn) List(filepath string) ([]Entry, error) {
+	res, err := c.ServerConn.List(filepath)
+
+	var entries []Entry
+	for _, entry := range res {
+		entries = append(entries, entry)
```
ccoVeille, line 38: I thought you had to do this 🤔

```suggestion
		entries = append(entries, Entry(entry))
```

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -142,121 +171,41 @@ func (f *fileSystem) Create(name string) (file_interface.File, error) {
...
-		dirs = append(dirs, dir)
-
-		if parentDir == "" || parentDir == "/" {
-			break
-		}
-
-		currentdir = path.Clean(parentDir)
+	t := time.Time{}
+	mt := fl.ModTime()
+	if mt != t {
```
ccoVeille, line 186: Use Zero

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -276,33 +225,49 @@ func (f *fileSystem) Open(name string) (file_interface.File, error) {
...
+		entryType: ftp.EntryTypeFile,
+		conn:      f.conn,
+		logger:    f.logger,
+		metrics:   f.metrics,
+	}
+
+	t := time.Time{}
+
+	mt := fl.ModTime()
+	if mt != t {
```
ccoVeille, line 241: Use Zero

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -379,6 +325,13 @@ func (f *fileSystem) Rename(oldname, newname string) error {
 
 	msg = fmt.Sprintf("Renamed file %q to %q", oldname, newname)
 	status = "SUCCESS"
+	tempFile.path = newFilePath
+
+	t := time.Time{}
+	mt := tempFile.ModTime()
+	if mt != t {
```
ccoVeille, line 332: Use Zero

pkg/gofr/datasource/file/ftp/fs_dir.go
```diff
@@ -0,0 +1,291 @@
...
+import (
+	"fmt"
+	"os"
+	"path"
+	"slices"
+	"time"
+
+	"github.com/jlaffaye/ftp"
+	file_interface "gofr.dev/pkg/gofr/datasource/file"
+)
```
ccoVeille, line 12:
```suggestion
import (
	"fmt"
	"os"
	"path"
	"slices"
	"time"

	"github.com/jlaffaye/ftp"

	file_interface "gofr.dev/pkg/gofr/datasource/file"
)
```

pkg/gofr/datasource/file/fs.go
coolwednesday: If an error occurs reading the directory, ReadDir returns the entries it was able to read before the error, along with the error. Hence I have returned fileInfo instead of nil. I have added a comment about it.

pkg/gofr/datasource/file/ftp/fs.go
coolwednesday: Entry is already of type ftp.Entry. Hence type asserting it shows redundant type.

pkg/gofr/datasource/file/ftp/fs.go
```diff
@@ -79,7 +92,12 @@ func (f *fileSystem) Connect() {
 
 	ftpServer := fmt.Sprintf("%v:%v", f.config.Host, f.config.Port)
 
-	defer f.processLog(&FileLog{Operation: "Connect", Location: ftpServer, Status: &status, Message: &msg}, time.Now())
+	defer f.processLog(&FileLog{
+		Operation: "Connect",
+		Location:  ftpServer,
+		Status:    &status,
+		Message:   &msg,
```
ccoVeille, line 99: `msg` is never overloaded

Why returning an empty string, I mean the field could stay unset no?

pkg/gofr/datasource/file/ftp/fs_test.go
```diff
@@ -556,30 +569,346 @@ func TestRemoveDir(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
-			path := fmt.Sprintf("%v/%v", tt.basePath, tt.removePath)
+			fs.config.RemoteDir = tt.basePath
 
-			tt.mockRemoveExpect(mockFtpConn, path)
+			tt.mockRemoveExpect(mockFtpConn, tt.basePath, tt.removePath)
 
 			err := fs.RemoveAll(tt.removePath)
 
 			assert.Equal(t, tt.expectError, err != nil, tt.name)
```
ccoVeille, line 594: This should provide a better error reporting if a test fails

```suggestion
			if tt.expectError {
				assert.Error(t, err)
 			} else {
 				assert.NoError(t, err)
 			}
```

ccoVeille:
A few suggestions, but otherwise 👍

ccoVeille (approved):

pkg/gofr/datasource/file/ftp/interface.go
coolwednesday: I have a doubt regarding this implementation. Revisiting the code, I realised that..... what the user is handed over through Stat or ReadDir is FileInfo Interface that allows the user to call methods defined on them. 

They will never be  exposed to ftp.Entry, because I add all the information that ftp.Entry returns in the file struct itself inside the directory functions. This file struct implements the FileInfo interface. So is there really a need to define a type like this for ftp.Entry to mask it ? 

What is the benefit of this, that I might be missing ?

docs/advanced-guide/handling-file/page.md
```diff
@@ -46,6 +46,48 @@ To create subdirectories as well
...
+```go
+currentDir, err := ctx.File.Chdir("..")
+```
+
+To switch to another directory in same parent directory
+```go
+currentDir, err := ctx.File.Chdir("../my_dir2")
+```
+
+To switch to child directory
```
ccoVeille, line 67: What about this?

```suggestion
To switch to a subfolder of the current directory
```

`child` could let think there is a kind of magic that detects the first child or something

docs/advanced-guide/handling-file/page.md
```diff
@@ -46,6 +46,48 @@ To create subdirectories as well
...
+
+for _, entry := range entries {
+entryType := "File"
+
+if entry.IsDir() {
+entryType = "Dir"
+} 
+
+fmt.Printf("%v: %v Size: %v Last Modified Time : %v\n" entryType, entry.Name(), entry.Size(), entry.ModTime())
+}
```
ccoVeille, line 88: Indentation is missing

```suggestion
entries, err := ctx.File.ReadDir("../testdir")

for _, entry := range entries {
    entryType := "File"

    if entry.IsDir() {
        entryType = "Dir"
    } 

    fmt.Printf("%v: %v Size: %v Last Modified Time : %v\n" entryType, entry.Name(), entry.Size(), entry.ModTime())
}
```

docs/advanced-guide/handling-file/page.md
```diff
@@ -76,16 +118,41 @@ for reader.Next() {
 
 
 ### Opening and Reading Content from a File
-
-To open file with default options
+To open a file with default settings, use the `Open` command, which provides read and seek permissions only. For write permissions, use `OpenFile` with the appropriate file modes.
+> Note: In FTP, file permissions are not differentiated; both `Open` and `OpenFile` allow full file operations regardless of specified permissions.
```
ccoVeille, line 122: I'm unsure what you meant, but I would like to suggest this

```suggestion
> Note: In FTP, file permissions are not differentiated; both `Open` and `OpenFile` allow full-file operations regardless of specified permissions.
```

Or


```suggestion
> Note: In FTP, file permissions are not differentiated; both `Open` and `OpenFile` allow all file operations regardless of specified permissions.
```

docs/advanced-guide/handling-file/page.md
```diff
@@ -76,16 +118,41 @@ for reader.Next() {
...
+csvFile.Close()
+
+csvFile, err = ctx.File.OpenFile("my_file.csv", os.O_RDWR, os.ModePerm)
+
+// WriteAt writes the buffer content at the specified offset.
+_, err = csvFile.WriteAt([]byte("test content"), 4)
+if err != nil {
+return nil, err
+}
+
```
ccoVeille, line 140:
```suggestion
// WriteAt writes the buffer content at the specified offset.
_, err = csvFile.WriteAt([]byte("test content"), 4)
if err != nil {
     return nil, err
}

```

docs/advanced-guide/handling-file/page.md
```diff
@@ -76,16 +118,41 @@ for reader.Next() {
...
+Stat retrieves details of a file or directory, including its name, size, last modified time, and type (such as whether it is a file or folder)
+```go
+file, _ := ctx.File.Stat("my_file.text")
+entryType := "File"
+
+if entry.IsDir() {
+entryType = "Dir"
+}
+
+fmt.Printf("%v: %v Size: %v Last Modified Time : %v\n" entryType, entry.Name(), entry.Size(), entry.ModTime())
```
ccoVeille, line 154:
```suggestion
file, _ := ctx.File.Stat("my_file.text")
entryType := "File"

if entry.IsDir() {
     entryType = "Dir"
}

fmt.Printf("%v: %v Size: %v Last Modified Time : %v\n" entryType, entry.Name(), entry.Size(), entry.ModTime())
```

ccoVeille (approved):

pkg/gofr/datasource/file/ftp/interface.go
ccoVeille: The servercon interface is exported.

But I conceded the servercon might not be used outside gofr implementation.

I'm OK with doing without the type alias

It's not a real problem also.

So yes, you can rollback  and clean as you did

ccoVeille: I saw you changed the visibility of ServerConn

The interface was already there, so it was previously exported.

It might be OK, and better.

I would like to get a feedback from 

@aryanmehrotra @vipul-rawat @srijan-27 @Umang01-hash

aryanmehrotra: It should be fine if it is unexported, because it was not used before as well by user. As user would have access to the File interface and File will be used testing. So, I guess it should be fine being unexported.

P.S: We should be vigilant more on what has to be exported going forward, so we don't end up changing later - Starting with keeping unexported is better @ccoVeille @coolwednesday

ccoVeille: Thanks for your feedback.

I now tend to use internal packages when starting a new project, it avoids the classic debate about "how can keep this struct unexported but use it in another package of the same repository" sometimes it's impossible to solve the issue because on cyclic import it may lead, so using internal help. But it's easier to start with it, than adding it to an existing project with large code base

ccoVeille (approved):
Thanks again @coolwednesday for baring with my iterative feedbacks.

aryanmehrotra (approved):

vipul-rawat (approved):

aryanmehrotra (approved):

srijan-27 (approved):

Issue #888, opened 2024-08-02
How to make primary key autoincrement in postgres


Naveenbhargav-Mallampati:
```
type users struct {
	Id         int    `json:"id" gofr:"primary_key,auto_increment"`
	Name       string `json:"name"`
	Age        int    `json:"age"`
	IsEmployed bool   `json:"isEmployed"`
}
```

I have added a addRestHandler for this struct.
cruds are working fine.

The only problem is , in create api --> without passing id : first time creates and row with 0 id , after that , it is returning postgres deplicate id contraint voliation error.

when create if we pass id explicitly it is working.

in Auto cruds , is there a way to make id autoincrement in any way.

sorry if it is a silly  question.

aryanmehrotra:
Yes, it will be supported. It is work in progress in PR830

PR #930, opened 2024-08-14
Fix/auto increment add rest handlers

Umang01-hash:
## Pull Request Template

**Description:**

-   Resolves issue #888 in GoFr
-   Introduces new sql tags for making id `auto-increment` and `not_null` in structs.

**Example:**
```go
type user struct {
	Id         int    `json:"id"  sql:"auto_increment"`
	Name       string `json:"name"  sql:"not_null"`
	Age        int    `json:"age"`
	IsEmployed bool   `json:"isEmployed"`
}
```

**Checklist:**

### DIFF
--- a/pkg/gofr/crud_handlers.go
+++ b/pkg/gofr/crud_handlers.go
@@ -54,11 +54,12 @@ type CRUD interface {
 
 // entity stores information about an entity.
 type entity struct {
-	name       string
-	entityType reflect.Type
-	primaryKey string
-	tableName  string
-	restPath   string
+	name        string
+	entityType  reflect.Type
+	primaryKey  string
+	tableName   string
+	restPath    string
+	constraints map[string]sql.FieldConstraints
 }
 
 // scanEntity extracts entity information for CRUD operations.
@@ -86,13 +87,51 @@ func scanEntity(object interface{}) (*entity, error) {
 	tableName := getTableName(object, structName)
 	restPath := getRestPath(object, structName)
 
-	return &entity{
-		name:       structName,
-		entityType: entityType,
-		primaryKey: primaryKeyFieldName,
-		tableName:  tableName,
-		restPath:   restPath,
-	}, nil
+	e := &entity{
+		name:        structName,
+		entityType:  entityType,
+		primaryKey:  primaryKeyFieldName,
+		tableName:   tableName,
+		restPath:    restPath,
+		constraints: make(map[string]sql.FieldConstraints),
+	}
+
+	for i := 0; i < entityType.NumField(); i++ {
+		field := entityType.Field(i)
+		fieldName := toSnakeCase(field.Name)
+
+		sqlTag := field.Tag.Get("sql")
+		if sqlTag != "" {
+			constraints, err := parseSQLTag(sqlTag)
+			if err != nil {
+				return nil, err
+			}
+
+			e.constraints[fieldName] = constraints
+		}
+	}
+
+	return e, nil
+}
+
+func parseSQLTag(tag string) (sql.FieldConstraints, error) {
+	var constraints sql.FieldConstraints
+	tags := strings.Split(tag, ",")
+
+	for _, t := range tags {
+		t = strings.ToLower(t) // Convert to lowercase for case-insensitivity
+
+		switch t {
+		case "auto_increment":
+			constraints.AutoIncrement = true
+		case "not_null":
+			constraints.NotNull = true
+		default:
+			return constraints, fmt.Errorf("invalid sql tag: %s", t)
+		}
+	}
+
+	return constraints, nil
 }
 
 func getTableName(object any, structName string) string {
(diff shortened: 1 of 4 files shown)


pkg/gofr/crud_handlers.go
```diff
@@ -86,13 +89,52 @@ func scanEntity(object interface{}) (*entity, error) {
...
+		}
+	}
+
+	return e, nil
+}
+
+func parseSQLTag(tag string) (sql.FieldConstraints, error) {
+	var constraints sql.FieldConstraints
+
+	tags := strings.Split(tag, ",")
```
ccoVeille, line 122: what about this?

```suggestion
	for i := 0; i < entityType.NumField(); i++ {
		field := entityType.Field(i)
		fieldName := toSnakeCase(field.Name)
		constraints, err := parseSQLTag(field.Tag)
		if err != nil {
				return nil, err
		}
		e.constraints[fieldName] = constraints
	}

	return e, nil
}

func parseSQLTag(tag whateverItCouldBe) (sql.FieldConstraints, error) {
	var constraints sql.FieldConstraints
	sqlTag := tag.Get("sql")
	if sqlTag == "" {
		return constraints, nil
	}

	tags := strings.Split(tag, ",")
```

pkg/gofr/crud_handlers.go
```diff
@@ -86,13 +89,52 @@ func scanEntity(object interface{}) (*entity, error) {
...
+
+	for _, t := range tags {
```
ccoVeille, line 124: I would use `tag`

```suggestion
	for _, tag := range tags {
```

`t` is connoted for` *testing.T`

pkg/gofr/crud_handlers.go
```diff
@@ -148,30 +190,84 @@ func (a *App) registerCRUDHandlers(e *entity, object interface{}) {
...
+	if hasAutoIncrementID(e.constraints) { // Check for auto-increment ID
+		lastID, err = result.LastInsertId()
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		lastID = fieldValues[0]
+	}
+
+	return fmt.Sprintf("%s successfully created with id: %v", e.name, lastID), nil
```
ccoVeille, line 221: I would have returned only lastID

```suggestion
	return lastID, nil
```

The sprintf can be done in the code calling the method as the entity is the receiver here, e.name is available when calling the method

pkg/gofr/crud_handlers.go
```diff
@@ -148,30 +190,84 @@ func (a *App) registerCRUDHandlers(e *entity, object interface{}) {
...
+			return nil, fmt.Errorf("%w: %s", errFieldCannotBeNull, fieldName)
+		}
+	}
+
+	return newEntity, nil
+}
+
+func (e *entity) extractFields(newEntity interface{}) (fieldNames []string, fieldValues []interface{}) {
+	fieldNames = make([]string, 0, e.entityType.NumField())
+	fieldValues = make([]interface{}, 0, e.entityType.NumField())
```
ccoVeille, line 246:
```suggestion
func (e *entity) extractFields(newEntity any) (fieldNames []string, fieldValues []any) {
	fieldNames = make([]string, 0, e.entityType.NumField())
	fieldValues = make([]any, 0, e.entityType.NumField())
```

pkg/gofr/datasource/sql/query_builder.go
```diff
@@ -1,27 +1,51 @@
 package sql
 
 import (
+	"errors"
 	"fmt"
+
+	"reflect"
 	"strings"
```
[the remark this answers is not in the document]
Umang01-hash: sorted the imports.

pkg/gofr/crud_handlers.go
Umang01-hash: yups renamed to alias `any`

pkg/gofr/crud_handlers.go
Umang01-hash: sure renamed to tag

pkg/gofr/crud_handlers.go
Umang01-hash: sure done this refactoring

pkg/gofr/crud_handlers.go
Umang01-hash: returning the **lastID** here is not a very good option as this is the response what user will get by default until the user doesn't override the CRUD methods. The Create is called by POST here:

```go
if fn, ok := object.(Create); ok {
		a.POST(basePath, fn.Create)
	} else {
		a.POST(basePath, e.Create)
	}
```

thatswhy placed at last of Create so if user chooses to follow the default implementation path then it should give a proper response

```go
func (e *entity) Create(c *Context) (interface{}, error) {
	newEntity, err := e.bindAndValidateEntity(c)
	if err != nil {
		return nil, err
	}

	fieldNames, fieldValues := e.extractFields(newEntity)

	stmt, err := sql.InsertQuery(c.SQL.Dialect(), e.tableName, fieldNames, fieldValues, e.constraints)
	if err != nil {
		return nil, err
	}

	result, err := c.SQL.ExecContext(c, stmt, fieldValues...)
	if err != nil {
		return nil, err
	}

	var lastID interface{}

	if hasAutoIncrementID(e.constraints) { // Check for auto-increment ID
		lastID, err = result.LastInsertId()
		if err != nil {
			return nil, err
		}
	} else {
		lastID = fieldValues[0]
	}

	return fmt.Sprintf("%s successfully created with id: %v", e.name, lastID), nil
}
```

ccoVeille:
There is something I don't get.

You said you changed things and you asked for review, but I see nothing new.

Did I miss something or you forgot pushing?

Umang01-hash:
> There is something I don't get.
> 
> You said you changed things and you asked for review, but I see nothing new.
> 
> Did I miss something or you forgot pushing?

I forgot to push @ccoVeille. Did it now

ccoVeille (approved):

docs/quick-start/add-rest-handlers/page.md
```diff
@@ -49,6 +49,21 @@ func (u *userEntity) TableName() string {
 }
 ```
 
+## Adding Database Constraints
+By default, the GoFr assumes to have manual insertion of id for a given struct, but to support sql constraints like `auto-increment`,
+`not-null` user can use the `sql` tag while declaring the struct fields.
+
+```go
+type user struct {
+	Id         int    `json:"id"  sql:"auto_increment"`
```
aryanmehrotra, line 58: Id should be ID

aryanmehrotra (changes requested):

docs/quick-start/add-rest-handlers/page.md
Umang01-hash: sure changed

aryanmehrotra (approved):

Umang01-hash:
@reski-rukmantiyo Thankyou for your contributions also. Keep using GoFr !

PR #957, opened 2024-08-28
Add support for SFTP

aryanmehrotra:
## Pull Request Template

Screenshot 2024-08-28 at 4 19 10 PM

**Description:**

**Additional Information:**

**Checklist:**


docs/advanced-guide/handling-file/page.md
```diff
@@ -6,7 +6,7 @@ GoFr simplifies the complexity of working with different file stores by offering
 
 By default, local file-store is initialised and user can access it from the context.
 
-Gofr also supports FTP file-store. The file-store can be initialised with FTP as follows:
+Gofr also supports FTP,SFTP file-store. The file-store can be initialised with FTP as follows:
```
ccoVeille, line 9: I would use this

```suggestion
Gofr also supports FTP/SFTP file-store. The file-store can be initialised as follows:
```

or


```suggestion
Gofr also supports FTP and SFTP file-stores. The file-store can be initialised as follows:
```

pkg/gofr/datasource/file/sftp/file.go
```diff
@@ -0,0 +1,142 @@
...
+
+import (
+	"bufio"
+	"encoding/json"
+	"errors"
+	"os"
+	"strings"
+
+	"github.com/pkg/sftp"
+	File "gofr.dev/pkg/gofr/datasource/file"
```
ccoVeille, line 11: Usually package names are in lowercase

pkg/gofr/datasource/file/sftp/file.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+var errNotPointer = errors.New("input should be a pointer to a string")
```
ccoVeille, line 14:
```suggestion
var errNotStringPointer = errors.New("input should be a pointer to a string")
```

pkg/gofr/datasource/file/sftp/file.go
```diff
@@ -0,0 +1,142 @@
...
+// Next checks if there is data available in next line otherwise returns false.
+func (f textReader) Next() bool {
+	return f.scanner.Scan()
+}
+
+// Scan binds the line to provided pointer to string.
+func (f textReader) Scan(i interface{}) error {
+	switch target := i.(type) {
+	case *string:
+		*target = f.scanner.Text()
```
ccoVeille, line 137: target is a local variable, with a string type.

I'm unsure what the code is supposed to do.

I mean storing something in the pointer of a variable that will be no longer exist when this code will return nil

I'm insure this would work, but I would have understand if the code had been something like that

```suggestion
		*i = f.scanner.Text()
```

pkg/gofr/datasource/file/sftp/fs_test.go
```diff
@@ -0,0 +1,304 @@
...
+		{"directory created successfully", nil},
+		{"directory creation failed", errors.New("failed to create file")},
+	}
+
+	for _, tc := range testCases {
+		mocks.client.EXPECT().Mkdir("/test").Return(tc.err)
+
+		err := files.Mkdir("/test", 1)
+
+		require.Equal(t, tc.err, err, "TEST[%d] Failed. Desc %v")
```
ccoVeille, line 57: I would have used `require.ErrorContains`  and do not define local errors with errors.New, or maybe `require.ErrorEqual`

This comment is valid in various places in this test file

pkg/gofr/datasource/file/sftp/fs_test.go
```diff
@@ -0,0 +1,304 @@
...
+	sftpFile := sftp.File{}
+
+	testCases := []struct {
+		desc     string
+		name     string
+		expFile  File.File
+		expError error
+	}{
+		{"File Opened Successfully", "text.csv", file{File: &sftpFile, logger: mocks.logger}, nil},
+		{"File Open Failed", "text.csv", nil, errors.New("File Creation Failed")},
```
ccoVeille, line 235: It's strange it talks about creation when it's file open

pkg/gofr/datasource/file/sftp/file.go
aryanmehrotra: This would work because - we are checking if the `i` variable is a pointer and now target would store the same location so when we update the content at that memory it is also reflected by `i`. 

i would anyway be a pointer, Keeping it this way, it is made sure that we are assigning to a pointer

pkg/gofr/datasource/file/sftp/fs_test.go
aryanmehrotra: I have multiple cases where some cases also have nil error and ErrorEqual expects an error and a string - I would need to check empty string with error? should that be better compared to what I am doing now? @ccoVeille

ccoVeille: I would say you could use a string in the struct, then when the variable is empty you use require.NoError, then when the string is not empty, you can use ErrorContains or EqualError

ccoVeille (approved):

Umang01-hash (approved):

docs/advanced-guide/handling-file/page.md
ccoVeille: You resolved, but I didn't see a change

There were two changes I was suggesting:
- FTP/SFTP or FTP and SFTP
- remove the mention of FTP at the end of sentence as the block is now for both

Umang01-hash: @aryanmehrotra please make these changes and update the documentation

pkg/gofr/datasource/file/sftp/fs_test.go
aryanmehrotra: I am not fully convinced with this.
I believe it's better to check the error as nil in case of no error - there could be a mistake by the person who is writing the code and return an empty string - in that case also it will pass.

ccoVeille:
This comment had not been addressed

https://github.com/gofr-dev/gofr/pull/957#discussion_r1734511563

Not a big issue, that's why I left my review status as "approved"

ccoVeille:
This comment had not been addressed

https://github.com/gofr-dev/gofr/pull/957#discussion_r1734511563

Not a big issue, that's why I left my review status as "approved"

docs/advanced-guide/handling-file/page.md
ccoVeille: This was not done.

I commented another PR about FTP to suggest addressing it

https://github.com/gofr-dev/gofr/pull/958#issuecomment-2324018097

aryanmehrotra: Missed, will fix in a separate PR.

Umang01-hash:
@ccoVeille Sorry this comment didn't get resolved before merging. We will surely address it in the PR #958.

Issue #968, opened 2024-08-30
Improper logs  in sqlite db


aryanmehrotra:
When connecting to sqlite db - logs are coming incorrectly.

Port and host are coming in spite of no port and host required.

To Reproduce: Connect with the following configs - 
DB_DIALECT=sqlite
DB_NAME=test

SkMishra77:
Hi @aryanmehrotra I would like to work on this. Could your please Assign this issue to me?

SkMishra77:
@aryanmehrotra are you referring to this line in log 
,"message":"connected to 'test_db' database at 'localhost:3306'"

here ip:port should not be logged for sqlite

aryanmehrotra:
Yes and if you can set log level to DEBUG you would be seeing more of these

SkMishra77:
ok so for sqlite only changing it to **connected to 'test_db' database**

Umang01-hash:
@SkMishra77 yups you can raise a PR for these changes

SkMishra77:
@aryanmehrotra @Umang01-hash 
Code Climate fails because Cognitive Complexity exceeds so should i refactor this fuction ?

coolwednesday:
@Umang01-hash , I see the linked PR has no latest progress. I would like to work on this. Please assign this to me.

PR #973, opened 2024-09-02
Fixed :: Improper logs in sqlite db

SkMishra77:
**Description:**

Added a condition to check if the Dialect is sqlite then do not log the addr and port.
Closes : #968 

image


aryanmehrotra (changes requested):
Cognitive complexity is high - rather than lot of if else it can be modified to create and log the fields provided by the user

SkMishra77:
@aryanmehrotra Have a Look !!

srijan-27:
@SkMishra77 - There are few linter errors, try running `golangci-lint run` on local. Also, I think we can add tests for these logs.

SkMishra77:
@srijan-27 Can you Help me to identify that how to write testcase for the retryConnection function as it is infinite running function !

srijan-27:
> @srijan-27 Can you Help me to identify that how to write testcase for the retryConnection function as it is infinite running function !

you don't need to specifically write tests for that function, it can be tests along with its master function (NewSQL).
Reference: https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/datasource/sql/sql_test.go#L259

aryanmehrotra:
@SkMishra77 apologies for a late response, could you please check how we are logging using structs - such that you can remove those if else.
https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/datasource/cassandra/logger.go

Here we are pretty printing but you If you can check the QueryLog struct.
As it hampers the readability to have these many if else.

vipul-rawat:
Closing due to stale activity.

PR #983, opened 2024-09-03
downgrade otelhttp dependency

Umang01-hash:


srijan-27 (approved):

srijan-27:
https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6053
This is the issue from otel's side, they have fixed it and will be released in next version, so will be upgrading then.

PR #985, opened 2024-09-04
Add Solr support

srijan-27:
## Pull Request Template

**Description:**

**Additional Information:**

**Checklist:**


pkg/gofr/container/datasources.go
```diff
@@ -266,3 +267,24 @@ type KVStoreProvider interface {
 
 	provider
 }
+
+type Solr interface {
+	Search(ctx context.Context, collection string, params map[string]interface{}) (interface{}, error)
```
ccoVeille, line 272: Please use any, here and everywhere


```suggestion
	Search(ctx context.Context, collection string, params map[string]any) (any, error)
```

pkg/gofr/datasource/solr/solr_test.go
```diff
@@ -0,0 +1,168 @@
...
+	testClientCreate(t, s)
+	testClientDelete(t, s)
+	testClientDeleteField(t, s)
+	testClientListFields(t, s)
+	testClientRetrieve(t, s)
+	testClientUpdate(t, s)
+	testClientUpdateField(t, s)
+}
+
+func testClientSearch(t *testing.T, s *Client) {
```
ccoVeille, line 67: This is a strange way to do

I would add `setupClient(t *testing)` that does everything before you call the test client

Then I would add a TestClientSearch that calls setupClient and get `s` from it

pkg/gofr/datasource/solr/logger.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"fmt"
+	"io"
+	"regexp"
+	"strings"
+)
+
+type Logger interface {
+	Debug(args ...interface{})
```
ccoVeille, line 11:
```suggestion
	Debug(args ...any)
```

Here and everywhere

pkg/gofr/datasource/solr/solr.go
```diff
@@ -0,0 +1,215 @@
...
+
+	return call(ctx, "POST", url, params, document)
+}
+
+// ListFields retrieves all the fields in the schema for the specified collection.
+// params can be used to send query parameters like wt, fl, includeDynamic etc.
+func (c *Client) ListFields(ctx context.Context, collection string, params map[string]interface{}) (interface{}, error) {
+	url := c.url + collection + "/schema/fields"
+
+	defer c.sendOperationStats(&QueryLog{Type: "ListField", Url: url}, time.Now())
```
ccoVeille, line 112: Shouldn't it match the method name?

```suggestion
	defer c.sendOperationStats(&QueryLog{Type: "ListFields", Url: url}, time.Now())
```

pkg/gofr/datasource/solr/solr.go
```diff
@@ -0,0 +1,215 @@
...
+
+	return call(ctx, "GET", url, params, nil)
+}
+
+// Retrieve retrieves the entire schema that includes all the fields,field types,dynamic rules and copy field rules.
+// params can be used to specify the format of response
+func (c *Client) Retrieve(ctx context.Context, collection string, params map[string]interface{}) (interface{}, error) {
```
ccoVeille, line 119: Here and everywhere


```suggestion
func (c *Client) Retrieve(ctx context.Context, collection string, params map[string]any) (any, error) {
```

pkg/gofr/datasource/solr/solr.go
```diff
@@ -0,0 +1,215 @@
...
+// call forms the http request and makes a call to solr and populates the solr response
+func call(ctx context.Context, method, url string, params map[string]interface{}, body io.Reader) (interface{}, error) {
+	req, err := http.NewRequest(method, url, body)
+	if err != nil {
+		return nil, err
+	}
+
+	if method != "GET" {
+		req.Header.Add("content-type", "application/json")
+	}
```
ccoVeille, line 169: Why having no need for me his header with GET?

A comment would be appreciated

pkg/gofr/datasource/solr/solr.go
```diff
@@ -0,0 +1,215 @@
...
+	q := req.URL.Query()
+
+	for k, val := range params {
+		switch v := val.(type) {
+		case []string:
+			for _, val := range v {
+				q.Add(k, val)
+			}
+		default:
+			q.Add(k, fmt.Sprintf("%v", val))
```
ccoVeille, line 180: This is a Sprint


```suggestion
			q.Add(k, fmt.Sprint(val))
```

pkg/gofr/datasource/solr/solr_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+func Test_InvalidRequest(t *testing.T) {
+	_, err := call(context.TODO(), "GET", ":/localhost:", nil, nil)
```
ccoVeille, line 16:
```suggestion
	_, err := call(context.Background()), "GET", ":/localhost:", nil, nil)
```

There is no reason to use a TODO context here (and everywhere in the tests). TODO is there when adding a context somewhere there was no context, but they might be a context at an higher level in the stack, so the TODO is there to say, there are things to do.

Here you are writing a test

pkg/gofr/datasource/solr/solr_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+func setupClient(t *testing.T) *Client {
+	t.Helper()
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, _ = w.Write([]byte(`{
+  		"responseHeader": {
+    	"rf": 1,
+    	"status": 0}}`))
+	}))
```
ccoVeille, line 51: You removed the ts.Close from the code. Maybe because you face a problem with the scope of the defer.

Here you can simply use t.Cleanup


```suggestion
	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		_, _ = w.Write([]byte(`{
  		"responseHeader": {
    	"rf": 1,
    	"status": 0}}`))
	}))

	t.Cleanup(func() {
		ts.Close()
	})
```

pkg/gofr/datasource/solr/solr_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+func Test_ClientUpdate(t *testing.T) {
+	s := setupClient(t)
+
+	body := bytes.NewBuffer([]byte(`{
+		"id": "1234567",
+		"cat": [
+			"Book"
+		]}`))
+	resp, err := s.Update(context.TODO(), "test", body, map[string]interface{}{"commit": "true"})
```
ccoVeille, line 103:
```suggestion
	resp, err := s.Update(context.TODO(), "test", body, map[string]any{"commit": "true"})
```

But in fact, this should be OK


```suggestion
	resp, err := s.Update(context.TODO(), "test", body, map[string]string{"commit": "true"})
```

aryanmehrotra (changes requested):
Shouldn't we utilise gofr's http client?

pkg/gofr/datasource/solr/solr.go
```diff
@@ -0,0 +1,215 @@
...
+// New initializes Solr driver with the provided configuration.
+// The Connect method must be called to establish a connection to Solr.
+// Usage:
+// client := New(config)
+// client.UseLogger(loggerInstance)
+// client.UseMetrics(metricsInstance)
+// client.Connect()
+func New(conf Config) *Client {
+	s := &Client{}
+	s.url = "http://" + conf.Host + ":" + conf.Port + "/solr/"
```
ccoVeille, line 34: Is solr always in http?, so not in https

ccoVeille, line 34: I would remove the trailing slash here

```suggestion
	s.url = "http://" + conf.Host + ":" + conf.Port + "/solr"
```


And then add it back when handling API calls

See https://github.com/gofr-dev/gofr/pull/985#discussion_r1760613769

pkg/gofr/datasource/solr/solr.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+func (c *Client) HealthCheck(ctx context.Context) (any, error) {
+	return nil, nil
+}
+
+// Search searches documents in the given collections based on the parameters specified.
+// This can be used for making any queries to SOLR
+func (c *Client) Search(ctx context.Context, collection string, params map[string]any) (any, error) {
+	url := c.url + collection + "/select"
```
ccoVeille, line 70: https://github.com/gofr-dev/gofr/pull/985#discussion_r1760613131


```suggestion
	url := c.url + "/" + collection + "/select"
```

pkg/gofr/datasource/solr/solr.go
```diff
@@ -0,0 +1,215 @@
...
+
+	defer c.sendOperationStats(&QueryLog{Type: "Search", Url: url}, time.Now())
+
+	return call(ctx, "GET", url, params, nil)
```
ccoVeille, line 74:
```suggestion
	return call(ctx, http.MethodGet, url, params, nil)
```

Same for others

pkg/gofr/datasource/solr/solr.go
```diff
@@ -0,0 +1,215 @@
...
+
+	resp, err := client.Do(req.WithContext(ctx))
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	var respBody any
+
+	b, _ := io.ReadAll(resp.Body)
```
ccoVeille, line 196: Why ignoring the error?


```suggestion
	b, err := io.ReadAll(resp.Body)
	if err != nil {
		return nil, err
	}
```

ccoVeille (approved):
Some minor feedbacks

pkg/gofr/datasource/solr/solr.go
srijan-27: I believe it can be supported with https as well.. but since these calls will be within the same cluster, hence using the `http`

aryanmehrotra (approved):

aryanmehrotra (approved):

pkg/gofr/datasource/solr/solr.go
mfreeman451: shouldn't that be configurable then?

PR #986, opened 2024-09-04
En/fix multipart form types

Umang01-hash:
## Pull Request Template

**Description:**

-   Add support for more types in method `setFieldValue` for multip-part form in **http** package

**Checklist:**


pkg/gofr/http/multipart_file_bind.go
```diff
@@ -146,13 +162,195 @@ func (uf *formData) setFieldValue(value reflect.Value, data string) (bool, error
...
+		reflect.Map, reflect.Pointer, reflect.UnsafePointer:
 		return false, nil
+	}
+
+	return false, nil
+}
+
+func (uf *formData) setInterfaceValue(value reflect.Value, data string) (bool, error) {
+	// If the interface is not set to a concrete value, we can't modify it directly
+	if value.Kind() == reflect.Interface && value.IsNil() {
```
vipul-rawat, line 181: the condition value.Kind() == reflect.Interface is not required as we are already checking it in the switch case.

pkg/gofr/http/multipart_file_bind.go
```diff
@@ -1,14 +1,30 @@
...
 	"strconv"
+	"strings"
 
 	"gofr.dev/pkg/gofr/file"
 )
 
+var (
+	errNilInterface             = errors.New("cannot set value on a nil interface")
+	errUnsupportedInterfaceType = errors.New("unsupported interface value type")
+	errDataLengthExceeds        = errors.New("data length exceeds array capacity")
```
ccoVeille, line 18: I would use past tense for the variable name


```suggestion
	errDataLengthExceeded        = errors.New("data length exceeds array capacity")
```

pkg/gofr/http/multipart_file_bind.go
```diff
@@ -1,14 +1,30 @@
...
+	errUnsupportedKind          = errors.New("unsupported kind")
+	errSettingValue             = errors.New("error setting value at index")
+	errNotAStruct               = errors.New("provided value is not a struct")
+	errFieldNotFound            = errors.New("field not found in struct")
+	errFieldUnexported          = errors.New("cannot set field; it might be unexported")
+	errUnsupportedFieldType     = errors.New("unsupported type for field")
```
ccoVeille, line 24:
```suggestion
	errUnexportedField          = errors.New("cannot set field; it might be unexported")
	errUnsupportedFieldType     = errors.New("unsupported type for field")
```

pkg/gofr/http/multipart_file_bind.go
```diff
@@ -1,14 +1,30 @@
...
 
 	"gofr.dev/pkg/gofr/file"
 )
 
+var (
+	errNilInterface             = errors.New("cannot set value on a nil interface")
+	errUnsupportedInterfaceType = errors.New("unsupported interface value type")
+	errDataLengthExceeds        = errors.New("data length exceeds array capacity")
+	errUnsupportedKind          = errors.New("unsupported kind")
+	errSettingValue             = errors.New("error setting value at index")
```
ccoVeille, line 20: The error name is not consistent with the others. It's about something that failed

Something like this would sound better

```suggestion
	errCannotSetValue             = errors.New("error setting value at index")
```


```suggestion
	errSettingValueFailure             = errors.New("error setting value at index")
```

pkg/gofr/http/multipart_file_bind.go
```diff
@@ -1,14 +1,30 @@
...
 	"mime/multipart"
 	"reflect"
 	"strconv"
+	"strings"
 
 	"gofr.dev/pkg/gofr/file"
 )
 
+var (
+	errNilInterface             = errors.New("cannot set value on a nil interface")
```
ccoVeille: This maybe

```suggestion
	errNilInterfaceFound             = errors.New("cannot set value on a nil interface")
```

pkg/gofr/http/multipart_file_bind.go
```diff
@@ -146,13 +162,195 @@ func (uf *formData) setFieldValue(value reflect.Value, data string) (bool, error
...
+		newSlice = reflect.MakeSlice(value.Type(), len(elements), len(elements))
+	} else if value.Kind() == reflect.Array {
+		if len(elements) > value.Len() {
+			return false, errDataLengthExceeds
+		}
+
+		newSlice = reflect.New(value.Type()).Elem() // Create a new zero-valued array
+	} else {
+		return false, fmt.Errorf("%w: %s", errUnsupportedKind, value.Kind())
+	}
```
ccoVeille, line 220: Using a switch on value.Kind would be easier to read

pkg/gofr/http/multipart_file_bind.go
```diff
@@ -146,13 +162,195 @@ func (uf *formData) setFieldValue(value reflect.Value, data string) (bool, error
...
+// UnmarshalJSON is a custom unmarshaller because json package in Go unmarshal numbers to float64 by default.
+func (c *customUnmarshaller) UnmarshalJSON(data []byte) error {
+	var rawData map[string]interface{}
+
+	err := json.Unmarshal(data, &rawData)
+	if err != nil {
+		return err
+	}
+
+	dataMap := make(map[string]interface{}, len(rawData))
```
ccoVeille, line 316:
```suggestion
	dataMap := make(map[string]any, len(rawData))
```

pkg/gofr/http/multipart_file_bind.go
```diff
@@ -146,13 +163,194 @@ func (uf *formData) setFieldValue(value reflect.Value, data string) (bool, error
...
+		switch val := val.(type) {
+		case float64:
+			if val == float64(int(val)) {
+				dataMap[key] = int(val)
+			} else {
+				dataMap[key] = val
+			}
+		default:
+			dataMap[key] = val
+		}
```
ccoVeille, line 328: This seems very complicated

What about this?

```suggestion
		if valFloat, ok := val.(float64); ok {
			valInt := int(val)
			if valFloat == float64(valInilt) {
				val = valInt
			}
		}
		dataMap[key] = val
```

pkg/gofr/http/multipart_file_bind_test.go
```diff
@@ -54,3 +55,126 @@ func TestGetFieldName(t *testing.T) {
...
+
+	// Test with a slice
+	value := reflect.ValueOf(&testStruct{Slice: nil}).Elem().FieldByName("Slice")
+
+	data := "a,b,c"
+
+	ok, err := uf.setSliceOrArrayValue(value, data)
+	if !ok || err != nil {
+		t.Errorf("setSliceOrArrayValue failed: %v", err)
+	}
```
ccoVeille, line 142: The error message  will be empty if ok is false

pkg/gofr/http/multipart_file_bind_test.go
```diff
@@ -54,3 +55,126 @@ func TestGetFieldName(t *testing.T) {
...
+
+	if len(value.Interface().([]string)) != 3 {
+		t.Errorf("slice not set correctly")
```
ccoVeille, line 145: Shouldn't you fail here?


```suggestion
		t.Fatal("slice not set correctly")
```

ccoVeille, line 145:
```suggestion
		t.Error("slice not set correctly")
```

pkg/gofr/http/multipart_file_bind_test.go
```diff
@@ -54,3 +55,126 @@ func TestGetFieldName(t *testing.T) {
...
+		Field1 string
+		Field2 int
+	}
+
+	uf := &formData{}
+
+	// Test with a valid input string
+	value := reflect.ValueOf(&testStruct{}).Elem()
+
+	data := "{\"Field1\":\"value1\",\"Field2\":123}"
```
ccoVeille, line 170:
```suggestion
	data := `{"Field1":"value1","Field2":123}`
```

pkg/gofr/http/multipart_file_bind_test.go
```diff
@@ -54,3 +55,126 @@ func TestGetFieldName(t *testing.T) {
...
+
+	ok, err := uf.setStructValue(value, data)
+	if !ok || err != nil {
+		t.Errorf("setStructValue failed: %v", err)
+	}
+
+	if value.FieldByName("Field1").String() != "value1" || value.FieldByName("Field2").Int() != 123 {
+		t.Errorf("struct fields not set correctly")
```
ccoVeille, line 178:
```suggestion
		t.Error("struct fields not set correctly")
```

Or you provide information about the struct via the error

pkg/gofr/http/multipart_file_bind.go
```diff
@@ -146,13 +163,194 @@ func (uf *formData) setFieldValue(value reflect.Value, data string) (bool, error
...
+	if value.Kind() != reflect.Struct {
+		return false, errNotAStruct
+	}
+
+	dataMap, err := parseStringToMap(data)
+	if err != nil {
+		return false, err
+	}
+
+	anyFieldSet := false
```
ccoVeille, line 250: I would return an error when no field is set this way, the return signature will be a simple err

pkg/gofr/http/multipart_file_bind.go
```diff
@@ -146,13 +163,194 @@ func (uf *formData) setFieldValue(value reflect.Value, data string) (bool, error
...
+		field, err := getFieldByName(value, key)
+		if err != nil {
+			return false, err
+		}
+
+		if err := setFieldValueFromData(field, val); err != nil {
+			return false, err
+		}
+
+		anyFieldSet = true
```
ccoVeille, line 262: Is there a need to continue when one is set?
It could be ok, but I'm asking

pkg/gofr/http/multipart_file_bind.go
Umang01-hash: yups sure we can use this

pkg/gofr/http/multipart_file_bind.go
Umang01-hash: updated error name to  `errSettingValueFailure`

pkg/gofr/http/multipart_file_bind.go
Umang01-hash: replaced with a switch

pkg/gofr/http/multipart_file_bind.go
Umang01-hash: sure did the change

pkg/gofr/http/multipart_file_bind_test.go
Umang01-hash: sure updated it

pkg/gofr/http/multipart_file_bind.go
Umang01-hash: changed to **errUnexportedField**

pkg/gofr/http/multipart_file_bind.go
Umang01-hash: this error has been removed

pkg/gofr/http/multipart_file_bind_test.go
Umang01-hash: used **Fatal** here

pkg/gofr/http/multipart_file_bind_test.go
Umang01-hash: Sure updated

pkg/gofr/http/multipart_file_bind_test.go
Umang01-hash: changed to **Error** from **Errorf**

pkg/gofr/http/multipart_file_bind.go
Umang01-hash: simplied it

pkg/gofr/http/multipart_file_bind.go
Umang01-hash: I’ve updated the function to return an error if no fields were set, simplifying the return signature.

pkg/gofr/http/multipart_file_bind.go
Umang01-hash: I’ve modified the loop to stop processing once a field is successfully set, as you suggested continuing further isn’t necessary.

pkg/gofr/http/multipart_file_bind_test.go
Umang01-hash: updated that

ccoVeille (approved):
thanks for applying the changes I suggested

pkg/gofr/http/multipart_file_bind_test.go
```diff
@@ -54,3 +55,129 @@ func TestGetFieldName(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		f := &formData{}
+		val := reflect.New(reflect.TypeOf(tc.valueType.value)).Elem()
+
+		set, err := f.setFieldValue(val, tc.data)
+		if err != nil {
+			t.Errorf("Unexpected error for value kind %v and data %q: %v", val.Kind(), tc.data, err)
+		}
```
ccoVeille, line 92: You are already using testify in the file

```suggestion
		require.NoErrorf(t, err, "Unexpected error for value kind %v and data %q", val.Kind(), tc.data)
```

pkg/gofr/http/multipart_file_bind_test.go
```diff
@@ -54,3 +55,129 @@ func TestGetFieldName(t *testing.T) {
...
+
+		if set != tc.expected {
+			t.Errorf("Expected set to be %v for value kind %v and data %q, got %v", tc.expected, val.Kind(), tc.data, set)
+		}
```
ccoVeille, line 96: assert.Equalf(t, tc.expected, set, "Expected set to be %v for value kind %v and data %q", tc.expected, val.Kind(), tc.data)

pkg/gofr/http/multipart_file_bind_test.go
```diff
@@ -54,3 +55,129 @@ func TestGetFieldName(t *testing.T) {
...
+
+	data := "a,b,c"
+
+	ok, err := uf.setSliceOrArrayValue(value, data)
+
+	assert.True(t, ok, "setSliceOrArrayValue failed")
+
+	if err != nil {
+		t.Errorf("setSliceOrArrayValue failed: %v", err)
+	}
```
ccoVeille, line 145: require.NoError

I'm sure you can review the file and find others

pkg/gofr/http/multipart_file_bind_test.go
Umang01-hash: sure updated

pkg/gofr/http/multipart_file_bind_test.go
Umang01-hash: used `assert.Equal` instead of if

pkg/gofr/http/multipart_file_bind_test.go
Umang01-hash: Yups i will and update all of them

ccoVeille (approved):

aryanmehrotra (changes requested):
Lets add docs as well!

pkg/gofr/http/form_data_binder.go
```diff
@@ -0,0 +1,176 @@
...
+		if err := setFieldValueFromData(field, val); err != nil {
+			return false, err
+		}
+
+		// Return true and nil error once a field is set
+		return true, nil
+	}
+
+	// Return false and an error if no fields were set
+	return false, errFieldsNotSet
```
ccoVeille, line 89: There is something strange with this loop

Every single branch of the loop does a return

So it means there is no possible second iteration.

Also, the return at the end can only be reached if the dataMap is empty

So either, there is a bug, or it's an attempt to solve the "I have a map with one entry, but I don't know its key" problem.

But the code is a bit strange, as it let think the code is not linear.

Here is what I would suggest.

```suggestion
	dataMap, err := parseStringToMap(data)
	if err != nil {
		return false, err
	}
	
	if len(dataMap) == 0 {
		// Return false and an error if no fields were set
		return false, errFieldsNotSet
	}
	
	var key string
	var val any
	for key, val = range dataMap {
		// we only need to iterate to get one element
		break
	}
	field, err := getFieldByName(value, key)
	if err != nil {
		return false, err
	}

	if err := setFieldValueFromData(field, val); err != nil {
		return false, err
	}

	// Return true and nil error once a field is set
	return true, nil
```

Another solution is to use this but it might be overkill

```suggestion
	dataMap, err := parseStringToMap(data)
	if err != nil {
		return false, err
	}
	
	if len(dataMap) == 0 {
		// Return false and an error if no fields were set
		return false, errFieldsNotSet
	}
	
	key := maps.Keys(dataMap)[0]
	var := dataMap[key]

	field, err := getFieldByName(value, key)
	if err != nil {
		return false, err
	}

	if err := setFieldValueFromData(field, val); err != nil {
		return false, err
	}

	// Return true and nil error once a field is set
	return true, nil
```

Umang01-hash: Ahmm ya i refactored the code and used your first approach.

ccoVeille (approved):

pkg/gofr/http/form_data_binder.go
vipul-rawat: The implementation is wrong in the sense that it is only setting values for a single field when trying to set the values of a struct. There could be cases where multiple fields need to be set, currently only one field of the struct is being set by the function which is not the intended use.

This function should set values for all the fields for which form data is present and are settable by reflect.

vipul-rawat: Also, this method of setting field values is quite inefficient as we are iterating over all the fields of a struct for every data map entry. It would be efficient to iterate over the fields and fetch is values are present in the data map.

Umang01-hash: sure modified the `setStructValue` method accordingly

pkg/gofr/http/form_data_binder.go
```diff
@@ -0,0 +1,211 @@
...
+	switch value.Kind() {
+	case reflect.Slice:
+		newSlice = reflect.MakeSlice(value.Type(), len(elements), len(elements))
+	case reflect.Array:
+		if len(elements) > value.Len() {
+			return false, errDataLengthExceeded
+		}
+
+		newSlice = reflect.New(value.Type()).Elem()
+	case reflect.Invalid, reflect.Bool, reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64, reflect.Uint,
```
vipul-rawat, line 37: The case seems un-necessary since it is already checked before that if value.Kind is either slice or array

Umang01-hash: Sure i have replaced the `switch` here with if-else

vipul-rawat (approved):

aryanmehrotra (dismissed):

aryanmehrotra (approved):

PR #988, opened 2024-09-04
Fixing sql mocks

### LINKED ISSUE #960: Unable to test all SQL database operations
Since we have HealthCheck in the interface, we are not able to assign sqlmock to the DB interface, due to which QueryRow and QueryRowContext cannot be tested

coolwednesday:
**Description:**
-  Closes #960 
-   Previously the mock was not able to mock Query, Queryrow, Select, Dialect, HealthCheck methods. Hence, replaced            MockGen generated sql mocks with go-mock-sql package in the mock container.
-   Added Mock functions for Select, Dialect, HealthCheck as they are extra methods added by GoFr's DB interface.
-  Corrected existing tests to use the new mocking behaviour of sql

**Checklist:**


pkg/gofr/container/container_test.go
```diff
@@ -4,11 +4,11 @@ import (
 	"context"
 	"testing"
 
+	"go.uber.org/mock/gomock"
+
 	"github.com/gorilla/websocket"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"go.uber.org/mock/gomock"
```
ccoVeille, line 10: this line should remain, there is no reason to have a distinct block for gomock before the github.com packages

pkg/gofr/container/datasources.go
```diff
@@ -5,13 +5,12 @@ import (
 	"database/sql"
 
 	"github.com/redis/go-redis/v9"
+	gofrSQL "gofr.dev/pkg/gofr/datasource/sql"
 
 	"gofr.dev/pkg/gofr/datasource"
-	gofrSQL "gofr.dev/pkg/gofr/datasource/sql"
```
ccoVeille, line 10: same here

pkg/gofr/container/health_test.go
```diff
@@ -9,7 +9,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-
```
ccoVeille, line 12: you should not remove this line

I'm surprised gci linter didn't report an issue

https://github.com/gofr-dev/gofr/blob/eedaec521dac7b4d624b7f4cab510e8eab9fde11/.golangci.yml#L10-L14

pkg/gofr/container/mock_container.go
```diff
@@ -2,10 +2,14 @@ package container
...
 	"fmt"
 	"net/http"
 	"testing"
 
+	"github.com/DATA-DOG/go-sqlmock"
+	"gofr.dev/pkg/gofr/datasource/sql"
+
 	"go.uber.org/mock/gomock"
 
 	"gofr.dev/pkg/gofr/datasource"
```
ccoVeille, line 15: Please install [gci](https://github.com/daixiang0/gci/) and run

```shell
gci write -s standard -s default -s localmodule .
```

pkg/gofr/container/mock_container.go
```diff
@@ -75,7 +114,114 @@ func NewMockContainer(t *testing.T) (*Container, Mocks) {
...
+		m.logger.Fatalf("expected query: %s, actual query: %s", query, expectedText)
+	}
+
+	if len(args) != len(expectedArgs) {
+		m.logger.Fatalf("expected %d args, actual %d", len(expectedArgs), len(args))
+	}
+
+	for i := range args {
+		if args[i] != expectedArgs[i] {
+			m.logger.Fatalf("expected arg %d, actual arg %d", args[i], expectedArgs[i])
```
ccoVeille, line 136: maybe error would be enough here

```suggestion
			m.logger.Errorf("expected arg %d, actual arg %d", args[i], expectedArgs[i])
```

This way the error message will be clearer if something fails

pkg/gofr/container/mockcontainer_test.go
```diff
@@ -0,0 +1 @@
+package container
```
ccoVeille, line 1: why adding this empty file ?

did you plan to add tests and forgot them ?

pkg/gofr/container/mock_container.go
```diff
@@ -75,7 +114,114 @@ func NewMockContainer(t *testing.T) (*Container, Mocks) {
...
+	}
+
+	m.queryWithArgs = m.queryWithArgs[:lastIndex]
+}
+
+func (m sqlMockDB) HealthCheck() *datasource.Health {
+	if m.expectedHealthCheck == nil || len(m.expectedHealthCheck) == 0 {
+		m.logger.Fatal("Did not expect any mock calls for HealthCheck")
+	}
+
```
ccoVeille, line 147: I'm surprised this file and a few others do not use testify but plain testing

I mean here, I would have expected something like this

```suggestion
	require.NotEmpty(m.expectedHealthCheck, "Did not expect any mock calls for HealthCheck")
```

pkg/gofr/migration/sql_test.go
```diff
@@ -4,123 +4,153 @@ import (
...
+		mocks.SQL.ExpectExec("DELETE FROM users WHERE id = ?").WithArgs(1).WillReturnResult(expectedResult)
+		sqlDB := mockContainer.SQL
 
 		result, err := sqlDB.Exec("DELETE FROM users WHERE id = ?", 1)
+		require.NoError(t, err)
 
-		if err != nil {
-			t.Errorf("Exec should return no error, got: %v", err)
-		}
+		expectedLastInserted, err := expectedResult.LastInsertId()
```
ccoVeille, line 132: a last insertid on a delete ?

I understand everything is mocked but it's strange

ccoVeille (approved):
sounds good except a few minor things

pkg/gofr/container/mock_container.go
coolwednesday: Here, I would like to stop the execution instead of just an error log, as proceeding even after the query and arguments of the function do not match, would not be the ideal behaviour in case of sqlmocks.

pkg/gofr/container/mockcontainer_test.go
coolwednesday: Sorry for the confusion , I forgot to push the recent changes.

pkg/gofr/container/mock_container.go
ccoVeille: OK, indeed

pkg/gofr/container/mockcontainer_test.go
ccoVeille: 😂👍

pkg/gofr/container/mock_container.go
coolwednesday: I don't pass the t *Testing.T in the functions that is required for require and assert functions.

pkg/gofr/migration/sql_test.go
```diff
@@ -4,123 +4,153 @@ import (
...
-			t.Errorf("Query should return the expected rows, got: %v", rows)
+		for rows.Next() {
+			err = rows.Scan(&id, &name)
+			require.NoError(t, err)
+			require.NoError(t, rows.Err())
+			assert.Equal(t, expectedResult[i].id, id)
+			assert.Equal(t, expectedResult[i].name, name)
+
+			i++
 		}
```
ccoVeille, line 51: I know these are tests, but usually after a `for rows->Next` loop

It's a good thing to check if rows->Error() don't return an error

pkg/gofr/migration/sql_test.go
```diff
@@ -4,123 +4,153 @@ import (
...
-			t.Errorf("Query should return the expected error, got: %v", err)
+		rows, err := sqlMockDB.Query("SELECT * FROM unknown_table")
+		require.NoError(t, err)
+
+		for rows.Next() {
+			err = rows.Scan(&id, &name)
+			require.NoError(t, rows.Err())
+			require.Error(t, err)
+			assert.Equal(t, expectedErr, err)
 		}
```
ccoVeille, line 74: Same remark

https://github.com/gofr-dev/gofr/pull/988#discussion_r1744867529

pkg/gofr/migration/sql_test.go
coolwednesday: I suppose Last insertID does not change with a delete query...but as all exec commands return Result that has last inserted and rowsAffected .....I think a non-zero value would be returned. I might be wrong....but I can't find any definite answer about this, around on the internet.

pkg/gofr/container/container_test.go
```diff
@@ -160,15 +160,16 @@ func TestContainer_Close(t *testing.T) {
...
 	mockRedis := NewMockRedis(controller)
 	mockPubSub := &MockPubSub{}
 
-	mockDB.EXPECT().Close().Return(nil)
 	mockRedis.EXPECT().Close().Return(nil)
+	sqlMock.ExpectClose()
 
 	c := NewContainer(config.NewMockConfig(nil))
-	c.SQL = mockDB
+	c.SQL = &sqlMockDB{mockDB, &e, logging.NewLogger(logging.DEBUG)}
```
Umang01-hash, line 172: Can we not directly write it like this? Is there any specific need of variable `e` ?

`c.SQL = &sqlMockDB{mockDB, &expectedQuery{}, logging.NewLogger(logging.DEBUG)}`

pkg/gofr/container/mock_container.go
```diff
@@ -32,8 +33,16 @@ func NewMockContainer(t *testing.T) (*Container, Mocks) {
 
 	ctrl := gomock.NewController(t)
 
-	sqlMock := NewMockDB(ctrl)
-	container.SQL = sqlMock
+	mockDB, sqlMock, _ := sql.NewSQLMocks(t)
+	// initialisation of expectations
+	e := expectedQuery{}
+
+	sql2 := &mockSQL{sqlMock, &e}
```
Umang01-hash, line 40: `sql2` name is not very clear. Maybe we can name it to `sqlMockWrapper`, `testSQLMock` or anything more relevant?

pkg/gofr/container/mock_container.go
```diff
@@ -32,8 +33,16 @@ func NewMockContainer(t *testing.T) (*Container, Mocks) {
 
 	ctrl := gomock.NewController(t)
 
-	sqlMock := NewMockDB(ctrl)
-	container.SQL = sqlMock
+	mockDB, sqlMock, _ := sql.NewSQLMocks(t)
+	// initialisation of expectations
+	e := expectedQuery{}
```
Umang01-hash, line 38: No need of this variable, just use `expectedQuery` struct inside `&mockSQL{}`

pkg/gofr/container/mockcontainer_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"gofr.dev/pkg/gofr/datasource"
+)
+
+func TestMockSQL_Select(t *testing.T) {
+	ids := make([]string, 0)
```
Umang01-hash, line 13: Why not to use this instead?
`ids := []string{"1", "2"}`

pkg/gofr/container/sql_mock.go
```diff
@@ -0,0 +1,163 @@
+package container
+
+import (
+	"context"
+	gosql "database/sql"
```
Umang01-hash, line 5: Instead of adding alias to the official library, let's add it to our GoFr Sql package.

pkg/gofr/migration/sql_test.go
```diff
@@ -4,123 +4,153 @@ import (
...
+
+		i := 0
 
-		if rows != expectedRows {
-			t.Errorf("Query should return the expected rows, got: %v", rows)
+		for rows.Next() {
+			err = rows.Scan(&id, &name)
+			require.NoError(t, err)
+			require.NoError(t, rows.Err())
+			assert.Equal(t, expectedResult[i].id, id)
```
Umang01-hash, line 47: We can use replace all the `assert` with `require` so as to use only one of them for tests.

Umang01-hash (changes requested):

pkg/gofr/container/container_test.go
coolwednesday: I am initialising a struct and then I am passing the pointer to the same struct in mocksql as well as mocksqlDB. S I intialised it seperately. Though I will change e to a better variable name.

Umang01-hash (approved):

aryanmehrotra (approved):

Issue #999, opened 2024-09-06
Refactor Select method in SQL DB.


aryanmehrotra:
Method DB.Select has a Cognitive Complexity of 15 (exceeds 12 allowed).

vigneshp91:
I'm willing to work on this @aryanmehrotra

srijan-27:
@vigneshp91 - sure, assigning this to you.

PR #1019, opened 2024-09-13
fixes: issue 999 refactor db select to reduce cognitive complexity

vigneshp91:
Fixes https://github.com/gofr-dev/gofr/issues/999
**Description:**

-  To reduce the cognitive complexity created two functions to handle the select operation.


aryanmehrotra (approved):

pkg/gofr/datasource/sql/db.go
```diff
@@ -231,37 +230,44 @@ func (d *DB) Select(ctx context.Context, data interface{}, query string, args ..
...
-		}
+	if rvo.Elem().CanSet() {
+		rvo.Elem().Set(rv)
+	}
+}
 
-	default:
-		d.logger.Debugf("a pointer to %v was not expected.", rv.Kind().String())
+func (d *DB) selectStruct(ctx context.Context, query string, args []interface{}, rv reflect.Value) {
+	rows, _ := d.QueryContext(ctx, query, args...)
```
ccoVeille, line 268: Why ignoring the error?

If it's legitimate, please add a comment

I noted previous code was not doing it, but it looks odd

ccoVeille: Thanks

vigneshp91:
@aryanmehrotra Can you approve and merge this?

pkg/gofr/datasource/sql/db.go
```diff
@@ -231,37 +230,49 @@ func (d *DB) Select(ctx context.Context, data interface{}, query string, args ..
...
+func (d *DB) selectStruct(ctx context.Context, query string, args []interface{}, rv reflect.Value) {
+	rows, err := d.QueryContext(ctx, query, args...)
+	if err != nil {
+		d.logger.Errorf("error running query: %v", err)
+		return
+	}
+
+	for rows.Next() {
+		d.rowsToStruct(rows, rv)
 	}
```
ccoVeille, line 276: No error management to do with something like this?


```suggestion
	for rows.Next() {
		d.rowsToStruct(rows, rv)
 	}

	if err:= rows.Err() {
		return err
 	}
```

pkg/gofr/datasource/sql/db.go
```diff
@@ -231,37 +230,49 @@ func (d *DB) Select(ctx context.Context, data interface{}, query string, args ..
...
+			_ = rows.Scan(val.Interface())
 		}
 
-	case reflect.Struct:
-		rows, _ := d.QueryContext(ctx, query, args...)
-		for rows.Next() {
-			d.rowsToStruct(rows, rv)
-		}
+		rv = reflect.Append(rv, val.Elem())
+	}
```
ccoVeille, line 260: Same question as here

https://github.com/gofr-dev/gofr/pull/1019#discussion_r1760304447

Umang01-hash:
@vigneshp91 seems like your PR has some review comments unresolved. Please address them

pkg/gofr/datasource/sql/db.go
Umang01-hash: I have added a check for `rows.Err()` and if it is not nil returning from there after logging

pkg/gofr/datasource/sql/db.go
Umang01-hash: Also added a check here

vipul-rawat (approved):

aryanmehrotra (approved):

PR #1041, opened 2024-09-23
chore: connect mongoDB

1ch0:
## Pull Request Template

**Description:**
> connect mongoDB

- add hidePass
- add ping check
- chore log

### DIFF
--- a/pkg/gofr/datasource/mongo/mongo.go
+++ b/pkg/gofr/datasource/mongo/mongo.go
@@ -67,8 +67,6 @@ func (c *Client) UseMetrics(metrics interface{}) {
 
 // Connect establishes a connection to MongoDB and registers metrics using the provided configuration when the client was Created.
 func (c *Client) Connect() {
-	c.logger.Logf("connecting to mongoDB at %v to database %v", c.config.URI, c.config.Database)
-
 	uri := c.config.URI
 
 	if uri == "" {
@@ -79,7 +77,11 @@ func (c *Client) Connect() {
 	m, err := mongo.Connect(context.Background(), options.Client().ApplyURI(uri))
 	if err != nil {
 		c.logger.Errorf("error connecting to mongoDB, err:%v", err)
+		return
+	}
 
+	if err = m.Ping(context.Background(), nil); err != nil {
+		c.logger.Errorf("error connecting mongoDB %s, err:%v", c.config.Host, err)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -67,8 +68,6 @@ func (c *Client) UseMetrics(metrics interface{}) {
 
 // Connect establishes a connection to MongoDB and registers metrics using the provided configuration when the client was Created.
 func (c *Client) Connect() {
-	c.logger.Logf("connecting to mongoDB at %v to database %v", c.config.URI, c.config.Database)
```
Umang01-hash, line 70: Since we have already added the comment to deprecate the URI method to connect to MongoDB , instead we will use the HOST, PORT and PASSWORD fields.

```
// Deprecated Provide Host User Password Port Instead and driver will generate the URI
```

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -79,14 +78,24 @@ func (c *Client) Connect() {
 	m, err := mongo.Connect(context.Background(), options.Client().ApplyURI(uri))
 	if err != nil {
 		c.logger.Errorf("error connecting to mongoDB, err:%v", err)
-
+		return
+	}
+	err = m.Ping(context.Background(), nil)
```
Umang01-hash, line 83: Please leave a line space here

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -79,14 +78,24 @@ func (c *Client) Connect() {
...
+	if err != nil {
+		c.logger.Errorf("error connecting mongoDB %s, err:%v", hidePass(uri), err)
 		return
 	}
 
 	mongoBuckets := []float64{.05, .075, .1, .125, .15, .2, .3, .5, .75, 1, 2, 3, 4, 5, 7.5, 10}
 	c.metrics.NewHistogram("app_mongo_stats", "Response time of MONGO queries in milliseconds.", mongoBuckets...)
 
 	c.Database = m.Database(c.config.Database)
+	c.logger.Logf("connecting to mongoDB at %v to database %v", hidePass(uri), c.config.Database)
```
Umang01-hash, line 93: We can only mention the host and port while connecting no need to show other information

Umang01-hash (changes requested):

Umang01-hash:
@1ch0 We don't need to show up password and user while logging to connect to Mongo. Also we don't intend to support the `URI` input directly from user so this regex parsing is not needed.

Please correct me if i am understanding anything wrong.
Thankyou

PR #1044, opened 2024-09-23
Making Traces configurable to improve performance

coolwednesday:
**Description:**

-   Idea is to make traces configurable to avoid unnecessary overhead when configs of traces are not provided by the user.
-   This solves the performance of GoFr while handling such request.
-   This PR also introduces sampling to allow user to set the level of tracing they require through the configs.
-    For requests with tracing disabled, correlation id is generated through `github.com/google/uuid`

**Checklist:**

### DIFF
--- a/pkg/gofr/http/middleware/logger.go
+++ b/pkg/gofr/http/middleware/logger.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
+
 	"go.opentelemetry.io/otel/trace"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/gofr/http/middleware/logger.go
```diff
@@ -65,38 +69,39 @@ type logger interface {
 	Error(...interface{})
 }
 
+func getIDs(requestCtx context.Context) (ctx context.Context, correlationID, spanID string) {
+	requestSpan := trace.SpanFromContext(requestCtx).SpanContext()
+	hasTraceID := requestSpan.HasTraceID()
+
+	if hasTraceID {
```
vipul-rawat, line 76:
```suggestion
	if requestSpan.HasTraceID() {
	}
```

pkg/gofr/http/middleware/logger.go
```diff
@@ -107,6 +112,46 @@ func Logging(logger logger) func(inner http.Handler) http.Handler {
...
 
+func getCorrelationAndSpanID(request *http.Request) (r *http.Request, correlationID, spanID string) {
+	correlationID = request.Header.Get("X-Correlation-ID")
+
+	var ctx context.Context
+
+	if correlationID == "" {
+		ctx, correlationID, spanID = getIDs(request.Context())
+		// returning deep copy of request with added context
+		request = request.Clone(ctx)
```
aryanmehrotra, line 123: can we put the context in the request? Clone would be a complex operation - might not give the desired perforamnce results

aryanmehrotra (changes requested):
1. Sampling should be part of a different PR

PR #1052, opened 2024-09-25
En/mongo tracing

Umang01-hash:
## Pull Request Template

**Description:**

-   Add tracing support for MongoDB datasource.

Screenshot 2024-09-25 at 3 13 35 PM

**Checklist:**


pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -101,28 +102,36 @@ func (c *Client) Connect() {
 
 // InsertOne inserts a single document into the specified collection.
 func (c *Client) InsertOne(ctx context.Context, collection string, document interface{}) (interface{}, error) {
-	defer c.sendOperationStats(&QueryLog{Query: "insertOne", Collection: collection, Filter: document}, time.Now())
+	tracedCtx, span := c.addTrace(ctx, "insert", collection)
```
srijan-27, line 105: I think we should not rename `insertOne` to `insert` as it would cause parity with other places as every other place we are still using `One` as suffix.

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -101,28 +102,36 @@ func (c *Client) Connect() {
...
 
-	return c.Database.Collection(collection).InsertOne(ctx, document)
+	result, err := c.Database.Collection(collection).InsertOne(tracedCtx, document)
+
+	defer c.pushObservability(&QueryLog{Query: "insertOne", Collection: collection, Filter: document}, time.Now(),
+		"insert", span)
```
srijan-27, line 110: not necessary, but IMO we should not move `defer pushObservability` from beginning to end..
if updating, please update everywhere..

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -101,28 +102,36 @@ func (c *Client) Connect() {
...
+
+	return result, err
 }
 
 // InsertMany inserts multiple documents into the specified collection.
 func (c *Client) InsertMany(ctx context.Context, collection string, documents []interface{}) ([]interface{}, error) {
-	defer c.sendOperationStats(&QueryLog{Query: "insertMany", Collection: collection, Filter: documents}, time.Now())
+	tracedCtx, span := c.addTrace(ctx, "insertMany", collection)
```
srijan-27, line 117: wouldn't `tracerCtx` be better than `tracedCtx`?

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -133,94 +142,129 @@ func (c *Client) Find(ctx context.Context, collection string, filter, results in
...
+
+	defer c.pushObservability(&QueryLog{Query: "createCollection", Collection: name}, time.Now(), "createCollection",
+		span)
 
-	return c.Database.CreateCollection(ctx, name)
+	return err
 }
 
-func (c *Client) sendOperationStats(ql *QueryLog, startTime time.Time) {
+func (c *Client) pushObservability(ql *QueryLog, startTime time.Time, method string, span trace.Span) {
```
srijan-27, line 267: this method name is a convention followed by GoFr and being used everywhere. If renaming the name, please take note that we would need to update this everywhere later...

pkg/gofr/datasource/mongo/mongo_test.go
```diff
@@ -79,7 +80,7 @@ func Test_InsertCommands(t *testing.T) {
 		resp, err := cl.InsertOne(context.Background(), mt.Coll.Name(), doc)
 
 		assert.NotNil(t, resp)
-		assert.Nil(t, err)
+		assert.NoError(t, err)
```
srijan-27, line 83: just curious, were the tests failing everywhere with the latest changes? or is this just changed as part of code quality?

pkg/gofr/datasource/mongo/mongo_test.go
```diff
@@ -514,6 +516,7 @@ func TestClient_StartSession(t *testing.T) {
 		// Call the StartSession method
 		sess, err := cl.StartSession()
 		ses, ok := sess.(Transaction)
+
```
srijan-27, line 519: IMO, if block should be clubbed with the variable declaration.. (in this case  var`ok`)

pkg/gofr/external_db_test.go
```diff
@@ -40,6 +40,7 @@ func TestApp_AddMongo(t *testing.T) {
 
 		mock.EXPECT().UseLogger(app.Logger())
 		mock.EXPECT().UseMetrics(app.Metrics())
+		mock.EXPECT().UseTracer(gomock.Any())
```
srijan-27, line 43: please correct me if I am wrong.. but I assume `app.Tracer()` would be better instead of `gomock.Any()` just like we are using Logger and Metrics.

srijan-27 (changes requested):

pkg/gofr/datasource/mongo/mongo.go
Umang01-hash: sure will use `insert` only

pkg/gofr/datasource/mongo/mongo.go
Umang01-hash: I have updated it everywhere.

pkg/gofr/external_db_test.go
Umang01-hash: Currently `app` doesn't have a method called `Tracer`.  Although it has `initTracer` but it serves a different purpose

pkg/gofr/datasource/mongo/mongo_test.go
Umang01-hash: sure done

pkg/gofr/datasource/mongo/mongo.go
Umang01-hash: renamed back to `sendOperationStats `

aryanmehrotra (approved):

pkg/gofr/datasource/mongo/mongo.go
srijan-27: okay, but won't the method name `InsertOne` differ from the `method` being shown in tracer. Also, in-case of `FindOne`, `UpdateOne`, `DeleteOne` we are still using the same, so why only change for `Insert`..

Umang01-hash: Ohh sorry my bad, I wanted to change the name of span to `InsertOne` only. You are right we should use the same naming conventions as we are using for other methods having `One` as suffix.

Changed the span name to `insertOne`.

srijan-27 (approved):

vipul-rawat (approved):

Issue #991, opened 2024-09-06
Incorrect overview for container package


aryanmehrotra:
Incorrect overview for container package
<img width="925" alt="Screenshot 2024-09-06 at 11 38 52 PM" src="https://github.com/user-attachments/assets/b2c33f39-4044-4c4f-b56b-9fcb8ffa59a0">

aryanmehrotra:
Hey @JonasBakys0 we would to make it more suitable.

JonasBakys0:
Cool, can I try it? Can you assign me to this issue?

PR #1065, opened 2024-10-01
doc: Update container package overview

JonasBakys0:
**Description:**

- Updates incorrect overview for container package
- [issue](https://github.com/gofr-dev/gofr/issues/991)

**Checklist:**

### DIFF
--- a/pkg/gofr/container/container.go
+++ b/pkg/gofr/container/container.go
@@ -1,3 +1,6 @@
+// The container package provides a centralized structure to manage common application-level concerns such as
+// logging, connection pools, and service management. This package is designed to facilitate the sharing and
+// management of these concerns across different parts of an application.
 package container
 
 import (
--- a/pkg/gofr/container/container.go
+++ b/pkg/gofr/container/container.go
@@ -1,6 +1,15 @@
-// The container package provides a centralized structure to manage common application-level concerns such as
-// logging, connection pools, and service management. This package is designed to facilitate the sharing and
-// management of these concerns across different parts of an application.
+/*
+Package container provides a centralized structure to manage common application-level concerns such as
+logging, connection pools, and service management. This package is designed to facilitate the sharing and
+management of these concerns across different parts of an application.
+
+Supported data sources:
+  - Databases (Cassandra, ClickHouse, MongoDB, DGraph, MySQL, PostgreSQL, SQLite)
+  - Key-value storages (Redis, BadgerDB)
+  - Pub/Sub systems (Azure Event Hub, Google as backend, Kafka, MQTT)
+  - Search engines (Solr)
+  - File systems (FTP, SFTP, S3)
+*/
 package container
 
 import (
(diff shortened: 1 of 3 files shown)


pkg/gofr/container/container.go
```diff
@@ -1,3 +1,6 @@
+// The container package provides a centralized structure to manage common application-level concerns such as
+// logging, connection pools, and service management. This package is designed to facilitate the sharing and
```
aryanmehrotra, line 2: can you also include data sources which comprises of different databases, pubsub, filesystem etc.

aryanmehrotra (changes requested):

pkg/gofr/container/container.go
JonasBakys0: Sure. Also, I have a question about documentation in go. I saw that `package container` is used in many files which made me wonder what is the best way to document such package? How do I choose which file to add documentation?

aryanmehrotra: `container.go` is the central package which also has the container struct. So it is the best file.

JonasBakys0: I wanted to add some references but couldn't figure out how exactly it works in godoc 😄 
```
Supported data sources:
  - Databases ([cassandra], [clickhouse], [mongo], [solr], [dgraph], [sql])
  - Key value storages [redis] [kafka]
  - Pub/Sub systems ([google], [kafka], [mqtt])
  - File system [file]
```
Results:
image

Even more links to other packages generate false(not found) URL at least locally
Can you help me with this?

aryanmehrotra: You may refer to the following documentation 
https://tip.golang.org/doc/comment#links

JonasBakys0:
I dropped the link idea because it seemed too complicated. For example:
- If I want to use the doc link I face this issue:
image
- And with Links, it's hard to maintain them. What if cassandra package location change
```
// [Cassandra]: https://pkg.go.dev/gofr.dev/pkg/gofr/datasource/cassandra
```

Either way, please correct me if I'm wrong. Go is new to me and don't know how idiomatic documentation should look like in go 😄

aryanmehrotra (dismissed):

pkg/gofr/container/container.go
```diff
@@ -47,6 +59,8 @@ type Container struct {
 	Solr       Solr
 	DGraph     Dgraph
 
+	Dito cassandra
```
coolwednesday, line 62: What is this for ?

coolwednesday (changes requested):

Umang01-hash:
@JonasBakys0 the tests and the pipeline is failing due to these errors:
Screenshot 2024-10-03 at 3 07 34 PM


Please fix them.

pkg/gofr/container/container.go
JonasBakys0: Good catch, some testing code left.

Umang01-hash (approved):

aryanmehrotra (approved):

Issue #989, opened 2024-09-06
Create DOCKERFILE for examples - with observability setup


aryanmehrotra:
Currently, DOCKERFILE are not present for examples - observability is needed to be setup manually.

Docker files needs to be added with observability setup.

vigneshp91:
Do you want to add the docker setup for observability in 
/examples/http-server folder?  or in any specific example @aryanmehrotra

aryanmehrotra:
It would be similar for all of those - but let's first add for http-server

vigneshp91:
@aryanmehrotra Can you assign this to me?

PR #1079, opened 2024-10-04
issue 989 added docker-compose for the http-server example with obser…

vigneshp91:
Added docker compose file which run mysql,redis and jaeger and the sample application in docker containers.
Fixes [#989](https://github.com/gofr-dev/gofr/issues/989)


aryanmehrotra (changes requested):
Can you also run grafana with the dashboard from : https://grafana.com/grafana/dashboards/19905-gofr-application-services-monitoring/

Moreover, we can use gofr tracer instead of jaeger as it is a public api as well.

vigneshp91:
> Can you also run grafana with the dashboard from : https://grafana.com/grafana/dashboards/19905-gofr-application-services-monitoring/
> 
> Moreover, we can use gofr tracer instead of jaeger as it is a public api as well.

I can remove Jaeger but
if we are including Grafana then there will be too many setups has to be done like import dashboard, add datasource from prometheus it will make the docker file complex

aryanmehrotra:
@vigneshp91 yes, but we would want to keep in at least one example so that it becomes reference point for new users to setup

vigneshp91:
@aryanmehrotra I have added the required changes. But the  PKG Unit Testing (v1.21) stage is failing. Not sure because of my change

examples/http-server/Dockerfile
```diff
@@ -1,15 +1,16 @@
...
+# Build stage
+FROM golang:1.22 AS build
 
-RUN mkdir /src/
-WORKDIR /src/
+WORKDIR /src
 COPY . .
 RUN go get ./...
-RUN go build -ldflags "-linkmode external -extldflags -static" -a main.go
+RUN go build -ldflags "-linkmode external -extldflags -static" -a -o /app/main main.go
```
aryanmehrotra, line 7: Why not build the image in local-system and just copy the build file and the configs - which would reduce the size of the container.

i.e we can use alpine container?

examples/http-server/docker-compose.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - DB_PORT=3306
+      - DB_DIALECT=mysql
+    ports:
+      - "9000:9000"
+      - "2121:2121"
+    depends_on:
+      - redisdb
+      - mysqldb
+      - grafana
+      - prometheus
```
aryanmehrotra, line 26: were you able to see the metrics in the prometheus or grafana dashboard?
as it is not appearing for me

aryanmehrotra (changes requested):

examples/http-server/Dockerfile
aryanmehrotra: Lets discuss the approach on this first before actually doing it

examples/http-server/docker-compose.yaml
vigneshp91: its visible in grafana dashboard
Screenshot 2024-10-09 at 8 40 23 PM
Screenshot 2024-10-09 at 8 40 57 PM
Screenshot 2024-10-09 at 8 40 57 PM

examples/http-server/Dockerfile
vigneshp91: @aryanmehrotra IMO Building the app outside the docker is also a good approach which will reduce the docker build time and simplify the docker file.
Building the app inside the docker also have the advantage that we don't need to depend on the host machine for the dependencies. users can simply install docker and they can start the app.
I'm ok with both the approaches

examples/http-server/docker-compose.yaml
aryanmehrotra: my bad, I was trying to test in GitHub code space, probably it had some port forwarding issue.
It works fine in local

examples/http-server/Dockerfile
aryanmehrotra: seems fine, lets build in the docker itself if needed will change in future

examples/http-server/README.md
```diff
@@ -4,17 +4,13 @@ This GoFr example demonstrates a simple HTTP server which supports Redis and MyS
...
 ```
 
-- Run the docker image of MySQL
-```console
-docker run --name gofr-mysql -e MYSQL_ROOT_PASSWORD=password -e MYSQL_DATABASE=test -p 2001:3306 -d mysql:8.0.30
-```
+To test the example, follow these steps:
+
+1. Open your browser and navigate to `http://localhost:9000/hello`.
+2. To view the trace in Jaeger, open `http://localhost:16686`.
```
aryanmehrotra, line 15: now it won't be visible in jaeger we can update the link in message to view on the gofr tracer website

https://tracer.gofr.dev

aryanmehrotra (changes requested):
1. Can we move prometheus and provisioning in one directory may be keep docker-compose.yml as well there.

It would be clean that way

once this is fixed we can merge this PR.

vigneshp91:
@aryanmehrotra Can we merge this?

aryanmehrotra (approved):

Umang01-hash (approved):

vipul-rawat (approved):

PR #1092, opened 2024-10-09
Adding support for OpentsDB

### LINKED ISSUE #962: Support for Opentsdb

coolwednesday:
**Description:**

-   This PR introduces opentsdb and its various rest api responses support in GoFr.
-   Closes #962 
-   Currently supporting rest apis endpoints as follows :
```
GET 			           /api/aggregators
GET,POST,PUT,DELETE               /api/annotation
POST 			           /api/put
GET 			           /api/query
GET 			           /api/query/last
GET 			           /api/version
```

**Checklist:**


ccoVeille (commented):
I reviewed quickly, I will need other pass, but at least here are things that could be addressed

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,353 @@
...
+	if c.opentsdbCfg.DetectDeltaNum <= 0 {
+		c.opentsdbCfg.DetectDeltaNum = DefaultDetectDeltaNum
+	}
+
+	if c.opentsdbCfg.MaxContentLength <= 0 {
+		c.opentsdbCfg.MaxContentLength = DefaultMaxContentLength
+	}
+
+	// Initialize the OpenTSDB client with the given configuration.
+	c.tsdbEndpoint = fmt.Sprintf("http://%s", c.opentsdbCfg.OpentsdbHost)
```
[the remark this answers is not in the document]
coolwednesday: According to the OpenTSDB REST API documentation, it uses a basic HTTP-based API. All available examples also reference HTTP rather than HTTPS, which suggests that HTTPS is not currently supported. This is likely because OpenTSDB was originally designed for internal use, where the need for encrypted communication is less critical.

pkg/gofr/datasource/opentsdb/annotation.go
```diff
@@ -0,0 +1,334 @@
...
+		func(resp []byte) error {
+			originContents := string(resp)
+
+			var resultBytes []byte
+
+			if strings.Contains(originContents, "error") || strings.Contains(originContents, "totalDeleted") {
+				resultBytes = resp
+			} else if strings.Contains(originContents, "startTime") {
+				resultBytes = []byte(fmt.Sprintf("{%s:%s}", `"InvolvedAnnotations"`, originContents))
+			} else {
```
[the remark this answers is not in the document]
coolwednesday: > I would decode in a struct with the field you need, then apply the logic of checking the fields

Here it gives error when error or totalDeleted is returned, in case of DeleteAnnotation ....while a struct starting with startTime or an error is returned in case of UpdateAnnotation. I tried generalizing the custom-parser for both of the operations.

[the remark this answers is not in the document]
ccoVeille: First, please bare with me while I have zero knowledge with OpenTSDB

But, imagine the code returns a valid reply that contains the "error" string, not as a JSON node, but somewhere in the reply.

Your code will fail.

pkg/gofr/datasource/opentsdb/dropcache.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"go.opentelemetry.io/otel/trace"
+)
+
+type DropcachesResponse struct {
```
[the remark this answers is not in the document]
ccoVeille: You resolved but didn't change the code

ccoVeille:
Just to make sure, you marked things as resolved, but you haven't pushed changes yet, right?

coolwednesday:
> Just to make sure, you marked things as resolved, but you haven't pushed changes yet, right?

yes I am yet to resolve few other changes. Will push it all together.

pkg/gofr/datasource/opentsdb/annotation.go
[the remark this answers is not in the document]
coolwednesday: @coolwednesday I feel the same, would test with a few more cases and check.

pkg/gofr/datasource/opentsdb/annotation.go
```diff
@@ -0,0 +1,334 @@
...
+
+	for k, v := range queryAnnoParam {
+		fmt.Fprintf(buffer, "%s=%v", k, v)
+
+		if i < size-1 {
+			buffer.WriteString("&")
+		}
+
+		i++
+	}
```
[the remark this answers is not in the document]
coolwednesday: > I'm a bit worried about how things could need to be escaped
> 
> Like an equal sign or an ampersand.
> 

Can you explain more ? .....I am not able to get you. I understand the url.Values part, but its just adding more complexity. I can add a comment saying that this is to prepare the url something of that sort, but I feel that is also not needed.

pkg/gofr/container/health_test.go
```diff
@@ -79,6 +79,12 @@ func TestContainer_Health(t *testing.T) {
...
 				},
 			},
+			"opentsdb": datasource.Health{
+				Status: tc.datasourceHealth, Details: map[string]interface{}{
+					"host":  "localhost:8000",
+					"error": "opentsdb not connected",
+				},
+			},
 			"test-service": &service.Health{
 				Status: "UP", Details: map[string]interface{}{
```
ccoVeille, line 89:
```suggestion
				Status: "UP", Details: map[string]any{
```

pkg/gofr/datasource/opentsdb/annotation.go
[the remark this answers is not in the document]
ccoVeille: My problem is the fact the following characters would lead to oddities

& ? #

If you are manipulating things that are like URL parameters, you should use url.Values

But once again, I'm unsure what you are trying to achieve here, or what are the annotations about

pkg/gofr/datasource/opentsdb/annotation.go
```diff
@@ -0,0 +1,329 @@
...
+
+			if strings.Contains(tag, "error") || strings.Contains(tag, "totalDeleted") {
+				resultBytes = resp
+			} else if strings.Contains(tag, "tsuid") {
+				resultBytes = []byte(fmt.Sprintf(`{"InvolvedAnnotations":%s}`, originContents))
+			} else {
+				return fmt.Errorf("unrecognized bulk annotation response: %s", originContents)
+			}
+
+			return json.Unmarshal(resultBytes, &bulkAnnotResp)
```
ccoVeille, line 246: This way to parse is an hack.

I would recommend using something better than this.

I will try to suggest you something when I will have a computer. Only a phone for now

Are these covered by a unit tests?

pkg/gofr/datasource/opentsdb/annotation.go
```diff
@@ -0,0 +1,329 @@
...
+	bulkAnnoResp := BulkAnnotationResponse{logger: c.logger, tracer: c.tracer, ctx: c.ctx}
+	if err = c.sendRequest(http.MethodPost, bulkAnnoEndpoint, reqBodyCnt, &bulkAnnoResp); err != nil {
+		message = fmt.Sprintf("error while processing update bulk annotations request to url %q: %s", bulkAnnoEndpoint, err)
+		return nil, err
+	}
+
+	status = StatusSuccess
+	message = fmt.Sprintf("Bulk annotation updation processed successfully at url %q", bulkAnnoEndpoint)
+
+	c.logger.Logf("Bulk annotation updated successfully")
```
ccoVeille, line 285: Why not adding Log to the Logger interface?

I would use this here

```suggestion
	c.logger.Log("Bulk annotation updated successfully")
```

pkg/gofr/datasource/opentsdb/annotation.go
```diff
@@ -0,0 +1,329 @@
...
+	bulkAnnoResp := BulkAnnotationResponse{logger: c.logger, tracer: c.tracer, ctx: c.ctx}
+	if err = c.sendRequest(http.MethodDelete, bulkAnnoEndpoint, string(resultBytes), &bulkAnnoResp); err != nil {
+		message = fmt.Sprintf("Bulk annotation delete request failed at url %q: %v", bulkAnnoEndpoint, err)
+		return nil, err
+	}
+
+	status = StatusSuccess
+	message = fmt.Sprintf("Bulk annotations deleted successfully at url %q", bulkAnnoEndpoint)
+
+	c.logger.Logf("Bulk annotation deleted successfully")
```
ccoVeille, line 316:
```suggestion
	c.logger.Log("Bulk annotation deleted successfully")
```

pkg/gofr/datasource/opentsdb/annotation.go
```diff
@@ -0,0 +1,329 @@
...
+
+	return &bulkAnnoResp, nil
+}
+
+// marshalAnnotations converts a slice of annotations into a JSON string for bulk operations.
+func marshalAnnotations(annotations []Annotation) (string, error) {
+	resultBytes, err := json.Marshal(annotations)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal annotations: %v", err)
```
ccoVeille, line 325: Why not wrapping?

```suggestion
		return "", fmt.Errorf("failed to marshal annotations: %w", err)
```

pkg/gofr/datasource/opentsdb/dropcache.go
```diff
@@ -0,0 +1,60 @@
...
+func (dropResp *DropCachesResponse) setStatusCode(code int) {
+	dropResp.StatusCode = code
+}
+
+func (dropResp *DropCachesResponse) GetCustomParser() func(respCnt []byte) error {
+	return getCustomParser(dropResp.ctx, dropResp, "GetCustomParser-DropCaches", dropResp.logger,
+		func(resp []byte) error {
+			return json.Unmarshal([]byte(fmt.Sprintf(`{"DropcachesInfo":%s}`, string(resp))), &dropResp)
+		})
+}
```
ccoVeille, line 33: I'm not convinced by this pattern

Please consider this

```suggestion
func (dropResp *DropCachesResponse) GetCustomParser() func(respCnt []byte) error {
	return getCustomParser(dropResp.ctx, dropResp, "GetCustomParser-DropCaches", dropResp.logger,
		func(resp []byte) error {
			var j map[string]string
			err := json.Unmarshal(resp, &j)
			if err != nil {
				return err
			}
			dropResp.DropCachesInfo = j
			return nil
		})
}
```

pkg/gofr/datasource/opentsdb/interface.go
```diff
@@ -0,0 +1,57 @@
...
+	// String returns the contents of the specific Response instance with
+	// the string format
+	String() string
+}
+
+// Logger interface is used by opentsdb package to log information about request execution.
+type Logger interface {
+	Debug(args ...interface{})
+	Logf(pattern string, args ...interface{})
+	Errorf(pattern string, args ...interface{})
```
ccoVeille, line 50:
```suggestion
	Debug(args ...any)
	Logf(pattern string, args ...any)
	Errorf(pattern string, args ...any)
```

pkg/gofr/datasource/opentsdb/observability.go
```diff
@@ -0,0 +1,134 @@
...
+func (aggreResp *AggregatorsResponse) addTrace(ctx context.Context, operation string) trace.Span {
+	return addTracer(ctx, aggreResp.tracer, operation, "AggregatorRes")
+}
+
+func (annotResp *AnnotationResponse) addTrace(ctx context.Context, operation string) trace.Span {
+	return addTracer(ctx, annotResp.tracer, operation, "AnnotationRes")
+}
+
+func (bulkAnnotResp *BulkAnnotationResponse) addTrace(ctx context.Context, operation string) trace.Span {
+	return addTracer(ctx, bulkAnnotResp.tracer, operation, "BulkAnnotatResponse")
```
ccoVeille, line 81:
```suggestion
	return addTracer(ctx, bulkAnnotResp.tracer, operation, "BulkAnnotationResponse")
```

pkg/gofr/datasource/opentsdb/observability.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (queryResp *QueryResponse) addTrace(ctx context.Context, operation string) trace.Span {
+	return addTracer(ctx, queryResp.tracer, operation, "QueryResponse")
+}
+
+func (dropResp *DropCachesResponse) addTrace(ctx context.Context, operation string) trace.Span {
+	return addTracer(ctx, dropResp.tracer, operation, "DropcacheResponse")
```
ccoVeille, line 89:
```suggestion
	return addTracer(ctx, dropResp.tracer, operation, "DropCachesResponse")
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,353 @@
...
+
+const (
+	StatusFailed      = "FAIL"
+	StatusSuccess     = "SUCCESS"
+	DefaultDialTime   = 5 * time.Second  // Default time for establishing TCP connections.
+	ConnectionTimeout = 30 * time.Second // Timeout for keeping connections alive.
+	GetMethod         = "GET"            // HTTP GET method.
+	PostMethod        = "POST"           // HTTP POST method.
+	PutMethod         = "PUT"            // HTTP PUT method.
+	DeleteMethod      = "DELETE"         // HTTP DELETE method.
```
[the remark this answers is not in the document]
ccoVeille: Why didn't you remove them then?

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,353 @@
...
+	logger       Logger
+	metrics      Metrics
+	tracer       trace.Tracer
+}
+
+type Config struct {
+
+	// The host of the target opentsdb, is a required non-empty string which is
+	// in the format of ip:port without http:// prefix or a domain.
+	OpentsdbHost string
```
ccoVeille, line 85:
```suggestion
	OpenTSDBHost string
```

or

```suggestion
	Host string
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,353 @@
...
+		tsdbEndpoint: c.tsdbEndpoint,
+		client:       c.client,
+		ctx:          ctx,
+		opentsdbCfg:  c.opentsdbCfg,
+	}
+}
+
+type Health struct {
+	Status  string                 `json:"status,omitempty"`
+	Details map[string]interface{} `json:"details,omitempty"`
```
ccoVeille, line 208:
```suggestion
	Details map[string]any `json:"details,omitempty"`
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,353 @@
...
+	span := c.addTrace(c.ctx, "HealthCheck")
+
+	status := StatusFailed
+
+	var message string
+
+	defer sendOperationStats(c.logger, time.Now(), "HealthCheck", &status, &message, span)
+
+	h := Health{
+		Details: make(map[string]interface{}),
```
ccoVeille, line 221:
```suggestion
		Details: make(map[string]any),
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,353 @@
...
+// Client is the implementation of the OpentsDBClient interface,
+// which includes context-aware functionality.
+type Client struct {
+	tsdbEndpoint string
+	client       HTTPClient
+	ctx          context.Context
+	opentsdbCfg  Config
+	logger       Logger
+	metrics      Metrics
+	tracer       trace.Tracer
```
ccoVeille, line 78:
```suggestion
	endpoint     string
	client       HTTPClient
	ctx          context.Context
	config       Config
	logger       Logger
	metrics      Metrics
	tracer       trace.Tracer
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,353 @@
...
+	defer sendOperationStats(c.logger, time.Now(), "sendRequest", &status, &message, span)
+
+	// Create the HTTP request, attaching the context if available.
+	req, err := http.NewRequest(method, url, strings.NewReader(reqBodyCnt))
+	if c.ctx != nil {
+		req = req.WithContext(c.ctx)
+	}
+
+	if err != nil {
+		errRequestCreation := fmt.Errorf("failed to create request for %s %s: %v", method, url, err)
```
ccoVeille, line 271:
```suggestion
		errRequestCreation := fmt.Errorf("failed to create request for %s %s: %w", method, url, err)
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,353 @@
...
+		return errRequestCreation
+	}
+
+	// Set the request headers.
+	req.Header.Set("Content-Type", "application/json; charset=UTF-8")
+
+	// Send the request and handle the response.
+	resp, err := c.client.Do(req)
+	if err != nil {
+		errSendingRequest := fmt.Errorf("failed to send request for %s %s: %v", method, url, err)
```
ccoVeille, line 284:
```suggestion
		errSendingRequest := fmt.Errorf("failed to send request for %s %s: %w", method, url, err)
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,353 @@
...
+
+		return errSendingRequest
+	}
+
+	defer resp.Body.Close()
+
+	// Read and parse the response.
+	jsonBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		errReading := fmt.Errorf("failed to read response body for %s %s: %v", method, url, err)
```
ccoVeille, line 296:
```suggestion
		errReading := fmt.Errorf("failed to read response body for %s %s: %w", method, url, err)
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,353 @@
...
+		return errReading
+	}
+
+	parsedResp.SetStatus(resp.StatusCode)
+
+	parser := parsedResp.GetCustomParser()
+	if parser == nil {
+		// Use the default JSON unmarshaller if no custom parser is provided.
+		if err := json.Unmarshal(jsonBytes, parsedResp); err != nil {
+			errUnmarshaling := fmt.Errorf("failed to unmarshal response body for %s %s: %v", method, url, err)
```
ccoVeille, line 309:
```suggestion
			errUnmarshaling := fmt.Errorf("failed to unmarshal response body for %s %s: %w", method, url, err)
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,353 @@
...
+
+			message = fmt.Sprint(errUnmarshaling)
+
+			return errUnmarshaling
+		}
+	} else {
+		// Use the custom parser if available.
+		if err := parser(jsonBytes); err != nil {
+			message = fmt.Sprintf("failed to parse response body through custom parser %s %s: %v", method, url, err)
```
ccoVeille, line 318:
```suggestion
			message = fmt.Sprintf("failed to parse response body through custom parser %s %s: %w", method, url, err)
```

pkg/gofr/datasource/opentsdb/put.go
```diff
@@ -0,0 +1,354 @@
...
+	Metric string `json:"metric"`
+
+	// A Unix epoch style timestamp in seconds or milliseconds.
+	// The timestamp must not contain non-numeric characters.
+	// One can use time.Now().Unix() to set this attribute.
+	// This attribute is also required with non-zero value.
+	Timestamp int64 `json:"timestamp"`
+
+	// The real type of Value only could be int, int64, float64, or string, and is required.
+	Value interface{} `json:"value"`
```
ccoVeille, line 29:
```suggestion
	Value any `json:"value"`
```

pkg/gofr/datasource/opentsdb/put.go
```diff
@@ -0,0 +1,354 @@
...
+
+	status := StatusFailed
+
+	var message string
+
+	defer sendOperationStats(c.logger, time.Now(), "splitProperGroups-Put", &status, &message, span)
+
+	datasBytes, err := json.Marshal(&datapoints)
+	if err != nil {
+		message = fmt.Sprintf("failed to marshal the datapoints to be put: %v", err)
```
ccoVeille, line 184:
```suggestion
		message = fmt.Sprintf("failed to marshal the datapoints to be put: %w", err)
```

pkg/gofr/datasource/opentsdb/put.go
```diff
@@ -0,0 +1,354 @@
...
+
+		if index < size-1 {
+			buffer.WriteString(",")
+		}
+	}
+
+	buffer.WriteString("]")
+
+	return buffer.String(), nil
+}
```
ccoVeille, line 302: I don't get why you are not using

json.Marshall(datas)

A list of DataPoint will be serialized the way you expect no?

Unless, if you have a nil list, but here you should handle a nil and assign an empty list

pkg/gofr/datasource/opentsdb/query.go
```diff
@@ -0,0 +1,414 @@
...
+	// The start time for the query. This can be a relative or absolute timestamp.
+	// The data type can only be string, int, or int64.
+	// The value is required with non-zero value of the target type.
+	Start interface{} `json:"start"`
+
+	// An end time for the query. If not supplied, the TSD will assume the local
+	// system time on the server. This may be a relative or absolute timestamp.
+	// The data type can only be string, or int64.
+	// The value is optional.
+	End interface{} `json:"end,omitempty"`
```
ccoVeille, line 28:
```suggestion
	Start any `json:"start"`

	// An end time for the query. If not supplied, the TSD will assume the local
	// system time on the server. This may be a relative or absolute timestamp.
	// The data type can only be string, or int64.
	// The value is optional.
	End any `json:"end,omitempty"`
```

pkg/gofr/datasource/opentsdb/query.go
```diff
@@ -0,0 +1,414 @@
...
+	// continuously incrementing counter, and you want to view the rate of change between data points.
+	Rate bool `json:"rate,omitempty"`
+
+	// rateOptions represents monotonically increasing counter handling options.
+	// The value is optional.
+	// Currently, there is only three kind of value can be set to this map:
+	// Only three keys can be set into the rateOption parameter of the QueryParam is
+	// QueryRateOptionCounter (value type is bool),  QueryRateOptionCounterMax (value type is int,int64)
+	// QueryRateOptionResetValue (value type is int,int64)
+	RateParams map[string]interface{} `json:"rateOptions,omitempty"`
```
ccoVeille, line 100:
```suggestion
	RateParams map[string]any `json:"rateOptions,omitempty"`
```

pkg/gofr/datasource/opentsdb/query.go
```diff
@@ -0,0 +1,414 @@
...
+	// by the value as an integer or a floating point.
+	// For example, the default output is "dps"{"<timestamp>":<value>}.
+	// By default, the timestamps will be in seconds. If the msResolution flag is set, then the
+	// timestamps will be in milliseconds.
+	//
+	// Because the elements of map is out of order, using common way to iterate Dps will not get
+	// data points with timestamps out of order.
+	// So be aware that one should use '(qri *QueryRespItem) GetDataPoints() []*DataPoint' to
+	// acquire the real ascending data points.
+	Dps map[string]interface{} `json:"dps"`
```
ccoVeille, line 192:
```suggestion
	Dps map[string]any `json:"dps"`
```

pkg/gofr/datasource/opentsdb/version.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (verResp *VersionResponse) String() string {
+	return toString(verResp.ctx, verResp, "ToString-VersionResp", verResp.logger)
+}
+
+func (verResp *VersionResponse) GetCustomParser() func(respCnt []byte) error {
+	return getCustomParser(verResp.ctx, verResp, "GetCustomParser-VersionResp", verResp.logger,
+		func(resp []byte) error {
+			return json.Unmarshal([]byte(fmt.Sprintf("{%s:%s}", `"VersionInfo"`, string(resp))), &verResp)
+		})
```
ccoVeille, line 37: Same here about using a map[string]string, unmarshalling then assign the value to verResp.VersionInfo

ccoVeille (commented):
Here comes a new round of review that might need a new one later

pkg/gofr/datasource/opentsdb/put.go
coolwednesday: %w is used with fmt.Errorf() calls .....

pkg/gofr/datasource/opentsdb/opentsdb.go
coolwednesday: "%w" is used with fmt.Errorf() calls....

pkg/gofr/datasource/opentsdb/version.go
```diff
@@ -0,0 +1,62 @@
...
+func (verResp *VersionResponse) String() string {
+	return toString(verResp.ctx, verResp, "ToString-VersionResp", verResp.logger)
+}
+
+func (verResp *VersionResponse) GetCustomParser() func(respCnt []byte) error {
+	return getCustomParser(verResp.ctx, verResp, "GetCustomParser-VersionResp", verResp.logger,
+		func(resp []byte) error {
+			return json.Unmarshal([]byte(fmt.Sprintf(`{"VersionInfo":%s}`, string(resp))), &verResp)
+		})
+}
```
ccoVeille, line 39: I who have expected the same logic to use a map[string]any, use unmarshal then set VersionInfo instead of this

Maybe there are still other remaining in the code. Could you check?

pkg/gofr/container/datasources.go
```diff
@@ -395,3 +395,276 @@ type DgraphProvider interface {
...
+type OpenTSDB interface {
+
+	// HealthChecker checks if the target OpenTSDB server is reachable.
+	// It returns an error if the server is unreachable, otherwise returns nil.
+	HealthChecker
+
+	// Put handles the 'POST /api/put' endpoint, allowing the storage of data in OpenTSDB.
+	//
+	// Parameters:
+	// - data: A slice of DataPoint objects, which must contain at least one instance.
```
ccoVeille, line 416: You added the context as parameter

Maybe you should update the documentation

pkg/gofr/container/datasources.go
```diff
@@ -395,3 +395,276 @@ type DgraphProvider interface {
...
+	Suggest(ctx context.Context, sugParm any) (any, error)
+
+	// Dropcaches is the implementation of 'GET /api/dropcaches' endpoint.
+	// It purges the in-memory data cached in OpenTSDB. This includes all UID to name
+	// and name to UID maps for metrics, tag names and tag values.
+	//
+	// When query operation is successful, a pointer of DropcachesResponse will be returned with the corresponding
+	// status code and response info. Otherwise, an error instance will be returned, when it failed to parese the
+	// response, or OpenTSDB is un-connectable right now.
+	Dropcaches(ctx context.Context) (any, error)
```
ccoVeille, line 482: Shouldn't it be DropCaches here

pkg/gofr/container/datasources.go
```diff
@@ -395,3 +395,276 @@ type DgraphProvider interface {
...
+	//   - An empty string (""): Indicates no additional response details are required.
+	//
+	// Return:
+	// - On success, it returns a pointer to a PutResponse, along with the HTTP status code and relevant response information.
+	// - On failure (due to invalid parameters, response parsing errors, or OpenTSDB connectivity issues), it returns an error.
+	//
+	// Notes:
+	// - Use 'PutRespWithSummary' to receive summarized information about the data that was stored.
+	// - Use 'PutRespWithDetails' for a more comprehensive breakdown of the put operation.
+	Put(ctx context.Context, data any, queryParam string) (any, error)
```
ccoVeille, line 429: Here and everywhere

I'm surprised by the type `any` and `any` reply

The documentation is pretty clear 

http://opentsdb.net/docs/build/html/api_http/put.html

pkg/gofr/datasource/opentsdb/aggregators.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"go.opentelemetry.io/otel/trace"
+)
+
+// AggregatorsResponse acts as the implementation of Response in the /api/aggregators.
+// It holds the status code and the response values defined in the
+// [OpenTSDB Official Docs]: http://opentsdb.net/docs/build/html/api_http/aggregators.html.
+type AggregatorsResponse struct {
+	StatusCode  int
+	Aggregators []string `json:"aggregators"`
```
ccoVeille, line 18:
```suggestion
	Aggregators []string
```

pkg/gofr/datasource/opentsdb/annotation.go
```diff
@@ -0,0 +1,326 @@
...
+
+			return json.Unmarshal(resp, &annotResp)
+		})
+}
+
+func (annotResp *AnnotationResponse) String() string {
+	return toString(annotResp.ctx, annotResp, "ToString-Annotation", annotResp.logger)
+}
+
+func (c *Client) QueryAnnotation(ctx context.Context, queryAnnoParam map[string]interface{}) (*AnnotationResponse, error) {
```
ccoVeille, line 77:
```suggestion
func (c *Client) QueryAnnotation(ctx context.Context, queryAnnoParam map[string]any) (*AnnotationResponse, error) {
```

coolwednesday:
@ccoVeille , Thank you for your review. I am currently closing this PR and implement your advice along with design changes that I feel needs to be made in the code, and then reopen it.

ccoVeille:
I hope your code won't be lost, because I made a lot of reviews and suggestions, haha

pkg/gofr/container/datasources.go
ccoVeille: Usually one was to do check the interface is satisfied is to add something like this in code

```
var _ gofr.Request = YourObject{}
```

This way, you will get a typecheck error aside your implementation

coolwednesday: I am using any to prevent adding opentsdb dependency in the container package where I am defining the interface. I have changed the function definitions so that it works .....

pkg/gofr/datasource/opentsdb/go.mod
```diff
@@ -0,0 +1,19 @@
+module awesomeProject2/gofr/pkg/gofr/datasource/opentsdb
```
vipul-rawat, line 1: please remove the awsomeProject2, it should be gofr.dev

vipul-rawat (dismissed):

pkg/gofr/datasource/opentsdb/response.go
```diff
@@ -0,0 +1,439 @@
...
+
+		var respStr string
+
+		if len(resp) != 0 && resp[0] == '[' && resp[len(resp)-1] == ']' {
+			respStr = fmt.Sprintf(`{"queryRespCnts":%s}`, originRespStr)
+		} else {
+			respStr = originRespStr
+		}
+
+		return json.Unmarshal([]byte(respStr), obj)
```
ccoVeille, line 388: I dislike how you do, but if you want to keep this, you can do this at least

```suggestion
		if strings.HasPrefix(resp, "[") && strings.HasSuffix(resp, "]") {
			respStr = fmt.Sprintf(`{"queryRespCnts":%s}`, originRespStr)
		} else {
			respStr = originRespStr
		}

		return json.Unmarshal([]byte(respStr), obj)
```

I would have done something like this, instead of parsing the data like this


```suggestion

		var cnts []any
		err := json.Unmarshal([]byte(originRespStr), &cnts)
		if err != nil {
			return json.Unmarshal([]byte(originRespStr), obj)
		}
		obj. = cnts
		return nil
```

Umang01-hash: Sure changed it to this as per your suggestion:

```go
var respStr string

		if strings.HasPrefix(string(resp), "[") && strings.HasSuffix(string(resp), "]") {
			respStr = fmt.Sprintf(`{"queryRespCnts":%s}`, originRespStr)
		} else {
			respStr = originRespStr
		}

		return json.Unmarshal([]byte(respStr), obj)
```

pkg/gofr/datasource/opentsdb/preprocess.go
```diff
@@ -0,0 +1,408 @@
...
+	case string:
+		if v == "" {
+			return false
+		}
+
+	default:
+		return false
+	}
+
+	return true
```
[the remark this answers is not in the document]
Umang01-hash: yups, changed as per your suggestion

pkg/gofr/datasource/opentsdb/preprocess.go
```diff
@@ -0,0 +1,408 @@
...
+	return &putResp, nil
+}
+
+func parsePutErrorMsg(resp *PutResponse) error {
+	buf := bytes.Buffer{}
+	buf.WriteString(fmt.Sprintf("Failed to put %d datapoint(s) into opentsdb \n", resp.Failed))
+
+	if len(resp.Errors) > 0 {
+		for _, putError := range resp.Errors {
+			str, _ := json.Marshal(putError)
```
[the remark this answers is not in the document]
Umang01-hash: returned the error from here

pkg/gofr/datasource/opentsdb/preprocess.go
```diff
@@ -0,0 +1,408 @@
...
+
+// DefaultTransport defines the default HTTP transport settings,
+// including connection timeouts and idle connections.
+var DefaultTransport = &http.Transport{
+	MaxIdleConnsPerHost: 10,
+	DialContext: (&net.Dialer{
+		Timeout:   DefaultDialTime,
+		KeepAlive: ConnectionTimeout,
+	}).DialContext,
+}
```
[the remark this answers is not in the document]
Umang01-hash: sure made it unexported

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,433 @@
...
+
+	AnQueryStartTime = "start_time"
+	AnQueryTSUid     = "tsuid"
+
+	// The below three constants are used in /put.
+	DefaultMaxPutPointsNum = 75
+	DefaultDetectDeltaNum  = 3
+	// Unit is bytes, and assumes that config items of 'tsd.http.request.enable_chunked = true'
+	// and 'tsd.http.request.max_chunk = 40960' are all in the opentsdb.conf.
+	DefaultMaxContentLength = 40960
```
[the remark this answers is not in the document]
Umang01-hash: Correct, made them all unexported

pkg/gofr/datasource/opentsdb/observability.go
```diff
@@ -0,0 +1,106 @@
...
+	})
+
+	if span != nil {
+		defer span.End()
+		span.SetAttributes(attribute.Int64(fmt.Sprintf("opentsdb.%v.duration", operation), duration))
+	}
+}
+
+func addTracer(ctx context.Context, tracer trace.Tracer, operation, typeName string) trace.Span {
+	if tracer != nil {
```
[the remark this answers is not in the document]
Umang01-hash: Sure returned nil if tracer is nil

pkg/gofr/datasource/opentsdb/interface.go
```diff
@@ -0,0 +1,51 @@
...
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+// Response defines the common behaviors all the specific response for
+// different rest-apis should obey.
+// Currently, it is an abstraction used in Client.sendRequest()
+// to stored the different kinds of response contents for all the rest-apis.
+type Response interface {
+	// GetCustomParser can be used to retrieve a custom-defined parser.
```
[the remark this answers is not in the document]
Umang01-hash: Sure changed the comment

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,433 @@
...
+		status = StatusSuccess
+		message = fmt.Sprintf("Put request to url %q processed successfully", putEndpoint)
+		putResp.Success = tempResp.Success
+		putResp.Failed = tempResp.Failed
+		putResp.Errors = tempResp.Errors
+
+		return nil
+	}
+
+	return parsePutErrorMsg(tempResp)
```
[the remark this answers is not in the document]
Umang01-hash: right, set the message's default value to `message := "Put request failed"`

pkg/gofr/datasource/opentsdb/preprocess.go
```diff
@@ -0,0 +1,408 @@
...
+
+// PutResponse acts as the implementation of Response in the /api/put scene.
+// It holds the status code and the response values defined in
+// the [OpenTSDB Official Docs]: http://opentsdb.net/docs/build/html/api_http/put.html.
+type PutResponse struct {
+	Failed  int64      `json:"failed"`
+	Success int64      `json:"success"`
+	Errors  []PutError `json:"errors,omitempty"`
+	logger  Logger
+	tracer  trace.Tracer
```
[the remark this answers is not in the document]
coolwednesday: This seems a viable option.

pkg/gofr/datasource/opentsdb/interface.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"context"
+	"net"
+	"net/http"
+	"time"
+
+	"go.opentelemetry.io/otel/trace"
+)
+
+type Conn interface {
```
ccoVeille, line 12: I would use


```suggestion
type Connection interface {
```

ccoVeille, line 12: Also, do you need this interface to be exported?

I didn't find where it's used

coolwednesday: it is being used to write mocks for test...I will make it unexported tho.

pkg/gofr/datasource/opentsdb/preprocess.go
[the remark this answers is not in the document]
ccoVeille: Thank you for applying the changes

pkg/gofr/datasource/opentsdb/observability.go
```diff
@@ -0,0 +1,106 @@
...
+	logger.Debug(&QueryLog{
+		Operation: operation,
+		Status:    status,
+		Duration:  duration,
+		Message:   message,
+	})
+
+	if span != nil {
+		defer span.End()
+		span.SetAttributes(attribute.Int64(fmt.Sprintf("opentsdb.%v.duration", operation), duration))
```
ccoVeille, line 50: No need for a defer


```suggestion
		span.SetAttributes(attribute.Int64(fmt.Sprintf("opentsdb.%v.duration", operation), duration))
		span.End()
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,418 @@
...
+		message = "The given query param is invalid."
+		return errors.New(message)
+	}
+
+	var putEndpoint string
+	if !isEmptyPutParam(queryParam) {
+		putEndpoint = fmt.Sprintf("%s%s?%s", c.endpoint, putPath, queryParam)
+	} else {
+		putEndpoint = fmt.Sprintf("%s%s", c.endpoint, putPath)
+	}
```
ccoVeille, line 180: What about this?

```suggestion
	putEndpoint := fmt.Sprintf("%s%s", c.endpoint, putPath)
	if !isEmptyPutParam(queryParam) {
		putEndpoint = fmt.Sprintf("%s?%s", putEndpoint, queryParam)
	}
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,418 @@
...
+	status := statusFailed
+
+	message := "Put request failed"
+
+	defer sendOperationStats(c.logger, time.Now(), "Put", &status, &message, span)
+
+	putResp, ok := resp.(*PutResponse)
+	if !ok {
+		return errors.New("invalid response type. Must be *PutResponse")
+	}
```
ccoVeille, line 157: This is something I didn't get with the package signature

Everything is passed as any, then the type is checked

What is the reason behind this?

Is it for avoiding coupling in the interface used in containers package?

I would have expected something like this?

```suggestion
func (c *Client) PutDataPoints(ctx context.Context, datas any, queryParam string, resp *PutResponse) error {
	span := c.addTrace(ctx, "Put")

	status := statusFailed

	message := "Put request failed"

	defer sendOperationStats(c.logger, time.Now(), "Put", &status, &message, span)
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,418 @@
...
+	if err != nil {
+		return err
+	}
+
+	if len(tempResp.Errors) == 0 {
+		status = statusSuccess
+		message = fmt.Sprintf("Put request to url %q processed successfully", putEndpoint)
+		putResp.Success = tempResp.Success
+		putResp.Failed = tempResp.Failed
+		putResp.Errors = tempResp.Errors
```
ccoVeille, line 192: PutResp has only 3 fields

I'm surprised to see this here

Are you trying to do something like

```suggestion
		*putResp = tempResp
```

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,418 @@
...
+
+		return nil
+	}
+
+	return parsePutErrorMsg(tempResp)
```
ccoVeille, line 197: I would invert

```go
	if len(tempResp.Errors) > 0 {
		return parsePutErrorMsg(tempResp)
	}
	status = statusSuccess
	// ...

	return nil
```

pkg/gofr/datasource/opentsdb/preprocess.go
```diff
@@ -0,0 +1,395 @@
...
+		return true
+	case float64:
+		return true
+	case float32:
+		return true
+	case string:
+		return true
+	default:
+		return false
+	}
```
ccoVeille, line 242: Maybe this


```suggestion
	switch data.Value.(type) {
	case int64:
	case int:
	case float64:
	case float32:
	case string:
	default:
		return false
	}
	return true
```

I don't think we can do that with a switch type

```suggestion
	switch data.Value.(type) {
	case int64, int, float64, float32, string:
	default:
		return false
	}
	return true
```

pkg/gofr/datasource/opentsdb/response.go
```diff
@@ -0,0 +1,416 @@
...
+		message = fmt.Sprint(errSendingRequest)
+
+		return errSendingRequest
+	}
+
+	if resp.StatusCode >= 400 && resp.StatusCode < 500 {
+		return fmt.Errorf("client error: %d", resp.StatusCode)
+	}
+
+	defer resp.Body.Close()
```
ccoVeille, line 246: The body might need to be closed even on 400 and 500

This could be enough

```suggestion
	defer resp.Body.Close()
	if resp.StatusCode >= 400 && resp.StatusCode < 500 {
		return fmt.Errorf("client error: %d", resp.StatusCode)
	}
```
But this would be safer, I think

```suggestion
	if resp.Body != nil {
		defer resp.Body.Close()
	}

	if resp.StatusCode >= 400 && resp.StatusCode < 500 {
		return fmt.Errorf("client error: %d", resp.StatusCode)
	}
```

pkg/gofr/datasource/opentsdb/response.go
```diff
@@ -0,0 +1,416 @@
...
+		err := unmarshalFunc(result)
+		if err != nil {
+			message = fmt.Sprintf("unmarshal %s error: %s", operation, err)
+			logger.Errorf(message)
+
+			return err
+		}
+
+		status = statusSuccess
+		message = fmt.Sprintf("%s custom parsing was successful.", operation)
```
ccoVeille, line 350: I don't think other messages have a leading dot

```suggestion
		message = fmt.Sprintf("%s custom parsing was successful", operation)
```

pkg/gofr/datasource/opentsdb/response.go
```diff
@@ -0,0 +1,416 @@
...
+
+	resultBytes, err := json.Marshal(annotation)
+	if err != nil {
+		message = fmt.Sprintf("marshal annotation response error: %s", err)
+		return errors.New(message)
+	}
+
+	if err = c.sendRequest(ctx, method, annoEndpoint, string(resultBytes), annResp); err != nil {
+		message = fmt.Sprintf("%s: error while processing %s annotation request to url %q: %s", operation, method, annoEndpoint, err.Error())
+		return err
```
ccoVeille, line 407: I would recommend the following pattern everywhere


```suggestion
		err = fmt.Errorf("%s error while processing %s annotation request to url %q: %w", operation, method, annoEndpoint, err.Error)
		message = err.Error()
		return err
```

@Umang01-hash what do you think?

Error should be wrapped with information especially in the error returned and not only the log

pkg/gofr/datasource/opentsdb/opentsdb.go
coolwednesday: If I define actual type in the containers package....goFr package will have dependency added of opentsdb. I want to avoid that, hence as the interface is defined there, I am using any....

pkg/gofr/datasource/opentsdb/response.go
coolwednesday: GoFr has different log levels .....message is actually a debug message only appearing at debug level, which I feel must give information about the action in which the operation is being performed, So, I don't feel the context of the error should be the same as the debug level. It will diminish the value of debug level, if user gets all the info through errors....

But I am open to suggestions....

ccoVeille:
@coolwednesday You resolved some of the threads, then asked for a new review.

But you didn't push your last changes, right?

Or you were asking me to take a look at the replies you did on some threads?

coolwednesday:
> @coolwednesday You resolved some of the threads, then asked for a new review.
> 
> But you didn't push your last changes, right?
> 
> Or you were asking me to take a look at the replies you did on some threads?

@ccoVeille, can you check and review now ?

pkg/gofr/datasource/opentsdb/opentsdb.go
```diff
@@ -0,0 +1,418 @@
...
+
+	status := statusFailed
+
+	var message string
+
+	defer sendOperationStats(c.logger, time.Now(), "GetAggregators", &status, &message, span)
+
+	aggreResp, ok := resp.(*AggregatorsResponse)
+	if !ok {
+		return errors.New("invalid response type. Must be a *AggregatorsResponse")
```
ccoVeille, line 358: No message set here.

please consider a placeholder message

pkg/gofr/datasource/opentsdb/preprocess.go
```diff
@@ -0,0 +1,392 @@
...
+	// supply one or more map values in the same format as the query string. Tags are converted to filters in 2.2.
+	// Note that if no tags are specified, all metrics in the system will be aggregated into the results.
+	// It will be deprecated in OpenTSDB 2.2.
+	Tags map[string]string `json:"tags,omitempty"`
+}
+
+func getQueryBodyContents(param any) (string, error) {
+	result, err := json.Marshal(param)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal query param: %v", err)
```
ccoVeille, line 296:
```suggestion
		return "", fmt.Errorf("failed to marshal query param: %w", err)
```

pkg/gofr/datasource/opentsdb/response.go
```diff
@@ -0,0 +1,374 @@
...
+		// Use the default JSON unmarshaller if no custom parser is provided.
+		if err = json.Unmarshal(jsonBytes, parsedResp); err != nil {
+			errUnmarshaling := fmt.Errorf("failed to unmarshal response body for %s %s: %w", method, url, err)
+
+			return errUnmarshaling
+		}
+	} else {
+		// Use the custom parser if available.
+		if err := parser(jsonBytes); err != nil {
+			return fmt.Errorf("failed to parse response body through custom parser %s %s: %v", method, url, err)
```
ccoVeille, line 250:
```suggestion
			return fmt.Errorf("failed to parse response body through custom parser %s %s: %w", method, url, err)
```

pkg/gofr/datasource/opentsdb/response.go
```diff
@@ -0,0 +1,374 @@
...
+	status := statusFailed
+
+	var message string
+
+	defer sendOperationStats(c.logger, time.Now(), "Version", &status, &message, span)
+
+	verEndpoint := fmt.Sprintf("%s%s", c.endpoint, versionPath)
+
+	if err := c.sendRequest(ctx, http.MethodGet, verEndpoint, "", verResp); err != nil {
+		message = fmt.Sprintf("error while processing request at URL %s: %s", verEndpoint, err)
```
ccoVeille, line 273:
```suggestion
		message = fmt.Sprintf("error while processing request at URL %s: %e", verEndpoint, err)
```

ccoVeille (approved):

Umang01-hash (approved):

aryanmehrotra (approved):

PR #1118, opened 2024-10-17
Adding context to database provider

### LINKED ISSUE #1117: mongo driver doesnt support contexts
**Describe the bug**
```go	
m, err := mongo.Connect(context.Background(), options.Client().ApplyURI(uri))
if err != nil {
	c.logger.Errorf("error connecting to mongoDB, err:%v", err)

	return
}
```

The context should be passed on here, there is no way to set a timeout or cancellation on this.

The Connect() interface should accept a context and return an error.

**To Reproduce**

1. The code is

   ```go
   
   ```

2. The error is

   ```
   
   ```

**Expected behavior**
Right now it will sit there and try and connect to a mongodb instance forever. users should have the ability to set a timeout context and properly return errors, instead of just printing stuff with the logger.

**Screenshots**

- OS: [e.g. Linux]
- gofr version [e.g. v1.5.0]
- go version [e.g. 1.21]

**More description**

mfreeman451:
## Pull Request Template

**Description:**

-   Added the ability to pass a context down through the DB provider to the Connect() call
-   Connect() now returns an error.
-   Currently the Connect() interface doesn't accept a context, invalid connections to databases will not be known.

-   Connect() calls in the DB provider will now return an error and require a context to be passed.
-   This breaks the ability to daisy chain method calls.

**Additional Information:**

-   Closes #1117 

**Checklist:**


pkg/gofr/container/datasources.go
```diff
@@ -285,8 +285,8 @@ type provider interface {
 	// UseTracer sets the tracer for the Cassandra client.
 	UseTracer(tracer any)
 
-	// Connect establishes a connection to Cassandra and registers metrics using the provided configuration when the client was Created.
-	Connect()
+	// Connect establishes a connection to a DB and registers metrics using the provided configuration when the client was Created.
+	Connect(ctx context.Context) error
```
aryanmehrotra, line 289: why connect would need a context?

Also this is breaking a lot of datastores - can you check all the different datastores as all of them implements the provider interface.

aryanmehrotra (changes requested):

pkg/gofr/external_db.go
```diff
@@ -33,16 +39,20 @@ func (a *App) AddFTP(fs file.FileSystemProvider) {
 }
 
 // AddPubSub sets the PubSub client in the app's container.
-func (a *App) AddPubSub(pubsub container.PubSubProvider) {
+func (a *App) AddPubSub(ctx context.Context, pubsub container.PubSubProvider) error {
```
aryanmehrotra, line 42: This would result in a breaking change - user would need to handle the errors on their own rather than gofr taking care of that.

aryanmehrotra (changes requested):

pkg/gofr/external_db.go
mfreeman451: gofr isn't really "taking care of" anything, it just hides the error in logging messages..

pkg/gofr/datasource/mongo/errors.go
```diff
@@ -0,0 +1,17 @@
...
+	ErrInvalidURI = errors.New("invalid MongoDB URI")
+
+	// ErrAuthentication is returned when authentication fails.
+	ErrAuthentication = errors.New("authentication failed")
+
+	// ErrDatabaseConnection is returned when the client fails to connect to the specified database.
+	ErrDatabaseConnection = errors.New("failed to connect to database")
+
+	// ErrGenericConnection is returned for general connection issues.
+	ErrGenericConnection = errors.New("MongoDB connection error")
```
ccoVeille, line 16: Here is a suggestion, feel free to consider it or ignore it


```suggestion
	// ErrInvalidURI is returned when the MongoDB URI is invalid or cannot be parsed.
	ErrInvalidURI = fmt.Errorf("%w: invalid MongoDB URI", ErrConnection)

	// ErrAuthentication is returned when authentication fails.
	ErrAuthentication = fmt.Errorf("%w: authentication failed", ErrConnection)

	// ErrDatabaseConnection is returned when the client fails to connect to the specified database.
	ErrDatabaseConnection = fmt.Errorf("%w: failed to connect to database", ErrConnection)

	// ErrGenericConnection is returned for general connection issues.
	ErrConnection = errors.New("MongoDB connection error")
```

This way the caller can catch all mongodb error with an errors.Is(err, ErrConnection)

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -77,27 +78,103 @@ func (c *Client) UseTracer(tracer any) {
...
+
+func (*Client) isTimeoutError(err error) bool {
+	return strings.Contains(err.Error(), "connection timeout") || mongo.IsTimeout(err)
+}
 
-		return
+func (*Client) isAuthenticationError(err error) bool {
+	return strings.Contains(err.Error(), "authentication failed") ||
+		strings.Contains(err.Error(), "AuthenticationFailed")
+}
```
ccoVeille, line 141: I have a big problem with this

Validating errors with what they contains is a bad idea

If you have no other way to do that, then you have to add a clear comment about it

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -77,27 +78,103 @@ func (c *Client) UseTracer(tracer any) {
...
+	return nil
+}
+
+func (c *Client) handlePingError(err error) error {
+	if mongo.IsTimeout(err) {
+		return fmt.Errorf("%w: connection timeout", ErrGenericConnection)
 	}
 
+	if errors.Is(err, mongo.ErrClientDisconnected) {
+		return fmt.Errorf("%w: client disconnected", ErrGenericConnection)
```
ccoVeille, line 157:
```suggestion
		return fmt.Errorf("%w: client disconnected: %w", ErrGenericConnection, err)
```

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -77,27 +78,103 @@ func (c *Client) UseTracer(tracer any) {
...
+	if err := client.Ping(ctx, nil); err != nil {
+		return c.handlePingError(err)
+	}
+
+	return nil
+}
+
+func (c *Client) handlePingError(err error) error {
+	if mongo.IsTimeout(err) {
+		return fmt.Errorf("%w: connection timeout", ErrGenericConnection)
```
ccoVeille, line 153:
```suggestion
		return fmt.Errorf("%w: connection timeout: %w", ErrGenericConnection, err)
```

You should never hide low level error with yours. You should wrap them with yours

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -77,27 +78,103 @@ func (c *Client) UseTracer(tracer any) {
...
+
+	if c.isTimeoutError(err) {
+		return fmt.Errorf("%w: connection timeout", ErrGenericConnection)
+	}
+
+	return fmt.Errorf("%w: %w", ErrGenericConnection, err)
+}
+
+func (*Client) isTimeoutError(err error) bool {
+	return strings.Contains(err.Error(), "connection timeout") || mongo.IsTimeout(err)
```
ccoVeille, line 135: Is there a reason to do not rely *only* on [mongo.IsTimeOut](https://pkg.go.dev/go.mongodb.org/mongo-driver/mongo#pkg-functions) ?

It seems strange to check if error contains a text

If you have to, something I'm not sure, you should add a comment explaining it.

pkg/gofr/datasource/mongo/mongo.go
ccoVeille: Take a look at the code

https://github.com/mongodb/mongo-go-driver/blob/b823409f1094fc54099bf64799a1e5a1d9e6fa67/x/mongo/driver/auth/auth.go#L148

An error could be caught with an errors.As, looking for auth.Error

https://pkg.go.dev/go.mongodb.org/mongo-driver@v1.17.1/x/mongo/driver/auth#Error

aryanmehrotra:
@mfreeman451 can you fix the reveiw comments? 
We would not want to keep a PR open for more than 2 weeks - for better maintainability.

pkg/gofr/datasource/mongo/mongo.go
vikash: If we write new code which does not adhere to best practices, it should not be treated as outside the scope. Error validation should not be based on string matching.

pkg/gofr/datasource/cassandra/cassandra.go
```diff
@@ -61,14 +61,15 @@ func New(conf Config) *Client {
 }
 
 // Connect establishes a connection to Cassandra and registers metrics using the provided configuration when the client was Created.
-func (c *Client) Connect() {
-	c.logger.Debugf("connecting to Cassandra at %v on port %v to keyspace %v", c.config.Hosts, c.config.Port, c.config.Keyspace)
+func (c *Client) Connect(_ context.Context) error {
+	c.logger.Logf("connecting to cassandra at %v on port %v to keyspace %v", c.config.Hosts, c.config.Port, c.config.Keyspace)
```
ccoVeille, line 65: I think this change is unintended

```suggestion
	c.logger.Logf("connecting to Cassandra at %v on port %v to keyspace %v", c.config.Hosts, c.config.Port, c.config.Keyspace)
```

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -82,31 +82,96 @@ func (c *Client) UseTracer(tracer any) {
 }
 
 // Connect establishes a connection to MongoDB and registers metrics using the provided configuration when the client was Created.
-func (c *Client) Connect() {
-	c.logger.Debugf("connecting to MongoDB at %v to database %v", c.config.URI, c.config.Database)
+func (c *Client) Connect(ctx context.Context) error {
+	c.logger.Logf("connecting to mongoDB at %v to database %v", c.config.URI, c.config.Database)
```
ccoVeille, line 86:
```suggestion
	c.logger.Logf("connecting to MongoDB at %v to database %v", c.config.URI, c.config.Database)
```

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -82,31 +82,96 @@ func (c *Client) UseTracer(tracer any) {
...
+	if c.isTimeoutError(err) {
+		return fmt.Errorf("%w: connection timeout", ErrGenericConnection)
+	}
+
+	return fmt.Errorf("%w: %w", ErrGenericConnection, err)
+}
+
+func (*Client) isTimeoutError(err error) bool {
+	return strings.Contains(err.Error(), "connection timeout") || mongo.IsTimeout(err)
+}
```
ccoVeille, line 141: See https://github.com/gofr-dev/gofr/pull/1118#discussion_r1836078938

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -116,6 +181,12 @@ func (c *Client) Connect() {
 
 	mongoBuckets := []float64{.05, .075, .1, .125, .15, .2, .3, .5, .75, 1, 2, 3, 4, 5, 7.5, 10}
 	c.metrics.NewHistogram("app_mongo_stats", "Response time of MONGO queries in milliseconds.", mongoBuckets...)
```
ccoVeille, line 183:
```suggestion
 	c.metrics.NewHistogram("app_mongo_stats", "Response time of MongoDB queries in milliseconds.", mongoBuckets...)
```

Minor typo that could be fixed even if out of the PR

mfreeman451:
cleaned up and updated

ccoVeille (approved):

coolwednesday:
Hey @mfreeman451,

I was going through the code and I believe the current implementation might introduce a breaking change. To prevent that, here's what I suggest:

- Let's focus solely on MongoDB for now. We should avoid making changes in other datasources for now. Please revert those changes.

- We can create an interface that embeds the existing `provider` interface, say `providerWithContext` and introduces an additional method, `ConnectWithContext(ctx context.Context)`. This can help us to handle the only MongoDB-specific logic without affecting other providers in this PR.

- Replace the current provider implementation with this new interface for `MongoProvider`, enabling the optional use of the `ConnectWithContext` method.

- In `externaldb.go`, where we are injecting the `MongoDB` provider using `AddMongo`, we will replace the call to `Connect() `with `ConnectWithContext(ctx context.Context)`.

- To handle context passing efficiently, we can use the functional options pattern. This way, context can be optionally passed to the connection method without breaking the existing flow.

I believe the approach I suggested is one way to address this. However, if you have a better or more efficient solution, feel free to go ahead and use it to resolve the issue. I’m open to exploring alternatives that might work better for our use case.

Let me know your thoughts on this!

Umang01-hash:
@mfreeman451 are you proceeding towards the review-changes suggested by @coolwednesday ?

aryanmehrotra:
Please reopen once the changes are fixed and commits are resolved as it has been more than 2 weeks of last activity.

Issue #1080, opened 2024-10-04
Invalid Cassandra godocs 


aryanmehrotra:
godocs is selecting docs from the mocks generated, should be proper.

Along with this re-verify the other go-docs comments present in the package

https://pkg.go.dev/gofr.dev/pkg/gofr/datasource/cassandra

<img width="462" alt="Screenshot 2024-10-04 at 7 04 33 PM" src="https://github.com/user-attachments/assets/46619626-47a9-4be5-9098-84b2df58804d">

FilledEther20:
@aryanmehrotra Greetings Sir Could u assign this issue to me?

aryanmehrotra:
We can assign one issue at a time to you, once the other issue is done let us know we will assign

PR #1128, opened 2024-10-18
Fix issue#1080 invalid cassandra docs

FilledEther20:
## Pull Request 

**Description:**

-   This pull request addresses issue Invalid Cassandra godocs  #1080 by updating the Cassandra package to resolve the problem where GoDocs were incorrectly selecting from generated mocks.
-   The update ensures that only the relevant Cassandra documentation is highlighted, improving both readability and maintainability of the code.

**Checklist:**


coolwednesday (approved):

pkg/gofr/datasource/cassandra/mock_interfaces.go
```diff
@@ -1,3 +1,6 @@
+//go:build exclude
```
vipul-rawat, line 1: I just have one question, will this change get reverted when someone generates a new mock using mockgen - then this change has to be kept in mind.

FilledEther20: Yes, these changes will be reverted when a new mock is generated, as they are manually added. Once mockgen is run, it will overwrite these modifications automatically. To address this, I considered creating an internal directory to hold all test and mock files, which would be excluded from generating the GoDocs. However, this led to circular dependency issues, requiring changes across the entire codebase. That's why I opted for the current approach of manually adding the modifications.

vipul-rawat: Other contributors would have to make sure that these are included whenever there is any change in the mock files.

Issue #1146, opened 2024-10-23
Fix AddStaticFiles method


Umang01-hash:
**Describe the bug**
The method `AddStaticFiles` is not wokring correctly.

If we try to run this code:

```go
func main() {
    app := gofr.New()

    app.AddStaticFiles("", "./static")

    app.Run()
}
```

And have some files in our static directory , when trying to access the route `http://localhost:9000/abc.jpeg` we are getting route not registered.

**To Reproduce**
Steps to reproduce the behavior, if applicable:

1. The code is:

```go
package main

import "gofr.dev/pkg/gofr"

func main() {
	app := gofr.New()

	app.AddStaticFiles("/", "./static")

	app.Run()
}
```

2. The error is

```
{
    "error": {
        "message": "route not registered"
    }
} 
```

**Expected behavior**
We should be able to successfully route it.

Umang01-hash:
@IvanGael Thank you for your interest, but I have already raised a PR for the same. You can please check any other existing issues will assign that to you.

PR #1147, opened 2024-10-23
Fix AddStaticFiles method

Umang01-hash:
## Pull Request Template

**Description:**
- Resolve issue #1146 
-   Move the setup the default routes of `/.well-known` to `New` method of app, instead of `Run` so as to ensure proper handlers are called for the specific routes instead of **staticHandler**.

**Checklist:**


pkg/gofr/gofr.go
```diff
@@ -88,6 +88,17 @@ func New() *App {
 	app.httpServer.certFile = app.Config.GetOrDefault("CERT_FILE", "")
 	app.httpServer.keyFile = app.Config.GetOrDefault("KEY_FILE", "")
 
+	// Add Default routes
+	app.add(http.MethodGet, "/.well-known/health", healthHandler)
+	app.add(http.MethodGet, "/.well-known/alive", liveHandler)
```
ccoVeille, line 93: Just a question, not about your changes in this PR

Are these well known URI somehow standard?

The only documentation I found was the gofr documentation

https://gofr.dev/docs/advanced-guide/monitoring-service-health

The documentation doesn't talk about it

I don't find them also on Wikipedia

https://en.wikipedia.org/wiki/Well-known_URI

Doesn't it mean gofr created its own "standard", but that is the only one to use?

I'm unsure I would have used .well-known URI for these

Unless there is an rfc or other project using this

pkg/gofr/http/router.go
```diff
@@ -56,7 +59,12 @@ func (rou *Router) AddStaticFiles(endpoint, dirName string) {
 	cfg := staticFileConfig{directoryName: dirName}
 
 	fileServer := http.FileServer(http.Dir(cfg.directoryName))
-	rou.Router.NewRoute().PathPrefix(endpoint + "/").Handler(http.StripPrefix(endpoint, cfg.staticHandler(fileServer)))
+
+	if endpoint == "/" {
+		rou.Router.NewRoute().PathPrefix("/").Handler(cfg.staticHandler(fileServer))
+	} else {
+		rou.Router.NewRoute().PathPrefix(endpoint + "/").Handler(http.StripPrefix(endpoint, cfg.staticHandler(fileServer)))
+	}
```
ccoVeille, line 67: What about this

```suggestion
	if endpoint == "/" {
		rou.Router.NewRoute().PathPrefix("/").Handler(cfg.staticHandler(fileServer))
		return
	}
	rou.Router.NewRoute().PathPrefix(endpoint + "/").Handler(http.StripPrefix(endpoint, cfg.staticHandler(fileServer)))
```

ccoVeille, line 67: I'm unsure it's the right place (lines of code or even the right PR) to comment this, but I'm commenting here, when I think about it.

If my inputs lead to a conclusion, please consider opening an issue.

So here is my point:

Are we sure the URL path is cleaned?

I found many framework having issues when a path in URL was provided with `//`

Like `//favicon.ico` or `/.well-known//health` it's often caused by a URL of the server containing a trailing slash in a environment variable.

This often leads to 404, while an Apache2 or NGinx server would have served the content

My remark is valid for static files but also for API endpoints

pkg/gofr/gofr.go
```diff
@@ -88,6 +88,17 @@ func New() *App {
...
+	app.add(http.MethodGet, "/favicon.ico", faviconHandler)
+
+	if _, err = os.Stat("./static/openapi.json"); err == nil {
+		app.add(http.MethodGet, "/.well-known/openapi.json", OpenAPIHandler)
+		app.add(http.MethodGet, "/.well-known/swagger", SwaggerUIHandler)
+		app.add(http.MethodGet, "/.well-known/{name}", SwaggerUIHandler)
```
ccoVeille, line 99: This line requires a comment

Is it a kind of catchall to send anything that is not openapi.json and swagger to the SwaggerUIHandler

pkg/gofr/gofr.go
```diff
@@ -88,6 +88,17 @@ func New() *App {
 	app.httpServer.certFile = app.Config.GetOrDefault("CERT_FILE", "")
 	app.httpServer.keyFile = app.Config.GetOrDefault("KEY_FILE", "")
 
+	// Add Default routes
+	app.add(http.MethodGet, "/.well-known/health", healthHandler)
+	app.add(http.MethodGet, "/.well-known/alive", liveHandler)
+	app.add(http.MethodGet, "/favicon.ico", faviconHandler)
+
+	if _, err = os.Stat("./static/openapi.json"); err == nil {
+		app.add(http.MethodGet, "/.well-known/openapi.json", OpenAPIHandler)
```
ccoVeille, line 97: Why not using the default swagger constant you defined? For these 2 lines.

Umang01-hash: Sure used the constant here

pkg/gofr/gofr.go
Umang01-hash: Sure added the comments

pkg/gofr/http/router.go
Umang01-hash: Sure changed the code as per your suggestion

pkg/gofr/gofr.go
Umang01-hash: Hi @ccoVeille 

Thank you for your insightful question regarding the `.well-known` URIs.

You are correct that these URIs are not part of an official RFC or standard for the specific use cases we've implemented in GoFr. We developed these endpoints to streamline access to important API information such as health status, swagger documentation, and more. Our goal was to provide a consistent and convenient way for developers to obtain vital information about their APIs without the need for additional authentication.

While the .well-known URIs in GoFr may not align with a globally recognized standard, we believe they offer a practical solution for common development needs. We appreciate your feedback and will continue to consider how we can improve our approach to be more in line with industry standards and best practices.

Please let us know if you have any further questions or suggestions. We're here to help!

ccoVeille: I think this information is solid. I would suggest to add them in the code documentation but also in the official documentation

pkg/gofr/http/router.go
Umang01-hash: Sure will add that issue

pkg/gofr/gofr.go
Umang01-hash: Sure @ccoVeille will add this to the documentations also

ccoVeille: Side remark, I would have used a gofr prefix in the URI then
Something like /.well-known/gofr

But I assume it's either too late and opinionated

ccoVeille (approved):

vipul-rawat (approved):

PR #1149, opened 2024-10-24
Add zop cli documentation to showcase deployment tool

vipul-rawat:


vikash:
We need to change certain things in the implementation which will affect the documentation. So, closing it for now.

PR #1161, opened 2024-10-29
Feature to fetch Auth Info from context when auth is enabled

### LINKED ISSUE #1094: Feature to fetch claims when authentication is enabled.
We can have a feature to get claims from GoFr context ....no matter the kind of authentication we use. 

The usecase can be generalized for various kind of authentication like jwt or api key etc .....

coolwednesday:
**Description:**
-   Get claims/username/apikey through the context when auth is enabled.
-   Closes #1094 

**Checklist:**


pkg/gofr/context.go
```diff
@@ -28,6 +31,12 @@ type Context struct {
 	responder Responder
 }
 
+type AuthInfo interface {
+	GetClaims() jwt.MapClaims
+	GetUsername() string
```
ccoVeille, line 36: I would suggest this "username" is term made of two words

```suggestion
	GetUserName() string
```

pkg/gofr/context.go
```diff
@@ -75,6 +84,40 @@ func (c *Context) WriteMessageToSocket(data any) error {
...
+		return ""
+	}
+
+	return claims
+}
+
+func (c *Context) GetAPIKey() string {
+	claims, ok := c.Request.Context().Value(middleware.APIKey).(string)
+	if !ok {
+		c.Warn("failed to get api key from context")
```
ccoVeille, line 114:
```suggestion
		c.Warn("failed to get API key from context")
```

pkg/gofr/context.go
```diff
@@ -75,6 +84,40 @@ func (c *Context) WriteMessageToSocket(data any) error {
...
+		return nil
+	}
+
+	return claims
+}
+
+func (c *Context) GetUsername() string {
+	claims, ok := c.Request.Context().Value(middleware.Username).(string)
+	if !ok {
+		c.Warn("failed to get username from context")
```
ccoVeille, line 104: I wouldn't use the failed wording on a non-blocking error.

You are using a warning error level, so I would use something like


```suggestion
		c.Warn("no username found in context")
```

ccoVeille: But I'm unsure about the logic of logging here

For me, there should be no log, it should be either:
- returning an error when missing
- returning simply a void result, like you do

In both cases, it should be from caller responsibility to handle this case

Imagine this log entry could be repeated fof ever and ever

pkg/gofr/context_test.go
```diff
@@ -7,6 +7,9 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"github.com/golang-jwt/jwt/v5"
```
ccoVeille, line 10: Please fix this

You can use gci

pkg/gofr/http/middleware/basic_auth.go
```diff
@@ -16,6 +17,8 @@ type BasicAuthProvider struct {
 	Container                   *container.Container
 }
 
+const Username authMethod = 1
```
ccoVeille, line 20:
```suggestion
const UserName authMethod = 1
```

ccoVeille (commented):
Hello, it's me again 😅

pkg/gofr/context.go
coolwednesday: I feel username is a single word , so it should not be UserName.

pkg/gofr/http/middleware/basic_auth.go
coolwednesday: username is a single word.

pkg/gofr/context.go
```diff
@@ -3,6 +3,9 @@ package gofr
 import (
 	"context"
 
+	"github.com/golang-jwt/jwt/v5"
+	"gofr.dev/pkg/gofr/http/middleware"
+
 	"github.com/gorilla/websocket"
```
ccoVeille, line 9:
```suggestion
	"gofr.dev/pkg/gofr/http/middleware"

	"github.com/golang-jwt/jwt/v5"
	"github.com/gorilla/websocket"
```

ccoVeille (approved):

Umang01-hash (approved):

pkg/gofr/context.go
```diff
@@ -3,6 +3,9 @@ package gofr
 import (
 	"context"
 
+	"gofr.dev/pkg/gofr/http/middleware"
```
ccoVeille, line 6: Move it at the end with the other gofr.dev one

vipul-rawat (approved):

PR #1163, opened 2024-10-30
Enhance CMD: introduce TUI elements(Spinners and ProgressBar) and terminal painting

### LINKED ISSUE #1164: Add CMD TUI elements for gofr CMD
Add some TUI elements for the CMD applications and allow users to modify and paint the terminal using the terminal escape sequence.

vipul-rawat:
closes #1164 

image
image


examples/sample-cmd/main.go
```diff
@@ -23,6 +25,31 @@ func main() {
...
+		p := terminal.NewProgressBar(ctx.Out, 100)
+
+		for i := 1; i <= 100; i++ {
+			// do a time taking process or compute a small subset of a bigger problem,
+			// this could be processing batches of a data set.
+			time.Sleep(50 * time.Millisecond)
+
+			// increment the progress to display on the progress bar.
+			p.Incr(int64(1))
+		}
```
ccoVeille, line 48: This sub command won't expire if the context is canceled

I would use a check of ctx.Err() to get out the loop

pkg/gofr/cmd/terminal/colors.go
```diff
@@ -0,0 +1,30 @@
...
+	BrightGreen
+	BrightYellow
+	BrightBlue
+	BrightMagenta
+	BrightCyan
+	BrightWhite
+)
+
+func (o *Output) SetColor(colorCode int) {
+	o.Printf(csi + "38;5;" + strconv.Itoa(colorCode) + "m")
```
ccoVeille, line 25: Why a sprintf if you concatenating string, and do not use %s or %d

It would avoid the call to strings.Itoa

pkg/gofr/cmd/terminal/output.go
```diff
@@ -0,0 +1,83 @@
...
+	CursorPrevLine(n int)
+	CursorUp(n int)
+	DeleteLines(n int)
+	ExitAltScreen()
+	HideCursor()
+	InsertLines(n int)
+	MoveCursor(row int, column int)
+	Print(messages ...interface{})
+	Printf(format string, args ...interface{})
+	Println(messages ...interface{})
```
ccoVeille, line 44:
```suggestion
	Print(messages ...any)
	Printf(format string, args ...any)
	Println(messages ...any)
```

pkg/gofr/cmd/terminal/output_test.go
```diff
@@ -0,0 +1,18 @@
+package terminal
+
+import (
+	"github.com/stretchr/testify/assert"
+	"os"
+	"testing"
```
ccoVeille, line 6: Fix the import order please

pkg/gofr/cmd/terminal/output_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+func TestNewOutput(t *testing.T) {
+	// intialize a new standard output stream
+	o := New()
+
+	assert.Equal(t, os.Stdout, o.out)
+	assert.Equal(t, uintptr(1), o.fd)
+
+	// for tests, the os.Stdout do not directly outputs to the terminal
+	assert.Equal(t, false, o.isTerminal)
```
ccoVeille, line 17:
```suggestion
	assert.False(t, o.isTerminal)
```

pkg/gofr/cmd/terminal/progress.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (p *ProgressBar) Incr(i int64) bool {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if p.current < p.total {
+		p.current += i
+		if p.current > p.total {
+			p.current = p.total
+		}
```
ccoVeille, line 50: NitPick: You can call max now


```suggestion
		p.current = max(p.current, p.total)
```

pkg/gofr/cmd/terminal/progress.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return p.current != p.total
+}
+
+func (p *ProgressBar) updateProgressBar() {
+	// perform the TUI update of the progress bar
+	p.stream.Print("\r")
+
+	pString := p.getString()
+	p.stream.Printf("%s", pString)
```
ccoVeille, line 63: This is a print then, no?


```suggestion
	p.stream.Print(pString)
```

pkg/gofr/cmd/terminal/progress_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+func TestProgressBar_Fail(t *testing.T) {
+	out := testutil.StdoutOutputForFunc(func() {
+		var out bytes.Buffer
+		stream := &Output{terminal{isTerminal: true, fd: 1}, &out}
+		bar := NewProgressBar(stream, int64(-1))
+
+		assert.Equal(t, bar.total, int64(0))
```
ccoVeille, line 47: Expected and actual are inverted no?


```suggestion
		assert.Equal(t, int64(0), bar.total)
```

But I would use Zero anyway


```suggestion
		assert.Zero(t, bar.total)
```

pkg/gofr/cmd/terminal/progress_test.go
```diff
@@ -0,0 +1,131 @@
...
+	// Increment the progress by 20
+	b := bar.Incr(int64(20))
+	if !b && bar.current != 20 {
+		t.Errorf("fail: bar incremented value not correct current: %v should be 20", bar.current)
+	}
+
+	expectedOut := "\r[█████████░                                        ] 20.000%"
+	if out.String() != expectedOut {
+		t.Errorf("Unexpected progress bar output: got %q, want %q", out.String(), expectedOut)
+	}
```
ccoVeille, line 71: Here and everywhere else

Why not using ?

require.Equal(t, expectedOut, out.String())

pkg/gofr/cmd/terminal/spinner.go
```diff
@@ -0,0 +1,81 @@
...
+
+	s.Stream.HideCursor()
+
+	go func() {
+		for range t.C {
+			if s.started {
+				s.Stream.Print("\r")
+			} else {
+				break
+			}
```
ccoVeille, line 61:
```suggestion
			if !s.started {
				break
			}
			s.Stream.Print("\r")
```

pkg/gofr/cmd/terminal/spinner.go
```diff
@@ -0,0 +1,81 @@
...
+
+			s.Stream.Printf("%s"+"", s.Frames[i%len(s.Frames)])
```
ccoVeille, line 63: Why appending an empty string?

Also this? No?

```suggestion
		s.Stream.Print(s.Frames[i%len(s.Frames)])
```

pkg/gofr/cmd/terminal/spinner_test.go
```diff
@@ -0,0 +1,73 @@
...
+	// Testing Dot spinner
+	b := &bytes.Buffer{}
+	output := &Output{out: b}
+	spinner := NewDotSpinner(output)
+
+	// Start the spinner
+	spinner.Spin()
+
+	// Let it run for a bit
+	time.Sleep(time.Duration(waitTime) * time.Second)
```
ccoVeille, line 22:
```suggestion
	var waitTime = 3 * time.Second

	// Testing Dot spinner
	b := &bytes.Buffer{}
	output := &Output{out: b}
	spinner := NewDotSpinner(output)

	// Start the spinner
	spinner.Spin()

	// Let it run for a bit
	time.Sleep(waitTime)
```

pkg/gofr/cmd/terminal/terminalOut.go
ccoVeille, line 1: out.go would be enough in terminal package

Also, it would comply Go convention about filenames

pkg/gofr/cmd/terminal/terminalOut.go
```diff
@@ -0,0 +1,188 @@
...
+	"strings"
+)
+
+const (
+	// escape Escape character.
+	escape = '\x1b'
+	// csi Control Sequence Introducer.
+	csi = string(escape) + "["
+	// osc Operating System Command.
+	osc = string(escape) + "]"
```
ccoVeille, line 14:
```suggestion
	// escape Escape character.
	escape = string('\x1b')
	// csi Control Sequence Introducer.
	csi = escape + "["
	// osc Operating System Command.
	osc = escape + "]"
```

pkg/gofr/cmd/terminal/terminalOut.go
```diff
@@ -0,0 +1,188 @@
...
+	cursorPositionSeq        = "%d;%dH"
+	eraseDisplaySeq          = "%dJ"
+	eraseLineSeq             = "%dK"
+	saveCursorPositionSeq    = "s"
+	restoreCursorPositionSeq = "u"
+	changeScrollingRegionSeq = "%d;%dr"
+	insertLineSeq            = "%dL"
+	deleteLineSeq            = "%dM"
+
+	// Explicit values ersasing lines
```
ccoVeille, line 35:
```suggestion
	// Explicit values erasing lines
```

examples/sample-cmd/main.go
```diff
@@ -23,6 +25,31 @@ func main() {
 		return fmt.Sprintf("Hello %s!", c.Param("name")), nil
 	})
 
+	app.SubCommand("spinner", func(ctx *gofr.Context) (interface{}, error) {
+		// intialize the spinner and defer stop it
```
Umang01-hash, line 29: typo in word `initialize`

Umang01-hash (changes requested):

examples/sample-cmd/main.go
```diff
@@ -23,6 +25,36 @@ func main() {
...
+			case <-ctx.Done():
+				return nil, ctx.Err()
+			default:
+				// do a time taking process or compute a small subset of a bigger problem,
+				// this could be processing batches of a data set.
+				time.Sleep(50 * time.Millisecond)
+
+				// increment the progress to display on the progress bar.
+				p.Incr(int64(1))
+			}
```
ccoVeille, line 52: This is the way to do, otherwise the time.Sleep is still blocking

```suggestion
			select {
			case <-ctx.Done():
				return nil, ctx.Err()
			case <-time.After(50 * time.Millisecond):
				// do a time taking process or compute a small subset of a bigger problem,
				// this could be processing batches of a data set.

				// increment the progress to display on the progress bar.
				p.Incr(int64(1))
			}
```

ccoVeille (changes requested):

Umang01-hash (dismissed):

examples/sample-cmd/main.go
```diff
@@ -23,6 +25,35 @@ func main() {
 		return fmt.Sprintf("Hello %s!", c.Param("name")), nil
 	})
 
+	app.SubCommand("spinner", func(ctx *gofr.Context) (interface{}, error) {
+		// initialize the spinner and defer stop it
+		defer terminal.NewDotSpinner(ctx.Out).Spin().Stop()
+
+		// stimulate a time-taking process
+		time.Sleep(2 * time.Second)
```
ccoVeille, line 33: Here also I would use the following pattern


```go
                        case <-ctx.Done(): 
                                 
                         case <-time.After(2 * time.Second): 
                                
                         }
```

pkg/gofr/cmd/terminal/progress.go
```diff
@@ -0,0 +1,105 @@
...
+
+type Term interface {
+	IsTerminal(fd int) bool
+	GetSize(fd int) (width, height int, err error)
+}
+
+func NewProgressBar(out Output, total int64) *ProgressBar {
+	w, _, err := out.getSize()
+	if err != nil {
+		fmt.Printf("error getting terminal size, err : %v, could not initialize progress bar\n", err)
```
ccoVeille, line 25: Why not printing to output?

But also, why not simply returning the error , and let the caller handles it?

pkg/gofr/cmd/terminal/progress.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	if total < 0 {
+		fmt.Println("error initializing progress bar, total should be > 0")
```
ccoVeille, line 29: Same here

pkg/gofr/cmd/terminal/progress.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		if roundedPercent > 0 && roundedPercent < 50 {
+			pbBox = fmt.Sprintf("[%s%s%s] ", strings.Repeat("█", roundedPercent-1), "░", strings.Repeat(" ", numSpaces))
+		} else if roundedPercent <= 0 {
+			pbBox = fmt.Sprintf("[%s] ", strings.Repeat(" ", numSpaces))
+		} else {
+			pbBox = fmt.Sprintf("[%s%s] ", strings.Repeat("█", roundedPercent), strings.Repeat(" ", numSpaces))
+		}
+	}
```
ccoVeille, line 100: I would move this to an unexported method,allowing you to use early returns

pkg/gofr/cmd/terminal/progress_test.go
```diff
@@ -0,0 +1,127 @@
...
+	assert.Equal(t, expectedCompletion, out.String())
+}
+
+func TestProgressBar_Fail(t *testing.T) {
+	out := testutil.StdoutOutputForFunc(func() {
+		var out bytes.Buffer
+		stream := &Out{terminal{isTerminal: true, fd: 1}, &out}
+		bar := NewProgressBar(stream, int64(-1))
+
+		assert.Equal(t, int64(0), bar.total)
```
ccoVeille, line 44: As the type is known, I would have used


```suggestion
		assert.Zero(t, bar.total)
```

pkg/gofr/cmd/terminal/spinner.go
```diff
@@ -0,0 +1,80 @@
...
+func NewGlobeSpinner(o Output) *Spinner {
+	return &Spinner{
+		frames:    []string{"🌍", "🌎", "🌏"},
+		fps:       time.Second / 4,
+		outStream: o,
+	}
+}
+
+func (s *Spinner) Spin() *Spinner {
+	t := time.NewTicker(s.fps)
```
ccoVeille, line 48: I'm a bit worried a ticker could be created, but never cleaned,if no one call Stop

I'm unsure about the best way,but I would to raise a possible memory leak

ccoVeille (approved):
A few remarks but OK

pkg/gofr/cmd/terminal/spinner.go
vipul-rawat: Could having a context in Spin() function and stopping the ticker on context.Done resolve this issue?

vipul-rawat: And when someone does not call Stop() the spinner will always spin and will never stop as the control flow will not be passed back to the main go routine.

pkg/gofr/cmd/terminal/progress.go
vipul-rawat: Returning error to let users figure out if the progress bar was initialized! 
Initial thoughts were to not handle any error as we would like the process to continue even if the progress bar is not initialized as it is just a visualization effect.

examples/sample-cmd/main.go
```diff
@@ -23,6 +25,43 @@ func main() {
 		return fmt.Sprintf("Hello %s!", c.Param("name")), nil
 	})
 
+	app.SubCommand("spinner", func(ctx *gofr.Context) (interface{}, error) {
+		// initialize the spinner
+		spinner := terminal.NewDotSpinner(ctx.Out).Spin(ctx).Spin(ctx)
```
ccoVeille, line 30: Spin().Spin() ?

The methods names could be confusing

examples/sample-cmd/main.go
```diff
@@ -23,6 +25,43 @@ func main() {
...
+
+		select {
+		case <-ctx.Done():
+			spinner.Stop()
+			return nil, ctx.Err()
+		case <-time.After(2 * time.Second):
+			spinner.Stop()
+		}
```
ccoVeille, line 38: I would use this,it would prevent people to forget the call to Stop if a new `case` is added

```suggestion
		defer spinner.Stop()
		select {
		case <-ctx.Done():
			return nil, ctx.Err()
		case <-time.After(2 * time.Second):
		}
```

pkg/gofr/cmd/terminal/progress.go
```diff
@@ -0,0 +1,120 @@
...
+	return p, nil
+}
+
+func (p *ProgressBar) Incr(i int64) bool {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if p.current < p.total {
+		p.current += i
+		p.current = min(p.current, p.total)
```
ccoVeille, line 58: current could be a sync/atomic.Int64

It would avoid the mutex and the defer

https://pkg.go.dev/sync/atomic#Int64

But the min thing might need to be rewritten, maybe when reading the value

```suggestion
		p.current.Add(i)
		p.current = min(p.current, p.total)
```

vipul-rawat: But we also want to synchronize the painting of the progress bar that is being called in the updateProgressBar(), so mutex is either way mutex may be required.

examples/sample-cmd/main.go
vipul-rawat: I wanted to chain the Stop method too, cause doing that we can write one defer command to start and stop the spinner.

I can return nil from the Spin() which starts the spinner.

pkg/gofr/cmd/terminal/progress.go
ccoVeille: Oh that's right. It's a valuable comment you could add to the code

vipul-rawat: Added.

pkg/gofr/cmd/terminal/spinner.go
vipul-rawat: Closing the ticker when the context is done.

aryanmehrotra (approved):

PR #1165, opened 2024-10-30
Custom http response headers

### LINKED ISSUE #1130: how to return custom http response data and header
Hey, 

I tried gofr for one of my Projects. I'm realy happy with it, but now I faced a big blocking point to continue with gofr. 

I want to return a custom http response. 

Data: string
And a lot of http headers

But at the moment I can't find the API to do it. I analyzed the source code and found the Responder and also the response.Raw and response.File but that's not sufficient. I have to add more custom headers. So only way I see at the moment is CustomMiddleware but from my point of view this feels wrong. 

Because it's not possible to attach a middleware to a dedicated route only globally. 

Has anybody a hit / idea for me?

rigved-telang:
**Description:**

-   Added a new struct called Response which can be used to set custom headers to be sent back with response
-   Closes #1130 

**Additional Information:**

-   Usage:
```
headers := map[string]string{
	"X-Custom-Header":  "CustomValue",
	"X-Another-Header": "AnotherValue",
}

return response.Response{
	Data:    "Hello World from new Server",
	Headers: headers,
}, nil
```

**Checklist:**


pkg/gofr/handler_test.go
```diff
@@ -99,6 +99,71 @@ func TestHandler_ServeHTTP_Panic(t *testing.T) {
...
+			err:        nil,
+			statusCode: http.StatusOK,
+			body:       `{"message":"Hello, World!"}`,
+		},
+		{
+			desc:       "No headers, method is GET, data is simple string, no error",
+			method:     http.MethodGet,
+			data:       "simple string",
+			headers:    nil,
+			err:        nil,
```
ccoVeille, line 137: I wouldn't have set parameters to their default value

```suggestion
```

pkg/gofr/http/response/response.go
```diff
@@ -0,0 +1,6 @@
+package response
+
+type Response struct {
+	Data    interface{}       `json:"data"`
```
ccoVeille, line 4:
```suggestion
	Data    any               `json:"data"`
```

pkg/gofr/handler_test.go
rigved-telang: I have removed the setting of default values explicitly.

pkg/gofr/http/response/response.go
rigved-telang: Done

go.sum
shashank-zopsmart, line 1: Any specific reason to update the go.sum?

pkg/gofr/handler.go
```diff
@@ -109,6 +109,14 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		err = gofrHTTP.ErrorPanicRecovery{}
 	}
 
+	// **Handle Custom Headers** if `result` is a `Response`.
+	if resp, ok := result.(response.Response); ok {
+		for key, value := range resp.Headers {
+			w.Header().Set(key, value)
```
shashank-zopsmart, line 115: Can we add logic to skip headers that gofr already sets, such as 'Content-type'?

shashank-zopsmart (changes requested):

go.sum
rigved-telang: None. When I ran `go install golang.org/x/tools/cmd/goimports@latest`, it updated the go.sum, I can. remove it if you think it is necessary.

pkg/gofr/handler.go
rigved-telang: `if resp, ok := result.(response.Response); ok {

	for key, value := range resp.Headers {
		if w.Header().Get(key) == "" {
			w.Header().Set(key, value)
		}
	}
	result = resp.Data
}`
How about this, we'll check if a Header has been set and then set it only if it hasn't been set.

Umang01-hash: Yes maybe this can help

go.sum
shashank-zopsmart: Sure, in my opinion, go.mod & go.sum should only be updated if there are dependency update that are adding some value to the feature/enhancement/bug fix in the PR

pkg/gofr/handler.go
shashank-zopsmart: LGTM.

shashank-zopsmart: @Umang01-hash @aryan-mehrotra-zs @vipul-rawat How about having a list of non overridable response headers (or response headers that can only be overridden by some special methods provided by Gofr)? So that a check can be implemented for those headers only. Intention behind is that, user deliberately wants to override those headers (such as content-type) but since Gofr is internally setting those headers, so the overall working of the framework should not break.

**PS:** If required, this can be done as future enhancement. For now, this PR is good.

go.sum
rigved-telang: Understood. Will remove the additions in go.sum.

pkg/gofr/handler.go
aryanmehrotra: I would suggest if the user has provided some header which is already set by gofr that should also be over-ridden, if user doesn't set - the default which is set by gofr is used.

rigved-telang:
@ccoVeille @aryanmehrotra @Umang01-hash @shashank-zopsmart I have extracted logError and handleWebSocketUpgrade to reduce the Cognitive Complexity error which was being raised by codeclimate. Kindly review that as well.

pkg/gofr/handler.go
```diff
@@ -101,14 +91,21 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		if websocket.IsWebSocketUpgrade(r) {
-			// Do not respond with HTTP headers since this is a WebSocket request
-			return
-		}
+		handleWebSocketUpgrade(r)
 	case <-panicked:
 		err = gofrHTTP.ErrorPanicRecovery{}
 	}
 
+	// **Handle Custom Headers** if `result` is a `Response`.
```
ccoVeille, line 99: I'm unsure about the formatting here

In a godoc, I would have understand

Here I don't get what it brings.

I might not have raised this, but I also face a misunderstanding about the fact I don't get why there is a need to use emphasis on this

I could have understood if it was


```suggestion
	// handle **Custom Headers** if `result` is a `Response`.
```

But then, I would be lost anyway. Why would we need to emphasis on the fact it's custom headers, maybe you could explain

And finally, why using emphasis and not simply use quote.

```suggestion
	// handle "Custom Headers" if `result` is a `Response`.
```

Thanks for reading this almost useless feedback about a comment 😅😬😂

pkg/gofr/handler.go
```diff
@@ -101,14 +91,21 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if resp, ok := result.(response.Response); ok {
+		for key, value := range resp.Headers {
+			if w.Header().Get(key) == "" {
+				w.Header().Set(key, value)
+			}
+		}
+		result = resp.Data
+	}
```
ccoVeille, line 107: Maybe a comment explaining the reason behind this.

Here is my understanding:

There is a need to propagate the header to the caller.

Then explain existing headers are not overwritten.

For this exact reason, I would create a method in


```suggestion
	if resp, ok := result.(response.Response); ok {
		resp.PropagateHeaders(w)
		result = resp.Data
	}
```

With something like this in response.Response struct

```go
func(response.Response resp) PropagateHeaders(w *http.Response)
	for key, value := range resp.Headers {
		if w.Header().Get(key) != "" {
			// do not overwrite existing header
			continue
		}
		w.Header().Set(key, value)
	}
}
```

I wrote this on a phone, there might be some errors 😅

pkg/gofr/handler.go
```diff
@@ -150,3 +147,18 @@ func panicRecoveryHandler(re any, log logging.Logger, panicked chan struct{}) {
 		StackTrace: string(debug.Stack()),
 	})
 }
+
+// Log the error(if any) with traceID and errorMessage.
+func (h handler) logError(traceID string, err error) {
+	if err != nil {
+		errorLog := &ErrorLogEntry{TraceID: traceID, Error: err.Error()}
+		h.container.Logger.Error(errorLog)
+	}
```
ccoVeille, line 156:
```suggestion
	if err != nil {
		return
	}
	errorLog := &ErrorLogEntry{TraceID: traceID, Error: err.Error()}
	h.container.Logger.Error(errorLog)
```

pkg/gofr/handler_test.go
```diff
@@ -99,6 +99,68 @@ func TestHandler_ServeHTTP_Panic(t *testing.T) {
 	assert.Contains(t, w.Body.String(), http.StatusText(http.StatusInternalServerError), "TestHandler_ServeHTTP_Panic Failed")
 }
 
+func TestHandler_ServeHTTP_WithHeaders(t *testing.T) {
+	testCases := []struct {
+		desc       string
+		method     string
+		data       interface{}
```
ccoVeille, line 106:
```suggestion
		data       any
```

pkg/gofr/handler.go
rigved-telang: @shashank-zopsmart @Umang01-hash what do you guys think, should we extract the setting of custom headers to another function or a simple if condition should suffice in the ServerHTTP func itself

pkg/gofr/handler.go
rigved-telang: I'd just extracted the existing code to a new function, also isn't the logic wrong here, the error should get logged when it is not nil right?

ccoVeille: Indeed, I edited my post it should be OK now

pkg/gofr/handler.go
Umang01-hash: @rigved-telang extracting the setting of custom headers to another function is not a bad idea as it would lead to a more clear code and auto explain what the function is intended to do

Umang01-hash:
@rigved-telang  can you please also resolve this linter ?

Screenshot 2024-11-05 at 3 57 05 PM

pkg/gofr/handler.go
rigved-telang: Hi @Umang01-hash apologies for the inactiivity on the issue, I came to my hometown for Diwali and thus was not able to work on it. I will make the required changes within 4-5 days as I will be vacationing for another day. Thanks.

ccoVeille: It's OK, but thanks for letting us know

rigved-telang:
@Umang01-hash I have resolved all the comments, kindly review

pkg/gofr/http/response/response.go
```diff
@@ -0,0 +1,21 @@
+package response
+
+import (
+	"net/http"
+)
+
+type Response struct {
+	Data    any               `json:"data"`
+	Headers map[string]string `json:"-"`
```
ccoVeille, line 9: Maybe rename to this

```suggestion
	CustomHeaders map[string]string `json:"-"`
```

This way the variable and method SetCustomHeaders names will be consistent.

Just a suggestion, it's up to you

ccoVeille (approved):

ccoVeille (approved):

ccoVeille:
@Umang01-hash you changed back from CustomHeaders to Headers

I'm fine with it, buf maybe you could drop a comment explaining the reason 😅

Umang01-hash:
> @Umang01-hash you changed back from CustomHeaders to Headers
> 
> I'm fine with it, but maybe you could drop a comment explaining the reason 😅

Hey @ccoVeille I just did that in order to correct the explanation of the PR. @rigved-telang mentioned the usage like:

```go
headers := map[string]string{
	"X-Custom-Header":  "CustomValue",
	"X-Another-Header": "AnotherValue",
}

return response.Response{
	Data:    "Hello World from new Server",
	Headers: headers,
}, nil
```

But when i was using it i need to do :

```go
return response.Response{
	Data:    "Hello World from new Server",
	CustomHeaders: headers,
}, nil
```

So i thought maybe he mistakenly left it. Sorry if this is wrong. Should i revert it back?

ccoVeille:
Thanks for your reply @Umang01-hash 

Honestly I don't know. I raised the point [here](https://github.com/gofr-dev/gofr/pull/1165#discussion_r1835683157)

Then @rigved-telang changed it according to my suggestion.

Now, you changed it back. I thought it was made on purpose, but apparently no.

I'm fine with both, maybe @aryanmehrotra @vipul-rawat @srijan-27 could share their thoughts.

Umang01-hash:
> Thanks for your reply @Umang01-hash
> 
> Honestly I don't know. I raised the point [here](https://github.com/gofr-dev/gofr/pull/1165#discussion_r1835683157)
> 
> Then @rigved-telang changed it according to my suggestion.
> 
> Now, you changed it back. I thought it was made on purpose, but apparently no.
> 
> I'm fine with both, maybe @aryanmehrotra @vipul-rawat @srijan-27 could share their thoughts.

IMHO, I think we don't need to add prefix `Custom` to `Headers` field of the response struct because the main objective of this struct is to allow users to send headers that are not present by default in GoFr or that may give the API clients additional information regarding the response.

@aryanmehrotra @vipul-rawat your suggestions are welcomed here!

aryanmehrotra:
Just Header seems Okay, Having the Key CustomHeader seems overkill.

rigved-telang:
> @rigved-telang also there is one comment left by @aryanmehrotra i.e.
> 
> **I would suggest if the user has provided some header which is already set by gofr that should also be over-ridden, if user doesn't set - the default which is set by gofr is used.**
> 
> Please address it, as per the current behaviour we are not overwritting any headers set by GoFr:
> 
> Screenshot 2024-11-11 at 2 20 56 PM

This would essentially mean that we set the header without checking for whether there is an existing value for that header right? I will make the change immediately but just wanted to confirm that that is what we want.

rigved-telang:
@Umang01-hash @aryanmehrotra I have changed the logic such that manually entered headers always override the default headers, kindly review

ccoVeille (approved):

shashank-zopsmart:
> @rigved-telang also there is one comment left by @aryanmehrotra i.e.
> 
> **I would suggest if the user has provided some header which is already set by gofr that should also be over-ridden, if user doesn't set - the default which is set by gofr is used.**
> 
> Please address it, as per the current behaviour we are not overwritting any headers set by GoFr:
> 
> Screenshot 2024-11-11 at 2 20 56 PM

@Umang01-hash In context to above comment, `Content-type` header is not being overridden. Is it fine? If yes, then it could be possible that in near future, there could be some other headers as well that will not be overridden using this feature.

Umang01-hash:
> > @rigved-telang also there is one comment left by @aryanmehrotra i.e.
> > **I would suggest if the user has provided some header which is already set by gofr that should also be over-ridden, if user doesn't set - the default which is set by gofr is used.**
> > Please address it, as per the current behaviour we are not overwritting any headers set by GoFr:


> @Umang01-hash In context to above comment, `Content-type` header is not being overridden. Is it fine? If yes, then it could be possible that in near future, there could be some other headers as well that will not be overridden using this feature.

@shashank-zopsmart the `Content-Type` header is not being overridden because GoFr currently supports only JSON format for exporting responses, we need to add support for other formats like `XML` so that we can add this functionality to override the content-type header else it will not make much sense to do it right now.

Currently treating this as an exception we can move forward with the PR and add this support later.

rigved-telang:
> > > @rigved-telang also there is one comment left by @aryanmehrotra i.e.
> > > **I would suggest if the user has provided some header which is already set by gofr that should also be over-ridden, if user doesn't set - the default which is set by gofr is used.**
> > > Please address it, as per the current behaviour we are not overwritting any headers set by GoFr:
> 
> > @Umang01-hash In context to above comment, `Content-type` header is not being overridden. Is it fine? If yes, then it could be possible that in near future, there could be some other headers as well that will not be overridden using this feature.
> 
> @shashank-zopsmart the `Content-Type` header is not being overridden because GoFr currently supports only JSON format for exporting responses, we need to add support for other formats like `XML` so that we can add this functionality to override the content-type header else it will not make much sense to do it right now.
> 
> Currently treating this as an exception we can move forward with the PR and add this support later.

@Umang01-hash FYI, currently even the default headers set by gofr will be overwritten as you had pointed out in the last conversation to execute what @aryanmehrotra, my code block I simply write to the Headers() whatever headers are entered by the users.

Umang01-hash (approved):

aryanmehrotra (approved):

aryanmehrotra (approved):

PR #1173, opened 2024-11-06
Configurable MySQL Charset via DB_CHARSET Environment Variable

surajit-zs:
- Feature: Added a new environment variable `DB_CHARSET` to make the MySQL character set configurable.
- Default Value: The default value for `DB_CHARSET` is set to `utf8`.
- Usage: Users can set `DB_CHARSET=utf8mb4` to enable full Unicode support, including emojis and special characters.


pkg/gofr/datasource/sql/sql.go
```diff
@@ -159,18 +160,24 @@ func getDBConfig(configs config.Config) *DBConfig {
...
 	}
 }
 
 func getDBConnectionString(dbConfig *DBConfig) (string, error) {
 	switch dbConfig.Dialect {
 	case "mysql":
-		return fmt.Sprintf("%s:%s@tcp(%s:%s)/%s?charset=utf8&parseTime=True&loc=Local&interpolateParams=true",
+		if dbConfig.Charset == "" {
+			dbConfig.Charset = "utf8"
+		}
```
ccoVeille, line 172: While I like your change, I think you somehow spotted an existing issue.

It might be considered as a breaking changes, but I would expect the default value to be `utf8mb4`

But it's more a question for @aryanmehrotra @vipul-rawat @srijan-27 @Umang01-hash that could and should be addressed in another PR.

Except that, your PR is fine and could be merged as-is.

pkg/gofr/datasource/sql/sql.go
```diff
@@ -159,18 +160,24 @@ func getDBConfig(configs config.Config) *DBConfig {
...
 		// only for postgres
 		SSLMode: configs.GetOrDefault("DB_SSL_MODE", "disable"),
+		Charset: configs.Get("DB_CHARSET"),
 	}
 }
 
 func getDBConnectionString(dbConfig *DBConfig) (string, error) {
 	switch dbConfig.Dialect {
 	case "mysql":
-		return fmt.Sprintf("%s:%s@tcp(%s:%s)/%s?charset=utf8&parseTime=True&loc=Local&interpolateParams=true",
```
ccoVeille, line 168: Also, from my perspective we could consider adding also a constant for being to configure the client `loc`

So it could also be set to UTC or anything needed

I had this exact need for an old database.

This is also a suggestion that could be addressed in another PR

Cc @aryanmehrotra @srijan-27 @vipul-rawat @Umang01-hash

ccoVeille (approved):

pkg/gofr/datasource/sql/sql.go
Umang01-hash: @ccoVeille I agree with your suggestion to change the default value to `utf8mb4`as it supports a broader range of Unicode characters and aligns with modern applications needs.

pkg/gofr/datasource/sql/sql.go
Umang01-hash: Here is the issue for both of your suggestions:

https://github.com/gofr-dev/gofr/issues/1174

Umang01-hash (approved):

PR #1179, opened 2024-11-07
Add Middleware Support with Container Access to UseMiddlewareWithContainer

### LINKED ISSUE #1176: Support Middleware with UseContainerMiddleware Function
Add a method, UseContainerMiddleware, that allows adding a single middleware function that takes both the application’s container and the HTTP request as parameters. This will allow us to apply additional context-aware processing, such as authorization, logging, or validation, on a per-request basis.

Proposed Solution:
Implement UseContainerMiddleware, which will add a single middleware function utilizing the container.
Ensure UseContainerMiddleware does not interfere with the existing UseMiddleware, enabling both methods to be used interchangeably as needed.
The UseContainerMiddleware method would look like this:

```go
func (a *App) UseContainerMiddleware(fn func(c *container.Container, r *http.Request) (int, error)) {
	a.httpServer.router.Use(func(h http.Handler) http.Handler {
		return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {

			// Execute user-defined middleware function with the shared container
			status, err := fn(a.container, request)
			if err != nil {
				// Use the returned status code for custom responses
				http.Error(writer, err.Error(), status)
				return
			}

			// Call the next handler in the chain if no error
			h.ServeHTTP(writer, request)
		})
	})
}

```
I’m currently testing this implementation to ensure it meets all requirements. If the tests are successful, I’ll create a PR for review. Please let me know if you have any concerns.

surajit-zs:
- Closes #1176 
This pull request introduces a new method UseMiddlewareWithContainer to the App structure, allowing middleware functions to have access to the container instance. This feature is useful for scenarios where middleware needs to interact with shared resources, such as databases or caches, which are stored in the container.

- New Method: Added UseMiddlewareWithContainer to App, which allows middleware functions to access the container and the handler.
- Middleware Signature: The middleware accepts two parameters:
`c *container.Container`: The container instance.
`handler http.Handler`: The next handler in the chain.
- Routing Enhancement: This method enables any middleware to modify or enhance request handling using container services or resources before passing control to the next handler.


Umang01-hash:
> @Umang01-hash could you please check why this workflow is failing?
> 
> I don't think it's because of this PR changes.

Sure will check and let you know

Umang01-hash (approved):

aryanmehrotra (approved):

vipul-rawat (approved):

PR #1199, opened 2024-11-13
Update page.md to included MariaDB

robertsilen:
**Description:**

Suggesting to include MariaDB in page.md, the Quick Start guide for Connecting MariaDB/MySQL

**Additional Information:**

- MariaDB is the default used by most distros, so might as well have it default in this guide too. 
- I have not tested Gofr compatibility with MariaDB - I'm presuming this works.


docs/quick-start/connecting-mysql/page.md
```diff
@@ -1,26 +1,32 @@
...
 ```bash
 docker run --name gofr-mysql -e MYSQL_ROOT_PASSWORD=root123 -e MYSQL_DATABASE=test_db -p 3306:3306 -d mysql:8.0.30
 ```
 
-Access `test_db` database and create table customer with columns `id` and `name`
+Access `test_db` database and create table customer with columns `id` and `name`. Change mysql to mariadb as needed: 
 
 ```bash
-docker exec -it gofr-mysql mysql -uroot -proot123 test_db -e "CREATE TABLE customers (id INT AUTO_INCREMENT PRIMARY KEY, name VARCHAR(255) NOT NULL);"
+docker exec -it gofr-mariadb mariadb -uroot -proot123 test_db -e "CREATE TABLE customers (id INT AUTO_INCREMENT PRIMARY KEY, name VARCHAR(255) NOT NULL);"
```
Umang01-hash, line 22: IMHO, `MySQL` is a more popular and used database but since we are supporting both what about we keep the docker command to use `gofr-mysql` i.e. **by default let's show the commands for MySQL db** and in a note we can mention to run `mariaDB` please change the name to this...

docs/quick-start/connecting-mysql/page.md
```diff
@@ -1,26 +1,32 @@
-# Connecting MySQL
+# Connecting MariaDB/MySQL
 
-Just like Redis GoFr also supports connection to SQL(MySQL and Postgres) databases based on configuration variables.
+Just like Redis GoFr also supports connection to SQL(MariaDB, MySQL and Postgres) databases based on configuration variables.
 
 ## Setup
 
-Users can run MySQL and create a database locally using the following Docker command:
+Users can run MariaDB/MySQL and create a database locally using the following Docker command:
```
Umang01-hash, line 7: I will love if we can just reverse the order of showing docker commands to run `mariaDB` and `MySQL`, since the title of page is **connecting with mysql** we should keep it above and shift the `MariaDB` command beneath that.

Umang01-hash (changes requested):

Umang01-hash:
@robertsilen Please resolve the review comments so that we can proceed to merge your PR.
Let us know in case of any help required.

Umang01-hash (approved):

coolwednesday (approved):

PR #1211, opened 2024-11-16
Fix:  panic in google pubsub, recieving on closed channel

### LINKED ISSUE #1210: Google Pubsub panic during message handling
Encountering a panic error with the message “send on closed channel” while messages are being handled. This issue is occuring when events arrive in quick succession, leading to unexpected channel closure during message processing
Screenshot 2024-11-16 at 19 32 06
Screenshot 2024-11-15 at 18 40 36

vipul-rawat:
## Pull Request Template

**Description:**

closes #1210 

Using a map of channels with the topic as key, calling the Google client subscribe function only when the subscription has not been started. The channel's closing is done in the close function when shutdown is initiated.


aryanmehrotra (approved):

pkg/gofr/datasource/pubsub/google/google_test.go
```diff
@@ -174,17 +175,17 @@ func Test_validateConfigs(t *testing.T) {
...
+		receiveChan: make(map[string]chan *pubsub.Message),
 	}
 
 	err := g.Close()
 
 	require.NoError(t, err)
 
 	// client empty
-	g = &googleClient{}
+	g = &googleClient{receiveChan: make(map[string]chan *pubsub.Message)}
```
ccoVeille, line 188: I tend to define the things like that in newConfig function you can pass parameters or an Init method on the googleClient

This way the private data prerequisites are not spread in multiple files

pkg/gofr/datasource/pubsub/google/google.go
```diff
@@ -117,63 +122,71 @@ func (g *googleClient) Publish(ctx context.Context, topic string, message []byte
...
 	defer span.End()
 
-	g.metrics.IncrementCounter(ctx, "app_pubsub_subscribe_total_count", "topic", topic, "subscription_name", g.Config.SubscriptionName)
+	g.metrics.IncrementCounter(spanCtx, "app_pubsub_subscribe_total_count", "topic", topic, "subscription_name", g.Config.SubscriptionName)
 
-	t, err := g.getTopic(ctx, topic)
-	if err != nil {
-		return nil, err
-	}
+	if _, ok := g.subStarted[topic]; !ok {
```
ccoVeille, line 132: This if is covering so many lines.

I would have moved the code currently inside to an unexported method

PR #1214, opened 2024-11-18
Fix: Correcting time unit for logs of http-service as well as datasources

### LINKED ISSUE #1197: BUG: Change Time Unit from Milliseconds to Microseconds
Change Time Unit from Milliseconds to Microseconds

Ref: https://github.com/gofr-dev/gofr/blob/368e3f5db21994be1a8b9a360567b689b23aec93/pkg/gofr/service/new.go#L173

coolwednesday:
## Pull Request Template

**Description:**
-   Closes #1197 
-  Previously, all the duration to be displayed in the logs, were calculated in milliseconds but while printing it through prettyprint, we use to display as microseconds, which is wrong. Hence corrected the calculations to be done in microseconds..

**Checklist:**


ccoVeille:
Please update the PR description

```diff
-we use to display as microseconds, which is wrong. Hence corrected the calculations to be done in Microseconds.
+we use to display as milliseconds, which is wrong. Hence corrected the calculations to be done in microseconds.
```

ccoVeille (approved):

ccoVeille:
> > Please update the PR description
> > 
> > ```diff
> > -we use to display as microseconds, which is wrong. Hence corrected the calculations to be done in Microseconds.
> > +we use to display as milliseconds, which is wrong. Hence corrected the calculations to be done in microseconds.
> > ```
> 
> @ccoVeille , I mean what I wrote. The log displays 8 microseconds where in actually gofr calculates it as 8 milliseconds. So we are correcting that, to corrrectly show ....8,000 µs.

OK, so I think I'm lost.

Let's reboot.

Your text says:

>  we use to display as microseconds, which is wrong. Hence corrected the calculations to be done in microseconds.

So it means
- it was in **microseconds**
- we fixed to use **microseconds**

The code :
- was calling **milliseconds**
- is updated to call **microseconds**

So for me, there is a problem with the text. The code is fine

aryanmehrotra (changes requested):
Missing changes for transaction in SQL DB and openTSDB

aryanmehrotra (approved):

PR #1228, opened 2024-11-20
deprecate UseMiddlewarWithContainer and remove it from tests and docs

Umang01-hash:
## Pull Request Template

**Description:**

-   Deprecate feature `UseMiddlewareWithContainer`.

### Reason for Deprecation:
We are deprecating the `UseMiddlewareWithContainer` feature due to architectural considerations. Allowing middleware to access databases or caches introduces unnecessary complexities.

By removing this support, we aim to encourage a cleaner separation of concerns, ensuring that middleware remains focused on its intended purpose. This change will help in maintaining the simplicity and integrity of our framework.

**Checklist:**


pkg/gofr/gofr_test.go
```diff
@@ -620,57 +620,6 @@ func Test_UseMiddleware(t *testing.T) {
 	assert.Equal(t, "applied", testHeaderValue, "Test_UseMiddleware Failed! header value mismatch.")
 }
 
-// Test the UseMiddlewareWithContainer function.
```
ccoVeille, line 623: I disagree

As long it's here, so as long the code is not removed, the test must remain

People are using the deprecated method, and might never migrate

pkg/gofr/gofr.go
```diff
@@ -648,6 +648,8 @@ func (a *App) UseMiddleware(middlewares ...gofrHTTP.Middleware) {
 //
 // The `middleware` function receives the container and the handler, allowing
 // the middleware to modify the request processing flow.
+// Deprecated: UseMiddlewareWithContainer will be removed in a future release.
+// Please use the UseMiddleware method that does not depend on the container.
```
ccoVeille, line 652:
```suggestion
// Please use the [*App.UseMiddleware] method that does not depend on the container.
```

ccoVeille (changes requested):

pkg/gofr/gofr_test.go
Umang01-hash: @ccoVeille I completely agree that Test should remain as long as code is present.

But since we were using the `app.UseMiddlewareWithContainer` methods in test we were getting a warning that **deprecated code is still being used.**

So removed the test for the same.

pkg/gofr/gofr.go
Umang01-hash: changed

pkg/gofr/gofr_test.go
ccoVeille: The call can be ignored with a nolint

ccoVeille (approved):

pkg/gofr/gofr_test.go
ccoVeille: > @ccoVeille I completely agree that Test should remain as long as code is present.
> 
> But since we were using the `app.UseMiddlewareWithContainer` methods in test we were getting a warning that **deprecated code is still being used.**
> 
> So removed the test for the same.

So the warning was not blocking for the CI ?

Or the warning was only local to your IDE?

Umang01-hash: Yes i checked it was local for my Goland IDE.
Did the change and ran linter, all checks passed

coolwednesday (approved):

Issue #1234, opened 2024-11-22
Dgraph giving panic


Umang01-hash:
**Describe the bug**
When connecting DGraph and calling it's method getting a panic:

<img width="1048" alt="Screenshot 2024-11-22 at 1 33 03 PM" src="https://github.com/user-attachments/assets/4726a3d6-b98c-485a-aa4b-10e90cab6ec6">

**To Reproduce**
Steps to reproduce the behavior, if applicable:

1. The code is

```go
   package main

import (
	"encoding/json"
	"fmt"

	"github.com/dgraph-io/dgo/v210/protos/api"

	"gofr.dev/pkg/gofr"

	"gofr.dev/pkg/gofr/datasource/dgraph"
)

func main() {
	// Create a new application
	app := gofr.New()

	db := dgraph.New(dgraph.Config{
		Host: "localhost",
		Port: "9080",
	})

	// Connect to Dgraph running on localhost:9080
	app.AddDgraph(db)

	// Add routes for Dgraph operations
	app.POST("/dgraph", DGraphInsertHandler)
	app.GET("/dgraph", DGraphQueryHandler)

	// Run the application
	app.Run()
}

// DGraphInsertHandler handles POST requests to insert data into Dgraph
func DGraphInsertHandler(c *gofr.Context) (interface{}, error) {
	// Example mutation data to insert into Dgraph
	mutationData := `
		{
			"set": [
				{
					"name": "GoFr Dev"
				},
				{
					"name": "James Doe"
				}
			]
		}
	`

	// Create an api.Mutation object
	mutation := &api.Mutation{
		SetJson:   []byte(mutationData), // Set the JSON payload
		CommitNow: true,                 // Auto-commit the transaction
	}

	// Run the mutation in Dgraph
	response, err := c.DGraph.Mutate(c, mutation)
	if err != nil {
		return nil, err
	}

	return response, nil
}

// DGraphQueryHandler handles GET requests to fetch data from Dgraph
func DGraphQueryHandler(c *gofr.Context) (interface{}, error) {
	// A simple query to fetch all persons with a name in Dgraph
	response, err := c.DGraph.Query(c, "{ persons(func: has(name)) { uid name } }")
	if err != nil {
		return nil, err
	}

	// Cast response to *api.Response (the correct type returned by Dgraph Query)
	resp, ok := response.(*api.Response)
	if !ok {
		return nil, fmt.Errorf("unexpected response type")
	}

	// Parse the response JSON
[...]

PR #1235, opened 2024-11-22
add missing tracer for Dgraph datasource

Umang01-hash:
## Pull Request Template

**Description:**

-   Closes issue #1234 
-  Add missing tracer for Dgraph datasource.

**Checklist:**


pkg/gofr/external_db.go
```diff
@@ -121,6 +121,10 @@ func (a *App) AddDgraph(db container.DgraphProvider) {
 	db.UseLogger(a.Logger())
 	db.UseMetrics(a.Metrics())
 
+	tracer := otel.GetTracerProvider().Tracer("gofr-dgraph")
+
+	db.UseTracer(tracer)
```
ccoVeille, line 126: While this change might fix the issue, there might be a problem somewhere else.

A tracer cannot be considered as mandatory.

I mean it's a feature to have a tracer. What was the panic ?

There must be code that implies a tracer to be always present, while it could be simply absent.

Umang01-hash: @ccoVeille I agree with your suggestion that making tracer necessary can be avoided, but since we are implementing these external datasources ourselves and as a framework we feel that observability is crucial to application, so we try to add it everywhere possible. 

We just missed it in `DGraph` when adding for other datasources thatswhy this issue and this solution is here, but we will be extra careful when implementing datasources on our end so that it doesn't repeat.

ccoVeille: My problem is not functional.

I'm OK with having a tracer.

My problem is having code that panic if a tracer is not set.

Did you find where the code was panicking?

I mean was it something like foo.Bar where bar is nil, or a code that raise a panic when some variable is checked.

If it's the first option, the code needs to be updated to raise an error. Either a non blocking one, or a functional panic if the tracer is mandatory.

Also, this panic should be covered by a test that will help to either check there is no panic, or we have the expected panic

Here your change is about putting a patch on a wooden leg.

My 2 cents

Umang01-hash: @ccoVeille addressing your comments, I have added the checks before setting the attributes to the span(which was causing panic earlier). Now even if you don't provide a tracer code will not panic and work as expected.

pkg/gofr/datasource/dgraph/dgraph.go
```diff
@@ -283,9 +269,29 @@ func (d *Client) HealthCheck(ctx context.Context) (any, error) {
...
-	d.metrics.RecordHistogram(ctx, metricName, float64(query.Duration))
+func (d *Client) addTrace(ctx context.Context, method string) (context.Context, trace.Span) {
+	if d.tracer != nil {
+		tracedCtx, span := d.tracer.Start(ctx, fmt.Sprintf("dgraph-%v", method))
+
+		return tracedCtx, span
+	}
+
+	return ctx, nil
+}
```
ccoVeille, line 280: The code would be cleaner


```suggestion
	if d.tracer == nil {
		return ctx, nil
	}
	tracedCtx, span := d.tracer.Start(ctx, fmt.Sprintf("dgraph-%v", method))
	return tracedCtx, span
}
```

ccoVeille (approved):

ccoVeille:
It's a good refactoring

I think we could use it as a base for a metric struct that could be reused everywhere

Avoiding the boiler plate we have everywhere.

I think about a  short lived struct

If you agree about the need, and maybe the idea, we could create a new issue

ccoVeille:
By boiler plate, I mean code that is copy pasted and is pain to fix

Like

https://github.com/gofr-dev/gofr/pull/1214

Umang01-hash:
> It's a good refactoring
> 
> I think we could use it as a base for a metric struct that could be reused everywhere
> 
> Avoiding the boiler plate we have everywhere.
> 
> I think about a short lived struct
> 
> If you agree about the need, and maybe the idea, we could create a new issue

Sure @ccoVeille we can do that. Please go ahead and create the issue will be happy to assist resolving it.

ccoVeille:
In other framework it would be a middleware that is added on demand, without interference with the code

ccoVeille:
> > It's a good refactoring
> > 
> > I think we could use it as a base for a metric struct that could be reused everywhere
> > 
> > Avoiding the boiler plate we have everywhere.
> > 
> > I think about a short lived struct
> > 
> > If you agree about the need, and maybe the idea, we could create a new issue
> 
> Sure @ccoVeille we can do that. Please go ahead and create the issue will be happy to assist resolving it.

I created 

- #1244

pkg/gofr/datasource/dgraph/dgraph.go
```diff
@@ -26,7 +26,6 @@ type Config struct {
 // Client represents the Dgraph client with logging and metrics.
 type Client struct {
 	client  DgraphClient
-	conn    *grpc.ClientConn
```
coolwednesday, line 29: Can we get a context of why it existed and now why is it removed ?

pkg/gofr/datasource/dgraph/dgraph.go
```diff
@@ -200,18 +191,14 @@ func (d *Client) Mutate(ctx context.Context, mu any) (any, error) {
...
-
 	if err != nil {
 		d.logger.Error("dgraph mutation failed: ", err)
 		ql.PrettyPrint(d.logger)
+
 		return nil, err
 	}
 
-	d.sendOperationStats(ctx, ql, "dgraph_mutate_duration")
+	defer d.sendOperationStats(tracedCtx, start, mutationToString(mutation), "mutate", span, ql, "dgraph_mutate_duration")
```
coolwednesday, line 201: I feel we should remove defer as we are just returning the response and error after this.

pkg/gofr/datasource/dgraph/dgraph.go
```diff
@@ -283,9 +269,29 @@ func (d *Client) HealthCheck(ctx context.Context) (any, error) {
...
+	}
+
+	tracedCtx, span := d.tracer.Start(ctx, fmt.Sprintf("dgraph-%v", method))
+
+	return tracedCtx, span
+}
+
+func (d *Client) sendOperationStats(ctx context.Context, start time.Time, query, method string,
+	span trace.Span, queryLog *QueryLog, metricName string) {
+	duration := time.Since(start).Milliseconds()
```
coolwednesday, line 284: Microseconds

pkg/gofr/datasource/dgraph/dgraph.go
```diff
@@ -241,18 +227,18 @@ func (d *Client) Alter(ctx context.Context, op any) error {
 		Duration: duration,
 	}
 
-	span.SetAttributes(
-		attribute.String("dgraph.alter.operation", operation.String()),
-		attribute.Int64("dgraph.alter.duration", duration),
-	)
+	if span != nil {
```
coolwednesday, line 230: I might be wrong, but we are doing it in `line 286`, through `sendOperationStats()` called in `line 241` , here in this function. Is this necessary here ?

vipul-rawat (dismissed):

coolwednesday (changes requested):

pkg/gofr/datasource/dgraph/dgraph.go
Umang01-hash: It existed because we thought of storing the connection inside the `Client` struct but later changed the implementation and used it like this:

```go
conn, err := grpc.Dial(address, grpc.WithTransportCredentials(insecure.NewCredentials()))
if err != nil {
	d.logger.Errorf("error while connecting to Dgraph, err: %v", err)
	return
}
	
dg := dgo.NewDgraphClient(api.NewDgraphClient(conn))
d.client = NewDgraphClient(dg)

```

So this field **conn was unused**, hence removed it.

pkg/gofr/datasource/dgraph/dgraph.go
Umang01-hash: sure removed defer call to this method, and instead called it directly

pkg/gofr/datasource/dgraph/dgraph.go
Umang01-hash: thankyou changed

coolwednesday (approved):

pkg/gofr/datasource/dgraph/dgraph.go
Umang01-hash: We were doing squash and merge till now but decided to merge all commits from next release

aryanmehrotra (approved):

PR #1237, opened 2024-11-25
En/redis lock

### LINKED ISSUE #637: Redis lock
## Problem:
Now we have [go-redis](http://github.com/redis/go-redis/v9) and for implementing distributed lock (common case for Redis usage) users have to write custom implementations of Locks.

## Proposal:
To simplify the locks aquiring / releasing process would be better to have [redislock](https://github.com/bsm/redislock) out of the box.

## In the code usage may look like:
```go
lock, err := ctx.Redis().Locker().Obtain(ctx, "my-key", 100*time.Millisecond, nil)
```

Umang01-hash:
## Pull Request Template

**Description:**

-   Closes #637 

**Checklist:**


examples/http-server-using-redis/main.go
```diff
@@ -71,3 +76,31 @@ func RedisPipelineHandler(c *gofr.Context) (interface{}, error) {
...
+	if err != nil && !errors.Is(err, redis.Nil) {
+		return nil, err
+	}
+
+	// Increment the count under the lock
+	err = c.Redis.Set(c, countKey, count+1, redisExpiryTime*time.Minute).Err()
+	if err != nil {
+		return nil, err
+	}
+
```
aryanmehrotra, line 104: I am unsure if this would work
Because in the example we are acquiring a lock and then using c.Redis to get and set the values.

So, if two parallel requests come in - both of them will be processed without wait inspite of having the lock.

If this would have been container/pod level maybe this would have worked - but it should not be like that.

or are we acquiring the lock only on the my-lock-key - if so how it is being demonstrated here?

aryanmehrotra (changes requested):

Umang01-hash:
As mentioned by @ccoVeille we already have distributed lock implemented in redis's client so we won't need a separate library or method for using it.

Here is the link to documentation: https://redis.io/glossary/redis-lock/

Closing this issue as of now.

Issue #964, opened 2024-08-29
Fix Pipeline for external datasources


aryanmehrotra:
Linters and tests are not running for external datasources as they are different modules.
Need to check if the current command can be changed somehow to incorporate it or we need to run them separately.

coolwednesday:
To fix the pipeline :
- We wish to fix it in multiple PRs to first fix the datasources (linters & tests)
- Introducing new workflow to run tests and linters in datasources also.

Umang01-hash:
@coolwednesday assigned the issue to you.

PR #1245, opened 2024-11-27
Fixed linters & tests in Mongo, Solr, Badger datasources

coolwednesday:
## Pull Request Template

**Description:**

-   This is a prerequisite to #964
-    We need to fix linters and tests in all datasources before adding workflow for external datasources

**Checklist:**


pkg/gofr/datasource/solr/solr.go
```diff
@@ -153,70 +151,70 @@ func (c *Client) ListFields(ctx context.Context, collection string, params map[s
...
+	resp, span, err := c.call(ctx, http.MethodPost, url, nil, document)
 
-	c.sendOperationStats(ctx, &QueryLog{Type: "DeleteField", Url: url}, startTime, "delete-field", span)
+	c.sendOperationStats(ctx, &QueryLog{Type: "DeleteField", URL: url}, startTime, "delete-field", span)
 
 	return resp, err
 }
 
-// Response stores the response from SOLR
+// Response stores the response from SOLR.
```
ccoVeille, line 210:
```suggestion
// Response stores the response from Solr.
```

Solr is not an acronym, but a brand, written Solr

pkg/gofr/datasource/solr/solr_test.go
```diff
@@ -177,9 +177,9 @@ func Test_ClientUpdateField(t *testing.T) {
 func Test_ClientDeleteField(t *testing.T) {
 	s := setupClient(t)
 
-	body := bytes.NewBuffer([]byte(`{"delete-field":{
+	body := bytes.NewBufferString(`{"delete-field":{
```
ccoVeille, line 180: Did you use mirror linter to detect these?

ccoVeille (approved):

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -54,7 +54,7 @@ i.e. by default observability features gets initialised when used with GoFr.
 // client.UseLogger(loggerInstance)
 // client.UseMetrics(metricsInstance)
 // client.Connect().
-func New(c Config) *Client {
+func New(c *Config) *Client {
 	return &Client{config: c}
 }
```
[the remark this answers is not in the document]
coolwednesday: the golangci-lint returns : `hugeParam: c is heavy (96 bytes); consider passing it by pointer (gocritic)`
Hence made it a pointer.

pkg/gofr/datasource/solr/solr_test.go
coolwednesday: Yes, I did check that each such instance is changed according to the linter.

ccoVeille (approved):

vikash (approved):

pkg/gofr/datasource/solr/solr.go
```diff
@@ -35,8 +35,8 @@ type Client struct {
 // client := New(config)
 // client.UseLogger(loggerInstance)
 // client.UseMetrics(metricsInstance)
-// client.Connect()
-func New(conf Config) *Client {
+// client.Connect().
+func New(conf *Config) *Client {
```
vikash, line 39: This change is unnecessary and does not feel like a good idea. Config should not be a pointer.

ccoVeille: I agree.

https://github.com/gofr-dev/gofr/pull/1245#discussion_r1860112867

coolwednesday: I was trying to be consistent across all the datasources. Some datasources that have a heavy config struct are passed by reference. So was being consistent across datasources passing each config struct by reference.

ccoVeille: The same point has been raised by myself, on another PR

https://github.com/gofr-dev/gofr/pull/1254#discussion_r1862569427

ccoVeille: I don't know what should be done. I agree it's better to have a consistent say to do.

But having a pointer for a configuration looks strange, not, yes it has performance/memory issues

coolwednesday: Would be changing it to a pointer to struct.

ccoVeille (approved):

pkg/gofr/datasource/mongo/mongo_test.go
```diff
@@ -536,16 +536,16 @@ func TestClient_StartSession(t *testing.T) {
...
 		err = ses.CommitTransaction(context.Background())
 
-		assert.Nil(t, err)
+		require.NoError(t, err)
 
 		ses.EndSession(context.Background())
 
 		// Assert that there was no error
-		assert.Nil(t, err)
+		require.NoError(t, err)
```
Umang01-hash, line 548: We are using both `assert` as well as `require` in the test file. We can use any one of them throughout.

Umang01-hash (changes requested):

pkg/gofr/datasource/mongo/mongo_test.go
ccoVeille: Do you mean stop using assert?abd only use require package?

It seems a regression. Having tests that fails on first issue is not the best thing

Thi k about checking the http.Status, then the content of the JSON returned

If the check on http.Status fail, you have no idea what the problem is without tjejson content

coolwednesday: @Umang01-hash thoughts ?

Umang01-hash: ahhh! Got it 
Sorry my bad! Let's use both

Umang01-hash (approved):

vipul-rawat (approved):

PR #1246, opened 2024-11-27
Dockerfile Update: Build and Serve GoFr Website as Static Site Using GoFr static-server

aryanmehrotra:


Umang01-hash (dismissed):

vikash:
Instead of nginx, let's use the static server written using Gofr.

docs/Dockerfile
```diff
@@ -9,20 +9,19 @@ COPY  docs/references /app/src/app/docs/references
...
 
 RUN npm install
 RUN npm run build
 
-RUN addgroup -g 1001 -S nodejs
-RUN adduser -S nextjs -u 1001
-
-USER nextjs
+# Stage 2: Serve with Static Server
+FROM zopdev/static-server:latest
```
vikash, line 18: Using latest may have unintended consequences. We should fix the version instead.

vipul-rawat (approved):

Issue #1251, opened 2024-11-28
HTTP Metrics are not being exported for Static Server 


aryanmehrotra:
When a directory is added to serve the files from it. metrics should be exported about which endpoint was called - but no metrics are being exported.
<img width="1387" alt="Screenshot 2024-11-28 at 11 36 46 AM" src="https://github.com/user-attachments/assets/977ef7a7-a137-440e-afb7-ec86e481c94d">

PR #1255, opened 2024-11-29
Fix incorrect metrics when serving static files

Umang01-hash:
## Pull Request Template

**Description:**

-   Resolve issue #1251 

Screenshot 2024-11-29 at 2 59 57 PM

**Checklist:**


vikash:
We need to be careful here. If we put file path for each file instead of a pattern while calculating the http related metrics, it will become very very large.

pkg/gofr/http/middleware/metrics.go
```diff
@@ -21,11 +22,19 @@ type metrics interface {
...
 			start := time.Now()
 
 			srw := &StatusResponseWriter{ResponseWriter: w}
 
 			path, _ := mux.CurrentRoute(r).GetPathTemplate()
+
+			if path == "/" || strings.HasPrefix(path, "/static") || staticFileRegex.MatchString(r.URL.Path) {
+				path = r.URL.Path
+			}
+
```
ccoVeille, line 37: Please don't use regexp for this.

You can use [filepath.Ext](https://pkg.go.dev/path/filepath#Ext)

Something like this could work

```suggestion
			switch strings.ToLower(filepath.Ext(r.URL.Path)) {
				case "css", "js", "png", "jpg","jpeg","gif","ico","svg","txt","html","json","woff","woff2","ttf","eot","pdf":
					path = r.URL.Path
			}
			if path == "/" || strings.HasPrefix(path, "/static") {
				path = r.URL.Path
			}
```

pkg/gofr/http/middleware/metrics.go
```diff
@@ -21,11 +22,19 @@ type metrics interface {
 func Metrics(metrics metrics) func(inner http.Handler) http.Handler {
 	return func(inner http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Compile regex for common file extensions
+			staticFileRegex := regexp.MustCompile(`\.(css|js|png|jpg|jpeg|gif|ico|svg|txt|html|json|woff|woff2|ttf|eot|pdf)$`)
```
ccoVeille, line 26: also, if you want to keep a regex, move it away from the method to a var in package level

Otherwise, you will compute the regex on each metrics call

ccoVeille (changes requested):
regexp are evil

pkg/gofr/http/middleware/metrics.go
Umang01-hash: Sure i have used `filepath.Ext()` method to match it instead of regex as per your suggestion.

pkg/gofr/http/middleware/metrics_test.go
```diff
@@ -54,3 +55,36 @@ func TestMetrics(t *testing.T) {
...
+
+func TestMetrics_StaticFile(t *testing.T) {
+	mockMetrics := &mockMetrics{}
+
+	mockMetrics.On("RecordHistogram", mock.Anything, "app_http_response", mock.Anything,
+		[]string{"path", "/static/example.js", "method", "GET", "status", "200"}).Return(nil)
+
+	// Create a temporary static file for the test
+	tempDir := t.TempDir()
+	staticFilePath := tempDir + "/example.js"
```
ccoVeille, line 67: Can you add a test where the file is accessed with a query parameter. Something like ?v=42

Umang01-hash: @ccoVeille added the test where request is formed like this:

```go
	req := httptest.NewRequest(http.MethodGet, "/static/example.js?v=42", http.NoBody)
```

ccoVeille (approved):

vipul-rawat (approved):

PR #1225, opened 2024-11-19
Update gofr version in examples

Umang01-hash:
## Pull Request Template

**Description:**

-   Updated GoFr version to latest (v1.27.1) in examples.

**Checklist:**

-   [ ] I have formatted my code using  `goimport`  and  `golangci-lint`.
-   [ ] All new code is covered by unit tests.
-   [ ] This PR does not decrease the overall code coverage.
-   [ ] I have reviewed the code comments and documentation for clarity.

**Thank you for your contribution!**

PR #1256, opened 2024-11-29
remove go.mod files from examples

Umang01-hash:
## Pull Request Template

**Description:**

-   Remove `go.mod` files from remaining examples.

**Checklist:**


ccoVeille:
> Remove go.mod files from remaining examples.

This is the "what?", but nothing explain the "why?"

I mean, you explain what the commit is about, but I have no idea why you do such changes

Umang01-hash:
> > Remove go.mod files from remaining examples.
> 
> This is the "what?", but nothing explain the "why?"
> 
> I mean, you explain what the commit is about, but I have no idea why you do such changes

Did it regarding this comment:

https://github.com/gofr-dev/gofr/pull/1225#issuecomment-2498696256

go.mod
```diff
@@ -68,7 +68,10 @@ require (
 	github.com/googleapis/enterprise-certificate-proxy v0.3.4 // indirect
 	github.com/googleapis/gax-go/v2 v2.14.0 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.23.0 // indirect
+	github.com/hashicorp/errwrap v1.1.0 // indirect
+	github.com/hashicorp/go-multierror v1.1.1 // indirect
 	github.com/hashicorp/golang-lru/v2 v2.0.7 // indirect
+	github.com/jlaffaye/ftp v0.2.0 // indirect
```
aryanmehrotra, line 74: this should not be part of the gofr go.mod

aryanmehrotra (changes requested):

go.mod
ccoVeille: As the go.mod are removed in examples, isn't normal things are going up, and are added to main one no?

If this is the problem, we can leave a go.mod at the root of examples

Another solution is to move the examples folder into a testdata folder, as it's a special folder

aryanmehrotra: For examples like http-server we are using the datasources(sql,redis) which are already part part of `go.mod`.
and in example like add-filestore, we had a different go.mod as we were importing ftp which is not part of the gofr `go.mod` but it is implemented as an external datasource.

ccoVeille: Yes, exactly. So I agree there is something strange

Umang01-hash: Will do one thing, we will remove the go.mod from **http-server example** only so that this unwanted dependency doesn't get added to GoFr's `go.mod`

aryanmehrotra (approved):

Issue #1275, opened 2024-12-06
Interservice calls do not correctly set request headers


coolwednesday:
When making interservice calls no headers are set - i.e content-type as well.
Because of which when request reaches a gofr server without content-type body is not binded.

ssshekhu53:
@aryan-mehrotra-zs Can you please assign this issue to me

Umang01-hash:
@ssshekhu53 assigned the issue to you.

PR #1277, opened 2024-12-09
Updated createAndSendRequest to add headers

ssshekhu53:
- Added check for if the `Content-type` header is not set, then `application/json` will be set by default.
- Updated the test as well
- Refer #1275


coolwednesday (approved):

Umang01-hash (approved):

pkg/gofr/service/new_test.go
```diff
@@ -74,7 +74,7 @@ func TestHTTPService_createAndSendRequest(t *testing.T) {
 	// when params value is of type []string then last value is sent in request
 	resp, err := service.createAndSendRequest(ctx,
 		http.MethodPost, "test-path", map[string]interface{}{"key": "value", "name": []string{"gofr", "test"}},
-		[]byte("{Test Body}"), map[string]string{"header1": "value1"})
+		[]byte("{Test Body}"), map[string]string{"header1": "value1", "content-type": "application/json"})
```
aryanmehrotra, line 77: test needs to be updated with the following cases - 
1. We have added content-type but we are not verifying if the server which received the request received it with the content-type and the correct value.
2. We should be also adding a case where content-type is not set by the user and it automatically gets set.
3. Content-type is set with application/xml or something and it is not overrided.

aryanmehrotra (changes requested):

aryanmehrotra (approved):

Issue #1279, opened 2024-12-09
Avoid Logging 4XX Status Codes as Errors


surajit-zs:
### Problem
Currently, 4XX HTTP status codes are logged as errors, which can clutter the logs and mislead monitoring systems. Client-side issues should not be considered server errors.

### Expected Behavior
- 4XX status codes should be treated as info or excluded from error logs.
- Only 5XX status codes should be logged as errors.

### Proposed Solution
- Introduce conditional logging:
  - Log **5XX** status codes as errors.
  - Optionally log **4XX** as info or ignore them entirely.
<img width="1416" alt="image" src="https://github.com/user-attachments/assets/413ad72e-ea21-4029-87b2-14226f0409b2">

PR #1295, opened 2024-12-11
Fix/4xx logs

Umang01-hash:
## Pull Request Template

**Description:**

-   Closes issue #1279 

**Checklist:**

### DIFF
--- a/pkg/gofr/handler.go
+++ b/pkg/gofr/handler.go
@@ -158,28 +158,21 @@ func (h handler) logError(traceID string, err error) {
 	}
 }
 
-// Helper function to check if the error is one of the known HTTP Error types.
+//nolint:govet // We are using this function to check if the error is of type HTTP error.
 func isHTTPError(err error) bool {
-	var (
-		entityNotFoundError     gofrHTTP.ErrorEntityNotFound
-		entityAlreadyExistError gofrHTTP.ErrorEntityAlreadyExist
-		invalidParamError       gofrHTTP.ErrorInvalidParam
-		missingParamError       gofrHTTP.ErrorMissingParam
-		invalidRouteError       gofrHTTP.ErrorInvalidRoute
-		requestTimeoutError     gofrHTTP.ErrorRequestTimeout
-	)
-
-	switch {
-	case errors.Is(err, entityNotFoundError),
-		errors.Is(err, entityAlreadyExistError),
-		errors.Is(err, invalidParamError),
-		errors.Is(err, missingParamError),
-		errors.Is(err, invalidRouteError),
-		errors.Is(err, requestTimeoutError):
-		return true
-	default:
-		return false
+	httpErrors := []error{
+		&gofrHTTP.ErrorEntityAlreadyExist{},
+		&gofrHTTP.ErrorEntityNotFound{},
+		&gofrHTTP.ErrorInvalidParam{},
+		&gofrHTTP.ErrorMissingParam{},
 	}
+	for _, knownErr := range httpErrors {
+		if errors.As(err, &knownErr) {
+			return true
+		}
+	}
+
+	return false
 }
 
 func handleWebSocketUpgrade(r *http.Request) {
--- a/pkg/gofr/handler.go
+++ b/pkg/gofr/handler.go
@@ -150,7 +150,6 @@ func (h handler) logError(traceID string, err error) {
 	if err != nil {
 		errorLog := &ErrorLogEntry{TraceID: traceID, Error: err.Error()}
 
-		// Check if the error is one of the known types
 		if isHTTPError(err) {
 			h.container.Logger.Info(errorLog)
 		} else {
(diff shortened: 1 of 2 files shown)


pkg/gofr/handler_test.go
```diff
@@ -36,6 +36,8 @@ func TestHandler_ServeHTTP(t *testing.T) {
 			`{}`},
 		{"method is get, data is mil, error is not nil", http.MethodGet, nil, errTest, http.StatusInternalServerError,
 			`{"error":{"message":"some error"}}`},
+		{"method is get, data is mil, error is http error", http.MethodGet, nil, gofrHTTP.ErrorEntityNotFound{}, http.StatusNotFound,
+			`{"error":{"message":"No entity found with : "}}`},
```
ccoVeille, line 40: Why the log look strange?

`found with : "`

pkg/gofr/handler.go
```diff
@@ -149,8 +149,30 @@ func panicRecoveryHandler(re any, log logging.Logger, panicked chan struct{}) {
 func (h handler) logError(traceID string, err error) {
 	if err != nil {
 		errorLog := &ErrorLogEntry{TraceID: traceID, Error: err.Error()}
-		h.container.Logger.Error(errorLog)
+
+		if isHTTPError(err) {
+			h.container.Logger.Info(errorLog)
+		} else {
+			h.container.Logger.Error(errorLog)
+		}
```
ccoVeille: What about this?

```suggestion
		logHelper := h.container.Logger.Error
		if isHTTPError(err) {
			logHelper = h.container.Logger.Info
		}
		logHelper(errorLog)
```

pkg/gofr/handler.go
```diff
@@ -149,8 +149,30 @@ func panicRecoveryHandler(re any, log logging.Logger, panicked chan struct{}) {
...
+		&gofrHTTP.ErrorMissingParam{},
+	}
+	for _, knownErr := range httpErrors {
+		if errors.As(err, &knownErr) {
+			return true
+		}
 	}
+
+	return false
 }
```
ccoVeille: If you have to do this, it means you could do better with the struct definition

```suggestion
func isHTTPError(err error) bool {
	return errors.Is(err, gofrHTTP.Error) {
}
```

Now let's move to gofr http package

Here is the current code for ErrorEntityNotFound

```go
// ErrorEntityNotFound represents an error for when an entity is not found in the system.
type ErrorEntityNotFound struct {
        Name  string
        Value string
}

func (e ErrorEntityNotFound) Error() string {
        // For ex: "No entity found with id: 2"
        return fmt.Sprintf("No entity found with %s: %s", e.Name, e.Value)
}

func (ErrorEntityNotFound) StatusCode() int {
        return http.StatusNotFound
}
```

Let's define a error, then add a method to ErrorEntityNotFound

```go
var ErrCommon = errors.New("HTTP error")

func (ErrorEntityNotFound) UnWrap() error {
        return ErrCommon
}
```

ccoVeille:
I'm writing a feedback. I need time

ccoVeille (changes requested):

pkg/gofr/handler.go
ccoVeille: You changed the implementation

Please consider what I suggested anyway

ccoVeille: You can even define errCommon as unexported

And add this to http package

```go
func IsHTTPError(err error) bool {
    return errors.Is(err, errCommon)
}
```

pkg/gofr/handler_test.go
Umang01-hash: sorry for the empty values, corrected that

pkg/gofr/handler.go
Umang01-hash: I have changed the implementation to this:

```go
func (h handler) logError(traceID string, err error) {
	if err != nil {
		errorLog := &ErrorLogEntry{TraceID: traceID, Error: err.Error()}

		if errors.As(err, &gofrHTTP.ErrorEntityAlreadyExist{}) {
			h.container.Logger.Info(errorLog)
		} else if errors.As(err, &gofrHTTP.ErrorEntityNotFound{}) {
			h.container.Logger.Info(errorLog)
		} else if errors.As(err, &gofrHTTP.ErrorInvalidParam{}) {
			h.container.Logger.Info(errorLog)
		} else if errors.As(err, &gofrHTTP.ErrorMissingParam{}) {
			h.container.Logger.Info(errorLog)
		} else {
			h.container.Logger.Error(errorLog)
		}
	}
}
```

pkg/gofr/handler.go
Umang01-hash: @ccoVeille we can't use `errors.Is` as it will match the exact value (then the populated fields in the error struct will be treated as different error)

So based on your comment can we do something like this:

```go
import (
    "errors"
    gofrHTTP "path/to/gofrHTTP"
)

// Log the error (if any) with traceID and errorMessage.
func (h handler) logError(traceID string, err error) {
    if err != nil {
        errorLog := &ErrorLogEntry{TraceID: traceID, Error: err.Error()}

        if isKnownError(err) {
            h.container.Logger.Info(errorLog)
        } else {
            h.container.Logger.Error(errorLog)
        }
    }
}

func isKnownError(err error) bool {
    switch {
    case isEntityAlreadyExistError(err):
        return true
    case isEntityNotFoundError(err):
        return true
    case isInvalidParamError(err):
        return true
    case isMissingParamError(err):
        return true
    default:
        return false
    }
}

func isEntityAlreadyExistError(err error) bool {
    var target *gofrHTTP.ErrorEntityAlreadyExist
    return errors.As(err, &target)
}

func isEntityNotFoundError(err error) bool {
    var target *gofrHTTP.ErrorEntityNotFound
    return errors.As(err, &target)
}

func isInvalidParamError(err error) bool {
    var target *gofrHTTP.ErrorInvalidParam
    return errors.As(err, &target)
}

func isMissingParamError(err error) bool {
    var target *gofrHTTP.ErrorMissingParam
    return errors.As(err, &target)
}

```

ccoVeille: I disagree, mainly because that's the purpose of adding Unwrap

I could work on a better code, but I don't have time yet

Umang01-hash: @ccoVeille will wait for your suggestion on a better code.

ccoVeille: I opened #1298

Umang01-hash:
Closing this PR as the work of this PR has been addressed in PR #1298.

PR #1298, opened 2024-12-12
Decrease log level for 4xx HTTP errors

ccoVeille:
## Pull Request Template

**Description:**

-  Closes issue #1279 
-  Replaces #1295 

**Checklist:**

### DIFF
--- a/pkg/gofr/http/errors.go
+++ b/pkg/gofr/http/errors.go
@@ -2,10 +2,42 @@
 package http
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
 )
+
+// sentinel errors used for HTTP errors
+// they are not exposed to the user.
+var (
+	// errCommon is the base error for all HTTP errors.
+	errCommon = errors.New("HTTP issue")
+	// errRegular is the base error for all regular errors.
+	// errRegular wraps [errCommon], this way anything that matches errRegular will also match errCommon.
+	errRegular = fmt.Errorf("%w: %s", errCommon, "regular error")
+	// errCritical is the base error for all critical errors.
+	// errCritical wraps [errCommon], this way anything that matches errCritical will also match errCommon.
+	errCritical = fmt.Errorf("%w: %s", errCommon, "critical error")
+)
+
+// IsHTTPError returns true if the error is an HTTP error.
+// This uses the [Unwrapper] interface functions defined on the struct.
+func IsHTTPError(err error) bool {
+	return errors.Is(err, errCommon)
+}
+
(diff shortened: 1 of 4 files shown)


coolwednesday:
@ccoVeille, a better name for the PR would be appreciated.

Umang01-hash:
@ccoVeille I tested your solution and it is working perfectly fine.
I have updated the PR name as well as description.

And also it would be nice if you can add test for the newly added code.

ccoVeille:
This PR target branch was your PR one @Umang01-hash 

- #1285

It explains why the PR description was so basic. The tests were the ones you added.

I should have mentioned and use draft.

I was planning you to merge them in your branch, then keep the discussion running in your PR.

I'll sort that, no worries.

pkg/gofr/http/errors.go
```diff
@@ -2,11 +2,18 @@
...
 	"fmt"
 	"net/http"
 	"strings"
 )
 
+var errCommon = errors.New("HTTP error")
+
+func IsHTTPError(err error) bool {
+	return errors.Is(err, errCommon)
+}
```
ccoVeille, line 15: I think this might be OK for now, but a better name and sentinel could be provided.

I  say that because I didn't add the errCommon to ErrPanicRecovery, yet.

I could add another sentinel then, but also a new methods.

I struggle to find a name for the "is non critical error" and the sentinel.

Also, I dislike the obesity it brings.

Another solution could be to export the sentinels. But it's a different direction and it might be more difficult to maintain, also the magic behind and advantage of the Unwrap could be less visible.

Another way could be to provide another way to catch this, and only rely on the interface of having a status code and having a logic of 400 < code < 499

Anyway, if I keep IsHTTPError I will have to include PanicRecovery maybe

Here is the status of a Work in progress in my mind, what do you think guys

Umang01-hash: @ccoVeille  I think for now we can keep `IsHTTPError` as it is simplifying the things and at the same time serving the logic we want to implement.
You can include a different error for panicRecovery as it will need to be logged on Error level only.

ccoVeille: I would use something like "IsRegularError()"

ccoVeille: I'm working on it again

ccoVeille: I implemented here what I initially planned

ccoVeille:
I implemented here what I initially planned.

But I felt like it was not a good way to do. Something that couldn't scale easily,

So I provided an alternative implementation that would allow more than the current one.

See
- #1304

For me, it supersedes the current PR, that now becomes pointless.

Issue #1248, opened 2024-11-27
Refactor unit tests with hardcoded ports


ccoVeille:
The tests are using hardcoded port, it's not a good pattern. Unit tests cannot be launched in parallel.

Any of these packages could be used

https://github.com/facebookarchive/freeport
https://github.com/projectdiscovery/freeport
https://github.com/phayes/freeport
https://github.com/Konstantin8105/FreePort
https://github.com/jamielxcarter/dort
https://github.com/prithviramesh/free-port-finder

Of course, as the feature is very basic, it could be added via a simple helper. By copy pasting something like this.

https://gist.github.com/sevkin/96bdae9274465b2d09191384f86ef39d

From a library with a copyleft license

vikash:
This is a great idea - whenever we are actually testing a call over a port.

PR #1304, opened 2024-12-13
Decrease log level for 4xx HTTP errors : alternative option

### LINKED ISSUE #1279: Avoid Logging 4XX Status Codes as Errors

ccoVeille:
## Pull Request Template

**Description:**

This PR is an alternative implementation of what was initiated with

- #1295
- #1298 

It goes further than #1298 as it allows defining the log level for any error, not only the HTTP errors

It also provides a way for user to provide a log level for their custom errors (See [docs/advanced-guide/gofr-errors/page.md](https://github.com/gofr-dev/gofr/blob/development/docs/advanced-guide/gofr-errors/page.md))

Fixes #1279
Replaces #1295
Replaces #1298

**Checklist:**

### DIFF
--- a/docs/advanced-guide/gofr-errors/page.md
+++ b/docs/advanced-guide/gofr-errors/page.md
@@ -44,6 +44,8 @@ dbErr2 := datasource.ErrorDB{Message : "database connection timed out!"}
 GoFr's error structs implements an interface with `Error() string` and `StatusCode() int` methods, users can override the 
 status code by implementing it for their custom error.
 
+You can optionally define a log level for your error with the `LogLevel() logging.Level` methods
+
 #### Usage:
 ```go
 type customError struct {
@@ -56,5 +58,9 @@ func (c customError) Error() string {
 
 func (c customError) StatusCode() int {
 	return http.StatusMethodNotAllowed
+}
+
+func (c customError) LogLevel() logging.Level {
+	return logging.WARN
 }
 ```
--- a/pkg/gofr/handler.go
+++ b/pkg/gofr/handler.go
@@ -158,28 +158,21 @@ func (h handler) logError(traceID string, err error) {
 	}
 }
 
-// Helper function to check if the error is one of the known HTTP Error types.
+//nolint:govet // We are using this function to check if the error is of type HTTP error.
 func isHTTPError(err error) bool {
-	var (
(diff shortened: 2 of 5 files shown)


ccoVeille:
Can someone tell me why the tests are failing ?

I think it's about port already in use, so it could be related to:
- #1248

Umang01-hash (commented):
I think you missed adding the `LogLevel()` method to the **errors** of the `http` package. Without that all the http errors will be printed at ERROR level.

aryanmehrotra (changes requested):
I am doubtful and not convinced.
Right now, all the errors are logged at error level, will it be really beneficial from the user perspective to write a method to set the log level for their errors.

Because errors should be logged at error levels or based on the status code.

I.E 4xx being logged at INFO

vipul-rawat:
In my opinion, since all 4xx logs are client-side errors, they are not actually errors related to the server and should be logged at the INFO level like 2xx logs. Just 5xx logs should be considered errors for the server as they are unexpected and server-side errors.

ccoVeille:
> I think you missed adding the `LogLevel()` method to the **errors** of the `http` package. Without that all the http errors will be printed at ERROR level.

Indeed, I might have forgotten to commit a local file 🤦‍♂️😅

I will take a look.

ccoVeille:
> In my opinion, since all 4xx logs are client-side errors, they are not actually errors related to the server and should be logged at the INFO level like 2xx logs. Just 5xx logs should be considered errors for the server as they are unexpected and server-side errors.

Exactly, having a distinct error level for http error is important.

For example 404 could be a Debug, a 403 could be a warning.

A custom error could have a need to set the log level, the same way they are allowed to have a custom status code.

But also please note the change allows providing a log level to any error in the stack

Umang01-hash (approved):

aryanmehrotra (approved):

PR #1318, opened 2024-12-18
En/redis db support

### LINKED ISSUE #1316: Issue while changing the redis DB
## Problem
We are unable to change the redis DB dynamically in this framework such that each and everytime, the redis instance is connected, by default it will point it to 0. That would cause a problem for other microservices to change the DB value in the redis.

## Suggested Solution
- We can have it as env based in the redis options to change the db value such that we can configure it in the env level for a service.

Umang01-hash:
## Pull Request Template

**Description:**

-   Resolves #1316 .
-   Introduces a new config `REDIS_DB` to change the DB value for redis server.

**Checklist:**

### DIFF
--- a/pkg/gofr/datasource/redis/redis.go
+++ b/pkg/gofr/datasource/redis/redis.go
@@ -23,6 +23,7 @@ type Config struct {
 	Username string
 	Password string
 	Port     int
+	DB       int
 	Options  *redis.Options
 }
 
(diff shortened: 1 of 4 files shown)


aryanmehrotra (approved):

pkg/gofr/datasource/redis/redis.go
```diff
@@ -102,6 +110,8 @@ func getRedisConfig(c config.Config) *Config {
 		options.Password = redisConfig.Password
 	}
 
+	options.DB = redisConfig.DB
```
coolwednesday, line 113: Can we log the db index to which the redis DB is connected ? in connection successful logs ?

coolwednesday (changes requested):

pkg/gofr/datasource/redis/redis.go
Umang01-hash: done

coolwednesday (approved):

aryanmehrotra (approved):

PR #1317, opened 2024-12-17
Fix/hardcoded ports

Umang01-hash:
## Pull Request Template

**Description:**

-   Resolves issue #1248 

**Checklist:**

-   [x] I have formatted my code using  `goimport`  and  `golangci-lint`.
-   [x] All new code is covered by unit tests.
-   [x] This PR does not decrease the overall code coverage.
-   [x] I have reviewed the code comments and documentation for clarity.

**Thank you for your contribution!**

PR #1319, opened 2024-12-19
Fix : Server shouldn't start if HTTP_PORT is blocked

### LINKED ISSUE #1243: Server shouldn't start if HTTP_PORT is blocked
If the HTTP_PORT is blocked the server should not start. Currently, it appears the server attempts to start even if the port is in use which can lead to issues.
This issue proposes that the server should perform a check before starting to ensure the HTTP_PORT is available. If the port is blocked, the server should fail to start and log an error message.
This enhancement will improve the reliability of the application and prevent unexpected behavior.

Screenshot 2024-11-27 at 11 40 53 AM

coolwednesday:
**Description:**

-   Closes #1243 

**Checklist:**


pkg/gofr/gofr.go
```diff
@@ -244,6 +254,17 @@ func (a *App) Shutdown(ctx context.Context) error {
 	return err
 }
 
+func (*App) isPortAvailable(port int) bool {
+	conn, err := net.DialTimeout("tcp", fmt.Sprintf(":%d", port), checkPortTimeout)
+	if err != nil {
```
Umang01-hash, line 259: What about this `fmt.Sprint(port)` ?

pkg/gofr/gofr.go
```diff
@@ -94,16 +100,7 @@ func New() *App {
...
-	if _, err = os.Stat("./static/" + gofrHTTP.DefaultSwaggerFileName); err == nil {
-		// Route to serve the OpenAPI JSON specification file.
-		app.add(http.MethodGet, "/.well-known/"+gofrHTTP.DefaultSwaggerFileName, OpenAPIHandler)
-		// Route to serve the Swagger UI, providing a user interface for the API documentation.
-		app.add(http.MethodGet, "/.well-known/swagger", SwaggerUIHandler)
-		// Catchall route: any request to /.well-known/{name} (e.g., /.well-known/other)
-		// will be handled by the SwaggerUIHandler, serving the Swagger UI.
-		app.add(http.MethodGet, "/.well-known/{name}", SwaggerUIHandler)
-	}
+	app.addOpenAPIDocumentation()
```
Umang01-hash, line 103: Can we add verify or check as prefix before the method name `addOpenAPIDocumentation` as the method is also first checking and then adding the documentation.

Umang01-hash (commented):
Can we also add some test covering the new statements of code?

pkg/gofr/grpc.go
```diff
@@ -66,6 +66,10 @@ var (
 
 // RegisterService adds a gRPC service to the GoFr application.
 func (a *App) RegisterService(desc *grpc.ServiceDesc, impl any) {
+	if a.grpcRegistered && !a.isPortAvailable(a.grpcServer.port) {
+		a.container.Logger.Fatalf("grpc port %d is blocked or unreachable", a.grpcServer.port)
```
ccoVeille, line 70:
```suggestion
		a.container.Logger.Fatalf("gRPC port %d is blocked or unreachable", a.grpcServer.port)
```

pkg/gofr/gofr.go
```diff
@@ -244,6 +254,17 @@ func (a *App) Shutdown(ctx context.Context) error {
...
+func (*App) isPortAvailable(port int) bool {
+	conn, err := net.DialTimeout("tcp", fmt.Sprintf(":%d", port), checkPortTimeout)
+	if err != nil {
+		return true
+	}
+
+	conn.Close()
+
+	return false
+}
```
ccoVeille, line 266: This method do not use App. It could be a simple function.

But it could also be moved to `internal/port.go` that was just added with #1317 by @Umang01-hash

pkg/gofr/gofr.go
coolwednesday: there is a colon ...I don't think just port works ...as that is not a valid format.

pkg/gofr/gofr.go
ccoVeille: Did you forget to push?

I don't see the changes, but this was resolved

pkg/gofr/gofr.go
```diff
@@ -244,6 +254,17 @@ func (a *App) Shutdown(ctx context.Context) error {
...
+func isPortAvailable(port int) bool {
+	conn, err := net.DialTimeout("tcp", fmt.Sprintf(":%d", port), checkPortTimeout)
+	if err != nil {
+		return true
+	}
+
+	conn.Close()
+
+	return false
+}
```
ccoVeille, line 266: Reopening as previous was resolved without changes

https://github.com/gofr-dev/gofr/pull/1319#discussion_r1891210360

Please move this and the relative test to internal/port.go and port_test.go

coolwednesday: I did make it a function. And I will not be moving it to internal....as it does not exist as of now ....testutils exist... that is more related to helper functions for test. So adding it there does not make sense.

ccoVeille: I might have misunderstood something then. I'm sorry

aryanmehrotra (dismissed):

Umang01-hash (approved):

examples/using-add-rest-handlers/main_test.go
```diff
@@ -12,6 +13,8 @@ import (
 
 func TestIntegration_AddRESTHandlers(t *testing.T) {
 	const host = "http://localhost:9090"
+	t.Setenv("METRICS_PORT", fmt.Sprint(2023))
```
ccoVeille, line 16:
```suggestion
	t.Setenv("METRICS_PORT", "2023")
```

examples/using-cron-jobs/main_test.go
```diff
@@ -1,13 +1,16 @@
...
 import (
+	"fmt"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func Test_UserPurgeCron(t *testing.T) {
+	t.Setenv("METRICS_PORT", fmt.Sprint(2022))
```
ccoVeille, line 12:
```suggestion
	t.Setenv("METRICS_PORT", "2022")
```

examples/using-publisher/main_test.go
```diff
@@ -12,6 +13,8 @@ import (
 
 func TestExamplePublisher(t *testing.T) {
 	const host = "http://localhost:8100"
+	t.Setenv("METRICS_PORT", fmt.Sprint(2032))
```
ccoVeille, line 16:
```suggestion
	t.Setenv("METRICS_PORT", "2032")
```

examples/using-publisher/main_test.go
```diff
@@ -64,6 +67,7 @@ func TestExamplePublisher(t *testing.T) {
 func TestExamplePublisherError(t *testing.T) {
 	t.Setenv("PUBSUB_BROKER", "localhost:1012")
 	t.Setenv("HTTP_PORT", "8200")
+	t.Setenv("METRICS_PORT", fmt.Sprint(2038))
```
ccoVeille, line 70:
```suggestion
	t.Setenv("METRICS_PORT", "2038")
```

examples/using-web-socket/main_test.go
```diff
@@ -11,6 +11,7 @@ import (
 
 func Test_WebSocket_Success(t *testing.T) {
 	wsURL := fmt.Sprintf("ws://%s/ws", "localhost:8001")
+	t.Setenv("METRICS_PORT", fmt.Sprint(2030))
```
ccoVeille, line 14:
```suggestion
	t.Setenv("METRICS_PORT", "2030")
```

ccoVeille:
Usually, framework/app don't request a port to be available unless if a port is defined in the conf.

Usually the code does things like that
- a port is defined in the conf, the port is not available, app doesn't start or log an error if it's a metric port
- no port is provided in the conf: either the app takes the first available port and report it in the logs, or the feature behind the port is ignored, such as no port for metrics,
no metrics+ warning in the log

While the PR scope is clear,bI'm unsure what the app should do normally

@aryanmehrotra @Umang01-hash 

What do you think?

Umang01-hash:
Thank you for your feedback @ccoVeille .

Currently, if any of the **metrics/HTTP/gRPC** ports are blocked, **the server will not start.** GoFr emphasizes observability from the beginning of application development to ensure that all necessary services are available and running as expected.

We understand your concern about the complexity of writing tests when every single file requires a new port. We will consider adding an option where, if no port is provided in the configuration, the observability setup could be turned off. This would simplify the process and make it easier to write tests without needing to handle multiple ports.

For now, ensuring the specified ports are available before starting the server helps maintain a robust and observable application environment.

Thank you again for your valuable input.

ccoVeille:
I agree. I raised the point, I'm not asking for one solution or another.

With your feedbacks, I think you could simply consider the following:
- if no port is provided in the conf (via env or anything), the code uses :0 then a free port is computed

It could help for tests.

But you could also use a TestMain and set to freeport, and then no need to provide some in each tests

Of course, it's out of the current PR

I can help on this

aryanmehrotra (changes requested):
Let's use free port

examples/http-server-using-redis/main_test.go
```diff
@@ -78,6 +89,9 @@ func TestRedisSetHandler(t *testing.T) {
 }
 
 func TestRedisPipelineHandler(t *testing.T) {
+	t.Setenv("HTTP_PORT", "8086")
+	t.Setenv("METRICS_PORT", "2043")
```
ccoVeille, line 93: This one differs from the others

examples/http-server-using-redis/main_test.go
```diff
@@ -55,6 +63,12 @@ func TestHTTPServerUsingRedis(t *testing.T) {
 }
 
 func TestRedisSetHandler(t *testing.T) {
+	metricsPort := testutil.GetFreePort(t)
+	httpPort := testutil.GetFreePort(t)
+
+	t.Setenv("METRICS_PORT", fmt.Sprint(metricsPort))
+	t.Setenv("HTTP_PORT", fmt.Sprint(httpPort))
+
```
ccoVeille, line 71: Why alternating between strconv.Itoa and fmt.Sprint? (The previous files vs this one and next ones)

examples/using-add-rest-handlers/main_test.go
```diff
@@ -2,7 +2,10 @@ package main
 
 import (
 	"bytes"
+	"fmt"
+	"gofr.dev/pkg/gofr/testutil"
```
ccoVeille, line 6: Fix import order

examples/using-cron-jobs/main_test.go
```diff
@@ -1,13 +1,18 @@
 package main
 
 import (
+	"fmt"
+	"gofr.dev/pkg/gofr/testutil"
```
ccoVeille, line 5: Fix import order,please

examples/using-custom-metrics/main_test.go
```diff
@@ -1,16 +1,25 @@
 package main
 
 import (
+	"fmt"
+	"gofr.dev/pkg/gofr/testutil"
```
ccoVeille, line 5: Here too

examples/using-file-bind/main_test.go
```diff
@@ -2,10 +2,13 @@ package main
 
 import (
 	"bytes"
+	"fmt"
+	"gofr.dev/pkg/gofr/testutil"
```
ccoVeille, line 6: You got the point 😅

examples/http-server/main_test.go
```diff
@@ -20,9 +22,14 @@ import (
 	"gofr.dev/pkg/gofr/testutil"
 )
 
-const host = "http://localhost:9000"
+var host string
 
 func TestIntegration_SimpleAPIServer(t *testing.T) {
+	host = fmt.Sprint("http://localhost:9000")
```
ccoVeille, line 28: I'm unsure I get the point

```suggestion
	host = "http://localhost:9000"
```

examples/using-custom-metrics/main_test.go
```diff
@@ -31,11 +40,11 @@ func TestIntegration(t *testing.T) {
 		t.Fatalf("request to /transaction failed %v", err)
 	}
 
-	req, _ = http.NewRequest(http.MethodGet, "http://localhost:2120/metrics", nil)
+	req, _ = http.NewRequest(http.MethodGet, fmt.Sprint("http://localhost:", metricsPort, "/metrics"), nil)
 
 	resp, err := c.Do(req)
 	if err != nil {
-		t.Fatalf("request to localhost:2120/metrics failed %v", err)
+		t.Fatalf("request to fmt.Sprint(\"http://localhost:\", metricsPort,\"/metrics\") failed %v", err)
```
ccoVeille, line 47: I don't think it will do what you expect

examples/using-web-socket/main_test.go
```diff
@@ -10,7 +12,12 @@ import (
 )
 
 func Test_WebSocket_Success(t *testing.T) {
-	wsURL := fmt.Sprintf("ws://%s/ws", "localhost:8001")
+	port := testutil.GetFreePort(t)
+	t.Setenv("HTTP_PORT", strconv.Itoa(port))
+	wsURL := fmt.Sprintf("ws://%s/ws", fmt.Sprint("localhost:", port))
```
ccoVeille, line 17: fmt using fmt as argument



```suggestion
	wsURL := fmt.Sprintf("ws://localhost:%d/ws", port)
```

ccoVeille (approved):

examples/http-server/main_test.go
```diff
@@ -20,9 +21,14 @@ import (
 	"gofr.dev/pkg/gofr/testutil"
 )
 
-const host = "http://localhost:9000"
+var host string
```
ccoVeille, line 24: Can we refactor to remove this variable?

I have no idea how difficult it could be. I'm only spotting it by looking at the partial diff

examples/using-file-bind/main_test.go
```diff
@@ -2,19 +2,28 @@ package main
...
 )
 
 func TestMain_BindError(t *testing.T) {
-	const host = "http://localhost:8300"
+	httpPort := testutil.GetFreePort(t)
+	t.Setenv("HTTP_PORT", strconv.Itoa(httpPort))
+	host := fmt.Sprint("http://localhost:", httpPort)
+
+	port := testutil.GetFreePort(t)
+	t.Setenv("METRICS_PORT", strconv.Itoa(port))
```
ccoVeille, line 25: Not for this PR, but what do you guys think about adding an helper for setting up all this?

It could set up the ports, the env and return a struct, the port and url and host and metrics could be getter on this struct.

I can open an issue for this and suggest a PR, or it could be addressed by anyone if someone is interested.

ccoVeille: I can also suggest something using TestMain that could help you to make tests even cleaner. But the things will be hidden from sight and somehow magical. So I don't think TestMain should be used for things in examples folder. It's important that code can be easily copy pasted by people who doesn't understand how stuffs work.

But test main could be an hammer for a scratch, maybe a simple NewT(*testing.T) that set up the ports via env and return an app like New() does

pkg/gofr/external_db_test.go
```diff
@@ -1,8 +1,11 @@
 package gofr
 
 import (
+	"strconv"
 	"testing"
 
+	"gofr.dev/pkg/gofr/testutil"
```
ccoVeille, line 7: What is your IDE?

I would like to help you setting it up.

I'm not sarcastic, I do want to help

ccoVeille (approved):

examples/using-file-bind/main_test.go
coolwednesday: Sure, we can handle that in another PR.

pkg/gofr/external_db_test.go
coolwednesday: sorted the imports ...

examples/using-file-bind/main_test.go
ccoVeille: I created 
- https://github.com/gofr-dev/gofr/issues/1352

examples/http-server/main_test.go
```diff
@@ -20,9 +21,12 @@ import (
 	"gofr.dev/pkg/gofr/testutil"
 )
 
-const host = "http://localhost:9000"
-
 func TestIntegration_SimpleAPIServer(t *testing.T) {
+	host := "http://localhost:9000"
```
ccoVeille, line 25: Thanks for removing the package level const.

It's way clearer now.

One question, do we really need to have localhost:9000 in the 3 integration tests?

Can't it be moved to a random ports like the other tests?

coolwednesday: In the` main.go` file, the same server is added as an external `httpService` for demonstration purposes in one of the examples. To keep the setup straightforward, the port has been hardcoded since dynamically altering the `httpService` URL during tests is not feasible in this case.

aryanmehrotra (approved):

PR #1320, opened 2024-12-19
remove gofr's dependency from ftp package

### LINKED ISSUE #1259: Remove datasources' dependency on GoFr
Few datasources like ftp, sftp, nats import GoFr. This dependency must be removed to ensure that datasources are not dependent on any particular versions of GoFr.

Umang01-hash:
## Pull Request Template

**Description:**

-   Resolves #1259 partially.

**Checklist:**


pkg/gofr/datasource/file/ftp/interface.go
```diff
@@ -3,11 +3,82 @@ package ftp
...
 	"context"
 	"io"
+	"os"
 	"time"
 
 	"github.com/jlaffaye/ftp"
 )
 
+// FTPFile represents a file in the filesystem.
+type FTPFile interface {
```
ccoVeille, line 13: There is no need to name this FTPFile in ftp package


```suggestion
// File represents a file in the filesystem.
type File interface {
```

Especially when next interface is named FileInfo

pkg/gofr/datasource/file/ftp/interface.go
```diff
@@ -3,11 +3,82 @@ package ftp
...
+	IsDir() bool        // abbreviation for Mode().IsDir()
+}
+
+type RowReader interface {
+	Next() bool
+	Scan(interface{}) error
+}
+
+// FTPFileSystem : Any simulated or real filesystem should implement this interface.
+type FTPFileSystem interface {
```
ccoVeille, line 38:
```suggestion
// FileSystem : Any simulated or real filesystem should implement this interface.
type FileSystem interface {
```

pkg/gofr/datasource/file/ftp/file.go
```diff
@@ -14,8 +14,6 @@ import (
 	"time"
 
 	"github.com/jlaffaye/ftp"
-
-	file_interface "gofr.dev/pkg/gofr/datasource/file"
```
ccoVeille, line 18: Is this file still needed then?

pkg/gofr/datasource/file/ftp/interface.go
Umang01-hash: I completely agree with your point but we can't keep it just File as then it is colliding with the struct named `File` that has all the implementation of the interface.

pkg/gofr/datasource/file/ftp/interface.go
Umang01-hash: `FileSystem` is also a struct so we need to rename it to `FTPFileSystem`

pkg/gofr/datasource/file/ftp/file.go
Umang01-hash: no with the updated implementation that's not needed

Umang01-hash:
The failing `S3` tests and build for the workflow run can be addressed by this PR: https://github.com/gofr-dev/gofr/pull/1324

PR #1324, opened 2024-12-20
Remove/gofr dependency s3

### LINKED ISSUE #1259: Remove datasources' dependency on GoFr

Umang01-hash:
## Pull Request Template

**Description:**

-   Resolves #1259 partially.

**Checklist:**


aryanmehrotra (approved):

aryanmehrotra (dismissed):

pkg/gofr/datasource/file/s3/fs.go
```diff
@@ -71,7 +69,7 @@ func (f *FileSystem) UseMetrics(metrics interface{}) {
 // Connect initializes and validates the connection to the S3 service.
 //
 // This method sets up the S3 client using the provided configuration, including access key, secret key, region, and base endpoint.
-// It loads the AWS configuration and creates an S3 client, which is then assigns it to the `fileSystem` struct.
+// It loads the AWS configuration and creates an S3 client, which is then assigns it to the `FileSystem` struct.
```
ccoVeille, line 72:
```suggestion
// It loads the AWS configuration and creates an S3 client, which is then assigns it to the [FileSystem] struct.
```

pkg/gofr/datasource/file/s3/fs.go
```diff
@@ -50,7 +48,7 @@ type Config struct {
 }
 
 // New initializes a new instance of FTP fileSystem with provided configuration.
```
ccoVeille, line 50:
```suggestion
// New initializes a new instance of S3 fileSystem with provided configuration.
```

pkg/gofr/datasource/file/s3/fs.go
```diff
@@ -122,7 +120,7 @@ func (f *FileSystem) Connect() {
 // This method creates an empty file at the specified path in the S3 bucket. It first checks if the parent directory exists;
 // if the parent directory does not exist, it returns an error. After creating the file, it retrieves the file metadata
 // and returns a `file` object representing the newly created file.
```
ccoVeille, line 122: It no longer returns a file object.

You should talk about type assertion for getting result.

But for me it's strange. Returning an interface of with a few method could be an option, why removing the type and use any?

ccoVeille, line 122:
```suggestion
// and returns a [S3File] object representing the newly created file.
```

Umang01-hash: changed

pkg/gofr/datasource/file/s3/fs.go
Umang01-hash: done

pkg/gofr/datasource/file/s3/fs.go
Umang01-hash: sure corrected

pkg/gofr/datasource/file/s3/fs.go
Umang01-hash: We did it to complete remove the GoFr's dependency from the s3 datasource. Since the `file.File` interface was GoFr's interface defined in file package we made this changes to cut that dependency

ccoVeille (approved):

pkg/gofr/datasource/pubsub/mqtt/default_client.go
```diff
@@ -0,0 +1,81 @@
+package mqtt
+
+import (
+	"fmt"
+	"sync"
+
+	mqtt "github.com/eclipse/paho.mqtt.golang"
+
+	"github.com/google/uuid"
```
ccoVeille, line 9:
```suggestion
	mqtt "github.com/eclipse/paho.mqtt.golang"
	"github.com/google/uuid"
```

pkg/gofr/http/errors.go
```diff
@@ -94,3 +119,26 @@ func (ErrorPanicRecovery) Error() string {
...
+	_ statusCodeResponder = ErrorPanicRecovery{}
+
+	_ logging.LogLevelResponder = ErrorEntityNotFound{}
+	_ logging.LogLevelResponder = ErrorEntityAlreadyExist{}
+	_ logging.LogLevelResponder = ErrorInvalidParam{}
+	_ logging.LogLevelResponder = ErrorMissingParam{}
+	_ logging.LogLevelResponder = ErrorInvalidRoute{}
+	_ logging.LogLevelResponder = ErrorRequestTimeout{}
+	_ logging.LogLevelResponder = ErrorPanicRecovery{}
+)
```
ccoVeille, line 144: wait I don't get why you get so many changes in a PR about S3.

Is it a broken rebase? Or the bad target branch?

ccoVeille (changes requested):

pkg/gofr/datasource/file/s3/file_parse.go
```diff
@@ -41,7 +39,7 @@ type jsonReader struct {
 }
 
 // ReadAll reads either JSON or text files based on file extension and returns a corresponding RowReader.
-func (f *S3File) ReadAll() (file.RowReader, error) {
+func (f *S3File) ReadAll() (any, error) {
```
ccoVeille, line 42: I have a problem with the any here

Can't you define an interface that jsonReader and csvReader would satisfy?

I don't know at least something with Next, Scan and other if needed.

My main problem is a fact an exported method returns any in its signature, but also unexported structures.

I'm not even sure someone will be able to use the code once refactored

Shouldn't you document things?

Umang01-hash: @ccoVeille We can do it maybe but then also if the return type would be file.RowReader or something we have defined in GoFr then we can't remove GoFr's dependency from this datasource.

And even doing it with `any` is not working here.

ccoVeille: Your problems is about dependency import cycle?

Or about depending on external sources, so it leads to add things to gofr go.mod ?

If it's the first one, the solution is often to add a new package where you will move the interface definitions. This package will depends on nothing inside gofr.

Then for compatibility purpose, you replace the existing type at gofr level by aliases to the one at package level.

But here, once again, I might have totally misunderstood what you are trying to achieve

Umang01-hash: @ccoVeille 
> Or about depending on external sources, so it leads to add things to gofr go.mod ?

We aim to externalize all the datasources. The code in the datasource packages shouldn't depend on GoFr, so we don't have to worry about updating GoFr versions in these packages. This approach also maintains modularity and helps in keeping our binary size light.

However, if we keep the interface definition anywhere inside GoFr, we would need to import it inside the datasource packages for return types, which we want to avoid.

ccoVeille: OK,  but having data sources that returns any seems a problem to me.

But maybe other users of gofr could tell what they think about it, maybe a discussion could happen on the discord server

ccoVeille: Can't we move the file.RowHeader and other interface in data sources package?

Another solution is to duplicate the interface definitions which is acceptable.

Then unit tests can be added to validate  interface are compatible

PR #1342, opened 2024-12-26
En/system env overrides others

### LINKED ISSUE #1341: System Env not overriding other envs.
**Describe the bug**
If we set any env at `os` level, then it should override all the other env's like `.env` , `.local.env` etc.
It should basically override the env values set by `APP_ENV `config.

Umang01-hash:
## Pull Request Template

**Description:**

-   Resolve #1341.

**Checklist:**

### DIFF
--- a/pkg/gofr/config/godotenv.go
+++ b/pkg/gofr/config/godotenv.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
+	"strings"
 
 	"github.com/joho/godotenv"
 )
@@ -12,6 +13,7 @@ import (
 const (
 	defaultFileName         = "/.env"
 	defaultOverrideFileName = "/.local.env"
+	splitParts              = 2
 )
 
 type EnvLoader struct {
@@ -63,6 +65,14 @@ func (e *EnvLoader) read(folder string) {
 		}
 	} else {
 		e.logger.Infof("Loaded config from file: %v", overrideFile)
+	}
+
+	// Reload system environment variables to ensure they override any previously loaded values
+	for _, envVar := range os.Environ() {
+		keyValue := strings.SplitN(envVar, "=", splitParts)
+		if len(keyValue) == splitParts {
+			os.Setenv(keyValue[0], keyValue[1])
+		}
 	}
 }
 
--- a/pkg/gofr/config/godotenv.go
+++ b/pkg/gofr/config/godotenv.go
@@ -69,9 +69,9 @@ func (e *EnvLoader) read(folder string) {
 
 	// Reload system environment variables to ensure they override any previously loaded values
 	for _, envVar := range os.Environ() {
-		keyValue := strings.SplitN(envVar, "=", splitParts)
-		if len(keyValue) == splitParts {
-			os.Setenv(keyValue[0], keyValue[1])
+		key, value, found := strings.Cut(envVar, "=")
+		if found {
+			os.Setenv(key, value)
 		}
 	}
 }
--- a/pkg/gofr/config/godotenv.go
+++ b/pkg/gofr/config/godotenv.go
@@ -13,7 +13,6 @@ import (
 const (
 	defaultFileName         = "/.env"
 	defaultOverrideFileName = "/.local.env"
-	splitParts              = 2
 )
 
 type EnvLoader struct {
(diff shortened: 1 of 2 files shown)


pkg/gofr/config/godotenv.go
```diff
@@ -64,6 +66,14 @@ func (e *EnvLoader) read(folder string) {
 	} else {
 		e.logger.Infof("Loaded config from file: %v", overrideFile)
 	}
+
+	// Reload system environment variables to ensure they override any previously loaded values
+	for _, envVar := range os.Environ() {
+		keyValue := strings.SplitN(envVar, "=", splitParts)
+		if len(keyValue) == splitParts {
```
ccoVeille, line 73: There is a problem with the changes here, not the code but the logic, so maybe with the whole PR

Think about a password with `=` character in it

```dotenv
SECRET="foo="
```

ccoVeille, line 73: I don't think you will get a number bigger than 2 with SplitN 2

Is the idea about testing at least equal was present?

pkg/gofr/config/godotenv.go
```diff
@@ -64,6 +66,14 @@ func (e *EnvLoader) read(folder string) {
...
+			os.Setenv(keyValue[0], keyValue[1])
```
ccoVeille, line 74: Env parsing is somehow as fun as parsing JSON

You can have quote

```dotenv
FOO="bar" # this is fun
```

You can  have comments

```dotenv
FOO=bar # this is fun
```

You can have commented line

```dotenv
#FOO=bar
```

I would recommend using a well tested lib for this, godotenv is already used here. I'm unsure why you are using os.Environ and try to parse them "manually"

pkg/gofr/config/godotenv.go
```diff
@@ -64,6 +66,14 @@ func (e *EnvLoader) read(folder string) {
 	} else {
 		e.logger.Infof("Loaded config from file: %v", overrideFile)
 	}
+
+	// Reload system environment variables to ensure they override any previously loaded values
+	for _, envVar := range os.Environ() {
+		keyValue := strings.SplitN(envVar, "=", splitParts)
```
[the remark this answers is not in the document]
Umang01-hash: Sure @ccoVeille 
I have refactored the code and used ` strings.Cut` instead.

pkg/gofr/config/godotenv.go
Umang01-hash: I have tested this scenario with updated implementation and it's working fine.

Umang01-hash: Updated this code to this:
```go

	// Reload system environment variables to ensure they override any previously loaded values
	for _, envVar := range os.Environ() {
		key, value, found := strings.Cut(envVar, "=")
		if found {
			os.Setenv(key, value)
		}
	}
```

pkg/gofr/config/godotenv.go
Umang01-hash: @ccoVeille tried using the **godotenv** library to solve the issue, but since we already used the methods `Load()` and `Overload()` once in the code, it was not able to re-load the system environment variables. Therefore, as per the requirement of the issue, I used the above implementation manually to ensure that the system environment variables are correctly reloaded and override any previously loaded values.

ccoVeille: You should add a valuable comment like this one in the code.

It's not clear otherwise

pkg/gofr/config/godotenv.go
ccoVeille: Do you tji k you could add these tests to the PR then?

pkg/gofr/config/godotenv.go
```diff
@@ -5,13 +5,15 @@ import (
...
 	"os"
+	"strings"
 
 	"github.com/joho/godotenv"
 )
 
 const (
 	defaultFileName         = "/.env"
 	defaultOverrideFileName = "/.local.env"
+	splitParts              = 2
```
ccoVeille, line 16: Are you still using this?

Maybe unused linter could be added to golangci-lint

ccoVeille (approved):

pkg/gofr/config/godotenv.go
Umang01-hash: I have removed the const.

thankyou

vipul-rawat (approved):

PR #1355, opened 2024-12-31
Fix : Incorrect MongoDB Connection Logs

### LINKED ISSUE #1326: Incorrect Mongo Connection Logs
We don't use uri to connect , all the time anymore. We need to check if we are connecting through uri or making the connection string accordingly and then log accordingly.

coolwednesday:
- Closes #1326 
- Replacing uri in logs with Host for mongo

- This PR also updates the go.mod to fix security vulnerability issue, quoted as 
`1.22 in pkg/gofr/datasource/mongo/go.mod does not match this syntax and there is no additional toolchain directive, which may cause some go commands to fail.`

- Description : As uri ,also has password and username .... it should not be present in the logs....hence replaced it with host config.


Umang01-hash (dismissed):

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -108,18 +119,31 @@ func (c *Client) Connect() {
...
+	// regular expression to extract the host
+	re := regexp.MustCompile(`mongodb://.*?:.*?@(.*?):\d+/.*`)
+
+	matches := re.FindStringSubmatch(uri)
+
+	if len(matches) > 1 {
+		host = matches[1]
+	}
+
+	return host
```
ccoVeille, line 146: Regexp are evil, and also they are slow

Here you are parsing something that look a like an URI

I mean it's `scheme://user:pass@host:port/path`

Can't you use net/url package ? 

https://pkg.go.dev/net/url#ParseRequestURI

Then Hostname https://pkg.go.dev/net/url#URL.Hostname

Umang01-hash (dismissed):

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -83,13 +84,23 @@ func (c *Client) UseTracer(tracer any) {
 
 // Connect establishes a connection to MongoDB and registers metrics using the provided configuration when the client was Created.
 func (c *Client) Connect() {
-	c.logger.Debugf("connecting to MongoDB at %v to database %v", c.config.URI, c.config.Database)
+	var host string
+
+	c.logger.Debugf("connecting to MongoDB at %v to database %v", c.config.Host, c.config.Database)
```
aryanmehrotra, line 89: we are first logging it and then we are checking the host and later based on uri or host we are printing log, which might print and empty host

aryanmehrotra (changes requested):

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -108,18 +113,60 @@ func (c *Client) Connect() {
...
+	case config.Host == "":
+		return "", "", fmt.Errorf("%w: host is empty", errMissingField)
+	case config.Port == 0:
+		return "", "", fmt.Errorf("%w: port is empty", errMissingField)
+	case config.Database == "":
+		return "", "", fmt.Errorf("%w: database is empty", errMissingField)
+	}
+
+	uri = fmt.Sprintf("mongodb://%s:%s@%s:%d/%s?authSource=admin",
+		config.User, config.Password, config.Host, config.Port, config.Database)
```
ccoVeille, line 150: Be careful here

This approach could lead to bugs

user and password (but more likely the password can contain character that needs to be escaped, such as @#: it would lead to invalid URL, or password that won't be readable by mongodb client

They should be escaped I'm unsure what is the RFC for this.

Also the database may contains slash or ? (I'm usure it's possible with mongoDB), so here url.PathEscape should be used for  comfig.Database

I would like to suggest creating a url.URL with all required fields then use the .String

Please note, this could be done in a separate PR as the parttern of forging URL by concatenation might be used somewhere else

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -108,18 +113,71 @@ func (c *Client) Connect() {
...
+		return "", "", fmt.Errorf("%w: host is empty", errMissingField)
+	case config.Port == 0:
+		return "", "", fmt.Errorf("%w: port is empty", errMissingField)
+	case config.Database == "":
+		return "", "", fmt.Errorf("%w: database is empty", errMissingField)
+	}
+
+	u := &url.URL{
+		Scheme: "mongodb",
+		Host:   fmt.Sprintf("%s:%d", config.Host, config.Port),
```
ccoVeille, line 151: there is https://pkg.go.dev/net#JoinHostPort for this

it could lead to issue with IPv6 otherwise

ccoVeille (approved):

Umang01-hash (dismissed):

pkg/gofr/datasource/mongo/mongo_test.go
```diff
@@ -36,17 +36,155 @@ func Test_NewMongoClient(t *testing.T) {
...
+		config        Config
+		expectedURI   string
+		expectedHost  string
+		expectedError string
+	}{
+		{
+			name: "Valid Config",
+			config: Config{
+				User:     "admin",
+				Password: "password",
```
ccoVeille, line 51: Could you please update the password to something that cont wins characters that need to be encoded

Something like


```suggestion
				Password: "p@##word:",
```


This way you will validate they are escaped

ccoVeille, line 51: Could you add a test without authentication

I mean with user and password being empty

The code supports it

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -108,18 +115,71 @@ func (c *Client) Connect() {
...
+
+		return config.URI, host, nil
+	}
+
+	switch {
+	case config.Host == "":
+		return "", "", fmt.Errorf("%w: host is empty", errMissingField)
+	case config.Port == 0:
+		return "", "", fmt.Errorf("%w: port is empty", errMissingField)
+	case config.Database == "":
```
ccoVeille, line 147: Shouldn't you prevent (error) or warn (log)

URI and host, port, database, username and port cannot be used together.

I'm a bit afraid someone will use the URI, then will provide user and password.

Another way to consider this is to populate url.URL from the ParseURI instead of creating a URL object, then to set the field if present in the config

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -108,18 +115,71 @@ func (c *Client) Connect() {
...
+
+	q := u.Query()
+	q.Set("authSource", "admin")
+	u.RawQuery = q.Encode()
+
+	return u.String(), u.Hostname(), nil
+}
+
+func getDBHost(uri string) (host string, err error) {
+	parsedURL, err := url.Parse(uri)
```
ccoVeille, line 169: https://pkg.go.dev/net/url#ParseRequestURI would be safer

pkg/gofr/datasource/mongo/mongo.go
coolwednesday: That is the reason I am checking if URI is present or not ....
If it is ...host name is extracted for logs and that uri is only used for connection....

If it is not ..it checks for the other fields of the config to create the uri itself ... What is the issue here ? I am not able to understand ....

pkg/gofr/datasource/mongo/mongo_test.go
```diff
@@ -36,17 +36,166 @@ func Test_NewMongoClient(t *testing.T) {
...
+			name:        "Invalid URI with no host",
+			uri:         "mongodb://username:password@:27017/database?authSource=admin",
+			expected:    "",
+			expectedErr: "failed to parse host from MongoDB URI",
+		},
+		{
+			name:        "Empty URI",
+			uri:         "",
+			expected:    "",
+			expectedErr: "parse \"\": empty url",
```
ccoVeille, line 164: The change you add to do here because of my suggestion make me think this error should be wrapped

See https://github.com/gofr-dev/gofr/pull/1355#discussion_r1906560707

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -108,18 +115,71 @@ func (c *Client) Connect() {
...
+	q.Set("authSource", "admin")
+	u.RawQuery = q.Encode()
+
+	return u.String(), u.Hostname(), nil
+}
+
+func getDBHost(uri string) (host string, err error) {
+	parsedURL, err := url.ParseRequestURI(uri)
+	if err != nil {
+		return "", err
```
ccoVeille, line 171:
```suggestion
		return "", fmt.Errorf("%w: %w", errParseHost, err)
```
See https://github.com/gofr-dev/gofr/pull/1355#discussion_r1906559791

coolwednesday: I don't feel the same, 
If the url is empty....empty url error is already wrapped around `"error generating Mongo URI"`, errParseHost error seems irrelevant in that context.

pkg/gofr/datasource/mongo/mongo_test.go
coolwednesday: See [#1355 comment](https://github.com/gofr-dev/gofr/pull/1355#discussion_r1906567159)

pkg/gofr/datasource/mongo/mongo.go
ccoVeille: Indeed

ccoVeille (approved):

pkg/gofr/datasource/mongo/mongo.go
```diff
@@ -108,18 +115,71 @@ func (c *Client) Connect() {
...
+	}
+
+	if parsedURL.Scheme != "mongodb" {
+		return "", errIncorrectURI
+	}
+
+	if parsedURL.Hostname() == "" {
+		return "", errParseHost
+	}
```
ccoVeille, line 180: Could this happen now we use ParseRequestURI ?

Is errParseHost still needed?

coolwednesday: In one of the test cases, you see, it can be for this use case,

`mongodb://username:password@:27017/database?authSource=admin`

ccoVeille: Indeed. Thanks for checking

aryanmehrotra (approved):

Issue #1350, opened 2024-12-27
Bug: can't test path params 


coolwednesday:

vipul-rawat:
Users can write tests in the format, using the request from mux.SetURLVars to create a new gofr Request.
```
func TestHandler(t *testing.T) {
	req := httptest.NewRequest(http.MethodGet, "/hello", nil)
	req = mux.SetURLVars(req, map[string]string{"name": "New"})

	ctx := &gofr.Context{
		Request:   gofrHttp.NewRequest(req),
		Container: &container.Container{Logger: logging.NewLogger(logging.DEBUG)},
	}

	resp, err := HelloHandler(ctx)
	fmt.Println(resp, err)
}
```

PR #1359, opened 2025-01-03
Added function SetPathParam for setting path parameters in HTTP requests

vaidehiadhi:
## Pull Request Template

**Description:**

-  Added a function SetPathParam to streamline the process of setting path parameters in HTTP requests. This function uses mux.SetURLVars to associate a map of key-value pairs as path parameters with an HTTP request.

-   Fixes Bug: can't test path params (#1350), which prevented efficient testing of path parameters in HTTP handlers.

-   The inability to set path parameters in test scenarios created significant challenges for unit testing API routes, especially when path parameters were critical for routing logic. By introducing the SetPathParam function, you can now inject path parameters into mock HTTP requests, facilitating more robust and reliable testing workflows.

**Checklist:**

### DIFF
--- a/pkg/gofr/http/request.go
+++ b/pkg/gofr/http/request.go
@@ -179,3 +179,8 @@ func (r *Request) bindBinary(raw interface{}) error {
 
 	return nil
 }
+
+// SetPathParam sets URL path parameters for the given HTTP request.
+func SetPathParam(r *http.Request, prm map[string]string) *http.Request {
+	return mux.SetURLVars(r, prm)
+}
--- a/pkg/gofr/http/request_test.go
+++ b/pkg/gofr/http/request_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"github.com/gorilla/mux"
 	"io"
 	"mime/multipart"
 	"net/http"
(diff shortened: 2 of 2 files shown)


pkg/gofr/http/request_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"github.com/gorilla/mux"
```
ccoVeille, line 7: Fix import order please

Import out of std lib are separated from std lib ones with a line feed

pkg/gofr/http/request.go
```diff
@@ -179,3 +179,8 @@ func (r *Request) bindBinary(raw interface{}) error {
 
 	return nil
 }
+
+// SetPathParam sets URL path parameters for the given HTTP request.
+func SetPathParam(r *http.Request, prm map[string]string) *http.Request {
+	return mux.SetURLVars(r, prm)
+}
```
ccoVeille, line 186: I'm unsure about the function purpose, while mux one provides it

Maybe it's about making this easier to find what to do, but the comment could explain why it's a simple wrapper/alias

vipul-rawat:
This PR is not required, as the users can test the Handlers using the mux package, but I think it is important to document this as users may find it difficult where to set the path params(gofr using mux.Vars to get the path params)

PR #1394, opened 2025-01-16
Support Tracing for Interservice gRPC calls

coolwednesday:
**Description:**

-   Prerequisite to closing #1392 
-   This PR adds a check to verify if the incoming metadata for gRPC has traceID and spanID. If present then the span created is a child span of the existing span to ensure proper tracing in interservice gRPC calls.
-   The next step would be to develop the gofr-cli command that gives boilerplate code to add trace-id and span-id to the sent context's metadata.
-  Further changes have been initiated in gofr-cli : https://github.com/gofr-dev/gofr-cli/pull/26
[Screenshot 2025-01-18 at 10 23 15 AM](url)

**Checklist:**


pkg/gofr/grpc/log.go
```diff
@@ -56,40 +57,76 @@ func (l RPCLog) String() string {
...
+
+func initializeSpanContext(ctx context.Context) (context.Context, trace.SpanContext) {
+	md, _ := metadata.FromIncomingContext(ctx)
+
+	traceIDHex := getMetadataValue(md, "x-gofr-traceid")
+	spanIDHex := getMetadataValue(md, "x-gofr-spanid")
+
+	if traceIDHex != "" && spanIDHex != "" {
+		traceID, _ := trace.TraceIDFromHex(traceIDHex)
+		spanID, _ := trace.SpanIDFromHex(spanIDHex)
```
ccoVeille, line 86: Why ignoring those errors?

https://pkg.go.dev/go.opentelemetry.io/otel/trace#TraceIDFromHex

pkg/gofr/grpc/log.go
```diff
@@ -56,40 +57,76 @@ func (l RPCLog) String() string {
...
+		documentRPCLog(ctx, logger, info.FullMethod, start, err)
+
+		span.End()
 
 		return resp, err
 	}
 }
+
+func initializeSpanContext(ctx context.Context) (context.Context, trace.SpanContext) {
+	md, _ := metadata.FromIncomingContext(ctx)
```
ccoVeille, line 79: Why ignoring the fact the context might be not present

https://pkg.go.dev/google.golang.org/grpc/metadata#FromIncomingContext

You can get a nil context

https://github.com/grpc/grpc-go/blob/v1.69.4/metadata%2Fmetadata.go#L197-L203

And so code using it will panic

pkg/gofr/grpc/log.go
```diff
@@ -56,40 +57,76 @@ func (l RPCLog) String() string {
...
+			Remote:     true,
+		})
+
+		ctx = trace.ContextWithRemoteSpanContext(ctx, spanContext)
+
+		return ctx, spanContext
+	}
+
+	return ctx, trace.SpanContext{}
+}
```
ccoVeille, line 101:
```suggestion
	if traceIDHex == "" || spanIDHex == "" {
		return ctx, trace.SpanContext{}
	}

	traceID, _ := trace.TraceIDFromHex(traceIDHex)
	spanID, _ := trace.SpanIDFromHex(spanIDHex)

	spanContext := trace.NewSpanContext(trace.SpanContextConfig{
		TraceID:    traceID,
		SpanID:     spanID,
		TraceFlags: trace.FlagsSampled,
		Remote:     true,
	})

	ctx = trace.ContextWithRemoteSpanContext(ctx, spanContext)

	return ctx, spanContext
}
```

pkg/gofr/grpc/log.go
```diff
@@ -56,40 +57,76 @@ func (l RPCLog) String() string {
...
+
+	if err != nil {
+		if statusErr, ok := status.FromError(err); ok {
+			//nolint:gosec // gRPC codes are typically under the range.
+			logEntry.StatusCode = int32(statusErr.Code())
+		}
+	} else {
+		logEntry.StatusCode = int32(codes.OK)
+	}
+
```
ccoVeille, line 119: I think there is a problem here

If the error is not OK, you will get no status code, no?

```suggestion
	if err != nil {
		if statusErr, ok := status.FromError(err); ok {
			//nolint:gosec // gRPC codes are typically under the range.
			logEntry.StatusCode = int32(statusErr.Code())
		} else {
			// here?
		}
	} else {
		logEntry.StatusCode = int32(codes.OK)
	}

```

What about this

```suggestion
	logEntry.StatusCode = int32(codes.OK)
	if err != nil {
		if statusErr, ok := status.FromError(err); ok {
			logEntry.StatusCode = int32(codes.Internal)

			//nolint:gosec // gRPC codes are typically under the range.
			logEntry.StatusCode = int32(statusErr.Code())
		}
	}

```

But please note this

> Otherwise, err is an error not compatible with this package. In this case, a Status is returned with codes.Unknown and err's Error() message, and ok is false. 
> Source https://pkg.go.dev/google.golang.org/grpc/status#FromError

So maybe, you might simply get the status and ignore the `ok` returned value

pkg/gofr/grpc/log.go
coolwednesday: because traceID and spanID is something that is set by the framework itself while making interservice gRPC calls, and cannot really be anything else, here returning the error and checking for them would be just more code  which I feel is unnecessary.

pkg/gofr/grpc/log.go
coolwednesday: Even if we get a nil metadata , being a map, they are safe for read operations. And I am not performing any operations on metadata if I don't get traceID or spanID so the code does not panic. But, I will add a testcase for the same.

pkg/gofr/grpc/log.go
```diff
@@ -56,40 +57,75 @@ func (l RPCLog) String() string {
...
+		Method:       method,
+	}
+
+	if err != nil {
+		statusErr, _ := status.FromError(err)
+		//nolint:gosec // gRPC codes are typically under the range.
+		logEntry.StatusCode = int32(statusErr.Code())
+	} else {
+		logEntry.StatusCode = int32(codes.OK)
+	}
```
ccoVeille, line 117: What about this?


```suggestion
	statusCode := status.OK
	if err != nil {
		statusErr, _ := status.FromError(err)
		statusCode = statusErr.Code()
	}
	//nolint:gosec // gRPC codes are typically under the range.
	logEntry.StatusCode = int32(statusCode)
```

.github/workflows/go.yml
```diff
@@ -259,26 +259,43 @@ jobs:
...
-      - name: Lint Root Module
         run: |
-          echo "Linting root module..."
-          golangci-lint run --timeout 9m0s
+         go mod tidy
+
+      - name: Lint Root Module
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.62
```
ccoVeille, line 279: Why forcing a version ?

I mean it requires to maintain it

Also, if you want to force one. Please use the latest one, so the 1.63.4

.github/workflows/go.yml
```diff
@@ -259,26 +259,43 @@ jobs:
...
-            cd -
-          done    
+            golangci-lint run --timeout 9m0s || total_errors=$((total_errors + 1))
+            cd -  # Return to the root directory
+          done
+          echo "Total submodule lint errors: $total_errors"
+          if [ $total_errors -gt 0 ]; then
+            echo "Linting failed for $total_errors submodule(s)."
+            exit 1  # Fail the job if there are linting errors in submodules
+          fi
```
ccoVeille, line 298: Not directly related to your changes, but at some point, when the things to do becomes as complicated as this. Writing an sh script is better, it could be launched quickly

pkg/gofr/grpc/log.go
```diff
@@ -56,40 +57,77 @@ func (l RPCLog) String() string {
...
+		statusCode = statusErr.Code()
+	}
+
+	//nolint:gosec // gRPC codes are typically under the range.
+	logEntry.StatusCode = int32(statusCode)
+
+	if logger != nil {
+		logger.Info(logEntry)
+	}
+}
```
ccoVeille, line 124: Now, the code is way clearer.

Do we agree, we could do this?


```suggestion
func documentRPCLog(ctx context.Context, logger Logger, method string, start time.Time, err error) {
	if logger == nil {
		// nothing to do
		return
	}

	logEntry := RPCLog{
		ID:           trace.SpanFromContext(ctx).SpanContext().TraceID().String(),
		StartTime:    start.Format("2006-01-02T15:04:05.999999999-07:00"),
		ResponseTime: time.Since(start).Microseconds(),
		Method:       method,
	}

	statusCode := codes.OK
	if err != nil {
		statusErr, _ := status.FromError(err)
		statusCode = statusErr.Code()
	}

	//nolint:gosec // gRPC codes are typically under the range.
	logEntry.StatusCode = int32(statusCode)
	logger.Info(logEntry)
}
```

ccoVeille (approved):

.github/workflows/go.yml
coolwednesday: That can be done in a PR that addresses this very specific issue. This change in workflow was just created because using a container to launch golangc-lint run was leading in error , `Error obtaining VCS status`. We can update the golang version or any modifying in the workflow in a dedicated PR, specifically for it.

.github/workflows/go.yml
coolwednesday: Sure would refactor workflow, in another PR.

.github/workflows/go.yml
ccoVeille: I created:
- https://github.com/gofr-dev/gofr/issues/1398

.github/workflows/go.yml
ccoVeille: I created:
- https://github.com/gofr-dev/gofr/issues/1399

aryanmehrotra (commented):
1. Readme can be update either to refer the docs at gofr.dev/docs such that user can understand how these files are created.

aryanmehrotra:
Logs needs to be fixed
Screenshot 2025-01-19 at 1 56 46 AM


Metadata should also contain the correaltionID - can be part of different issue ticket
Screenshot 2025-01-19 at 1 59 31 AM

coolwednesday:
> Logs needs to be fixed - BADWIDTH Screenshot 2025-01-19 at 1 56 46 AM
> 
> Metadata should also contain the correaltionID - can be part of different issue ticket Screenshot 2025-01-19 at 1 59 31 AM
> 
> When an external request is made - we should also log it similar to interservice http calls - can be part of a different ticket Screenshot 2025-01-19 at 2 02 24 AM

In my implementation, I am only adding correlation ID that is the traceID, when the user actually initiates a call using gofr-CLI's generated gRPC Client. So only incoming context, while sending request carries metadata and none of the corresponding response/outgoing context include correlation ID in the metadata.

So, is it necessary in any case to have metadata with correlation ID in the outgoing context received ?

coolwednesday:
> Logs needs to be fixed - BADWIDTH Screenshot 2025-01-19 at 1 56 46 AM
> 
> Metadata should also contain the correaltionID - can be part of different issue ticket Screenshot 2025-01-19 at 1 59 31 AM
> 
> When an external request is made - we should also log it similar to interservice http calls - can be part of a different ticket Screenshot 2025-01-19 at 2 02 24 AM

Fixed :
Screenshot 2025-01-20 at 12 06 53 PM

ccoVeille:
I'm a bit surprised to see my recent commits  I made in this PR.

It looks like to me you had to fight conflicts at some points. You apparently sorted them. 👍




I'm used to rebase-only workflow. Maybe it's OK in a merge-workflow 🤔 🤷

But I prefer to report it

coolwednesday:
As there are multiple changes in files ... that are not related to the PR itself. I will be closing the PR and creating a new one.

Link to the new PR : https://github.com/gofr-dev/gofr/pull/1408

PR #1411, opened 2025-01-20
chore: update godoc and configure a new linter

ccoVeille:
I'm enabling a rule that is disabled by default by golangci-lint

the rule is useful for well-structured project like gofr

It would avoid some errors in godoc

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -68,6 +68,12 @@ linters-settings:
       - name: error-return
       - name: error-strings
       - name: errorf
+      - name: exported
+        arguments:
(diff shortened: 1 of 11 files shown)


ccoVeille:
Note for myself, there is something strange with the project, I will have to take a look later

But for now, it's easy to reproduce the issue locally by launching something like this

```bash
for f in $(find -name 'go.mod' | xargs dirname); do cd $f ; golangci-lint run ./... ; cd - > /dev/null ; done
```

Umang01-hash:
The work of this PR has been addressed and continued in PR #1436.

PR #1474, opened 2025-02-10
Storing the number of running GoFr servers

### LINKED ISSUE #1393: Display Active User Count on Website When GoFr Server is Running
**Description:** When a GoFr server is started, a ping should be sent to the GoFr website , say `/ping` endpoint, to update and display the current active user count in real-time. 

**Proposed Feature:**

- When a GoFr server starts, it sends a ping request to a designated API endpoint on the GoFr website, which will record and update the active user count.
- The website should display this active user count on the `/ping` status page.
- The active user count should be updated periodically (e.g., every 30 seconds) to reflect the number of currently running GoFr servers.

**Additional Notes:**
- Ensure that user privacy is respected, and no personal data is exposed in the process of counting active users.
- Consider rate-limiting the pings to prevent overloading the server.

coolwednesday:
## Pull Request Template

**Description:**
- Closes #1393 
- This [PR](https://github.com/gofr-dev/website/pull/153) must be closed after this PR.
- After merging this PR, we need to track the count of pings sent here to ensure real-time insights are reflected on our Grafana dashboard.

**Checklist:**


pkg/gofr/gofr.go
```diff
@@ -42,6 +42,11 @@ const (
 	gofrTraceExporter      = "gofr"
 	gofrTracerURL          = "https://tracer.gofr.dev"
 	checkPortTimeout       = 2 * time.Second
+	gofrHost               = "https://gofr.dev"
+	startServerPing        = "/api/ping/up"
+	shutServerPing         = "/api/ping/down"
+	pingTimeout            = 5 * time.Second
+	TelemetryTrue          = "true"
```
Umang01-hash, line 49: Can we rename it to `defaultTelemetry` instead?

pkg/gofr/gofr_test.go
```diff
@@ -71,6 +71,56 @@ func TestGoFr_isPortAvailable(t *testing.T) {
...
+	tests := []struct {
+		name           string
+		input          string
+		expectedURL    string
+		expectedMethod string
+		mockStatus     int
+	}{
+		{"Ping Start Server", "start", gofrHost + startServerPing, http.MethodGet, http.StatusOK},
+		{"Ping Shut Server", "stop", gofrHost + shutServerPing, http.MethodGet, http.StatusOK},
+	}
```
Umang01-hash, line 97: Since the values of `expectedMethod` and `mockStatus` is same in both the test cases we can remove these fields.

examples/grpc/grpc-server/main_test.go
```diff
@@ -19,7 +19,7 @@ func TestGRPCServer(t *testing.T) {
 	host := configs.GRPCHost
 
 	go main()
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(1000 * time.Millisecond)
```
Umang01-hash, line 22: Why we have did this? Can you please explain

Umang01-hash (changes requested):

pkg/gofr/gofr.go
```diff
@@ -224,6 +237,33 @@ func (a *App) Run() {
...
 }
 
+func (a *App) pingGoFr(client *http.Client, s string) {
+	url := fmt.Sprint(gofrHost, shutServerPing)
+
+	if s == "start" {
+		url = fmt.Sprint(gofrHost, startServerPing)
+
+		a.container.Info("GoFr tracks running servers via telemetry. Set GOFR_TELEMETRY=false in configs to disable it.")
+	}
```
ccoVeille, line 247: the same code is duplicated to check the variable in current implementation 

I suggest either 
- having a `func (a *App) hasTelemetry() bool`
- always calling pingGofr, and check for GOFR_TELEMETRY then return

pkg/gofr/gofr.go
```diff
@@ -224,6 +237,33 @@ func (a *App) Run() {
 	wg.Wait()
 }
 
+func (a *App) pingGoFr(client *http.Client, s string) {
```
ccoVeille, line 240: Here are some suggestions for a better name

```suggestion
func (a *App) handleTelemetry(client *http.Client, s string) {
```

Or


```suggestion
func (a *App) sendTelemetry(client *http.Client, s string) {
```

ccoVeille, line 240: I'm not fan of a string here, using a iota and constant might be better

pkg/gofr/gofr.go
```diff
@@ -224,6 +237,33 @@ func (a *App) Run() {
...
+	url := fmt.Sprint(gofrHost, shutServerPing)
+
+	if s == "start" {
+		url = fmt.Sprint(gofrHost, startServerPing)
+
+		a.container.Info("GoFr tracks running servers via telemetry. Set GOFR_TELEMETRY=false in configs to disable it.")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), pingTimeout)
```
ccoVeille, line 249:
```suggestion
func (a *App) pingGoFr(ctx context.Context, client *http.Client, s string) {
	url := fmt.Sprint(gofrHost, shutServerPing)

	if s == "start" {
		url = fmt.Sprint(gofrHost, startServerPing)

		a.container.Info("GoFr tracks running servers via telemetry. Set GOFR_TELEMETRY=false in configs to disable it.")
	}

	ctx, cancel := context.WithTimeout(context.WithoutCancel(), pingTimeout)
```

You should receive the context from the caller, that's the way to do

pkg/gofr/gofr.go
```diff
@@ -224,6 +237,33 @@ func (a *App) Run() {
...
+	if err != nil {
+		return
+	}
+
+	req.Header.Set("Connection", "close")
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return
+	}
```
ccoVeille, line 262: The error should be logged, either with a warning level or a debug

pkg/gofr/gofr.go
coolwednesday: It is an independent call, does not need to have access to the existing context.

Umang01-hash (dismissed):

pkg/gofr/gofr.go
```diff
@@ -224,6 +239,41 @@ func (a *App) Run() {
 	wg.Wait()
 }
 
+func (a *App) hasTelemetry() bool {
+	return a.Config.GetOrDefault("GOFR_TELEMETRY", "true") == defaultTelemetry
+}
```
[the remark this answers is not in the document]
coolwednesday: It looks like a bit over-engineering to me. If user does mention this config, I am assuming it will be from the intent of diasabling it.

ccoVeille, line 244: If it's the default value,it should be this no?


```suggestion
func (a *App) hasTelemetry() bool {
	return a.Config.GetOrDefault("GOFR_TELEMETRY", defaultTelemetry) == "true"
}
```

pkg/gofr/gofr.go
```diff
@@ -224,6 +237,33 @@ func (a *App) Run() {
...
+	}
+
+	req.Header.Set("Connection", "close")
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return
+	}
+
+	resp.Body.Close()
```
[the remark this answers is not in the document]
coolwednesday: https://github.com/gofr-dev/gofr/pull/1474#discussion_r1959112262

Umang01-hash (approved):

Umang01-hash (approved):

PR #1506, opened 2025-02-19
add validatePath method to sanitize fileNames

Umang01-hash:


coolwednesday (approved):

aryanmehrotra:
this will be breaking change as it is bringing a strict change on how path's are given by the user i.e we are not allowing user to give relative path and only relying on strict path.

PR #1510, opened 2025-02-20
Fix/mqtt default subscription

Umang01-hash:
## Pull Request Template

**Description:**

-   Add `IsConnected()` method to all pub-sub clients so that during subscription we can have a check of it before actually trying to subscribe.
- Add **retryConnect** to default MQTT client so that re-subscription happens after defaultClient reconnects. 

**Checklist:**


coolwednesday (approved):

vikash:
This looks unnecessary to be exposed. Subscribe can internally decide this. @Umang01-hash

Umang01-hash:
> This looks unnecessary to be exposed. Subscribe can internally decide this. @Umang01-hash 

Sure @vikash will do it internally.

coolwednesday (changes requested):
If we are internally checking whether the client is connected or not...the IsConnected() function should be unexported.

pkg/gofr/datasource/pubsub/google/google.go
```diff
@@ -333,3 +339,17 @@ func retryConnect(conf Config, logger pubsub.Logger, g *googleClient) {
 		time.Sleep(defaultRetryInterval)
 	}
 }
+
+func (g *googleClient) IsConnected() bool {
```
aryanmehrotra, line 343: this can be unexported is not being used outside

aryanmehrotra (changes requested):

pkg/gofr/datasource/pubsub/google/google.go
Umang01-hash: sure done

Umang01-hash:
> If we are internally checking whether the client is connected or not...the `IsConnected()` function should be unexported.

Sure done.

aryanmehrotra (approved):

PR #1518, opened 2025-02-24
719 refactoring gofr.go file

vikash:
Just broken down the gofr.go files into multiple files for now. Because refactoring needs to be done carefully but at the same time, we would want to merge changes as soon as possible. Every change in future might potentially make the issue larger. This should be fairly straight forward to merge for now. Eventually we need to also break down the relevant tests and figure out opportunities to create sub-packages wherever possible.


Umang01-hash:
> @coolwednesday @Umang01-hash - can we please do this one first as we wouldn't want gofr.go files to keep on increasing. 

Sure we will prioritise it and will fix it.

vikash:
Screenshot 2025-02-24 at 11 04 17 PM

Linter is not working properly on my system. While I debug this, feel free to take this PR forward by editing the files.

vikash:
Finally passed all checks. Merging.

PR #1529, opened 2025-02-26
En/graceful shutdown config

### LINKED ISSUE #1508: Complete inflight requests after SIGTERM
Unless I'm mistaken, when a SIGTERM is recieved gofr immediately exits closing all connections.

Would it be possible to gracefully handle the shutdown?

Off the top of my head the following might be useful, especially for those of us who plan to scale up/down:
- Ability to set a threshold (seconds) where gofr will still accept new connections after a sigterm
- Ability to set a threshold (seconds) where gofr will continue to handle inflight requests after the threshold for new connections (above)

Traefik's lifecycle procedure might be a good example of how it could be handled: https://doc.traefik.io/traefik/routing/entrypoints/#lifecycle

Umang01-hash:
## Pull Request Template

**Description:**

-   Closes #1508.
- Introduces a new config `SHUTDOWN_GRACE_PERIOD` to change the default graceful shutdown timeout from **30sec.**

**Checklist:**


pkg/gofr/run.go
```diff
@@ -19,19 +19,31 @@ func (a *App) Run() {
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
 
+	timeout, err := getShutdownTimeoutFromConfig(a.Config)
+	if err != nil {
+		a.Logger().Errorf("error parsing value of shutdown timeout from config: %v. Setting default timeout of 30 sec.", err)
+
+		timeout = shutDownTimeout
```
coolwednesday, line 26: As we are setting the timeout in function while sending the error. Again setting it here, doesn't seem required.

pkg/gofr/run.go
```diff
@@ -19,19 +19,31 @@ func (a *App) Run() {
...
+	}
+
 	// Goroutine to handle shutdown when context is canceled
 	go func() {
 		<-ctx.Done()
 
 		// Create a shutdown context with a timeout
-		shutdownCtx, done := context.WithTimeout(context.WithoutCancel(ctx), shutDownTimeout)
+		shutdownCtx, done := context.WithTimeout(context.Background(), timeout)
```
aryanmehrotra, line 34: why did we change the context?

aryanmehrotra (changes requested):

pkg/gofr/run.go
Umang01-hash: In the original code, `context.WithoutCance`l was used to ensure the shutdown process had a fixed timeout (**shutDownTimeout**) to complete before termination.

In the updated code, the `getShutdownTimeoutFromConfig` function fetches the shutdown timeout from the configuration, allowing it to be configurable. 

If `context.WithCancel` were used instead, the shutdown context would be canceled when the parent context is canceled, potentially cutting short the shutdown process and not respecting the full timeout value.

pkg/gofr/run.go
Umang01-hash: done resolved.

pkg/gofr/run.go
Umang01-hash: corrected

aryanmehrotra (approved):

PR #1530, opened 2025-02-26
Rearranging Middlewares to allow Authentication before setting WebSocket Conn

### LINKED ISSUE #1526: Cannot Close websocket conneciton in case of unauthorised request
**Describe the bug**
I have validations set for my server to authorise users based on Bearer tokens, when using the websocket, The server is showing connection successfull and I get error logs for connection highjack because I am writing 401 errors to the response writter. 

Even If I ignore the un-authorized request and not move forward, the connection remains open.

**To Reproduce**
Add a middleware which returns and writes to the http.ResponseWriter and hijack can be reproduced.
Check for authorization and return without calling inner.ServeHTTP(w,r), the connection to websocket remains open.

1. The code is

   ```go
   
   ```

2. The error is

   ```
   
   ```

**Expected behavior**
I should be able to get the connection in middleware for the particular websocket request and close connection in case of unauthorised request

**Screenshots**

- OS: [e.g. Linux]
- gofr version [e.g. v1.5.0]
- go version [e.g. 1.21]

**More description**

coolwednesday:
- Closes #1526 
- As websocket connection does not inherently provide authentication support. It is favourable to authenticate before starting a websocket connection. This can only be done, if user's added authentication or other middlewares are executed before, upgrading to a websocket connection. This PR addresses the same.


Umang01-hash (changes requested):

Umang01-hash (approved):

pkg/gofr/run.go
```diff
@@ -19,6 +21,18 @@ func (a *App) Run() {
...
+	//	WebSocket connections do not inherently support authentication mechanisms.
+	//	It is recommended to authenticate users before upgrading to a WebSocket connection.
+	//	Hence, we are registering middlewares here, to ensure that authentication or other
+	//	middleware logic is executed during the initial HTTP handshake request, prior to upgrading
+	//	the connection to WebSocket.
+	if a.httpServer != nil {
+		a.httpServer.router.Use(middleware.WSHandlerUpgrade(a.container, a.httpServer.ws))
+		a.httpServer.router.Use(middleware.Logging(a.Logger()))
+		a.httpServer.router.Use(middleware.Metrics(a.Metrics()))
+	}
```
aryanmehrotra, line 34: When other default middleware's are being registered in httpServer, we should not be registering any default middleware here.

aryanmehrotra (changes requested):

pkg/gofr/run.go
coolwednesday: I have added a developer note ... explaining the necessity to register it here ....

aryanmehrotra (dismissed):

aryanmehrotra (dismissed):

pkg/gofr/context_test.go
```diff
@@ -78,42 +79,40 @@ func TestContext_AddTrace(t *testing.T) {
...
 	defer ws.Close()
 
 	_, message, err := ws.ReadMessage()
-	require.NoError(t, err)
+	require.NoError(t, err, "Failed to read WebSocket message")
 
-	expectedResponse := "Hello! GoFr"
-	assert.Equal(t, expectedResponse, string(message))
+	expected := "Hello! GoFr"
+	assert.Equal(t, expected, string(message))
```
Umang01-hash, line 115: I think instead of using the expected variable we can directly compare the strings here.

pkg/gofr/context_test.go
```diff
@@ -78,42 +79,40 @@ func TestContext_AddTrace(t *testing.T) {
...
 
 		// TODO: returning error here to close the connection to the websocket
 		// as the websocket close error is not caught because we are using no bind function here.
 		// this must not be necessary. We should put an actual check in handleWebSocketConnection method instead.
-		return nil, &websocket.CloseError{Code: websocket.CloseNormalClosure, Text: "Error closing"}
+		return "", &websocket.CloseError{Code: websocket.CloseNormalClosure, Text: "Closing"}
 	})
 
 	go app.Run()
+	time.Sleep(200 * time.Millisecond) // Wait for the server to boot
```
Umang01-hash, line 101: Can we use a lesser sleep time maybe `100 milliseconds`?

pkg/gofr/http_server_test.go
```diff
@@ -40,13 +57,13 @@ func TestRun_ServerStartsListening(t *testing.T) {
...
 	// Start the server
-	go server.Run(c)
+	go server.run(c, middleware.GetConfigs(getConfigs(t)))
 
 	// Wait for the server to start listening
 	time.Sleep(100 * time.Millisecond)
 
 	var netClient = &http.Client{
-		Timeout: 200 * time.Millisecond,
+		Timeout: 200 * time.Second,
```
Umang01-hash, line 66: Can we use a lesser time here also

Umang01-hash (approved):

aryanmehrotra (approved):

PR #1534, opened 2025-02-28
En/add exists method arango

### LINKED ISSUE #1507: ArangoDB Migration - Method to determine if a DB, Collection or Graph exists
Would it be possible to add additonal methods to the ArangoDB datasource to support determining if a DB, Collection of Graph exists?

If you're in a position where the resources already exist, the migration will fail. I ran into this today with the latest 1.34 release.

From what I understand, the Arango GO SDK does have unique methods for determining if something exists.

p.s. just incase this helps other users, the migrations are stored in the default system DB.

Umang01-hash:
## Pull Request Template

**Description:**

-   Resolves #1507 
-   Introduce a new method `Exists(ctx context.Context, name string, resourceType any) (bool, error)` to check if a database, collection or a graph exists.

```go
package migrations

import (
	"context"
	"gofr.dev/pkg/gofr/datasource/arangodb"

	"gofr.dev/pkg/gofr/migration"
)

func createCollection() migration.Migrate {
	return migration.Migrate{
		UP: func(d migration.Datasource) error {
			// .....
			
			ans, err := d.ArangoDB.Exists(context.Background(), "employee_db", arangodb.ResourceTypeCollection)
			if err != nil || ans == false {
				return err
			}

			return nil
		},
	}
}

```

**Checklist:**

### DIFF
--- a/pkg/gofr/container/datasources.go
+++ b/pkg/gofr/container/datasources.go
@@ -637,7 +637,10 @@ type ArangoDB interface {
 	// Parameters:
 	//   - ctx: Request context for tracing and cancellation.
 	//   - name: Name of the database, collection, or graph.
-	//   - resourceType: Type of the resource ("database", "collection", "graph").
+	//   - resourceType: Type of the resource (any). Allowed values are:
+	//     - arangodb.ResourceTypeDatabase
+	//     - arangodb.ResourceTypeCollection
+	//     - arangodb.ResourceTypeGraph
 	//
 	// Returns true if the resource exists, otherwise false.
 	Exists(ctx context.Context, name string, resourceType any) (bool, error)
--- a/pkg/gofr/container/datasources.go
+++ b/pkg/gofr/container/datasources.go
@@ -633,6 +633,15 @@ type ArangoDB interface {
 	// the result parameter is not a pointer to a slice of maps.
 	Query(ctx context.Context, dbName string, query string, bindVars map[string]any, result any) error
 
+	// Exists checks if a database, collection, or graph exists.
+	// Parameters:
+	//   - ctx: Request context for tracing and cancellation.
+	//   - name: Name of the database, collection, or graph.
+	//   - resourceType: Type of the resource ("database", "collection", "graph").
+	//
+	// Returns true if the resource exists, otherwise false.
+	Exists(ctx context.Context, name, resourceType string) (bool, error)
+
 	HealthChecker
 }
 
--- a/pkg/gofr/container/datasources.go
+++ b/pkg/gofr/container/datasources.go
@@ -640,7 +640,7 @@ type ArangoDB interface {
 	//   - resourceType: Type of the resource ("database", "collection", "graph").
 	//
 	// Returns true if the resource exists, otherwise false.
-	Exists(ctx context.Context, name, resourceType string) (bool, error)
+	Exists(ctx context.Context, name string, resourceType any) (bool, error)
 
 	HealthChecker
 }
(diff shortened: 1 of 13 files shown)


pkg/gofr/datasource/arangodb/arango.go
```diff
@@ -203,6 +204,31 @@ func (c *Client) Query(ctx context.Context, dbName, query string, bindVars map[s
...
 }
 
+// Exists checks if a database, collection, or graph exists.
+// Parameters:
+//   - ctx: Request context for tracing and cancellation.
+//   - name: Name of the database, collection, or graph.
+//   - resourceType: Type of the resource ("database", "collection", "graph").
+//
+// Returns true if the resource exists, otherwise false.
+func (c *Client) Exists(ctx context.Context, name, resourceType string) (bool, error) {
```
vikash, line 214: We should not be taking resourceType as a string. Because this kind of implementation results into people relying on documentation for possible values. This also means that we will accept an invalid value and check for the valid values in the method definition. In a strong typed language,  defining a custom type is a better approach. Consider following: 

```
// Define a custom type for ResourceType
type ResourceType string

// Define allowed values
const (
	Database ResourceType = "Database"
        Collection ResourceType = "Collection"
        Graph ResourceType = "Graph"
)

func (c *Client) Exists(ctx context.Context, name string, resourceType ResourceType) bool
```

Umang01-hash: Sure @vikash I have addressed the review comment by defining a custom `ResourceType` within the `arangodb` package. 

To avoid importing **arangodb** in our `datasources.go`, I kept the **resourceType parameter as any** in the `Exists` method and perform a type check inside the method. 

And also updated the documentation.

coolwednesday (approved):

pkg/gofr/container/datasources.go
```diff
@@ -633,6 +633,18 @@ type ArangoDB interface {
...
+	// Parameters:
+	//   - ctx: Request context for tracing and cancellation.
+	//   - name: Name of the database, collection, or graph.
+	//   - resourceType: Type of the resource (any). Allowed values are:
+	//     - arangodb.ResourceTypeDatabase
+	//     - arangodb.ResourceTypeCollection
+	//     - arangodb.ResourceTypeGraph
+	//
+	// Returns true if the resource exists, otherwise false.
+	Exists(ctx context.Context, name string, resourceType any) (bool, error)
```
vikash, line 646: why 'resourceType any' why not 'resourceType ResourceType'

pkg/gofr/datasource/arangodb/arango.go
vikash: That defeats the purpose of having a strong type. The type should be defined by the package which is requiring it. This can also be treated as a soft indication if the method is really required to be part of interface or not.

pkg/gofr/container/datasources.go
Umang01-hash: We are trying to keep GoFr and the related datasource packages decoupled.
If i use `resourceType  arangodb. ResourceType` i will need to import the arangodb package in GoFr and that will tightly bind it.

vikash: If interface is defines by gofr package, the resourceType should be defined by it as well.

Umang01-hash: Sure @vikash got it.
Also was thinking to avoid this we can refactor the related `CreateDB`, `CreateCollection` and `CreateGraph` methods to have a check if the db, collection or graph with the same name exists and then only proceed with the creation of it.

Umang01-hash:
The work of this. PR has been contuinued in PR #1548. Closing here to avoid duplicacy.

PR #1537, opened 2025-03-02
 aded an option to get all query parameter as a key value map

aviscode:
## Pull Request Template

**Description:**
be able to access  all query parameter as a key value map

**Checklist:**

-   [v] I have formatted my code using  `goimport`  and  `golangci-lint`.
-   [v] All new code is covered by unit tests.
-   [v] This PR does not decrease the overall code coverage.

### DIFF
--- a/pkg/gofr/http/request.go
+++ b/pkg/gofr/http/request.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


Umang01-hash:
Hey @aviscode!
Thanks for raising the PR.

In GoFr, you can get the Param using `c.Param(key)` or `ctx.Params(key)` that will return you a **string** or a **[]string** respectively. While accessing all params at once offers convenience, we aim to avoid unintended dependencies on arbitrary inputs.

Could you share specific use cases where this approach is essential? This helps us evaluate better.

Thanks again for your input!

aviscode:
OK, here's the use case I build a service that's accessing my database and all other services are talking to this service and they passing many filters in query Params and I don't know the filters. I just need to pass them to the execution against the database so that's why I wanna get all the filters from the Quarry Perams and just get the key value and pass it to the exec

Umang01-hash:
Hi @aviscode,

Thank you for sharing your use case! We appreciate the convenience this feature offers, but GoFr’s design prioritizes security and maintainability. Directly passing arbitrary query parameters risks vulnerabilities like SQL injection and creates unclear service contracts.

We believe a safer approach would involve validating or restricting parameters upfront.

danysz:
I need exactly the same feature.
Any other solution. You provided an explanation why you don't like it but there is proposal for a better solution. I will be really happy for one.

PR #1541, opened 2025-03-04
Updating WorkFlow with Retries for Failing Tests

coolwednesday:
We are adding retry logic for the failing tests in the workflow, so that we do not have to rerun the pipeline again and agian.


Umang01-hash (approved):

.github/workflows/go.yml
```diff
@@ -176,55 +221,90 @@ jobs:
...
-
-            # Return to the root directory
-            cd -
-          done
-
+            # Create a directory for coverage reports
+            mkdir -p coverage_reports
+            
+            # Get the list of submodules
+            SUBMODULES='${{ steps.detect_submodules.outputs.submodules }}'
```
Umang01-hash, line 246: Here we are getting the list of submodules and above :
`echo "Detected submodules: $SUBMODULES"`

We are printing it again. I think they can be merged into a single command or step

.github/workflows/go.yml
```diff
@@ -176,55 +221,90 @@ jobs:
...
 
+  # Job for uploading coverage to external services (CodeClimate)
   upload_coverage:
     name: Upload Coverage📊
     runs-on: ubuntu-latest
-    needs: [Example-Unit-Testing,PKG-Unit-Testing]
-    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/development'}}
+    # This job only needs example and pkg test results, not submodules
+    needs: [Example-Unit-Testing, PKG-Unit-Testing]
+#    # Only run this job on pushes to the development branch
```
Umang01-hash, line 283: Please uncomment it.

.github/workflows/go.yml
```diff
@@ -176,55 +221,90 @@ jobs:
...
+              go mod download
+              go mod tidy
+            
+              # Run tests with a focus on failed tests first
+              go test ./... -v -short -coverprofile=${module_name}.cov -coverpkg=./...
+            
+              # Copy coverage file to the coverage_reports directory
+              cp ${module_name}.cov ../../../coverage_reports/
+            
+              cd -
```
Umang01-hash, line 267: I think this `cd -` can be remoed since each `xargs` command runs in a subshell

coolwednesday: Hey, I tried running the workflow with and without it .... The workflow fails in cases when I do not change the directory back.

Umang01-hash (dismissed):

Umang01-hash (approved):

PR #1544, opened 2025-03-05
Resolve correct endpoints to serve Static Files

### LINKED ISSUE #1543: Static Server re-registers the files already being served on user specified Endpoint when the directory is Static
GoFr’s server serves files from a folder named "static" through two endpoints, currently:

- /static/ (default endpoint)
- A user-specified endpoint
This needs to be fixed. If the user specifies "static" as the directory, we should ensure that files are not re-registered under multiple endpoints, avoiding duplication.

coolwednesday:
## Pull Request Template

**Description**  

- Closes #1543  
- This update ensures the following:  

1. Files in the `static/` directory (at the root of any GoFr project) are served by default at `/static/`.  
2. **Favicon, Swagger, and OpenAPI.json** can only be served if they exist inside the `static` directory.  
3. **OpenAPI.json and Swagger** are only accessible through `/.well-known/swagger` or `/.well-known/openapi.json`.  
4. If a user registers `AddStaticFiles({custom-endpoint}, ./static)`, GoFr automatically serves files at the **custom endpoint** instead of `/static/`.  
5. Multiple static file directories can be registered at different endpoints.  

**Checklist:**


pkg/gofr/http/router.go
```diff
@@ -53,18 +53,16 @@ type staticFileConfig struct {
 	directoryName string
 }
 
-func (rou *Router) AddStaticFiles(endpoint, dirName string) {
+func (rou *Router) AddStaticFiles(dirName, endpoint string) {
```
vikash, line 56: This change looks unnecessary and makes it backward incompatible.

coolwednesday: Users access the AddStaticFiles method on the App, not the router, so i think it's not a backward compatibility issue. However, the change is unnecessary, so I will revert it.

Umang01-hash (approved):

PR #1554, opened 2025-03-10
Fixing Static File Server to serve NotFound file if present with proper Status Code

### LINKED ISSUE #1555: GoFr returns wrong Status Code in case of static server rendering a NotFound file for 404 response
This issue is linked to discussion as follows : https://github.com/orgs/gofr-dev/discussions/1536

Key points in the discussion that needs to be addressed :
- A redundant `WriteHeader` call occurs because the framework attempts to set the header again, even if it has already been written in a middleware.
- The static server returns a `200 OK` response when say a `404.html` file is present, when we try to render it for any missing file requests, instead of required behavior of  a `404 Not Found` status.

coolwednesday:
**Description:**

-   Closes #1555 

-  This PR rearranges the middleware that were previously changed in PR #1530, in order to allow authentication in web-sockets, as the merged PR, introduced a loophole as all standard middlewares, being registered after all the user-defined middlewares, created the case where, the de-facto standard of GoFr that is logs. traces, metrics will not be applied in case we set and return the response in user defined middlewares itself. In this PR, standard middleware is executed first, and WebSocket connection upgrades at the very last, still ensuring authentication without breaking GoFr's expected behavior.

- The change is necessary because GoFr's `statusResponseWriter`, set in the `Logging middleware` at the start, tracks `WriteHeader` calls. This ensures any subsequent middleware can check if `WriteHeader` was already invoked, preventing redundant writes.

- Additionally, static file handling now includes a check for missing files. If a requested file is not found, the server checks for `404.html`, sets a `404 Not Found` status, and serves it if available. If `404.html `is absent, a plain `404 Not Found` response is returned.

- Added integration tests for the same.

**Checklist:**


pkg/gofr/http/router.go
```diff
@@ -85,6 +86,30 @@ func (staticConfig staticFileConfig) staticHandler(fileServer http.Handler) http
 			return
 		}
 
+		// Check if the requested file exists
+		if _, err = os.Stat(absPath); os.IsNotExist(err) {
+			// Set response status before writing to avoid GoFr overriding it
+			w.WriteHeader(http.StatusNotFound)
+
+			// Serve 404.html for missing HTML pages
+			path, err := filepath.Abs(filepath.Join(staticConfig.directoryName, "404.html"))
```
akshat-zopsmart, line 95: @coolwednesday Would be best if we can avoid hard coding the 404 file name

pkg/gofr/http/router.go
```diff
@@ -85,6 +86,30 @@ func (staticConfig staticFileConfig) staticHandler(fileServer http.Handler) http
...
+			if err != nil {
+				_, _ = w.Write([]byte("404 not found"))
+				return
+			}
+
+			if _, err = os.Stat(path); !os.IsNotExist(err) {
```
akshat-zopsmart, line 101: We should also handle other errors like file not readable

pkg/gofr/http/router_test.go
akshat-zopsmart, line 1: @coolwednesday we should have a test scenario added for file permissions issue as well (file exists but read permission missing)

akshat-zopsmart (changes requested):

pkg/gofr/http/router.go
coolwednesday: Done

pkg/gofr/http/router.go
coolwednesday: It would require adding an optional parameter in the function to check that. I don't think it needs to be complicated that way for now ....However with a usecase please do open a discussion, and we will take up the issue from there.

akshat-zopsmart: Can't we have it in `staticConfig` itself?

pkg/gofr/http_server.go
```diff
@@ -30,10 +30,17 @@ var (
 	errInvalidKeyFile         = errors.New("invalid key file")
 )
 
-func newHTTPServer(port int) *httpServer {
+func newHTTPServer(c *container.Container, port int, middlewareConfigs map[string]string) *httpServer {
```
akshat-zopsmart, line 33: @coolwednesday what's the advantage of moving the middleware here?

pkg/gofr/http/router.go
coolwednesday: Sure.

pkg/gofr/http/router.go
```diff
@@ -73,18 +77,61 @@ func (staticConfig staticFileConfig) staticHandler(fileServer http.Handler) http
 
 		fileName := filePath[len(filePath)-1]
 
+		absPath, err := filepath.Abs(filepath.Join(staticConfig.directoryName, url))
```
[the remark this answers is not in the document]
coolwednesday: If I move it in another function, it would become difficult for a developer to understand the whole flow ... As I would be using something like :
` func (cfg staticFileConfig) staticHandler(fileServer http.Handler) http.Handler {
		status, serveFile, response := handleStaticRequest(cfg.directoryName, r.URL.Path, r)

		if serveFile != "" {
			http.ServeFile(w, r, serveFile)
			return
		}
		
		w.WriteHeader(status)
		
		if response != "" {
			_, _ = w.Write([]byte(response))
		} else {
			fileServer.ServeHTTP(w, r)
		}
	})
}
`
I feel there can be confusion regarding when `http.ServeFile` is used and where `fileServer.ServerHTTP`. However I am open to discussion.

pkg/gofr/http/middleware/logger.go
```diff
@@ -12,17 +15,34 @@ import (
...
+		return
+	}
+
 	w.status = status
+	w.wroteHeader = true
 	w.ResponseWriter.WriteHeader(status)
 }
 
+// Hijack implements the http.Hijacker interface.
+func (w *StatusResponseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
```
Umang01-hash, line 38: Can we also try and add test for this newly added Hijack() method?

pkg/gofr/gofr.go
```diff
@@ -100,7 +100,7 @@ func (a *App) httpServerSetup() {
 	}
 
 	for dirName, endpoint := range a.httpServer.staticFiles {
-		a.httpServer.router.AddStaticFiles(endpoint, dirName)
+		a.httpServer.router.AddStaticFiles(a.container.Logger, endpoint, dirName)
```
Umang01-hash, line 103: Why not to use `a.Logger()` here since method is directly on the app struct we can use it's logger method.?

pkg/gofr/http/router_test.go
```diff
@@ -76,70 +77,143 @@ func TestRouterWithMiddleware(t *testing.T) {
...
-	router.AddStaticFiles("/gofr", currentWorkingDir+"/testDir")
+	runStaticFileTests(t, tempDir, "/static", testCases)
+}
 
-	// Send a request to the test handler
-	req := httptest.NewRequest(http.MethodGet, "/gofr/indexTest.html", http.NoBody)
-	rec := httptest.NewRecorder()
-	router.ServeHTTP(rec, req)
+// Testing files being served at an endpoint named public.
+func TestStaticFileServing_Public(t *testing.T) {
```
Umang01-hash, line 141: i don't really feel that we need a TDD for these test methods. You are have only one case in both the test why not to write it normally?

```go
func TestStaticFileServing_Public(t *testing.T) {
	tempDir := t.TempDir()
	err := os.WriteFile(filepath.Join(tempDir, "test.txt"), []byte("Hello, Public!"), 0600)
	if err != nil {
		t.Fatalf("Failed to create test file: %v", err)
	}

	logger := logging.NewMockLogger(logging.DEBUG)
	router := NewRouter()
	router.AddStaticFiles(logger, "/public", tempDir)

	req := httptest.NewRequest(http.MethodGet, "/public/test.txt", http.NoBody)
	w := httptest.NewRecorder()

	router.ServeHTTP(w, req)

	assert.Equal(t, http.StatusOK, w.Code)
	assert.Equal(t, "Hello, Public!", strings.TrimSpace(w.Body.String()))
}

func TestStaticFileServing_Root(t *testing.T) {
	tempDir := t.TempDir()
	err := os.WriteFile(filepath.Join(tempDir, "test.txt"), []byte("Hello, Root!"), 0600)
	if err != nil {
		t.Fatalf("Failed to create test file: %v", err)
	}

	logger := logging.NewMockLogger(logging.DEBUG)
	router := NewRouter()
	router.AddStaticFiles(logger, "/", tempDir)

	req := httptest.NewRequest(http.MethodGet, "/test.txt", http.NoBody)
	w := httptest.NewRecorder()

	router.ServeHTTP(w, req)

	assert.Equal(t, http.StatusOK, w.Code)
	assert.Equal(t, "Hello, Root!", strings.TrimSpace(w.Body.String()))
}
```

Umang01-hash (changes requested):

Umang01-hash (commented):
@coolwednesday i think we can start the logs with small case. We need to maintain consistency for logs all around GoFr.

Umang01-hash (approved):

aryanmehrotra (approved):

PR #1571, opened 2025-03-12
Disable GOFR_TELEMETRY to avoid adding count of GoFr servers run in the pipeline.

### LINKED ISSUE #1572: GoFr Telemetry is capturing analytics data from tests executed as part of the pipeline runs.

coolwednesday:
- Closes #1572
- Previously, each time the pipeline was executed, there was a sudden spike in the number of running GoFr servers. This resulted in misleading data regarding GoFr’s actual usage. To address this, we’re now configuring the environment for all tests that spin up GoFr servers, ensuring they don’t impact usage analytics.

- Previous number of hits on website because of pipeline:
Screenshot 2025-03-17 at 3 58 52 PM

- After the PR Change :
Screenshot 2025-03-17 at 4 00 34 PM


ccoVeille (changes requested):
I would like you to discourage such a change.

You will always face something where it will be forgotten.

Please consider using [testing.Testing](https://pkg.go.dev/testing#Testing) in [hasTelemetry](https://github.com/gofr-dev/gofr/blob/32f31a92517978df1a6e5c0e42e1c1264b784936/pkg/gofr/telemetry.go#L9)

Please note it may has consequence on how hasTelemtry is tested but a small refactoring could solve this.

coolwednesday:
> I would like you to discourage such a change.
> 
> You will always face something where it will be forgotten.
> 
> Please consider using [testing.Testing](https://pkg.go.dev/testing#Testing) in [hasTelemetry](https://github.com/gofr-dev/gofr/blob/32f31a92517978df1a6e5c0e42e1c1264b784936/pkg/gofr/telemetry.go#L9)
> 
> Please note it may has consequence on how hasTelemtry is tested but a small refactoring could solve this.

Hey I am checking that it is ensured...in the existing tests ....

Umang01-hash (approved):

aryanmehrotra (changes requested):
To fix/improve tests we should not make any changes in non-test files.

Umang01-hash:
> To fix/improve tests we should not make any changes in non-test files.

@aryanmehrotra the only change that the PR introduces is checking if the method is being called inside a test, that would save us from adding cofigs for disabling it in many test files. I don't feel that's a bad trade-off.

Umang01-hash:
@aryanmehrotra but when thinking of including the testing library in non test files is not a good practice as per standard go practices. We should keep test-related code and dependencies excluded from the final production binary, keeping it lean and free from unintended flags.

coolwednesday:
> @aryanmehrotra but when thinking of including the testing library in non test files is not a good practice as per standard go practices. We should keep test-related code and dependencies excluded from the final production binary, keeping it lean and free from unintended flags.

Sure will do this ...

Umang01-hash:
@coolwednesday Can we please try avoiding adding the "test" library in non test files and think of other alternative.

ccoVeille:
I'm sorry if the idea I suggested wasn't accepted.

I'm also sorry @coolwednesday spent time on working on this.

The solution is used 

https://github.com/search?q=language%3Ago+%22testing.Testing%28%29%22+&type=code&p=2


Here is an example about people using testing.Testing especially for telemetry

https://github.com/YounixM/signoz/blob/6e20fbb174f08fd4f132d4d22d88f4a411c8d931/pkg/query-service/telemetry/telemetry.go#L145


Then about the "do not import testing in non test file" I'm mitigated.

The pattern is pretty common

https://github.com/search?q=language%3Ago+%22testing.T%22+-path%3A_test.go&type=code

But I concede, most of the time it's for adding testing helper that will be exported to be reused.

ccoVeille:
Anyway, the other solution I can think about without being a pain to maintain would be to set the GOFR_TELEMETRY  env variable to false via [TestMain](https://pkg.go.dev/testing#hdr-Main)

So adding this to each package could be enough

```go
func TestMain(m *testing.M) {
	os.SetEnv("GOFR_TELEMETRY", "false")
	m.Run()
}
```

ccoVeille:
> We should keep test-related code and dependencies excluded from the final production binary

This argument is valid.

It was debated when testing.Testing() was added to std lib

https://github.com/golang/go/issues/52600#issuecomment-1489306050

ccoVeille:
So here, I would like to suggest checking the consequence in term of binary size with 

https://github.com/Zxilly/go-size-analyzer

(Hi @Zxilly 👋)

By using both methods:
- the testing.Testing() one
- the one that will be used at last.

Please note, I'm fine with the idea of using something else than the testing.Testing I initially suggested. I'm simply curious.

By the way using @Zxilly (great) tool is always a chance to catch unexpected consequences of some imports.

So no matter how this PR will solve the issue it tries to solve, I'm fine. No worries.

I'm just a bit disappointed for @coolwednesday who is doing back and forth changes. So sorry again. But sometimes, the journey is more interesting than the destination, so you might learn and discover things during this.

Umang01-hash (approved):

Issue #1573, opened 2025-03-12
Error while concurrently writing to WebSocket using WriteMessageToSocket


deadland2002:
**Describe the bug**

When multiple goroutines attempt to write data to a WebSocket simultaneously using `WriteMessageToSocket(wsMessage)`, it throws an error.  

**To Reproduce**
- Use multiple goroutines to write data to a WebSocket.  
- Call `gofrCtx.WriteMessageToSocket(wsMessage)` inside each goroutine.  
- Observe the error.  

1. The code is

```go
func test(gofrCtx *gofr.Context){
     err = gofrCtx.WriteMessageToSocket(wsMessage)
}   
```

2. The error is

```
panic: concurrent write to websocket connection

goroutine 434 [running]: github.com/gorilla/websocket.(*messageWriter).flushFrame(0x14000c93d28, 0x1, {0x140006c8953?, 0x1033e6634?, 0x1039f79e0?})
```

**Expected behavior**
Concurrent writes to a Web socket should not cause errors. The function should handle multiple goroutines writing data safely without requiring external synchronization.

**Environments (please complete the following information):**
- OS : MacOS
- gofr version : v1.35.0
- go version :  go1.24.1 darwin/arm64

**More description**
i solved the problem temporarily by using
```go
func test(gofrCtx *gofr.Context){
     mu.Lock()
     err = gofrCtx.WriteMessageToSocket(wsMessage)
     mu.Unlock()
}   
```

kuldeep-bishnoi:
can work on this, assign it to me. @Umang01-hash

Umang01-hash:
sure @kuldeep-bishnoi done!

Ale-Cas:
Hi @deadland2002, I'm not sure the assumptions here are correct, in particular WriteMessageToSocket seems to just wrap gorilla/websocket WriteMessage, which the documentation mentions explicitly as a not thread-safe method, see https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency

So, since this method is just a wrapper, it inherits the concurrency requirements from gorilla's implementation, and I think this should not be labeled as a bug, but as a possible improvement over the current implementation. 

Adding a mutex as you suggested could be an improvement, but could also lead to unexpected behavior in some cases. 
With a mutex, you wouldn't be sure which goroutine took the lock and the ordering of how messages are sent to the connection would be undefined. 
[...]

PR #1601, opened 2025-03-24
Add thread safety to socket write methods

Ale-Cas:
## Pull Request Template

**Description:**

Resolves: https://github.com/gofr-dev/gofr/issues/1573
And also adds support for WriteJSON and concurrent calls to both WriteMessage and WriteJSON through a mutex on the websocket connection wrapper.

**Additional Information:**

Even though the race detector is not used for all tests, I verified that the tests I added are passing using the `--race` flag.
Screenshot 2025-03-24 at 13 13 53

**Checklist:**

I run `goimports -w .` but it looks like it also re-formatted examples and other unrelated files, should I revert these changes manually? 

I also can't run `golangci-lint` with the current package configuration, for this reason I opened an issue on the lint package https://github.com/golangci/golangci-lint/issues/5596


coolwednesday:
@Ale-Cas, I see the pipeline is failing at linters. Can you please fix them ?

pkg/gofr/websocket/websocket.go
```diff
@@ -76,6 +78,24 @@ func (w *Connection) Bind(v any) error {
...
+
+// WriteJSON writes the JSON encoding of data as a message.
+//
+// This method is thread-safe and be called concurrently with WriteMessage
+func (w *Connection) WriteJSON(data interface{}) error {
+	w.writeMutex.Lock()
+	defer w.writeMutex.Unlock()
+	return w.Conn.WriteJSON(data)
+}
+
```
coolwednesday, line 98: Can I know the reason of adding this method ? As if I send any struct in `WriteMessageToSocket `method , it anyways serializes it and I recieve a JSON response similar to the behaviour of `WriteJSONToSocket`

coolwednesday (changes requested):

Ale-Cas:
Hi @coolwednesday, yes I'm working on fixing the linting, even though unfortunately I still cannot run them locally because of https://github.com/gofr-dev/gofr/issues/1573#issuecomment-2751082916 

> Also can I know the reason of changes in grpc example files ?
Since I couldn't run the golangci-lint, I manually run `goimports -w .` and it changed the formatting on unrelated files, I'll try to revert those changes manually.

> I would also suggest not pushing the go.work.sum file if it is directly not related to any changes in the websocket code.
Sure will revert that too, it changed when I run `go mod tidy`, but I'm not entirely sure why it was not in sync

coolwednesday:
> Hi @coolwednesday, yes I'm working on fixing the linting, even though unfortunately I still cannot run them locally because of [#1573 (comment)](https://github.com/gofr-dev/gofr/issues/1573#issuecomment-2751082916)
> 
> > Also can I know the reason of changes in grpc example files ?
> 
> Since I couldn't run the golangci-lint, I manually run `goimports -w .` and it changed the formatting on unrelated files, I'll try to revert those changes manually.
> 
> > I would also suggest not pushing the go.work.sum file if it is directly not related to any changes in the websocket code.
> 
> Sure will revert that too, it changed when I run `go mod tidy`, but I'm not entirely sure why it was not in sync

I would suggest not doing `git add .`(I am a guilty too, lol) and commit.

Ale-Cas:
> I would suggest not doing `git add .`(I am a guilty too, lol) and commit. Also we need to check the syncing issue, but we will address it in another PR. So unfortunately you need to revert those changes here.

Indeed, and thanks for the initial review, I'm definitely guilty of adding everything, it's a bad old habit. 
I'll revert and answer to your other comments.

pkg/gofr/websocket/websocket.go
Ale-Cas: This method is using the underlying gorilla implementation that is optimized to directly write the message on the underlying mesage writer. 
This is slightly more efficient than the current implementation which first marshals the json into an array of bytes and then uses the generic WriteMessage. 

Also this way it's easier to simply switch from the gorilla connection to directly use the gofr Connection, since the methods have the same name and the gofr one just adds thread-safety.

This was my rationale when adding this feature, but if you think it's not needed or a good use case, I can revert this too.

Screenshot 2025-03-26 at 15 17 31

pkg/gofr/websocket/websocket_test.go
```diff
@@ -150,3 +151,117 @@ func dereference(v any) any {
...
+					}
+				}()
+			}
+		}
+
+		wg.Wait()
+	}))
+	defer server.Close()
+}
+
```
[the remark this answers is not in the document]
coolwednesday: Ohh okay, got it. As the methods' names were similar, I got confused. My bad !

coolwednesday:
@Ale-Cas , everything is fine in the PR. However , can you remove the changes in unrelated files like gRPC examples and go.work.sum .... and then we can get the approval of other maintainers to get it merged.

pkg/gofr/websocket/websocket.go
coolwednesday: Sure, we can keep this.

Ale-Cas:
Hi @coolwednesday thanks for the review! sorry but it's been a busy time at work and haven't had much spare time to contribute, if you're ok with it I'll actually move the core changes to a separate PR instead of manually reverting the formatting for the unrelated files.

Ale-Cas:
@coolwednesday as you suggested, I moved the core change to https://github.com/gofr-dev/gofr/pull/1642

PR #1614, opened 2025-03-27
Added support to handle multi-page apps from gofr

im-ng:
## Pull Request Template

**Description:**

- Added support to enable the gofr to be used in multi page application
- Added support to handle redirect headers and cookie headers for easy server redirection
- This helps me to write better app using the templ engine

- None

**Additional Information:**

- None


coolwednesday:
@im-ng, please address any linter or Code Climate issues in the PR.

It would also be helpful to include the motivation behind this PR. If possible, consider creating a GitHub issue explaining the problem you faced and linking it in the PR. If an issue already exists, you can reference it using **fixes #issue_number** or **closes #issue_number**.

This makes it easier for reviewers and maintainers to understand the context and prioritize the changes accordingly. Thanks! 😊

im-ng:
@coolwednesday Updated necessary comments, lint part in this PR.

vikash:
For gofr's focus on being the framework for Microservices, is cookie a valid use-case? Specially when it changes (adds fields and thus complexity) to the common structs. This does not feel a right addition to the framework.

im-ng:
Thanks for the feedback. This PR intends to enable session management for the **multi-page application**, as the framework does not have to limit itself to a stateless authorization header. 

I see the custom headers were introduced in the 1.28.x (common struct) version, as Gofr finds more use cases to adjust itself to support the x-refresh-token or more. As the framework grows in the hands of users, I thought it would be a good addition to allow one to set the needed JWT or secure tokens using the cookie itself for MPA.

As we observe increasing traction surrounding server-side rendering, templates, and HTMX, enabling such capabilities assists one in utilizing Gofr as their framework.

coolwednesday:
Hi @im-ng ,

We appreciate the effort you've put into this. We'll take some more time to evaluate the relevance of the proposed use case and consider the implementation details you've suggested.

Since this PR has been open for quite some time, we'll be closing it for now. Thank you for your understanding, and we look forward to reconnecting once we have more clarity on the direction.

PR #1615, opened 2025-03-27
En/error fields

Umang01-hash:
## Pull Request Template

**Description:**

-  Allows users to add more fields in their error response.

### Example

Screenshot 2025-03-27 at 10 14 21 AM

**Checklist:**


docs/advanced-guide/gofr-errors/page.md
```diff
@@ -64,3 +64,35 @@ func (c customError) LogLevel() logging.Level {
 	return logging.WARN
 }
 ```
+
+## Extended Error Responses
+
+For RFC 7807-style error responses with additional fields, implement the ResponseMarshaller interface:
+
```
coolwednesday, line 71: 7807 has been obsoleted by 9457 ....

```suggestion
## Extended Error Responses

For RFC 7807/9457 - style error responses with additional fields, implement the ResponseMarshaller interface:

```

coolwednesday (changes requested):

coolwednesday:
@Umang01-hash , when we send an empty struct, ideally we should get a `500` status code and some details(that are hardcoded and do not need to be initialised) in the error response . However in such case it returns a `200` status code ...a new field status as `500` and status as Error. Seems a bit weird. 

Screenshot 2025-03-27 at 3 16 19 PM

Also whenever the status code is not set, can we handle to return `500` ? As in current case it causes panic in `writeHeader`.

docs/advanced-guide/gofr-errors/page.md
Umang01-hash: sure done

pkg/gofr/http/responder.go
```diff
@@ -58,6 +64,54 @@ func (r Responder) Respond(data any, err error) {
 	_ = json.NewEncoder(r.w).Encode(resp)
 }
 
+func (r Responder) determineResponse(data any, err error) (statusCode int, errObj any) {
+	// Handle empty struct case first
+	if err == nil && isEmptyStruct(data) {
+		return http.StatusInternalServerError, createErrorResponse(errEmptyResponse)
+	}
+
```
coolwednesday, line 72: I feel it should be checking for not nil.

```suggestion
	// Handle empty struct case first
	if err != nil && isEmptyStruct(data) {
		return http.StatusInternalServerError, createErrorResponse(errEmptyResponse)
	}

```

coolwednesday (changes requested):

pkg/gofr/http/responder.go
```diff
@@ -58,6 +64,54 @@ func (r Responder) Respond(data any, err error) {
...
+	// Ensure valid status code
+	if statusCode == 0 {
+		statusCode = http.StatusInternalServerError
+
+		if errorObj == nil {
+			errorObj = createErrorResponse(errMissingStatusCode)
+		}
+	}
+
+	return statusCode, errorObj
```
ccoVeille, line 84: What about this?

```suggestion
	statusCode, errorObj := getStatusCode(r.method, data, err)

	// Ensure valid status code
	if statusCode == 0 {
		statusCode = http.StatusInternalServerError

		if errorObj == nil {
			errorObj = createErrorResponse(errMissingStatusCode)
		}
	}

	if errorObj == nil {
			errorObj = createErrorResponse(errMissingError)
		}

	return statusCode, errorObj
```

I'm unsure there is a need, but I prefer to raise the point

pkg/gofr/http/responder_test.go
```diff
@@ -207,6 +208,122 @@ func TestResponder_TemplateResponse(t *testing.T) {
...
+	assert.Equal(t, http.StatusNotFound, resp.StatusCode)
+
+	var body map[string]any
+	err := json.NewDecoder(resp.Body).Decode(&body)
+	require.NoError(t, err)
+
+	errorObj := body["error"].(map[string]any)
+
+	assert.Equal(t, "resource not found", errorObj["message"])
+	assert.Equal(t, int(http.StatusNotFound), int(errorObj["code"].(float64)))
```
ccoVeille, line 235: You can use EqualValues here


```suggestion
	assert.EqualValues(t, http.StatusNotFound, errorObj["code"])
```

pkg/gofr/http/responder_test.go
```diff
@@ -207,6 +208,122 @@ func TestResponder_TemplateResponse(t *testing.T) {
...
+	resp := w.Result()
+	defer resp.Body.Close()
+
+	assert.Equal(t, http.StatusNotFound, resp.StatusCode)
+
+	var body map[string]any
+	err := json.NewDecoder(resp.Body).Decode(&body)
+	require.NoError(t, err)
+
+	errorObj := body["error"].(map[string]any)
```
ccoVeille, line 232:
```suggestion
	errorObj, ok := body["error"].(map[string]any)
```

Then report an error if !ok

pkg/gofr/http/responder.go
Umang01-hash: sure done

pkg/gofr/http/responder_test.go
```diff
@@ -207,6 +208,122 @@ func TestResponder_TemplateResponse(t *testing.T) {
...
+	responder.Respond(nil, msgErr)
+
+	resp := w.Result()
+	defer resp.Body.Close()
+
+	var body map[string]any
+	err := json.NewDecoder(resp.Body).Decode(&body)
+	require.NoError(t, err)
+
+	errorObj := body["error"].(map[string]any)
```
ccoVeille, line 268: Same

pkg/gofr/http/responder.go
Umang01-hash: @ccoVeille there is no need of this change because:


- The current implementation follows REST API best practice where successful responses don't include error fields
- The json:"error,omitempty" tag is intentionally designed to omit the error field when nil.

pkg/gofr/http/responder_test.go
Umang01-hash: sure done

pkg/gofr/http/responder_test.go
Umang01-hash: sure done

pkg/gofr/http/responder_test.go
```diff
@@ -207,6 +208,129 @@ func TestResponder_TemplateResponse(t *testing.T) {
...
+	assert.Equal(t, http.StatusNotFound, resp.StatusCode)
+
+	var body map[string]any
+	err := json.NewDecoder(resp.Body).Decode(&body)
+	require.NoError(t, err)
+
+	errorObj, ok := body["error"].(map[string]any)
+	if !ok {
+		t.Fatalf("Expected error object to be of type map[string]any, got %T", body["error"])
+	}
```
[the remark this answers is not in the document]
Umang01-hash: Sure used `require.Truef`

pkg/gofr/http/responder_test.go
```diff
@@ -207,6 +208,129 @@ func TestResponder_TemplateResponse(t *testing.T) {
...
+
+	assert.Equal(t, "resource not found", errorObj["message"])
+	assert.EqualValues(t, http.StatusNotFound, int(errorObj["code"].(float64)))
```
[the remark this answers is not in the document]
Umang01-hash: No it is passing. I corrected that. Thankyou

ccoVeille (approved):

coolwednesday (approved):

coolwednesday (approved):

pkg/gofr/http/responder_test.go
```diff
@@ -207,6 +208,123 @@ func TestResponder_TemplateResponse(t *testing.T) {
...
+	var body map[string]any
+	err := json.NewDecoder(resp.Body).Decode(&body)
+	require.NoError(t, err)
+
+	errorObj, ok := body["error"].(map[string]any)
+	require.Truef(t, ok, "Expected error object to be of type map[string]any, got %T", body["error"])
+
+	assert.Equal(t, "resource not found", errorObj["message"])
+	assert.EqualValues(t, http.StatusNotFound, errorObj["code"])
+	assert.Equal(t, "Custom Error", errorObj["title"])
```
ccoVeille, line 237: I was annoyed by this code.

I was likely to suggest you to use unserialize the json to a struct to avoid all this by hand.

Something like

```go
var body struct {
Message string `json:message`
Code int `json:code`
Errors []string `json:errors`
}
```

But then I realized it could be way simpler to consider using assert.JSONEq directly with the expected returned json

ccoVeille, line 237: I was annoyed by this code.

I was likely to suggest you to use unserialize the json to a struct to avoid all this by hand.

Something like

```suggestion
var body struct {
Error CustomError `json:error`
}
err := json.NewDecoder(resp.Body).Decode(&body)
	require.NoError(t, err)

	errObj := body.Error
	assert.Equal(t, "resource not found", errObj.Message)
	assert.EqualValues(t, http.StatusNotFound, errorObj.Code)
	assert.Equal(t, "Custom Error", errorObj.Title)
```

This is pseudo code

But then I realized it could be way simpler to consider using assert.JSONEq directly with the expected returned json


```suggestion
	var body json.RawMessage
	err := json.NewDecoder(resp.Body).Decode(&body)
	require.NoError(t, err)

	assert.JSONEq(t,`{"error":{"code": 404, "title": "Custom Error", "message": "resource not found"}}`, body)
```

Pseudo code too

ccoVeille, line 237: I was annoyed by this code.

I was likely to suggest you to use unserialize the json to a struct to avoid all this by hand.

Something like

```suggestion
var body struct {
Error CustomError `json:error`
}
err := json.NewDecoder(resp.Body).Decode(&body)
	require.NoError(t, err)

	errObj := body.Error
	assert.Equal(t, "resource not found", errObj.Message)
	assert.EqualValues(t, http.StatusNotFound, errorObj.Code)
	assert.Equal(t, "Custom Error", errorObj.Title)
```

This is pseudo code

But then I realized it could be way simpler to consider using assert.JSONEq directly with the expected returned json


```suggestion
	var body json.RawMessage
	err := json.NewDecoder(resp.Body).Decode(&body)
	require.NoError(t, err)

	assert.JSONEq(t,`{"error":{"code": 404, "title": "Custom Error", "message": "resource not found"}}`, body)
```

Pseudo code too

pkg/gofr/http/responder.go
```diff
@@ -58,6 +64,54 @@ func (r Responder) Respond(data any, err error) {
...
+
+	statusCode, errorObj := getStatusCode(r.method, data, err)
+
+	// Ensure valid status code
+	if statusCode == 0 {
+		statusCode = http.StatusInternalServerError
+
+		if errorObj == nil {
+			errorObj = createErrorResponse(errMissingStatusCode)
+		}
```
aryanmehrotra, line 81: will the user be getting this error that errMissingStatusCode?

aryanmehrotra (changes requested):

pkg/gofr/http/responder.go
Umang01-hash: This code was added for a failSafe condition if system produces an invalid status code (0) AND the error object is nil. Will remove if not needed

pkg/gofr/http/responder_test.go
Umang01-hash: Sure changed to your seconf suggestion seems much more readable and efficient.

pkg/gofr/http/responder_test.go
```diff
@@ -207,6 +209,125 @@ func TestResponder_TemplateResponse(t *testing.T) {
...
+
+	var body map[string]any
+	err := json.NewDecoder(resp.Body).Decode(&body)
+	require.NoError(t, err)
+
+	errorObj, ok := body["error"].(map[string]any)
+	require.Truef(t, ok, "Expected error object to be of type map[string]any, got %T", body["error"])
+
+	assert.Equal(t, "original message", errorObj["message"])
+	assert.Equal(t, "additional info", errorObj["info"])
```
ccoVeille, line 276: I thought you would used JSONEq here too

pkg/gofr/http/responder.go
```diff
@@ -58,6 +63,45 @@ func (r Responder) Respond(data any, err error) {
...
 
+func (r Responder) determineResponse(data any, err error) (statusCode int, errObj any) {
+	// Handle empty struct case first
+	if err != nil && isEmptyStruct(data) {
+		return http.StatusInternalServerError, createErrorResponse(errEmptyResponse)
+	}
+
+	statusCode, errorObj := getStatusCode(r.method, data, err)
+
+	return statusCode, errorObj
```
coolwednesday, line 74:
```suggestion

	statusCode, errorObj := getStatusCode(r.method, data, err)
	
	if statusCode == 0 {
		statusCode = http.StatusInternalServerError
	}

	return statusCode, errorObj
```

We need to check for statusCode being 0, in which case WriteHeader will panic.

coolwednesday (changes requested):

pkg/gofr/http/responder.go
Umang01-hash: Sure corrected

coolwednesday (approved):

aryanmehrotra (approved):

PR #1617, opened 2025-03-30
FIX/NATS connection setup

### LINKED ISSUE #1613: Report a bug of connect to NATS
**Describe the bug**
The current code uses an asynchronous way to connect to the NATS server：
```
func (cm *ConnectionManager) Connect() error {
	go cm.retryConnect()

	return nil
}

```
In Client.Connect() ，calling connManager.Connect() and then immediately getting jetStream() is likely to fail, resulting in the errJetStreamNotConfigured , because the asynchronous connection has not been completed yet.
```
      if err := connManager.Connect(); err != nil {
		c.logger.Errorf("failed to connect to NATS server at %v: %v", c.Config.Server, err)
		return err
	}

	c.connManager = connManager

	js, err := c.connManager.jetStream()
	if err != nil {
		return err
	}
```
**To Reproduce**

1. The code is

   ```go
   app := gofr.New()
    app.AddPubSub(nats.New(nats.Config{
        Server:     "nats://localhost:4222",
        Stream: nats.StreamConfig{
            Stream:   "mystream",
            Subjects: []string{"orders.*", "shipments.*"},
        },
        MaxWait:     5 * time.Second,
        MaxPullWait: 500 * time.Millisecond,
        Consumer:    "my-consumer",
        CredsFile:   "/path/to/creds.json",
    }))
      ...
    app.Run()
  ```

2. The error is

   ```
   
   ```

**Expected behavior**

**Screenshots**

- OS: [e.g. Linux]
- gofr version [e.g. v1.5.0]
- go version [e.g. 1.21]

**More description**

Sumit-Kumar1:
## Pull Request Template

**Description:**
- Modified NATS connection setup to synchronous execution, preventing panics by ensuring operations occur only after a valid connection is established.
- Closes #1613 

**Checklist:**


pkg/gofr/datasource/pubsub/nats/connection_manager.go
```diff
@@ -79,7 +79,7 @@ func NewConnectionManager(
 
 // Connect establishes a connection to NATS and sets up JetStream.
 func (cm *ConnectionManager) Connect() error {
-	go cm.retryConnect()
+	cm.retryConnect()
 
 	return nil
 }
```
coolwednesday, line 85: Won't this make it a blocking call ? 

A seperate goroutine was introduced to allow an infinite loop to run in case of no connection to retry connecting). So in case of no connection here ....it would enter in a blocking call, not executing other processes. Any reasoning as to why that can be okay to introduce ?, As I feel the opposite....

coolwednesday (changes requested):

pkg/gofr/datasource/pubsub/nats/connection_manager.go
Sumit-Kumar1: Hi @coolwednesday , I am currently trying to leverage the NATS sdk's reconnect according to this document : https://docs.nats.io/using-nats/developer/connecting/reconnect.

Will update my PR and test out the flow according to your comment too.

coolwednesday: Setting up a small test project for NATS based on the [documentation](https://gofr.dev/docs/advanced-guide/using-publisher-subscriber#nats-jet-stream) and evaluating the following scenarios in both the current implementation and your implementation can help:
- Failing to reconnect right from the start
- Losing connection during the application's runtime

Let me know if you need any assistance.

Sumit-Kumar1: Updated the retry logic for NATS client altogether and tested out the publisher and subscriber for the above scenarios. Please review it.

Meanwhile I'll add the testing for newly added changes

Umang01-hash:
@Sumit-Kumar1 Please resolve the code quality issue:
Screenshot 2025-04-04 at 11 44 56 AM

Also it will be good to add tests for the newly added code.

pkg/gofr/datasource/pubsub/nats/client.go
```diff
@@ -8,14 +8,15 @@ import (
...
+	"github.com/nats-io/nats.go"
 	"github.com/nats-io/nats.go/jetstream"
 	"go.opentelemetry.io/otel/trace"
 	"gofr.dev/pkg/gofr/datasource/pubsub"
 )
 
 //go:generate mockgen -destination=mock_tracer.go -package=nats go.opentelemetry.io/otel/trace Tracer
 
-const defaultRetryTimeout = 10 * time.Second
+const defaultRetryTimeout = nats.DefaultTimeout
```
vikash, line 19: To import an entire package for a constant value may not be required.

pkg/gofr/datasource/pubsub/nats/client.go
```diff
@@ -64,6 +68,36 @@ func (c *Client) Connect() error {
 	return nil
 }
 
+func (c *Client) retryConnect() {
+	for {
+		c.logger.Debugf("connecting to NATS server at %v", c.Config.Server)
+
+		connManager := NewConnectionManager(c.Config, c.logger, c.natsConnector, c.jetStreamCreator)
```
vikash, line 75: This code is duplicated. We could refactor this.

Sumit-Kumar1: Refactored retry connect and connect methods added tests too.

Umang01-hash (approved):

PR #1640, opened 2025-04-10
En/http redirects

### LINKED ISSUE #1531: How to Perform Redirects in Gofr?
I am trying to redirect users to a frontend URL after authentication in Gofr. However, I couldn't find a built-in Redirect method in ctx.

Here’s my current code:

```
app.GET("/auth/google/callback", func(ctx *gofr.Context) (interface{}, error) {
	sessionManager := session.GetSession()

	email := sessionManager.GetString(ctx.Context, "email")
	name := sessionManager.GetString(ctx.Context, "name")

	// Expected to redirect but ctx does not have a Redirect method
	frontendURL := fmt.Sprintf("https://frontend.com/dashboard?email=%s&name=%s", email, name)

	// What is the correct way to redirect?
	ctx.Redirect(frontendURL, http.StatusFound) // This does not exist in Gofr

	return nil, nil
})
```
**Expected Behavior**
A way to redirect users to the frontend URL after authentication.

**Actual Behavior**
ctx.Redirect() does not exist in Gofr. like in echo or Fiber
There is no clear documentation on how to handle redirects.

**Questions**
What is the correct way to redirect users in Gofr?
Is there an alternative method to set the Location header?

Umang01-hash:
## Pull Request Template

**Description:**

-   Resolves #1531 

#### Example:

```go
package main

import (
	"gofr.dev/pkg/gofr"

	"gofr.dev/pkg/gofr/http/response"
)

func main() {
	app := gofr.New()

	app.GET("/old-page", func(ctx *gofr.Context) (any, error) {
		// Redirect to a new URL 
		return response.Redirect{URL: "https://example.com/new-page"}, nil
	})

	app.Run()
}

```

Screenshot 2025-04-10 at 1 15 45 PM

**Additional Information:**

**Checklist:**


coolwednesday:
@Umang01-hash , Can we also check and add tests for url with path and query params get handled successfully through the redirect. I feel that is not the case right now.

docs/advanced-guide/overriding-default/page.md
```diff
@@ -115,6 +115,33 @@ func listHandler(ctx *gofr.Context) (any, error) {
...
+)
+
+func main() {
+    app := gofr.New()
+    
+    app.GET("/old-page", func(ctx *gofr.Context) (any, error) {
+        // Redirect to a new URL with 301 Moved Permanently status
+        return ctx.Redirect("/new-page", http.StatusMovedPermanently)
+    })
+
```
coolwednesday, line 140: Let's not have a redirect method. We can just expose response of redirect Type ..... Also, we must ensure the traces are in hierarchy.

coolwednesday (changes requested):

docs/advanced-guide/overriding-default/page.md
vikash: I agree. Till now, all responses in Gofr are a type which is returned. ctx.Redirect is coming from other framework's behaviour where things like ctx.JSON etc are method. Also, we do not ask people to set manual status codes - but, we automatically decide. We should do the same and use a proper status code. The usecase in APIs for redirects are never going to be 301 - I think.

pkg/gofr/http/responder.go
```diff
@@ -42,6 +43,19 @@ func (r Responder) Respond(data any, err error) {
 		v.Render(r.w)
 
 		return
+	case *resTypes.Redirect:
+		statusCode = http.StatusFound // 302 by default
+
+		if r.method == http.MethodPost || r.method == http.MethodPut ||
+			r.method == http.MethodPatch {
+			statusCode = http.StatusSeeOther
+		}
```
ccoVeille, line 52: Please consider using a switch for readability purpose

pkg/gofr/http/response/redirect.go
```diff
@@ -0,0 +1,13 @@
+package response
+
+type Redirect struct {
+	URL string
+}
+
+// NewRedirect creates a redirect response with the specified URL and status code.
+// If statusCode is 0, it defaults to 302 (Found).
```
ccoVeille, line 8:
```suggestion
```

I feel like this comment is now outdated, no?

Umang01-hash: Yess removed it

pkg/gofr/http/responder.go
Umang01-hash: refactored with a `switch` case.

coolwednesday:
> @Umang01-hash , Can we also check and add tests for url with path and query params as well as request Bodies, headers etc.....get handled successfully through the redirect. I feel that is not the case right now.

@Umang01-hash, Can we add test for these cases ? I feel this implementation can fall short in fitting these usecases.

docs/advanced-guide/overriding-default/page.md
coolwednesday: In redirects, on the client-side :

- For 301, 302, 303 → body is usually dropped; method becomes GET
- For 307, 308 → body is preserved; method is unchanged

I think instead of setting internally, we should allow mentioning it in the response itself so that the client browser can process the urls accordingly.

pkg/gofr/http/responder_test.go
```diff
@@ -348,3 +348,21 @@ func removeTemplateDir(t *testing.T) {
...
 	require.NoError(t, err)
 }
+
+func TestResponder_RedirectResponse(t *testing.T) {
+	recorder := httptest.NewRecorder()
+	r := NewResponder(recorder, http.MethodGet)
+
+	// Set up redirect with specific URL and status code
+	redirectURL := "/new-location"
+	statusCode := http.StatusFound // 301
```
ccoVeille, line 358:
```suggestion
	statusCode := http.StatusFound // 302
```

ccoVeille:
While I get the logic of having a default code for get, post, ...

I don't understand why it's not possible to use a permanent redirect (HTTP 301) for example

pkg/gofr/http/responder_test.go
```diff
@@ -348,3 +348,21 @@ func removeTemplateDir(t *testing.T) {
 
 	require.NoError(t, err)
 }
+
+func TestResponder_RedirectResponse(t *testing.T) {
+	recorder := httptest.NewRecorder()
+	r := NewResponder(recorder, http.MethodGet)
```
ccoVeille, line 354: You should also test at least POST and HEAD

pkg/gofr/http/responder.go
```diff
@@ -47,6 +48,19 @@ func (r Responder) Respond(data any, err error) {
 		v.Render(r.w)
 
 		return
+	case *resTypes.Redirect:
+		switch r.method {
+		case http.MethodPost, http.MethodPut, http.MethodPatch:
+			statusCode = http.StatusSeeOther // 303
+		default:
+			statusCode = http.StatusFound // 302 by default
+		}
```
ccoVeille, line 57: While I get the logic of having a default code for get, post, ...

I don't understand why it's not possible to use a permanent redirect (HTTP 301) for example.

Here the code is hard coded. I can live with it. I'm simply raising the point

ccoVeille (changes requested):
@Umang01-hash cc @coolwednesday 

I got not feedback on these

- https://github.com/gofr-dev/gofr/pull/1640#discussion_r2053982471
- https://github.com/gofr-dev/gofr/pull/1640#discussion_r2055374025
- https://github.com/gofr-dev/gofr/pull/1640#discussion_r2053990282


so I'm moving my comment to "request changes"

pkg/gofr/http/responder.go
vikash: @ccoVeille Permanent redirect will probably be handled at the infra level itself. Like entire domain or URL prefix shifting from one place to another. Can you think of an API use-case for 301?

ccoVeille: Here my question was more about the fact there is a redirect component that doesn't allow you to do what you want.

I would have expected to be able to provide the http code if I need it.

ccoVeille: I don't have such examples right now.

It appears I'm not able to convince you. It's not a problem for me.

I have a master degree in "over thinking things", so don't worries 😅

It can be merged like this.

If someone has a need for being able to use 301, 308 or another code than the one you hard code in the code, they will open an issue and discussion will start from there.

pkg/gofr/http/responder_test.go
Umang01-hash: I used the **POST** method in test and then used status code `303` for the same.

pkg/gofr/http/response/redirect.go
```diff
@@ -0,0 +1,12 @@
...
+type Redirect struct {
+	URL string
+}
+
+// NewRedirect creates a redirect response with the specified URL and status code.
+func NewRedirect(url string) Redirect {
+	return Redirect{
+		URL: url,
+	}
+}
```
coolwednesday, line 12: This can be removed.

coolwednesday (changes requested):

pkg/gofr/http/responder.go
```diff
@@ -47,6 +48,19 @@ func (r Responder) Respond(data any, err error) {
 		v.Render(r.w)
 
 		return
+	case resTypes.Redirect:
+		switch r.method {
+		case http.MethodPost, http.MethodPut, http.MethodPatch:
+			statusCode = http.StatusSeeOther // 303
+		default:
+			statusCode = http.StatusFound // 302 by default
+		}
```
ccoVeille, line 57: There is something I cannot explain: why we computed this line 30

```
```

And we here don't care about statusCode we computed

Shouldn't we check if data.type is not redirect before calling determineResponse ? And so redirect without calling it

pkg/gofr/http/response/redirect.go
Umang01-hash: Sure done

pkg/gofr/http/responder.go
```diff
@@ -30,6 +30,7 @@ func (r Responder) Respond(data any, err error) {
 	statusCode, errorObj := r.determineResponse(data, err)
 
 	var resp any
+
 	switch v := data.(type) {
```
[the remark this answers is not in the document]
Umang01-hash: Sure applied this suggestion

pkg/gofr/http/responder.go
Umang01-hash: I reordered it to:

```go
	var resp any

	switch v := data.(type) {
	case resTypes.File:
		r.w.Header().Set("Content-Type", v.ContentType)
		r.w.WriteHeader(http.StatusOK)

		_, _ = r.w.Write(v.Content)

		return
	case resTypes.Template:
		r.w.Header().Set("Content-Type", "text/html")
		v.Render(r.w)

		return
	case resTypes.Redirect:
		// HTTP 302 by default
		statusCode := http.StatusFound

		switch r.method {
		case http.MethodPost, http.MethodPut, http.MethodPatch:
			// HTTP 303
			statusCode = http.StatusSeeOther
		}

		r.w.Header().Set("Location", v.URL)
		r.w.WriteHeader(statusCode)

		return
	}

	switch v := data.(type) {
	case resTypes.Raw:
		resp = v.Data
	case resTypes.Response:
	default:
		if isNil(data) {
			data = nil
		}

	}

	r.w.WriteHeader(statusCode)

	_ = json.NewEncoder(r.w).Encode(resp)
}
```

coolwednesday (approved):
@aryanmehrotra, This PR has been tested and looks good to me. Let me know your thoughts

pkg/gofr/http/responder.go
ccoVeille: Sounds way better indeed

Umang01-hash: Sure i reordered it as suggested.

pkg/gofr/http/response/redirect.go
Umang01-hash: sure removed.

pkg/gofr/http/responder.go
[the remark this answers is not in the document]
Umang01-hash: sure done

pkg/gofr/http/responder.go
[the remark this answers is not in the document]
Umang01-hash: Applied the suggestion. Thankyou

ccoVeille (approved):

aryanmehrotra (approved):

aryanmehrotra (approved):

PR #1662, opened 2025-04-17
doc: fix Go version in README and quick-start guide

ccoVeille:
go.mod fixes minimal Go version to 1.24

So the documentation needs to be updated to reflect the new minimum Go version.


Umang01-hash (approved):

coolwednesday (approved):

aryanmehrotra:
Shouldn't this be 1.22 as we are running the tests in the three versions in the pipeline starting version 1.22

ccoVeille:
> Shouldn't this be 1.22 as we are running the tests in the three versions in the pipeline starting version 1.22

go.mod enforces the minimum version that is needed to use the gofr modules within a project.

So it means it's the minimal version is Go 1.24

The fact you are running the tests on older versions means either:
- the minimal version in go.mod was previously updated for wrong reasons
- the tests are run on a version that cannot be used when using gofr as a module
- the tests are run on Go 1.24 even if you are using Go 1.22 to launch the tests. (I'm 80% sure about this)

I didn't reply earlier because I was busy with family things and I need a computer to check all this.

But what I can tell is that right now the minimal version to run GoFr is Go 1.24

We can discuss about what could be done to fix the minimal version, but for now this PR updates the documentation to match what the code expects.

Umang01-hash (approved):

PR #1663, opened 2025-04-17
fix: typo in opentsdb SubQuery struct

ccoVeille:
This is a break change.


ccoVeille:
Based on the fact that opentsdb is broken because of #1664 

- #1664 

I would say it's OK

coolwednesday:
> Based on the fact that opentsdb is broken because of #1664
> 
> * [gofr bump gofr.dev/pkg/gofr/datasource/opentsdb to v0.2.1 #1664](https://github.com/gofr-dev/gofr/issues/1664)
> 
> I would say it's OK

I somewhat agree. Since opentsdb has not been released since, that fix and we are not on our stable version in cases of datasources, I think this change is acceptable.

@aryanmehrotra , @Umang01-hash , let me know your thoughts.

coolwednesday (approved):

Umang01-hash:
@coolwednesday i agree with you. We will release a stable version of opentsdb with the next gofr release.

Umang01-hash (approved):

PR #1668, opened 2025-04-21
Enabled usetesting linter in golangci-lint

### LINKED ISSUE #1612: consider enabling usetesting linter in golangci-lint
https://golangci-lint.run/usage/linters/#usetesting

https://github.com/ldez/usetesting

Now you are using Go 1.24, we can activate it easily.

257 errors about the usage of [t.Context](https://pkg.go.dev/testing#T.Context)
2 about [t.Setenv](https://pkg.go.dev/testing#T.Setenv)
1 about [t.TempDir](https://pkg.go.dev/testing#T.TempDir)

Yadavanurag13:
This PR closes #1612 

**Description:**

-   Enabled usetesting

**Checklist:**


.golangci.yml
```diff
@@ -147,6 +147,40 @@ linters:
...
+    
+        # Enable/disable `context.Background()` detections.
+        # Disabled if Go < 1.24.
+        # Default: true
+        context-background: false
+    
+        # Enable/disable `context.TODO()` detections.
+        # Disabled if Go < 1.24.
+        # Default: true
+        context-todo: false
```
ccoVeille, line 183: Please remove this that is not used.

The config is based on golangci-lint  v1, so the node for this would be linter-settings.usetesting not `settings.usetesting`

Also, you are disabling all usetesting configuration by reverting every single setting to the opposite of their default value


```suggestion
```

Also, you are breaking the comment about the linters that is disabled on purpose like godot

go.work.sum
ccoVeille, line 1: I don't understand how this file could be modified without having a change in go.mod

Maybe the go.mod/go.sum is not clean, but then it should be a separate PR

My guess is this was altered during your local tests and was committed by error

ccoVeille (commented):
LGTM 👍 except minors remarks

.golangci.yml
Yadavanurag13: I have removed these changes.

go.work.sum
Yadavanurag13: I have rebased my pr on latest development branch. may be there I have committed it by err.

Fixing it.

Yadavanurag13:
@ccoVeille I have pushed the final changes and addressed all the comments. However, there's still a diff showing in the go.work.sum file, even though the contents are exactly the same. Could you please review it once?

Thanks !!

ccoVeille (approved):

Umang01-hash (approved):

aryanmehrotra (approved):

Issue #1667, opened 2025-04-21
Document American English is preferred in CONTRIBUTIONS.md


ccoVeille:
As stated here

> we follow American English conventions in this project, so the use of "favor" is intentional and not a typo.
_Originally posted by @coolwednesday in https://github.com/gofr-dev/gofr/pull/1656#issuecomment-2818107218_

This should be mentioned in the contributions information

Omar8345:
will work on this, @vikash

PR #1679, opened 2025-04-24
CI: add typos-action to GHA

### LINKED ISSUE #1666: Set up a typo detectors, focused on American English
As stated here

> we follow American English conventions in this project, so the use of "favor" is intentional and not a typo.
_Originally posted by @coolwednesday in https://github.com/gofr-dev/gofr/pull/1656#issuecomment-2818107218_

A tool like the following ones could be set up to enforce typo detection:
- cspell
- codespell
- typos

All of them supports:
- American vs British/Australian English
- GitHub action

It would help people to contribute to project, it would also avoid typos

ccoVeille:
- **chore: rename file with typo in it**
- **fix: typos in documentation and code**
- **ci: add typos detection in GitHub actions**

Fixes #1666 
- #1666 

It follows 
- #1678
- #1667


ccoVeille:
This should help from now to avoid typos in code and documentation.

yshngg:
Hi @ccoVeille , this isn't a typo: "initialize" (US) and "initialise" (UK) have the same meaning.

Ref: https://dictionary.cambridge.org/us/dictionary/english/initialize

ccoVeille:
It's funny it caught the exact example that was added to CONTRIBUTING via https://github.com/gofr-dev/gofr/commit/d3716681e01753e6dd5555c9bddab6ac5643b54f that @Omar8345 added.

I will add an inline comment to remove this one for being reported

Omar8345:
@ccoVeille how funny - the CI definitely overdid its job here! 😂

ccoVeille:
@Umang01-hash @vikash @coolwednesday @aryanmehrotra 

Is there anything blocking with thie PR?

Omar8345:
@ccoVeille nice catch for this issue, all the best

Umang01-hash:
@ccoVeille the PR looks good to me.
One minor suggestion is can we run this typo's job parallel to our other jobs instead of a new workflow. This way we can complete all the checks in a single workflow run.

ccoVeille:
> One minor suggestion is can we run this typo's job parallel to our other jobs instead of a new workflow. This way we can complete all the checks in a single workflow run.

I'm afraid I don't understand the point.

Here is my understanding of the GitHub CI via GHA
- all workflows are run in parallel, there is no logic of ordering
- all jobs in a workflow are run in parallel
- steps are run sequentially in a job

Right now, there are two workflows:
- one for the website (only on docs/)
- one for the code (docs/ is excluded)

I set up the typos workflows in a way typos detection is launched only once, in a way one PR containing one typo in the code and one typo in the documentation will lead to one job failing and not two, to make it easier to follow.

Maybe I don't understand what you meant,
so please let me know if it's clearer

ccoVeille:
As discussed here, and then on Discord in DM with @Umang01-hash , the current workflows files cannot help us.

The main raison are that:
- `website` workflow is only called with something is tagged. So it's too late to catch things
- `go` workflow is not called when the documentation is updated via a PR

coolwednesday:
> As discussed here, and then on Discord in DM with @Umang01-hash , the current workflows files cannot help us.
> 
> The main raison are that:
> 
> * `website` workflow is only called with something is tagged. So it's too late to catch things
> * `go` workflow is not called when the documentation is updated via a PR

Hey @ccoVeille , I would suggest having the docs ignore filter added in each job, instead of at the very start of the workfkow file. In that case you would be able to control what jobs should run how.

coolwednesday (changes requested):
https://github.com/gofr-dev/gofr/pull/1679#issuecomment-2853412640

Umang01-hash:
> > > As discussed here, and then on Discord in DM with @Umang01-hash , the current workflows files cannot help us.
> > > The main raison are that:
> > > 
> > > * `website` workflow is only called with something is tagged. So it's too late to catch things
> > > * `go` workflow is not called when the documentation is updated via a PR
> > 
> > 
> > Hey @ccoVeille , I would suggest having the docs ignore filter added in each job, instead of at the very start of the workfkow file. In that case you would be able to control what jobs should run how.
> 
> That's one way, indeed. I'm fine with doing the following changes, if two people confirm me they are OK with the following plan:
> 
> * include typos detection as a job in the file currently named go.yml
> * update all jobs currently in go.yml to exclude docs/ path
> * prefix all jobs currently in go.yml to mention they are about Go
> * rename go.yml workflow with another name (maybe code.yml)
> 
> I can do it, but for me, there is no logic in keeping the number of workflow low.
> 
> And here these changes are pretty important for almost no gain.
> 
> That said, if there is a consensus in such need, I can do it.

@coolwednesday I think what @ccoVeille suggested is right. Doing all the above changes just at cost of not making new workflow maybe isn't the best idea.

Umang01-hash (approved):

coolwednesday:
> > > > As discussed here, and then on Discord in DM with @Umang01-hash , the current workflows files cannot help us.
> > > > The main raison are that:
> > > > 
> > > > * `website` workflow is only called with something is tagged. So it's too late to catch things
> > > > * `go` workflow is not called when the documentation is updated via a PR
> > > 
> > > 
> > > Hey @ccoVeille , I would suggest having the docs ignore filter added in each job, instead of at the very start of the workfkow file. In that case you would be able to control what jobs should run how.
> > 
> > 
> > That's one way, indeed. I'm fine with doing the following changes, if two people confirm me they are OK with the following plan:
> > 
> > * include typos detection as a job in the file currently named go.yml
> > * update all jobs currently in go.yml to exclude docs/ path
> > * prefix all jobs currently in go.yml to mention they are about Go
> > * rename go.yml workflow with another name (maybe code.yml)
> > 
> > I can do it, but for me, there is no logic in keeping the number of workflow low.
> > And here these changes are pretty important for almost no gain.
> > That said, if there is a consensus in such need, I can do it.
> 
> @coolwednesday I think what @ccoVeille suggested is right. Doing all the above changes just at cost of not making new workflow maybe isn't the best idea.

Sure, I was suggesting a way around to have it in the same workflow. It is a lot of work, indeed.

coolwednesday (approved):

aryanmehrotra (approved):

Issue #1672, opened 2025-04-23
Disable Logs for Health and Alive Routes


leonardobiffi:
**Is your feature request related to a problem? Please describe.**  
I'm always frustrated when my logs are cluttered with repetitive entries from health check requests. This makes it difficult to identify more critical issues in the logs since health and alive routes are constantly being called.

```
{"level":"INFO","time":"2025-04-23T17:53:53.76753377Z","message":{"trace_id":"59357705d3dcaa0b86617d4909a953cf","span_id":"2050085ec4e91954","start_time":"2025-04-23T17:53:53.76641276+00:00","response_time":1117,"method":"GET","user_agent":"kube-probe/1.32+","ip":"10.10.20.130:39200","uri":"/.well-known/health","response":200},"gofrVersion":"v1.36.0"}

{"level":"INFO","time":"2025-04-23T17:53:54.711014144Z","message":{"trace_id":"6525ab152bfb93c9141b9fa6f2814c4f","span_id":"a1fde3166a3eaf7e","start_time":"2025-04-23T17:53:54.710916671+00:00","response_time":94,"method":"GET","user_agent":"kube-probe/1.32+","ip":"10.10.20.130:39204","uri":"/.well-known/alive","response":200},"gofrVersion":"v1.36.0"}
```

**Describe the solution you'd like**  
I would like an option to disable logging for the health (`/.well-known/health`) and alive (`/.well-known/alive`) routes. This would help reduce log noise and allow for easier monitoring of important events.

**Describe alternatives you've considered**  
An alternative solution could be to change the log level for these specific routes, but completely disabling logs would be more effective in reducing clutter.

**Additional context**  
This feature would be beneficial for environments where health checks are frequent, such as in Kubernetes, and would improve the overall log management experience.

vikash:
We can introduce an environment variable LOG_DISABLE_PROBES, which will default to false. But when set to true, will not log these lines.

PR #1681, opened 2025-04-28
fix: disable logs for healthcheck endpoints

leonardobiffi:
## Pull Request Template

**Description:**

This PR implements changes to disable healthcheck logs within the application.

- **Changes Made:**
  - Modified the logging configuration to suppress healthcheck log messages

- **Issue:**
  - This PR addresses issue #1672

- **Motivation Behind the Changes:**
  - The motivation for disabling healthcheck logs stems from the need to reduce log verbosity and improve readability Healthcheck logs can generate excessive noise in the log files, making it harder to identify critical issues

- **Expected Benefits:**
  - By disabling these logs, we expect:
    - Improved clarity in log files
    - Easier troubleshooting and monitoring of application performance
    - Reduced storage usage for log data

**Checklist:**


pkg/gofr/http/middleware/logger.go
```diff
@@ -93,6 +93,12 @@ type logger interface {
 func Logging(logger logger) func(inner http.Handler) http.Handler {
 	return func(inner http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// disable logging for health check endpoints
+			if r.URL.Path == "/.well-known/health" || r.URL.Path == "/.well-known/alive" {
+				inner.ServeHTTP(w, r)
+				return
+			}
+
```
ssshekhu53, line 101: This will disable the logging every time. As @vikash mentioned [in this comment](https://github.com/gofr-dev/gofr/issues/1672#issuecomment-2826314364), we need to have an environment variable `LOG_DISABLE_PROBES` that will act as the flag to enable/disable logging.

ssshekhu53 (changes requested):

pkg/gofr/http/middleware/logger.go
Umang01-hash: @leonardobiffi what @ssshekhu53 suggested is correct. We need a config based approach to disable these logs for the users who want it.

leonardobiffi: @ssshekhu53 @Umang01-hash I added the suggestion, let me know if it’s something like what you thought.

pkg/gofr/http/middleware/logger_test.go
```diff
@@ -3,19 +3,42 @@ package middleware
...
+
+	// Write data to the env file
+	for key, value := range envData {
+		_, err := fmt.Fprintf(envFile, "%s=%s\n", key, value)
+		if err != nil {
+			t.Fatalf("unable to write to file: %v", err)
+		}
+	}
+}
+
```
ssshekhu53, line 41: Is is just for creating env file for test? If so, then can `t.SetEnv` be used for the same?

ssshekhu53 (changes requested):
Passing whole config to the `Logging` function doesn't seem a good idea to me. I would suggest to just pass a flag. We can make use of `middleware.GetConfigs`. There, that `LOGS_DISABLE_PROBE` can be read and put in the map returned by it. And in the `newHTTPServer` function, we can get the value from `middlewareConfigs` map and pass into the `Logging` function. In this way, we can achieve a sort of abstraction by passing only the necessary values to `Logging` function rather that giving it access to the whole configs. 

Also, since `Logging` is an exported function, I think it will be a breaking change to update it's signature. Possibly, a new function can be created and `newHTTPServer` can be updated to inject the respective middleware on the basis of flag.

@Umang01-hash what's your thought on this?

pkg/gofr/http/middleware/logger_test.go
leonardobiffi: I removed this, based on suggestion to dont pass whole config to the Logging function

pkg/gofr/http_server.go
```diff
@@ -29,17 +29,22 @@ var (
...
+		middleware.CORS(middlewareConfigs.CorsHeaders, r.RegisteredRoutes),
 		middleware.Metrics(c.Metrics()),
 	)
 
+	if middlewareConfigs.LogDisableProbes == "true" {
+		// r.Use(middleware.LoggingWithoutProbes(c.Logger))
+	} else {
+		r.Use(middleware.Logging(c.Logger))
+	}
+
```
ssshekhu53, line 47: Didn't understand this. The `if` block is commented then why is it even required?

ssshekhu53:
> @ssshekhu53 I made changes so that we don't pass the entire config in `Logging`, and to not alter the function's signature. To do this, I made a change in the return of `middleware.GetConfigs` and added a new middleware for when `LOG_DISABLE_PROBES` is enabled. What do you think of these changes?

@leonardobiffi This one seems good to me. Just a small comment. If that's resolved, rest I can approve.

pkg/gofr/http_server.go
leonardobiffi: @ssshekhu53 Sorry, my bad
Its fixed now :)

leonardobiffi:
> > @ssshekhu53 I made changes so that we don't pass the entire config in `Logging`, and to not alter the function's signature. To do this, I made a change in the return of `middleware.GetConfigs` and added a new middleware for when `LOG_DISABLE_PROBES` is enabled. What do you think of these changes?
> 
> @leonardobiffi This one seems good to me. Just a small comment. If that's resolved, rest I can approve.

@ssshekhu53 Nice, thank you
I already fix the problem that you point. I hope everything is okay now :)

Umang01-hash:
@leonardobiffi There are code qality and linter issues in the PR:
Screenshot 2025-05-05 at 9 49 10 AM

Screenshot 2025-05-05 at 9 46 59 AM

Please fix it.

pkg/gofr/http/middleware/config_test.go
```diff
@@ -14,16 +14,27 @@ func TestGetConfigs(t *testing.T) {
...
+}
+
+func TestLogDisableProbesConfig(t *testing.T) {
+	mockConfig := config.NewMockConfig(map[string]string{
+		"LOG_DISABLE_PROBES": "true",
+	})
+
+	middlewareConfigs := GetConfigs(mockConfig)
+
+	assert.Equal(t, "true", middlewareConfigs.LogDisableProbes, "TestGetConfigs Failed!")
```
Umang01-hash, line 39: The error message on assert failure should be "**TestLogDisableProbesConfig** Failed".

pkg/gofr/http/middleware/logger_test.go
```diff
@@ -49,8 +49,8 @@ func Test_LoggingMiddleware(t *testing.T) {
 		req, _ := http.NewRequestWithContext(t.Context(), http.MethodGet, "http://dummy", http.NoBody)
 
 		rr := httptest.NewRecorder()
-
-		handler := Logging(logging.NewMockLogger(logging.DEBUG))(http.HandlerFunc(testHandler))
+		logger := logging.NewMockLogger(logging.DEBUG)
```
Umang01-hash, line 52: Is this change really needed?

pkg/gofr/http/middleware/logger_test.go
```diff
@@ -63,15 +63,29 @@ func Test_LoggingMiddlewareError(t *testing.T) {
 		req, _ := http.NewRequestWithContext(t.Context(), http.MethodGet, "http://dummy", http.NoBody)
 
 		rr := httptest.NewRecorder()
-
-		handler := Logging(logging.NewMockLogger(logging.ERROR))(http.HandlerFunc(testHandlerError))
+		logger := logging.NewMockLogger(logging.ERROR)
+		handler := Logging(logger)(http.HandlerFunc(testHandlerError))
```
Umang01-hash, line 67: Can we revert this change is not needed?

pkg/gofr/http/middleware/config.go
```diff
@@ -9,8 +9,15 @@ import (
 	"gofr.dev/pkg/gofr/config"
 )
 
-func GetConfigs(c config.Config) map[string]string {
-	middlewareConfigs := make(map[string]string)
+type MiddlewareConfig struct {
+	CorsHeaders      map[string]string
+	LogDisableProbes string
```
Umang01-hash, line 14: Shouldn't we keep the `LogDisableProbes` field as **bool** instead of string since the bool ensures that the field can only have true or false values, reducing the risk of invalid data and also avoid string comparisons later.

Umang01-hash (changes requested):

pkg/gofr/http/middleware/config.go
leonardobiffi: Yes, it makes sense. I kept it as a _string_ because I saw it that way in other parts of the code, but it really is better as a _bool_.

pkg/gofr/http/middleware/logger_test.go
leonardobiffi: Reverted :)

pkg/gofr/http/middleware/logger_test.go
leonardobiffi: Reverted :)

pkg/gofr/http/middleware/config_test.go
leonardobiffi: Fixed

Umang01-hash:
@leonardobiffi Thanks for making the review changes. One suggestion that was also being discussed in #1686 was regarding the use of two separate middleware instead of modifying the behaviour of `Logging` middleware.

I tried it by adding a flag there and than ran the application it was working as expected. Maybe we have a more cleaner approach for the same. Please try and use the Logging middleware for the same.

Screenshot 2025-05-06 at 2 42 47 PM

Umang01-hash:
> > @leonardobiffi Thanks for making the review changes. One suggestion that was also being discussed in #1686 was regarding the use of two separate middleware instead of modifying the behaviour of `Logging` middleware.
> > I tried it by adding a flag there and than ran the application it was working as expected. Maybe we have a more cleaner approach for the same. Please try and use the Logging middleware for the same.
> > Screenshot 2025-05-06 at 2 42 47 PM
> 
> @Umang01-hash @leonardobiffi Would like to have discussion on [this](https://github.com/gofr-dev/gofr/pull/1686#discussion_r2070820270) before moving further?

@kevin-lockhart-1001 Logging middeware is exported i understand that. Changing it will not be a breaking change because no-one else is calling it apart from us inside `http_server.go`.

leonardobiffi:
> > > @leonardobiffi Thanks for making the review changes. One suggestion that was also being discussed in #1686 was regarding the use of two separate middleware instead of modifying the behaviour of `Logging` middleware.
> > > I tried it by adding a flag there and than ran the application it was working as expected. Maybe we have a more cleaner approach for the same. Please try and use the Logging middleware for the same.
> > > Screenshot 2025-05-06 at 2 42 47 PM
> > 
> > 
> > @Umang01-hash @leonardobiffi Would like to have discussion on [this](https://github.com/gofr-dev/gofr/pull/1686#discussion_r2070820270) before moving further?
> 
> @kevin-lockhart-1001 Logging middeware is exported i understand that. Changing it will not be a breaking change because no-one else is calling it apart from us inside `http_server.go`.

Here's a revised version of your text with improved grammar and flow:

---

@Umang01-hash @kevin-lockhart-1001 My first approach was to change the `Logging` middleware to handle and ignore the health check routes. However, after @ssshekhu53's [suggestion](https://github.com/gofr-dev/gofr/pull/1681#pullrequestreview-2803365324), which I found compelling, it makes sense to have a different logging middleware to avoid changing the signature of the `Logging` function.

I think we have a dilemma: we can either keep the `Logging` function simple and change its signature, or we can have two logging middlewares to maintain the current signature but increase complexity. 

Personally, I find both approaches acceptable, but we need to decide which one to use.

Umang01-hash:
> > > > @leonardobiffi Thanks for making the review changes. One suggestion that was also being discussed in #1686 was regarding the use of two separate middleware instead of modifying the behaviour of `Logging` middleware.
> > > > I tried it by adding a flag there and than ran the application it was working as expected. Maybe we have a more cleaner approach for the same. Please try and use the Logging middleware for the same.
> > > > Screenshot 2025-05-06 at 2 42 47 PM
> > > 
> > > 
> > > @Umang01-hash @leonardobiffi Would like to have discussion on [this](https://github.com/gofr-dev/gofr/pull/1686#discussion_r2070820270) before moving further?
> > 
> > 
> > @kevin-lockhart-1001 Logging middeware is exported i understand that. Changing it will not be a breaking change because no-one else is calling it apart from us inside `http_server.go`.
> 
> @Umang01-hash @kevin-lockhart-1001 My first approach was to change the `Logging` middleware to handle and ignore the health check routes. However, after @ssshekhu53's [suggestion](https://github.com/gofr-dev/gofr/pull/1681#pullrequestreview-2803365324), which I found compelling, it makes sense to have a different logging middleware to avoid changing the signature of the `Logging` function.
> 
> I think we have a dilemma: we can either keep the `Logging` function simple and change its signature, or we can have two logging middlewares to maintain the current signature but increase complexity.
> 
> Personally, I find both approaches acceptable, but we need to decide which one to use.

@leonardobiffi  According to me changing Logging function is better.

coolwednesday (changes requested):

kevin-lockhart-1001:
> one

I am convinced with @kevin-lockhart-1001`s comment. I personally feel that adding a separate middleware isn't affecting the complexity that much that we compromise by having breaking change. 

Also, I see very good point from @coolwednesday's [comment](https://github.com/gofr-dev/gofr/pull/1681#discussion_r2074823906), that probably we missed, you might need to pass the health check endpoints to the middleware so as to cover the case of overridden health check endpoints which again convinces me to having a separate logging function.

Umang01-hash:
@kevin-lockhart-1001 @leonardobiffi we need to decide carefully on how can we check if a custom heath-endpoint is given and then we pass it on to the Logging function. 

Also @kevin-lockhart-1001 again i want to know how changing the siganture of `Logging` middleware will be a breaking change? Who else is calling or using it apart from GoFr codebase. In the codebase also we are calling it only once.

Can you please explaing me how this is a breaking change?

kevin-lockhart-1001:
> @kevin-lockhart-1001 @leonardobiffi we need to decide carefully on how can we check if a custom heath-endpoint is given and then we pass it on to the Logging function.
> 
> Also @kevin-lockhart-1001 again i want to know how changing the siganture of `Logging` middleware will be a breaking change? Who else is calling or using it apart from GoFr codebase. In the codebase also we are calling it only once.
> 
> Can you please explaing me how this is a breaking change?

Well, here's the deal: messing with that signature is kind of like opening a can of worms, and it's definitely a breaking change. Think of it this way: even though we in the **GoFr** codebase might only call it once (and you're right on the money there!), because it's an exported function – meaning it's out there in the wild with a big ol' "come and get it" sign – we can't be a hundred percent sure who else might be hitching their wagon to it.

So, yeah, even if our own house is in order, we gotta be real careful about what's out there in the neighborhood. Changing an exported function's signature is generally a no-go unless we're planning a major overhaul and are ready to deal with some potentially unhappy campers whose code suddenly breaks.

Hope that makes sense!

Umang01-hash:
@leonardobiffi Please go ahead with the changes.

leonardobiffi:
> @leonardobiffi Please go ahead with the changes.

@Umang01-hash I made the changes, and add this idea to set on `LOG_PATHS_PROBES` additional paths to be ignored on `Logging`, make sense to you?

Umang01-hash:
@leonardobiffi Thankyou for making the changes. While the idea of `LOG_PATHS_PROBES ` is not bad but if you see when we register an HTTP service like this:

```go
 app.AddHTTPService("abc", "http://localhost:9000",
        &service.HealthConfig{
            HealthEndpoint: "/health",
        },
        &service.CircuitBreakerConfig{
            Threshold: 1,
            Interval:  1 * time.Second,
        },
        &service.RetryConfig{
            MaxRetries: 3,
        },
    )
```

We already need to tell the custom health endpoint here. Then again specifying it in configs will be a bit repetitive. It will be much better if we handle it internally.

leonardobiffi:
> @leonardobiffi Thankyou for making the changes. While the idea of `LOG_PATHS_PROBES ` is not bad but if you see when we register an HTTP service like this:
> 
> ```go
>  app.AddHTTPService("abc", "http://localhost:9000",
>         &service.HealthConfig{
>             HealthEndpoint: "/health",
>         },
>         &service.CircuitBreakerConfig{
>             Threshold: 1,
>             Interval:  1 * time.Second,
>         },
>         &service.RetryConfig{
>             MaxRetries: 3,
>         },
>     )
> ```
> 
> We already need to tell the custom health endpoint here. Then again specifying it in configs will be a bit repetitive. It will be much better if we handle it internally.
> 
> Also we have a code climate issue here: **Cognitive Complexity of Logging is 14 (which exceeds allowed limit 12)**
> 
> Please resolve this also.

@Umang01-hash I reverted the changes on my current branch, because I was already quite confused with the changes. 
I made some adjustments to the `Logging` function, but the **Cognitive Complexity of Logging is still 13**. What do you suggest?

Also, I didn't understand why we would have to handle the `HealthEndpoint` when a customService is added, because looking at the code, the application’s healthcheck routes are fixed and by default they are always these: 
https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/factory.go#L47 

This `HealthEndpoint` config of the customService will be used in the `healthHandler` to make the request to the endpoint of the customService 
https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/container/health.go#L46 

But the routes `/.well-known/alive` and `/.well-known/health` will not change. 
I tested this on `examples/using-http-service`
image

Umang01-hash:
@leonardobiffi sorry my bad! I got the context wrong. You are absolutely right that the application's health-check endpoints are fixed and we don't need to handle the case of external http service's health endpoint.

Reagrding the cognitive complexity of logging i think we can break it down below 12 and then the issue from code climate will not occur.

Once again sorry for my previous suggestion regarding healthCheck endpoint and thankyou for bearing.

leonardobiffi:
> @leonardobiffi sorry my bad! I got the context wrong. You are absolutely right that the application's health-check endpoints are fixed and we don't need to handle the case of external http service's health endpoint.
> 
> Reagrding the cognitive complexity of logging i think we can break it down below 12 and then the issue from code climate will not occur.
> 
> Once again sorry for my previous suggestion regarding healthCheck endpoint and thankyou for bearing.

@Umang01-hash I created the function `handleRequestLog`, now the code complexity was reduced and pass.
I hope that all changes is ok now :)

Umang01-hash:
@leonardobiffi the `PKG Unit Testing` is failing and also we have a linter error.
Please resolve them.

coolwednesday (approved):

kevin-lockhart-1001 (changes requested):

pkg/gofr/http/middleware/logger.go
```diff
@@ -90,47 +90,69 @@ type logger interface {
...
-						logger.Log(l)
-					}
-				}
-			}(srw, r)
-
-			defer func() {
-				panicRecovery(recover(), srw, logger)
-			}()
+			defer handleRequestLog(srw, r, start, traceID, spanID, logger)
+			defer func() { panicRecovery(recover(), srw, logger) }()
```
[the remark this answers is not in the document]
Umang01-hash: @leonardobiffi please place the `panicRecovery` call correctly so that it should be called in both the cases if we receive panic.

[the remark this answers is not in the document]
leonardobiffi: @Umang01-hash @kevin-lockhart-1001 yes make sense, I changed to `panicRecovery` be define before the probe log check

Umang01-hash (approved):

kevin-lockhart-1001 (approved):

PR #1684, opened 2025-04-30
Adding examples and modifying grpc to add interceptor for streaming

coolwednesday:


examples/grpc/grpc-streaming-server/main_test.go
```diff
@@ -0,0 +1,131 @@
...
+			_ = stream.Send(&server.Request{Message: msg})
+			time.Sleep(100 * time.Millisecond)
+		}
+		_ = stream.CloseSend()
+	}()
+
+	var responses []string
+	for {
+		resp, err := stream.Recv()
+		if err == io.EOF {
```
ccoVeille, line 116:
```suggestion
		if errors.Is(err, io.EOF) {
```

examples/grpc/grpc-unary-server/main_test.go
```diff
@@ -2,32 +2,19 @@ package main
 
 import (
 	"context"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/credentials/insecure"
 	"os"
 	"testing"
 	"time"
```
ccoVeille, line 10: Please fix the go imports orders

examples/grpc/grpc-unary-server/main_test.go
```diff
@@ -82,3 +69,12 @@ func TestHelloProtoMethods(t *testing.T) {
...
 	assert.Equal(t, "message:\"Hello World\"", resp.String())
 }
+
+func TestMain(m *testing.M) {
+	os.Setenv("GOFR_TELEMETRY", "false")
+
+	go main()
+	time.Sleep(300 * time.Millisecond) // wait for server to boot
+
+	os.Exit(m.Run())
```
ccoVeille, line 79:
```suggestion
	m.Run()
```

This is the recommended way now. Previously calling os.Exit was necessary in TestMain, it's no longer the case

examples/grpc/grpc-streaming-client/main.go
```diff
@@ -0,0 +1,231 @@
...
+	"gofr.dev/examples/grpc/grpc-streaming-client/client"
+	"gofr.dev/pkg/gofr"
+	"io"
+	"time"
+)
+
+func main() {
+	app := gofr.New()
+
+	//Create a gRPC client for the Chat Streaming service
```
ccoVeille, line 15:
```suggestion
	// Create a gRPC client for the Chat Streaming service
```

pkg/gofr/grpc/log.go
```diff
@@ -150,7 +201,8 @@ func logRPC(ctx context.Context, logger Logger, metrics Metrics, start time.Time
 	}
 
 	if logger != nil {
-		if method == debugMethod {
+		if method == debugMethod || strings.Contains(method, "/Send") ||
+			strings.Contains(method, "/Recv") || strings.Contains(method, "/SendAndClose") {
 			logger.Debug(logEntry)
 		} else {
 			logger.Info(logEntry)
```
ccoVeille, line 208: A switch could be used here

examples/grpc/grpc-streaming-client/main.go
```diff
@@ -0,0 +1,241 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"gofr.dev/examples/grpc/grpc-streaming-client/client"
+	"gofr.dev/pkg/gofr"
+	"io"
+	"time"
```
ccoVeille, line 9: Fix imports order please

ccoVeille (approved):

examples/grpc/grpc-streaming-client/main.go
ccoVeille: I'm surprised golangci-lint didn't catch this.

ccoVeille: It comes from the setting `issues.exclude-dirs-use-default` that is set to true by default, and that excludes `examples` folder to be linted.

pkg/gofr/grpc.go
```diff
@@ -53,21 +54,32 @@ func (a *App) AddGRPCUnaryInterceptors(grpcInterceptors ...grpc.UnaryServerInter
 	a.grpcServer.interceptors = append(a.grpcServer.interceptors, grpcInterceptors...)
 }
 
+func (a *App) AddGRPCServerStreamInterceptors(grpcInterceptors ...grpc.StreamServerInterceptor) {
```
Umang01-hash, line 57: Since we are already inside `grpc.go`, i think we can name the function arguments as **interceptors** only 

Wherever applicable...

pkg/gofr/grpc/log.go
```diff
@@ -74,13 +75,62 @@ func (l gRPCLog) String() string {
 	return string(line)
 }
 
+// StreamObservabilityInterceptor handles logging, metrics, and tracing for streaming RPCs.
+func StreamObservabilityInterceptor(logger Logger, metrics Metrics) grpc.StreamServerInterceptor {
+	tracer := otel.GetTracerProvider().Tracer("gofr-stream", trace.WithInstrumentationVersion("v0.1"))
```
Umang01-hash, line 80: Since we can support multiple streaming types i think we should rename the span to `gofr-grpc-stream`.

pkg/gofr/grpc/log.go
```diff
@@ -74,13 +75,62 @@ func (l gRPCLog) String() string {
...
+		// Wrap the stream to propagate context with tracing
+		wrappedStream := &wrappedServerStream{
+			ServerStream: ss,
+			ctx:          ctx,
+		}
+
+		// Process the stream
+		err := handler(srv, wrappedStream)
+
+		fullMethod := info.FullMethod
```
Umang01-hash, line 101: Can we rename it to `grpcMethodName`?

Umang01-hash (changes requested):

Umang01-hash (approved):

ccoVeille (approved):

PR #1686, opened 2025-04-30
En/1672 disable health check log

kevin-lockhart-1001:
**Description:**

- Added a dedicated logging middleware to address the issue described in the ticket https://github.com/gofr-dev/gofr/issues/1672
- Refactored the newHTTPServer function to conditionally include the logging middleware based on the LOG_DISABLE_PROBES environment variable

**Checklist:**


pkg/gofr/gofr_test.go
```diff
@@ -1162,3 +1162,39 @@ func TestApp_Subscribe(t *testing.T) {
 		assert.False(t, ok)
 	})
 }
+
+func Test_NewDisableHealthCheckLog(t *testing.T) {
+	port := testutil.GetFreePort(t)
+
+	t.Setenv("HTTP_PORT", strconv.Itoa(port))
+	t.Setenv("LOG_DISABLE_PROBES", "true")
```
ccoVeille, line 1170: You should use the constant you define in the other file

pkg/gofr/http/middleware/config.go
```diff
@@ -9,6 +9,8 @@ import (
 	"gofr.dev/pkg/gofr/config"
 )
 
+const LogDisableProbeKey = "LOG_DISABLE_PROBES"
```
ccoVeille, line 12: The constant should have a clear comment about its purpose

Or simpl prefix it with EnvVar

```suggestion
const EnvVarLogDisableProbeKey = "LOG_DISABLE_PROBES"
```

pkg/gofr/http_server.go
```diff
@@ -33,9 +34,15 @@ func newHTTPServer(c *container.Container, port int, middlewareConfigs map[strin
 	r := gofrHTTP.NewRouter()
 	wsManager := websocket.New()
 
+	r.Use(middleware.Tracer)
+
+	if ok, _ := strconv.ParseBool(middlewareConfigs[middleware.LogDisableProbeKey]); !ok {
+		r.Use(middleware.Logging(c.Logger))
+	} else {
+		r.Use(middleware.LoggingSkipHealthCheck(c.Logger))
+	}
```
ccoVeille, line 43: This code looks strange to me

I mean, why having two middlewares?

pkg/gofr/http/middleware/config.go
ccoVeille, line 12: Does this one need to be exported ?

ccoVeille, line 12: I would have used


```suggestion
const LogDisableProbeKey = "LOG_PROBES_DISABLED"
```

Or something like that. Maybe there are other env variables about disabling something

pkg/gofr/http_server.go
vikash: True. Same middleware behaviour should be changed.

kevin-lockhart-1001: Okay, so changing how that logging middleware works now would mean we'd have to toss in a flag as an argument, and since that Logging function is out there for everyone, that'd be a big ol' breaking change. So, wouldn't it make more sense to just spin up a separate logging middleware and then just kinda decide whether to use it or not when we're setting up the server, based on that flag? 

@ccoVeille @vikash

pkg/gofr/http/middleware/config.go
kevin-lockhart-1001: If followed [this](https://github.com/gofr-dev/gofr/pull/1686/files#r2068996110) suggestion, then yes

vikash:
@kevin-lockhart-1001 Someone else has been implementing this as well: https://github.com/gofr-dev/gofr/pull/1681 It might be a good idea to collaborate on this.

Umang01-hash:
@kevin-lockhart-1001 Thankyou for contributing but since the issue has been resolved i am closing this PR.

PR #1704, opened 2025-05-11
Refactor logger to reduce cognitive complexity

anshuhimanshu001:
This PR refactors the logf function in logger.go to reduce its cognitive complexity and meet code quality standards. It also includes a condition to skip logging for health check endpoints when LOG_DISABLE_PROBES=true.

### DIFF
--- a/pkg/gofr/logging/logger.go
+++ b/pkg/gofr/logging/logger.go
@@ -1,241 +1,211 @@
 package logging
 
 import (
-	"encoding/json"
-	"fmt"
-	"io"
-	"os"
-	"time"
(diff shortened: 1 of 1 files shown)


ccoVeille:
This looks alike an AI generated PR…

coolwednesday:
@anshuhimanshu001, thanks for the effort, but this PR currently doesn’t appear to add significant value. Additionally, the issue regarding disabling health check logs is already being addressed by another contributor, so there's no need to include it in this PR.

Do checkout the other issues if you are interested in contributing. Let us know if you need any help, navigating.

PR #1795, opened 2025-05-28
Added http.StatusNotFound (404) in case of no rows found 

akshat-kumar-singhal:
**Description:**
Return http.StatusNotFound (404) from ErrorDB.StatusCode() in case there are no rows returned by methods for getting a single row, instead of the blanket http.StatusInternalServerError (500).
This change would allow the calling method to handle the error from store layer method in a better way. Current support:
- SQL
- Cassandra
- Mongo

N/A

**Additional Information:**
Support can be added for other data stores in future.

**Checklist:**


Umang01-hash:
@akshat-kumar-singhal Thankyou for contributing. We strongly encourage to raise the PR only when ready. Requesting you to please close it as of now and reopen whenever you feel it is ready.

Thankyou

Umang01-hash:
@akshat-kumar-singhal Thankyou for your Contribution While the idea of standardizing error responses is valuable, considering GoFr's support for multiple databases, integrating a generic error handling mechanism into the core framework may not be optimal. It could introduce unnecessary coupling and complexity.

A better approach would be to implement specific error handling within each database module. This way, we maintain the framework's modularity and ensure that each database's unique behaviors are correctly managed.

PR #1807, opened 2025-06-04
Feat: add GetGRPCServer public method

ace3:
## Pull Request Template

**Description:**

-   Added a unit test for the `GetGRPCServer` method in the `App` struct.
-   The new test ensures that the method returns a non-nil `*grpc.Server` instance after initialization.
-   This improves test coverage and validates the public API for accessing the underlying gRPC server.

-   No breaking changes are introduced by this PR.
-   All existing functionality remains backward compatible.

**Additional Information:**

-   No new dependencies or external libraries were added.
-   The test follows the existing test structure and uses `testify/require` for assertions.

**Checklist:**

### DIFF
--- a/pkg/gofr/grpc.go
+++ b/pkg/gofr/grpc.go
@@ -7,8 +7,8 @@ import (
 	"reflect"
 	"strconv"
 
-	"github.com/grpc-ecosystem/go-grpc-middleware"
-	"github.com/grpc-ecosystem/go-grpc-middleware/recovery"
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
+	grpc_recovery "github.com/grpc-ecosystem/go-grpc-middleware/recovery"
 	"google.golang.org/grpc"
 
 	"gofr.dev/pkg/gofr/container"
@@ -56,6 +56,22 @@ func (a *App) AddGRPCUnaryInterceptors(interceptors ...grpc.UnaryServerIntercept
 
 func (a *App) AddGRPCServerStreamInterceptors(interceptors ...grpc.StreamServerInterceptor) {
 	a.grpcServer.streamInterceptors = append(a.grpcServer.streamInterceptors, interceptors...)
+}
+
+// GetGRPCServer returns the gRPC server instance.
+// This is useful for accessing the server directly, for example, to register services or perform custom operations.
+// Note: This should be used with caution as it exposes the internal gRPC server instance.
+// It is recommended to use RegisterService method for registering gRPC services instead of directly manipulating the server.
+// Example usage:
+// grpcServer := app.GetGRPCServer()
+// grpcServer.RegisterService(&myServiceDesc, myServiceImpl)
+// For adding reflection
+// support, you can use the reflection package:
+// import "google.golang.org/grpc/reflection"
+// reflection.Register(grpcServer)
+
+func (a *App) GetGRPCServer() *grpc.Server {
+	return a.grpcServer.server
 }
 
 func newGRPCServer(c *container.Container, port int) *grpcServer {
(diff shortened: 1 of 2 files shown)


ace3:
This PR is used for my use case to enable the gRPC reflection of the service.

Screenshot 2025-06-04 at 10 10 13


```go

import (
	"google.golang.org/grpc/reflection"
)

func main() {

	app := gofr.New()

	greetServer.RegisterGreetServerWithGofr(app, greetServer.NewGreetGoFrServer())

        app.WithReflection()
	app.Run()
}
```

coolwednesday:
@ace3, Just a heads-up:
Please avoid making any new changes directly to the you fork's development branch. Any future branches created from development will inherit its commit history — including any recent changes. This can cause confusion as those commits will appear in your new branches as well.

Let’s keep development clean to ensure a smooth workflow going forward.

pkg/gofr/grpc.go
```diff
@@ -58,6 +59,20 @@ func (a *App) AddGRPCServerStreamInterceptors(interceptors ...grpc.StreamServerI
...
+//   app.WithReflection()
+
+func (a *App) WithReflection() {
+	if a.grpcServer.server == nil {
+		a.grpcServer.createServer()
+	}
+
+	reflection.Register(a.grpcServer.server)
+}
+
```
coolwednesday: After reviewing the use case for enabling gRPC reflection, here’s my suggestion:

Reflection should be enabled in the following scenarios:
- Development and Debugging: Tools like grpcurl, Postman, and Evans rely on reflection to explore available services and make requests without needing .proto files.
- Internal, Trusted Environments: In secure, internal networks that are not exposed to the public internet, reflection is generally safe and helpful for observability and tooling.
- CI/CD and Automation: Useful in staging or automated testing environments where dynamic service inspection is required.

Reflection should be disabled in the following scenarios:
- Production or Public-Facing Services: Enabling reflection in production introduces a security risk by exposing service definitions. This can help attackers map your API surface and craft targeted exploits.
- Zero-trust or Multi-tenant Environments: In such cases, reflection increases the potential attack surface and should be avoided.

Recommendation:
We should make reflection configurable via an environment variable. You can use `a.Config.GetOrDefault()` to toggle reflection based on the environment. This approach provides flexibility across different stages of deployment.

If this approach sounds good, we should also update the documentation accordingly.

coolwednesday (changes requested):

pkg/gofr/grpc.go
ace3: Hi, I just adjusted the code, if there is `GRPC_ENABLE_REFLECTION` set, and the value is true, it will enable the GRPC Reflection.

Set the environment variable when starting your app:

- For local/dev: export GRPC_ENABLE_REFLECTION=true
- For prod: export GRPC_ENABLE_REFLECTION=false (or just unset it)

The server will only register reflection if the variable is "true".

coolwednesday: Thank you so much !

coolwednesday (approved):

kairavaibhav (approved):

pkg/gofr/grpc.go
```diff
@@ -58,6 +60,25 @@ func (a *App) AddGRPCServerStreamInterceptors(interceptors ...grpc.StreamServerI
...
+		return
+	}
+
+	if a.grpcServer.server == nil {
+		a.grpcServer.createServer()
+	}
+
+	reflection.Register(a.grpcServer.server)
+}
+
```
coolwednesday: Hey! Since the `GRPC_ENABLE_REFLECTION` config is already defined, having a separate method to enable reflection feels redundant. Instead, the server should automatically handle it based on the config value.

It would be cleaner to move this logic directly into the `createServer()` function, so that reflection is enabled transparently whenever the config flag is set—no need for an additional method call.

Umang01-hash:
@ace3 Just a reminder you have a review comment left from @coolwednesday. Let's resolve it so that we can go ahead and merge your PR.

pkg/gofr/grpc.go
ace3: i just move it.
but need additional changes. since on grpcServer there is no config.Config.

ace3:
> @ace3 Just a reminder you have a review comment left from @coolwednesday. Let's resolve it so that we can go ahead and merge your PR.

hi, i just pushed the changes.

PR #1808, opened 2025-06-04
Oauth improvements

akshat-kumar-singhal:
**Description:**
_Background_:
In case the tokenURL is not provided in the OAuthConfig, an error is encountered while trying to fetch the token. The error returned is from the url package itself and does not provide much guidance to the user on what needs to be fixed. Error message: `Post "" unsupported protocol scheme ""`
_Fix Made_:
A new method NewOAuthConfig() has been created which is responsible for the validation of OAuth config. Users are required to use this method instead of directly initialising the OAuthConfig struct.

None

**Additional Information:**
We should consider converting the OAuthConfig struct to un-exported to enforce the usage of `NewOAuthConfig()` function

**Checklist:**

### DIFF
--- a/pkg/gofr/service/oauth.go
+++ b/pkg/gofr/service/oauth.go
@@ -11,7 +11,7 @@ import (
 	"golang.org/x/oauth2/clientcredentials"
 )
 
-const AuthorisationHeader = "Authorization"
+const AuthHeader = "Authorization"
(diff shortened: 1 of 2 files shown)


akshat-kumar-singhal:
Will re-create the PR once the base changes are merged

PR #1815, opened 2025-06-05
Release/v1.40.1

Umang01-hash:
# v1.40.1

## 🔧 Dependency Updates

Upgraded all externalized datasource dependencies to their latest stable versions for improved performance, security, and compatibility across supported services.


aryanmehrotra (approved):

pkg/gofr/version/version.go
```diff
@@ -1,3 +1,3 @@
 package version
 
-const Framework = "v1.40.0"
+const Framework = "v1.41.1"
```
aryanmehrotra: The version is not correct

aryanmehrotra (changes requested):

pkg/gofr/version/version.go
Umang01-hash: resolved

aryanmehrotra (approved):

Issue #1816, opened 2025-06-07
Make gauge metrics thread-safe


ashwingopalsamy:
In `pkg/gofr/metrics/register.go`, the `float64Gauge` type maintains a map of observations without any synchronization mechanism. 

For example: `SetGauge` modifies this map while HTTP handlers and other `goroutines` may concurrently read or update gauge values. As Go maps aren’t thread-safe for concurrent writes, simultaneous calls to `SetGauge` can lead to data races or undefined behaviors.

abrarr21:
Hey, I'd like to take this on, can you assign it to me

PR #1817, opened 2025-06-07
fix: ensure thread-safe handling of gauge metrics

abrarr21:
**Description:**

-   This PR addresses issue #1816.
-   Modified the metrics store and register logic to make gauge metrics thread-safe.
-   Changed gauge maps to store pointers (*float64Gauge) instead of values, preventing data races during concurrent access.
-   Updated all relevant interfaces and methods to use pointer receivers for consistency and safety.
-   Motivation: Fix race conditions and warnings related to concurrent gauge metric updates, improving reliability and correctness of metrics handling.

-  Store and register interfaces now use pointers for gauges, which might affect downstream code if directly accessing gauges.
-  These changes are necessary to ensure safe concurrent metric updates and prevent race conditions.

**Additional Information:**

-  No new dependencies added.

**Checklist:**


abrarr21:
@Umang01-hash @aryanmehrotra Please review this!!

abrarr21:
I wrote a dedicated unit test file to verify the behavior of the metric store, specifically focusing on the new changes. The test cases include:

1. Set and Get operations for all metric types (counter, upDownCounter, histogram, and the updated gauge).

2. Duplicate registration errors.

3. Fetching non-existent metrics.

4. Concurrency test for gauge metric using multiple goroutines to simulate real-world parallel usage.

I also ran all tests with -race to ensure there are no data races.
```go test -v -race ./pkg/gofr/metrics -run ^TestStore```

**Comment by Umang01-hash** [comment]
Hey @abrarr21 ! Your PR has some linter errors. Can you please fix them before we re-review your PR.

Screenshot 2025-06-11 at 10 54 56 AM

File: pkg/gofr/version/version.go [hunk 1]
```diff
@@ -1,3 +1,3 @@
 package version
 
-const Framework = "dev"
+const Framework = "v1.40.1"
```
coolwednesday: This is actually unrelated. Please remove this.

pkg/gofr/metrics/store.go
```diff
@@ -102,7 +102,7 @@ func (s store) setHistogram(name string, m metric.Float64Histogram) error {
 	return metricsAlreadyRegistered{metricsName: name}
 }
 
-func (s store) setGauge(name string, m float64Gauge) error {
+func (s *store) setGauge(name string, m *float64Gauge) error {
 	_, ok := s.gauge[name]
 	if !ok {
 		s.gauge[name] = m
```
coolwednesday: If we are trying to prevent race conditions, should we also not handle this case here ?

coolwednesday (changes requested):

pkg/gofr/metrics/store.go
abrarr21: You're right, thanks for pointing that out!
We're guarding against race conditions elsewhere, so it makes sense to do the same here. I’ll add a mutex lock around this setGauge function as well to ensure thread safety.

abrarr21: @coolwednesday Since each method in the store accesses shared maps (like counter, gauge, etc.),
should I apply locking (i.e., Lock() for writes and RLock() for reads) in every method — both getX() and setX() — to ensure thread safety?

Just want to confirm if internal locking is expected in all these functions, or if it's assumed that the caller will handle synchronization externally.

abrarr21:
@coolwednesday  @Umang01-hash @aryanmehrotra Added mutex locks to all getter and setter methods in the store to prevent race conditions when accessing metric maps concurrently. Getters use read locks, setters use write locks. This makes the store thread-safe.

coolwednesday (approved):

Umang01-hash (approved):

Umang01-hash:
Hey @abrarr21 , thanks for your contribution! **This PR contains 307 commits**, which makes it quite large for merging directly into `development`. To keep the commit history clean and avoid unnecessary conflicts, please **rebase your branch** onto the latest development branch

abrarr21:
> Hey @abrarr21 , thanks for your contribution! **This PR contains 307 commits**, which makes it quite large for merging directly into `development`. To keep the commit history clean and avoid unnecessary conflicts, please **rebase your branch** onto the latest development branch

@Umang01-hash Done

Umang01-hash (approved):

pkg/gofr/metrics/store_unit_test.go
```diff
@@ -0,0 +1,122 @@
+package metrics
```
aryanmehrotra: @abrarr21 store_unit_test.go needs to be renamed to store_test.go as that is the convention being followed everywhere

Issue #1833, opened 2025-06-11
Support to run a synchronous Job on app start


aryanmehrotra:
**Is your feature request related to a problem? Please describe.**
Sometimes the app needs to maintain an in-memory cache - but currently no method is there for the same and container or context using which we can make API calls the gofr way.

**Describe the solution you'd like**
A method probably on the App, or something like how migrations are there.

**Describe alternatives you've considered**
I can directly use NewHTTPService or NewMysql or any DB and then initialise them manually but this wont be a scalable approach

vikash:
Why is this even required?

Why can't this just be written normally before calling app.Run()

aryanmehrotra:
We’re looking to support a use case where data needs to be fetched on application startup — specifically, to populate an in-memory cache by calling external REST APIs or querying the database. This data must be available before the app starts serving traffic.

Currently, the gofr way recommends using ctx.HTTPService or ctx.DB, but those require a gofr.Context, which is only available during request handling — not during startup.
While one workaround is to create a cron job, it doesn't solve the need for blocking the app start until the job finishes.

Technically, yes — we could call NewHTTPService() or open DB connections manually before app.Run(). But this breaks the standard GoFr flow:

- It bypasses GoFr’s DI and lifecycle management
- Introduces duplicate config handling
- Makes the codebase inconsistent with other GoFr-based services

vikash:
```golang
        app := gofr.New()

	app.OnStart(func(a *gofr.App) error {
		// Do your synchronous startup work here
		// e.g., initialize cache, make API calls, etc.
		return nil
	})

	app.GET("/greet", func(ctx *gofr.Context) (any, error) {
		return "Hello World!", nil
	})
```

is equivalent to:

```golang
        app := gofr.New()

	// Do your synchronous startup work here
	// e.g., initialize cache, make API calls, etc.


	app.GET("/greet", func(ctx *gofr.Context) (any, error) {
		return "Hello World!", nil
	})
```

What is the proposed solution here?

aryanmehrotra:
`app.OnStart(func(a *gofr.App) error {` won't solve the usecase, as http client, db connection can only be accessed from gofr container or context (as context is uniform across handlers and cron jobs).

We should also not be exposing container directly from the app, rather similar to how cron jobs are present we can have `OnStart` or `PreRun` method.


```        
        app := gofr.New()

	app.OnStart(func(ctx *gofr.Context) error {
		// Do your synchronous startup work here
[...]

PR #1849, opened 2025-06-15
 feat: add OnStart hook to run synchronous jobs before server starts

HeerakKashyap:
**Concise explanation:**

This PR adds a new OnStart hook feature to the GoFr framework, allowing users to register and run synchronous jobs before the server starts serving requests.
Issue addressed:
Addresses issue #1833  Support to run a synchronous Job on app start.
Motivation & benefits:
Enables initialization tasks (e.g., in-memory cache, API calls, pre-loading data) to be performed before the app begins handling traffic, improving reliability and flexibility for microservice developers.

**Breaking Changes :**
No breaking changes are introduced by this PR.
The new feature is fully backward compatible and opt-in.

**Additional Information:**
No new dependencies or external libraries were added.

Example usage:
image
All startup hooks are executed in order before the server starts. If any hook returns an error, the app logs the error and exits.

**Checklist:**

 I have formatted my code using goimports and golangci-lint.
 All new code is covered by unit tests.
 This PR does not decrease the overall code coverage.
 I have reviewed the code comments and documentation for clarity.


pkg/gofr/run.go
```diff
@@ -95,3 +93,13 @@ func (a *App) Run() {
 
 	wg.Wait()
 }
+
+// runStartupHooks executes all registered startup hooks before the server starts.
+func (a *App) runStartupHooks() {
+	for _, hook := range a.startupHooks {
+		if err := hook(a); err != nil {
+			a.Logger().Errorf("startup hook failed: %v", err)
+			os.Exit(1) // or handle error as you see fit
```
ccoVeille: I'm not fan of the deep exit here.

I would at least expect to return the error here, and log in app.Run, and exit there.

I would prefer if App.Run wasn't using an exit, but I feel like it's already a problem in Run signature

HeerakKashyap: @ccoVeille I've updated the implementation so that runStartupHooks now returns an error, and the logging and os.Exit(1) are handled in Run as you suggested , This should make the error handling cleaner and avoid deep exits in helper methods. Let me know if i got anaything wrong..

vikash:
Why not just use the hook logic between gofr.New() and app.Run(). Having a method app.onStart is not really adding any value here.

pkg/gofr/run.go
coolwednesday: Can we do an optional functional parameter in `app.Run()` that takes a function similar to how we pass in `AddCronJob` method on app ? Just a suggestion. Would like to know other people thoughts ....

HeerakKashyap:
@vikash Sir , i believe .OnStart is just a way for users to say, “run this function before the server starts". it’s similar to how many frameworks allow you to register middleware, shutdown hooks, etc.

vikash:
@HeerakKashyap - Please check the issue discussion. It's important for the usecase to have context available. I have raised a PR (using chatGPT codex) here: https://github.com/gofr-dev/gofr/pull/1854  This is in no way complete and has linting and other issues. But, can give you directional ideas. I will close my PR as you started working on it - however, going forward it might be better if you do not pick the issues with 'triage' label.

coolwednesday:
@HeerakKashyap, Since this PR has been replaced by a new PR to address the same issue. Can you close this PR and reference it probably in the new one for added context.

Thanks !

PR #1853, opened 2025-06-16
Refactor: Make `Run` Return Error

HeerakKashyap:
**What does this PR do?**
Refactors the Run method to return an error instead of handling errors internally.
Updates all usages and tests to handle the new error return value.
Improves error handling and makes the codebase more idiomatic and testable.

**Why is this needed?**
Returning errors from Run allows callers to handle failures more flexibly.
Aligns with Go best practices for error handling.
Makes it easier to write tests and debug issues.

**What changed?**
Run method signature now returns an error.
All invocations of Run updated to check and handle the error.
Related tests updated to expect and assert on errors.

*Additional notes**
No breaking changes to public APIs except for the Run method signature.
Please review the test changes to ensure all edge cases are covered.

This PR is an updated version of my earlier pull request:
#1849 – feat: add OnStart hook to run synchronous jobs before server starts

I was unable to make further changes or updates to the previous PR due to branch or workflow issues. All feedback and improvements from the previous PR have been incorporated here.
#1833


Umang01-hash:
@HeerakKashyap Thanks for the PR, but I have concerns about this change.

**GoFr** is designed to handle **server startup failures, panic recovery, and graceful shutdown automatically** so developers don't need to write boilerplate error handling code. Making Run() return an error shifts this responsibility back to the user, which goes against our core philosophy of providing framework that handles operational concerns.

**This would also be a breaking change for existing applications.**

Could you help me understand what specific problem this is trying to solve? 
Perhaps we can find an alternative approach that maintains GoFr's developer-friendly design.

HeerakKashyap:
@Umang01-hash 
Thank you for your detailed feedback, @Umang01-hash!
I’m new to the GoFr community and participating in GoFr Summer of Code, so I really appreciate your guidance and the opportunity to learn.
My intention with this change was to address the issue of making error handling during server startup more explicit and testable, as I noticed some related discussions/issues. I thought returning an error from Run() would give users more control and make it easier to handle startup failures in their own code, especially for testing and automation scenarios.
However, I understand your point about GoFr’s philosophy of handling operational concerns internally to keep things simple for developers. I definitely don’t want to introduce breaking changes or go against the framework’s core design principles.
If there’s a more GoFr-friendly way to achieve better error visibility or testability for startup hooks, I’d love to hear your suggestions or work on an alternative approach!
Thank you again for your feedback and for helping me learn the best practices for contributing to GoFr.

Umang01-hash:
@HeerakKashyap Thankyou for understanding. As of now i don't see any usecase of returning error from `app.Run()` method. We can close this PR now and see if that's required in future scope we can open an issue/discussion regarding that.

PR #1867, opened 2025-06-17
Oauth improvements

akshat-kumar-singhal:
### Description:
**Change 1**: Added function NewOAuthConfig()
_Background_:
In case the tokenURL is not provided in the OAuthConfig, an error is encountered while trying to fetch the token. The error returned is from the url package itself and does not provide much guidance to the user on what needs to be fixed. Error message: Post "" unsupported protocol scheme ""
_Fix Made_:
A new function NewOAuthConfig() has been added which is responsible for the validation of OAuth config. Users are required to use this method instead of directly initialising the OAuthConfig struct.

**Change 2**: AuthStyle Made Configurable
Previously AuthStyle was hardcoded to `oauth2.AuthStyleInHeader` which has now been made configurable. In case no value is passed, it defaults to [oauth2.AuthStyleAutoDetect](https://pkg.go.dev/golang.org/x/oauth2@v0.30.0#AuthStyleAutoDetect).

### Breaking Changes (if applicable):
None

### Additional Information:
We should consider converting the OAuthConfig struct to un-exported to enforce the usage of NewOAuthConfig() function

### Checklist:


pkg/gofr/service/oauth_test.go
```diff
@@ -1,25 +1,42 @@
...
 )
 
+// #nosec G101
+const clientID = "0iyeGcLYWudLGqZfD6HvOdZHZ5TlciAJ"
+
+// #nosec G101
+const clientSecret = "GQXTY2f9186nUS3C9WWi7eJz8-iVEsxq7lKxdjfhOJbsEPPtEszL3AxFn8k_NAER"
+
+// #nosec G101
+const tokenURL = "https://dev-zq6tvaxf3v7p0g7j.us.auth0.com/oauth/token"
```
vikash: Makes the test indeterministic. Token URL may not always work. Also, putting secret (even in test) promotes bad practice. We should figure out a better way to test. It was present before as well - but, since we are refactoring it. :)

akshat-kumar-singhal: This is has been addressed by creating a mock server which issues as well as validates tokens.

Umang01-hash (approved):

pkg/gofr/service/errors.go
```diff
@@ -0,0 +1,20 @@
+package service
+
+import "github.com/pkg/errors"
```
ccoVeille: This sounds like a bad idea to use something else than the std error

pkg/gofr/service/errors.go
```diff
@@ -0,0 +1,20 @@
...
+}
+
+func (o OAuthErr) Error() string {
+	if o.Message == "" && o.Err == nil {
+		return "unknown error"
+	} else if o.Message == "" {
+		return o.Err.Error()
+	} else if o.Err == nil {
+		return o.Message
+	}
```
ccoVeille: This should be preferred

```suggestion
	if o.Message == "" && o.Err == nil {
		return "unknown error"
	}
	if o.Message == "" {
		return o.Err.Error()
	}

	if o.Err == nil {
		return o.Message
	}
```

ccoVeille: The logic is strange, try this

```suggestion
	if o.Message != "" {
		return o.Message
	}
	if o.Err != nil {
		return o.Err.Error()
	}
	return "unknown error"
```

ccoVeille: But consider this instead

https://github.com/gofr-dev/gofr/pull/1867#discussion_r2156249499

pkg/gofr/service/oauth.go
```diff
@@ -28,6 +31,58 @@ type OAuthConfig struct {
 
 	// EndpointParams specifies additional parameters for requests to the token endpoint.
 	EndpointParams url.Values
+
+	// AuthStyle represents how requests for tokens are authenticated to the server
+	// Defaults to oauth2.AuthStyleAutoDetect
```
ccoVeille: godoc is your friend

```suggestion
	// Defaults to [oauth2.AuthStyleAutoDetect]
```

pkg/gofr/service/oauth_test.go
```diff
@@ -1,209 +1,272 @@
...
-func oAuthHTTPServer(t *testing.T) *httptest.Server {
-	t.Helper()
-
-	// Start a test HTTP server
-	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		header := r.Header.Get("Authorization")
-		token := strings.Split(header, " ")
+var err1 = errors.New("unsupported protocol scheme \"\"")
+
+var err2 = errors.New("unsupported protocol scheme \"abc\"")
```
ccoVeille: Use string literals for clarity


```suggestion
var err1 = errors.New(`unsupported protocol scheme ""`)

var err2 = errors.New(`unsupported protocol scheme "abc"`)
```

pkg/gofr/service/oauth.go
```diff
@@ -28,6 +31,58 @@ type OAuthConfig struct {
...
+func validateTokenURL(tokenURL string) error {
+	if tokenURL == "" {
+		return OAuthErr{nil, "token url is mandatory"}
+	}
+
+	u, err2 := url.Parse(tokenURL)
+
+	switch {
+	case err2 != nil:
+		return OAuthErr{err2, "error in token URL"}
```
ccoVeille: There is no reason to use err2 as a variable name here

```suggestion
	u, err := url.Parse(tokenURL)

	switch {
	case err != nil:
		return OAuthErr{err, "error in token URL"}
```

pkg/gofr/service/errors.go
akshat-kumar-singhal: How do we handle the scenario where both Message and Err are non empty?

pkg/gofr/service/errors.go
akshat-kumar-singhal: need this for `errors.Wrap`, similar to datasource.ErrorDB

pkg/gofr/service/errors.go
```diff
@@ -0,0 +1,21 @@
...
+func (o OAuthErr) Error() string {
+	switch {
+	case o.Message == "" && o.Err == nil:
+		return "unknown error"
+	case o.Message == "":
+		return o.Err.Error()
+	case o.Err == nil:
+		return o.Message
+	default:
+		return errors.Wrap(o.Err, o.Message).Error()
```
ccoVeille: Use something simple instead of wrapping to convert back to a string

```suggestion
		return fmt.Printf("%s: %s", o.Err, o.Message)
```

pkg/gofr/service/errors_test.go
```diff
@@ -0,0 +1,28 @@
+package service
+
+import (
+	"testing"
+
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+)
```
ccoVeille:
```suggestion
import (
	"errors"
	"testing"

	"github.com/stretchr/testify/assert"
)
```

pkg/gofr/service/errors.go
ccoVeille: See this

https://github.com/gofr-dev/gofr/pull/1867#discussion_r2156826262

akshat-kumar-singhal: are we moving away from the convention used in datasource.ErrorDB? If yes, will make the change here as well as datasource.ErrorDB.Error()

ccoVeille: There is one slight difference

Yes, [datasource/errors.go](https://github.com/gofr-dev/gofr/blob/development/pkg%2Fgofr%2Fdatasource%2Ferrors.go) is doing something strange.

But this file also uses a method [errors.WithStack](https://pkg.go.dev/github.com/pkg/errors#WithStack) that is only available in github.com/pkg/errors # package.

So you could and should change the Wrap().Error oddity, but you won't be able to remove pkg/errors package unfortunately

pkg/gofr/service/errors.go
ccoVeille: I don't know if it can happen, but it can be done with

```go
return fmt.Sprintf("%s: %s", o.Message, o.Err)
```

pkg/gofr/service/errors.go
akshat-kumar-singhal: Updated the implementation here. Left the datasource.ErrorDB as is for now - can take that up separately.

pkg/gofr/service/oauth_test.go
```diff
@@ -1,209 +1,272 @@
...
-	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		header := r.Header.Get("Authorization")
-		token := strings.Split(header, " ")
+var err1 = errors.New(`unsupported protocol scheme ""`)
+
+var err2 = errors.New(`unsupported protocol scheme "abc"`)
+
+func TestHttpService_RequestsOAuth(t *testing.T) {
+	server := setupOAuthHTTPServer(t)
+	defer server.httpServer.Close()
```
ccoVeille: Please remove them all of them, and consider using a cleanup in setupOAuthHTTPServer

See https://github.com/gofr-dev/gofr/pull/1867#discussion_r2159310342

pkg/gofr/service/mock_oauth_server.go
```diff
@@ -0,0 +1,207 @@
...
+
+		assert.Equal(t, server.audienceClaim, claims[0])
+
+		w.WriteHeader(http.StatusOK)
+	})
+
+	server.httpServer = httptest.NewServer(mux)
+
+	return &server
+}
```
ccoVeille:
```suggestion
	server.httpServer = httptest.NewServer(mux)
	t.Cleanup(func() {
		err := server.httpServer.Close()
		assert.NoError(t, err, "error while closing http server")
	}
	return &server
}
```

pkg/gofr/service/oauth_test.go
```diff
@@ -1,209 +1,272 @@
...
+	tokenURL := server.getTokenURL()
+
+	invalidURL := "abc://invalid-url"
+
+	testCases := []struct {
+		method     string
+		headers    bool
+		tokenURL   string
+		err        error
+		statusCode int
```
ccoVeille, line 35: Don't add `statusCode` to the struct, you are using the same value for all of them.

```suggestion
```

See https://github.com/gofr-dev/gofr/pull/1867#discussion_r2159312561

pkg/gofr/service/oauth_test.go
```diff
@@ -1,209 +1,272 @@
...
+		}
 
-	// Apply circuit breaker option to the HTTP service
-	httpSvc := oauthConfig.AddOption(&service)
+		assert.Equalf(t, tc.err, err, "failed test case #%d", i)
 
-	return httpSvc
-}
+		if resp != nil {
+			assert.Equalf(t, tc.statusCode, resp.StatusCode, "failed test case #%d", i)
```
ccoVeille:
```suggestion
			assert.Equalf(t, http.StatusOK, resp.StatusCode, "failed test case #%d", i)
```

pkg/gofr/service/oauth_test.go
```diff
@@ -1,209 +1,272 @@
...
-func oAuthHTTPServer(t *testing.T) *httptest.Server {
-	t.Helper()
-
-	// Start a test HTTP server
-	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		header := r.Header.Get("Authorization")
-		token := strings.Split(header, " ")
+var err1 = errors.New(`unsupported protocol scheme ""`)
+
+var err2 = errors.New(`unsupported protocol scheme "abc"`)
```
ccoVeille: These names are not useful, also it leads to confusion when reading the map used for tests

```suggestion
const invalidURL := "abc://invalid-url"

var (
	errTokenURL = &url.Error{Op: "Post", URL: "", Err: errors.New(`unsupported protocol scheme ""`)}
	errInvalidURL = &url.Error{Op: "Post", URL: invalidURL, Err: err2}
)
```

Then use `errTokenURL` and `errInvalidURL` everywhere

pkg/gofr/service/oauth_test.go
```diff
@@ -1,209 +1,272 @@
...
+		headers    bool
+		tokenURL   string
+		err        error
+		statusCode int
+	}{
+		{http.MethodGet, false, tokenURL, nil, http.StatusOK},
+		{http.MethodPost, false, tokenURL, nil, http.StatusOK},
+		{http.MethodPut, false, tokenURL, nil, http.StatusOK},
+		{http.MethodDelete, false, tokenURL, nil, http.StatusOK},
+		{http.MethodPatch, false, tokenURL, nil, http.StatusOK},
```
ccoVeille: a struct should be used with the fields name, not like this

pkg/gofr/service/oauth_test.go
```diff
@@ -1,209 +1,272 @@
...
+func callHTTPServiceGet(ctx context.Context, service HTTP, headers bool) (resp *http.Response, err error) {
+	if headers {
+		resp, err = service.GetWithHeaders(ctx, "test", nil, nil)
+	} else {
+		resp, err = service.Get(ctx, "test", nil)
+	}
 
-	_ = resp.Body.Close()
+	return resp, err
 }
```
ccoVeille: Please consider doing something like this

```go
func callHTTPService(t *testing.T, ctx context.Context, service HTTP, method string, headers bool) (resp *http.Response, err error) {
	t.Helper()

	if headers {
		resp, err = callHTTPServiceWithHeaders(t, ctx, service, method)
	} else {
		resp, err = callHTTPServiceWithoutHeaders(t, ctx, service, method)
	}

	if resp != nil {
		t.Cleanup(func() {
			require.NoError(t, resp.Body.Close())
		})
	}

	return resp, err
}

func callHTTPServiceWithHeaders(t *testing.T, ctx context.Context, service HTTP, method string, headers bool) (resp *http.Response, err error) {
		switch method {
			case http.MethodGet:
				return resp, err = service.GetWithHeaders(ctx, "test", nil, nil)
			case http.Method:Post:
				return resp, err = service.PostWithHeaders(ctx, "test", nil, nil, nil)
			case http.MethodDelete:
				return resp, err = service.DeleteWithHeaders(ctx, "test", nil, nil, nil)
			case http.MethodGet:
				return resp, err = service.GetWithHeaders(ctx, "test", nil, nil, nil)
		}
		t.Fatalf("Unsupported method: %s", method)
}

func callHTTPServiceWithoutHeaders(t *testing.T, ctx context.Context, service HTTP, method string, headers bool) (resp *http.Response, err error) {
	switch method {
		case http.MethodGet:
			return resp, err = service.Get(ctx, "test", nil)
		case http.Method:Post:
			return resp, err = service.Post(ctx, "test", nil, nil)
		case http.MethodDelete:
			return resp, err = service.Delete(ctx, "test", nil, nil)
		case http.MethodPatch:
			return resp, err = service.Patch(ctx, "test", nil, nil)
	}

	t.Fatalf("Unsupported method: %s", method)
}
```

This way, you don't have to play with a switch everywhere

pkg/gofr/service/oauth_test.go
```diff
@@ -1,209 +1,272 @@
...
+
+		if err == nil {
+			authHeader, ok := headers[AuthHeader]
+			assert.True(t, ok, "failed test case #%d", i)
+			assert.NotEmptyf(t, authHeader, "failed test case #%d", i)
+			assert.True(t, strings.HasPrefix(authHeader, "Bearer"))
+			delete(headers, AuthHeader)
+			assert.Equal(t, tc.response, headers, "failed test case #%d", i)
+		}
 	}
```
ccoVeille: the way to do it to use a t.Run here

```suggestion
	for i, tc := range testCases {
		t.Run(fmt.Sprintf("test case #%d", d), func(t *testing) {
			service, ok := getOAuthService(server.httpService(), server.clientID, server.clientSecret, tc.tokenURL, server.audienceClaim).(*oAuth)
			assert.True(t, ok, "unable to get oAuth object")
	
			headers, err := service.addAuthorizationHeader(tc.ctx, tc.headers)
			assert.Equal(t, tc.err, err)
	
			if err == nil {
				authHeader, ok := headers[AuthHeader]
				assert.True(t, ok)
				assert.NotEmptyf(t, authHeader)
				assert.True(t, strings.HasPrefix(authHeader, "Bearer"))
				delete(headers, AuthHeader)
				assert.Equal(t, tc.response, headers)
			}
		})
	}
```

pkg/gofr/service/oauth_test.go
```diff
@@ -1,209 +1,272 @@
...
+		{clientID, "", "", nil, nil, 0, OAuthErr{nil, "client secret is mandatory"}},
+		{clientID, "", tokenURL, nil, nil, 0, OAuthErr{nil, "client secret is mandatory"}},
+		{clientID, clientSecret, "", nil, nil, 0, OAuthErr{nil, "token url is mandatory"}},
+		{clientID, clientSecret, "invalid_url_format", nil, nil, 0, OAuthErr{nil, "empty host"}},
+		{clientID, clientSecret, tokenURL, nil, nil, 0, nil},
+		{clientID, "some_random_client_secret", tokenURL, nil, nil, 0, nil},
+		{"some_random_client_id", clientSecret, tokenURL, nil, nil, 0, nil},
+		{clientID, clientSecret, tokenURL, nil, nil, 1, nil},
+		{clientID, "some_random_client_secret", tokenURL, nil, nil, 1, nil},
+		{"some_random_client_id", clientSecret, tokenURL, nil, nil, 2, nil},
```
ccoVeille: Here is why you shouldn't use a struct without field names, you are providing a lot of fields with their zero values

pkg/gofr/service/oauth_test.go
ccoVeille: See https://github.com/gofr-dev/gofr/pull/1867#discussion_r2159311980

pkg/gofr/service/oauth_test.go
akshat-kumar-singhal: done

pkg/gofr/service/oauth_test.go
akshat-kumar-singhal: Had missed adding test cases for status code, added them now

pkg/gofr/service/oauth_test.go
akshat-kumar-singhal: updated

pkg/gofr/service/oauth_test.go
akshat-kumar-singhal: updated

pkg/gofr/service/oauth_test.go
akshat-kumar-singhal: updated

pkg/gofr/service/oauth_test.go
akshat-kumar-singhal: updated

pkg/gofr/service/oauth_test.go
akshat-kumar-singhal: added test cases for statusCode hence retaining

pkg/gofr/service/oauth_test.go
akshat-kumar-singhal: Updated

pkg/gofr/service/errors.go
akshat-kumar-singhal: We'd need to have the given switch case to incorporate this - the recommended if else block won't work. 

> I don't know if it can happen, but it can be done with
> 
> ```go
> return fmt.Sprintf("%s: %s", o.Message, o.Err)
> ```

pkg/gofr/service/mock_oauth_server.go
```diff
@@ -0,0 +1,211 @@
...
+	return s.httpServer.URL + s.tokenURL
+}
+
+func getClaims(r *http.Request) map[string]any {
+	claims := make(map[string]any, 0)
+
+	for key, value := range r.Form {
+		if key == "client_id" || key == "client_secret" || key == "grant_type" {
+			continue
+		}
```
ccoVeille, line 165: You can use a switch here, the code would be clearer

ccoVeille (approved):
Thanks for applying the changes.

A minor remark, but 👍

coolwednesday (approved):

Umang01-hash (dismissed):

aryanmehrotra (approved):

aryanmehrotra (changes requested):
Documentation needs to be update in this file: https://github.com/gofr-dev/gofr/blob/development/docs/advanced-guide/http-communication/page.md

aryanmehrotra (approved):

PR #1901, opened 2025-06-26
feat: Azure Blob Storage Integration Module

PAVAN2627:
## Summary
This PR adds support for Azure Blob Storage to the GoFr framework as part of the unified file interface. The module is implemented under `pkg/datasource/file/azure` and aligns with the existing design used for AWS S3 and SFTP.

## Features
- ✅ Upload, download, delete, and list file operations
- ✅ Uses Azure Blob SDK for Go
- ✅ Implements the `file.File` interface for compatibility
- ✅ Clean, idiomatic Go with error handling and modular structure
- ✅ Plug-and-play: no tight coupling with GoFr core

## Folder Structure
- `pkg/datasource/file/azure/` — main Azure module
- `azure.go` — implementation of file operations
- `azure_test.go` — unit tests for each method

## Usage Example
```go
import (
    "github.com/gofr-dev/gofr/pkg/datasource/file/azure"
)

azureClient := azure.NewAzureBlob(azure.Config{
    AccountName: "your-account-name",
    AccountKey:  "your-account-key",
    Container:   "your-container",
})

err := azureClient.Upload(ctx, "sample.txt", []byte("Hello from GoFr!"))

### CODE REVIEW

File: go.mod [hunk 1]
```diff
@@ -1,9 +1,10 @@
-module gofr.dev
+module github.com/PAVAN2627/gofr
```
Umang01-hash, line 1: You cannot change the path of main go.mod. It will break everything. Please revert this.

go.mod
```diff
@@ -1,9 +1,10 @@
...
 
 go 1.24
 
 require (
 	cloud.google.com/go/pubsub v1.49.0
+	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.1
```
Umang01-hash, line 7: There should be a separate `go.mod` created for this azure blob integration inside the `pkg/datsource/file/azureblobl`. If we add this dependency here then the whole motive of separating modules to keep the initial binary small become ineffective.

pkg/datasource/file/azureblob/README.md
```diff
@@ -0,0 +1,44 @@
+# Azure Blob File Module for GoFr
```
Umang01-hash, line 1: The `README.md` here is not required. We need to updated [this](https://gofr.dev/docs/advanced-guide/handling-file#usage) documentation.  Refer to the `docs` directory for that.

pkg/datasource/file/azureblob/azureblob.go
```diff
@@ -0,0 +1,65 @@
+package azureblob
+
+import (
+    "context"
+    "fmt"
+    "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
```
Umang01-hash, line 6: Imports are not sorted.

Umang01-hash (changes requested):
- You have added your `azureblob` directoy at wrong place. It doesn't need to be preset at `gofr/pkg/datasource/file` but instead [here](https://github.com/gofr-dev/gofr/tree/development/pkg/gofr/datasource/file)  at `gofr/pkg/gofr/datasource/file`.

- The godoc comments are missing from the client file `azureblob.go`

- The PR is failing the code quality issues.

- Test are showing 0% code coverage.


A lot of issues in the PR is there. Can we please mark it as **DRAFT** fix all the things and then re-raise it.

PR #1904, opened 2025-06-27
 Add delay in subscriber loop to prevent tight retry on error

bhagavan04:
Fix: Add delay in startSubscriber loop to prevent tight retry on subscription error.

Description: This PR fixes a potential performance issue in the SubscriptionManager where the startSubscriber method retries immediately after a subscription failure without delay.
current behaviour: ->Immediate retry on subscription errors.
->Leads to high CPU usage and excessive logs during repeated failures

after fix behaviour: ->Added time.Sleep(2 * time.Second) after a failed subscription attempt
->Imported Go standard time package
->Improves resilience and prevents retry storm

Future enhancement: Make retry delay configurable via container or env.

Before Change (No Delay – Tight Loop):
log output:
ERROR error in subscription for topic user-events: connection refused
ERROR error in subscription for topic user-events: connection refused
ERROR error in subscription for topic user-events: connection refused
...
(CPU usage high, log flooding rapidly)

 After Change (With Delay):
Log output
ERROR error in subscription for topic user-events: connection refused
(wait 2s)
ERROR error in subscription for topic user-events: connection refused
(wait 2s)
ERROR error in subscription for topic user-events: connection refused
...
(Much smoother, reduced load)


pkg/gofr/subscriber.go
```diff
@@ -33,6 +33,8 @@ func (s *SubscriptionManager) startSubscriber(ctx context.Context, topic string,
 			err := s.handleSubscription(ctx, topic, handler)
 			if err != nil {
 				s.container.Logger.Errorf("error in subscription for topic %s: %v", topic, err)
+				// Add delay to prevent tight retry loop
+				time.sleep(2 * time.second)
```
ccoVeille: This is not a good pattern as it's blocking to exit when context expire.

Please consider something like this


```go
func (s *SubscriptionManager) startSubscriber(ctx context.Context, topic string, handler SubscribeFunc) error {
				var delay time.Duration
        for {
                select {
                case <-ctx.Done():
                        s.container.Logger.Infof("shutting down subscriber for topic %s", topic)
                        return nil
                case <-time.After(delay):
                        err := s.handleSubscription(ctx, topic, handler)
                        if err != nil {
                                s.container.Logger.Errorf("error in subscription for topic %s: %v", topic, err)
                                delay = 2 * time.second
                        }
                }
        }
}
```

Here also something like this could be tried

delay += 2 * time.second

or

delay =  max(1.1 * delay, 2 * time.Second)

ccoVeille (changes requested):

pkg/gofr/subscriber.go
```diff
@@ -24,15 +24,27 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
 func (s *SubscriptionManager) startSubscriber(ctx context.Context, topic string, handler SubscribeFunc) error {
+	var delay time.duration = 2 * time.second
+	
 	for {
 		select {
 		case <-ctx.Done():
 			s.container.Logger.Infof("shutting down subscriber for topic %s", topic)
 			return nil
-		default:
+		case <-time.after(delay):
```
ccoVeille: I don't understand this change. When the app starts it now wait 2s ?

pkg/gofr/subscriber.go
```diff
@@ -24,15 +24,27 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
+
+                               // Exponential backoff: slow down retry after repeated failures
+				delay += 2 * time.second
+				if(delay > 30*time.second {
+				delay = 30 * time.second // capture max delay
+			       }
+				   else{
+					   // reset delay after success
+					   delay = 2 * time.second
 			}
```
ccoVeille: This is not formatted, and I feel like the else is not at the right place

Here is the code

delay += 2s
if delay > 30s
    delay = 30s
else
    delay = 2s

ccoVeille: The else sounds suspicious here. Please look at the code.

Here is what I would suggest

```suggestion
 			if err == nil {
				// reset delay after success
				delay = 2 * time.second
				continue
			}
 			s.container.Logger.Errorf("error in subscription for topic %s: %v", topic, err)

			// Exponential backoff: slow down retry after repeated failures
			delay += 2 * time.second
			delay = min(delay, 30 time.Second)

```

pkg/gofr/subscriber.go
bhagavan04: You're right -with the current change, the subscriber waits 2 seconds even on the first run, which is not ideal.
I'll fix it by keeping the first attempt immediate and only applying the delay after an error occurs.
I'll move the delay logic inside the error block, and restore default: for the initial trigger.

Umang01-hash:
Screenshot 2025-07-02 at 12 37 33 PM

@bhagavan04 Thankyou for your contribution i found your code still have some linter errors. Please resolve the same. The examples are not even working right now for testing your change.

pkg/gofr/subscriber.go
```diff
@@ -24,20 +25,30 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
...
+				continue
 			}
-		}
-	}
+ 			s.container.Logger.Errorf("error in subscription for topic %s: %v", topic, err)
+			// Exponential backoff: slow down retry after repeated failures
+			delay += 2 * time.Second
+			delay = time.Duration(math.Min(float64(delay), float64(30*time.Second)))
+			// actually wait before retrying
+			time.Sleep(delay)
```
ccoVeille: This is wrong and blocking.

Code was reverted to something that is not good.

there should be something like this

```go
func (s *SubscriptionManager) startSubscriber(ctx context.Context, topic string, handler SubscribeFunc) error {
        var delay time.Duration // zero delay by default

        for {
                select {
                case <-ctx.Done():
                        s.container.Logger.Infof("shutting down subscriber for topic %s", topic)
                        return nil
                 case <-time.After(delay):
                        err := s.handleSubscription(ctx, topic, handler)
                        if err == nil {
                                // reset delay after success
                                delay = 0
                                continue
                        }
                        s.container.Logger.Errorf("error in subscription for topic %s: %v", topic, err)
                        // Exponential backoff: slow down retry after repeated failures
                        delay += 2 * time.Second
                        if delay >= 30 * time.Second {
                           delay = 30 * time.Second
                        }
                  }
          }
}
```

I removed the min because I felt it was bringing too much complexity than a regular if

Umang01-hash:
Hey @bhagavan04 
The linter and code quality issues are still not resolved. Please resolve the pipleine failures.

bhagavan04:
Fix all syntax issues (time.Second, time.After, etc.)
Remove unused imports like math

ccoVeille (approved):

ccoVeille:
Do you mind adding a Co-authored-by tag to your commits ?

Here is mine

```
Co-authored-by: ccoVeille <3875889+ccoVeille@users.noreply.github.com>`
```

Some context here: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

Thanks

bhagavan04:
"I've added the Co-authored-by tag. Please review and approve the workflows."
Co-authored-by: ccoVeille <3875889+ccoVeille@users.noreply.github.com>`

coolwednesday:
@bhagavan04 , There are Code quality issue. Check them out and fix them.

pkg/gofr/subscriber.go
```diff
@@ -3,7 +3,9 @@ package gofr
 import (
 	"context"
 	"runtime/debug"
-
+
```
ccoVeille: This line should be removed

Umang01-hash:
Screenshot 2025-07-09 at 1 54 02 PM
@bhagavan04 Still this issue exists in Code Quality step.

Please resolve it.

Umang01-hash:
@bhagavan04 The fix was tested, and the logs show that the **delay mechanism in startSubscriber is not functioning as expected**. While the delay is applied between loop iterations, it doesn't handle concurrent subscriptions to multiple topics properly. This results in errors being logged without the intended delays.

Screenshot 2025-07-09 at 3 30 39 PM

bhagavan04:
Thanks for the feedback @Umang01-hash! You're right — the delay logic inside startSubscriber currently applies only within a single loop execution per topic, and since subscriptions are launched concurrently per topic, retries across topics can still hammer the system tightly.
 To fix this, I’m planning to move the delay mechanism into the handleSubscription function itself, so that each topic’s subscription handles its own retry pacing — even when run concurrently.
Alternatively, we can also consider wrapping each startSubscriber call with a throttling mechanism or using a retry/backoff utility like go-retryable to decouple retries from the main loop.

Will push a patch shortly — happy to hear any other ideas too!

Umang01-hash:
@bhagavan04 Any updates on the above?

coolwednesday:
@bhagavan04 , As there have been no updates fro your side, would be closing the PR for now. Feel free to reopen it after the comments above have been addressed.

PR #2079, opened 2025-07-19
feat: add OnStart hook for synchronous startup jobs 

### LINKED ISSUE #1833: Support to run a synchronous Job on app start

HeerakKashyap:
**Description:**

- Implements an `OnStart` hook for the GoFr framework, allowing users to register synchronous startup jobs that execute before the server starts.
- The hook receives a fully initialized `*gofr.Context` with all DI-managed services, as requested by maintainers.
- Removes any previous custom context logic and follows the GoFr way for dependency injection.
- Updates documentation and provides usage examples.
- Includes proof of working feature (see attached screenshots).

**Testing:**
- Verified that OnStart hooks can access DI-managed services (e.g., SQL, Redis).
- All relevant tests pass with MySQL and Redis running in Docker.

**Fixes** #1833

---

Let me know if you want to add or change anything, or if you need help attaching screenshots or filling out the checklist!


go.work.sum
ccoVeille, line 1: I'm unsure why this was changed. As if code using a different logger was added

pkg/gofr/run.go
```diff
@@ -19,6 +23,27 @@ func (a *App) Run() {
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
 
+	// --- Add this block for OnStart hooks ---
+	onStartCtx, span := otel.GetTracerProvider().Tracer("gofr-onstart").Start(context.Background(), "OnStart")
+	defer span.End()
```
ccoVeille: Why creating a context here?

Can't you use ctx?

Is that a problem of context cancelation?

If so using context.WithoutCancel should be considered

pkg/gofr/run.go
```diff
@@ -19,6 +23,27 @@ func (a *App) Run() {
...
+
+	gofrCtx := &Context{
+		Context:       onStartCtx,
+		Container:     a.container,
+		Request:       noopRequest{}, // Make sure noopRequest{} is available
+		ContextLogger: *logger,
+	}
+
+	for _, hook := range a.onStartHooks {
+		if err := hook(gofrCtx); err != nil {
```
ccoVeille: Here this hook receive a Context that comes from a context that cannot expires, it doesn't make sense to me.

See also https://github.com/gofr-dev/gofr/pull/2079#discussion_r2217398061

Why is there a need for this?

pkg/gofr/run.go
```diff
@@ -19,6 +23,27 @@ func (a *App) Run() {
...
+			a.Logger().Errorf("OnStart hook failed: %v", err)
+			os.Exit(1)
+		}
+	}
+	// --- End OnStart block ---
```
ccoVeille: I would move this to a method dedicated to onStartHooks

pkg/gofr/run.go
```diff
@@ -19,6 +23,27 @@ func (a *App) Run() {
...
+	defer span.End()
+
+	logger := logging.NewContextLogger(onStartCtx, a.container.Logger)
+
+	gofrCtx := &Context{
+		Context:       onStartCtx,
+		Container:     a.container,
+		Request:       noopRequest{}, // Make sure noopRequest{} is available
+		ContextLogger: *logger,
+	}
```
ccoVeille: Also, I feel like this code is inspired from something that is created later in the stack.

Which means the code is duplicated.

Imagine a new field added to the gofr.Context hooks won't have it.

I might be wrong but I feel like there is a problem

go.work.sum
HeerakKashyap: I was building a new docker container, it automatically changes if u run so "go build " kinds of prompts

pkg/gofr/run.go
HeerakKashyap: @ccoVeille  I created a new context for the OnStart hook to ensure it was independent, but I understand your concern about context cancellation and expiry.  I can update the implementation to use the main ctx (the one created with signal.NotifyContext) or use context.WithoutCancel(ctx) if you prefer, so the OnStart hook context is consistent with the rest of the app and respects cancellation.

pkg/gofr/run.go
HeerakKashyap: @ccoVeille 
Thank you for the feedback!
You’re right—using a context that never expires for the OnStart hook doesn’t make sense, especially if cancellation or timeouts are important for startup logic.
I can update the implementation to use the main app context (ctx) or, if cancellation is a concern, use context.WithoutCancel(ctx) as you suggested, so the OnStart hook context is consistent with the rest of the app and can be cancelled if needed.
Also, I agree that moving the OnStart hook logic to a dedicated method will improve code clarity and maintainability.

pkg/gofr/run.go
HeerakKashyap: I duplicated the context construction logic because, as umang sir mentioned in their feedback, the OnStart hook should provide access to GoFr’s DI-managed services like ctx.DB or ctx.HTTPService. My intention was to ensure the context passed to the hook would have these dependencies available, similar to request handlers and cron jobs.

ccoVeille: Let's wait for @coolwednesday @Umang01-hash feedbacks then

pkg/gofr/run.go
ccoVeille: My feedbacks were only mine

Let's wait for @coolwednesday @Umang01-hash ones

pkg/gofr/run.go
ccoVeille: I feel like you replied the wrong thread, and you wanted to reply to

https://github.com/gofr-dev/gofr/pull/2079#discussion_r2217398574

ccoVeille: > I would move this to a method dedicated to onStartHooks

I'm still curious about your point of view here

pkg/gofr/run.go
ccoVeille: I feel like a reply to this was provided here

> Thank you for the feedback!
> You’re right—using a context that never expires for the OnStart hook doesn’t make sense, especially if cancellation or timeouts are important for startup logic.
> I can update the implementation to use the main app context (ctx) or, if cancellation is a concern, use context.WithoutCancel(ctx) as you suggested, so the OnStart hook context is consistent with the rest of the app and can be cancelled if needed.

https://github.com/gofr-dev/gofr/pull/2079#discussion_r2217694345

pkg/gofr/run.go
ccoVeille: I just notice you replied, I'm sorry

> Also, I agree that moving the OnStart hook logic to a dedicated method will improve code clarity and maintainability.

pkg/gofr/run.go
HeerakKashyap: yes sure sir

pkg/gofr/run.go
HeerakKashyap: > I just notice you replied, I'm sorry

not a problem sir, i am also looking to provide best possible approach

pkg/gofr/run.go
Umang01-hash: @HeerakKashyap Using the **main ctx (created with signal.NotifyContext)** ensures consistency and proper lifecycle management, as all components share the same context. I recommend proceeding with the main ctx for now to maintain alignment with the framework's design and ensure proper cancellation handling.

pkg/gofr/run.go
Umang01-hash: @HeerakKashyap The intent to provide DI-managed services like ctx.DB or ctx.HTTPService in the OnStart hook is valid. However, duplicating the context construction logic here introduces maintainability issues. If a new field is added to `gofr.Context`, it would require updates in multiple places, increasing the risk of inconsistencies.

Instead, consider reusing the existing `newContext` method.

pkg/gofr/run.go
Umang01-hash: @HeerakKashyap the suggestion from @ccoVeille is right, letsmove this to a method dedicated to **onStartHooks**

examples/http-server/main.go
```diff
@@ -28,6 +28,14 @@ func main() {
 	a.GET("/trace", TraceHandler)
 	a.GET("/mysql", MysqlHandler)
 
+	// Register an OnStart Hook
+	a.OnStart(func(ctx *gofr.Context) error {
+		// Try accessing a DI-managed service, e.g., DB or HTTPService
```
Umang01-hash: The example is too basic and doesn't showcase a real-world use case. Consider demonstrating a meaningful initialization task, like seeding a database or warming up a cache, to better illustrate the OnStart hook's practical value.

And if not atleast we will use our **logger** instead of `fmt.Println` here.

pkg/gofr/run.go
HeerakKashyap: ok @Umang01-hash i am fixing it , i'll update this branch soon

HeerakKashyap:
> * There are no tests added for the newly added code.
> * No documentation updates for the new feature.
> * We have following code quality/linter issues in your PR:
> 
> Screenshot 2025-07-21 at 4 33 49 PM
> Please resolve them along with other review comments. Thankyou

should i update iin readme.md ,  it would be okay to do so right ? @Umang01-hash

examples/http-server/main.go
HeerakKashyap: ok i'll look for this to add a realistic use case

Umang01-hash:
> > * There are no tests added for the newly added code.
> > * No documentation updates for the new feature.
> > * We have following code quality/linter issues in your PR:
> > 
> > Screenshot 2025-07-21 at 4 33 49 PM
> > Please resolve them along with other review comments. Thankyou
> 
> should i update iin readme.md , it would be okay to do so right ? @Umang01-hash

@HeerakKashyap I think we can update the existing docs or maybe create a new one for the same. Both can work depending upon the content we are sharing.

HeerakKashyap:
ss for testing
Test case are working fine @Umang01-hash @ccoVeille

pkg/gofr/run.go
```diff
@@ -15,10 +15,15 @@ func (a *App) Run() {
...
 
-	// Create a context that is canceled on receiving termination signals
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
 
+	// Running startup hooks and exit if they fail.
+	// Using the main app context to ensure proper lifecycle management.
+	if err := a.runOnStartHooks(ctx); err != nil {
+		os.Exit(1)
+	}
```
ccoVeille: os.Exit is dangerous

This would be a problem if context is canceled, as the server shutdown won't be triggered


```suggestion
	// Running startup hooks and exit if they fail.
	// Using the main app context to ensure proper lifecycle management.
	if err := a.runOnStartHooks(ctx); err != nil {
		if !errors.Is(ctx, context.Canceled) {
			os.Exit(1)
		}
	}
```

This is pseudo code, it should be improved, but I hope it helps to understand the issue

HeerakKashyap:
kidnly have a look @ccoVeille @Umang01-hash

pkg/gofr/run.go
```diff
@@ -15,10 +16,22 @@ func (a *App) Run() {
...
+	if err := a.runOnStartHooks(ctx); err != nil {
+		// If the error is due to context cancellation, allow graceful shutdown.
+		if errors.Is(err, context.Canceled) {
+			a.Logger().Info("Startup canceled by context, shutting down gracefully.")
+			return
+		}
+		// For any other error, log and exit.
+		a.Logger().Errorf("Startup failed: %v", err)
+		os.Exit(1)
+	}
```
ccoVeille: Read my comment and the comment you added, the look at the code that handles the graceful shutdown.

It should be like this

```suggestion
	// Run startup hooks and exit if they fail for reasons other than context cancellation.
	// Using the main app context to ensure proper lifecycle management.
	if err := a.runOnStartHooks(ctx); err != nil {
		// If the error is due to context cancellation, allow graceful shutdown.
		if !errors.Is(err, context.Canceled) {
			// For any other error, log and exit.
			a.Logger().Errorf("Startup failed: %v", err)
			os.Exit(1)
		}

		a.Logger().Info("Startup canceled by context, shutting down gracefully.")
	}
```

ccoVeille:
The code is in, but I would prefer if the context was not forged twice.

Can't we create the gofr.Context before calling startup hooks, then pass it through all the layers as argument up to the place where the gofr.Context is created now?

Or would this implies breaking changes,? Or simply too much changes?

Please note, it doesn't necessary have to be addressed in this PR.

pkg/gofr/run.go
ccoVeille: This still needs to be addressed

HeerakKashyap:
> The code is in, but I would prefer if the context was not forged twice.
> 
> Can't we create the gofr.Context before calling startup hooks, then pass it through all the layers as argument up to the place where the gofr.Context is created now?
> 
> Or would this implies breaking changes,? Or simply too much changes?
> 
> Please note, it doesn't necessary have to be addressed in this PR.

Thank you for the suggestion Sir! @ccoVeille 

 I agree that creating the gofr.Context only once and passing it through would be cleaner and more efficient.
 However, implementing this would require significant refactoring and could potentially introduce breaking changes, especially if other parts of the framework expect to construct their own context.

 For this PR, I’ll keep the current approach to avoid introducing breaking changes, but I think this is a great idea for future improvement and can be considered in a dedicated refactor.

HeerakKashyap:
@Umang01-hash sir, kindly run the workflows and let me know is there now any linter errors etc.

go.work.sum
Umang01-hash: @HeerakKashyap Please reomve these changes. There should be no changes to `go.work.sum` file.

pkg/gofr/gofr.go
```diff
@@ -46,6 +46,30 @@ type App struct {
...
+	return &Context{
+		Context:       ctx,
+		Container:     a.container,
+		Request:       noopRequest{},
+		ContextLogger: *logger,
+	}
+}
+
+func (a *App) runOnStartHooks(ctx context.Context) error {
+	gofrCtx := a.newContextForHooks(ctx)
```
Umang01-hash: Consider reusing the existing `newContext` or `newCMDContext` functions instead of creating a new `newContextForHooks` method. This would improve consistency and reduce code duplication. You could adapt one of the existing methods to handle the hook case by providing appropriate parameters.

pkg/gofr/gofr.go
```diff
@@ -46,6 +46,30 @@ type App struct {
 	httpRegistered bool
 
 	subscriptionManager SubscriptionManager
+	onStartHooks        []func(ctx *Context) error
+}
+
+func (a *App) newContextForHooks(ctx context.Context) *Context {
```
Umang01-hash: We don't really need this `newContextForHooks`. Please remove it.

Umang01-hash:
Hey @HeerakKashyap only a few things:
- Changes in `go.work.sum` file needs to be reverted.
- We dont need a `newContextForHooks` methods, we can use exitisting methods to create new context.
- A lot of linter errors are there in the code we need to resolve them.

Please do the above changes and we should be ready to go with your PR.

HeerakKashyap:
@Umang01-hash kidnly , check the changes out

Umang01-hash:
Screenshot 2025-07-24 at @HeerakKashyap 0 53 08 AM
Hey @HeerakKashyap. Please resolve the linter issues on your PR.

HeerakKashyap:
> Screenshot 2025-07-24 at @HeerakKashyap 0 53 08 AM Hey @HeerakKashyap. Please resolve the linter issues on your PR.

i had to make couple of changes to fix the complexity issue , kindly see if this works @Umang01-hash

go.work.sum
```diff
@@ -865,6 +865,7 @@ github.com/go-logfmt/logfmt v0.5.0 h1:TrB8swr/68K7m9CcGut2g3UOihhbcbiMAYiuTXdEih
 github.com/go-logfmt/logfmt v0.5.1 h1:otpy5pqBCBZ1ng9RQ0dPu4PN7ba75Y/aA+UpowDyNVA=
 github.com/go-logfmt/logfmt v0.5.1/go.mod h1:WYhtIu8zTZfxdn5+rREduYbwxfcBr/Vr6KEVveWlfTs=
 github.com/go-logr/logr v1.4.1/go.mod h1:9T104GzyrTigFIr8wt5mBrctHMim0Nb2HLGrmQ40KvY=
+github.com/go-logr/logr v1.4.2/go.mod h1:9T104GzyrTigFIr8wt5mBrctHMim0Nb2HLGrmQ40KvY=
```
Umang01-hash: Please revert this change back.

pkg/gofr/run.go
```diff
@@ -15,16 +17,42 @@ func (a *App) Run() {
 		a.cmd.Run(a.container)
 	}
 
-	// Create a context that is canceled on receiving termination signals
```
Umang01-hash: Why you removed this comment ?

pkg/gofr/run.go
```diff
@@ -15,16 +17,42 @@ func (a *App) Run() {
...
+	if !a.handleStartupHooks(ctx) {
+		return
+	}
+
 	timeout, err := getShutdownTimeoutFromConfig(a.Config)
 	if err != nil {
 		a.Logger().Errorf("error parsing value of shutdown timeout from config: %v. Setting default timeout of 30 sec.", err)
 	}
 
-	// Goroutine to handle shutdown when context is canceled
```
Umang01-hash, line 27: If removed here i think this comment can be added inside `startShutdownHandler`

HeerakKashyap:
@Umang01-hash So this PR is ready to get merged ?

Umang01-hash (changes requested):

pkg/gofr/run.go
HeerakKashyap: @Umang01-hash resolved , I was just thinking of following different approach.

HeerakKashyap:
@Umang01-hash CAN U LET me know how can i run linter test locally, if there's some application in Golang for running linter tests

Umang01-hash:
> @Umang01-hash CAN U LET me know how can i run linter test locally, if there's some application in Golang for running linter tests

Hey @HeerakKashyap! You can follow this command to install linter : 
```cmd
go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1 
```

then export it's path and you can run the linters locally using `golangci-lint run` command inside the GoFr directory.

HeerakKashyap:
> > @Umang01-hash CAN U LET me know how can i run linter test locally, if there's some application in Golang for running linter tests
> 
> Hey @HeerakKashyap! You can follow this command to install linter :
> 
> ```batchfile
> go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1 
> ```
> 
> then export it's path and you can run the linters locally using `golangci-lint run` command inside the GoFr directory.
 
okay Sir, let me do this then i'll let u know
Thanks!

HeerakKashyap:
@Umang01-hash Code quality tests are passed, what's next ?

HeerakKashyap:
> > > @Umang01-hash CAN U LET me know how can i run linter test locally, if there's some application in Golang for running linter tests
> > 
> > 
> > Hey @HeerakKashyap! You can follow this command to install linter :
> > ```batchfile
> > go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > then export it's path and you can run the linters locally using `golangci-lint run` command inside the GoFr directory.
> 
all linter issues are solved, I tested it locally
@Umang01-hash

HeerakKashyap:
@Umang01-hash . KINDLY REVIEW

docs/advanced-guide/startup-hooks.md
```diff
@@ -0,0 +1,55 @@
+# Startup Hooks
```
Umang01-hash: This needs to be inside a directory `startup-hooks` see all the other directories inside `docs`.

pkg/gofr/gofr.go
```diff
@@ -296,3 +311,7 @@ func (a *App) AddStaticFiles(endpoint, filePath string) {
 
 	a.httpServer.staticFiles[filePath] = endpoint
 }
+
+func (a *App) OnStart(hook func(ctx *Context) error) {
```
Umang01-hash, line 341: Can we please add a good **godoc comment** above this method `OnStart` so users can easily know it's purpose, usage etc.?

Umang01-hash (changes requested):
- In order to include your documentation for a new page you also need to make changes inside `navigation.js` inside **docs** directory.
- Please remove the changes from `go.work.sum` as they are not related to your PR.

examples/http-server/main.go
```diff
@@ -28,6 +28,26 @@ func main() {
...
+		err := ctx.Redis.Set(ctx, cacheKey, cacheValue, 0).Err()
+		if err != nil {
+			ctx.Container.Logger.Errorf("Failed to warm up cache: %v", err)
+			return err // Return the error to halt startup if caching fails.
+		}
+
+		ctx.Container.Logger.Info("Cache warmed up successfully!")
+
+		return nil
+	})
```
aryanmehrotra: we should be doing this before the registration of routes, as that is how it is intended to be used

pkg/gofr/run.go
```diff
@@ -43,15 +73,29 @@ func (a *App) Run() {
...
 		go a.sendTelemetry(http.DefaultClient, true)
 	}
+}
 
+// startAllServers starts all registered servers concurrently.
+func (a *App) startAllServers(ctx context.Context) {
 	wg := sync.WaitGroup{}
 
-	// Start Metrics Server
-	// running metrics server before HTTP and gRPC
```
aryanmehrotra: lets not remove the comments

aryanmehrotra (changes requested):

pkg/gofr/run.go
```diff
@@ -19,11 +21,39 @@ func (a *App) Run() {
...
+	if err := a.runOnStartHooks(ctx); err != nil {
+		if !errors.Is(err, context.Canceled) {
+			a.Logger().Errorf("Startup failed: %v", err)
+
+			return false
+		}
+		// If the error is context.Canceled, do not exit; allow graceful shutdown.
+		a.Logger().Info("Startup canceled by context, shutting down gracefully.")
+
+		return false
```
aryanmehrotra, line 49: In runOnStartHooks we are not using this context, then how are we making sure that if the startup us cancelled how the operation initiated will be stopped?

examples/http-server/main.go
```diff
@@ -21,6 +21,27 @@ func main() {
...
+
+		err := ctx.Redis.Set(ctx, cacheKey, cacheValue, 0).Err()
+		if err != nil {
+			ctx.Container.Logger.Errorf("Failed to warm up cache: %v", err)
+			return err // Return the error to halt startup if caching fails.
+		}
+
+		ctx.Container.Logger.Info("Cache warmed up successfully!")
+
+		return nil
```
coolwednesday: Container is not exposed through ctx that is `gofr.Context`. Please anything that you write in docs make sure you first actually execute it. This will prevent any such errors related to syntax in the documentation.

go.work.sum
coolwednesday: @HeerakKashyap , No go.work changes please. Please revert them.

pkg/gofr/gofr.go
```diff
@@ -46,6 +46,21 @@ type App struct {
...
+
+	for _, hook := range a.onStartHooks {
+		if err := hook(gofrCtx); err != nil {
+			a.Logger().Errorf("OnStart hook failed: %v", err)
+			return err
+		}
+	}
+
+	return nil
 }
```
coolwednesday: Why do we need to add `context.Context` when it is not being used ? we can take no parameter.

pkg/gofr/run.go
coolwednesday: @HeerakKashyap, the context used is GoFr's context. You can instead make a wrapper function to retrieve the context present inside the gofr context value if we need to. 

Can you let me know what is the use case for context.Cancellation in case for graceful shutdown and simply returning false in the other case ?

coolwednesday (changes requested):

pkg/gofr/run.go
HeerakKashyap: > In runOnStartHooks we are not using this context, then how are we making sure that if the startup us cancelled how the operation initiated will be stopped?

@aryanmehrotra SIr, You raised a valid point about context propagation. Currently, the `runOnStartHooks` function receives a context but doesn't pass it to the individual hooks. This means if startup is cancelled, long-running operations in hooks won't be notified.

We should modify the hook execution to pass the context through. Here's what I propose:

```go
func (a *App) runOnStartHooks(ctx context.Context) error {
    gofrCtx := newContext(ctx, noopRequest{}, a.container) // Pass ctx here
    
    for _, hook := range a.onStartHooks {
        if err := hook(gofrCtx); err != nil {
            a.Logger().Errorf("OnStart hook failed: %v", err)
            return err
        }
        
        // Check if context was cancelled
        if ctx.Err() != nil {
            return ctx.Err()
        }
    }
    return nil
}
```

This way, hooks can check `ctx.Err()` or use `ctx.Done()` to handle cancellation gracefully.

pkg/gofr/gofr.go
HeerakKashyap: > Why do we need to add `context.Context` when it is not being used ? we can take no parameter.

@coolwednesday The current implementation doesn't use the context parameter. However, based on @aryanmehrotra's feedback, we should actually use it for proper cancellation handling.

The context parameter serves two purposes:
1. Cancellation propagation: Allows hooks to respond to startup cancellation
2. Timeout handling: Could be used to set timeouts for long-running startup operations

If we remove the context parameter entirely, we lose the ability to gracefully handle startup cancellation. Instead, I suggest we keep it and actually use it as shown in my response to @aryanmehrotra above.

HeerakKashyap:
@coolwednesday The distinction between `context.Canceled` and other errors is :

 context.Canceled (graceful shutdown):
- User initiated shutdown (Ctrl+C, SIGTERM)
- Application should exit cleanly without error messages
- No need to log as "Startup failed" since it's intentional

  Other errors (startup failure):
- Actual problems during startup (DB connection failed, config errors, etc.)
- Should log as "Startup failed" for debugging
- Application should exit with error status

**Example scenarios:**
- `context.Canceled`: User hits Ctrl+C during startup → graceful exit
- Other error: Redis connection fails → log error and exit with failure status

I believe this allows users to distinguish between intentional shutdowns and actual startup problems.

coolwednesday:
@HeerakKashyap , Please address all the review comments, so that we can close this ASAP.

HeerakKashyap:
> @HeerakKashyap , Please address all the review comments, so that we can close this ASAP.

I've given replies above for the same , so should i do all the changes given by the maintainers ?

coolwednesday:
> @coolwednesday The distinction between `context.Canceled` and other errors is :
> 
> context.Canceled (graceful shutdown):
> 
> * User initiated shutdown (Ctrl+C, SIGTERM)
> * Application should exit cleanly without error messages
> * No need to log as "Startup failed" since it's intentional
>   Other errors (startup failure):
> * Actual problems during startup (DB connection failed, config errors, etc.)
> * Should log as "Startup failed" for debugging
> * Application should exit with error status
> 
> **Example scenarios:**
> 
> * `context.Canceled`: User hits Ctrl+C during startup → graceful exit
> * Other error: Redis connection fails → log error and exit with failure status
> 
> I believe this allows users to distinguish between intentional shutdowns and actual startup problems.

If we are thinking about graceful exit. It should happen in all cases. That is all the DB/Stream Connections started should be closed with graceful exit. Keep Context.Canceled only.

coolwednesday:
> > @HeerakKashyap , Please address all the review comments, so that we can close this ASAP.
> 
> I've given replies above for the same , so should i do all the changes given by the maintainers ?

Yes Please address them.

Umang01-hash (approved):

go.work.sum
ccoVeille: This is still expected to be reverted I think

What to you think @coolwednesday ?

pkg/gofr/gofr_test.go
```diff
@@ -1162,3 +1163,38 @@ func TestApp_Subscribe(t *testing.T) {
...
+	t.Run("error", func(t *testing.T) {
+		app := New()
+
+		app.OnStart(func(_ *Context) error {
+			return errHookFailed
+		})
+
+		err := app.runOnStartHooks(t.Context())
+
+		require.Equal(t, errHookFailed, err, "Expected an error from runOnStartHooks")
```
ccoVeille:
```suggestion
		require.ErrorIs(t, err, errHookFailed, "Expected an error from runOnStartHooks")
```

coolwednesday (dismissed):

PR #2084, opened 2025-07-21
Fix/context logger

### LINKED ISSUE #2083: Panic in case of NewCMD app while logging
**Describe the bug**
I am running a command where I am logging using ctx.Warnf - due to which my code panic

**To Reproduce**
Use ctx.Warnf instead of ctx.Logger.Warnf

**Expected behavior**
Code should not panic and the logging should happen

**Screenshots**
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1cd463d]

goroutine 99 [running]:
gofr.dev/pkg/gofr/logging.(*ContextLogger).Warnf(...)
	/home/runner/go/pkg/mod/gofr.dev@v1.41.1-0.20250709062323-22838df194db/pkg/gofr/logging/ctx_logger.go:61

// trace of code in my codebase where I was calling ctx.Warnf
```

Umang01-hash:
## Pull Request Template

**Description:**

-   Closes #2083. 

### Cause of Panic:
```go
	app.SubCommand("hello", func(c *gofr.Context) (any, error) {
		c.Warn("this is a warn log!")
		return "Hello World!", nil
	},
```

This will panic as `c.Warn` uses the **ContextLogger** present inside context to log.
Since in CMD applications, ContextLogger was not initialzied we got into a panic.

### Solution:

-   Move initilization of `NewContextLogger` inside function `newContext` so we don't need to do it manaually.

**Checklist:**

### DIFF
--- a/pkg/gofr/cmd.go
+++ b/pkg/gofr/cmd.go
@@ -13,7 +13,6 @@ import (
 type cmd struct {
 	routes []route
 	out    terminal.Output
-	ctx    Context
 }
 
 type route struct {
--- a/pkg/gofr/cmd.go
+++ b/pkg/gofr/cmd.go
@@ -13,6 +13,7 @@ import (
 type cmd struct {
 	routes []route
 	out    terminal.Output
+	ctx    Context
 }
 
 type route struct {
--- a/pkg/gofr/cmd/request_test.go
+++ b/pkg/gofr/cmd/request_test.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"os"
 	"testing"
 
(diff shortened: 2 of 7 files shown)


pkg/gofr/cmd.go
```diff
@@ -13,6 +13,7 @@ import (
 type cmd struct {
 	routes []route
 	out    terminal.Output
+	ctx    Context
```
ccoVeille: Embeding context is ~forbidden by the Geneva convention~ 😁 _highly discouraged as it's an anti-pattern_

https://github.com/sivchari/containedctx

ccoVeille: One solution could be to embed a ContextLogger instead

ccoVeille (changes requested):

pkg/gofr/cmd.go
Umang01-hash: @ccoVeille Thankyou for your suggestions!!

vipul-rawat (approved):

Issue #2050, opened 2025-07-11
RFC: Standard Error Handling Utility Package in gofr


gizmo-rt:
## Summary
This RFC proposes the introduction of a centralized error handling utility package within the gofr framework. At present, error handling across the framework is somewhat inconsistent and decentralized, with individual packages employing their own mechanisms for capturing and emitting errors. This proposal seeks to establish a standardized approach to error definition, propagation, and handling across the system, with the goal of enabling structured logging, consistent error categorization, and enhanced observability.

## Motivation

- **Inconsistency:** Varying error-handling patterns across packages, which can introduce challenges in maintaining and scaling the codebase.
- **Lack of standardisation:** No unified severity model or error code mapping exists.
- **Instrumentation gaps:** Disparate error formats hinder effective structured logging and telemetry.
- **Security:** Risk of leaking sensitive information due to lack of standardised redaction protocols.
- **Developer friction:** Absence of a shared error vocabulary and mechanism creates additional overhead for new and existing developers.

## Goals

- Provide a common language and model for errors.
- Enable flexibility in defining and configuring application-specific errors.
- Standardize severity levels to make errors more predictable and actionable.
- Enable structured logging and instrumentation by making errors machine-readable and codified.
- Separate internal errors from user-facing messages to avoid data leakage or confusion.

## Design Requirements

- Use standard Go error interface as the foundation.
- Support a configuration schema for:
       - Error codes
       - Internal vs external classification
       - Severity levels (Info, Warning, Critical)
- Allow application developers to extend base error definitions with domain-specific ones.

## Error Model
A standard error object should include the following fields:

```
type AppError struct {
    // Unique error code, e.g., "USER_NOT_FOUND"
    Code        string                
    // user-facing
    ExternalMessage  string
    // developer-facing 
    InternalMessage string 
    Severity    SeverityLevel
    // Wrapped error, optional
    Cause       error     
    // Additional context (safe for logging)
    Meta        map[string]interface{} 
}

```
Severity levels could be defined as:
```

type SeverityLevel string

const (
    // Indicates a potential problem that needs attention, but is not immediately critical. Example: low disk space or high CPU usage
    SeverityWarning  SeverityLevel = "warning" 
    
    // Issue that could lead to data loss or significant service disruption requiring immediate action. Example: application crashes 
    SeverityError    SeverityLevel = "error" 

    // system-crashing error, requiring immediate action. Example: loss of database connection
    SeverityCritical SeverityLevel = "critical" 
[...]

PR #2089, opened 2025-07-23
feat: adds error package

gizmo-rt:
## Pull Request Template

**Description:**

https://github.com/gofr-dev/gofr/issues/2050

NA

**Additional Information:**

No external library used

**Checklist:**


pkg/gofr/serrors/base.go
```diff
@@ -0,0 +1,82 @@
+package serrors
```
ccoVeille, line 1: I'm unsure about the package name

serrors sounds bad

pkg/gofr/serrors/error.go
```diff
@@ -0,0 +1,129 @@
+package serrors
+
+import (
+	"fmt"
+	"time"
+)
+
+const UNSET = "NA"
```
ccoVeille, line 8: Go doesn't use uppercase constants

pkg/gofr/serrors/error.go
```diff
@@ -0,0 +1,129 @@
...
+// Usage:
+//
+//	if err := io.Read(file); err != nil {
+//	    return errors.Wrap(err, "config load failed")
+//	}
+func New(err error, message string) *Error {
+	meta := make(map[string]any)
+
+	// Capture precise timestamp for error tracing
+	meta["timestamp"] = fmt.Sprintf("%d", (time.Now().UTC()).UnixNano())
```
ccoVeille:
```suggestion
	meta["timestamp"] = fmt.Sprint(time.Now().UnixNano())
```

pkg/gofr/serrors/level.go
```diff
@@ -0,0 +1,25 @@
+package serrors
+
+type Level int
+
+const (
+	INFO Level = iota + 1
+	WARNING
+	ERROR
+	CRITICAL
+)
```
ccoVeille, line 10: I would use a common prefix, such as Level

pkg/gofr/serrors/base.go
```diff
@@ -0,0 +1,82 @@
...
+func (err *Error) WithExternalMessage(msg string) ErrorSchema {
+	err.externalMessage = msg
+	return err
+}
+
+func getMetaString(meta map[string]any) string {
+	jsonBytes, err := json.Marshal(meta)
+
+	if err != nil {
+		panic("failed to marshal map to JSON: " + err.Error())
```
ccoVeille: A panic here is dangerous.

pkg/gofr/serrors/error.go
gizmo-rt: https://go.dev/doc/effective_go#constants

pkg/gofr/serrors/error.go
gizmo-rt: I initially made this more explicit to help catch any underlying changes. On reflection, it might be more than required, so I'm considering simplifying it

pkg/gofr/serrors/error.go
ccoVeille: Indeed, I had mixed cap section in mind, with the exception of initialism.

So

Foo ✔️
FOO ✔️
FOO_BAR ❌
FooBar ✔️

pkg/gofr/serrors/base.go
gizmo-rt: Nice catch !

pkg/gofr/serrors/base.go
gizmo-rt: I had a initial thought that since we are adding structure to error, hence the naming `serrors`. Other package names i had in mind were 
`xerr -> extensible error`
`richerr -> rich error`
`errplus -> additional error (similar to extensible)`
Besides,  go has a `errors` library and adding an `s` as a prefix ...i hope Rob may not trigger  a blog

pkg/gofr/serrors/base.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (err *Error) ExternalStatus() int {
+	return err.externalStatusCode
+}
+
+func (err *Error) ExternalMessage() string {
+	return err.externalMessage
+}
+
+func (err *Error) WithStatusCode(value string) ErrorSchema {
```
Umang01-hash, line 32: It might improve clarity to use Go’s conventional setter/getter naming patterns: for example, we can use `StatusCode()` or `SetStatusCode()` if mutators are required. In the current form, the difference between `WithStatusCode` and `Code` can be confusing

pkg/gofr/serrors/config.go
```diff
@@ -0,0 +1,14 @@
+package serrors
+
+type Error struct {
```
Umang01-hash, line 3: The error struct has a lot of different fields, which can be tricky to understand and use. In Go, errors are meant to be as simple as possible. Maybe we could start with the most important fields only, and add more if needed later.

docs/advanced-guide/serrors/page.md
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"fmt"
+	"gofr.dev/pkg/gofr/serrors"
+	"net/http"
+)
+
+const EntityNotFound = "EntityNotFound"
+
+var registry = map[string]serrors.Registry{
```
Umang01-hash, line 47: Defining 'registry' as a package-level variable like this
makes it global and shared across the whole application.
This can create bugs especially in concurrent code or tests. Should we prefer  declaring registries locally within the function or module where they are needed ??

pkg/gofr/serrors/interface.go
```diff
@@ -0,0 +1,18 @@
+package serrors
+
+type ErrorSchema interface {
```
akshat-kumar-singhal, line 3: Is this intended to be only used in the HTTP/delivery layer? If not, are we looking at a separate set of "Status Codes" for store and service layers?

pkg/gofr/serrors/level.go
```diff
@@ -0,0 +1,25 @@
+package serrors
+
+type Level int
+
+const (
```
akshat-kumar-singhal, line 5: We shouldn't be redefining log levels

pkg/gofr/serrors/level_test.go
```diff
@@ -0,0 +1,36 @@
...
+		{"critical level", CRITICAL, TestCritical},
+
+		{"positive invalid level", Level(100), TestUnknown},
+
+		{"negative invalid level", Level(-1), TestUnknown},
+		{"out of range invalid level", Level(9999999999999), TestUnknown},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := tt.level.GetErrorLevel(); got != tt.want {
```
akshat-kumar-singhal, line 31: user assert.Equal/Equalf to make it make it less verbose

pkg/gofr/serrors/config.go
akshat-kumar-singhal: avoid exporting the struct and just have the interface exported

pkg/gofr/serrors/examples/example.go
akshat-kumar-singhal, line 1: shouldn't this be with the other examples in `gofr/examples` folder?

pkg/gofr/serrors/examples/example.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"gofr.dev/pkg/gofr/serrors"
+)
+
+func Read() {
+	err := errors.New("db connection error")
+	serror := serrors.New(err, err.Error())
```
akshat-kumar-singhal, line 11: why should we pass the error as well as error message string from the same error? Are we looking at this as error wrapping functionality? If yes, then the example needs to be fixed.

pkg/gofr/serrors/error.go
```diff
@@ -0,0 +1,129 @@
...
+		level:              0,       // Default error severity
+		retryable:          false,   // Default non retryable
+		meta:               meta,    // Metadata with timestamp
+		externalStatusCode: 0,       // Unset external status code
+		externalMessage:    UNSET,   // Unset external message
+	}
+}
+
+// GetInternalError used when logging errors.
+func GetInternalError(err *Error, addMeta bool) string {
```
akshat-kumar-singhal, line 60: any advantage of not creating this function as a receiver method?

gizmo-rt:
> * Right now, errors need to be built by calling a bunch of methods one after another (WithStatusCode, WithLevel, etc.). Method Chaining is a good approach here but Go code usually sets fields directly or uses simple helper functions. Can we think about a simpler way, so errors are easy to create in one place? Maybe use functional options?
> 
> A few implementation details—like the method chaining, use of the global error registry, and the number of fields—and I wonder if there's a way we could simplify things a bit more ?

@Umang01-hash  Thanks for sharing your insights. 

Here's few sample references
https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/service/auth.go
https://github.com/golang/go/blob/master/src/net/http/request.go

pkg/gofr/serrors/level.go
gizmo-rt: These are error levels. While both log levels and error levels are related to handling system behavior, they serve distinct purposes. 
Log levels primarily help manage log output verbosity, while error levels help manage the flow and handling of errors in the application.

pkg/gofr/serrors/interface.go
gizmo-rt: Design goals as outlined here https://github.com/gofr-dev/gofr/issues/2050

docs/advanced-guide/serrors/page.md
gizmo-rt: Its intended behaviour

pkg/gofr/serrors/error.go
gizmo-rt: to segregate build from use

pkg/gofr/serrors/level.go
akshat-kumar-singhal: The nomenclature for error levels could be different. serrors.ERROR doesn't sound right..

Umang01-hash:
> > * Right now, errors need to be built by calling a bunch of methods one after another (WithStatusCode, WithLevel, etc.). Method Chaining is a good approach here but Go code usually sets fields directly or uses simple helper functions. Can we think about a simpler way, so errors are easy to create in one place? Maybe use functional options?
> > 
> > A few implementation details—like the method chaining, use of the global error registry, and the number of fields—and I wonder if there's a way we could simplify things a bit more ?
> 
> @Umang01-hash Thanks for sharing your insights.
> 
> Here's few sample references https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/service/auth.go https://github.com/golang/go/blob/master/src/net/http/request.go

@gizmo-rt Thanks for sharing those examples!  I see how method chaining fits into our existing patterns. Maybe we can also provide a simpler constructor or helper for the most common cases—just to make error creation a bit quicker for day-to-day use.

pkg/gofr/serrors/error.go
```diff
@@ -0,0 +1,129 @@
...
+		err.level.GetErrorLevel(),
+		err.statusCode,
+		err.subStatusCode,
+		err.message,
+		errorCause,
+	)
+
+	if addMeta {
+		msg += fmt.Sprintf("| %s", getMetaString(err.meta))
+	}
```
akshat-kumar-singhal, line 78: This would output `INFO | 404 | CODE2 | MessageString | ErrorCauseString| MetaString` instead of `INFO | 404 | CODE2 | MessageString | ErrorCauseString | MetaString` (note the space before the last pipe).
Consider create a function which accepts variadic input and returns pipe separated output, internally doing something similar to `strings.Join()` 

Also, consider returning a struct, which can implement Stringer interface to print this format. The struct would allow better handling of error by the receiver instead of being forced to use the raw string.

gizmo-rt:
Update: There has been new changes in the design. Shall be opening a new PR with the updated implementation

PR #2120, opened 2025-08-04
fix:Add delay in startSubscriber function

### LINKED ISSUE #1882: Infinite Loop Without Delay on Error
issue: Add delay in startSubscriber to avoid tight retry loop on error.

About this issue: Currently, the startSubscriber function in SubscriptionManager continuously loops and immediately retries if handleSubscription returns an error. This can cause high CPU usage, log flooding, and unstable behavior in production.

use case:
Adding this delay prevents:
Excessive CPU usage.
Fast log flooding when an error is persistent.
More stable retry behavior.

Beeram12:
## Pull Request Template

fixes #1882

**Description:**

- In the current `SubscriptionManager.startSubscriber` implementation, if `handleSubscription` returns an error, the function immediately retries without any delay. Which causes high CPU usage, log flooding, System instability.
- Fix : Introduced a delay after an error occurs before retrying the subscription.Prevents excessive log spam.

- None

**Checklist:**


coolwednesday:
Hey !

Please fix the code quality issues. 

Thanks !

pkg/gofr/subscriber.go
```diff
@@ -33,6 +34,7 @@ func (s *SubscriptionManager) startSubscriber(ctx context.Context, topic string,
 			err := s.handleSubscription(ctx, topic, handler)
 			if err != nil {
 				s.container.Logger.Errorf("error in subscription for topic %s: %v", topic, err)
+				time.Sleep(time.Second * 2)
```
ccoVeille: time.Sleep is blocking, it's jot a good thing when tge context is canceled

I assume you wanted to do something like this.

```go
// startSubscriber continuously subscribes to a topic and handles messages using the provided handler.
func (s *SubscriptionManager) startSubscriber(ctx context.Context, topic string, handler SubscribeFunc) error {
        var delay time.Duration
        for {
                select {
                case <-ctx.Done():
                        s.container.Logger.Infof("shutting down subscriber for topic %s", topic)
                        return nil
                case <- time.After(delay):
                        err := s.handleSubscription(ctx, topic, handler)
                        if err != nil {
                                s.container.Logger.Errorf("error in subscription for topic %s: %v", topic, err)
                                delay = time.Second * 2
                        }
                }
        }
}
```

Beeram12: Yeah It is something like this. I wanted to add delay.
Thanks for correcting mistake I'll fix them

pkg/gofr/subscriber.go
```diff
@@ -24,6 +25,7 @@ func newSubscriptionManager(c *container.Container) SubscriptionManager {
 
 // startSubscriber continuously subscribes to a topic and handles messages using the provided handler.
 func (s *SubscriptionManager) startSubscriber(ctx context.Context, topic string, handler SubscribeFunc) error {
+	const delay = 2
```
ccoVeille: This is not what I suggested

https://github.com/gofr-dev/gofr/pull/2120#discussion_r2250986663

Beeram12: But as per your suggestion, delay is being declared and initialized but not being used anywhere

ccoVeille: Here is what I suggested

```go
// startSubscriber continuously subscribes to a topic and handles messages using the provided handler.
func (s *SubscriptionManager) startSubscriber(ctx context.Context, topic string, handler SubscribeFunc) error {
        var delay time.Duration
        for {
                select {
                case <-ctx.Done():
                        s.container.Logger.Infof("shutting down subscriber for topic %s", topic)
                        return nil
                case <-time.After(delay):
                        err := s.handleSubscription(ctx, topic, handler)
                        if err != nil {
                                s.container.Logger.Errorf("error in subscription for topic %s: %v", topic, err)
                                delay = time.Second * 2
                        }
                }
        }
}
```

`delay` is used in the ` case <-time.After(delay):` I used to replace the `default ` you were using

Beeram12: okay 
My mistake I didnt look into it

ccoVeille:
The feature could evolve later to make this handle with a setting, instead of hard coded 2 seconds, or use a backoff lib (there are some good module on GitHub)

But for now, I think the PR is good enough.

coolwednesday (approved):
@Beeram12 , Even though this PR is good to go for the start. We can have a logic added for exponential backoff added in retries. 

I would appreciate if you also work on the same.

PR #2140, opened 2025-08-07
En/db resolver

Umang01-hash:
## Pull Request Template

**Description:**

**What is DBResolver?**
- Adds a DBResolver module to GoFr, which provides automatic read/write splitting for SQL databases.

- **HTTP Method**: Write operations (`POST`, `PUT`, `PATCH`, `DELETE`) → Primary | 
Read operations (`GET`) → Replicas

- Seamlessly wraps the existing SQL datasource: does not require any application code changes for existing queries.

- Developers interact with c.SQL exactly as before; all routing and failover are fully transparent.

**Motivation & Benefits**
- Scalable horizontal read performance with multiple replicas
- Reduced load on primary database
- Fault-tolerant: automatic fallback to primary if all replicas fail
- Clean metrics and tracing support for operational visibility

## Example Usage:
```go
package main

import (
	"gofr.dev/pkg/gofr"
	"gofr.dev/pkg/gofr/datasource/dbresolver"
)

type Customer struct {
	ID   int    `db:"id"`
	Name string `db:"name"`
}

func main() {
	a := gofr.New()

	// Initialize DB resolver with default settings
	err := dbresolver.InitDBResolver(a, &dbresolver.Config{
		Strategy:      dbresolver.StrategyRoundRobin, // use round-robin strategy or random strategy
		ReadFallback:  true, // allow reads on primary if all replicas are down
		MaxFailures:   3, 			  // number of allowed failures before marking a replica as down
		TimeoutSec:    30, // timeout for marking a replica as down
		PrimaryRoutes: []string{"/admin", "/api/payments/*"}, // routes that should go to primary

		Replicas: []dbresolver.ReplicaCredential{
			{
				Host:     "localhost:3307",
				User:     "replica_user1",
				Password: "pass1",
			},
			{
				Host:     "replica2.example.com:3308",
				User:     "replica_user2",
				Password: "pass2",
			},
			{
				Host:     "replica3.example.com:3309",
				User:     "replica_user3",
				Password: "pass3",
			},
		},
	})
	if err != nil {
		a.Logger().Errorf("failed to initialize db resolver: %v", err)
	}

	// Read endpoint - goes to replica
	a.GET("/customers", func(c *gofr.Context) (interface{}, error) {
		var customers []Customer

		c.SQL.Select(c, &customers, "SELECT id, name FROM customers")

		return customers, err
	})

	// Write endpoint - goes to primary
	a.POST("/customers", func(c *gofr.Context) (interface{}, error) {
		var customer Customer

		c.Bind(&customer)

		_, err := c.SQL.Exec("INSERT INTO customers (name) VALUES (?)", customer.Name)

		return customer, err
	})

	// Admin endpoint - forced to primary
	a.GET("/admin/customers", func(c *gofr.Context) (interface{}, error) {
		var customers []Customer

		c.SQL.Select(c, &customers, "SELECT id, name FROM customers")

		return customers, err
	})

	a.Run()
}

```

## Configuration Example:

```env
DB_HOST=localhost
DB_USER=root
DB_PASSWORD=rootpassword
DB_NAME=testdb
DB_PORT=3306
DB_DIALECT=mysql
DB_MAX_IDLE_CONNECTION=2
DB_MAX_OPEN_CONNECTION=0
```

## Testing Strategy:

- Primary and Replicas launched with docker-compose (primary:3306, replicas:3307/3308)

- Replication automated by setup scripts; seed data from SQL dump and app endpoints

- Load Testing: Performed with Apache JMeter simulating high-concurrency API requests (both reads and writes)

**Results:**

- Zero error rate; throughput and latency showed no regression compared to the baseline.

- Read/write split is fully performant, and scaling is achieved without application change.

**Checklist:**


go.work
```diff
@@ -1,6 +1,4 @@
-go 1.24.0
-
-toolchain go1.24.0
+go 1.24.5
```
ccoVeille: Is there a need to enforce 1.24.5 ?

pkg/gofr/datasource/dbresolver/dbresolver_wrapper.go
```diff
@@ -0,0 +1,81 @@
+package dbresolver
+
+import (
+	"errors"
+
+	"go.opentelemetry.io/otel/trace"
+	"gofr.dev/pkg/gofr/container"
+)
```
ccoVeille:
```suggestion
import (
	"errors"

	"go.opentelemetry.io/otel/trace"

	"gofr.dev/pkg/gofr/container"
)
```

pkg/gofr/datasource/dbresolver/dbresolver_wrapper.go
```diff
@@ -0,0 +1,81 @@
...
+	// Create options slice
+	var opts []Option
+
+	var strategy Strategy
+	if r.strategyName == "random" {
+		strategy = NewRandomStrategy()
+	} else {
+		// Default to round-robin for any other value including "round-robin"
+		strategy = NewRoundRobinStrategy(len(replicas))
+	}
```
ccoVeille: I tend to avoid else when possible


```suggestion
	// Default to round-robin
	strategy := NewRoundRobinStrategy(len(replicas))
	if r.strategyName == "random" {
		strategy = NewRandomStrategy()
	}
```

pkg/gofr/datasource/dbresolver/dbresolver_wrapper_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel"
+	"go.uber.org/mock/gomock"
+	"gofr.dev/pkg/gofr/container"
+)
```
ccoVeille:
```suggestion
import (
	"sync"
	"testing"
	"time"

	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
	"go.opentelemetry.io/otel"
	"go.uber.org/mock/gomock"

	"gofr.dev/pkg/gofr/container"
)
```

pkg/gofr/datasource/dbresolver/logger.go
```diff
@@ -0,0 +1,41 @@
...
+	formattedLog := fmt.Sprintf("\u001B[38;5;8m%-32s \u001B[38;5;206m%-6s\u001B[0m %8d\u001B[38;5;8mµs\u001B[0m %s %s\n",
+		clean(ql.Type), "DBRESOLVER", ql.Duration,
+		clean(fmt.Sprintf("%s %s", ql.Target, ql.Query)), clean(ql.Query))
+
+	logger.Debug(formattedLog)
+}
+
+func clean(query string) string {
+	// Replace multiple consecutive whitespace characters with a single space
+	query = regexp.MustCompile(`\s+`).ReplaceAllString(query, " ")
```
ccoVeille: A must compile with a fixed regexp should be moved at package level

It would avoid to compute it on each function call

gizmo-rt:
@Umang01-hash  Can you add more details on sequence of R(Read) and W(Write) cases under which read and write replicas would be selected ?

Umang01-hash:
> @Umang01-hash Can you add more details on sequence of R(Read) and W(Write) cases under which read and write replicas would be selected ?

Hey @gizmo-rt we first determine is a query is read/write and if it is a read query we check if healthyReplica is available or not. If the replica is available we select it and send the read query to it and if the replica is not available we send it to primary 
database. Methods like `QueryContext`, `Select` have this logic. Methods like `Exec` , `Prepare` etc are directly using the primary db. Transaction methods (Begin, BeginTx) are always routed to the primary.

If all replicas are unhealthy at time of a read query and fallback is enabled, the read falls back to the primary. If fallback is disabled, the read fails with an error.

If a replica experiences multiple failures, it's circuit breaker opens and replica is temporarily skipped for queries. This timeout period is 30 seconds by default and we allow 5 failures for replica before cicuit breaker is opened. 

For an example sequence like **R,R,W,R**:

1st R: Routed to replica

2nd R: Routed to next replica (depends on which strategy is choosen random or round-robin)

1st W: Routed to primary

3rd R:  Routed to next replica

gizmo-rt:
> > @Umang01-hash Can you add more details on sequence of R(Read) and W(Write) cases under which read and write replicas would be selected ?
> 
> Hey @gizmo-rt we first determine is a query is read/write and if it is a read query we check if healthyReplica is available or not. If the replica is available we select it and send the read query to it and if the replica is not available we send it to primary database. Methods like `QueryContext`, `Select` have this logic. Methods like `Exec` , `Prepare` etc are directly using the primary db. Transaction methods (Begin, BeginTx) are always routed to the primary.
> 
> If all replicas are unhealthy at time of a read query and fallback is enabled, the read falls back to the primary. If fallback is disabled, the read fails with an error.
> 
> If a replica experiences multiple failures, it's circuit breaker opens and replica is temporarily skipped for queries. This timeout period is 30 seconds by default and we allow 5 failures for replica before cicuit breaker is opened.
> 
> For an example sequence like **R,R,W,R**:
> 
> 1st R: Routed to replica
> 
> 2nd R: Routed to next replica (depends on which strategy is choosen random or round-robin)
> 
> 1st W: Routed to primary
> 
> 3rd R: Routed to next replica

So if we have 10 replicas, then in the worst case with 5 retries and 30s timeout user has to wait 1500 sec i.e 25 mins before getting a response ??  Curious, observed request timeout is generally in range 30s - 5mins

Umang01-hash:
> > > @Umang01-hash Can you add more details on sequence of R(Read) and W(Write) cases under which read and write replicas would be selected ?
> > 
> > 
> > Hey @gizmo-rt we first determine is a query is read/write and if it is a read query we check if healthyReplica is available or not. If the replica is available we select it and send the read query to it and if the replica is not available we send it to primary database. Methods like `QueryContext`, `Select` have this logic. Methods like `Exec` , `Prepare` etc are directly using the primary db. Transaction methods (Begin, BeginTx) are always routed to the primary.
> > If all replicas are unhealthy at time of a read query and fallback is enabled, the read falls back to the primary. If fallback is disabled, the read fails with an error.
> > If a replica experiences multiple failures, it's circuit breaker opens and replica is temporarily skipped for queries. This timeout period is 30 seconds by default and we allow 5 failures for replica before cicuit breaker is opened.
> > For an example sequence like **R,R,W,R**:
> > 1st R: Routed to replica
> > 2nd R: Routed to next replica (depends on which strategy is choosen random or round-robin)
> > 1st W: Routed to primary
> > 3rd R: Routed to next replica
> 
> So if we have 10 replicas, then in the worst case with 5 retries and 30s timeout user has to wait 1500 sec i.e 25 mins before getting a response ?? Curious, observed request timeout is generally in range 30s - 5mins

No @gizmo-rt ! It doesn;t happens that way. 

When a read request comes in:

1. We check our list of replicas and only consider ones with closed circuit breakers
2. If a replica fails, we immediately try fallback (if enabled) or return error

The 30-second timeout is how long a circuit breaker stays OPEN after failures, not how long each request waits. Once a circuit breaker opens, we stop sending traffic to that replica for 30 seconds. There is no chained retry across all replicas per request.

The system actively avoids replicas with open circuit breakers during routing, so those replicas are simply skipped.

A typical request flow with failing replicas:

**Request comes in**
Check replica 1 → circuit open (skip)
Check replica 2 → circuit open (skip)
Try replica 3 → fails immediately
Fall back to primary → succeeds

go.work
```diff
@@ -22,4 +22,5 @@ use (
 	./pkg/gofr/datasource/scylladb
 	./pkg/gofr/datasource/solr
 	./pkg/gofr/datasource/surrealdb
+	pkg/gofr/datasource/dbresolver
```
ccoVeille: I would have kept the same format and I would have ordered them alphabeticaly, as they were, no?

Also, it would decrease the risk of conflicts

ccoVeille (approved):
I let the discussion going as I wasn't sure where it would land.

So functionally, I'm not sure whether it's OK.

But the code seems OK

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,610 @@
...
+		return false
+	}
+
+	// Fast string comparison for common cases
+	firstSix := strings.ToUpper(trimmed[:6])
+	switch firstSix {
+	case "SELECT":
+		return true
+	case "SHOW  ", "DESCRI", "EXPLAI":
+		return true
```
akshat-kumar-singhal: any reason for not clubbing all the scenarios into a single condition given that they're returning the same response?

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,610 @@
...
+	}
+
+	// Trim whitespace and get first word
+	trimmed := strings.TrimLeft(query, " \t\n\r")
+	if len(trimmed) < minQueryLength {
+		return false
+	}
+
+	// Fast string comparison for common cases
+	firstSix := strings.ToUpper(trimmed[:6])
```
akshat-kumar-singhal: Would this be any better than just having the strings as is in the switch case?

akshat-kumar-singhal:
@Umang01-hash Hope we are routing the reads within a transaction to the primary? I couldn't find any the corresponding code for it.

Umang01-hash:
> Hope we are routing the reads within a transaction to the primary? I couldn't find any the corresponding code for it.

Yes @akshat-kumar-singhal Reads within transactions are always routed to the primary database because the transaction itself is created on the primary connection. Since **Begin()** returns a transaction object from the **primary database**, all subsequent operations on this transaction (including reads) automatically use the primary connection.

```go
// Begin always routes to primary (transactions).
func (r *Resolver) Begin() (*gofrSQL.Tx, error) {
    r.stats.totalQueries.Add(1)
    return r.primary.Begin()
}
```

pkg/gofr/datasource/dbresolver/logger.go
akshat-kumar-singhal, line 1: why do we need to define logger interface everywhere? How about reusing the existing interface?

pkg/gofr/datasource/dbresolver/metrics.go
```diff
@@ -0,0 +1,13 @@
+package dbresolver
+
+import (
+	"context"
+)
+
+// Metrics interface for metrics operations.
+type Metrics interface {
```
akshat-kumar-singhal, line 8: Considered using existing metrics from gofr?

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,608 @@
...
+	defaultTimeoutSec    int   = 30
+	defaultCacheSize     int64 = 1000
+	minQueryLength       int   = 6
+)
+
+// Pre-compiled regex - compiled once at package init.
+var readQueryRegex = regexp.MustCompile(`(?i)^\s*(SELECT|SHOW|DESCRIBE|EXPLAIN)`)
+
+// queryCache implements a thread-safe, size-limited cache for query classification.
+type queryCache struct {
```
akshat-kumar-singhal: Consider separating the queryCache implementation

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,608 @@
...
+
+	// Check cache for edge cases
+	if cached, exists := r.queryCache.get(query); exists {
+		return cached
+	}
+
+	// Fallback to regex for complex queries
+	isRead := readQueryRegex.MatchString(query)
+
+	r.queryCache.set(query, isRead)
```
akshat-kumar-singhal: storing the query in sync map might end up impacting the concurrency. Did we do a benchmark to see the performance of regex vs switch vs syncMap usage?

pkg/gofr/datasource/dbresolver/strategy.go
```diff
@@ -0,0 +1,75 @@
...
+// Choose selects the next replica in round-robin fashion.
+// Choose selects the next replica in round-robin fashion.
+func (s *RoundRobinStrategy) Choose(replicas []container.DB) (container.DB, error) {
+	if len(replicas) == 0 {
+		return nil, errNoReplicasAvailable
+	}
+
+	count := s.current.Add(1)
+
+	replicaCount := int64(len(replicas))
```
akshat-kumar-singhal: Consider this to avoid counting the replica length twice

    replicaCount := int64(len(replicas))
    if replicaCount == 0 {
		return nil, errNoReplicasAvailable
	}

	count := s.current.Add(1)

pkg/gofr/datasource/dbresolver/strategy.go
```diff
@@ -0,0 +1,75 @@
+package dbresolver
+
+import (
+	"errors"
+	"math/rand"
+	"sync/atomic"
+
+	"gofr.dev/pkg/gofr/container"
+)
+
```
akshat-kumar-singhal, line 7: Move these here. 

```
    roundRobinStrategy         = "round-robin"
	randomStrategy             = "random"
````

Also, consider defining `type StrategyType string` instead of using raw strings

File: pkg/gofr/datasource/dbresolver/dbresolver_wrapper.go [hunk 14]
```diff
@@ -0,0 +1,80 @@
...
+type ResolverWrapper struct {
+	logger       Logger
+	metrics      Metrics
+	tracer       trace.Tracer
+	strategyName string
+	readFallback bool
+}
+
+// NewProvider creates a new ResolverWrapper with strategy and fallback config.
+func NewProvider(strategy string, readFallback bool) *ResolverWrapper {
```
akshat-kumar-singhal: Consider taking a `Strategy` instance instead of strategy name

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,608 @@
...
+	if failures >= cb.maxFailures {
+		cb.state.Store(1)
+	}
+}
+
+// Replica wrapper with circuit breaker.
+type replicaWrapper struct {
+	db      container.DB
+	breaker *circuitBreaker
+	index   int
```
akshat-kumar-singhal, line 51: This is redundant given that replicaWrapper would be used as an element in an array (indexed)

akshat-kumar-singhal (changes requested):
Try to break the files into smaller logic groups - readability is a bit low right now.

docs/quick-start/connecting-mysql/page.md
```diff
@@ -239,3 +239,42 @@ Now when we access {% new-tab-link title="http://localhost:9000/customer" href="
...
+- The first argument specifies the **load balancing strategy** for read queries. Supported values:
+   - `round-robin`: Distributes reads evenly across replicas.
+   - `random`: Selects a replica at random for each read.
+- The second argument enables **fallback** to the primary if all replicas are unavailable.
+
+###  Configuration
+Add replica hosts to your `.env` file using the `DB_REPLICA_HOSTS` variable:
+
+```env
+DB_REPLICA_HOSTS=localhost:3307,localhost:3308
```
aryanmehrotra: we would need other configurations as well.

pkg/gofr/datasource/dbresolver/dbresolver_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+	if r.strategyName == "random" {
+		strategy = NewRandomStrategy()
+	}
+
+	// Add options.
+	opts = append(opts, WithStrategy(strategy), WithFallback(r.readFallback))
+
+	// Create and return the resolver.
+	return NewResolver(primary, replicas, r.logger, r.metrics, opts...), nil
+}
```
aryanmehrotra: what if we use the existing Connect method?

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,608 @@
...
+	"gofr.dev/pkg/gofr/datasource"
+	gofrSQL "gofr.dev/pkg/gofr/datasource/sql"
+)
+
+// Constants for strategies and intervals.
+const (
+	healthStatusUP             = "UP"
+	healthStatusDOWN           = "DOWN"
+	roundRobinStrategy         = "round-robin"
+	randomStrategy             = "random"
```
aryanmehrotra: these can be types instead of string

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,608 @@
...
+	}
+}
+
+func (c *queryCache) get(key string) (value, exists bool) {
+	val, found := c.cache.Load(key)
+	if !found {
+		return false, false
+	}
+
+	return val.(bool), true
```
aryanmehrotra: we should not to in-line type casting

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,608 @@
...
+	maxFailures int32
+	timeout     time.Duration
+}
+
+func newCircuitBreaker(maxFailures int32, timeout time.Duration) *circuitBreaker {
+	return &circuitBreaker{
+		maxFailures: maxFailures,
+		timeout:     timeout,
+	}
+}
```
aryanmehrotra: circuit breaker things should exist in the separate file

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,608 @@
...
+	stats      *statistics
+
+	// Background task management.
+	stopChan chan struct{}
+	wg       sync.WaitGroup
+	once     sync.Once
+}
+
+// Option is a function type for configuring the resolver.
+type Option func(*Resolver)
```
aryanmehrotra: this can also exist in a separate file

pkg/gofr/external_db.go
```diff
@@ -1,10 +1,26 @@
...
+)
+
+const (
+	maxIdleReplicaCap  = 50
+	minIdleReplica     = 10
+	maxOpenReplicaCap  = 200
+	minOpenReplica     = 50
+	defaultIdleReplica = 10
+	defaultOpenReplica = 100
 )
```
aryanmehrotra: this should be configurable by the user, or how these should be decided or how the user can decide these should be documented

docs/quick-start/connecting-mysql/page.md
```diff
@@ -239,3 +239,42 @@ Now when we access {% new-tab-link title="http://localhost:9000/customer" href="
...
+```
```
coolwednesday: SQL has multiple configurations like max open connections, idle connections ..... Will the same be applied here also ?

docs/references/configs/page.md
```diff
@@ -195,6 +195,12 @@ This document lists all the configuration options supported by the GoFr framewor
 
 ---
 
+- DB_REPLICA_HOSTS
+- Comma-separated list of replica database hosts. Used for read replicas.
+- None
+
+---
```
coolwednesday, line 220: We can add  users and password and other configs also specific to the replicas.

pkg/gofr/datasource/dbresolver/metrics.go
coolwednesday: We are just defining the interface here, since it is a separate module .

docs/quick-start/connecting-mysql/page.md
```diff
@@ -239,3 +239,73 @@ Now when we access {% new-tab-link title="http://localhost:9000/customer" href="
...
+
+After importing the package, you can configure the DBResolver in your GoFr application using the `AddDBResolver` method. 
+You can choose the load balancing strategy and enable fallback to primary:
+
+```go
+// Add DB resolver with round-robin strategy and fallback enabled
+resolver := dbresolver.NewProvider("round-robin", true)
+a.AddDBResolver(resolver)
+```
```
coolwednesday: The documentation needs to be updated here.

Umang01-hash:
This PR implements SQL-level routing (parsing queries) to split reads vs writes. Another approach is to use the HTTP method (e.g. `GET`=**read**, `POST/PUT`=**write**) to choose the DB. Given cases where a handler might do `SELECT` then `UPDATE`, or a `GET`-route that writes, which strategy do people prefer? Any thoughts on pros/cons?

@aryanmehrotra @ccoVeille @akshat-kumar-singhal any suggestions on it??

pkg/gofr/datasource/dbresolver/circuit_breaker.go
```diff
@@ -0,0 +1,55 @@
...
+	case circuitStateHalfOpen:
+		return true
+	default:
+		return true
+	}
+}
+
+func (cb *circuitBreaker) recordSuccess() {
+	cb.failures.Store(0)
+	cb.state.Store(0)
```
ccoVeille:
```suggestion
	cb.state.Store(circuitStateClosed)
```

pkg/gofr/datasource/dbresolver/circuit_breaker.go
```diff
@@ -0,0 +1,55 @@
...
+			return cb.state.CompareAndSwap(circuitStateOpen, circuitStateHalfOpen)
+		}
+
+		return false
+	case circuitStateHalfOpen:
+		return true
+	default:
+		return true
+	}
+}
```
ccoVeille, line 58:
```suggestion
func (cb *circuitBreaker) allowRequest() bool {
	state := cb.state.Load()

	if state != circuitStateOpen {
		return true
	}

	if time.Since(time.Unix(0, cb.lastFailure.Load())) <= cb.timeout {
		return false
	}

	return cb.state.CompareAndSwap(circuitStateOpen, circuitStateHalfOpen)
}
```

pkg/gofr/datasource/dbresolver/circuit_breaker.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"sync/atomic"
+	"time"
+)
+
+// Circuit breaker for replica health with atomic operations.
+type circuitBreaker struct {
+	failures    atomic.Int32
+	lastFailure atomic.Int64
```
ccoVeille: If this is a time, then use atomic.Value

Here it looks like strange.

Maybe this

https://github.com/changkun/redir/blob/3464f00e4071fa80d9eb4a56135589d92d50f708/auth.go#L47

Or using atomic.Pointer[time.Time]

pkg/gofr/datasource/dbresolver/circuit_breaker.go
```diff
@@ -0,0 +1,55 @@
...
+	state       atomic.Int32 // 0=closed, 1=open, 2=half-open.
```
ccoVeille: Here also, I might use atomic.Value

You could also use atomic.Pointer[circuitBreakerState]

By defining circuitBreakerState as the toe of all circuitBreakerState

The pointer is overkill here, but it makes it clear what you are dealing with

pkg/gofr/datasource/dbresolver/circuit_breaker.go
```diff
@@ -0,0 +1,55 @@
...
+		return false
+	case circuitStateHalfOpen:
+		return true
+	default:
+		return true
+	}
+}
+
+func (cb *circuitBreaker) recordSuccess() {
+	cb.failures.Store(0)
```
ccoVeille, line 61: Here you could reset the last failure with time.Time{} or simply nil depending on what you apply from my previous comment

pkg/gofr/datasource/dbresolver/circuit_breaker.go
```diff
@@ -0,0 +1,55 @@
...
+	cb.state.Store(0)
+}
+
+func (cb *circuitBreaker) recordFailure() {
+	failures := cb.failures.Add(1)
+	cb.lastFailure.Store(time.Now().UnixNano())
+
+	if failures >= cb.maxFailures {
+		cb.state.Store(1)
```
ccoVeille: use your constant

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,504 @@
...
+const (
+	healthStatusUP             = "UP"
+	healthStatusDOWN           = "DOWN"
+	circuitStateClosed   int32 = 0
+	circuitStateOpen     int32 = 1
+	circuitStateHalfOpen int32 = 2
+	defaultMaxFailures   int32 = 5
+	defaultTimeoutSec    int   = 30
+	defaultCacheSize     int64 = 1000
+	minQueryLength       int   = 6
```
ccoVeille: Remove it, and love it where you need it

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,504 @@
...
+		return cached
+	}
+
+	// Fallback to regex for complex queries
+	isRead := readQueryRegex.MatchString(query)
+
+	r.queryCache.set(query, isRead)
+
+	return isRead
+}
```
ccoVeille: I reduced size to 4 because no one will use show followed by 2 spaces, especially after you trimmed.

```suggestion
	const minQueryLength = 4

	// Fast path: check first few characters for common patterns
	// Trim whitespace and get first word
	trimmed := strings.TrimLeft(query, " \t\n\r")
	if len(trimmed) < minQueryLength {
		return false
	}

	// Fast string comparison for common cases
	firstSix := strings.ToUpper(trimmed[:minQueryLength])
	switch firstSix {
	case "SELE", "SHOW", "DESC", "EXPL":
		return true
	}

	// Check cache for edge cases
	if cached, exists := r.queryCache.get(query); exists {
		return cached
	}

	// Fallback to regex for complex queries
	isRead := readQueryRegex.MatchString(query)

	r.queryCache.set(query, isRead)

	return isRead
}
```

docs/quick-start/connecting-mysql/page.md
```diff
@@ -239,3 +239,73 @@ Now when we access {% new-tab-link title="http://localhost:9000/customer" href="
...
+```shell
+go get gofr.dev/pkg/gofr/datasource/dbresolver@latest
+```
+
+After importing the package, you can configure the DBResolver in your GoFr application using the `AddDBResolver` method. 
+You can choose the load balancing strategy and enable fallback to primary:
+
+```go
+// Add DB resolver with round-robin strategy and fallback enabled
+resolver := dbresolver.NewProvider("round-robin", true)
```
ccoVeille: Use the constant you defined here

akshat-kumar-singhal:
> This PR implements SQL-level routing (parsing queries) to split reads vs writes. Another approach is to use the HTTP method (e.g. `GET`=**read**, `POST/PUT`=**write**) to choose the DB. Given cases where a handler might do `SELECT` then `UPDATE`, or a `GET`-route that writes, which strategy do people prefer? Any thoughts on pros/cons?
> 
> @aryanmehrotra @ccoVeille @akshat-kumar-singhal any suggestions on it??

@Umang01-hash Any issues you see with the current implementation of identifying it based on the query?
Another way to think of this implementation is by drawing a parallel to cache. Given that we'd want to limit the visibility of cache/read DB within the store layer, the store layer should be able to decide whether to use primary or replica/cache based on the operation (SELECT/UPDATE etc) and transaction.

aryanmehrotra:
> > This PR implements SQL-level routing (parsing queries) to split reads vs writes. Another approach is to use the HTTP method (e.g. `GET`=**read**, `POST/PUT`=**write**) to choose the DB. Given cases where a handler might do `SELECT` then `UPDATE`, or a `GET`-route that writes, which strategy do people prefer? Any thoughts on pros/cons?
> > @aryanmehrotra @ccoVeille @akshat-kumar-singhal any suggestions on it??
> 
> @Umang01-hash Any issues you see with the current implementation of identifying it based on the query? Another way to think of this implementation is by drawing a parallel to cache. Given that we'd want to limit the visibility of cache/read DB within the store layer, the store layer should be able to decide whether to use primary or replica/cache based on the operation (SELECT/UPDATE etc) and transaction.

One issue which might occur is if there is a POST request and user is first inserting the value and then fetching it.
based on the query pattern which we have currently followed there are good chances that while fetching we will not get the resource, so the correct approach seems to be based on the request method type rather than query type.

akshat-kumar-singhal:
> One issue which might occur is if there is a POST request and user is first inserting the value and then fetching it.
based on the query pattern which we have currently followed there are good chances that while fetching we will not get the resource, so the correct approach seems to be based on the request method type rather than query type.

@aryanmehrotra While selecting the primary/replica based on request method type feels safer, we may still have issues due to replication lag. Also, in case of poor implementation (by application dev), example - DB writes in GET call, we can have other issues.
We may want to consider supporting a config - list of end points (maybe) which would operate on the replica instead of primary. This would provide more control to the developers and heavier read operations can be safely routed to the replica.

aryanmehrotra:
> > One issue which might occur is if there is a POST request and user is first inserting the value and then fetching it.
> > based on the query pattern which we have currently followed there are good chances that while fetching we will not get the resource, so the correct approach seems to be based on the request method type rather than query type.
> 
> @aryanmehrotra While selecting the primary/replica based on request method type feels safer, we may still have issues due to replication lag. Also, in case of poor implementation (by application dev), example - DB writes in GET call, we can have other issues. We may want to consider supporting a config - list of end points (maybe) which would operate on the replica instead of primary. This would provide more control to the developers and heavier read operations can be safely routed to the replica.

We should document poor practices like DB writes in GET calls, since features like status code handling are based on request type and overriding them breaks best practices.

For routing queries, we could either:

Support a config with endpoints that always use replicas for heavy reads, or
Let users specify which queries run on which DB.

This covers more use cases, but ideally we want a cleaner approach where devs don’t have to manage this manually. If not, either of the options works.

Umang01-hash:
Hey! Here are my sugegstions for the above problems/comments:

1. **Transaction Scoping:** When developers use `Begin() → Commit() `(database transactions), all queries inside that block would **automatically go to the primary database.**  This would solve the POST → INSERT → SELECT pattern because both operations happen on the same database.

2. **Read-After-Write Consistency:** We could track when any write happens and for the next few seconds, automatically route all read queries to primary instead of replicas. This would handle cases where you INSERT data and immediately try to SELECT it outside of transactions.

3. **Why SQL-Based Is Still Better:** HTTP methods (GET/POST) don't tell us what the actual SQL query does. A GET endpoint might cache data (write), or a POST might only validate data (read). Our SQL-based approach looks at the actual database operation (SELECT vs INSERT) rather than guessing from the HTTP method.

Question to all: Does the above approach sound reasonable? Would these two additions solve the consistency concerns while keeping the implementation simple? Happy to explore HTTP method-based routing if you think it's a better path forward.

[@aryanmehrotra](https://github.com/aryanmehrotra) [@ccoVeille](https://github.com/ccoVeille) [@akshat-kumar-singhal](https://github.com/akshat-kumar-singhal)  [@gizmo-rt](https://github.com/gizmo-rt) - thoughts?

aryanmehrotra:
> Hey! Here are my sugegstions for the above problems/comments:
> 
> 1. **Transaction Scoping:** When developers use `Begin() → Commit() `(database transactions), all queries inside that block would **automatically go to the primary database.**  This would solve the POST → INSERT → SELECT pattern because both operations happen on the same database.

I highly doubt if someone in a transaction would do POST → INSERT → SELECT, as mostly we do insert/update in transactions, SELECT happens separetely.

Also, if we think about the three layer architecture, there are different method for creating an entry and getting that entry, right now we can't do that in a single transaction in gofr.

> 2. **Read-After-Write Consistency:** We could track when any write happens and for the next few seconds, automatically route all read queries to primary instead of replicas. This would handle cases where you INSERT data and immediately try to SELECT it outside of transactions.

It would be very expensive, imagine writes or updates happening from a cron-job every 15 minutes then mostly it would be happening in a particular table, then we would also need to be table aware in that case. otherwise even after having the read/write split most queries would go to the write replica.

> 3. **Why SQL-Based Is Still Better:** HTTP methods (GET/POST) don't tell us what the actual SQL query does. A GET endpoint might cache data (write), or a POST might only validate data (read). Our SQL-based approach looks at the actual database operation (SELECT vs INSERT) rather than guessing from the HTTP method.

What does the best-practise say about GET and POST request endpoints functioning and what we already follow in gofr?

Umang01-hash:
@aryanmehrotra Thanks for your suggestions. Your points seem valid - transaction scoping wouldn't work in three-layer architecture, and global read-after-write tracking would be too expensive.

 **Request-Scoped Routing:** Instead of query-level or global tracking, **we can route all queries within a single HTTP request based on the HTTP method**. So if someone hits POST /users, ALL database queries (both INSERT and SELECT) within that request handler automatically go to primary. GET requests continue using replicas.

**This solves the POST → INSERT → SELECT consistency issue and aligns with REST best practices - GET should be safe (no side effects) and POST/PUT for state changes.** GoFr already follows these patterns in its routing and status code handling. Thoughts on this approach?​


An extension on this as suggested by @akshat-kumar-singhal is: 
**Route-Level Configuration:** We can also provide optional endpoint-level control where developers can configure specific routes to always use primary:

```
# Optional: Force specific endpoints to always use primary
DB_PRIMARY_ROUTES=/users/search,/reports/heavy,/admin/*

```

akshat-kumar-singhal:
Routes will need to have the method as well. We would be better off
accepting this configuration via code instead of complicating the ENV file.
What we should ensure is the ability to switch seamlessly between
primary-replica setup vs primary only (local).

aryanmehrotra:
> @aryanmehrotra Thanks for your suggestions. Your points seem valid - transaction scoping wouldn't work in three-layer architecture, and global read-after-write tracking would be too expensive.
> 
> **Request-Scoped Routing:** Instead of query-level or global tracking, **we can route all queries within a single HTTP request based on the HTTP method**. So if someone hits POST /users, ALL database queries (both INSERT and SELECT) within that request handler automatically go to primary. GET requests continue using replicas.
> 
> **This solves the POST → INSERT → SELECT consistency issue and aligns with REST best practices - GET should be safe (no side effects) and POST/PUT for state changes.** GoFr already follows these patterns in its routing and status code handling. Thoughts on this approach?​
> 
> An extension on this as suggested by @akshat-kumar-singhal is: **Route-Level Configuration:** We can also provide optional endpoint-level control where developers can configure specific routes to always use primary:
> 
> ```
> # Optional: Force specific endpoints to always use primary
> DB_PRIMARY_ROUTES=/users/search,/reports/heavy,/admin/*
> ```

Yes this seems better, also env variable shouldn't be used for this we should give option from the code itself, this can be part as an enhancement based on user feedback once we rollout the feature just based on the route level.

pkg/gofr/external_db.go
```diff
@@ -221,6 +237,195 @@ func (a *App) AddCouchbase(db container.CouchbaseProvider) {
...
+	val, err := strconv.Atoi(valStr)
+
+	if err != nil || val <= 0 {
+		return fallback
+	}
+
+	return val
+}
+
+func optimizedIdleConnections(cfg config.Config) string {
```
ckshitij: [Suggestion] Both `optimizedIdleConnections()` and `optimizedOpenConnections()` share almost identical logic (load config ->  parse value -> apply multiplier -> clamp with min/max).
can we create a reusable helper function like this:

```go
func optimizedConnections(cfg config.Config, key, maxKey, minKey, defaultKey string, multiplier, maxDefault, minDefault, defaultValDefault int) string {
	maxCap := getReplicaConfigInt(cfg, maxKey, maxDefault)
	minVal := getReplicaConfigInt(cfg, minKey, minDefault)
	defaultVal := getReplicaConfigInt(cfg, defaultKey, defaultValDefault)

	val, err := strconv.Atoi(cfg.Get(key))
	if err != nil || val <= 0 {
		return strconv.Itoa(defaultVal)
	}

	optimized := val * multiplier

	switch {
	case optimized > maxCap:
		optimized = maxCap
	case optimized < minVal:
		optimized = minVal
	}

	return strconv.Itoa(optimized)
}

func optimizedIdleConnections(cfg config.Config) string {
	return optimizedConnections(cfg,
		"DB_MAX_IDLE_CONNECTION",
		"DB_REPLICA_MAX_IDLE_CONNECTIONS",
		"DB_REPLICA_MIN_IDLE_CONNECTIONS",
		"DB_REPLICA_DEFAULT_IDLE_CONNECTIONS",
		4,
		maxIdleReplicaCapDefault,
		minIdleReplicaDefault,
		defaultIdleReplicaDefault,
	)
}

func optimizedOpenConnections(cfg config.Config) string {
	return optimizedConnections(cfg,
		"DB_MAX_OPEN_CONNECTION",
		"DB_REPLICA_MAX_OPEN_CONNECTIONS",
		"DB_REPLICA_MIN_OPEN_CONNECTIONS",
		"DB_REPLICA_DEFAULT_OPEN_CONNECTIONS",
		2,
		maxOpenReplicaCapDefault,
		minOpenReplicaDefault,
		defaultOpenReplicaDefault,
	)
}
```

pkg/gofr/context.go
```diff
@@ -158,8 +158,10 @@ func (a *authInfo) GetAPIKey() string {
 // }
 
 func newContext(w Responder, r Request, c *container.Container) *Context {
```
ckshitij, line 160: In `newContext`, we’re creating a `baseCtx := context.WithValue(r.Context(), "request", r)` and passing that as the context for the Context struct.

What is the purpose of introducing this `baseCtx` instead of just using `r.Context()` directly, and is it actually needed here?

If yes, should we also do something similar in `newCMDContext()` i.e., wrap r.Context() in a context.WithValue (or similar derived context) before initializing the struct for consistency?

pkg/gofr/external_db.go
Umang01-hash: Sure i have refactored this code.

pkg/gofr/datasource/dbresolver/options.go
```diff
@@ -0,0 +1,36 @@
...
+
+func WithPrimaryRoutes(routes map[string]bool) Option {
+	return func(r *Resolver) {
+		r.primaryRoutes = routes
+	}
+}
+
+func WithCircuitBreaker(maxFailures int32, timeoutSec int) Option {
+	return func(r *Resolver) {
+		timeout := time.Duration(timeoutSec) * time.Second
```
coolwednesday: Is this even being used ? As I saw we are adding the dbresolver config. Then where are we adding configurations through Options ?

pkg/gofr/datasource/dbresolver/resolver.go
```diff
@@ -0,0 +1,561 @@
...
+	buckets := []float64{0.05, 0.1, 0.2, 0.5, 1.0, 2.0, 5.0, 10.0}
+	r.metrics.NewHistogram("dbresolver_query_duration", "Response time of DB resolver operations in microseconds", buckets...)
+
+	// Gauges for operation tracking
+	r.metrics.NewGauge("dbresolver_primary_reads", "Total reads routed to primary")
+	r.metrics.NewGauge("dbresolver_primary_writes", "Total writes routed to primary")
+	r.metrics.NewGauge("dbresolver_replica_reads", "Total reads routed to replicas")
+	r.metrics.NewGauge("dbresolver_fallbacks", "Total fallbacks to primary")
+	r.metrics.NewGauge("dbresolver_failures", "Total replica failures")
+}
```
coolwednesday, line 142: Replica specific metrics

Umang01-hash: I have added the same in **recordStats**:

```go

func (r *Resolver) recordStats(start time.Time, method, target string, span trace.Span, isRead bool, replicaIndex *int) {
	duration := time.Since(start).Microseconds()

	// Update trace if available.
	if span != nil {
		defer span.End()

		attrs := []attribute.KeyValue{
			attribute.String("dbresolver.target", target),
			attribute.Int64("dbresolver.duration", duration),
			attribute.Bool("dbresolver.is_read", isRead),
		}

		// Add replica index if available
		if replicaIndex != nil {
			attrs = append(attrs, attribute.Int("dbresolver.replica_index", *replicaIndex))
		}
```

coolwednesday:
As reported by @Umang01-hash, here is a metric for other people's reference. We are good to go with this PR.

Before :
Screenshot 2025-11-21 at 3 12 57 PM

Last 10 min , requests Served : 
Screenshot 2025-11-21 at 3 15 05 PM

After : 
Screenshot 2025-11-21 at 3 12 42 PM
Last 10 min Requests, Served :
Screenshot 2025-11-21 at 3 15 15 PM

coolwednesday (dismissed):

coolwednesday (approved):

Issue #2000, opened 2025-07-03
RBAC(Role-Based Access Control) Support


goginenibhavani2000:
### Problem: 
GoFr lacks built-in role-based access control, forcing developers to implement custom authorization logic for each route. This leads to inconsistent security implementations and code duplication.
### Goal : 
Create a declarative RBAC middleware that integrates seamlessly with GoFr.
1. JSON Configuration: Define roles and permissions declaratively
2. Multiple Usage Patterns:
-       app.Use(rbac.Middleware(config))  // Global
-      app.GET("/admin", rbac.RequireRole("admin")(handler)) // Route-specific

3.  Flexible Role Extraction: Support JWT claims, sessions, and custom extractors
4. Pattern Matching: Handle wildcards (*) and path patterns (/users/*)
5. Helper Functions: IsAdmin(), HasRole(), GetUserRoles()

### Implementation Plan
- [ ] Models- 

// RoleConfig defines the roles and their allowed paths/permissions.
type RoleConfig struct {
	Roles map[string][]string `json:"roles"` // e.g., {"admin": ["*"], "editor": ["/edit", "/update"]}
}

// Options allow flexible configuration.
type Options struct {
	RoleExtractor func(*Context) ([]string, error) // customized function can be passed
	Config        RoleConfig
}

- [ ] func to  load RoleConfig (LoadRoleConfig(path string) (rbac.RoleConfig, error)) and validate config
- [ ] Implement func Middleware(opt Options) http.Handler  for global use
- [ ] Implement route-specific like RequireRole(role string) http.Handler
- [ ] isAuthorized function should support pattern match(match(pattern, path string) bool) for wildcard and prefix matching 
- [ ] Implement helper functions: IsAdmin(c *Context) bool , HasRole(c *Context, role string), GetUserRoles(c *Context) ([]string, error)

coolwednesday:
Sure Assigning it to you. However I would need other maintainers @aryanmehrotra & @Umang01-hash, opinion on your implementation design.

coolwednesday:
You can start working on it. Please ensure you implement an optional parameter design pattern to avoid this from becoming a breaking change.

Tag me here ! if you have any doubts.

goginenibhavani2000:
@coolwednesday Can you please check the draft PR for RBAC support- https://github.com/gofr-dev/gofr/pull/2049

Follow up Questions:
[...]

Issue #2114, opened 2025-08-01
Remote Config Support


akshat-kumar-singhal:
**Background**
Gofr support remote log level configuration. This can be extended to allow remote updating other configurations as well.

**Proposal**
A new interface which can be implemented by any receiver which supports reloading based on configuration at runtime. Remote Config
High level implementation as below:

```
// structs implementing RemoteConfigurable interface are responsible to update/reload the internal config and make it available in-place, i.e. not require restarting the application
type RemoteConfigurable interface {
  refreshConfig(config map[string]string) error
}

type RemoteConfiguration interface {
  registerReceiver(receiver RemoteConfigurable) bool
  fetchRemoteConfig() error
}

// TODO: this can be moved to a common package as it's used in other places like JWKS fetch as well
// TODO: better name can be given for this interface 
type HTTPGetProvider interface {
   GetWithHeaders(ctx context.Context, path string, queryParams map[string]any,
		headers map[string]string) (*http.Response, error)
}

// GofrRemoteConfig implements RemoteConfiguration interface 
type GofrRemoteConfig struct {
  provider HTTPGetProvider
  receivers RemoteConfigurable
// refreshFrequency configuration
}

```

aryanmehrotra:
what kind of configurations do we expect users would want to change at runtime? From my perspective, database details are unlikely to be updated dynamically, and even service URLs may not be something users would want to modify frequently.

Could you share some examples of configurations you think would be practical to update remotely? That might help us evaluate how useful and flexible this feature could be.

akshat-kumar-singhal:
We currently have log level being updated remotely. In case the developer wishes to update any other configuration on a runtime basis, the same can be done via this. 
Also, this implementation would reduce the coupling we currently have in the remote log level update with Remote Config service

aryanmehrotra:
I agree this refactor can definitely help reduce the current coupling in the remote log level update with the Remote Config service. That improvement makes sense.

At the same time, I feel it doesn’t necessarily need to expose developers to changing other configs. Having that flexibility comes with a few risks — for example:

**Pros**

1. Easier to roll out non-critical changes without restarts.
2. Useful for toggling log levels, feature flags, or rate limits.
3. Reduces operational overhead in cases where frequent minor updates are required.

**Cons**

1. Risk of inconsistent states across services if configs don’t sync properly.
2. Harder to debug when configs can silently change at runtime.
3. Security concerns if sensitive configs (like DB credentials) are updated this way.
[...]

PR #2144, opened 2025-08-11
RBAC middleware support

goginenibhavani2000:
**Description:**

-   Added a declarative RBAC middleware with JSON‑defined roles, flexible role extraction, pattern matching, and helper functions enables consistent, reusable, and easy‑to‑maintain authorization across routes.
-   Addresses https://github.com/gofr-dev/gofr/issues/2000
-   GoFr has no native RBAC, forcing manual, repetitive route‑level authorization that risks inconsistency and increases maintenance. Expected Benefits include centralized, configurable, and reusable access control that’s easier to maintain, more flexible, and improves security.

-   None

**Additional Information:**

-   No relevant dependencies effected or external libraries used.

**Checklist:**


coolwednesday (dismissed):
We would like to keep RBAC as a separate module. Hence add go.mod in it. So that it does not get added in the binary of the gofr framework.

examples/rbac/config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "roles": {
+      "admin": ["*"],
+      "editor": ["/posts/*", "/dashboard"],
+      "user": ["/profile", "/home", "/sayhello*","/greet"]
+    }
+  }
```
akshat-kumar-singhal: Roles should be defined at route level, not the other way around. That would allow clarity on who all can access a given route.

pkg/gofr/rbac/match.go
```diff
@@ -0,0 +1,45 @@
+package rbac
+
+import (
+	"path"
+	"strings"
+)
+
+func isPathAllowed(role, route string, config *Config) bool {
```
akshat-kumar-singhal: We shouldn't reinvent the path matching, utilise existing packages.

pkg/gofr/rbac/helper.go
```diff
@@ -0,0 +1,17 @@
+package rbac
+
+import "gofr.dev/pkg/gofr"
+
+func HasRole(ctx *gofr.Context, role string) bool {
+	expRole, _ := ctx.Context.Value(userRole).(string)
+	return expRole == role
+}
+
+func IsAdmin(ctx *gofr.Context) bool {
```
akshat-kumar-singhal: Roles should not be hardcoded into the framework. They are just strings, should not have a literal meaning defined in the framework.

pkg/gofr/rbac/helper.go
```diff
@@ -0,0 +1,17 @@
...
+	return HasRole(ctx, "admin")
+}
+
+func GetUserRole(ctx *gofr.Context) string {
```
akshat-kumar-singhal: Errors should not be silently ignored

pkg/gofr/rbac/helper_test.go
```diff
@@ -0,0 +1,67 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Create base context with the userRole value
+			baseCtx := context.WithValue(t.Context(), userRole, tt.ctxRoleVal)
+
+			// Wrap baseCtx in gofr.Context
+			gofrCtx := &gofr.Context{Context: baseCtx}
+
+			got := HasRole(gofrCtx, tt.checkRole)
+			if got != tt.expectedRes {
```
akshat-kumar-singhal: Consider using assert.Equal/Equalf to make it more concise

examples/rbac/main.go
```diff
@@ -0,0 +1,39 @@
...
+	app.UseMiddleware(rbac.Middleware(rbacConfigs))
+
+	app.GET("/sayhello/123", handler)
+	app.GET("/greet", rbac.RequireRole("user1", handler))
+
+	app.Run() // listens and serves on localhost:8000
+}
+
+func extractor(req *http.Request, _ ...any) (string, error) {
+	return req.Header.Get("X-USER-ROLE"), nil
```
akshat-kumar-singhal: User role passed in the header cannot be trusted as is. While this may have been added as an example only, it'd be better to have a proper example.

examples/rbac/main.go
```diff
@@ -0,0 +1,39 @@
...
+	overrides := map[string]bool{"user1": true}
+
+	rbacConfigs.OverRides = overrides
+
+	rbacConfigs.RoleExtractorFunc = extractor
+
+	app.UseMiddleware(rbac.Middleware(rbacConfigs))
+
+	app.GET("/sayhello/123", handler)
+	app.GET("/greet", rbac.RequireRole("user1", handler))
```
akshat-kumar-singhal: This is defeating the purpose of middleware if every route needs to add this RequireRole

akshat-kumar-singhal:
> We would like to keep RBAC as a separate module. Hence add go.mod in it. So that it does not get added in the binary of the gofr framework.

@coolwednesday What's the advantage of having RBAC in another module, unless we move the auth middlewares into a separate module as well?

Umang01-hash:
@goginenibhavani2000 please go ahead and resolve the review comments given by the reviewer. While we decide whether to keep it inside or outside like a separate module i think we can focus on completing the implementation, testing, documentation etc. Please let us know if you know any more help.

Umang01-hash:
> > We would like to keep RBAC as a separate module. Hence add go.mod in it. So that it does not get added in the binary of the gofr framework.
> 
> @coolwednesday What's the advantage of having RBAC in another module, unless we move the auth middlewares into a separate module as well?

@akshat-kumar-singhal  We see RBAC as a layer beneath Auth—more specialized and not universally required. Auth is core to GoFr and widely used, which is why its middlewares are bundled. RBAC, on the other hand, is optional and domain-specific, so we’re keeping it modular to avoid unnecessary coupling and binary weight.

akshat-kumar-singhal:
Won’t it be taken care of by the dead code elimination done during build?
I agree that RBAC shouldn’t be a default, but an easy to include
module/package. What we could consider is having the interface for RBAC
within gofr and the external module implementing that interface.

Umang01-hash:
> taken care of by the dead code elimination done during build? I agree that RBAC shouldn’t be a default, but an easy to include module/package. What we could consider is having the interface for RBAC

Yes that makes sense. WE can have the interface in GoFr and implementation outside.

pkg/gofr/rbac/helper.go
goginenibhavani2000: it returns a bool, so have ignored it

examples/rbac/main.go
goginenibhavani2000: requireRole is not needed for every route, can be directly used as app.GET("/sayhello/123", handler)

Umang01-hash:
@goginenibhavani2000 Are you still working on the issue. Please let us know in case you need any further help for the same.

goginenibhavani2000:
> @goginenibhavani2000 Are you still working on the issue. Please let us know in case you need any further help for the same.

Yes, i have completed the suggested changes except the change in having interface. Would like to confirm if the RBAC interface is like below

type RBAC interface {
    RoleExtractor(r *http.Request, args ...any) (string, error)
    IsPathAllowed(role string, path string) bool
}

coolwednesday:
@goginenibhavani2000, So Sorry for the delay. I will be prioritising this issue review for the next week's release. Let me know if you are interested to tale this further if not, I myself will pick up this issue.

goginenibhavani2000:
@coolwednesday I’d be happy to take this further and help with the changes. Let me know if there's anything specific you'd like me to focus on.

go.mod
```diff
@@ -107,3 +107,5 @@ require (
 	modernc.org/mathutil v1.7.1 // indirect
 	modernc.org/memory v1.11.0 // indirect
 )
+
+replace gofr.dev/pkg/gofr/rbac => ./pkg/gofr/rbac
```
Umang01-hash: This can be removed after initial rounds of testing right??

pkg/gofr/auth.go
```diff
@@ -102,3 +108,294 @@ func (a *App) EnableOAuth(jwksEndpoint string,
 
 	a.httpServer.router.Use(middleware.OAuth(middleware.NewOAuth(oauthOption), options...))
 }
+
```
Umang01-hash: Should we create a separate file `rbac.go` that contains all these RBAC related new code, so our `auth.go` remains focused on existing auth mechanisms ??

pkg/gofr/rbac/middleware_test.go
```diff
@@ -0,0 +1,557 @@
...
+				value: func(key any) any {
+					if key == userRole {
+						return tc.contextRole
+					}
+					return nil
+				},
+			}
+			resp, err := wrappedHandler(ctx)
+
+			if tc.wantErr != nil {
```
Umang01-hash: Tests should not have nested if-else structure like this. If needed we can break them into smaller tests.

pkg/gofr/rbac/middleware_test.go
```diff
@@ -0,0 +1,557 @@
...
```
Umang01-hash: Tests should not have nested if-else structure like this. If needed we can break them into smaller tests.

pkg/gofr/rbac/permissions_test.go
```diff
@@ -0,0 +1,420 @@
...
+				value: func(key any) any {
+					if key == userRole {
+						return tt.ctxRole
+					}
+					return nil
+				},
+			}
+
+			result, err := wrapped(ctx)
+			if tt.wantErr {
```
Umang01-hash: Tests should not have nested if-else structure like this. If needed we can break them into smaller tests.

pkg/gofr/rbac/providers/jwt.go
```diff
@@ -0,0 +1,193 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/golang-jwt/jwt/v5"
+	"gofr.dev/pkg/gofr/http/middleware"
+)
+
+var (
+	// ErrJWTNotEnabled is returned when JWT role extraction is attempted but OAuth is not enabled.
+	ErrJWTNotEnabled = errors.New("JWT/OAuth middleware not enabled")
```
Umang01-hash: Why are these errors exported? If we intend to use all these errors outside of directory have we documented it??

pkg/gofr/rbac/permissions_test.go
```diff
@@ -0,0 +1,420 @@
...
+			testConfig := config
+			if tt.name == "No match and no default" {
+				testConfig = &PermissionConfig{
+					RoutePermissionMap: config.RoutePermissionMap,
+					// No DefaultPermission
+				}
+			}
+
+			got, err := GetRequiredPermission(tt.method, tt.route, testConfig)
+			if tt.wantErr {
```
Umang01-hash: Tests should not have nested if-else structure like this. If needed we can break them into smaller tests.

pkg/gofr/rbac/hierarchy_test.go
```diff
@@ -0,0 +1,362 @@
...
+			route:   "/api/users",
+			want:    false,
+			useHier: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var got bool
+			if tt.useHier {
```
Umang01-hash: Tests should not have nested if-else structure like this. If needed we can break them into smaller tests.

pkg/gofr/rbac/config_test.go
```diff
@@ -0,0 +1,396 @@
...
+			require.NoError(t, err)
+
+			defer os.Remove(tempFile.Name())
+
+			_, err = tempFile.WriteString(tt.content)
+			require.NoError(t, err)
+			tempFile.Close()
+
+			cfg, err := LoadPermissions(tempFile.Name())
+			if tt.wantErr {
```
Umang01-hash: Tests should not have nested if-else structure like this. If needed we can break them into smaller tests.

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+
+func main() {
+	app := gofr.New()
+
+	// Load base RBAC configuration
+	config, err := rbac.LoadPermissions("configs/rbac.json")
+	if err != nil {
+		app.Logger().Error("Failed to load RBAC config: ", err)
+		return
+	}
```
aryanmehrotra: we can do this internally by just taking the file path and also have a default file path so user need to give path unless he gives something different

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+
+	// Configure permission-based access control
+	config.PermissionConfig = &rbac.PermissionConfig{
+		Permissions: map[string][]string{
+			"users:read":   {"admin", "editor", "viewer"},
+			"users:write":  {"admin", "editor"},
+			"users:delete": {"admin"},
+			"posts:read":   {"admin", "author", "viewer"},
+			"posts:write":  {"admin", "author"},
```
aryanmehrotra: example doesnt seem correct, why author??
also should this not be 

admin : users:read, users:write etc etc

to figure out what an admin can do need to look in multiple places

need to think this through what can be a better way

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+
+	// Enable RBAC with permissions
+	app.EnableRBAC(
+		gofr.WithConfig(config),
+		gofr.WithRoleExtractor(func(req *http.Request, args ...any) (string, error) {
+			role := req.Header.Get("X-User-Role")
+			if role == "" {
+				return "", fmt.Errorf("role header not found")
+			}
+			return role, nil
```
aryanmehrotra: why not just take the key? rather than giving user access the complete request? should the role come from anywhere apart from X-User-Role

Also is there any standardised guide which we are following for rbac support? if so we should mention that

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+}
+
+// Check role-based access (if not already authorized by permissions)
+if !authorized {
+    if isRoleAllowed(role, route, config) {
+        authorized = true
+        authReason = "role-based"
+    }
+}
+```
```
aryanmehrotra: can the user make any changes to it?? if cannot what benefit do we get by keeping this in documentation? how is it beneficial to the users of the framework?

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+
+### Permission Flow
+
+1. **Request arrives** → Middleware extracts user role
+2. **Permission check** → Checks if route has a permission mapping
+3. **Role validation** → Verifies if user's role has the required permission
+4. **Fallback** → If no permission mapping exists, falls back to role-based check
```
aryanmehrotra: what is the benefit of having the fallback? can't we already do the role-based check in just single operation?

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+```go
+config.PermissionConfig = &rbac.PermissionConfig{
+    Permissions: map[string][]string{
+        // Permission: [roles that have this permission]
+        "users:read":   {"admin", "editor", "viewer"},
+        "users:write":  {"admin", "editor"},
+        "users:delete": {"admin"},
+        "posts:read":   {"admin", "author", "viewer"},
+        "posts:write":  {"admin", "author"},
+        "posts:delete": {"admin"},
```
aryanmehrotra: same issue as above

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+
+You can also define permissions in the JSON config file:
+
+```json
+{
+  "permissions": {
+    "permissions": {
+      "users:read": ["admin", "editor", "viewer"],
+      "users:write": ["admin", "editor"],
+      "users:delete": ["admin"]
```
aryanmehrotra: same would apply here

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+    },
+    "routePermissions": {
+      "GET /api/users": "users:read",
+      "POST /api/users": "users:write",
+      "DELETE /api/users": "users:delete"
+    }
+  },
+  "enablePermissions": true
```
aryanmehrotra: why this? it is by default if this config is given this will be used

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+
+### Wildcard Support
+
+Route patterns support wildcards:
+
+```go
+RoutePermissionMap: map[string]string{
+    "GET /api/users/*":    "users:read",   // All GET requests under /api/users/
+    "POST /api/*":         "admin:write",  // All POST requests under /api/
+    "* /api/admin/*":      "admin:all",    // All methods under /api/admin/
```
aryanmehrotra: can there be a better way to provide route METHOD and route itself? 

example
```
[
  {
    "methods": ["GET"],
    "regex": "^/api/users/.*$",
    "permission": "users:read"
  }
]

```

we can provide multiple methods as well, we can take a specific struct as well

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+
+## Handler-Level Permission Checks
+
+### Using `gofr.RequirePermission()`
+
+For explicit permission checks in handlers:
+
+```go
+app.DELETE("/api/users/:id", gofr.RequirePermission("users:delete", config.PermissionConfig, deleteUser))
+```
```
aryanmehrotra: we should not be checking this in handler, how is this even possible? why to pollute this, its an addon

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+```json
+{
+  "route": {
+    "/api/*": ["admin", "editor"]
+  },
+  "overrides": {
+    "/health": true,        // Public routes
+    "/metrics": true
+  },
+  "defaultRole": "viewer",  // Default when role extraction fails
```
aryanmehrotra: documentation needs to be updated for this as this can also result in security flaw

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+  "enablePermissions": true  // Enable permission checks
+}
+```
+
+#### 3. **Configuration Management Benefits**
+
+- **Environment Overrides**: Supports `RBAC_ROUTE_*` environment variables
+- **Separation of Concerns**: Route-level config in file, fine-grained permissions in code
+- **Version Control**: Config changes tracked separately from code
+
```
aryanmehrotra: what's this?

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+```go
+app.EnableRBAC(
+	gofr.WithConfig(config),
+	gofr.WithRoleExtractor(func(req *http.Request, args ...any) (string, error) {
+		role := req.Header.Get("X-User-Role")
+		if role == "" {
+			return "", fmt.Errorf("role header not found")
+		}
+		return role, nil
+	}),
```
aryanmehrotra: we should not give access to req etc, we can just take the header key name

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+```go
+config.RoleExtractorFunc = func(req *http.Request, args ...any) (string, error) {
+    userID := req.Header.Get("X-User-ID")
+    if userID == "" {
+        return "", fmt.Errorf("user ID not found")
+    }
+
+    // Get container from args (automatically injected for database-based extraction)
+    if len(args) > 0 {
+        if cntr, ok := args[0].(*container.Container); ok && cntr != nil && cntr.SQL != nil {
```
aryanmehrotra: container should not be accessed from middleware, also put a recommended way how users can still dynamically change user permission if there auth service is implemented separately

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+```json
+{
+  "route": {
+    "/api/*": ["admin", "editor"],
+    "*": ["viewer"]
+  }
+}
+```
+
+This ensures routes without explicit permission mappings are still protected.
```
aryanmehrotra: example would be needed in docs

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+
+Write integration tests to verify permission checks:
+
+```go
+func TestPermissionChecks(t *testing.T) {
+    // Test that admin can delete users
+    // Test that editor cannot delete users
+    // Test that viewer cannot write users
+}
+```
```
aryanmehrotra: this should be part of example

docs/advanced-guide/rbac-permissions/page.md
```diff
@@ -0,0 +1,663 @@
...
+    "GET /api/users":      "users:read",
+    "PUT /api/users/:id":  "users:update",
+    "DELETE /api/users/:id": "users:delete",
+},
+```
+
+### Pattern 2: Hierarchical Permissions
+
+```go
+Permissions: map[string][]string{
```
aryanmehrotra: this is getting confusing we should have one way to do this, multiple ways is not giving much benefit

docs/advanced-guide/rbac/page.md
```diff
@@ -0,0 +1,728 @@
...
+	)
+
+	app.GET("/api/users", handler)
+	app.Run()
+}
+```
+
+> **Note**: The RBAC module uses automatic registration via package-level variable initialization (not `init()`). When you import `gofr.dev/pkg/gofr/rbac`, the RBAC implementations are automatically registered with core GoFr. This allows the framework to remain lightweight - RBAC is only included when explicitly imported.
+
+> **⚠️ Security Note**: Header-based RBAC is **not secure** for public APIs. Use JWT-based RBAC for production applications.
```
aryanmehrotra: do we really need this? 
app.EnableRBAC is the primary command which should ideally take care of things, i.e we can take an interface as an Input

docs/advanced-guide/rbac/page.md
```diff
@@ -0,0 +1,728 @@
...
+app.EnableRBAC(
+	gofr.WithPermissionsFile("configs/rbac.json"),
+	gofr.WithJWT("role"),
+)
+// JWT: {"role": "admin", "sub": "user123"}
+
+// Array notation - extract first role
+app.EnableRBAC(
+	gofr.WithPermissionsFile("configs/rbac.json"),
+	gofr.WithJWT("roles[0]"),
```
aryanmehrotra: user role would either be one or many ideally based on implementation, in which case would the user want to get only one role? instead of all and then validate, if required role is there or not.

docs/advanced-guide/rbac/page.md
```diff
@@ -0,0 +1,728 @@
...
+```
+
+## Comparison Matrix
+
+| Feature | Simple | JWT | Permissions-Header | Permissions-JWT | Permissions-DB |
+|---------|--------|-----|-------------------|----------------|----------------|
+| **Security** | ⚠️ Low | ✅ High | ⚠️ Low | ✅ High | ✅ High |
+| **Flexibility** | ⚠️ Low | ⚠️ Low | ✅ High | ✅ High | ✅✅ Very High |
+| **Performance** | ✅ Fast | ✅ Fast | ✅ Fast | ✅ Fast | ⚠️ Slower* |
+| **Dynamic Roles** | ❌ No | ❌ No | ❌ No | ❌ No | ✅ Yes |
```
aryanmehrotra: even without giving db access dynamic roles can still be set, depends on how its implmented

pkg/gofr/rbac.go
```diff
@@ -0,0 +1,301 @@
...
+
+	// PermissionConfig enables permission-based access control
+	PermissionConfig PermissionConfig
+
+	// RequiresContainer indicates if container access is needed for role extraction
+	RequiresContainer bool
+}
+
+// RBACOption is a function that configures RBACOptions.
+type RBACOption func(*RBACOptions)
```
aryanmehrotra: why not use an interface, similar to options in http service

pkg/gofr/rbac/middleware.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	status := "denied"
+	if allowed {
+		status = "allowed"
+	}
+
+	logger.Infof("[RBAC Audit] %s %s - Role: %s - Route: %s - %s - Reason: %s",
+		req.Method, req.URL.Path, role, route, status, reason)
+}
```
aryanmehrotra: why not just keep it by default? without creating this function, what benefit it gives inspite of not logging? and should we not by default log it may be debug level?

pkg/gofr/rbac/adapter.go
```diff
@@ -0,0 +1,102 @@
...
+		ErrAccessDenied,          // ErrAccessDenied
+		ErrPermissionDenied,      // ErrPermissionDenied
+	)
+
+	return true
+}
+
+// _ is a package-level variable that triggers registration when the package is imported.
+// This avoids using init() while maintaining automatic registration behavior.
+var _ = registerRBAC()
```
aryanmehrotra: there is no need of this

aryanmehrotra (changes requested):
code needs to be simplified, once the structural fixes are done code needs to be reviewed again.

pkg/gofr/rbac/config.go
akshat-kumar-singhal, line 1: @coolwednesday Consider extracting the "Hot Reload" functionality into a separate package. This functionality appears to be similar to Remote Config update discussed [here](https://github.com/gofr-dev/gofr/issues/2114).

pkg/gofr/rbac/middleware.go
akshat-kumar-singhal, line 1: We should think of the RBAC support as an extension to the auth middleware instead of creating a new middleware. That way we'd be able to have RBAC for non-JWT use cases as well.

pkg/gofr/rbac/config.go
coolwednesday: Sure. Will remove it for now, till we merge the Remote Config implementation.

pkg/gofr/rbac/middleware.go
coolwednesday: We currently do it for non-JWT that is header based case. We can extend for other's auth mechanisms also in the rbac module.

pkg/gofr/rbac/metrics.go
```diff
@@ -0,0 +1,17 @@
+package rbac
+
+import "context"
+
+// Metrics defines methods for recording metrics that RBAC uses.
+// It matches GoFr's container.Metrics interface.
```
akshat-kumar-singhal: in case we don't want to import GoFr metrics package, let's update the comments

pkg/gofr/rbac/endpoint_matcher.go
akshat-kumar-singhal, line 1: can we consider using mux package instead of re-writing the route matching logic?

akshat-kumar-singhal, line 1: @coolwednesday Can we consider using mux package instead of re-writing the route matching logic?

pkg/gofr/rbac.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"net/http"
+
+	"go.opentelemetry.io/otel"
+)
+
+// RBACProvider is the interface for RBAC implementations.
+// External RBAC modules (like gofr.dev/pkg/gofr/rbac) implement this interface.
+type RBACProvider interface {
```
akshat-kumar-singhal: Could we look at having a MiddlewareProvider type of interface instead of making it specific for RBAC? The implementation feels repetitive

pkg/gofr/rbac/provider.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// NewProvider creates a new RBAC provider with the config file path.
+// If configPath is empty, it will try default paths: configs/rbac.json, configs/rbac.yaml, configs/rbac.yml
+//
+// Example:
+//
+//	provider := rbac.NewProvider("configs/rbac.json")
+//	app.EnableRBAC(provider)
+func NewProvider(configPath string) *Provider {
```
akshat-kumar-singhal: This should have an error returned as well - in case the config is not valid

pkg/gofr/rbac/middleware.go
```diff
@@ -0,0 +1,415 @@
...
+	return func(handler http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// If config is nil, allow all requests (fail open)
+			if config == nil {
+				handler.ServeHTTP(w, r)
+
+				return
+			}
+
+			route := r.URL.Path
```
akshat-kumar-singhal: would `route` contain the actual path with values? We should be just sending the parameterised path to avoid increasing label cardinality as well as exposing sensitive information to external systems

pkg/gofr/rbac/middleware.go
```diff
@@ -0,0 +1,415 @@
...
+			// Store role in context and continue
+			ctx := context.WithValue(r.Context(), userRole, role)
+			handler.ServeHTTP(w, r.WithContext(ctx))
+		})
+	}
+}
+
+// startTracing starts tracing for the request if tracer is available.
+// The span is stored in the context and should be ended at the end of the middleware function.
+func startTracing(r *http.Request, config *Config, route string) *http.Request {
```
akshat-kumar-singhal: Should this implemented be in RBAC package or reuse existing implementation?

pkg/gofr/rbac.go
coolwednesday: We intentionally don’t add a generic “middleware provider” interface here. GoFr already exposes App.UseMiddleware for plugging in custom middleware, and EnableOAuth / EnableRBAC are App-level helpers that load and wire config in ways only the App can. RBAC stays provider-specific to avoid duplicating the existing middleware extension point.

pkg/gofr/rbac/endpoint_matcher.go
coolwednesday: We don’t reuse GoFr’s mux router because the router instance is an unexported/internal detail of the HTTP server in GoFr.  

mux’s matching model (notably PathPrefix and unanchored patterns) doesn’t align with RBAC’s needs for anchored, config-driven endpoint matches (incl. explicit regex and exact path templates). Using prefix semantics would over-grant access in common RBAC scenarios.

Keeping RBAC matching self-contained avoids a hard dependency on mux, preserves router-agnostic behavior, and prevents pulling routing internals into the auth layer.

coolwednesday: https://github.com/gofr-dev/gofr/pull/2144#discussion_r2626391694

coolwednesday: We will analyse Mux Router and consider writing GoFr's own router if needed. However that will become a breaking change in that case. Can do it in v2 only.

akshat-kumar-singhal: > We will analyse Mux Router and consider writing GoFr's own router if needed. However that will become a breaking change in that case. Can do it in v2 only.

As a developer using any framework, I'd expect the behaviour for route matching to be consistent across the handlers as well as middleware. Another way of visualising the issue with this mismatch is consider an implementation without middleware - instead every handler calls a common method for performing the RBAC auth. The handler would have the path matched as per mux only.

I do understand reservations around mux and the challenge in re-writing it right now due to the breaking change. 
Recommendation: Create a wrapper around mux in gofr and use that in this new package. This would ensure that:
- We don't introduce breaking change in the current version
- RBAC module is not tightly coupled with mux
- Route pattern matching is consistent across handlers and middleware  


cc @Umang01-hash @aryanmehrotra

docs/navigation.js
```diff
@@ -83,6 +83,11 @@ export const navigation = [
                 href: '/docs/advanced-guide/http-authentication',
                 desc: "Implement various HTTP authentication methods to secure your GoFR application and protect sensitive endpoints."
             },
+            {
+                title: 'Role-Based Access Control (RBAC)',
+                href: '/docs/advanced-guide/rbac',
+                desc: "Implement comprehensive Role-Based Access Control with support for roles, permissions, hierarchy, JWT integration, hot reloading, and fine-grained permission-based authorization."
```
akshat-kumar-singhal: We have removed hot reloading, let's update the description accordingly.

pkg/gofr/rbac/config.go
```diff
@@ -0,0 +1,432 @@
...
+	compiledRegexMap      map[string]*regexp.Regexp `json:"-" yaml:"-"` // Key: "METHOD:/path", Value: compiled regex
+
+	// Mutex for thread-safe access to maps (for future hot-reload support)
+	mu sync.RWMutex `json:"-" yaml:"-"`
+}
+
+// LoadPermissions loads RBAC configuration from a JSON or YAML file.
+// The file format is automatically detected based on the file extension.
+// Supported formats: .json, .yaml, .yml.
+func LoadPermissions(path string) (*Config, error) {
```
akshat-kumar-singhal: @coolwednesday should we look at using the file package in gofr instead of using os? It'll allow more flexibility to the developer as well as minimise our usage of os package directly

coolwednesday: We can’t rely on the common filesystem because GoFr supports only one filesystem per service. Using it inside RBAC would:

- Tie RBAC to whatever filesystem is registered
- Prevent users from registering another filesystem for other use cases
- Break RBAC if AddFileStore() is called later (it would replace the filesystem RBAC depends on)
- This would significantly restrict flexibility for users.

docs/advanced-guide/rbac/page.md
```diff
@@ -0,0 +1,565 @@
...
+	
+	// Extract role using the same claim path as configured in rbac.json
+	// Example: if jwtClaimPath is "role"
+	role, ok := claims["role"].(string)
+	if !ok {
+		return nil, errors.New("role not found in JWT claims")
+	}
+	
+	// Use role for business logic (e.g., personalize UI, filter data)
+	return map[string]string{"userRole": role}, nil
```
aryanmehrotra: this is not how gofr recommends let rather create a struct

aryanmehrotra (changes requested):

aryanmehrotra (dismissed):
Looks good to me just 
Handle errors in examples so that we dont promote returning 500 and use gofr provided error (in case of request-response cycles) and in middleware as well

coolwednesday:
Screenshot 2025-12-24 at 6 08 05 PM

Umang01-hash (approved):

Issue #2015, opened 2025-07-04
[GoFr-SOC]: Oracle DB


Samar1110:
### Problem: Oracle DB Integration

GoFr currently supports popular relational databases like MySQL and PostgreSQL, but lacks native support for Oracle DB, which is widely used in enterprise systems. To broaden compatibility and adoption in enterprise environments, we need to integrate Oracle as a first-class relational datasource within the GoFr framework.

### Goal:

Implement Oracle DB as a new relational datasource in `pkg/gofr/datasource/oracle`.
Add a method on the `App` to plug and use the Oracle datasource.

Include:

* **Logging**: Log query executions and connection errors with appropriate context.
* **Tracing**: Wrap DB calls in OpenTelemetry spans for distributed tracing.
* **Metrics**: Emit GoFr-compatible metrics for query counts, failures, and latency.

Samar1110:
I’m currently contributing to this as part of the GoFr Summer of Code 2025. Kindly assign this issue to me. Thank you!

coolwednesday:
Sure ! Make sure to make it an externalized module (has its own go.mod).

PR #2197, opened 2025-08-23
feat: Oracle db migration (#2015)

Samar1110:
## Pull Request: OracleDB Migration Support Integration

**Description:**

-   Adds OracleDB migration support in pkg/gofr/migration/oracle.go and associated tests in oracle_test.go.
-   Provides robust unit tests with mocks covering success and failure scenarios for migration table creation, fetching last migration, commit, and transaction begin.
-   Uses Oracle driver interface methods (Exec, Select) for migration SQL execution.
-   Addresses [#2015](https://github.com/gofr-dev/gofr/issues/2015).

**Details of the Change**

-   Added oracleDS and oracleMigrator structs implementing migration interface for Oracle.
-   Migration table DDL and queries specific for Oracle PL/SQL with exception handling.
-   Migration lifecycle functions: checkAndCreateMigrationTable, getLastMigration, beginTransaction, commitMigration, and rollback.
-   Unit tests simulating Oracle behavior via mocks, validating all migration lifecycle operations with error and success cases.

**Motivation & Benefits:**

-   Enables native OracleDB support in GoFr migration framework.
-   Ensures consistent migration behavior for Oracle alongside other supported databases.
-   Improves reliability of schema migrations when using OracleDB.

**Checklist:**


pkg/gofr/migration/interface.go
```diff
@@ -39,6 +39,13 @@ type Clickhouse interface {
 	HealthCheck(ctx context.Context) (any, error)
 }
 
+type Oracle interface {
+	Select(ctx context.Context, dest any, query string, args ...any) error
+	Exec(ctx context.Context, query string, args ...any) error
+
+	HealthCheck(ctx context.Context) (any, error)
```
Umang01-hash: Why is this method needed for migrations?

pkg/gofr/migration/migration_test.go
```diff
@@ -207,3 +208,130 @@ func initializeClickHouseRunMocks(t *testing.T) (*MockClickhouse, *container.Con
...
+
+	mockOracle.EXPECT().Exec(gomock.Any(), CheckAndCreateOracleMigrationTable).Return(nil)
+	mockOracle.EXPECT().Select(gomock.Any(), gomock.Any(), getLastOracleGoFrMigration).Return(nil)
+	mockOracle.EXPECT().Exec(gomock.Any(), "CREATE TABLE test (id INT)").Return(nil)
+	mockOracle.EXPECT().Exec(gomock.Any(), insertOracleGoFrMigrationRow, int64(1), "UP", gomock.Any(), gomock.Any()).Return(nil)
+
+	Run(migrationMap, mockContainer)
+}
+
+func TestOracleMigration_FailCreateMigrationTable(t *testing.T) {
```
Umang01-hash: The tests for Oracle Migration should be placed inside `oracle_test.go` not here.

pkg/gofr/migration/oracle.go
```diff
@@ -0,0 +1,110 @@
...
+// Provides a wrapper to apply the oracle migrator logic.
+func (od oracleDS) apply(m migrator) migrator {
+	return oracleMigrator{
+		Oracle:   od.Oracle,
+		migrator: m,
+	}
+}
+
+const (
+	CheckAndCreateOracleMigrationTable = `
```
Umang01-hash: Why is this constant exported? Can we please unexport it if not needed outside of the package.

pkg/gofr/migration/migration_test.go
```diff
@@ -207,3 +208,130 @@ func initializeClickHouseRunMocks(t *testing.T) (*MockClickhouse, *container.Con
```
Umang01-hash, line 208: In the method `initializeClickHouseRunMocks ` we need to mark **Oracle** as **nil** else it's tests will fail.

pkg/gofr/migration/migration_test.go
Samar1110: Do I need to shift only this test -> TestOracleMigration_FailCreateMigrationTable or all test related to Oracle migration in this file to oracle_test.go ?

pkg/gofr/migration/oracle.go
Samar1110: Sure I would do that

pkg/gofr/migration/migration_test.go
Samar1110: Sure , I will mark it as nil

pkg/gofr/migration/interface.go
Samar1110: Added the `HealthCheck(ctx context.Context) (any, error)` method to the `Oracle` interface in the migration package to align it with the `OracleDB` interface defined in the container package. Without this method, the `*MockOracle` generated for testing cannot be used as a `container.OracleDB` value, causing the compiler error:

"cannot use mockOracle (variable of type *MockOracle) as container.OracleDB value in assignment: *MockOracle does not implement container.OracleDB (missing method HealthCheck)".

As I am using

```go
mockContainer.Oracle = mockOracle
```

in some functions, it takes the reference from `pkg/gofr/container/datasource.go`, which requires the `HealthCheck` method for compatibility. To resolve this, I have added the `HealthCheck` method in `pkg/gofr/migration/interface.go`.

By adding the `HealthCheck` method, the mock satisfies the interface requirements, ensuring type compatibility and eliminating this assignment error.

pkg/gofr/migration/migration_test.go
Umang01-hash: You should place all test related to Oracle migration in this file to oracle_test.go

pkg/gofr/migration/interface.go
Umang01-hash: @Samar1110 I agree but if you notice many of the interfaces like SQL, Redis, Mongo, ArrangoDB etc all have HealthCheck() method not implemented yet they align with their interface definition in container.

You can see how their test have been written or implemented.

Umang01-hash:
@Samar1110 Please resolve the merge conflicts in your PR. You are near to completing this PR. Let's close it soon and if you need any help please let us know.

Samar1110:
> @Samar1110 Please resolve the merge conflicts in your PR. You are near to completing this PR. Let's close it soon and if you need any help please let us know.

Sure @Umang01-hash , by today I would complete the work

Samar1110:
@Umang01-hash could you please let me know if there are any additional changes or improvements I should make so that the branch is fully ready for merge.

coolwednesday (changes requested):
@Samar1110,  Hey Samar!
The PR looks good overall 👍. Just one small detail: since OracleDB supports transactions, it should behave like SQL migrations instead of the NoSQL migrations you used.

Once you make the changes, please attach screenshots covering these cases:

- When migrations run successfully
- When migrations rollback (using transactions)
- When migrations are skipped because they’ve already been applied

Samar1110:
> @Samar1110, Hey Samar! The PR looks good overall 👍. Just one small detail: since OracleDB supports transactions, it should behave like SQL migrations instead of the NoSQL migrations you used.
> 
> Once you make the changes, please attach screenshots covering these cases:
> 
> * When migrations run successfully
> * When migrations rollback (using transactions)
> * When migrations are skipped because they’ve already been applied

@coolwednesday @Umang01-hash Apologies for the delayed response, I was caught up with office work over the last two weeks. Many thanks to you @Umang01-hash , for completing my work. Please find the attached screenshots below. Once again, sorry for the late response.
Screenshot 2025-09-20 175753
Screenshot 2025-09-20 175749

coolwednesday (changes requested):
The rollback does not work for transactions. Also the PR gives an error of failed to fetch last migration, `dest must be *[]map[string]any` which I solved by using the following code snippet :

```
// Get the last applied migration version.
func (om oracleMigrator) getLastMigration(c *container.Container) int64 {
	var results []map[string]any
	var lastMigration int64

	err := om.Oracle.Select(context.Background(), &results, getLastOracleGoFrMigration)
	if err != nil {
		c.Errorf("Failed to fetch last migration: %v", err)

		return 0
	}

	if len(results) > 0 {
		// Handle Oracle number type conversion
		if lastMigVal, ok := results[0]["LAST_MIGRATION"]; ok {
			switch v := lastMigVal.(type) {
			case float64:
				lastMigration = int64(v)
			case int64:
				lastMigration = v
			case int:
				lastMigration = int64(v)
			default:
				// Handle godror.Number type by converting to string then parsing
				if str := fmt.Sprintf("%v", v); str != "" && str != "<nil>" {
					if parsed, parseErr := strconv.ParseInt(str, 10, 64); parseErr == nil {
						lastMigration = parsed
					}
				}
			}
		}
	}

	c.Debugf("Oracle last migration fetched value is: %v", lastMigration)

	lm2 := om.migrator.getLastMigration(c)

	if lm2 > lastMigration {
		return lm2
	}

	return lastMigration
}

```

Also can we also improve Connect method to handle the dsn effectively ? Will that require changing it in method ? or should we expose it as a configurable measure for people to add it themselves and update the same in docs.

coolwednesday (changes requested):
The rollback works. However in case of rollback we would like to exit the application. Hence use Fatal log. Also below there are redundant logs, please address them :

Screenshot 2025-09-24 at 12 01 19 PM

Umang01-hash (approved):

PR #2234, opened 2025-09-04
Update workflow to handle panic in pipeline

coolwednesday:


aryanmehrotra:
Handle panic as in?? can you update the description what we are actually trying to do here?

coolwednesday:
> Handle panic as in?? can you update the description what we are actually trying to do here?

Done.

Umang01-hash (approved):

PR #2240, opened 2025-09-05
Fix config Overloading & its tests

coolwednesday:
- Currently, we are using Overload for overriding. This causes the system environment variables to be overwritten. To handle this, we now run a manual loop that skips all environment variables initially loaded from the system.

### DIFF
--- a/pkg/gofr/config/godotenv.go
+++ b/pkg/gofr/config/godotenv.go
@@ -34,42 +34,96 @@ func NewEnvFile(configFolder string, logger logger) Config {
 }
 
 func (e *EnvLoader) read(folder string) {
-	var (
-		defaultFile  = folder + defaultFileName
-		overrideFile = folder + defaultOverrideFileName
-		env          = e.Get("APP_ENV")
-	)
+	// Capture initial system environment keys
+	initialEnv := e.captureInitialEnv()
 
-	err := godotenv.Load(defaultFile)
-	if err != nil {
-		if !errors.Is(err, fs.ErrNotExist) {
-			e.logger.Fatalf("Failed to load config from file: %v, Err: %v", defaultFile, err)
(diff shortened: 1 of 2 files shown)


Umang01-hash:
@coolwednesday I feel the code can be refactored into a simpler implementation. The three file loading functions (loadBaseEnvFile, loadLocalOverrideFile, loadEnvSpecificFile) contain significant duplication - consider creating a generic loadEnvFile function that handles common loading logic and error handling.

pkg/gofr/config/godotenv.go
```diff
@@ -40,6 +40,9 @@ func (e *EnvLoader) read(folder string) {
 		env          = e.Get("APP_ENV")
 	)
 
+	// Capture initial system environment before any file loading
+	initialEnv := e.captureInitialEnv()
+
```
aryanmehrotra: we were doing this before, we have added a new code we should remove either of them, also can we explain what exactly cause this issue, as the code which was in the package and this code is also same.

Also can you add comment when we load the env variables, is it set in the application context or the system configs

coolwednesday: So we are only loading the system variables first . Using that map to skip overriding during applying the .env file changes and then loading that variables in our process's execution at the very last.

coolwednesday: Also os.environ gives the current process specifications which is going to remain same for all starting processes until we don't export some new specifications in our environment at the system level.

Umang01-hash (approved):

aryanmehrotra (approved):

coolwednesday:
@aryanmehrotra, Here is the running tested server :
Screenshot 2025-09-08 at 4 30 49 PM

PR #2243, opened 2025-09-08
Adds test for http auth middleware

gizmo-rt:
## Pull Request Template

**Description:**

Adds test for http auth middleware example

-   List any breaking changes introduced by this PR. N/A

**Additional Information:**

**Checklist:**

### DIFF
--- a/examples/using-http-auth-middleware/main_test.go
+++ b/examples/using-http-auth-middleware/main_test.go
@@ -1,0 +1,154 @@
+package main
+
+import (
+	"encoding/base64"
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"gofr.dev/pkg/gofr"
+	"net/http"
+	"os"
+	"testing"
+	"time"
+)
+
+func Test_main(t *testing.T) {
+	go main()
+}
+
+func Test_setupBasicAuthSuccess(t *testing.T) {
+	os.Setenv("METRICS_PORT", "2200")
+	os.Setenv("HTTP_PORT", "8200")
+	app := gofr.New()
+	setupBasicAuth(app)
+	app.GET("/basic-auth-success", func(_ *gofr.Context) (any, error) {
+		return "success", nil
+	})
+	go app.Run()
+	time.Sleep(100 * time.Millisecond)
+
+	var netClient = &http.Client{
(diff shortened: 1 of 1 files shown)


examples/using-http-auth-middleware/main_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"time"
+)
+
+func Test_main(t *testing.T) {
+	go main()
+}
+
+func Test_setupBasicAuthSuccess(t *testing.T) {
+	os.Setenv("METRICS_PORT", "2200")
+	os.Setenv("HTTP_PORT", "8200")
```
coolwednesday: We can use `NewServerConfigs` to get a free port. Since numerous tests are integration test it can lead to port blocking issues.

coolwednesday (changes requested):

examples/using-http-auth-middleware/main_test.go
gizmo-rt: Thanks for sharing. Updated

gizmo-rt:
@Umang01-hash The commented example seems like an alternative method to  use the auth middleware. 
Either we can keep one of those examples or decouple and include both.  Thoughts ?

Umang01-hash:
@gizmo-rt what i meant to say was we should add tests that directly cover the **main()** function. You don't need to uncomment the alternatives - instead, add a test like this:

```
func Test_MainEndToEnd(t *testing.T) {
    serverConfigs := testutil.NewServerConfigs(t)
    
    // Run main() in a goroutine to avoid blocking
    go main()
    
    // Allow time for server to start
    
    client := &http.Client{Timeout: 200 * time.Millisecond}
    
    // Test valid API key
    t.Run("Valid API Key", func(t *testing.T) {
        req, _ := http.NewRequestWithContext(t.Context(), http.MethodGet,
            serverConfigs.HTTPHost+"/test-auth", http.NoBody)
        req.Header.Set("X-Api-Key", "valid-api-key")
        
        resp, err := client.Do(req)
        if err != nil {
            t.Fatalf("Error making request: %v", err)
        }
        defer resp.Body.Close()
        
        assert.Equal(t, http.StatusOK, resp.StatusCode)
    })

}
```

PR #2263, opened 2025-09-12
En/rate limiter

Umang01-hash:
## Add Rate Limiting Support for HTTP Services

This PR adds rate limiting capabilities to # Add Rate Limiting Support for HTTP Services

This PR adds rate limiting capabilities to GoFr's HTTP service clients using the token bucket algorithm. It provides both local (in-memory) and distributed (Redis-based) implementations to suit different deployment scenarios.

## Features

- **Two Implementation Strategies**:
  - **Local (In-Memory)**: For single-instance deployments
  - **Distributed (Redis)**: For multi-instance production deployments using atomic Lua scripts

- **Core Features**:
  - Configurable requests per second (RPS) and burst limits
  - Custom service key extraction support

## Usage Example

```go
// For distributed rate limiting (multi-instance)
a.AddHTTPService("payment-api", "http://localhost:9005",
    	&service.RateLimiterConfig{
			Requests: 5,
			Window:   time.Minute,
			Burst:    10,
			KeyFunc: func(req *http.Request) string {
				if req == nil {
					return "unknown"
				}

				return req.URL.Host // Use path as key
			},
			Store: service.NewRedisRateLimiterStore(rc),
		},
)

// For local rate limiting (single instance)
a.AddHTTPService("payment-api", "http://localhost:9005",
    &service.RateLimiterConfig{
        RequestsPerSecond: 1,    // 1 request per second
        Burst:            3,     // Allow bursts up to 3 requests
    },
) 
```

## Usage Screenshots:

- **Logs**
Screenshot 2025-09-12 at 11 17 31 AM

**Checklist:**

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -12,7 +12,7 @@ COPY  docs/navigation.js /app/src/lib
 COPY  docs/events.json /app/src/app/events
 COPY  docs/testimonials.json /app/utils
 
-ENV NODE_ENV production
+ENV NODE_ENV=production
 
 RUN npm install
 RUN npm run build


coolwednesday (approved):

akshat-kumar-singhal:
@Umang01-hash We should consider having support for generic time window, i.e. not limit to per second. Most systems have per minute/hour limits.

Umang01-hash:
@akshat-kumar-singhal sure fixed the same 

Users can use it like:
```
// Per-second rate limiting
app.AddHTTPService("fast-api", "https://fast.api.com",
    &service.RateLimiterConfig{
        Requests: 10,           // 10 requests per second
        Window:   time.Second,  // Explicit time window
        Burst:    15,
    },
)

// Per-hour rate limiting
app.AddHTTPService("batch-api", "https://batch.api.com",
    &service.RateLimiterConfig{
        Requests: 1000,        // 1000 requests per hour
        Window:   time.Hour,   // Hourly window
        Burst:    1200,
    },
)
```

pkg/gofr/service/logger.go
```diff
@@ -8,6 +8,7 @@ import (
 
 type Logger interface {
 	Log(args ...any)
+	Debug(args ...any)
```
akshat-kumar-singhal: Altering an exported interface is a breaking change

pkg/gofr/service/rate_limiter_distributed.go
```diff
@@ -0,0 +1,288 @@
...
+	logger      Logger
+	metrics     Metrics
+	HTTP
+}
+
+func NewDistributedRateLimiter(config RateLimiterConfig, h HTTP) HTTP {
+	httpSvc := h.(*httpService)
+
+	rl := &distributedRateLimiter{
+		config:      config,
```
akshat-kumar-singhal: this config should be validated

pkg/gofr/service/rate_limiter.go
```diff
@@ -0,0 +1,119 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	gofrRedis "gofr.dev/pkg/gofr/datasource/redis"
+)
+
+var (
+	errInvalidRequestRate     = errors.New("requestsPerSecond must be greater than 0")
+	errInvalidBurstSize       = errors.New("burst must be greater than 0")
```
akshat-kumar-singhal: Additional validation: Burst size should be greater than requests per unit time

pkg/gofr/service/rate_limiter.go
```diff
@@ -0,0 +1,119 @@
...
+	errInvalidRedisResultType = errors.New("unexpected Redis result type")
+)
+
+// RateLimiterConfig with custom keying support.
+type RateLimiterConfig struct {
+	Requests    float64                    // Number of requests allowed
+	Window      time.Duration              // Time window (e.g., time.Minute, time.Hour)
+	Burst       int                        // Maximum burst capacity (must be > 0)
```
akshat-kumar-singhal: Burst can be optional - set equal to `Requests` in case not configured/zero.

pkg/gofr/service/rate_limiter.go
```diff
@@ -0,0 +1,119 @@
...
+	"errors"
+	"fmt"
+	"net/http"
+	"time"
+
+	gofrRedis "gofr.dev/pkg/gofr/datasource/redis"
+)
+
+var (
+	errInvalidRequestRate     = errors.New("requestsPerSecond must be greater than 0")
```
akshat-kumar-singhal: Error message should be updated as we are supporting other time windows as well

pkg/gofr/service/rate_limiter_distributed.go
```diff
@@ -0,0 +1,288 @@
...
+// DistributedRateLimiter implements Redis-based distributed rate limiting using Token Bucket algorithm.
+// Strategy: Token Bucket with Redis Lua scripts for atomic operations
+// - Suitable for: Multi-instance production deployments
+// - Benefits: True distributed limiting across all service instances
+// - Performance: Single Redis call per rate limit check with atomic Lua execution
+
+// distributedRateLimiter with metrics support.
+type distributedRateLimiter struct {
+	config      RateLimiterConfig
+	redisClient *gofrRedis.Redis
```
akshat-kumar-singhal: Should this be a "cache interface" or "rate limiter bucket" instance instead of specifically Redis? That way the implementation difference between Redis/Local would just be at the storage level and not in the core functionality

pkg/gofr/service/logger.go
Umang01-hash: This Logger interface is the subset of our original Logger interface:

```
type Logger interface {
	Debug(args ...any)
	Debugf(format string, args ...any)
	Log(args ...any)
	Logf(format string, args ...any)
	Info(args ...any)
	Infof(format string, args ...any)
	Notice(args ...any)
	Noticef(format string, args ...any)
	Warn(args ...any)
	Warnf(format string, args ...any)
	Error(args ...any)
	Errorf(format string, args ...any)
	Fatal(args ...any)
	Fatalf(format string, args ...any)
	ChangeLevel(level Level)
}
```

We have limited method in it just to make it lean. Adding any method here will not be a breaking change untill and unless we alter the original logger interface. It is present here just to inject **app.Logger()**

pkg/gofr/service/rate_limiter_distributed.go
Umang01-hash: The RateLimiterConfig is already validated in the **AddOption** method before NewDistributedRateLimiter is called:

```

// AddOption implements the Options interface.
func (config *RateLimiterConfig) AddOption(h HTTP) HTTP {
	
	if err := config.Validate(); err != nil {
		if httpSvc, ok := h.(*httpService); ok {
			httpSvc.Logger.Log("Invalid rate limiter config, disabling rate limiting", "error", err)
		}

		return h
	}

	// .........
}

```

pkg/gofr/service/rate_limiter.go
Umang01-hash: Sure added

pkg/gofr/service/rate_limiter.go
Umang01-hash: Done made it optional and equal to Requests in case not configured/zero.

pkg/gofr/service/rate_limiter.go
Umang01-hash: updated.

pkg/gofr/service/rate_limiter_distributed.go
Umang01-hash: I've implemented a RateLimiterStore interface:

```
// RateLimiterStore abstracts the storage for rate limiter buckets.
type RateLimiterStore interface {
	Allow(ctx context.Context, key string, config RateLimiterConfig) (allowed bool, retryAfter time.Duration, err error)
}
```

Both the local and distributed (Redis) stores implement this interface, so the core rate limiter logic stays the same and only the storage backend changes.

pkg/gofr/service/logger.go
akshat-kumar-singhal: while this may be a subset of another interface, it is still an independent exported interface. We should at least make sure we mention this breaking change in the release notes.

pkg/gofr/service/rate_limiter_distributed.go
akshat-kumar-singhal: given that NewDistributedRateLimiter is exported, it should have independent validation of config - if a wrong config can cause issues in it's response object.

Umang01-hash: Done , both local and distributed rate limiters are being validated independently in thier factory functions.

pkg/gofr/service/logger.go
Umang01-hash: Sure will mention it.

docs/advanced-guide/http-communication/page.md
```diff
@@ -95,10 +95,24 @@ GoFr provides its user with additional configurational options while registering
...
+**Rate Limiter Store: Customization**
+GoFr allows you to use a custom rate limiter store by implementing the RateLimiterStore interface.This enables integration with any backend (e.g., Redis, database, or custom logic)
+Interface:
+```go
+type RateLimiterStore interface {
+Allow(ctx context.Context, key string, config RateLimiterConfig) (allowed bool, retryAfter time.Duration, err error)
+StartCleanup(ctx context.Context)
+StopCleanup()
+}
+```
```
coolwednesday, line 111: Can you add as a best practice it is advised to use Redis if it is the case for distributed systems.

Umang01-hash: Sure dont

coolwednesday (approved):

aryanmehrotra (approved):

PR #2373, opened 2025-09-30
Refactor - Move auth related to code to separate package under service

akshat-kumar-singhal:
**Description:**
Moved auth related code to separate sub package under service package

No

**Additional Information:**
N/A

**Checklist:**


pkg/gofr/service/auth/apikey_auth.go
```diff
@@ -15,14 +17,14 @@ type APIKeyConfig struct {
 	APIKey string
 }
```
aryanmehrotra, line 18: Introduces BREAKING CHANGE as if the user was importing in there main.go with the following import -

	"gofr.dev/pkg/gofr/service"

It now changes to - 
	"gofr.dev/pkg/gofr/service/auth"

akshat-kumar-singhal: Good catch. Handling this would require lot of convolution - will add this in v2 list.

akshat-kumar-singhal:
Causing breaking changes, will take up in V2

PR #2411, opened 2025-10-13
Improve test coverage for example/http-server-using-redis

AkshayCdra:
**Description:**
- Extract cache warmup logic into standalone function for better testability
- Add unit tests for Redis GET handler

**Checklist:**


Umang01-hash:
@AkshayCdr the coverage of `examples/http-server-using-redis` was **94.6%** previously and after this PR also it is same **94.6%**. How is this a improvement in coverage? 

In the `main.go`:

```
		err := ctx.Redis.Set(ctx, cacheKey, cacheValue, 0).Err()
		if err != nil {
			ctx.Logger.Errorf("Failed to warm up cache: %v", err)
			return err // Return the error to halt startup if caching fails.
		}
```

Only this part is uncovered currenly, if we can cover it then it will be an improvement for sure!

Umang01-hash:
@AkshayCdr I am closing this PR due t inactivity. Please feel free to reopen when the review comments are addressed.

PR #2503, opened 2025-11-07
updated default SQL histogram buckets for better latency analysis

Lakshith7748:
### Description

This PR enhances the default SQL query histogram configuration by introducing a wider range of bucket boundaries to improve real-world latency visibility.
Previously, the upper bound was limited to 10 ms, compressing slower queries into a single bucket.
Now, when no custom buckets are provided, a comprehensive default set (0.001 s to 60 s) is applied to capture micro to long-running query latencies effectively.

**Key change:**

Updated NewHistogram() in metrics.go to use a broader default bucket range.

Added validation for monotonically increasing bucket boundaries.

**Benefits:**

Improved accuracy for query latency analysis.

Better differentiation between fast, slow, and critical queries.

More meaningful P95/P99 latency metrics for production workloads.

**Breaking Changes:**

None 
Existing code providing custom bucket values remains fully compatible.

**Additional Information:**

Changes are limited to metrics.go under the metrics package.

### DIFF
--- a/pkg/gofr/metrics/register.go
+++ b/pkg/gofr/metrics/register.go
@@ -108,12 +108,35 @@ func (m *metricsManager) NewUpDownCounter(name, desc string) {
 //	[10, 100), [100, 1000), [1000, +Inf), where each range represents response times
 //	within a certain range, and the last bucket includes all values above 1000ms (represented by +Inf,
 //	which stands for positive infinity).
+
+//-->
+// NewHistogram registers a new histogram metric with specified bucket boundaries.
+// If no buckets are provided, a default set is used.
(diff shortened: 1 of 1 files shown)


Umang01-hash:
@Lakshith7748 This implementation to update SQL histogram is completely wrong.

As per the issue we need to correct the SQL datasources or on a broader scope all the datasources metrics. The changes in **NewHistogram**  made by you are not needed actually. The reason for the same is we initialize these Histograms in `container.go` and if we make the changes there it wil fix the issue.

I have addressed this issue in this PR: https://github.com/gofr-dev/gofr/pull/2526
Since it was a bug in existing functionality we needed to escalate it.

Issue #2551, opened 2025-11-19
Implement Server Side Rate limiter


coolwednesday:
Build server side rate limiting for HTTP as well as gRPC support.

0xEbrahim:
I want to work on this but i don't fully understand what you mean... so can you help me?

sammy200-ui:
@coolwednesday! I'd like to work on this feature. Here's my proposed approach:


Implement rate limiting middleware for both HTTP and gRPC servers using a token bucket algorithm via `golang.org/x/time/rate` (already available in go.mod).

### Design

#### 1. **HTTP Rate Limiter Middleware**
Location: `pkg/gofr/http/middleware/rate_limiter.go`

**Features:**
- Configurable requests per second (RPS) limit
- Configurable burst size
- Per-IP rate limiting (default) with option for global limiting
- Returns HTTP 429 (Too Many Requests) when limit exceeded
- Metrics tracking for rate-limited requests

**Configuration:**
```go
RATE_LIMIT_ENABLED=true
RATE_LIMIT_REQUESTS_PER_SECOND=100
RATE_LIMIT_BURST=10
RATE_LIMIT_PER_IP=true


app := gofr.New()
// Rate limiter will be automatically applied based on config
// Or manually:
app.UseMiddleware(middleware.RateLimiter(100, 10, true)) // rps, burst, perIP
```

2. gRPC Rate Limiter Interceptor
Location: `pkg/gofr/grpc_rate_limiter.go` (interceptor functions)

Features:

Unary interceptor for single RPCs
Stream interceptor for streaming RPCs
Returns codes.ResourceExhausted (gRPC status) when exceeded
Shares configuration with HTTP middleware
Works with existing observability interceptors

coolwednesday:
@sammy200-ui , Since @0xEbrahim also want to contribute, maybe both of you can collaborate together and pick up one. One of you can pick up HTTP and other can pick up gRPC.

coolwednesday:
@0xEbrahim, @sammy200-ui , Please decide who is interested in implementing what ....and discuss the approach accordingly. @sammy200-ui , please do help @0xEbrahim with the details if required. If still any blockers, feel free to tag any of the maintainers.

sammy200-ui:
That sounds good @coolwednesday, Happy to collaborate with you @0xEbrahim

@0xEbrahim - How about i work on  **HTTP rate limiter**  and you can do the  **gRPC rate limiter**? 

 please let me know what you think 

[...]

PR #2563, opened 2025-11-20
feat(http): implement server-side rate limiter middleware

sammy200-ui:
Description
Implements HTTP server-side rate limiting middleware using token bucket algorithm for #2551 (HTTP portion).

 Changes
-  Add `RateLimiter` middleware using `golang.org/x/time/rate`
-  Support per-IP and global rate limiting modes
-  Add `ErrorTooManyRequests` (HTTP 429) error type
-  Include 9 comprehensive test cases (all passing)
-  Add example application with full documentation
-  Auto-skip rate limiting for health check endpoints
-  Automatic cleanup of stale per-IP limiters every 5 minutes

Features
- **Token Bucket Algorithm**: Smooth rate limiting with configurable burst
- **Per-IP Limiting**: Each client IP gets independent rate limit (configurable)
- **Health Check Exemption**: `/.well-known/alive` and `/.well-known/health` always accessible
- **Prometheus Metrics**: Track rate limit violations via `app_http_rate_limit_exceeded_total`
- **IP Extraction**: Supports `X-Forwarded-For`, `X-Real-IP`, and `RemoteAddr`

Testing
All tests pass:
```bash
go test ./pkg/gofr/http/middleware/... -v -run TestRateLimiter
# PASS: 9/9 tests
```

Additional Information:

Uses golang.org/x/time/rate (already in dependencies)
Metrics exposed: app_http_rate_limit_exceeded_total counter
Example application available at `using-rate-limiter`
Complete documentation in example README with testing instructions
This PR implements the HTTP portion of #2551. 

**Checklist:**


sammy200-ui:
@coolwednesday i have raised the pr , you can review it

sammy200-ui:
@Umang01-hash just a follow up on the pr, any changes or fixes for the feat

Umang01-hash:
@sammy200-ui We will review your PR soon. Once reviewed we will let you know about any changes required for it.

Thanks for contributing to GoFr.

coolwednesday:
@sammy200-ui , is this feature only for single pod servers. How do you think we should handle rate limiting in case of distributed systems ?

sammy200-ui:
@coolwednesday Yes, currently this works for single-pod deployments. For distributed systems, each pod would enforce limits independently, so a client could exceed the global limit by hitting different pods, to handle this we would need ,redis-backed limiter or api gateway rate limiting or have document the limitation and keep it simple for v1

coolwednesday:
@sammy200-ui, Let's keep it simple and create another issue for that linking it to this. Also document this behaviour. However, I would suggest at least to make the factory function configurable enough to be extensive to Redis in future.

sammy200-ui:
@coolwednesday ok so i will add limiterBackend interface that will makes it easy to add Redis/other backends in future , add notes in code comments and example README explaining single-pod behavior , would that be good for this?? ,also do you want me to create a follow up issue for redis backed limiter or would you like to assign to someone else?

Umang01-hash:
> @coolwednesday ok so i will add limiterBackend interface that will makes it easy to add Redis/other backends in future , add notes in code comments and example README explaining single-pod behavior , would that be good for this?? ,also do you want me to create a follow up issue for redis backed limiter or would you like to assign to someone else?


@sammy200-ui I already defined one RateLimiterStore interface while implementing it for http-service option. Maybe you can have a look and use it : 

```go
// RateLimiterStore abstracts the storage and cleanup for rate limiter buckets.
type RateLimiterStore interface {
	Allow(ctx context.Context, key string, config RateLimiterConfig) (allowed bool, retryAfter time.Duration, err error)
	StartCleanup(ctx context.Context)
	StopCleanup()
}
```

The goal then will be we will use a single implementation for rate limiter of http service also and http or grpc also. This helps us to  resuing the same code and having a consistent functionality throughout. You can make it a part of another ticket for time.

@coolwednesday thoughts?

sammy200-ui:
@Umang01-hash @coolwednesday  The RateLimiterStore interface is for outbound HTTP client rate limiting, while my implementation is for incoming HTTP middleware.
They serve different purposes:
Your interface: Limits outbound requests to other services
My implementation: Limits incoming requests to the server

However, I see the value in consistency. Two options:

 1 : Keep this PR simple (in-memory only) and create a follow-up ticket to align both implementations with a shared interface + add Redis support for both.

 2: Refactor now to use RateLimiterStore, but it may delay this PR.

I prefer 1 to keep this focused. Thoughts?

coolwednesday:
> @Umang01-hash @coolwednesday The RateLimiterStore interface is for outbound HTTP client rate limiting, while my implementation is for incoming HTTP middleware. They serve different purposes: Your interface: Limits outbound requests to other services My implementation: Limits incoming requests to the server
> 
> However, I see the value in consistency. Two options:
> 
> 1 : Keep this PR simple (in-memory only) and create a follow-up ticket to align both implementations with a shared interface + add Redis support for both.
> 
> 2: Refactor now to use RateLimiterStore, but it may delay this PR.
> 
> I prefer 1 to keep this focused. Thoughts?

You can implement the interface and return errors as not supported when redis configs are passed. You can then pick this up in a subsequent PR.

sammy200-ui:
@Umang01-hash @coolwednesday I've refactored the rate limiter to use the RateLimiterStore interface pattern.

Changes made:

Created `rate_limiter_store.go` with the RateLimiterStore interface and memoryRateLimiterStore implementation
Refactored `RateLimiter` middleware to use the store pattern via an optional Store field in config
In-memory implementation works now with automatic cleanup
All tests passing with backward compatibility maintained
Added documentation about single-pod limitation

Redis support: As you suggested, I've left Redis implementation for a future PR. The store interface makes it easy to add later without breaking changes.

Let me know if you'd like any adjustments!

sammy200-ui:
@coolwednesday @Umang01-hash any new changes to be expected ?

sammy200-ui:
@Umang01-hash @coolwednesday just a follow up on the pr

sammy200-ui:
Hey @Umang01-hash do i need to resolve the co-pilot requests?

Umang01-hash:
> Hey @Umang01-hash do i need to resolve the co-pilot requests?

Yes @sammy200-ui if the comments given by co-pilot seems meaningful you should definitely resolve it. If not please comment why you think the given review comment is not useful/meaninful.

sammy200-ui:
Hello @Umang01-hash I have addressed all 8 Copilot review comments , let me know if there are any more changes

pkg/gofr/http/middleware/rate_limiter_test.go
```diff
@@ -0,0 +1,507 @@
...
+
+func TestGetIP_Priority(t *testing.T) {
+	req := httptest.NewRequest(http.MethodGet, "/test", http.NoBody)
+	req.Header.Set("X-Forwarded-For", "203.0.113.1")
+	req.Header.Set("X-Real-IP", "203.0.113.2")
+	req.RemoteAddr = "192.168.1.1:12345"
+
+	ip := getIP(req, true)
+	assert.Equal(t, "203.0.113.1", ip, "X-Forwarded-For should have highest priority when trusting proxies")
+}
```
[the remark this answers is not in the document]
sammy200-ui: 1. **Test real-world scenarios**: Current tests cover valid use cases that occur 
   in production: properly formatted X-Forwarded-For, X-Real-IP, and RemoteAddr 
   parsing. These represent >99% of actual traffic.

2. **Avoid testing broken infrastructure**: Tests for whitespace-only headers 
   (e.g., "X-Forwarded-For: , , ") would validate behavior for misconfigured 
   proxies that violate HTTP standards. Such scenarios should fail during 
   infrastructure testing, not application testing.

3. **Existing fallback is tested**: TestGetIP_RemoteAddr already validates the 
   fallback path (using RemoteAddr when headers are absent/invalid), which 
   handles the malformed header case by design.

4. **Test maintenance cost**: Adding tests for every possible malformed input 
   increases maintenance burden without improving production reliability. The 
   explicit documentation warning ("only enable TrustedProxies behind trusted 
   proxies") shifts responsibility to infrastructure configuration.

5. **RemoteAddr is always valid**: The net.SplitHostPort fallback ensures 
   RemoteAddr extraction cannot fail, providing defense in depth.

If there's a specific production incident or CVE showing these edge cases cause 
real issues, I'm happy to add comprehensive validation. Otherwise, I believe 
current coverage is appropriate for the feature's security model.

pkg/gofr/http/middleware/rate_limiter.go
```diff
@@ -0,0 +1,153 @@
...
+			// X-Forwarded-For can contain multiple IPs, take the first one
+			ips := strings.Split(forwarded, ",")
+			if len(ips) > 0 {
+				return strings.TrimSpace(ips[0])
+			}
+		}
+
+		// Check X-Real-IP header
+		if realIP := r.Header.Get("X-Real-IP"); realIP != "" {
+			return realIP
```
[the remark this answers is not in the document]
sammy200-ui: 1. **Fail-safe fallback**: If X-Forwarded-For or X-Real-IP contain whitespace/empty 
   strings, the code already falls through to RemoteAddr (direct connection IP), 
   which is always valid and cannot be spoofed.

2. **Malformed headers indicate broken infrastructure**: Headers like 
   "X-Forwarded-For: , , " represent misconfigured proxies that violate HTTP 
   standards. Such scenarios should not occur in production environments where 
   TrustedProxies=true is used.

3. **Explicit security warning**: The documentation and README clearly state that 
   TrustedProxies should ONLY be enabled behind trusted reverse proxies (nginx, 
   ALB, etc.). Administrators enabling this feature are expected to ensure their 
   proxy infrastructure is properly configured.

4. **Defense in depth**: Even if an empty string becomes the rate limit key, the 
   worst case is multiple malformed requests share one rate limit bucket - they 
   still get rate limited, just collectively rather than individually. This is 
   not a security vulnerability.

5. **Simplicity vs complexity tradeoff**: Adding validation for malformed proxy 
   headers adds code complexity to handle scenarios that should never occur in 
   properly configured environments. The current implementation prioritizes 
   readability and handles the 99.9% valid use case.

If there's evidence that major reverse proxies (nginx, AWS ALB, Cloudflare) can 
send such malformed headers, I'm happy to add validation. Otherwise, I believe 
the current implementation with clear documentation is the right tradeoff.

sammy200-ui:
Hi @Umang01-hash ,
 I've addressed all 9 comments from the second review and pushed the changes. However, Copilot has now flagged 3 additional comments regarding edge cases in IP header validation. After careful consideration, I believe these represent diminishing returns rather than critical issues. I have also tried to add the explanation for the 3 new reviews on the comment section

sammy200-ui:
Hey @Umang01-hash any new updates ?

Umang01-hash:
@sammy200-ui Sorry for the delay, i will review the PR soon and let you knwo about any changes if required.

Thanks for the patience.

sammy200-ui:
@Umang01-hash Sounds good. Let me know if you need any changes from my end.

pkg/gofr/http/middleware/rate_limiter.go
```diff
@@ -0,0 +1,153 @@
...
+
+	gofrHttp "gofr.dev/pkg/gofr/http"
+)
+
+var (
+	// ErrInvalidRequestsPerSecond is returned when RequestsPerSecond is not positive.
+	ErrInvalidRequestsPerSecond = errors.New("requestsPerSecond must be positive")
+
+	// ErrInvalidBurst is returned when Burst is not positive.
+	ErrInvalidBurst = errors.New("burst must be positive")
```
Umang01-hash: Why are these errors exported? If we are not intending to use them outside of the directory can we please unexport it??

pkg/gofr/http/middleware/rate_limiter.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	if c.Burst <= 0 {
+		return ErrInvalidBurst
+	}
+
+	return nil
+}
+
+type rateLimiterMetrics interface {
```
Umang01-hash: Why we have defined a new interface **rateLimiterMetrics**. Shouldn't we try to reuse the existing interfaces.

We have already defined :

```go
// Metrics represents an interface for registering the default metrics in GoFr framework.
type Metrics interface {
	IncrementCounter(ctx context.Context, name string, labels ...string)
}

```

the above metrics interface in file `pkg/gofr/http/metrics.go`, you can reuse it.

Or even inside the **middleware** directory too:

we have this `metrics` interface defined:

```go
type metrics interface {
	IncrementCounter(ctx context.Context, name string, labels ...string)
	DeltaUpDownCounter(ctx context.Context, name string, value float64, labels ...string)
	RecordHistogram(ctx context.Context, name string, value float64, labels ...string)
	SetGauge(name string, value float64, labels ...string)
}
```

pkg/gofr/http/middleware/rate_limiter.go
```diff
@@ -0,0 +1,153 @@
...
+func getIP(r *http.Request, trustProxies bool) string {
+	// Only trust proxy headers if explicitly configured
+	if trustProxies {
+		// Check X-Forwarded-For header first
+		forwarded := r.Header.Get("X-Forwarded-For")
+		if forwarded != "" {
+			// X-Forwarded-For can contain multiple IPs, take the first one
+			ips := strings.Split(forwarded, ",")
+			if len(ips) > 0 {
+				return strings.TrimSpace(ips[0])
```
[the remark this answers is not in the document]
Umang01-hash: @sammy200-ui This comment is correct. If `X-Forwarded-For` contains only whitespace or starts with comma, this returns 
empty string. The code checks `if forwarded != ""` but "   " (whitespace) is not empty. 
This means strings.TrimSpace(ips[0]) can return "" without falling through 
to RemoteAddr.

pkg/gofr/http/middleware/rate_limiter_store.go
```diff
@@ -0,0 +1,126 @@
...
+
+	entry := val.(*limiterEntry)
+	atomic.StoreInt64(&entry.lastAccess, now)
+
+	// Check if request is allowed
+	if !entry.limiter.Allow() {
+		// Calculate retry-after duration
+		reservation := entry.limiter.Reserve()
+		if !reservation.OK() {
+			return false, time.Second, nil
```
Umang01-hash: When `reservation.OK()` returns false (burst exceeded), you return a fixed 1-second delay. This is inaccurate for low request rates.

Example: **If RequestsPerSecond=0.1 (1 req per 10 seconds), the client should wait ~10 seconds, not 1 second.**

Umang01-hash (commented):
@sammy200-ui Creating a separate example for this functionality may not be a very good option. Either we can try to fit it in an already existing example or only update the documentation. Also we always try and keep our documentation simple to the point and clear. I think we can remove **Running the example, Testing, monitoring**, etc from the docs you created. 

You should create a new documentation inside the docs directory and delete this new `using-rate-limiter` directory from examples.

Umang01-hash:
Screenshot 2025-12-16 at 1 17 25 PM

@sammy200-ui Please also resolve these code quality issues in the PR. Thanks for the patience.

sammy200-ui:
Hello @Umang01-hash All review feedback has been addressed! 
Let me know if you need any changes from my end.

sammy200-ui:
@Umang01-hash some code quality errors that are fixed

sammy200-ui:
@Umang01-hash  some code quality errors that are fixed , you can run the test again

sammy200-ui:
@Umang01-hash Hi! Just checking in on this PR since it’s already approved.
Let me know if there’s anything else needed from my side before merge. Thanks!

pkg/gofr/http/middleware/rate_limiter_store.go
```diff
@@ -0,0 +1,129 @@
...
+	cfg := m.config
+
+	// Get or create limiter for this key
+	val, _ := m.limiters.LoadOrStore(key, &limiterEntry{
+		limiter:    rate.NewLimiter(rate.Limit(cfg.RequestsPerSecond), cfg.Burst),
+		lastAccess: now,
+	})
+
+	entry := val.(*limiterEntry)
+	atomic.StoreInt64(&entry.lastAccess, now)
```
coolwednesday: When two goroutines call LoadOrStore with the same new key, both create new entries, only one is stored, and the other is lost (memory leak).

pkg/gofr/http/middleware/rate_limiter.go
```diff
@@ -0,0 +1,172 @@
...
+				return
+			}
+
+			// Determine the rate limit key (IP or global)
+			key := "global"
+			if config.PerIP {
+				key = getIP(r, config.TrustedProxies)
+			}
+
+			// Check rate limit
```
coolwednesday, line 149: If getIP() returns an empty string, all requests share the same bucket, bypassing per-IP limits.

pkg/gofr/http/middleware/rate_limiter_store.go
```diff
@@ -0,0 +1,129 @@
...
+	// Check if request is allowed
+	if !entry.limiter.Allow() {
+		// Calculate retry-after duration
+		reservation := entry.limiter.Reserve()
+		if !reservation.OK() {
+			// Burst exceeded - calculate delay based on request rate
+			// Time to wait for one token = 1 / RequestsPerSecond
+			delay := time.Duration(float64(time.Second) / cfg.RequestsPerSecond)
+			return false, delay, nil
+		}
```
coolwednesday: Calls Allow() then Reserve(), which is inefficient and can produce incorrect retry-after values due to races.

pkg/gofr/http/middleware/rate_limiter_store.go
```diff
@@ -0,0 +1,129 @@
...
+	atomic.StoreInt64(&entry.lastAccess, now)
+
+	// Check if request is allowed
+	if !entry.limiter.Allow() {
+		// Calculate retry-after duration
+		reservation := entry.limiter.Reserve()
+		if !reservation.OK() {
+			// Burst exceeded - calculate delay based on request rate
+			// Time to wait for one token = 1 / RequestsPerSecond
+			delay := time.Duration(float64(time.Second) / cfg.RequestsPerSecond)
```
coolwednesday: No bounds checking; can overflow or be zero.

pkg/gofr/http/middleware/rate_limiter_store.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// Use stored config for consistency across all keys
+	cfg := m.config
+
+	// Get or create limiter for this key
+	val, _ := m.limiters.LoadOrStore(key, &limiterEntry{
+		limiter:    rate.NewLimiter(rate.Limit(cfg.RequestsPerSecond), cfg.Burst),
+		lastAccess: now,
+	})
+
```
coolwednesday, line 73: No limit on unique keys; attacker can exhaust memory.

coolwednesday (changes requested):

Umang01-hash:
@sammy200-ui Hey! Your PR has some review comments on it. Can you please resolve it so we can re-review and proceed towards merging if all things have been addressed. Please let us know if you need any assistance for the same.

sammy200-ui:
@Umang01-hash @coolwednesday oks sure i will make these changes

sammy200-ui:
Hello @Umang01-hash @coolwednesday done with the changes , you can review the pr again. let me know if any more changes to be made

Issue #2480, opened 2025-10-28
Add response time unit in json logs


aryanmehrotra:
**Describe the bug**
Right now json logs dont have the unit for the response time, latency etc but the prettyprinted logs in terminal do have, JSON logs should also have the unit to make it pretty clear for debugging purpose.

Moreover an extension to this would be keeping consistent unit across, but this needs to be thought throughly as gofr has different units, it would be for a reason.

0xEbrahim:
Can you define the scope that needs to be edited, I don't clearly understand but i want to contribute here again, sorry

gizmo-rt:
What purpose does this serve  given we already have metrics ?

aryanmehrotra:
@0xEbrahim we need to first decide if we want to do this as its in triage phase right now.

aryanmehrotra:
@gizmo-rt 
1. Parity between json and terminal logs
2. Metrics bucket are capped at 30 seconds, so any response time above that comes in 30s only.

aaryyya:
take

gizmo-rt:
Parity between logs is fine, though if it is solely due to metrics bucket cap then we should explore making https://github.com/gofr-dev/gofr/blob/4e9dd129983c5e5c0a6d093b863c42730a300a62/pkg/gofr/container/container.go#L262 configurable for users

aryanmehrotra:
> Parity between logs is fine

we dont have unit of time in json logs but we have them in terminal pretty printed logs

DevaanshKathuria:
Hi @aryanmehrotra!
I’ve opened a PR to add human-readable response time units to the JSON logs while keeping the existing numeric field for backward compatibility.
This improves parity between JSON logs and the pretty-printed terminal logs, and aligns with the discussion in the related issue.
Whenever you get a moment, could you please review it?

202378030317:
I'd like to work on this. Please assign it to me.

202378030317:
I'm a new contributor and would like to work on this issue. Could you please assign it to me? Thanks!

Fah218:
Hi! I would like to work on this issue. Could you please assign it to me?

krishachra:
Hi! I'd like to work on this issue.

I'm a first-time contributor to GoFr and would like to implement the JSON response time unit while keeping the existing behavior in mind. If there isn't already someone actively working on it, I'll start investigating the logging code and submit a PR.

Thanks!

khushi0515:
[...]

PR #2597, opened 2025-11-25
Fix JSON logs missing time units by adding human-readable response time

DevaanshKathuria:
This PR adds human-readable time units to JSON-formatted logs to match the terminal pretty-printed logs.

Changes:-
- Added a new response_time_human field to JSON logs, using duration.String() to include units (µs, ms, s).
- Preserved the existing numeric response_time field (in microseconds) for backward compatibility.
- Ensures complete parity between JSON logs and terminal logs while maintaining non-breaking changes.

Fixes:- #2480


aryanmehrotra:
Lets first discuss on how this needs to be approached in the issue ticket itself, as it has to be changed at multiple places.
Right now the pretty print has the unit and not the values just updating the struct won't help moreover the ticket is still in the triage phase.

Issue #2524, opened 2025-11-12
Getting Error logs for traces with 201 status code


coolwednesday:
<img width="1461" height="393" alt="Image" src="https://github.com/user-attachments/assets/2f99b592-0e26-4816-b806-a9d0fc08805d" />

coolwednesday:
This is an issue with OTEL itself. We will see if as a package using it . Do we need to handle this or not.....

iNotSoCrisp:
@coolwednesday @Umang01-hash 
I have fixed the issue. Please review it

coolwednesday:
> [@coolwednesday](https://github.com/coolwednesday) [@Umang01-hash](https://github.com/Umang01-hash) I have fixed the issue. Please review it

Hi @iNotSoCrisp ,
Quick note: it’s best not to work on issues that are already assigned. If you notice inactivity, please tag a maintainer to get it reassigned first to avoid duplicate work.
That said, I’m assigning this issue to you this time as an exception. Thanks for your understanding, and looking forward to your contribution!

PR #2634, opened 2025-12-02
fix: ignore error logs for traces with 201 status code

iNotSoCrisp:
## Pull Request Template

**Description:**

- Fixed an issue where traces with a 201 status code were being logged as errors.
- Addresses Issue #2524.
- The Zipkin exporter treats 201 as an error, but it is a success status. This change ignores such errors in the error handler to prevent false positives in the logs.

- None.

**Additional Information:**

- No new dependencies.

**Checklist:**

### DIFF
--- a/pkg/gofr/datasource/file/gcs/fs.go
+++ b/pkg/gofr/datasource/file/gcs/fs.go
@@ -38,10 +38,11 @@ func New(config *Config, logger datasource.Logger, metrics file.StorageMetrics) 
 
 	fs := &fileSystem{
 		CommonFileSystem: &file.CommonFileSystem{
-			Provider: adapter,
-			Location: config.BucketName,
-			Logger:   logger,
-			Metrics:  metrics,
+			Provider:     adapter,
+			Location:     config.BucketName,
+			Logger:       logger,
+			Metrics:      metrics,
+			ProviderName: "GCS", // Set provider name for observability
 		},
 	}
 
--- a/pkg/gofr/datasource/file/gcs/fs_test.go
+++ b/pkg/gofr/datasource/file/gcs/fs_test.go
@@ -101,6 +101,45 @@ func TestNew_Success(t *testing.T) {
 	}
 }
 
+func TestGCSFileSystem_Observe_ProviderName(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockLogger := file.NewMockLogger(ctrl)
+	mockMetrics := file.NewMockMetrics(ctrl)
+
+	fs := &fileSystem{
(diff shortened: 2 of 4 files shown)


pkg/gofr/otel.go
```diff
@@ -148,5 +148,11 @@ type otelErrorHandler struct {
 }
 
 func (o *otelErrorHandler) Handle(e error) {
+	// 201 is a success status code, but the zipkin exporter treats it as an error.
+	// We should ignore this error.
+	if strings.Contains(e.Error(), "status 201") {
+		return
+	}
+
```
coolwednesday: Can we match on the type or inspect the HTTP response code directly.

coolwednesday (changes requested):

pkg/gofr/otel.go
iNotSoCrisp: > Can we match on the type or inspect the HTTP response code directly.

Fixed @coolwednesday

iNotSoCrisp:
fixed @coolwednesday

pkg/gofr/datasource/redis/redis_test.go
```diff
@@ -35,7 +35,10 @@ func Test_NewClient_InvalidPort(t *testing.T) {
 	mockMetrics := NewMockMetrics(ctrl)
 	mockConfig := config.NewMockConfig(map[string]string{"REDIS_HOST": "localhost", "REDIS_PORT": "&&^%%^&*"})
 
-	mockMetrics.EXPECT().RecordHistogram(gomock.Any(), "app_redis_stats", gomock.Any(), "hostname", gomock.Any(), "type", "ping")
+	// The go-redis library may send multiple commands during initialization (hello, client, ping, etc.)
```
Umang01-hash, line 38: Why this file has been changed as it is not related to the scope of the PR?

coolwednesday: It was failing locally.

coolwednesday:
> @up1 While the regex approach works, it's overengineered as Error message formats can change between OTEL versions also Running regex on every error (even if precompiled) is not an ideal approach. Consider using `strings.Contains(msg, "status 201")` and similar checks instead - it's simpler, faster, and more maintainable.

Using regex in case in future we can have a check for all success codes and not just 201 ....

Umang01-hash:
@coolwednesday I understand the concern, but the regex doesn't actually provide extra flexibility here. You're already checking all **2xx codes with code >= 200 && code < 300** - the regex just extracts the number.
For extracting a number from "status XXX", `strings.Split()` or `strings.Fields()` is simpler and just as effective.

The regex adds parsing overhead without adding maintainability.

Umang01-hash (approved):

PR #2638, opened 2025-12-04
Fix/nil correlation

deepakpathik:
Problem

In the HTTP service tests, logs were showing "Nil Correlation ID" because the tests were using a no-op tracer. Since no real tracer was configured, no valid TraceID values were generated during test execution. This made the test logs less clear and prevented proper validation of tracing behavior.

Solution

The tests have been updated to use a real OpenTelemetry tracer provider. Specifically, go.opentelemetry.io/otel/sdk/trace is now configured as the TracerProvider in pkg/gofr/service/new_test.go. This ensures that valid TraceIDs are created during tests and that the tracing logic is fully exercised.

Improvements

Test logs now contain meaningful correlation IDs instead of "Nil Correlation ID".

Tracing behavior is validated more accurately during tests.

Removed outdated TODO comments related to this issue.

Breaking Changes

None.

Additional Information

This change improves the clarity, reliability, and observability of HTTP service tests without impacting runtime behavior.


Umang01-hash:
@deepakpathik Your PR contains the following Code Quality Issues:

Screenshot 2025-12-05 at 1 08 20 PM

Please fix them and also resolve the `go.work.sum` conflict so that we can review the PR.

PR #2759, opened 2025-12-24
Fix/http double slash

### LINKED ISSUE #2643: Bug: Double Slash `//` Path Causes POST Route to Resolve to GET Handler
#### **Description**

When defining both `GET /hello` and `POST /hello` routes, sending a request to `//hello` results in **POST resolving incorrectly to the GET handler**.

#### **Steps to Reproduce**

Using the following code:

```go
package main

import (
	"gofr.dev/pkg/gofr"
)

func main() {
	a := gofr.New()

	a.GET("/hello", func(c *gofr.Context) (any, error) {
		return "Hello World!", nil
	})

	a.POST("/hello", func(c *gofr.Context) (any, error) {
		return "Hello World! From POST", nil
	})

	a.Run()
}
```

Execute:

```bash
curl --location --request POST 'http://localhost:9000//hello' --data ''
curl --location 'http://localhost:9000//hello'
```

#### **Current Behavior**

Both of the above reach the **GET `/hello`** endpoint.

---

#### **Expected Behavior**

* `POST //hello` should resolve to `POST /hello`
* `GET //hello` should resolve to `GET /hello`
* The HTTP method should **not be overridden or ignored** because of path normalization

---

#### **Additional Observation**

If POST `/hello` is removed:

```go
// only GET defined
a.GET("/hello", handler)
```

Then **both**:

```bash
curl -X POST http://localhost:9000//hello
curl http://localhost:9000//hello
```

→ correctly redirect to the GET handler, meaning the normalization logic redirects to an existing matching route, but incorrectly defaults to GET when both GET + POST exist.

---

#### **Impact**

This causes:

* **Unexpected insecure behavior**
* **Incorrect handler execution**
* Potential **API breaking behavior**
* POST operations being mistakenly processed as GET

---

#### **Root Cause Suspicion**

GoFr (via Gorilla Mux) normalizes paths with double slashes. But during resolution, it appears:

* `//hello` is matched as `/hello`
* Method resolution incorrectly prioritizes GET route when multiple HTTP methods exist for same normalized path.

---

I have used postman to test this.

Umang01-hash:
## Pull Request Template

**Description:**

-   Fixed #2643.
- **Original Bug**: Gorilla Mux's automatic path cleaning issued HTTP 301 redirects for double slashes (e.g., `//hello` → `/hello`), causing some HTTP clients like Postman to convert POST requests to GET during the redirect process, violating HTTP method preservation.

- Modified `Router.ServeHTTP()` to normalize consecutive slashes **(e.g., //hello → /hello)** in-memory before routing, eliminating HTTP 301 redirects that could cause POST requests to be converted to GET in some HTTP clients.
- Added **SkipClean(true)** to Gorilla Mux configuration to disable automatic path cleaning with redirects, ensuring consistent method-preserving behavior across all HTTP clients.

**Checklist:**


pkg/gofr/http/router.go
```diff
@@ -42,6 +42,61 @@ func NewRouter() *Router {
...
+
+	result := normalized.String()
+
+	// Ensure we always have at least "/"
+	if result == "" {
+		return "/"
+	}
+
+	return result
+}
```
ccoVeille: [path.Clean](https://pkg.go.dev/path#example-Clean) from std lib might be what you are looking for

Umang01-hash: @ccoVeille Thanks! I used this **path.Clean()** and completely eliminated our **normalizePathSlashes** function.

pkg/gofr/http/router.go
```diff
@@ -42,6 +43,34 @@ func NewRouter() *Router {
...
+
+	// path.Clean returns "." for empty paths, convert to "/" for HTTP routing
+	if normalizedPath == "." {
+		normalizedPath = "/"
+	}
+
+	// Ensure path starts with "/" for HTTP routing
+	if !strings.HasPrefix(normalizedPath, "/") {
+		normalizedPath = "/" + normalizedPath
+	}
```
ccoVeille:
```suggestion
	normalizedPath = "/" + strings.TrimLeft(normalizedPath, "/")
```

pkg/gofr/http/router_test.go
```diff
@@ -77,6 +77,108 @@ func TestRouterWithMiddleware(t *testing.T) {
...
+
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte("POST handler"))
+	}))
+
+	tests := []struct {
+		desc string
+		path string
+	}{
+		{desc: "GET request to /hello", path: "/hello"},
```
ccoVeille, line 107: Maybe test also

```
.
..
hello
foo/../bar
../hello
```

pkg/gofr/http/router.go
Umang01-hash: sure done

pkg/gofr/http/router_test.go
Umang01-hash: added the tests for it.

ccoVeille (approved):

pkg/gofr/http/router.go
```diff
@@ -42,6 +43,32 @@ func NewRouter() *Router {
 	return r
 }
 
+// ServeHTTP implements http.Handler interface with path normalization.
```
ccoVeille:
```suggestion
// ServeHTTP implements [http.Handler] interface with path normalization.
```

Umang01-hash: @ccoVeille done!

ccoVeille (approved):

PR #2766, opened 2025-12-27
fix(http): handle JSON encoding errors in responder

### LINKED ISSUE #2739: HTTP Responder silently fails in case of math.NaN() is returned
**Describe the bug**
json encoder fails inside responder when math.NaN() comes to be marshalled

**To Reproduce**

1. The code is

```
func HelloHandler(c *gofr.Context) (any, error) {
	return math.NaN(), nil
}   
   ```

2. The error is

Status Code is set as 200, but the server fails to give response neither there is any log as

```
	if r.handleSpecialResponseTypes(data, err) {
		return
	}

	var resp any

	switch v := data.(type) {
	case resTypes.Raw:
		resp = v.Data
	case resTypes.Response:
	default:
		if isNil(data) {
			data = nil
		}

	}

	if r.w.Header().Get("Content-Type") == "" {
	}

	r.w.WriteHeader(statusCode)

	_ = json.NewEncoder(r.w).Encode(resp) -> Error not being handled here
}
```

**Expected behavior**
Since golang support math.NaN() gofr should also support handling it, and in anycase server should tell the user and set the correct status code when anything like this happens

**Screenshots**
Image

Aggarwalmansi:
## Description

The HTTP responder was writing headers before attempting JSON encoding.
When the encoder failed (e.g. `math.NaN()`), the error was silently ignored,
resulting in a `200 OK` response with an empty body and no logs.

This PR ensures JSON encoding is attempted **before** writing headers and
properly handles encoding failures.

## What changed

- Encode response payload into a buffer before writing headers
- Handle JSON encoding errors explicitly
- Return `500 Internal Server Error` when encoding fails
- Prevent silent response failures

## Why this is needed

Go’s JSON encoder does not support values like `NaN` or `Inf`.
Previously, such cases caused the server to respond with a misleading success status.
This change ensures correct HTTP semantics and improves debuggability.

## How it was tested

- Manually verified using a handler returning `math.NaN()`
- Confirmed response returns `500 Internal Server Error` instead of `200`
- Ran `go test ./pkg/gofr/http/...`

## Related Issue

Fixes #2739

### DIFF
--- a/pkg/gofr/http/responder.go
+++ b/pkg/gofr/http/responder.go
@@ -1,6 +1,7 @@
 package http
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"net/http"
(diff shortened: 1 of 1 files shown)


Umang01-hash:
@Aggarwalmansi Thankyou for taking your time to improve GoFr but currently this approach breaks GoFr's API contract. 
Using **http.Error()** returns **plain text with text/plain content-type instead of our structured JSON response format**. {"error":{"message":"..."}} with application/json. 

This inconsistency will **break existing clients expecting JSON**. Closing in favor of a solution that maintains GoFr's response structure.
​

PR #2769, opened 2025-12-29
fix math.Nan() return bug by placing correct encoding failure status …

### LINKED ISSUE #2739: HTTP Responder silently fails in case of math.NaN() is returned

Umang01-hash:
## Pull Request Template

**Description:**

-   Fixes #2739.
- Pre-encode responses using `json.Marshal()` before writing HTTP status code. If encoding fails (NaN, Inf, channels, functions), r**eturn HTTP 500 with error message.** 
- Added explicit newline (`\n`) after `json.Marshal()` output to maintain backwards compatibility with existing `json.Encoder.Encode()` behavior.

Screenshot 2025-12-29 at 12 12 09 PM

**Checklist:**

### DIFF
--- a/docs/advanced-guide/using-publisher-subscriber/page.md
+++ b/docs/advanced-guide/using-publisher-subscriber/page.md
@@ -10,7 +10,7 @@ scaled and maintained according to its own requirement.
 ## Design choice
 
 In GoFr application if a user wants to use the Publisher-Subscriber design, it supports several message brokers, 
-including Apache Kafka, Google PubSub, MQTT, and NATS JetStream.
+including Apache Kafka, Google PubSub, MQTT, NATS JetStream, and Redis Pub/Sub.
 The initialization of the PubSub is done in an IoC container which handles the PubSub client dependency.
 With this, the control lies with the framework and thus promotes modularity, testability, and re-usability.
 Users can do publish and subscribe to multiple topics in a single application, by providing the topic name.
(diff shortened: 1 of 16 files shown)


pkg/gofr/http/responder.go
```diff
@@ -53,9 +53,22 @@ func (r Responder) Respond(data any, err error) {
...
+
+		errorResp := response{Error: map[string]any{"message": "failed to encode response as JSON"}}
+		errorJSON, _ := json.Marshal(errorResp)
 
-	_ = json.NewEncoder(r.w).Encode(resp)
+		_, _ = r.w.Write(errorJSON)
+		_, _ = r.w.Write([]byte("\n"))
+
+		return
+	}
```
ccoVeille: You are ignoring the error maybe you could simply hardcore things

ccoVeille, line 63: What about this

```go
        if encodeErr != nil { 
                 r.w.WriteHeader(http.StatusInternalServerError) 
                 _, _ = r.w.Write([]byte(`{"error":{"message": "failed to encode response as JSON"}}`+"\n"))
                 return 
         }

File: pkg/gofr/http/responder_test.go [hunk 2]
```diff
@@ -533,3 +534,71 @@ func TestResponder_XMLFileTemplate_ErrorStatusCodes(t *testing.T) {
...
+		assert.Equal(t, tc.expectedCode, result.StatusCode, "TEST[%d] Failed: %s", i, tc.desc)
+
+		body := new(bytes.Buffer)
+		_, err := body.ReadFrom(result.Body)
+
+		require.NoError(t, err, "TEST[%d] Failed: %s", i, tc.desc)
+
+		assert.NotEmpty(t, body.String(), "TEST[%d] Failed: %s", i, tc.desc)
+
+		result.Body.Close()
```
ccoVeille: The error is never checked, but also the body is not closed if a require fails above this line.

Please consider adding a t.Cleanup

pkg/gofr/http/responder.go
Umang01-hash: Sure done

pkg/gofr/http/responder_test.go
Umang01-hash: Sure added and updated the test.

pkg/gofr/http/responder.go
```diff
@@ -53,9 +53,18 @@ func (r Responder) Respond(data any, err error) {
...
+		r.w.WriteHeader(http.StatusInternalServerError)
+
+		_, _ = r.w.Write([]byte(`{"error":{"message": "failed to encode response as JSON"}}` + "\n"))
+
+		return
+	}
 
-	_ = json.NewEncoder(r.w).Encode(resp)
+	r.w.WriteHeader(statusCode)
+	_, _ = r.w.Write(jsonData)
```
ccoVeille, line 66: It's a bit out of the scope of the PR, but do you have a way like a logger to report you got an error when writting to output.

Usually, it's a connection closed issue, but it could be interesting to catch them

pkg/gofr/http/responder_test.go
```diff
@@ -533,3 +534,73 @@ func TestResponder_XMLFileTemplate_ErrorStatusCodes(t *testing.T) {
...
+	for i, tc := range tests {
+		recorder := httptest.NewRecorder()
+		responder := NewResponder(recorder, http.MethodGet)
+
+		responder.Respond(tc.data, nil)
+
+		result := recorder.Result()
+
+		t.Cleanup(func() {
+			result.Body.Close()
```
ccoVeille: This maybe

```suggestion
			require.NoError(t, result.Body.Close())
```

pkg/gofr/http/responder.go
Umang01-hash: @ccoVeille  Great observation! However, **Write errors (connection closed, broken pipe, etc.)** occur after we've successfully set the correct status code and are unrecoverable at this point. 
Adding a logger would require modifying the `NewResponder` factory and struct signature for minimal value, since these transport-layer failures are already tracked by `net/http` and typically represent non-actionable client disconnections.

pkg/gofr/http/responder_test.go
Umang01-hash: done

pkg/gofr/http/responder.go
ccoVeille: Indeed. Thanks for the detailed reply

coolwednesday (approved):

PR #2790, opened 2026-01-02
add check to disable running metrics server when port is set to 0

### LINKED ISSUE #2789: Make metrics disabled when port is explicitly set to 0
When `METRICS_PORT` is not configured or set to an invalid value, GoFr defaults to port **2121** for the metrics server. There is no way to explicitly disable the metrics server, which may be undesirable in development environments or specific deployment scenarios.

Also if the PORT provided in configs is busy or unreachable the app stops with a **FATAL error metrics port blocked**.

 Allow users to disable the metrics server by setting `METRICS_PORT=0`. When this configuration is detected:
- Skip metrics server initialization (app.metricServer remains nil)
- Log a message: "Metrics server is disabled (**METRICS_PORT=0**)"
- The existing nil check in startMetricsServer handles graceful no-op

Umang01-hash:
## Pull Request Template

**Description:**

-   Closes #2789.
Screenshot 2026-01-05 at 1 59 00 PM

**Checklist:**


coolwednesday:
Can we instead have a disable variable introduce ...better serves the purpose. 0 is also a valid port. Hence I do not want to use it as a boolean.

Umang01-hash:
@coolwednesday Thanks for the valuable feedback. However, setting `Port 0 = disabled` is an established cloud-native convention, kubernetes also follow this convention. Introducing an additional variable for disabling it will lead to user remembering multiple configs for a single feature which is not the best behaviour.

PR #2812, opened 2026-01-07
Metrics in HttpService Options

### LINKED ISSUE #2813: metrics support for the circuit breaker
Currently, the circuit breaker does not expose any metrics. This makes it difficult to observe its behavior in production—such as how often it opens, how many requests are rejected, or whether failures are transient or persistent. Without metrics, troubleshooting service instability and validating circuit breaker configuration (thresholds, intervals) becomes difficult .

Expose standard, observable metrics for the circuit breaker that can be scraped by monitoring systems (e.g., Prometheus).

below is the code sniper where I have added circuit breaker

Image

coolwednesday:
# Fix Circuit Breaker Concurrency, Error Handling, and Metrics
- Closes #2813 

## Description
This PR addresses critical issues in the Circuit Breaker implementation regarding concurrency safety, error handling logic, and metric reporting.

### 1. Concurrency Safety & Race Condition Fixes
*   **`pkg/gofr/service/circuit_breaker.go`**: Added necessary locking in `startHealthChecks` and `tryCircuitRecovery`. Previously, `resetCircuit` was being called without holding the lock, leading to potential race conditions on `state` and `failureCount`.
*   **`pkg/gofr/service/metrics_helper.go`**: Implemented a global lock (`metricsMu`) for metric registration to prevent "concurrent map writes" when multiple services initialize metrics simultaneously.

### 2. Robust Edge Case Handling
*   **Thundering Herd Fix**: In `tryCircuitRecovery`, added a check for `cb.state == ClosedState` immediately after acquiring the lock. This prevents queued requests from re-running the health check if another request has already recovered the circuit.
*   **Busy Loop Prevention**: Updated `lastChecked` timestamp *before* attempting a health check in `tryCircuitRecovery`. This ensures that if a health check fails, subsequent requests respect the `interval` wait time instead of immediately retrying in a tight loop.

Screenshot 2026-01-09 at 6 25 59 PM

### 3. Functional Correctness
*   **HTTP 500 Handling**: Updated `executeWithCircuitBreaker` to treat HTTP 500+ status codes as failures. Previously, only network errors (`err != nil`) triggered the circuit breaker, causing it to remain closed during server-side outages.

**Checklist:**


docs/references/metrics/page.md
```diff
@@ -0,0 +1,57 @@
+# Metrics Reference
```
aryanmehrotra: lets also add the link to the grafana dashboard

pkg/gofr/service/circuit_breaker.go
```diff
@@ -122,6 +123,10 @@ func (cb *circuitBreaker) startHealthChecks() {
 func (cb *circuitBreaker) openCircuit() {
 	cb.state = OpenState
 	cb.lastChecked = time.Now()
+
+	if cb.metrics != nil {
```
aryanmehrotra, line 131: The metric addition is useful, but tracking only the open count feels a bit incomplete.

A counter for open events tells us how many times the breaker tripped, but not what state it’s currently in. From a debugging point of view, it’s hard to know whether the circuit is open right now, briefly flapping, or already recovered.

aryanmehrotra: Having both current state (gauge) and transitions (counters) makes alerting and debugging much clearer, while still keeping the metrics simple.

pkg/gofr/service/retry.go
```diff
@@ -98,11 +107,15 @@ func (rp *retryProvider) doWithRetry(reqFunc func() (*http.Response, error)) (*h
...
 
-	for i := 0; i < rp.maxRetries; i++ {
+	for i := 0; i <= rp.maxRetries; i++ {
 		resp, err = reqFunc()
 		if err == nil && resp.StatusCode != http.StatusInternalServerError {
 			return resp, nil
 		}
+
+		if i > 0 && rp.metrics != nil {
+			rp.metrics.IncrementCounter(context.Background(), "app_http_retry_count")
```
aryanmehrotra: should we not have label such that which service is causing retry etc? as there can be different serice, only the count won't help

pkg/gofr/service/circuit_breaker.go
aryanmehrotra: We should also tell for which service its open

docs/navigation.js
```diff
@@ -273,6 +273,11 @@ export const navigation = [
                 href: '/docs/references/testing',
                 desc: "GoFr provides a centralized collection of mocks to facilitate writing effective unit tests. Explore testing strategies and tools for GoFr applications, ensuring the code is robust, reliable, and maintainable."
             },
+            {
```
Umang01-hash: We already have this list here: https://gofr.dev/docs/quick-start/observability

No need to add a new entry we can update it in the above doc.

pkg/gofr/container/container.go
```diff
@@ -240,6 +240,9 @@ func (c *Container) registerFrameworkMetrics() {
 		httpBuckets := []float64{.001, .003, .005, .01, .02, .03, .05, .1, .2, .3, .5, .75, 1, 2, 3, 5, 10, 30}
 		c.Metrics().NewHistogram("app_http_response", "Response time of HTTP requests in seconds.", httpBuckets...)
 		c.Metrics().NewHistogram("app_http_service_response", "Response time of HTTP service requests in seconds.", httpBuckets...)
+
```
Umang01-hash: We don't need a line break here all of them are http service metrics only.

pkg/gofr/gofr.go
```diff
@@ -210,6 +210,8 @@ func (a *App) AddHTTPService(serviceName, serviceAddress string, options ...serv
 		a.container.Debugf("Service already registered Name: %v", serviceName)
 	}
 
+	options = append([]service.Options{service.WithName(serviceName)}, options...)
```
Umang01-hash: A small suggestion here, if we are adding this option only to populate the name in our metrics, let's make sure it is extensible such that if we need other things from service like URL etc we should not make a new option for the same.

pkg/gofr/service/circuit_breaker.go
```diff
@@ -74,7 +76,7 @@ func (cb *circuitBreaker) executeWithCircuitBreaker(ctx context.Context, f func(
 	}
 
 	result, err := f(ctx)
-	if err != nil {
+	if err != nil || (result != nil && result.StatusCode >= 500) {
```
aryanmehrotra: it should be >

PR #2825, opened 2026-01-08
feat: Add support for exporting metrics in Cron Jobs

### LINKED ISSUE #2651: Add Support for exporting metrics in GoFr Cron Jobs
Cron jobs Metrics

- app_cron_job_duration (histogram)
- app_cron_job_total (counter)
- app_cron_job_success (counter)
- app_cron_job_failures (counter)

Kind of metrics to expose are open to discussion.

iyashjayesh:
This PR implements metric collection for Cron Jobs in GoFr. 

It adds the following metrics to track the execution and health of cron jobs:

app_cron_job_duration (Histogram): Duration of each cron job execution.
app_cron_job_total (Counter): Total number of times a cron job has started.
app_cron_job_success (Counter): Total number of successful executions.
app_cron_job_failures (Counter): Total number of failed executions (panics).

Closes #2651

@coolwednesday ^

### DIFF
--- a/examples/using-add-filestore/go.mod
+++ b/examples/using-add-filestore/go.mod
@@ -17,7 +17,7 @@ require (
 	cloud.google.com/go/iam v1.5.3 // indirect
 	cloud.google.com/go/pubsub v1.50.1 // indirect
 	cloud.google.com/go/pubsub/v2 v2.0.0 // indirect
-	filippo.io/edwards25519 v1.1.0 // indirect
+	filippo.io/edwards25519 v1.1.1 // indirect
 	github.com/DATA-DOG/go-sqlmock v1.5.2 // indirect
 	github.com/XSAM/otelsql v0.41.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
--- a/examples/using-add-filestore/go.sum
+++ b/examples/using-add-filestore/go.sum
@@ -17,8 +17,8 @@ cloud.google.com/go/pubsub v1.50.1 h1:fzbXpPyJnSGvWXF1jabhQeXyxdbCIkXTpjXHy7xviB
 cloud.google.com/go/pubsub v1.50.1/go.mod h1:6YVJv3MzWJUVdvQXG081sFvS0dWQOdnV+oTo++q/xFk=
 cloud.google.com/go/pubsub/v2 v2.0.0 h1:0qS6mRJ41gD1lNmM/vdm6bR7DQu6coQcVwD+VPf0Bz0=
 cloud.google.com/go/pubsub/v2 v2.0.0/go.mod h1:0aztFxNzVQIRSZ8vUr79uH2bS3jwLebwK6q1sgEub+E=
-filippo.io/edwards25519 v1.1.0 h1:FNf4tywRC1HmFuKW5xopWpigGjJKiJSV0Cqo0cJWDaA=
-filippo.io/edwards25519 v1.1.0/go.mod h1:BxyFTGdWcka3PhytdK4V28tE5sGfRvvvRV7EaN4VDT4=
+filippo.io/edwards25519 v1.1.1 h1:YpjwWWlNmGIDyXOn8zLzqiD+9TyIlPhGFG96P39uBpw=
+filippo.io/edwards25519 v1.1.1/go.mod h1:BxyFTGdWcka3PhytdK4V28tE5sGfRvvvRV7EaN4VDT4=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/DATA-DOG/go-sqlmock v1.5.2 h1:OcvFkGmslmlZibjAjaHm3L//6LiuBgolP7OputlJIzU=
 github.com/DATA-DOG/go-sqlmock v1.5.2/go.mod h1:88MAG/4G7SMwSE3CeA0ZKzrT5CiOU3OJ+JlNzwDqpNU=
--- a/go.mod
+++ b/go.mod
@@ -56,7 +56,7 @@ require (
 	cloud.google.com/go/compute/metadata v0.9.0 // indirect
 	cloud.google.com/go/iam v1.5.3 // indirect
 	cloud.google.com/go/pubsub/v2 v2.0.0 // indirect
-	filippo.io/edwards25519 v1.1.0 // indirect
+	filippo.io/edwards25519 v1.1.1 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cenkalti/backoff/v5 v5.0.3 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -17,8 +17,8 @@ cloud.google.com/go/pubsub v1.50.1 h1:fzbXpPyJnSGvWXF1jabhQeXyxdbCIkXTpjXHy7xviB
 cloud.google.com/go/pubsub v1.50.1/go.mod h1:6YVJv3MzWJUVdvQXG081sFvS0dWQOdnV+oTo++q/xFk=
 cloud.google.com/go/pubsub/v2 v2.0.0 h1:0qS6mRJ41gD1lNmM/vdm6bR7DQu6coQcVwD+VPf0Bz0=
 cloud.google.com/go/pubsub/v2 v2.0.0/go.mod h1:0aztFxNzVQIRSZ8vUr79uH2bS3jwLebwK6q1sgEub+E=
-filippo.io/edwards25519 v1.1.0 h1:FNf4tywRC1HmFuKW5xopWpigGjJKiJSV0Cqo0cJWDaA=
-filippo.io/edwards25519 v1.1.0/go.mod h1:BxyFTGdWcka3PhytdK4V28tE5sGfRvvvRV7EaN4VDT4=
+filippo.io/edwards25519 v1.1.1 h1:YpjwWWlNmGIDyXOn8zLzqiD+9TyIlPhGFG96P39uBpw=
+filippo.io/edwards25519 v1.1.1/go.mod h1:BxyFTGdWcka3PhytdK4V28tE5sGfRvvvRV7EaN4VDT4=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/DATA-DOG/go-sqlmock v1.5.2 h1:OcvFkGmslmlZibjAjaHm3L//6LiuBgolP7OputlJIzU=
 github.com/DATA-DOG/go-sqlmock v1.5.2/go.mod h1:88MAG/4G7SMwSE3CeA0ZKzrT5CiOU3OJ+JlNzwDqpNU=
--- a/pkg/gofr/datasource/dbresolver/go.mod
+++ b/pkg/gofr/datasource/dbresolver/go.mod
@@ -18,7 +18,7 @@ require (
 	cloud.google.com/go/iam v1.5.3 // indirect
 	cloud.google.com/go/pubsub v1.50.1 // indirect
 	cloud.google.com/go/pubsub/v2 v2.0.0 // indirect
-	filippo.io/edwards25519 v1.1.0 // indirect
+	filippo.io/edwards25519 v1.1.1 // indirect
 	github.com/DATA-DOG/go-sqlmock v1.5.2 // indirect
 	github.com/XSAM/otelsql v0.41.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
--- a/pkg/gofr/datasource/dbresolver/go.sum
+++ b/pkg/gofr/datasource/dbresolver/go.sum
@@ -17,8 +17,8 @@ cloud.google.com/go/pubsub v1.50.1 h1:fzbXpPyJnSGvWXF1jabhQeXyxdbCIkXTpjXHy7xviB
 cloud.google.com/go/pubsub v1.50.1/go.mod h1:6YVJv3MzWJUVdvQXG081sFvS0dWQOdnV+oTo++q/xFk=
 cloud.google.com/go/pubsub/v2 v2.0.0 h1:0qS6mRJ41gD1lNmM/vdm6bR7DQu6coQcVwD+VPf0Bz0=
 cloud.google.com/go/pubsub/v2 v2.0.0/go.mod h1:0aztFxNzVQIRSZ8vUr79uH2bS3jwLebwK6q1sgEub+E=
-filippo.io/edwards25519 v1.1.0 h1:FNf4tywRC1HmFuKW5xopWpigGjJKiJSV0Cqo0cJWDaA=
-filippo.io/edwards25519 v1.1.0/go.mod h1:BxyFTGdWcka3PhytdK4V28tE5sGfRvvvRV7EaN4VDT4=
+filippo.io/edwards25519 v1.1.1 h1:YpjwWWlNmGIDyXOn8zLzqiD+9TyIlPhGFG96P39uBpw=
+filippo.io/edwards25519 v1.1.1/go.mod h1:BxyFTGdWcka3PhytdK4V28tE5sGfRvvvRV7EaN4VDT4=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/DATA-DOG/go-sqlmock v1.5.2 h1:OcvFkGmslmlZibjAjaHm3L//6LiuBgolP7OputlJIzU=
 github.com/DATA-DOG/go-sqlmock v1.5.2/go.mod h1:88MAG/4G7SMwSE3CeA0ZKzrT5CiOU3OJ+JlNzwDqpNU=
--- a/pkg/gofr/datasource/oracle/go.mod
+++ b/pkg/gofr/datasource/oracle/go.mod
@@ -20,7 +20,7 @@ require (
 	cloud.google.com/go/iam v1.5.3 // indirect
 	cloud.google.com/go/pubsub v1.50.1 // indirect
 	cloud.google.com/go/pubsub/v2 v2.0.0 // indirect
-	filippo.io/edwards25519 v1.1.0 // indirect
+	filippo.io/edwards25519 v1.1.1 // indirect
 	github.com/VictoriaMetrics/easyproto v0.1.4 // indirect
 	github.com/XSAM/otelsql v0.41.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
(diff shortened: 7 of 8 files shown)


Umang01-hash (commented):
@iyashjayesh Thanks for the effort to make GoFr better. 

But currently in the PR the pull-based metrics model **(:2121/metrics) requires app.Run() to start the server.** This doesn't work for short-lived/batch cron jobs—the process exits before Prometheus can scrape.


The [example](https://github.com/gofr-dev/gofr/tree/development/examples/using-cron-jobs) demonstrates this exact pattern (cron runs, sleeps, exits) but metrics are lost.

We need to handle both the cases:

1. Long-running crons → Current implementation works with app.Run()
2. Short-lived/batch crons →Maybe  add push-based export support.

The fix should be transparent to users. Just like HTTP apps don't require extra calls for metrics, cron jobs shouldn't either.

iyashjayesh:
Thanks for the feedback, @Umang01-hash 

I’ll try implementing this to handle it transparently:
- Config-based: If `METRICS_PUSH_GATEWAY_URL` is set, the code will automatically switch to a push exporter; otherwise, it will default to the existing pull model.
- Graceful flush: I’ll add a shutdown hook in `Container.Close()` to force a final metrics push, ensuring short-lived jobs are captured before exit.

Let me know if this makes sense.

iyashjayesh:
@Umang01-hash I’m facing issues running the lint checks locally.

I have `golangci-lint` installed with the version below, but when I run it on my machine I get a lot of lint errors that do not appear in the GitHub lint checks.

Could you please guide me on whether we follow a specific setup or command for running lint checks locally before committing code changes?

Also, I’d like to propose adding a simple Makefile to standardize common tasks like linting, testing, running the project, running tests for a specific package/file, and any required code generation. This would make local development more consistent.

```
(base) yash@Yashs-MacBook-Pro gofr % golangci-lint --version
golangci-lint has version 2.5.0 built with go1.25.1 from ff63786 on 2025-09-21T18:53:13Z
```

Umang01-hash:
> @Umang01-hash I’m facing issues running the lint checks locally.
> 
> I have `golangci-lint` installed with the version below, but when I run it on my machine I get a lot of lint errors that do not appear in the GitHub lint checks.
> 
> Could you please guide me on whether we follow a specific setup or command for running lint checks locally before committing code changes?
> 
> Also, I’d like to propose adding a simple Makefile to standardize common tasks like linting, testing, running the project, running tests for a specific package/file, and any required code generation. This would make local development more consistent.
> 
> ```
> (base) yash@Yashs-MacBook-Pro gofr % golangci-lint --version
> golangci-lint has version 2.5.0 built with go1.25.1 from ff63786 on 2025-09-21T18:53:13Z
> ```

Hi @iyashjayesh,

Thank you for the detailed breakdown and for suggesting improvements.

### Why you see different lint results locally

Our CI workflow (see below for reference) **always runs golangci-lint version v2.4.0** for both root module and for any changed submodules.  
If you use a different version (like v2.5.0), you may see extra or different errors due to changes in the tool’s default behaviors, rules, or how configs are interpreted.

**To avoid discrepancies, please:**

1. **Install golangci-lint v2.4.0:**  
   ```sh
   go install github.com/golangci/golangci-lint/cmd/golangci-lint@v2.4.0
   ```

2. **Run lint from repo root using our config:**  
   ```sh
   golangci-lint run  .golangci.yml
   ```
   (and for submodules, run from their respective folders if needed).

3. **Make sure your Go version matches what’s used in CI (Go 1.25 for linting).**

---

### **Regarding Makefile Proposal**

**We agree!**  
A Makefile helps developers run lint, test, etc. with the exact arguments/settings used in CI, ensuring consistency for everyone.  
Please feel free to propose one in a PR—we’re happy to help refine and standardize it for the team.

---

### **Relevant CI Workflow Section**  
For reference, here’s how our workflow runs lint checks:
```yaml
- name: Lint Root Module
  uses: golangci/golangci-lint-action@v9
  with:
    version: v2.4.0
    only-new-issues: true
```

Submodules are also linted with v2.4.0 if changes are detected.

---

#### **Summary**

- Always use v2.4.0 (not newer) for golangci-lint—this matches CI.
- Use the repo’s `.golangci.yml` for config.
- Matching Go/linter versions and flags solves ~all linting mismatches!
- We welcome your Makefile proposal for further consistency.

Let us know if these steps resolve your issue, or if you run into any other problems.

iyashjayesh:
@Umang01-hash I’ve fixed the lint issues and addressed the review suggestions.

Based on our discussion, I’ve also added a Makefile. Please review it and let me know your thoughts. Thanks 
https://github.com/gofr-dev/gofr/blob/57e460aa69c2e369654f3b8d19b6708416948023/Makefile

Umang01-hash:
@iyashjayesh Thank you for your thoughtful implementation and the clear rationale behind introducing the `METRICS_PUSH_GATEWAY_URL` config for short-lived cron jobs.

After an internal discussion and consideration of GoFr’s design philosophy, we’d like to propose an alternative approach and get your input before finalizing:

#### Prefer Log-Based Metric Export for Cron Jobs

- **Consistent GoFr Configuration**: GoFr typically abstracts configuration to simple ports, **_not full URLs or external endpoints_**. Requiring users to set up and manage a Pushgateway—and provide an explicit URL—marks a significant departure from our "batteries-included" approach.
- **Simplicity & Universality**: Exporting metrics to logs works anywhere, regardless of network/firewall/topology, and requires no extra infrastructure. It leverages existing log aggregation pipelines and works well for one-off jobs or environments (like serverless and batch) where exposing a port or running a push client isn’t practical.
- **Observability Best Practices**: Emitting metrics in a clear, structured log line (e.g., prefixed with `[GOFR_METRICS]`) allows users to easily parse and forward these to any backend (Prometheus via sidecar, Loki, ELK, Datadog, etc.).

#### Suggested Next Steps

- Replace the Pushgateway configuration with logic to print/export metrics to logs at the end of the cron job’s execution (ideally using a well-documented, parse-friendly format).
- Add documentation on how users can extract and ingest these metrics using standard log aggregation tooling.

We believe this approach strikes the best balance between operational simplicity, GoFr’s philosophy, and observability needs. Let’s also keep the Pushgateway idea in mind as an advanced, opt-in future capability!

Let us know your thoughts, and thank you again for pushing the observability envelope in GoFr!

iyashjayesh:
@Umang01-hash Thanks for the feedback. Agreed. Log-based metric export makes sense and is much simpler for short-lived jobs (serverless, batch, etc.).

I’ll update the PR to:
- Replace Pushgateway with a log exporter with [GOFR_METRICS]
- Flush metrics to logs at the end of each cron execution
- will also add docs on parsing and ingesting these metrics

I’ll keep Pushgateway aside as a possible advanced feature later. Let me know if this direction works.

Umang01-hash:
> @Umang01-hash Thanks for the feedback. Agreed. Log-based metric export makes sense and is much simpler for short-lived jobs (serverless, batch, etc.).
> 
> I’ll update the PR to:
> 
> * Replace Pushgateway with a log exporter with [GOFR_METRICS]
> * Flush metrics to logs at the end of each cron execution
> * will also add docs on parsing and ingesting these metrics
> 
> I’ll keep Pushgateway aside as a possible advanced feature later. Let me know if this direction works.

@iyashjayesh That sounds great, thanks for being flexible!

For the log formatting: Please take a look at how GoFr currently prints logs for requests and other activities. Ideally, you should print metrics as a structured log entry at the end of each cron execution—something like `[GOFR_METRICS] {json payload}` so it’s easy to parse and consistent with our existing log style.

Let’s keep this PR focused on exporting metrics via logs only—no Pushgateway integration for now. We can revisit Pushgateway as a separate, advanced feature in the future if there’s demand.

Looking forward to the updated PR!

iyashjayesh:
Thanks @Umang01-hash! Noted!

iyashjayesh:
@Umang01-hash pushed changes WRT log-based metric export mechanism for short-lived workloads like cron jobs. 

By setting `METRICS_LOG_EXPORT=true`, the application will emit metrics as a structured JSON payload to the logs when the metrics are flushed.

Please review it and let me know your thoughts. Thanks

Umang01-hash:
@iyashjayesh Thanks for the quick turnaround and for implementing the log-based metrics export for cron jobs.

**One important change before we move forward:**
- The config flag (`METRICS_LOG_EXPORT=true`) for enabling metrics in logs should not be required. In GoFr, metrics export is a default behavior (for HTTP/GRPC, metrics are always exported to a port). For short-lived jobs (cron), exporting metrics to logs at the end of execution should also be default, with *no extra config needed*.
- Please remove the toggle and ensure all cron jobs always emit the `[GOFR_METRICS] {json}` line at completion.

iyashjayesh:
> @iyashjayesh Thanks for the quick turnaround and for implementing the log-based metrics export for cron jobs.
> 
> **One important change before we move forward:**
> 
> * The config flag (`METRICS_LOG_EXPORT=true`) for enabling metrics in logs should not be required. In GoFr, metrics export is a default behavior (for HTTP/GRPC, metrics are always exported to a port). For short-lived jobs (cron), exporting metrics to logs at the end of execution should also be default, with _no extra config needed_.
> * Please remove the toggle and ensure all cron jobs always emit the `[GOFR_METRICS] {json}` line at completion.

@Umang01-hash thanks for the review. I've updated the approach. Please review it and let me know your thoughts. Thanks

pkg/gofr/metrics/exporters/exporter.go
```diff
@@ -1,9 +1,15 @@
 package exporters
 
 import (
+	"context"
+	"encoding/json"
+	"strconv"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
```
Umang01-hash: the alias name **dto** is not very clear. Can we please rename it?

pkg/gofr/metrics/exporters/exporter.go
```diff
@@ -1,9 +1,15 @@
...
 	"github.com/prometheus/otlptranslator"
 	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/exporters/prometheus"
+	otelprom "go.opentelemetry.io/otel/exporters/prometheus"
```
Umang01-hash: Why added alias to this already existing import?

pkg/gofr/metrics/exporters/exporter.go
```diff
@@ -12,12 +18,19 @@ import (
 	"gofr.dev/pkg/gofr/version"
 )
 
-func Prometheus(appName, appVersion string) metric.Meter {
-	exporter, err := prometheus.New(
-		prometheus.WithoutTargetInfo(),
-		prometheus.WithTranslationStrategy(otlptranslator.NoTranslation))
+type Logger interface {
```
Umang01-hash: We already have this **Logger** interface defined in file `resgiter.go.` Can we please use that only instead of defining a new one here:

```golang
// Logger defines a simple interface for logging messages at different log levels.
type Logger interface {
	Error(args ...any)
	Errorf(format string, args ...any)
	Warn(args ...any)
	Warnf(format string, args ...any)
	Info(args ...any)
	Infof(format string, args ...any)
}
```

pkg/gofr/metrics/exporters/exporter.go
```diff
@@ -12,12 +18,19 @@ import (
...
+	Infof(format string, args ...any)
+}
+
+func Prometheus(appName, appVersion string, logger Logger) (metric.Meter, func(context.Context) error, prometheus.Gatherer) {
```
Umang01-hash: The function named `Prometheus` is not very clear of what it is intended for. Can be name it in better way?

pkg/gofr/metrics/handler.go
```diff
@@ -6,15 +6,24 @@ import (
...
 
 	"github.com/gorilla/mux"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 // GetHandler creates a new HTTP handler that serves metrics collected by the provided metrics manager to the '/metrics' route.
 func GetHandler(m Manager) http.Handler {
-	var router = mux.NewRouter()
+	var (
```
Umang01-hash: @iyashjayesh Can you confirm that all changes in `metrics/handler.go` are necessary? Asking it because the file **handler.go** is responsible for exposing the metrics handler in case of http server etc , Since cron jobs don’t expose HTTP, we should strive to keep server and job logic separate.

Umang01-hash (changes requested):
@iyashjayesh Please provide screenshots showing metric output in a cron job scenario, and update documentation/release notes for new usage.

iyashjayesh:
> @iyashjayesh Please provide screenshots showing metric output in a cron job scenario, and update documentation/release notes for new usage.

Screenshot 2026-01-30 at 11 34 15 AM

```
INFO [11:32:33] Loaded config from file: ./configs/.env
INFO [11:32:33] Exporting traces to zipkin at http://localhost:2005/api/v2/spans
INFO [11:32:34] Starting cron job: counter
INFO [11:32:34] [Count: 1]
INFO [11:32:34] [GOFR_METRICS] {"app_cron_job_duration":[{"buckets":{"10":1,"100":1,"1000":1,"500":1,"5000":1},"count":1,"labels":{"job":"counter","otel_scope_name":"sample-cron","otel_scope_schema_url":"","otel_scope_version":"dev"},"sum":0}],"app_cron_job_success":[{"labels":{"job":"counter","otel_scope_name":"sample-cron","otel_scope_schema_url":"","otel_scope_version":"dev"},"value":1}],"app_cron_job_total":[{"labels":{"job":"counter","otel_scope_name":"sample-cron","otel_scope_schema_url":"","otel_scope_version":"dev"},"value":1}],"app_info":[{"labels":{"app_name":"sample-cron","app_version":"dev","framework_version":"dev","otel_scope_name":"sample-cron","otel_scope_schema_url":"","otel_scope_version":"dev"},"value":1}]}
INFO [11:32:34] Finished cron job: counter in 999.375µs
INFO [11:32:35] Starting cron job: counter
INFO [11:32:35] [Count: 2]
INFO [11:32:35] [GOFR_METRICS] {"app_cron_job_duration":[{"buckets":{"10":2,"100":2,"1000":2,"500":2,"5000":2},"count":2,"labels":{"job":"counter","otel_scope_name":"sample-cron","otel_scope_schema_url":"","otel_scope_version":"dev"},"sum":0}],"app_cron_job_success":[{"labels":{"job":"counter","otel_scope_name":"sample-cron","otel_scope_schema_url":"","otel_scope_version":"dev"},"value":2}],"app_cron_job_total":[{"labels":{"job":"counter","otel_scope_name":"sample-cron","otel_scope_schema_url":"","otel_scope_version":"dev"},"value":2}],"app_info":[{"labels":{"app_name":"sample-cron","app_version":"dev","framework_version":"dev","otel_scope_name":"sample-cron","otel_scope_schema_url":"","otel_scope_version":"dev"},"value":1}]}
INFO [11:32:35] Finished cron job: counter in 596.375µs
```

@Umang01-hash attached SS and logs for reference. Will address your other review comments meanwhile

pkg/gofr/metrics/exporters/exporter.go
iyashjayesh: @Umang01-hash The alias `otelprom` is necessary because there is a name collision in the imports. I'm importing two different packages that both use the name prometheus. Without the alias, the compiler wouldn't know which one to refer to. Let me know if this makes sense.

pkg/gofr/metrics/exporters/exporter.go
iyashjayesh: @Umang01-hash I've renamed the alias from dto to promModel. I've updated the code.

pkg/gofr/metrics/exporters/exporter.go
iyashjayesh: @Umang01-hash Updated the code to reuse the Logger interface.

pkg/gofr/metrics/exporters/exporter.go
iyashjayesh: I've renamed the function

pkg/gofr/metrics/handler.go
iyashjayesh: Yes, the changes in `handler.go` are necessary. The new metrics implementation uses a custom Prometheus registry to support metric gathering and flushing (used for the log-based metrics in cron jobs).

Because metrics are now registered in this custom registry instead of the `prometheus.DefaultRegistry`, we must update `GetHandler` to use the custom gatherer from the `metrics.Manager`. Without this change, the `/metrics` endpoint would stop showing application metrics when the new exporters are active.

While cron jobs themselves don't provide an HTTP endpoint, this change ensures that the GoFr metrics server correctly exposes metrics whenever the custom registry is used.

iyashjayesh:
@Umang01-hash, you can resolve the comments if you feel the updates make sense to you.

Umang01-hash:
Sure @iyashjayesh will re-review the PR and let you know.

iyashjayesh:
> @iyashjayesh, hey!
> 
> I was reviewing the implementation, and although we have a log-level–based approach, I feel it introduces unnecessary complexity. Since we already have a metrics server running, we could simply add a metric there instead. I’d lean toward not adopting @Umang01-hash’s suggestion for now. We can always introduce this added complexity later if we encounter a real need or issue that justifies it.
> 
> Apologies for repeated architectural changes.

Hi @coolwednesday, I’ll check this and get back to you. I’ve been occupied with some personal work over the past few days.
We’ll respond soon. 

@Umang01-hash I’d also appreciate your overview on this before we move ahead and finalize it. That will help us avoid repeated work.

Umang01-hash:
> > @iyashjayesh, hey!
> > I was reviewing the implementation, and although we have a log-level–based approach, I feel it introduces unnecessary complexity. Since we already have a metrics server running, we could simply add a metric there instead. I’d lean toward not adopting @Umang01-hash’s suggestion for now. We can always introduce this added complexity later if we encounter a real need or issue that justifies it.
> > Apologies for repeated architectural changes.
> 
> Hi @coolwednesday, I’ll check this and get back to you. I’ve been occupied with some personal work over the past few days. We’ll respond soon.
> 
> @Umang01-hash I’d also appreciate your overview on this before we move ahead and finalize it. That will help us avoid repeated work.

Yes @iyashjayesh sorry for the to and fro that has been done in this PR but for the cron i think log based approach was really not the best solution to it and we could have simply added the metric there instead. Sorry for my previous comment lets finalize this approachh now.

pkg/gofr/cron.go
```diff
@@ -104,16 +106,48 @@ func (j *job) run(cntnr *container.Container) {
...
 		}
 
 		c.Infof("Finished cron job: %s in %s", j.name, time.Since(start))
 	}()
 
+	if m := cntnr.Metrics(); m != nil {
+		m.IncrementCounter(ctx, "app_cron_job_total", "job", j.name)
+	}
+
 	j.fn(c)
```
coolwednesday, line 131: In `cron.go`, the deferred function computes `duration` correctly at the top, but then calls `time.Since(start)` **again** at the bottom for the log message:

```go
defer func() {
    duration := time.Since(start).Seconds()  // ← computed here (correct)

    if m := cntnr.Metrics(); m != nil {
        m.RecordHistogram(ctx, "app_cron_job_duration", float64(duration), ...)
        // ...
    }

    c.Infof("Finished cron job: %s in %s", j.name, time.Since(start))  // ← computed AGAIN (wrong)
}()
```

The log message will always show a slightly *longer* duration than the histogram because it re-computes `time.Since(start)` after all the deferred logic (metrics recording, error logging, etc.) has run.

Umang01-hash: Sure changed the log to use the `duration` declared above only.

Umang01-hash:
> The `else if r := recover()` branch in `cron.go` (the path taken when `cntnr.Metrics() == nil` AND the job panics) has **no test coverage**. The existing panic test always uses a `mockContainer` with metrics configured.

Added test case for this scenario also.

PR #2829, opened 2026-01-12
Fix/circuit breaker

Umang01-hash:
## Pull Request Template

**Bug:**

-  When a user configured `HealthConfig` with a custom health endpoint, the circuit breaker still used the default **/.well-known/alive endpoint** for recovery checks. If this default endpoint didn't exist on the downstream service, the circuit breaker would never recover (stay permanently open).

**Solution:**
- The circuit breaker now automatically uses the same health endpoint configured in HealthConfig for its recovery checks. The user only needs to configure the health endpoint once.

**Checklist:**

### DIFF
--- a/pkg/gofr/service/circuit_breaker.go
+++ b/pkg/gofr/service/circuit_breaker.go
@@ -99,14 +99,13 @@ func (cb *circuitBreaker) isOpen() bool {
 }
 
 func (cb *circuitBreaker) healthCheck(ctx context.Context) bool {
-	var resp *Health
-
-	// Read health config from parent httpService if available
 	if httpSvc := extractHTTPService(cb.HTTP); httpSvc != nil && httpSvc.healthEndpoint != "" {
-		resp = cb.HTTP.getHealthResponseForEndpoint(ctx, httpSvc.healthEndpoint, httpSvc.healthTimeout)
-	} else {
-		resp = cb.HTTP.HealthCheck(ctx)
+		resp := cb.HTTP.getHealthResponseForEndpoint(ctx, httpSvc.healthEndpoint, httpSvc.healthTimeout)
+
+		return resp.Status == serviceUp
 	}
+
+	resp := cb.HTTP.HealthCheck(ctx)
 
 	return resp.Status == serviceUp
 }


pkg/gofr/service/health_config.go
```diff
@@ -13,13 +13,36 @@ func (h *HealthConfig) AddOption(svc HTTP) HTTP {
...
+		updateCircuitBreakerHealthConfig(v.HTTP, endpoint, timeout)
+	case *rateLimiter:
+		updateCircuitBreakerHealthConfig(v.HTTP, endpoint, timeout)
+	case *customHeader:
+		updateCircuitBreakerHealthConfig(v.HTTP, endpoint, timeout)
+	case *customHealthService:
+		updateCircuitBreakerHealthConfig(v.HTTP, endpoint, timeout)
+	}
+}
+
```
coolwednesday: we can have case and default case only ?

pkg/gofr/service/circuit_breaker.go
```diff
@@ -76,7 +82,7 @@ func (cb *circuitBreaker) executeWithCircuitBreaker(ctx context.Context, f func(
 	}
 
 	result, err := f(ctx)
-	if err != nil || (result != nil && result.StatusCode > 500) {
+	if err != nil || (result != nil && result.StatusCode >= 500) {
```
coolwednesday: It was decided that we will open circuit breaker only on 500 + errors.

Umang01-hash: done

pkg/gofr/service/health_config.go
Umang01-hash: We could use **reflection in the default case** to auto-traverse wrapper types with an HTTP field. However, this introduces: 
(1) silent runtime failures if a field is renamed, 
(2) fragile FieldByName("HTTP") dependency, 
(3) performance overhead. 

The explicit type switch ensures compile-time safety and forces developers to consciously handle new wrapper types.

coolwednesday: ok sure.

coolwednesday (dismissed):

aryanmehrotra (changes requested):
You are currently updating the circuit breaker configuration inside the health-check, which is not a good practice.

At present, the HTTP configuration hierarchy looks like this:

Level 1: HTTP

Level 2: Individual Options

Each option at Level 2 is expected to be independent and should not depend on or modify another option. However, the current implementation violates this principle by allowing one option (health-check) to influence another (circuit breaker).

If a particular option requires a value that is also needed by another option, that value should be owned and managed by the parent HTTP configuration, not by one of the child options. This keeps the responsibilities clearly separated and maintains proper encapsulation.

With the current approach, if we introduce a new option in the future that also depends on health-check behavior, we would again need to modify the health-check configuration. This makes the design harder to extend and increases coupling.

Moving the shared configuration to the HTTP parent level will make the system more modular, easier to reason about, and more scalable for future extensions.

coolwednesday (dismissed):

pkg/gofr/service/circuit_breaker.go
```diff
@@ -98,9 +98,15 @@ func (cb *circuitBreaker) isOpen() bool {
...
 func (cb *circuitBreaker) healthCheck(ctx context.Context) bool {
-	resp := cb.HealthCheck(ctx)
+	var resp *Health
+
+	// Read health config from parent httpService if available
+	if httpSvc := extractHTTPService(cb.HTTP); httpSvc != nil && httpSvc.healthEndpoint != "" {
+		resp = cb.HTTP.getHealthResponseForEndpoint(ctx, httpSvc.healthEndpoint, httpSvc.healthTimeout)
+	} else {
+		resp = cb.HTTP.HealthCheck(ctx)
+	}
```
aryanmehrotra, line 106: if we are reading it from the parent HTTPService, they why do we need to extract the HTTP Service?
cb.HTTP.HealthCheck is a method from the parent that would already know which endpoint is required to be called to know the health check

Umang01-hash: With the current design, if **HealthConfig** is applied after **CircuitBreakerConfig**, the chain is:

**customHealthService -> circuitBreaker -> httpService**

In this case, `circuitBreaker.HealthCheck(ctx)` calls the embedded HTTP's HealthCheck, which is `httpService.HealthCheck(ctx)`— not the custom one. So, the custom endpoint is not used.

To ensure the correct endpoint is always called:
**The circuit breaker’s healthCheck should check if the parent httpService has a custom health endpoint set.**

If set, it should call `getHealthResponseForEndpoint` with the custom endpoint and timeout.

Otherwise, it should call the default **HealthCheck**.

aryanmehrotra: In that case for health-check specifically we should update the endpoint and that should be used, it shouldn't depend on the order

We can create a separate ticket for that

coolwednesday: Done. https://github.com/gofr-dev/gofr/issues/2847

PR #2901, opened 2026-02-02
Add GraphQL Support

### LINKED ISSUE #624: GraphQL support
REST and gRPC support now, any plan to consider GraphQL support?   REST, GraphQL and gRPC are most import protocols for service communication.

coolwednesday:
## Summary
- Adds native GraphQL support to GoFr using a **schema-first** approach
- Developers define their API contract via standard GraphQL SDL (`./configs/schema.graphqls`) and register resolvers using `app.GraphQLQuery()` and `app.GraphQLMutation()`
- Resolvers receive `*gofr.Context` with full access to databases, logger, and standard `c.Bind()` for argument binding
- Closes #624

### Schema-First
- Schema file (`./configs/schema.graphqls`) is parsed at startup using `gqlparser` for validation
- Resolver names must match fields defined in the schema's `Query` / `Mutation` types
- Supports scalar types, custom object types, enums, input types, and lists

### Single Endpoint
- All operations served via `POST /graphql` (per GraphQL-over-HTTP spec)
- Interactive GraphQL Playground available at `/.well-known/graphql/ui` (registered whenever GraphQL is enabled)
- Playground uses pinned GraphiQL v3 + React v18 for stability

### Resolver Architecture
- Resolvers are standard `gofr.Handler` functions — same signature as REST handlers
- Arguments are bound via a custom `graphQLRequest` that implements GoFr's `Request` interface, enabling `c.Bind()` to work with GraphQL arguments
- Each resolver runs with its own traced context and panic recovery

### Fail-Fast
- `setupGraphQL()` calls `Fatalf` if schema parsing or building fails — prevents silent startup with broken GraphQL
- Thread-safe lazy initialization of `graphqlManager` in `App`

### Observability
- Metrics (registered only when GraphQL is enabled — opt-in):
  - `app_graphql_operations_total` — counter by `operation_name` and `type`
  - `app_graphql_error_total` — counter for failed operations
  - `app_graphql_request_duration` — histogram for request latency
- Tracing: OTel spans for overall request and per-resolver (e.g., `graphql-resolver-user`)

### Security
- 32MB request body limit
- Panic recovery per resolver

## Files Changed
- `pkg/gofr/graphql.go` — Core implementation (manager, schema builder, resolver binding, handler, playground HTML)
- `pkg/gofr/graphql_test.go` — Unit tests covering schema building, request handling, error scenarios
- `pkg/gofr/gofr.go` — `GraphQLQuery()`, `GraphQLMutation()`, `setupGraphQL()`, playground route registration
- `pkg/gofr/responder.go` — `noopResponder` for resolver context consistency
- `pkg/gofr/http/middleware/metrics.go` — GraphQL route pattern for metrics middleware
- `pkg/gofr/container/container.go` — Minor addition
- `docs/advanced-guide/graphql/page.md` — Documentation with usage, monitoring, design, and limitations
- `examples/using-graphql/` — Working example with schema, resolvers, migrations, and integration tests

## Test Plan
- [x] Unit tests in `graphql_test.go` — schema building, request parsing, error handling, variables, named operations, malformed queries
- [x] Integration tests in `examples/using-graphql/main_test.go` — end-to-end query/mutation execution, error scenarios, playground availability
- [x] `golangci-lint run ./pkg/gofr/` — 0 issues
- [x] `go test ./pkg/gofr/` — all tests pass


pkg/gofr/graphql.go
```diff
@@ -0,0 +1,535 @@
...
+	// Detect operation type
+	opType := "query"
+	if astDoc, err := parser.ParseQuery(&ast.Source{Input: req.Query}); err == nil && len(astDoc.Operations) > 0 {
+		opType = string(astDoc.Operations[0].Operation)
+	}
+
+	ctx, span := m.tracer.Start(r.Context(), "graphql-request")
+	span.SetAttributes(attribute.String("graphql.operation_name", opName), attribute.String("graphql.operation_type", opType))
+
+	defer func() {
```
Umang01-hash, line 435: `OperationName` is parsed above (line 416) but never passed to the executor. Requests that send `operationName` will either fail or execute the wrong operation when a document has multiple named operations.

```go
result := graphql.Do(graphql.Params{
    Schema:         m.schema,
    RequestString:  req.Query,
    VariableValues: req.Variables,
    OperationName:  req.OperationName, // missing
    Context:        ctx,
})
```

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,535 @@
...
+		span.End()
+	}()
+
+	m.container.Metrics().IncrementCounter(ctx, "gofr_graphql_operations_total", "operation_name", opName, "type", opType)
+
+	result := graphql.Do(graphql.Params{
+		Schema:         m.schema,
+		RequestString:  req.Query,
+		VariableValues: req.Variables,
+		Context:        ctx,
```
Umang01-hash, line 453: `Content-Type: application/json` is only set on the 400 path. The 422 and 200 paths write JSON without setting the header. Add `w.Header().Set("Content-Type", "application/json")` before `w.WriteHeader` in all branches.

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,535 @@
...
+
+		// Prepare arguments for function call. Only Context is passed.
+		args := []reflect.Value{reflect.ValueOf(c)}
+
+		results := v.Call(args)
+
+		err := results[1].Interface()
+		if err != nil {
+			m.container.Errorf("GraphQL Resolver %s failed: %v", name, err)
+			return nil, err.(error)
```
Umang01-hash: Schema is built on the first HTTP request via `sync.Once`. GoFr initialises everything eagerly at startup and fails fast on misconfigurations — this should follow the same pattern. Move `buildSchema()` into `httpServerSetup()`. If the file is missing or unparseable, log the error and abort startup. An additional problem: if building fails, `sync.Once` permanently caches the error and the schema can never be retried without a full restart.

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,535 @@
...
+	c.Metrics().NewCounter("gofr_graphql_operations_total", "total Number of GraphQL operations received")
+	c.Metrics().NewCounter("gofr_graphql_error_total", "total Number of GraphQL operations that returned an error")
+	c.Metrics().NewHistogram("gofr_graphql_request_duration", "execution time of GraphQL requests",
+		0.001, 0.005, 0.01, 0.05, 0.1, 0.5, 1, 5, 10) //nolint:mnd // histogram buckets
+
+	return &graphQLManager{
+		container: c,
+		queries:   make(map[string]any),
+		mutations: make(map[string]any),
+		tracer:    otel.Tracer("gofr-graphql"),
```
Umang01-hash, line 95: GoFr already has `type Handler func(c *Context) (any, error)` in handler.go. Use that instead of `any`. It gives compile-time type safety and removes the need for reflection on every resolver call. Consistent with how `app.GET()`, `app.POST()` etc. register handlers.

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,535 @@
...
+	if ok && def.Kind == ast.InputObject {
+		fields := graphql.InputObjectConfigFieldMap{}
+
+		for _, f := range def.Fields {
+			fields[f.Name] = &graphql.InputObjectFieldConfig{
+				Type: m.mapInputType(f.Type, schema),
+			}
+		}
+
+		return graphql.NewInputObject(graphql.InputObjectConfig{
```
Umang01-hash, line 256: Enum types fall through here and silently get mapped to `String`. Same issue at line 317 for output types. A user with `enum Role { ADMIN USER }` in their schema gets no error — just broken behavior. Add proper `ast.Enum` handling, or at minimum return an error instead of falling back silently.

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,535 @@
...
+const (
+	gofrKey = "gofr"
+	stringT = "String"
+	idT     = "ID"
+	intT    = "Int"
+	floatT  = "Float"
+	boolT   = "Boolean"
+)
+
+type graphQLManager struct {
```
Umang01-hash, line 72: `schemaBuilt` is set to `true` at line 176 but never read anywhere. Dead field — remove it.

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,535 @@
...
+	b, err := json.Marshal(r.params)
+	if err != nil {
+		return err
+	}
+
+	return json.Unmarshal(b, v)
+}
+
+func (r *graphQLRequest) Context() context.Context { return r.ctx }
+func (*graphQLRequest) HostName() string           { return "" }
```
Umang01-hash, line 553: `graphql-playground-react` on npm is archived and unmaintained. Loading from a CDN also breaks in air-gapped environments and is a supply-chain risk. Switch to GraphiQL (actively maintained) and embed the assets using `embed.FS` so there's no external dependency at runtime.

pkg/gofr/gofr.go
```diff
@@ -233,6 +266,28 @@ func (a *App) Logger() logging.Logger {
 	return a.container.Logger
 }
 
+// Container returns the container instance associated with the App.
+func (a *App) Container() *container.Container {
```
Umang01-hash: The field has this comment right above it: *"container is unexported because this is an internal implementation and applications are provided access to it via Context"*. Adding a public `Container()` getter directly contradicts that design. This method is also not needed for the GraphQL feature — remove it.

pkg/gofr/gofr.go
```diff
@@ -233,6 +266,28 @@ func (a *App) Logger() logging.Logger {
...
+	return a.container
+}
+
+// Handle registers an http.Handler for a specific pattern.
+func (a *App) Handle(pattern string, handler http.Handler) {
```
Umang01-hash: Two problems:
1. `PathPrefix("/graphql")` matches `/graphql`, `/graphql/ui`, and `/graphqlsomething`. Use `Path` for exact matching.
2. This method exposes raw `http.Handler` registration that bypasses GoFr's middleware stack. It's not related to GraphQL — please don't add it here.

examples/using-graphql/main_test.go
```diff
@@ -0,0 +1,110 @@
...
+	metricsPort := testutil.GetFreePort(t)
+
+	t.Setenv("HTTP_PORT", strconv.Itoa(httpPort))
+	t.Setenv("METRICS_PORT", strconv.Itoa(metricsPort))
+	t.Setenv("APP_ENV", "dev")
+
+	host := fmt.Sprintf("http://localhost:%d", httpPort)
+
+	go main()
+	time.Sleep(500 * time.Millisecond) // Wait for server and migrations
```
Umang01-hash: `time.Sleep` for server readiness is flaky. Use a poll loop hitting `/health` with a timeout. Also both `TestIntegration_GraphQL` and `TestIntegration_GraphQL_Production` call `go main()` with no way to stop the server — they'll interfere with each other if run in the same binary.

Umang01-hash (changes requested):
Overall the approach is solid — schema-first, context parity, built-in observability. But there are several correctness bugs and architectural issues that need to be addressed before this is ready.

Blocking issues:
- Named operations are broken (OperationName not passed to executor)
- Missing Content-Type header on 200/422 responses
- Schema initialization is lazy (first request pays the cost, no fail-fast at startup)
- Handler type should be GoFr's own `Handler` type, not `any` with reflection
- `Container()` and `Handle()` on App are unrelated to GraphQL and break encapsulation
- Enum/Union/Interface types silently fall back to String

See inline comments for specifics.

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,520 @@
...
+	"net/http"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/graphql-go/graphql"
+	"github.com/vektah/gqlparser/v2"
+	"github.com/vektah/gqlparser/v2/ast"
+	"github.com/vektah/gqlparser/v2/parser"
```
gizmo-rt, line 19: Looks highly optimistic here. Is their 1 standard library which includes both parsing and execution  that can be used here ?

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,520 @@
...
+
+func (m *graphQLManager) RegisterMutation(name string, handler Handler) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	m.mutations[name] = handler
+}
+
+func (m *graphQLManager) buildSchema() error {
+	schemaContent, err := os.ReadFile("./configs/schema.graphqls")
```
gizmo-rt, line 116: Curious, how are we planning to standardise this among consumers ?

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,520 @@
...
+
+func (m *graphQLManager) getCustomOutputType(name string, schema *ast.Schema) graphql.Output {
+	def, ok := schema.Types[name]
+	if def != nil && def.Kind == ast.Enum {
+		return m.getEnum(def)
+	}
+
+	if !ok || def.Kind != ast.Object {
+		return graphql.String // Fallback
+	}
```
gizmo-rt, line 335: How would the consumer be notified if schema type checks fail ?

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,520 @@
...
+		typeCache: make(map[string]graphql.Output),
+		enumCache: make(map[string]*graphql.Enum),
+	}
+}
+
+func (m *graphQLManager) RegisterQuery(name string, handler Handler) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	m.queries[name] = handler
```
gizmo-rt, line 105: What are the tradeoffs of this implementation in a distributed context ?

pkg/gofr/graphql_test.go
```diff
@@ -0,0 +1,343 @@
...
+func TestGraphQL_Query(t *testing.T) {
+	t.Setenv("METRICS_PORT", "0")
+	setupSchema(t, `type Query { hello: String }`)
+
+	app := New()
+	app.GraphQLQuery("hello", func(_ *Context) (any, error) {
+		return "world", nil
+	})
+
+	reqBody := `{"query": "{ hello }"}`
```
gizmo-rt, line 43: Graphql queries are deeply nested. Can you add testcases for the same ?

pkg/gofr/graphql_test.go
```diff
@@ -0,0 +1,343 @@
...
+				ID   int    `json:"id"`
+				Role string `json:"role"`
+			} `json:"user"`
+		} `json:"data"`
+	}
+
+	err := json.Unmarshal(resp.Body.Bytes(), &result)
+	require.NoError(t, err)
+	assert.Equal(t, "ADMIN", result.Data.User.Role)
+}
```
gizmo-rt, line 333: Is it possible to include tests for the edge cases ?

coolwednesday:
Addressing the review-level comment from @gizmo-rt:

**1. Design Scope** — This is a **Phase 1 / minimal viable implementation**, covering:
- Queries + Mutations (no Subscriptions yet)
- Custom object types, input types, enums
- Schema-first with `./configs/schema.graphqls`
- Automatic health check injection
- Playground UI (non-production only)
- Full observability (OTel traces + Prometheus metrics)

**Not in scope** (future iterations): Subscriptions, custom scalars, DataLoader/N+1 prevention, field-level authorization, Federation.

**2. Assumptions & Tradeoffs** — We use two libraries (gqlparser for SDL parsing, graphql-go for execution) because GoFr registers resolvers dynamically via `app.GraphQLQuery("name", handler)` — this requires function-based resolvers that only graphql-go supports. Added a detailed architecture comment at the top of `graphql.go` documenting this decision and the alternatives evaluated.

pkg/gofr/graphql.go
coolwednesday: Dug deep into this one. Short answer: **not one that works with our API design.**

The main candidate is `graph-gophers/graphql-go` (v1.9.0, actively maintained, already in our go.mod as indirect). It handles both parsing and execution — but it resolves fields by matching Go method names on a struct, meaning you need compile-time methods like `func (r *Resolver) User(...)`.

GoFr registers handlers dynamically at runtime via `app.GraphQLQuery("user", func(c *gofr.Context) (any, error) { ... })`. Go does not support creating methods on structs at runtime, so there is a fundamental mismatch — we would need code generation or a completely different public API.

The current two-library setup (gqlparser for parsing + graphql-go/graphql for execution) is the necessary trade-off for keeping the Handler-based API that is consistent with `app.GET()`, `app.POST()` etc. Added a detailed architecture comment at the top of `graphql.go` explaining this decision and the alternatives we evaluated.

pkg/gofr/graphql.go
coolwednesday: The standardisation is through GoFr's existing `Handler` type — same one used for REST and CLI:

1. **Signature**: `func(c *gofr.Context) (any, error)` — consistent across all GoFr handlers
2. **Schema file**: `./configs/schema.graphqls` — single source of truth
3. **Registration**: `app.GraphQLQuery("name", handler)` / `app.GraphQLMutation("name", handler)` — mirrors `app.GET()`, `app.POST()`
4. **Args**: Always via `c.Bind(&args)` — same as HTTP handlers
5. **Fail-fast**: Schema field with no resolver → app won't start (Fatalf)

So if a developer knows how to write a REST handler in GoFr, they already know 90% of the GraphQL handler pattern. The schema file is the only new concept.

pkg/gofr/graphql.go
coolwednesday: Three levels of notification:

1. **Schema parse errors** (bad SDL syntax) → `Fatalf` at startup — app won't start
2. **Missing resolvers** (schema field with no handler) → `Fatalf` at startup — app won't start. Added `TestGraphQL_MissingResolver` to cover this.
3. **Unknown types** (not String/Int/Float/Boolean/Object/Enum) → Now logs `Warnf` during schema build so it's visible in startup logs:
   `WARN: unsupported GraphQL output type 'MyType': falling back to String.`
4. **Runtime mismatches** (resolver returns wrong shape) → Standard GraphQL `errors` array in the 200 response + `gofr_graphql_error_total` metric incremented

Previously the output type fallback was completely silent — fixed that.

pkg/gofr/graphql.go
coolwednesday: Good question. The main things to be aware of:

- **Schema consistency**: Each instance loads `schema.graphqls` from its local filesystem at startup. During rolling deploys, different instances might briefly serve different schemas. Same risk as any config file.
- **Stateless execution**: No server-side query caching or persisted queries. Each request is fully self-contained — actually good for horizontal scaling.
- **No federation**: Single-service GraphQL only. No Apollo Federation / schema stitching. Phase 2+ concern.
- **No subscriptions**: WebSocket-based real-time queries not supported yet.

For Phase 1, the stateless per-instance approach is the right call — simple, predictable, scales the same way REST endpoints do.

pkg/gofr/graphql_test.go
coolwednesday: Added two:

- **`TestGraphQL_NestedObjectTypes`** — 3-level deep query: `Author → [Book] → Publisher`, with list types and nested field assertions. Exercises the `typeCache` and recursive `mapType` paths.
- **`TestGraphQL_NestedInputTypes`** — Mutation with nested input: `CreateUserInput → AddressInput`, validates that nested argument binding works through `c.Bind()`.

Both pass and cover the areas that were not tested before (circular reference handling, list-of-objects, multi-level field resolution).

pkg/gofr/graphql_test.go
coolwednesday: Added 10 edge case tests:

| Test | What it covers |
|------|---------------|
| `TestGraphQL_Variables` | Variables via `{"variables": {"n": "GoFr"}}` |
| `TestGraphQL_OperationName` | Multiple named operations, selecting one via operationName |
| `TestGraphQL_EmptyQueryBody` | Empty query string |
| `TestGraphQL_UnknownFields` | Requesting fields not in schema |
| `TestGraphQL_IntrospectionQuery` | `__schema` introspection |
| `TestGraphQL_NullVariables` | `"variables": null` |
| `TestGraphQL_InvalidJSONBody` | Malformed JSON → 400 with JSON error body |
| `TestGraphQL_MutationWithoutMutationSchema` | Mutation on query-only schema |
| `TestGraphQL_MissingResolver` | Schema field with no handler → build error |
| `TestGraphQL_ContentTypeOnAllPaths` | Content-Type on 200, 400, 500 paths |

All 22 GraphQL tests pass.

pkg/gofr/graphql.go
gizmo-rt: @Umang01-hash Thoughts on this

gizmo-rt:
@coolwednesday Thanks for sharing. Is there a post launch ticket which consists of future enhancements ?

pkg/gofr/graphql.go
gizmo-rt: @Umang01-hash Do you see any hurdles with static path ?

Umang01-hash: @gizmo-rt No i dont see any hurdles as GoFr is opinionated, so we can preserve `/graphql` as fixed by design, just like `/health` and `/metrics` are not configurable by consumers.

pkg/gofr/graphql.go
Umang01-hash: @gizmo-rt the real reason two libraries are needed: `graphql-go/graphql`  does support **dynamic function-based resolvers without struct methods** (no built-in SDL parser — it only builds schemas programmatically).  , so `gqlparser` is needed purely to parse `schema.graphqls`. But since schema-first is a deliberate design choice here, both libraries are justified

docs/advanced-guide/graphql/page.md
```diff
@@ -0,0 +1,182 @@
...
+```go
+func main() {
+    app := gofr.New()
+
+    app.GraphQLQuery("user", func(c *gofr.Context) (any, error) {
+        // Extract arguments manually
+        var args struct {
+            ID int `json:"id"`
+        }
+        _ = c.Bind(&args)
```
aryanmehrotra: we should handle errors, should not promote this practise from our docs

docs/advanced-guide/graphql/page.md
```diff
@@ -0,0 +1,182 @@
...
+GoFr follows the standard GraphQL-over-HTTP convention by returning `200 OK` for all successfully processed requests, including those with resolver errors. This ensures that the response body is the source of truth for execution results.
+
+| Status Code | Condition |
+|---|---|
+| `200 OK` | The request was processed (regardless of whether it returned data or errors). |
+| `400 Bad Request` | The request body is not valid JSON. |
+
+**Error response body**:
+```json
+{
```
aryanmehrotra, line 106: is this in sync with our rest api responses?

docs/advanced-guide/graphql/page.md
```diff
@@ -0,0 +1,182 @@
...
+
+### 4. Supported Types
+GoFr supports all standard GraphQL scalar types (`String`, `Int`, `Float`, `Boolean`, `ID`), as well as `Object` and `Enum` types. These are automatically mapped between your GraphQL schema and Go structs/maps.
+
+---
+
+## Testing Your GraphQL API
+
+### 1. Interactive Exploration
+GoFr automatically hosts a **GraphQL Playground** at `/graphql/ui` in all non-production environments. Set `APP_ENV=production` to disable it. The route is not registered in production — it will return `404 Not Found`.
```
aryanmehrotra: should use the well-known endpoints

docs/quick-start/observability/page.md
```diff
@@ -228,6 +228,24 @@ GoFr publishes metrics to port: _2121_ on _/metrics_ endpoint in Prometheus form
...
+
+---
+
+- gofr_graphql_error_total
+- counter
+- Total number of GraphQL operations that returned an error
+
+---
+
+- gofr_graphql_request_duration
```
aryanmehrotra: what are the different labels?

examples/using-graphql/main.go
```diff
@@ -0,0 +1,74 @@
...
+			return nil, err
+		}
+
+		return User{ID: int(id), Name: args.Name, Role: args.Role}, nil
+	})
+
+	return app
+}
+
+func main() {
```
aryanmehrotra: why?

pkg/gofr/gofr.go
```diff
@@ -369,3 +395,19 @@ func (a *App) AddStaticFiles(endpoint, filePath string) {
...
+	if a.graphqlManager != nil {
+		a.graphqlManager.buildErr = a.graphqlManager.buildSchema()
+		if a.graphqlManager.buildErr != nil {
+			a.container.Logger.Fatalf("GraphQL build error: %v", a.graphqlManager.buildErr)
+		}
+
+		a.httpServer.router.NewRoute().Methods(http.MethodPost).Path("/graphql").Handler(a.graphqlManager.GetHandler())
+
+		// Only register GraphQL Playground UI in non-production environments
+		if a.Config.Get("APP_ENV") != "production" {
```
aryanmehrotra: will be removed

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,520 @@
...
+	errResolverMissing = errors.New("resolver missing for field")
+)
+
+const (
+	healthKey = "health"
+	stringT   = "String"
+	idT       = "ID"
+	intT      = "Int"
+	floatT    = "Float"
+	boolT     = "Boolean"
```
aryanmehrotra: can we have better names

pkg/gofr/graphql.go
```diff
@@ -0,0 +1,520 @@
...
+	mu        sync.RWMutex
+	tracer    trace.Tracer
+	typeCache map[string]graphql.Output
+	enumCache map[string]*graphql.Enum
+}
+
+func newGraphQLManager(c *container.Container) *graphQLManager {
+	c.Metrics().NewCounter("gofr_graphql_operations_total", "total Number of GraphQL operations received")
+	c.Metrics().NewCounter("gofr_graphql_error_total", "total Number of GraphQL operations that returned an error")
+	c.Metrics().NewHistogram("gofr_graphql_request_duration", "execution time of GraphQL requests",
```
aryanmehrotra: is it in sync with http bucket?

aryanmehrotra:
Lets make sure everything is in sync with how REST API's are such that the graphql/ui, is part of well-known and error struct is similar to the current REST specs, docs and comments is updated to clearly explain why we have new methods instead to reusing POST and GET,

docs/advanced-guide/graphql/page.md
coolwednesday: Errors in graphQL varies from HTTP semantics

pkg/gofr/graphql.go
coolwednesday: Yes. in HTTP Middleware we now exempt the graphql endpoints.

docs/quick-start/observability/page.md
coolwednesday: added

docs/advanced-guide/graphql/page.md
```diff
@@ -0,0 +1,221 @@
...
+        var args struct {
+            ID int `json:"id"`
+        }
+
+        if err := c.Bind(&args); err != nil {
+            return nil, err
+        }
+
+        // Return 'any' - GoFr validates this against the schema at runtime
+        return map[string]any{
```
aryanmehrotra: we should use struct to return

docs/advanced-guide/graphql/page.md
```diff
@@ -0,0 +1,221 @@
...
+      "path": ["fieldName"]
+    }
+  ]
+}
+```
+
+### 3. Argument Binding
+Instead of declarative arguments in the function signature, you use the standard `c.Bind()` method. GoFr automatically maps the GraphQL `args` map to your struct using JSON tags.
+
+### 4. Supported Types
```
aryanmehrotra, line 121: is it neccessary to tell, if we are supporting everything? we can give hyper link for graphql docs so use can refer and directly create from that, the schema file

examples/using-graphql/configs/.env
```diff
@@ -0,0 +1,13 @@
+METRICS_PORT=0
```
aryanmehrotra: lets send metrics as well

pkg/gofr/responder.go
```diff
@@ -5,3 +5,7 @@ package gofr
 type Responder interface {
 	Respond(data any, err error)
 }
+
+type noopResponder struct{}
+
+func (noopResponder) Respond(_ any, _ error) {}
```
aryanmehrotra, line 15: why do we need this

examples/using-graphql/main.go
```diff
@@ -0,0 +1,71 @@
...
+		if err := c.Bind(&args); err != nil {
+			return nil, err
+		}
+
+		var user User
+
+		err := c.SQL.QueryRowContext(c, "SELECT id, name, role FROM users WHERE id = ?", args.ID).
+			Scan(&user.ID, &user.Name, &user.Role)
+		if err != nil {
+			return nil, err
```
aryanmehrotra, line 42: should this be not binded in gofr errors?

coolwednesday:
## Observability Screenshots — GraphQL Example

Ran the `using-graphql` example with MySQL, executed queries (success + error + bad request), and captured logs and metrics.

### 1. Application Startup Logs
Screenshot 2026-03-13 at 12 38 07 PM

### 3. GraphQL-Specific Metrics (from `:2122/metrics`)

**Counters:**
Screenshot 2026-03-13 at 12 42 28 PM


Screenshot 2026-03-13 at 12 43 04 PM
Screenshot 2026-03-13 at 12 42 48 PM
Screenshot 2026-03-13 at 5 38 09 PM



Size |
|---|---|
| `http-server` (no GraphQL) | 51M |
| `http-sevrver` (with GraphQL) | 53M |

No binary size increase — GraphQL dependencies are already part of the GoFr core module.

### 6. Playground

GraphQL Playground is served at `/.well-known/graphql/ui` (uses GraphiQL 3.0.6).

coolwednesday:
**Re: [why do we need `noopResponder`](https://github.com/gofr-dev/gofr/pull/2901#discussion_r2912105710)**

GraphQL resolvers reuse `*gofr.Context` which requires a `Responder`. But GraphQL doesn't use the standard HTTP response path — the resolver result is collected by the `graphql-go` engine and serialized as part of the unified GraphQL JSON response (`{data, errors}`).

`noopResponder` satisfies the interface without writing anything, since the GraphQL handler (`handleGraphQLRequest`) owns the `http.ResponseWriter` directly.

pkg/gofr/responder.go
coolwednesday: https://github.com/gofr-dev/gofr/pull/2901#issuecomment-4053345269

PR #2958, opened 2026-02-05
Circuit breaker causing HTTP requests to execute sequentially instead of in parallel

### LINKED ISSUE #2889: Circuit Breaker causes parallel requests to execute sequentially
**Describe the bug**
When the circuit breaker is enabled, HTTP requests that normally execute in parallel start behaving sequentially.

For example, Without the circuit breaker, two simultaneous requests complete in ~10 seconds total. After enabling the circuit breaker, the same two requests take ~20 seconds total, indicating that the second request waits for the first one to finish.

**To Reproduce**

1. The code is

   ```go
   &gofrService.CircuitBreakerConfig{Threshold: 6, Interval: 5 * time.Second}
   ```

**Screenshots**

### 1️⃣ Without Circuit Breaker (Parallel Execution)
Without CB

### 2️⃣ With Circuit Breaker Enabled (Sequential Execution)
With CB

First Image (without CB): Both requests complete in ~10s and total time is ~10s, confirming parallel execution.
Second Image (with CB): One request completes at ~10s and the other at ~20s, total time ~20s, showing the circuit breaker caused sequential execution.

**Environments **
- gofr version 1.25.0
- go version 1.54.0

Umang01-hash:
## Pull Request Template

**Description:**

-   Resolves #2889
-   The `executeWithCircuitBreaker()` function held a mutex lock **during the entire HTTP request execution**, blocking all concurrent requests from proceeding.

```go
Request 1 enters executeWithCircuitBreaker(), acquires the mutex lock
Request 2 tries to enter, blocks waiting for the lock
Request 1 makes the HTTP call (takes 10s) while holding the lock 🔴
Request 1 releases lock after completion
Request 2 finally acquires lock and makes its call (takes another 10s)
Result: Sequential execution (20s total) instead of parallel (10s total)
```

-   Fix: Release mutex BEFORE making the HTTP request, Acquire mutex AFTER the HTTP request completes

**Checklist:**

### DIFF
--- a/pkg/gofr/datasource/redis/pubsub_test.go
+++ b/pkg/gofr/datasource/redis/pubsub_test.go
@@ -174,9 +174,10 @@ func setupMockTest(t *testing.T, conf map[string]string) (*testRedisClient, redi
 	redisConfig := getRedisConfig(config.NewMockConfig(conf), mockLogger)
 
 	r := &Redis{
-		Client: db,
-		config: redisConfig,
-		logger: mockLogger,
+		Client:     db,
+		config:     redisConfig,
+		logger:     mockLogger,
+		stopSignal: make(chan struct{}),
 	}
 	ps := newPubSub(db, redisConfig, mockLogger, mockMetrics)
 
--- a/pkg/gofr/datasource/redis/redis.go
+++ b/pkg/gofr/datasource/redis/redis.go
@@ -82,8 +82,10 @@ type StreamsConfig struct {
 
 type Redis struct {
 	*redis.Client
-	logger datasource.Logger
-	config *Config
+	logger     datasource.Logger
+	config     *Config
+	stopSignal chan struct{}
+	closeOnce  sync.Once
 }
 
 // PubSub handles Redis PubSub operations.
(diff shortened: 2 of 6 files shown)


pkg/gofr/service/circuit_breaker.go
```diff
@@ -60,33 +60,35 @@ func NewCircuitBreaker(config CircuitBreakerConfig, h HTTP) *circuitBreaker {
...
+	defer cb.mu.Unlock()
+
 	if err != nil || (result != nil && result.StatusCode > 500) {
 		cb.handleFailure()
+
+		if cb.failureCount > cb.threshold {
+			cb.openCircuit()
+
+			return nil, ErrCircuitOpen
+		}
```
coolwednesday, line 85: The `cb.handleFailure()` method (called on line 81) already checks the threshold and calls `cb.openCircuit()` if necessary. Calling `cb.openCircuit()` again here is redundant and may result in duplicate metrics being recorded for the same circuit break event.

**Suggestion:** 
Simplify the block to only check the state change:
```go
cb.handleFailure()

if cb.state == OpenState {
    return nil, ErrCircuitOpen
}
```

PR #3004, opened 2026-02-17
feat(migration): add Oracle migration locks management

### LINKED ISSUE #2893: Add migration locks for other RDBMS that is supported
We have added locks for MYSQL and Redis. we would like to add this for all datasources that have transactions like Oracle etc. You can refer to migrations package and implement this.

QUBITABHAY:
**Description:**

- Added SQL statements to create and manage a migration locks table in Oracle.
- Implemented logic to acquire, refresh, and release locks during migration processes.
- Introduced a background refresh mechanism for locks to ensure they remain valid.
- Enhanced error handling for lock acquisition and release operations.

fixes #2893


pkg/gofr/migration/oracle.go
```diff
@@ -198,10 +216,55 @@ func (om oracleMigrator) beginTransaction(c *container.Container) transactionDat
...
+
+		expiresAt := time.Now().UTC().Add(defaultLockTTL)
+
+		if err := om.Oracle.Exec(ctx, insertOracleLock, lockKey, ownerID, expiresAt); err != nil {
+			c.Debug("Oracle lock acquired successfully")
+			
+			go om.startRefresh(ctx, cancel, c, ownerID)
+		}
+
+		c.Debugf("Oracle lock already held, retrying in %v... (attempt %d)", defaultRetry, i+1)
```
coolwednesday, line 257: In `pkg/gofr/migration/oracle.go`, the `lock()` method incorrectly interprets the database result:

```go
if err := om.Oracle.Exec(ctx, insertOracleLock, lockKey, ownerID, expiresAt); err != nil {
    c.Debug("Oracle lock acquired successfully")
    go om.startRefresh(ctx, cancel, c, ownerID)
}
```

*   **Issue:** The condition `err != nil` is used to signal success. In reality, an error during `INSERT` usually signifies a `Unique Constraint` violation (meaning the lock is already held by someone else).
*   **Impact:** Multiple application instances will believe they have acquired the lock when they haven't. They will simultaneously run migrations, leading to collision and potential database state corruption.
*   **Fix:** Change to `if err == nil`.

pkg/gofr/migration/oracle.go
```diff
@@ -198,10 +216,55 @@ func (om oracleMigrator) beginTransaction(c *container.Container) transactionDat
...
 func (om oracleMigrator) lock(ctx context.Context, cancel context.CancelFunc, c *container.Container, ownerID string) error {
-	return om.migrator.lock(ctx, cancel, c, ownerID)
+	for i := 0; ; i++ {
+		_ = om.Oracle.Exec(ctx, deleteExpiredOracleLocks, time.Now().UTC())
+
+		expiresAt := time.Now().UTC().Add(defaultLockTTL)
+
+		if err := om.Oracle.Exec(ctx, insertOracleLock, lockKey, ownerID, expiresAt); err != nil {
+			c.Debug("Oracle lock acquired successfully")
+
```
coolwednesday: In `lock()`, the success condition is currently reversed:
```go
if err := om.Oracle.Exec(ctx, insertOracleLock, lockKey, ownerID, expiresAt); err != nil {
    // This block is entered if there is an error (e.g., lock already exists)
    c.Debug("Oracle lock acquired successfully")
}
```
The condition should be `if err == nil`. Currently, if the `INSERT` fails (which happens when another instance already holds the lock), this incorrectly treats it as a successful acquisition.

pkg/gofr/migration/oracle.go
```diff
@@ -198,10 +216,55 @@ func (om oracleMigrator) beginTransaction(c *container.Container) transactionDat
 }
 
 func (om oracleMigrator) lock(ctx context.Context, cancel context.CancelFunc, c *container.Container, ownerID string) error {
-	return om.migrator.lock(ctx, cancel, c, ownerID)
+	for i := 0; ; i++ {
+		_ = om.Oracle.Exec(ctx, deleteExpiredOracleLocks, time.Now().UTC())
```
coolwednesday: The `lock()` method lacks an exit path and delegation:
```go
if err == nil { // assuming fix from above
    c.Debug("Oracle lock acquired successfully")
    go om.startRefresh(ctx, cancel, c, ownerID)
    // Missing: return om.migrator.lock(ctx, cancel, c, ownerID)
}
```
Once the lock is acquired, the function must return by calling the next migrator in the chain: `return om.migrator.lock(ctx, cancel, c, ownerID)`. Without this, the code falls through to the retry logic and will loop indefinitely despite holding the lock.

pkg/gofr/migration/oracle.go
```diff
@@ -198,10 +216,64 @@ func (om oracleMigrator) beginTransaction(c *container.Container) transactionDat
...
+func (oracleMigrator) startRefresh(ctx context.Context, cancel context.CancelFunc, c *container.Container, ownerID string) {
+	ticker := time.NewTicker(defaultRefresh)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			expiresAt := time.Now().UTC().Add(defaultLockTTL)
+			if err := c.Oracle.Exec(ctx, updateOracleLock, expiresAt, lockKey, ownerID); err != nil {
+				c.Error("failed to refresh Oracle lock: %v", err)
```
coolwednesday: **`c.Error` does not format — `%v` will appear literally in logs.**

`c.Error(...)` is a variadic log function (like `fmt.Print`), not a format function. Passing a format string and `err` as separate arguments means the log output will literally print `%v` instead of the actual error value.

Change to `c.Errorf`:
```go
c.Errorf("failed to refresh Oracle lock: %v", err)
```
This is consistent with how every other error is logged in this file (lines 170, 179, 205, 234, 272).

pkg/gofr/migration/oracle.go
```diff
@@ -55,18 +55,36 @@ FROM gofr_migrations
...
 	err := om.Oracle.Exec(context.Background(), checkAndCreateOracleMigrationTable)
 	if err != nil {
-		c.Errorf("Failed to create Oracle migration table: %v", err)
-	} else {
-		c.Infof("Oracle migration table checked/created successfully")
+		return err
 	}
 
-	return err
+	if err := om.Oracle.Exec(context.Background(), checkAndCreateOracleMigrationLocksTable); err != nil {
```
coolwednesday, line 98: **Adding this `Exec` call breaks the existing `Test_OracleCheckAndCreateMigrationTable` test — it's not mocked.**

`checkAndCreateMigrationTable` now makes two `Exec` calls, but the test (`oracle_test.go:46`) only mocks one:
```go
// Current test only mocks the first call:
mockOracle.EXPECT().Exec(gomock.Any(), checkAndCreateOracleMigrationTable).Return(tc.err)
```

With gomock in strict mode, the unmocked second call (`checkAndCreateOracleMigrationLocksTable`) will cause the test to **panic/fail** on the `"no error"` path. The `"connection failed"` case passes only because the first `Exec` returning an error short-circuits before this line.

The test must be updated to mock both calls, and a new failure case should be added for when the locks table creation itself fails:
```go
// Success path:
mockOracle.EXPECT().Exec(gomock.Any(), checkAndCreateOracleMigrationTable).Return(nil)
mockOracle.EXPECT().Exec(gomock.Any(), checkAndCreateOracleMigrationLocksTable).Return(nil)

// Locks table failure path (new):
mockOracle.EXPECT().Exec(gomock.Any(), checkAndCreateOracleMigrationTable).Return(nil)
mockOracle.EXPECT().Exec(gomock.Any(), checkAndCreateOracleMigrationLocksTable).Return(sql.ErrConnDone)
```

pkg/gofr/migration/oracle.go
```diff
@@ -198,10 +216,64 @@ func (om oracleMigrator) beginTransaction(c *container.Container) transactionDat
 }
 
 func (om oracleMigrator) lock(ctx context.Context, cancel context.CancelFunc, c *container.Container, ownerID string) error {
```
coolwednesday, line 234: **Zero unit tests for `lock`, `unlock`, and `startRefresh`.**

These three functions implement distributed lock management — the most safety-critical logic in the migration system. A bug here can cause:
- Migrations running concurrently on multiple pods (data corruption)
- Locks that never release (migrations hang forever)
- Goroutine leaks if `startRefresh` isn't cancelled properly

None of the three new functions have any test coverage. Please add tests covering at minimum:
- `lock` acquires successfully on first attempt
- `lock` retries on a duplicate key error and succeeds on the next attempt
- `lock` returns `errLockAcquisitionFailed` on non-duplicate-key errors
- `lock` returns `ctx.Err()` when context is cancelled while waiting to retry
- `unlock` succeeds and calls base migrator
- `unlock` returns `errLockReleaseFailed` on `Exec` error

For reference, `redisMigrator` has analogous tests to follow.

pkg/gofr/migration/oracle.go
```diff
@@ -198,10 +216,64 @@ func (om oracleMigrator) beginTransaction(c *container.Container) transactionDat
...
+		err := om.Oracle.Exec(ctx, insertOracleLock, lockKey, ownerID, expiresAt)
+		if err == nil {
+			c.Debug("Oracle lock acquired successfully")
+
+			go om.startRefresh(ctx, cancel, c, ownerID)
+
+			return om.migrator.lock(ctx, cancel, c, ownerID)
+		}
+
+		if !isDuplicateKeyError(err) {
```
coolwednesday: **`isDuplicateKeyError` was written for MySQL/PostgreSQL/SQLite — not Oracle.**

The function (in `sql.go`) checks for: `"duplicate"`, `"unique constraint"`, `"integrity constraint"`, `"primary key constraint"`, `"constraint failed"`.

Oracle's ORA-00001 via the godror driver typically produces:
```
ORA-00001: unique constraint (SCHEMA.PK_GOFR_MIGRATION_LOCKS) violated
```
This happens to contain `"unique constraint"`, so it works today — but by coincidence. If the driver changes its wrapping or a different godror version emits `"ORA-00001"` without the substring, it will break silently and the lock loop will return `errLockAcquisitionFailed` instead of retrying.

Consider adding an explicit Oracle error code check:
```go
func isOracleDuplicateKeyError(err error) bool {
    msg := strings.ToLower(err.Error())
    return strings.Contains(msg, "ora-00001") || isDuplicateKeyError(err)
}
```

pkg/gofr/migration/oracle.go
```diff
@@ -198,10 +216,64 @@ func (om oracleMigrator) beginTransaction(c *container.Container) transactionDat
 }
 
 func (om oracleMigrator) lock(ctx context.Context, cancel context.CancelFunc, c *container.Container, ownerID string) error {
-	return om.migrator.lock(ctx, cancel, c, ownerID)
+	for i := 0; ; i++ {
+		_ = om.Oracle.Exec(ctx, deleteExpiredOracleLocks, time.Now().UTC())
```
coolwednesday: **Expired lock cleanup errors are silently discarded.**

If the `DELETE` fails (e.g. network blip, session limit hit), stale lock rows remain in the table. The next process trying to acquire the lock will still see the expired row and incorrectly treat it as an active lock until its `expires_at` naturally expires — which delays migrations unnecessarily.

At minimum, log a debug/warn so operators can detect persistent cleanup failures:
```go
if err := om.Oracle.Exec(ctx, deleteExpiredOracleLocks, time.Now().UTC()); err != nil {
    c.Debugf("failed to clean up expired Oracle locks: %v", err)
}
```
This matches the existing pattern in `sql.go` (see `c.Errorf("failed to clean up expired locks: %v", err)`).

pkg/gofr/migration/oracle.go
```diff
@@ -55,18 +55,36 @@ FROM gofr_migrations
...
+END;
+`
+	deleteExpiredOracleLocks = `DELETE FROM gofr_migration_locks WHERE expires_at < :1`
+	insertOracleLock         = `INSERT INTO gofr_migration_locks (lock_key, owner_id, expires_at) VALUES (:1, :2, :3)`
+	updateOracleLock         = `UPDATE gofr_migration_locks SET expires_at = :1 WHERE lock_key = :2 AND owner_id = :3`
+	deleteOracleLock         = `DELETE FROM gofr_migration_locks WHERE lock_key = :1 AND owner_id = :2`
 )
 
 // Create migration table if it doesn't exist.
 func (om oracleMigrator) checkAndCreateMigrationTable(c *container.Container) error {
```
coolwednesday, line 92: **Error logging removed for table creation — loss of diagnostics.**

The previous implementation logged both success and failure:
```go
c.Errorf("Failed to create Oracle migration table: %v", err)
c.Infof("Oracle migration table checked/created successfully")
```
The new version returns errors silently. With two `Exec` calls now in this function, operators have no way to tell from logs which table (`gofr_migrations` or `gofr_migration_locks`) failed. Add targeted error logs:
```go
if err := om.Oracle.Exec(ctx, checkAndCreateOracleMigrationLocksTable); err != nil {
    c.Errorf("Failed to create Oracle migration locks table: %v", err)
    return err
}
```

pkg/gofr/migration/oracle.go
Umang01-hash: Added a new function :
```
// isOracleDuplicateKeyError checks for Oracle unique constraint violation (ORA-00001) in addition
// to the generic patterns checked by isDuplicateKeyError, to avoid relying solely on driver-specific
// message formatting.
func isOracleDuplicateKeyError(err error) bool {
	msg := strings.ToLower(err.Error())
	return strings.Contains(msg, "ora-00001") || isDuplicateKeyError(err)
}
```

Will use this instead of **isDuplicateKeyError**.

pkg/gofr/migration/oracle.go
Umang01-hash: Added the targetted error logs for both of the Exec calls.

```golang
func (om oracleMigrator) checkAndCreateMigrationTable(c *container.Container) error {
	if err := om.Oracle.Exec(context.Background(), checkAndCreateOracleMigrationTable); err != nil {
		c.Errorf("failed to create Oracle migration table: %v", err)
		return err
	}

	if err := om.Oracle.Exec(context.Background(), checkAndCreateOracleMigrationLocksTable); err != nil {
		c.Errorf("failed to create Oracle migration locks table: %v", err)
		return err
	}

	return om.migrator.checkAndCreateMigrationTable(c)
}
```

pkg/gofr/migration/oracle.go
Umang01-hash: Added the tests for the same. `oracle.go` has now a coverage of **91%**.

pkg/gofr/migration/oracle.go
Umang01-hash: Fixed the test **Test_OracleCheckAndCreateMigrationTable** to include both the mock Exec calls.

pkg/gofr/migration/oracle.go
Umang01-hash: Sure corrected!

Umang01-hash (approved):

QUBITABHAY:
Hey @coolwednesday, is everything done or is something required to change?

coolwednesday:
> Hey @coolwednesday, is everything done or is something required to change?

Everything is done.

Issue #2758, opened 2025-12-24
Add support for file metadata and signed URL generation in Gofr GCS file store


NitinKumar004:
**Description**:

We use Gofr's GCS file store to upload  files. Gofr currently supports basic file uploads, but we need:

1. **File metadata support**:
   - Set `Content-Type` (e.g., "text/csv") — tells the browser how to handle the file
   - Set `Content-Disposition` (e.g., "attachment; filename='report.csv'") — forces download with a specific filename
   
   These metadata fields are needed so browsers download files correctly instead of displaying them.

2. **Signed URL generation**:
   - Generate signed URLs for uploaded files with configurable expiry time
   
   This creates time-limited, secure download links that can be shared via email. The URL expires after the set duration, providing secure access without making files public.

These features are supported by GCS, but Gofr's file store doesn't expose them. We currently use the GCS SDK directly, which bypasses Gofr's abstraction.

aryanmehrotra:
both halves of this are in — `pkg/gofr/datasource/file/gcs/storage_adapter.go`.

metadata goes through `CreateWithOptions` (Content-Type, Content-Disposition), and there's signed URL generation with a configurable expiry. the credentials get parsed once at connect time rather than per call, so signing isn't doing JSON+PEM work on every request.

closing. thanks for filing it with the actual use case spelled out — the Content-Disposition detail is the part that's easy to miss otherwise.

PR #3036, opened 2026-02-24
feat(file/gcs): add metadata upload and signed URL generation support

Umang01-hash:
## Summary

-   Adds cloud-specific file operations to GoFr's GCS file store:
         - **File metadata support** — set `Content-Type`, `Content-Disposition`, and custom key-value metadata when uploading files
          - **Signed URL generation** — create time-limited, pre-authenticated download URLs with configurable expiry and optional content-disposition override

-   Closes issue #2758.

----

GoFr's file store currently supports basic CRUD operations (`Create`, `Write`, `ReadDir`, `Remove`), but cloud providers like GCS offer richer functionality that users need:

1. **Metadata** — Setting `Content-Type` and `Content-Disposition` ensures browsers handle downloaded files correctly (e.g., downloading a CSV as an attachment instead of rendering it inline).
2. **Signed URLs** — Generating time-limited, pre-authenticated URLs allows sharing files securely via email or API responses without making bucket objects public or requiring client-side credentials.

Without this, users must drop down to the raw GCS SDK, bypassing GoFr's abstraction, logging, and metrics.

### Changes

- Introduced `file.CloudFileStore` interface with `CreateWithOptions` and `GenerateSignedURL`
- Added `file.AsCloud()` helper to safely type-assert a `FileSystem` to `CloudFileStore`
- Added `file.FileOptions` struct for `ContentType`, `ContentDisposition`, and `Metadata`
- Implemented `CreateWithOptions` and `GenerateSignedURL` in `datasource/file/gcs`

### Example main.go:

```golang
package main

import (
	"fmt"
	"os"
	"time"

	"gofr.dev/pkg/gofr"
	"gofr.dev/pkg/gofr/datasource/file"
	"gofr.dev/pkg/gofr/datasource/file/gcs"
)

func main() {
	app := gofr.New()

	fs := gcs.New(&gcs.Config{
		BucketName:      app.Config.Get("GCS_BUCKET_NAME"),
		CredentialsJSON: loadCredentials(app),
	})

	app.AddFileStore(fs)

	// Standard filesystem operations.
	app.POST("/files", uploadHandler)
	app.GET("/files", listHandler)
	app.DELETE("/files", deleteHandler)

	// Cloud-specific operations (metadata + signed URLs).
	app.POST("/files/metadata", uploadWithMetadataHandler)
	app.GET("/files/sign", signedURLHandler)

	app.Run()
}

func loadCredentials(app *gofr.App) string {
	if path := app.Config.Get("GCS_CREDENTIALS_FILE"); path != "" {
		data, err := os.ReadFile(path)
		if err == nil {
			return string(data)
		}
	}

	return app.Config.Get("GCS_CREDENTIALS_JSON")
}

// ─── Standard handlers ────────────────────────────────────────────────────────

func uploadHandler(c *gofr.Context) (any, error) {
	name := c.Param("name")
	if name == "" {
		return nil, fmt.Errorf("query param 'name' is required")
	}

	var body []byte
	if err := c.Bind(&body); err != nil {
		return nil, fmt.Errorf("read body: %w", err)
	}

	f, err := c.File.Create(name)
	if err != nil {
		return nil, fmt.Errorf("create file: %w", err)
	}
	defer f.Close()

	n, err := f.Write(body)
	if err != nil {
		return nil, fmt.Errorf("write file: %w", err)
	}

	return map[string]any{"file": name, "bytes_written": n}, nil
}

func listHandler(c *gofr.Context) (any, error) {
	prefix := c.Param("prefix")

	entries, err := c.File.ReadDir(prefix)
	if err != nil {
		return nil, fmt.Errorf("list files: %w", err)
	}

	type entry struct {
		Name  string `json:"name"`
		IsDir bool   `json:"is_dir"`
		Size  int64  `json:"size"`
	}

	result := make([]entry, 0, len(entries))
	for _, e := range entries {
		result = append(result, entry{
			Name:  e.Name(),
			IsDir: e.IsDir(),
			Size:  e.Size(),
		})
	}

	return map[string]any{"prefix": prefix, "entries": result}, nil
}

func deleteHandler(c *gofr.Context) (any, error) {
	name := c.Param("name")
	if name == "" {
		return nil, fmt.Errorf("query param 'name' is required")
	}

	if err := c.File.Remove(name); err != nil {
		return nil, fmt.Errorf("delete file: %w", err)
	}

	return map[string]string{"deleted": name}, nil
}

// ─── Cloud-specific handlers ──────────────────────────────────────────────────

type uploadWithMetadataRequest struct {
	Content     string            `json:"content"`
	ContentType string            `json:"content_type"`
	Disposition string            `json:"disposition"`
	Metadata    map[string]string `json:"metadata"`
}

func uploadWithMetadataHandler(c *gofr.Context) (any, error) {
	name := c.Param("name")
	if name == "" {
		return nil, fmt.Errorf("query param 'name' is required")
	}

	cfs, ok := file.AsCloud(c.File)
	if !ok {
		return nil, fmt.Errorf("configured file store does not support metadata; use a cloud provider")
	}

	var req uploadWithMetadataRequest
	if err := c.Bind(&req); err != nil {
		return nil, fmt.Errorf("bind request: %w", err)
	}

	opts := &file.FileOptions{
		ContentType:        req.ContentType,
		ContentDisposition: req.Disposition,
		Metadata:           req.Metadata,
	}

	f, err := cfs.CreateWithOptions(c.Context, name, opts)
	if err != nil {
		return nil, fmt.Errorf("create with metadata: %w", err)
	}
	defer f.Close()

	n, err := f.Write([]byte(req.Content))
	if err != nil {
		return nil, fmt.Errorf("write file: %w", err)
	}

	return map[string]any{
		"file":          name,
		"bytes_written": n,
		"content_type":  req.ContentType,
		"metadata":      req.Metadata,
	}, nil
}

func signedURLHandler(c *gofr.Context) (any, error) {
	name := c.Param("name")
	if name == "" {
		return nil, fmt.Errorf("query param 'name' is required")
	}

	cfs, ok := file.AsCloud(c.File)
	if !ok {
		return nil, fmt.Errorf("configured file store does not support signed URLs; use a cloud provider")
	}

	expiry := 15 * time.Minute
	if raw := c.Param("expiry"); raw != "" {
		parsed, err := time.ParseDuration(raw)
		if err != nil {
			return nil, fmt.Errorf("invalid expiry %q: %w", raw, err)
		}
		expiry = parsed
	}

	disposition := c.Param("disposition")

	var opts *file.FileOptions
	if disposition != "" {
		opts = &file.FileOptions{ContentDisposition: disposition}
	}

	signedURL, err := cfs.GenerateSignedURL(c.Context, name, expiry, opts)
	if err != nil {
		return nil, fmt.Errorf("generate signed URL: %w", err)
	}

	return map[string]any{
		"file":       name,
		"signed_url": signedURL,
		"expires_in": expiry.String(),
	}, nil
}

```

Screenshot 2026-02-24 at 11 35 25 AM

**Checklist:**


pkg/gofr/datasource/file/gcs/storage_adapter.go
```diff
@@ -76,6 +99,22 @@ func (s *storageAdapter) Connect(ctx context.Context) error {
...
+	// Parse and cache service-account credentials for signed URL signing.
+	// Done once at connect time so every SignedURL() call reuses the result instead of
+	// repeating JSON unmarshal + PEM decode + key parse on the hot path.
+	// When CredentialsJSON is absent (Workload Identity / ADC), saEmail and saPrivateKey
+	// remain empty and bucket.SignedURL will use the client's ambient credentials.
+	if s.cfg.CredentialsJSON != "" {
+		email, privateKey, parseErr := parseServiceAccountCredentials(s.cfg.CredentialsJSON)
+		if parseErr != nil {
+			_ = client.Close()
+			return fmt.Errorf("credentials cannot be used for signed URLs: %w", parseErr)
```
coolwednesday: **[Critical] Credential validation blocks GCS connectivity for all users, not just signed-URL users.**

If `CredentialsJSON` is provided but the private key is unparseable (e.g., non-standard format, Workload-Identity-style credentials, differently generated key), `Connect()` fails entirely — even though the GCS client was **already created and validated successfully** (the bucket attrs call above passed).

Users with working GCS credentials who never plan to call `GenerateSignedURL` will find their `Connect()` broken after upgrading. This is a silent regression.

**Suggested fix:** Downgrade to a warning and continue. Move key validation lazily into `SignedURL()` so it only fails when the feature is actually used:
```go
if parseErr != nil {
    // log Warnf: credentials cannot be used for signed URLs: %v; signed URL calls will fail
    // Do NOT return error — bucket access still works fine
}
```

pkg/gofr/datasource/file/gcs/storage_adapter.go
```diff
@@ -149,6 +188,31 @@ func (s *storageAdapter) NewWriter(ctx context.Context, name string) io.WriteClo
 	return s.bucket.Object(name).NewWriter(ctx)
 }
 
+// NewWriterWithOptions implements MetadataWriter.
+func (s *storageAdapter) NewWriterWithOptions(ctx context.Context, name string, opts *file.FileOptions) io.WriteCloser {
+	if name == "" {
+		return &failWriter{err: errEmptyObjectName}
+	}
+
+	w := s.bucket.Object(name).NewWriter(ctx)
```
coolwednesday, line 217: **[Critical] Nil pointer dereference if `Connect()` was never called or failed.**

`s.bucket` will be `nil` if `Connect()` has not succeeded. Calling `s.bucket.Object(name)` on a nil `*storage.BucketHandle` will **panic** at runtime.

Compare with `SignedURL()` which has an explicit guard:
```go
if s.bucket == nil {
    return "", errGCSClientNotInitialized
}
```
`NewWriterWithOptions` needs the same check before reaching `s.bucket.Object(...)`:
```go
if s.bucket == nil {
    return &failWriter{err: errGCSClientNotInitialized}
}
```
Also add a test for a non-empty name with a nil bucket to cover this panic path.

pkg/gofr/datasource/file/local_fs.go
```diff
@@ -179,26 +182,6 @@ func (*localProvider) ListDir(_ context.Context, prefix string) ([]ObjectInfo, [
 	return objects, dirs, nil
 }
 
-type fileSystem struct {
-	*localFileSystem
-}
-
-// New initializes local filesystem with logger (backward compatible wrapper).
-func New(logger datasource.Logger) FileSystem {
```
coolwednesday, line 187: **[Critical] Public API removed without deprecation.**

`file.New(logger)` is exported and explicitly documented as a *"backward compatible wrapper"*. Deleting it here is a breaking change for any caller using this function.

Even on a development branch, removing an API marked as backward-compatible should follow a deprecation cycle:
1. Add `// Deprecated: use NewLocalFileSystem instead.` godoc.
2. Remove it in the next major version bump.

If the intent is to fully consolidate into `NewLocalFileSystem`, please document the migration path in the PR description.

pkg/gofr/datasource/file/common_fs.go
```diff
@@ -646,3 +646,75 @@ func (c *CommonFileSystem) IsRetryDisabled() bool {
...
+// CreateWithOptions creates a file with optional metadata (content-type, content-disposition,
+// custom key-value pairs). If the underlying provider does not implement MetadataWriter,
+// the file is created without metadata and a warning is logged.
+func (c *CommonFileSystem) CreateWithOptions(ctx context.Context, name string, opts *FileOptions) (File, error) {
+	var msg string
+
+	st := StatusError
+	startTime := time.Now()
+
+	defer c.Observe(OpCreate, startTime, &st, &msg)
```
coolwednesday: **[Design] `CreateWithOptions` emits `OpCreate` — indistinguishable from `Create()` in metrics.**

Both `Create()` (line 401) and `CreateWithOptions()` use `OpCreate` as the histogram label. Operators cannot tell how many requests used metadata vs. plain creates in their dashboards.

Suggested fix — add to `observability.go`:
```go
OpCreateWithOptions = "CREATE_WITH_OPTIONS"
```
And use it here instead of `OpCreate`.

pkg/gofr/datasource/file/common_fs.go
```diff
@@ -646,3 +646,75 @@ func (c *CommonFileSystem) IsRetryDisabled() bool {
...
+
+	writer := c.Provider.NewWriter(ctx, name)
+	if writer == nil {
+		msg = "failed to create writer"
+		return nil, errWriterNil
+	}
+
+	st = StatusSuccess
+
+	msg = fmt.Sprintf("Created %q (metadata not supported)", name)
```
coolwednesday, line 689: **[Design] Silent metadata drop recorded as `SUCCESS` — misleading in metrics and logs.**

When `CreateWithOptions` falls back to a plain `NewWriter` (provider doesn't support metadata), the operation is recorded as `StatusSuccess`. But the caller's explicit request — `ContentType`, `ContentDisposition`, custom `Metadata` — was silently ignored.

Files uploaded with the wrong content-type will be invisible in metrics. A subtle production bug.

Consider either:
1. Returning a sentinel error (e.g., `ErrMetadataNotSupported`) so the caller can decide how to handle it.
2. Using a distinct status label (e.g., `"SUCCESS_DEGRADED"`) so it shows up differently in dashboards.

pkg/gofr/datasource/file/gcs/storage_adapter.go
```diff
@@ -290,3 +354,177 @@ func (s *storageAdapter) ListDir(ctx context.Context, prefix string) ([]file.Obj
...
+// of a custom endpoint. This is used to redirect signed URLs to a local emulator
+// (e.g., fake-gcs-server) during integration tests without changing the path or query.
+func rewriteSignedURLEndpoint(signedURL, endpoint string) string {
+	if endpoint == "" {
+		return signedURL
+	}
+
+	ep, err := url.Parse(endpoint)
+	if err != nil {
+		return signedURL
```
coolwednesday, line 535: **[Design] Parse error silently swallowed — emulator users get wrong host with no warning.**

If `endpoint` is misconfigured (e.g. `"://bad-url"`), the function silently returns the original URL pointing to `storage.googleapis.com`. In a test environment this means requests hit real GCS instead of the local emulator, with no indication that anything went wrong.

Since the function returns only `string`, the caller cannot detect failure. Options:
1. Accept a `datasource.Logger` parameter and emit a `Warnf`.
2. Change return type to `(string, error)` and propagate it through `SignedURL()`.

pkg/gofr/datasource/file/interface.go
```diff
@@ -132,7 +142,57 @@ type FileSystemProvider interface {
...
+// AsCloud is a convenience helper for checking if a FileSystemProvider
+// supports cloud-specific features. It's equivalent to type assertion
+// but provides a clearer intent in code.
+//
+// Example:
+//
+//	if cfs, ok := file.AsCloud(fs); ok {
+//	    url, _ := cfs.GenerateSignedURL(ctx, "file.csv", time.Hour, nil)
+//	}
+func AsCloud(fs FileSystem) (CloudFileSystem, bool) {
```
coolwednesday, line 186: **[Design] `AsCloud` accepts `FileSystem` but should accept `FileSystemProvider`.**

`CloudFileSystem` embeds `FileSystemProvider` which embeds `FileSystem`. Accepting the base `FileSystem` is too broad — a struct implementing only `FileSystem` (no `UseLogger`, `UseMetrics`, `Connect`) could be passed in and the assertion would silently return `false`.

The actual usage in the PR example is `file.AsCloud(c.File)` where `c.File` is a `FileSystemProvider`. Tightening the parameter type makes the function self-documenting:
```go
func AsCloud(fs FileSystemProvider) (CloudFileSystem, bool) {
```

pkg/gofr/datasource/file/gcs/storage_adapter.go
```diff
@@ -290,3 +354,177 @@ func (s *storageAdapter) ListDir(ctx context.Context, prefix string) ([]file.Obj
...
+	if expiry <= 0 {
+		return errExpiryMustBePositive
+	}
+
+	if expiry > maxGCSSignedURLExpiry {
+		return errExpiryTooLong
+	}
+
+	if opts != nil && opts.ContentType != "" {
+		if !strings.Contains(opts.ContentType, "/") {
```
coolwednesday: **[Minor] Content-type validation is too permissive.**

`strings.Contains(opts.ContentType, "/")` allows `"/"`, `"application/"`, and `"/json"` — all invalid MIME types. Consider validating both sides of the slash are non-empty:
```go
parts := strings.SplitN(opts.ContentType, "/", 2)
if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
    return fmt.Errorf("%w: %q", errInvalidContentType, opts.ContentType)
}
```

pkg/gofr/datasource/file/common_fs_signedurl_test.go
```diff
@@ -0,0 +1,141 @@
...
+	return "https://signed.example/obj", nil
+}
+
+// nopWriteCloser is a simple write closer used for tests.
+type nopWriteCloser struct{}
+
+func (*nopWriteCloser) Write([]byte) (int, error) { return 0, nil }
+func (*nopWriteCloser) Close() error              { return nil }
+
+func TestCreateWithOptions_UsesProviderNewWriterWithOptions(t *testing.T) {
```
coolwednesday, line 107: **[Minor] Missing test: `CreateWithOptions` when `NewWriterWithOptions` returns nil.**

The implementation has this untested code path:
```go
writer := mw.NewWriterWithOptions(ctx, name, opts)
if writer == nil {
    msg = "failed to create writer with options"
    return nil, errWriterNil  // ← no test covers this branch
}
```
Please add a test where `fakeProvider.NewWriterWithOptions` returns `nil` and assert the error is the expected sentinel.

pkg/gofr/datasource/file/gcs/storage_adapter.go
coolwednesday: Please check if addressing this comment is necessary or not.

pkg/gofr/datasource/file/common_fs.go
Umang01-hash: The operation *is* a success — the file was created exactly as the OS/cloud layer accepted it. Returning an error or a SUCCESS_DEGRADED status would break any caller that treats a successful write as a success, and would force every provider (including local FS, FTP, SFTP) to implement metadata support just to avoid a misleading status code. The existing warning log already surfaces the degradation clearly. The godoc on CreateWithOptions documents the fallback behaviour. Keeping StatusSuccess here is intentional — the file operation succeeded.

pkg/gofr/datasource/file/gcs/storage_adapter.go
Umang01-hash: The EndPoint field is exclusively for local test emulators (fake-gcs-server); it is never set in production deployments. If the endpoint URL is malformed, the test fails immediately and obviously because the signed URL still points to storage.googleapis.com instead of localhost — there is no silent data loss. Adding a logger parameter to this private helper or changing its return type to (string, error) would propagate a test-only concern into the public SignedURL() signature and double-wrap an error that only ever occurs during local development. The current fallback — returning the original valid GCS URL — is the correct tradeoff here.

pkg/gofr/datasource/file/interface.go
Umang01-hash: Changing the parameter to FileSystemProvider would break the primary usage pattern. c.File in the container is typed as file.FileSystem (not FileSystemProvider), so file.AsCloud(c.File) would fail to compile with a FileSystemProvider parameter. Since all handlers access the filesystem through c.File, the FileSystem parameter type is required. The runtime type assertion inside AsCloud still succeeds when the underlying value implements CloudFileSystem, regardless of the static interface type at the call site.

pkg/gofr/datasource/file/common_fs_signedurl_test.go
Umang01-hash: Added **returnNilWriter** bool to **fakeProvider**, updated **NewWriterWithOptions** to respect it, and added **TestCreateWithOptions_NilWriterReturnsError**.

pkg/gofr/datasource/file/gcs/storage_adapter.go
Umang01-hash: Replaced **strings.Contains** with **strings.SplitN(..., 2)** checking both sides are non-empty. Extracted the 2 into **contentTypePartsCount** to satisfy the mnd (magic number) linter.

pkg/gofr/datasource/file/common_fs.go
Umang01-hash: Added `OpCreateWithOptions = "CREATE_WITH_OPTIONS"` to `observability.go` and switched **CreateWithOptions** to use it.

pkg/gofr/datasource/file/local_fs.go
Umang01-hash: Restored `file.New()` in `local_fs.go` with **// Deprecated: godoc**. It now delegates to `NewLocalFileSystem()` — no duplicate struct needed.

pkg/gofr/datasource/file/gcs/storage_adapter.go
Umang01-hash: Added if `s.bucket == nil → &failWriter{err: errGCSClientNotInitialized}` guard, consistent with **SignedURL()**. Added a dedicated test `TestStorageAdapter_NewWriterWithOptions_NilBucket`.

pkg/gofr/datasource/file/gcs/storage_adapter.go
Umang01-hash: Added **credParseErr** error field to **storageAdapter**. `Connect()` now stores the parse error but does not fail — GCS client and all regular operations work normally. The error is returned lazily from **SignedURL()** only if called. Updated the test to match the new behavior.

coolwednesday (changes requested):
The changes looks good to me. Can we add documentation as well.

coolwednesday (dismissed):

docs/advanced-guide/handling-file/page.md
```diff
@@ -216,6 +216,76 @@ func main() {
...
+
+Use `file.AsCloud()` to safely check whether the configured file store supports cloud-specific operations. This avoids a raw type assertion and returns a typed interface:
+
+```go
+import "gofr.dev/pkg/gofr/datasource/file"
+
+cfs, ok := file.AsCloud(c.File)
+if !ok {
+    // File store is local, FTP, SFTP, or a cloud provider without cloud feature support
+    return nil, fmt.Errorf("cloud operations not supported by this file store")
```
aryanmehrotra: should be Error not Errorf

aryanmehrotra: should docs not promote usage of gofr errors

docs/advanced-guide/handling-file/page.md
```diff
@@ -216,6 +216,76 @@ func main() {
...
+
+`GenerateSignedURL` creates a time-limited, pre-authenticated URL that allows anyone with the link to download the file — no GCS credentials required on the client side:
+
+```go
+url, err := cfs.GenerateSignedURL(c, "reports/q1.csv", 15*time.Minute, nil)
+if err != nil {
+    return nil, err
+}
+
+return map[string]string{"download_url": url}, nil
```
aryanmehrotra: why are we returning like this?

pkg/gofr/datasource/file/gcs/storage_adapter.go
```diff
@@ -28,17 +34,40 @@ var (
...
+	errExpiryMustBePositive    = errors.New("expiry duration must be positive")
+	errExpiryTooLong           = errors.New("expiry cannot exceed 7 days for GCS signed URLs")
+	errInvalidContentType      = errors.New("invalid Content-Type format")
 )
 
 const (
-	contentTypeDirectory = "application/x-directory"
+	contentTypeDirectory  = "application/x-directory"
+	contentTypePartsCount = 2
+	maxGCSSignedURLExpiry = 7 * 24 * time.Hour // 7 days
```
aryanmehrotra: should this not be configured by the user?

aryanmehrotra: Gofr can provide a default though but not max

pkg/gofr/datasource/file/gcs/storage_adapter.go
```diff
@@ -61,7 +90,7 @@ func (s *storageAdapter) Connect(ctx context.Context) error {
 	case s.cfg.EndPoint != "":
 		client, err = storage.NewClient(ctx, option.WithEndpoint(s.cfg.EndPoint), option.WithoutAuthentication())
 	case s.cfg.CredentialsJSON != "":
-		client, err = storage.NewClient(ctx, option.WithCredentialsJSON([]byte(s.cfg.CredentialsJSON)))
+		client, err = storage.NewClient(ctx, option.WithCredentialsJSON([]byte(s.cfg.CredentialsJSON))) //nolint:staticcheck // deprecated
```
aryanmehrotra: we should not be keeping nolints, isnt there another way?

pkg/gofr/datasource/file/gcs/storage_adapter.go
```diff
@@ -28,17 +34,40 @@ var (
...
+	// SignedURL(), avoiding repeated JSON+PEM parsing per request.
+	// When empty, bucket.SignedURL falls back to the client's ambient credentials
+	// (Workload Identity, Application Default Credentials, etc.).
+	saEmail      string
+	saPrivateKey []byte
+
+	// credParseErr holds any error encountered when parsing CredentialsJSON during Connect().
+	// Connect() does not fail on a parse error so that users who never call GenerateSignedURL
+	// are unaffected. The error is returned lazily from SignedURL() if it is non-nil.
+	credParseErr error
```
aryanmehrotra: what benefit do we have of saving the error, we can log and move forward

pkg/gofr/datasource/file/interface.go
```diff
@@ -132,7 +142,57 @@ type FileSystemProvider interface {
 	Connect()
 }
 
+// ============================================================
```
aryanmehrotra, line 145: lets fix this comment, its not possible, we are extending the interface

docs/advanced-guide/handling-file/page.md
Umang01-hash: Fixed. Changed to use `file.ErrSignedURLsNotSupported` (defined in `interface.go`) which is GoFr's own sentinel error for this case, and removed `fmt.Errorf` since there is no formatting needed.

docs/advanced-guide/handling-file/page.md
Umang01-hash: Simplified to `return url, nil`.

pkg/gofr/datasource/file/gcs/storage_adapter.go
Umang01-hash: Removed `maxGCSSignedURLExpiry` and the corresponding validation. GoFr no longer enforces a max — if an expiry exceeds what GCS accepts (7 days for V4 signing), the GCS SDK will return an error directly.

pkg/gofr/datasource/file/gcs/storage_adapter.go
Umang01-hash: Removed the nolint. Replaced `option.WithCredentialsJSON` (deprecated) with `google.JWTConfigFromJSON` + `option.WithTokenSource`, which is not deprecated and avoids going through the deprecated `google.CredentialsFromJSON` path. The client creation logic was also extracted into a `createStorageClient` helper to reduce cyclomatic complexity.

pkg/gofr/datasource/file/gcs/storage_adapter.go
Umang01-hash: Removed the `credParseErr` field. `Connect()` now logs a `Errirf` when credentials can't be parsed for signed URLs (instead of storing the error), and `saEmail`/`saPrivateKey` remain empty — `SignedURL()` will fall back to IAM-based ambient credentials. Added a `logger` field to `storageAdapter` and overrode `UseLogger` in `fileSystem` to inject it.

pkg/gofr/datasource/file/interface.go
Umang01-hash: Fixed the comment. Removed the misleading claim about being 'non-breaking' — we are indeed extending with new interfaces. The comment now just describes what `CloudFileSystem` is.

PR #3041, opened 2026-02-26
fix: append path separator in isRestrictedFile to prevent sibling directory escape

### LINKED ISSUE #3040: Bug: isRestrictedFile path-prefix check allows access to sibling directories
### Summary

`staticHandler` in `pkg/gofr/http/router.go` uses a bare `strings.HasPrefix` to verify that a resolved file path is inside the registered static directory. This check can incorrectly pass for sibling directories that share a common name prefix.

### Reproduction

Suppose the app serves static files from `./public`:

```go
app.AddStaticFiles("/static", "./public")
```

If a directory `./publicother/` also exists on the host (e.g. another registered endpoint or an application-created directory), a request that resolves to `/app/publicother/secret.txt` passes the check:

```
strings.HasPrefix("/app/publicother/secret.txt", "/app/public") → true
```

because `/app/public` is a string prefix of `/app/publicother`, not a directory boundary.

### Root Cause

`pkg/gofr/http/router.go:139` — no path separator is appended before the prefix comparison:

```go
return !strings.HasPrefix(absPath, staticConfig.directoryName) || fileName == DefaultSwaggerFileName
```

### Fix

```go
return !strings.HasPrefix(absPath, staticConfig.directoryName+string(os.PathSeparator)) || fileName == DefaultSwaggerFileName
```

### Impact

Potential information disclosure if sibling directories with matching name prefixes exist alongside a registered static file directory.

coolwednesday:
## Summary

Fixes #3040.

`isRestrictedFile` in `pkg/gofr/http/router.go` used a bare `strings.HasPrefix` to verify that a resolved file path sits inside the registered static directory:

```go
// before
return !strings.HasPrefix(absPath, staticConfig.directoryName) || ...
```

Because the check operates on raw strings rather than directory boundaries, a path like `/app/publicother/secret.txt` incorrectly passes the `/app/public` prefix test — `"publicother"` shares the string prefix `"public"`.

The fix appends `os.PathSeparator` to the directory name before the comparison so the boundary is always respected:

```go
// after
return !strings.HasPrefix(absPath, staticConfig.directoryName+string(os.PathSeparator)) || ...
```

`os` was already imported; no new dependencies added.

## Test plan

- [x] `go test gofr.dev/pkg/gofr/http/...` passes
- [x] Manually verify: registering `./public` as a static directory no longer serves files from a sibling `./publicother` directory

### DIFF
--- a/pkg/gofr/http/router_test.go
+++ b/pkg/gofr/http/router_test.go
@@ -371,36 +371,6 @@ func Test_isRestrictedFile(t *testing.T) {
 	}
 }
 
-func Test_StaticFileServing_SiblingDirectoryEscape(t *testing.T) {
-	// Create two sibling directories: "public" and "publicother"
-	parentDir := t.TempDir()
-	publicDir := filepath.Join(parentDir, "public")
-	siblingDir := filepath.Join(parentDir, "publicother")
-
-	assert.NoError(t, os.MkdirAll(publicDir, 0755))
-	assert.NoError(t, os.MkdirAll(siblingDir, 0755))
-
-	assert.NoError(t, os.WriteFile(filepath.Join(publicDir, "allowed.txt"), []byte("allowed"), 0600))
-	assert.NoError(t, os.WriteFile(filepath.Join(siblingDir, "secret.txt"), []byte("secret"), 0600))
-
-	logger := logging.NewMockLogger(logging.DEBUG)
-	router := NewRouter()
-	router.AddStaticFiles(logger, "/static", publicDir)
-
-	// File inside the registered directory should be served
-	req := httptest.NewRequest(http.MethodGet, "/static/allowed.txt", http.NoBody)
-	w := httptest.NewRecorder()
-	router.ServeHTTP(w, req)
-	assert.Equal(t, http.StatusOK, w.Code)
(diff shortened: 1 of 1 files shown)


pkg/gofr/http/router_test.go
```diff
@@ -317,6 +317,90 @@ func Test_StaticFileServing_Static(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			cfg := staticFileConfig{directoryName: tc.directoryName}
+			result := cfg.isRestrictedFile(tc.url, tc.absPath)
+			assert.Equal(t, tc.expected, result)
+		})
+	}
+}
+
+func Test_StaticFileServing_SiblingDirectoryEscape(t *testing.T) {
```
Umang01-hash: The integration test **Test_StaticFileServing_SiblingDirectoryEscape** does not exercise `isRestrictedFile` because the traversal is blocked by `path.Clean` in the router before the
  static handler is invoked. 

We can drop **Test_StaticFileServing_SiblingDirectoryEscape** entirely — **Test_isRestrictedFile** already covers the fix correctly.

PR #3042, opened 2026-02-26
feat(rpc): Implement GRPC rate limiter

0xEbrahim:
## Description
Implements **gRPC server-side rate limiting** using token bucket algorithm via **gRPC interceptors** (Unary + Stream), reusing GoFr’s existing `RateLimiterConfig` and `RateLimiterStore` design (gRPC portion) for issue #2551.

## Changes
- Add **UnaryRateLimitInterceptor** using token bucket rate limiting (via existing store)
- Add **StreamRateLimitInterceptor** (rate limits **stream establishment**)
- Support **per-IP** and **global** rate limiting modes
- Add **Retry-After** hint as gRPC response metadata (`retry-after`)
- Add **Prometheus metric** for rate limit violations: `app_grpc_rate_limit_exceeded_total`
- IP extraction supports:
  - `x-forwarded-for` and `x-real-ip` (when `TrustedProxies=true`)
  - gRPC peer address fallback via `peer.FromContext(ctx)`
- Reuse existing store cleanup mechanism (started with `context.Background()`)

## Features
- **Token Bucket Algorithm**: Smooth rate limiting with configurable burst (via `RateLimiterStore`)
- **Per-IP Limiting**: Each client IP gets independent rate limit (configurable)
- **Global Limiting**: Shared limiter when `PerIP=false`
- **Retry Hint**: Returns `retry-after` as gRPC metadata (best-effort)
- **Prometheus Metrics**: Track rate limit violations via `app_grpc_rate_limit_exceeded_total`
- **IP Extraction**: Supports `x-forwarded-for`, `x-real-ip`, and gRPC peer address

## Notes
- gRPC equivalent of HTTP 429 is returned as **`codes.ResourceExhausted`** with message `"rate limit exceeded"`.
- Stream interceptor currently limits **stream creation**, not per-message throughput.
- **Security**: Only set `TrustedProxies=true` if behind a trusted reverse proxy/LB; otherwise forwarded headers can be spoofed.

## Testing
All tests pass:

```bash
go test ./... -v -run Test.*RateLimit.*
## Testing
All tests pass:

```bash
go test ./pkg/gofr/grpc -run 'Test.*RateLimit' -v
# ALL PASSED
```

**Checklist:**

### DIFF
--- a/pkg/gofr/grpc/log.go
+++ b/pkg/gofr/grpc/log.go
@@ -157,7 +157,7 @@ func ObservabilityInterceptor(logger Logger, metrics Metrics) grpc.UnaryServerIn
 		ctx, span := tracer.Start(ctx, info.FullMethod)
 
 		resp, err := handler(ctx, req)
-		if err != nil {
+		if err != nil && isServerError(err) {
 			logger.Errorf("error while handling gRPC request to method %q: %q", info.FullMethod, err)
 		}
 
@@ -294,6 +294,28 @@ func recordGRPCMetrics(ctx context.Context, metrics Metrics, name string, durati
 	}
 
 	metrics.RecordHistogram(ctx, name, durationMs, labels...)
+}
+
+// isServerError returns true if the gRPC error represents a server-side error.
+// Client errors like ResourceExhausted, InvalidArgument, NotFound, etc. are not
+// considered server errors and should not be logged at ERROR level.
+func isServerError(err error) bool {
+	s, ok := status.FromError(err)
+	if !ok {
+		return true
+	}
+
+	switch s.Code() {
+	case codes.InvalidArgument, codes.NotFound, codes.AlreadyExists,
+		codes.PermissionDenied, codes.Unauthenticated, codes.ResourceExhausted,
+		codes.FailedPrecondition, codes.OutOfRange, codes.Canceled:
+		return false
+	case codes.OK, codes.Unknown, codes.DeadlineExceeded, codes.Aborted,
+		codes.Unimplemented, codes.Internal, codes.Unavailable, codes.DataLoss:
+		return true
+	default:
+		return true
+	}
 }
 
 // Helper function to safely extract a value from metadata.
--- a/pkg/gofr/grpc/log_test.go
+++ b/pkg/gofr/grpc/log_test.go
@@ -687,6 +687,63 @@ func TestRecordGRPCMetrics_WithStreamType(t *testing.T) {
 	recordGRPCMetrics(ctx, mockMetrics, "test_metric", 100*time.Millisecond, "/test.Service/Method", "SERVER_STREAM")
 }
 
+func TestIsServerError(t *testing.T) {
+	tests := []struct {
+		name     string
+		err      error
+		expected bool
+	}{
+		{"Internal is server error", status.Error(codes.Internal, "internal"), true},
+		{"Unknown is server error", status.Error(codes.Unknown, "unknown"), true},
+		{"Unavailable is server error", status.Error(codes.Unavailable, "unavailable"), true},
+		{"DataLoss is server error", status.Error(codes.DataLoss, "data loss"), true},
+		{"DeadlineExceeded is server error", status.Error(codes.DeadlineExceeded, "timeout"), true},
+		{"Aborted is server error", status.Error(codes.Aborted, "aborted"), true},
+		{"Unimplemented is server error", status.Error(codes.Unimplemented, "unimplemented"), true},
+		{"ResourceExhausted is not server error", status.Error(codes.ResourceExhausted, "rate limit"), false},
+		{"InvalidArgument is not server error", status.Error(codes.InvalidArgument, "bad arg"), false},
+		{"NotFound is not server error", status.Error(codes.NotFound, "not found"), false},
+		{"PermissionDenied is not server error", status.Error(codes.PermissionDenied, "denied"), false},
+		{"Unauthenticated is not server error", status.Error(codes.Unauthenticated, "unauth"), false},
+		{"AlreadyExists is not server error", status.Error(codes.AlreadyExists, "exists"), false},
+		{"FailedPrecondition is not server error", status.Error(codes.FailedPrecondition, "precondition"), false},
+		{"OutOfRange is not server error", status.Error(codes.OutOfRange, "range"), false},
+		{"Canceled is not server error", status.Error(codes.Canceled, "canceled"), false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.expected, isServerError(tt.err))
+		})
+	}
+}
+
+func TestObservabilityInterceptor_WithClientError(t *testing.T) {
+	mockLogger, mockMetrics, ctrl := createMocks(t)
+	defer ctrl.Finish()
+
+	interceptor := ObservabilityInterceptor(mockLogger, mockMetrics)
+
+	ctx := t.Context()
+	req := "test request"
+	info := &grpc.UnaryServerInfo{
+		FullMethod: "/test.Service/Method",
+	}
+
+	handler := func(_ context.Context, _ any) (any, error) {
+		return nil, status.Error(codes.ResourceExhausted, "rate limit exceeded")
+	}
+
+	// Errorf should NOT be called for client errors like ResourceExhausted
+	mockLogger.EXPECT().Info(gomock.Any()).Times(1)
+	mockMetrics.EXPECT().RecordHistogram(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(1)
+
+	resp, err := interceptor(ctx, req, info, handler)
+
+	require.Error(t, err)
+	assert.Nil(t, resp)
+}
+
 func TestRecordGRPCMetrics_WithNilMetrics(t *testing.T) {
 	ctx := t.Context()
 


0xEbrahim:
@Umang01-hash Kindly, review this PR if you can

pkg/gofr/grpc/rate_limiter.go
```diff
@@ -0,0 +1,200 @@
...
+
+	if config.Store == nil {
+		config.Store = httpmw.NewMemoryRateLimiterStore(config)
+	}
+
+	config.Store.StartCleanup(context.Background())
+
+		return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
+			key := rateLimitKeyGlobal
+		if config.PerIP {
```
coolwednesday: **Skip gRPC health check methods**

The HTTP rate limiter skips well-known health endpoints via `isWellKnown()`. The gRPC interceptor has no equivalent. Kubernetes liveness/readiness probes hitting `/grpc.health.v1.Health/Check` will consume rate limit tokens, which can cause probe failures, pod restarts, and cascading outages in production.

Add a health check bypass at the top of both interceptor closures:

```go
if strings.HasPrefix(info.FullMethod, "/grpc.health.v1.Health/") {
    return handler(ctx, req)
}
```

Same change needed in `StreamRateLimitInterceptor`.

pkg/gofr/grpc/rate_limiter.go
```diff
@@ -0,0 +1,200 @@
...
+		panic(fmt.Sprintf("invalid rate limiter config: %v", err))
+	}
+
+	if config.Store == nil {
+		config.Store = httpmw.NewMemoryRateLimiterStore(config)
+	}
+
+	config.Store.StartCleanup(context.Background())
+
+		return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
```
coolwednesday: **Cleanup goroutine has no shutdown path**

`StartCleanup(context.Background())` starts a background goroutine that runs indefinitely. There is no `StopCleanup()` call wired into the gRPC server's `Shutdown()` method. During rolling restarts in production, this results in leaked goroutines.

Either accept a cancellable context from the caller, or wire `StopCleanup()` into the server shutdown path in `grpc.go`. Same issue at line 167 for the stream interceptor.

pkg/gofr/grpc/rate_limiter.go
```diff
@@ -0,0 +1,200 @@
...
+			key := rateLimitKeyGlobal
+		if config.PerIP {
+			key = getIP(ctx, config.TrustedProxies)
```
coolwednesday: **Include method in rate limit key for per-IP mode**

Currently the rate limit key is either `"global"` or the client IP. This means all gRPC methods share the same token bucket. A client calling `/Payment/Process` and `/Hello/SayHello` competes for the same tokens.

The HTTP server middleware has the same limitation, but since `info.FullMethod` is already available in the interceptor, consider including it in the key:

```go
key := rateLimitKeyGlobal
if config.PerIP {
    key = getIP(ctx, config.TrustedProxies)
    if key == "" {
        key = rateLimitKeyUnknown
    }
    key = key + ":" + info.FullMethod
}
```

This gives per-IP-per-method limiting with no additional configuration surface. If this is intentionally deferred, add a comment explaining that all methods share a single bucket.

pkg/gofr/grpc/rate_limiter_test.go
```diff
@@ -0,0 +1,846 @@
+package grpc
```
coolwednesday, line 1: **Add an integration test**

The unit tests are comprehensive, but there is no end-to-end test that starts a real gRPC server with rate limiting enabled, makes actual RPC calls, and verifies that the (N+1)th call returns `codes.ResourceExhausted`.

The existing examples in `examples/grpc/grpc-unary-server/` are the natural place for this. A minimal test would be:

```go
func TestIntegration_UnaryServer_RateLimited(t *testing.T) {
    // Start server with Burst=2
    // Make 3 RPC calls
    // Assert 3rd returns codes.ResourceExhausted
}
```

This validates the full interceptor chain (recovery, observability, rate limit, handler) works together, not just the interceptor in isolation.

docs/advanced-guide/grpc/page.md
```diff
@@ -222,6 +222,80 @@ func streamAuthInterceptor(srv any, ss grpc.ServerStream, info *grpc.StreamServe
...
+	app.AddGRPCUnaryInterceptors(gofrGrpc.UnaryRateLimitInterceptor(rateLimiterConfig, app.Metrics()))
+	app.AddGRPCServerStreamInterceptors(gofrGrpc.StreamRateLimitInterceptor(rateLimiterConfig, app.Metrics()))
+
+	// Register your gRPC service
+	packageName.Register<SERVICE_NAME>ServerWithGofr(app, &packageName.New<SERVICE_NAME>GoFrServer())
+
+	app.Run()
+}
+```
+
```
coolwednesday, line 282: **Document separate rate limits for unary and stream interceptors**

The docs currently show the same config for both interceptors. Add a note or example showing that users can pass different configs to each interceptor to enforce different limits — for example, allowing 100 unary calls/sec but only 5 new stream connections/sec:

```go
unaryCfg := middleware.RateLimiterConfig{
    RequestsPerSecond: 100,
    Burst:             50,
}

streamCfg := middleware.RateLimiterConfig{
    RequestsPerSecond: 5,
    Burst:             3,
}

app.AddGRPCUnaryInterceptors(gofrGrpc.UnaryRateLimitInterceptor(unaryCfg, app.Metrics()))
app.AddGRPCServerStreamInterceptors(gofrGrpc.StreamRateLimitInterceptor(streamCfg, app.Metrics()))
```

This makes it clear that unary and stream limits are independent and can be tuned separately based on the resource cost of each call type.

0xEbrahim:
@coolwednesday please take a look

coolwednesday:
## Changes Pushed — Rate Limiter Hardening

### 1. Rich Error Model
- Replaced simple `status.Error(codes.ResourceExhausted, ...)` with `errdetails.RetryInfo` via `rateLimitExhaustedError()`.
- Standard gRPC clients can now extract `RetryInfo.RetryDelay` natively for automatic retry backoff.
- `retry-after` metadata header is still sent alongside for non-proto clients.

### 2. Logging
- Added `Logger` parameter to both `UnaryRateLimitInterceptor` and `StreamRateLimitInterceptor`.
- Calls `l.Errorf(...)` on every rate-limit violation, logging the key and method for observability without requiring a metrics dashboard.

### 3. IP Optimization
- Merged `getForwardedIP()`, `getRealIP()`, and `first()` into a single `extractHeaderIP()` — reduces redundant `metadata.FromIncomingContext()` calls from 2 to 1 when proxies are trusted.
- `normalizeIP()` now skips `net.SplitHostPort()` for pure IPv4 addresses (no colon), avoiding unnecessary parsing overhead on every RPC.
- Decomposed `getIP()` into `getIPFromProxy()` / `getIPFromPeer()` to eliminate nested-if complexity (fixed `nestif` lint violation).

### 4. Tests
- Added `mockLogger` with `errorCount()` — asserts `Errorf` is called exactly once on violation.
- Added `errdetails.RetryInfo` assertions in both `TestUnaryRateLimitInterceptor_RetryAfterHeader` and `TestStreamRateLimitInterceptor_RetryAfterHeader` — verifies `RetryDelay` matches the expected duration.
- Fixed all 15+ call sites for the new `Logger` parameter.
- Replaced removed helper tests (`Test_first`, `Test_getForwardedIP`, `Test_getRealIP`) with `Test_extractHeaderIP`.

### 5. Linting
- 0 issues from `golangci-lint` — fixed `lll` (line length), `nestif`, and `revive` (unused receiver) violations.
- `goimports` clean.

**All 30 rate-limiter tests pass.**

0xEbrahim:
@coolwednesday  I fixed wrong function call on main_test.go, app.logger() was missing, you can take a look now

0xEbrahim:
@Umang01-hash @coolwednesday Hi, are there any changes should i apply on the current PR or it is ready already?

docs/advanced-guide/grpc/page.md
```diff
@@ -222,6 +222,126 @@ func streamAuthInterceptor(srv any, ss grpc.ServerStream, info *grpc.StreamServe
...
+	// IMPORTANT: create ONE shared store if you want a single budget
+	// for both unary and stream RPCs. If Store is left nil, each
+	// interceptor will create its own in-memory store and limits
+	// will be enforced independently.
+	store := middleware.NewMemoryRateLimiterStore(cfg)
+	cfg.Store = store
+
+	// Add rate limiter interceptors for gRPC
+	// Pass app.Metrics() to emit Prometheus counters, or nil to disable metrics.
+	app.AddGRPCUnaryInterceptors(gofrGrpc.UnaryRateLimitInterceptor(ctx, cfg, app.Metrics()))
```
Umang01-hash: This signature is wrong for **gofrGrpc.UnaryRateLimitInterceptor** and **gofrGrpc.StreamRateLimitInterceptor**
actual signature is `(ctx, cfg, Logger, CounterMetrics)` — app.Logger() is missing as the 3rd argument.

pkg/gofr/grpc/rate_limiter.go
```diff
@@ -0,0 +1,235 @@
...
+	httpmw "gofr.dev/pkg/gofr/http/middleware"
+)
+
+const (
+	rateLimitKeyGlobal      = "global"
+	rateLimitKeyUnknown     = "unknown"
+	grpcHealthServicePrefix = "/grpc.health.v1.Health/"
+)
+
+type CounterMetrics interface {
```
Umang01-hash: **log.go** already defines:
```golang
  type Metrics interface {
      RecordHistogram(ctx context.Context, name string, value float64, labels ...string)
  }
```

And, now the grpc package has two exported metric interfaces. Why?

pkg/gofr/grpc/rate_limiter.go
```diff
@@ -0,0 +1,235 @@
...
+	if err := cfg.Validate(); err != nil {
+		panic(fmt.Sprintf("invalid rate limiter config: %v", err))
+	}
+
+	if cfg.Store == nil {
+		cfg.Store = httpmw.NewMemoryRateLimiterStore(cfg)
+	}
+
+	cfg.Store.StartCleanup(ctx)
+
```
Umang01-hash, line 217: Code duplication between **Unary** and **Stream interceptors**.

```golang
if err := cfg.Validate(); err != nil { panic(...) }
  if cfg.Store == nil { cfg.Store = httpmw.NewMemoryRateLimiterStore(cfg) }
  cfg.Store.StartCleanup(ctx)
```

This should be extracted as helper for the setup,
  and a shared rate-check block.

docs/advanced-guide/grpc/page.md
```diff
@@ -222,6 +222,126 @@ func streamAuthInterceptor(srv any, ss grpc.ServerStream, info *grpc.StreamServe
...
+> **Security Warning**: Only set `TrustedProxies: true` if your application is behind a trusted reverse proxy (nginx, ALB, etc.).
+> Without a trusted proxy, clients can spoof metadata headers to bypass rate limits.
+
+### Behavior on Rate Limit Exceeded
+
+When a client exceeds the rate limit:
+
+1. The interceptor returns a gRPC error with status code `RESOURCE_EXHAUSTED`
+2. A `retry-after` response metadata header is set, indicating how many seconds the client should wait before retrying
+3. The `app_grpc_rate_limit_exceeded_total` Prometheus counter is incremented with `method` and `type` (`unary` or `stream`) labels
```
Umang01-hash, line 308: `app_grpc_rate_limit_exceeded_total`, where are we registering this metrics?

pkg/gofr/grpc/rate_limiter.go
```diff
@@ -0,0 +1,235 @@
...
+func rateLimitExhaustedError(ctx context.Context, retryAfter time.Duration) error {
+	st := status.New(codes.ResourceExhausted, "rate limit exceeded")
+	retryDetail := &errdetails.RetryInfo{
+		RetryDelay: durationpb.New(retryAfter),
+	}
+
+	st, _ = st.WithDetails(retryDetail)
+
+	_ = grpc.SendHeader(ctx, metadata.Pairs("retry-after", retryAfterSeconds(retryAfter.Seconds())))
```
Umang01-hash, line 151: The **retry-after header** works in the `unary interceptor` but is silently dropped in the `stream interceptor ` because the wrong API is used.

 No **stream test** catches this because tests also use plain `context.Background()`, masking the bug.

Umang01-hash:
The interceptor chain order is: **recovery → ObservabilityInterceptor → [user's UnaryRateLimitInterceptor]**

Because user interceptors are appended after the built-in ones, when a request is rate-limited, ObservabilityInterceptor sees the **ResourceExhausted** error and logs it as:

  `{"level":"ERROR","message":"error while handling gRPC request to method \"/Hello/SayHello\": \"rpc error: code = ResourceExhausted desc = rate limit exceeded\""}`

  This happens for every single rate-limited request. In production with aggressive rate limiting, this floods error logs with what look like internal server errors but are in fact intentional rejections.
  
   **ObservabilityInterceptor** should not call **logger.Errorf** for `codes.ResourceExhausted` — that's a client-side issue, not a server error.

coolwednesday:
Addressed in 275ccd00. 

`ObservabilityInterceptor` now calls `isServerError(err)` before logging at ERROR level. Client-side gRPC codes (`ResourceExhausted`, `InvalidArgument`, `NotFound`, `PermissionDenied`, `Unauthenticated`, `AlreadyExists`, `FailedPrecondition`, `OutOfRange`, `Canceled`) are no longer logged as errors — only true server errors (`Internal`, `Unknown`, `Unavailable`, `DataLoss`, `DeadlineExceeded`, `Aborted`, `Unimplemented`) are.

This prevents rate-limited requests from flooding error logs in production.

PR #3044, opened 2026-02-28
added built in SSE support in GoFr

### LINKED ISSUE #3038: Add Built-in SSE Support in GoFr
## Description

GoFr currently does not provide first-class support for Server-Sent Events (SSE). While SSE can be implemented manually using the underlying `net/http`, developers must handle headers, flushing, and connection lifecycle themselves.

It would be helpful to provide a built-in SSE helper or streaming API that:

* Automatically sets required headers (`text/event-stream`)
* Handles flushing
* Detects client disconnects
* Optionally supports named events and JSON payloads

This would reduce boilerplate and improve developer experience for real-time use cases such as notifications, dashboards, and live updates.

IronwallxR5:
## Pull Request Template

**Description:**

Adds built-in Server-Sent Events (SSE) support to the GoFr framework.

Fixes #3038

This introduces a seamless way to handle SSE by returning a new custom response type (`response.SSE`) from standard HTTP handlers. It utilizes an idiomatic closure-based streaming architecture that aligns with GoFr’s `(any, error)` handler pattern.

*Architecture*:

* No specialized routing methods: SSE is supported via standard registration (e.g., `app.GET()`), keeping routing consistent with HTTP verbs and avoiding framework bloat.

* Sequential Network Writes: `SSEStream.Send()` writes directly to the underlying http.ResponseWriter and flushes using http.ResponseController. This minimizes memory overhead and relies on standard library blocking for natural backpressure (removing the need for complex channel brokers).

* Separation of Concerns: Introduced gofr.SSEResponse as a helper that returns the framework's native response.SSE. This keeps user-facing APIs in the top-level package while keeping internal HTTP definitions in pkg/gofr/http/response.

*Middleware and Proxy compatibility*:

* Added `Flush()` and `Unwrap()` to `StatusResponseWriter` to allow `http.ResponseController` to reach the underlying flusher through the middleware chain.

* Sets `X-Accel-Buffering: no` to ensure real-time delivery through NGINX/Ingress proxies.

* Sets required headers: `Content-Type: text/event-stream`, `Cache-Control: no-cache`, and `Connection: keep-alive`.

None

**Additional Information:**

* Clean Lifecycle: Leverages standard Go context (`<-c.Context.Done()`) for idiomatic client disconnect detection within the handler closure.

* Comprehensive Testing: 26 tests covering event serialization, multi-line data formatting, direct flushing through middleware, and full HTTP integration.

* Documentation: Included an example application showing both infinite time-streaming and finite countdown streams.

**Checklist:**


coolwednesday:
@IronwallxR5, Let's finalise on the structure and implement accordingly. Feel free to reopen this pull request after that.

IronwallxR5:
Hey @Umang01-hash, I have made the latest changes according to the sequential callback design. Everything appears to work as planned, but I'm having trouble getting one specific CI check to pass. 

Can you tell me if this is a blocker or if you have any suggestions for fixing that failing case? Thanks!

pkg/gofr/http/response/sse.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"net/http"
+)
+
+// SSE represents a Server-Sent Events response.
+// Return this from a handler to stream events to the client.
+type SSE struct {
+	// Stream uses the provided ResponseWriter to send Server-Sent Events.
+	Stream func(w http.ResponseWriter) error
```
Umang01-hash: The **response.SSE** struct embeds a function: `Stream func(w http.ResponseWriter) error.` 

Every other response type **(File, XML, Raw, Redirect, Template)** is a pure data struct with no embedded behavior. Embedding a function here is an anomaly — it makes r`esponse.SSE `harder to inspect, log, or serialize. 

The function is also called with : 

`_ = sse.Stream(r.w)`


  with no nil-check, so a **zero-value response.SSE{} would panic.**

pkg/gofr/sse_test.go
```diff
@@ -0,0 +1,379 @@
...
+		Name: "notification",
+		ID:   "1",
+		Data: Notification{Title: "Hello", Message: "World"},
+	})
+
+	require.NoError(t, err)
+	assert.Equal(t, "id: 1\nevent: notification\ndata: {\"title\":\"Hello\",\"message\":\"World\"}\n\n", w.Body.String())
+}
+
+func TestSSEStream_ConcurrentSend(t *testing.T) {
```
Umang01-hash, line 239: This test documents **a data race as expected behavior** and then **tests nothing meaningful** `(assert.Positive(t, dataCount)). `

Running with **-race** would catch this. 
**The comment is a red flag:** _if concurrent use is unsupported, the type should be documented as such and this test should be removed_. If concurrent use is intended, a mutex is needed.

pkg/gofr/sse.go
```diff
@@ -0,0 +1,155 @@
...
+				w:  w,
+				rc: http.NewResponseController(w),
+			}
+			return callback(stream)
+		},
+	}
+}
+
+// Send writes a formatted SSE event to the stream and flushes.
+func (s *SSEStream) Send(event any) error {
```
Umang01-hash: **This is a type-safety problem.** Send accepts `any`, does a type assertion, and silently falls back to `SendData` if you pass a non-SSEEvent. 

With **SendData** and **SendEvent** already existing as clear named alternatives, `Send(any)` is confusing API surface. A user calling `stream.Send("hello")` would get the fallback behavior without any warning.

Please change it to:
`  func (s *SSEStream) Send(event SSEEvent) error`
  This is a public API that we'd have to support long-term, so getting the signature right now matters.

examples/using-sse/main.go
```diff
@@ -0,0 +1,61 @@
...
+			for i := 10; i >= 0; i-- {
+				select {
+				case <-c.Context.Done():
+					return nil
+				default:
+					if err := stream.SendEvent("countdown", map[string]int{"remaining": i}); err != nil {
+						return err
+					}
+
+					time.Sleep(500 * time.Millisecond)
```
Umang01-hash: - Fix the countdown example to use a ticker instead of `time.Sleep`

pkg/gofr/http/responder.go
```diff
@@ -276,3 +280,21 @@ func isNil(i any) bool {
...
+	r.w.Header().Set("Connection", "keep-alive")
+	r.w.Header().Set("X-Accel-Buffering", "no")
+	r.w.WriteHeader(http.StatusOK)
+
+	// Initial flush to establish connection
+	rc := http.NewResponseController(r.w)
+	_ = rc.Flush()
+
+	// Intercept and stream events using the handler's callback
+	_ = sse.Stream(r.w)
```
Umang01-hash: If the callback returns an error **(e.g., broken pipe, application-level failure)**, it's swallowed with no log entry. Please use the logger from the responder or add at minimum a debug log for the error.

Umang01-hash (commented):
Hey @IronwallxR5 Thanks for making the requested changes as per the issue discussion. After re-revieiwing i found a few gaps that needs to be filled before going ahead with the PR:

**1.** **Missing tests:**
  - `response.SSE` with nil Stream function (should test panic safety or nil guard)
  - Behavior when `sse.Stream` returns an error (currently discarded)
  - `Send(someNonSSEEventValue)` — the fallback behavior of the any parameter  
  
  
**2.** **Send(event any)** must be Send(event SSEEvent) — blocking API design issue (see inline comment on sse.go)                                                           
            
**3.** **Silence on callback** error in handleSSEResponse — see inline on responder.go

pkg/gofr/http/responder.go
coolwednesday: Beyond the error swallowing that Umang flagged — `sse.Stream(r.w)` blocks for the entire lifetime of the SSE connection (could be minutes/hours). During this time:

1. The logging middleware (`StatusResponseWriter`) is waiting for `ServeHTTP` to return to log the request. The log entry will only fire after the SSE stream ends, with a potentially enormous duration. Is this the desired behavior, or should SSE connections be logged differently (e.g., log on connect + log on disconnect)?

2. The `app_http_response` histogram will record the full connection lifetime as the response duration. A 30-minute SSE stream will skew all your percentiles. Consider either excluding SSE from the histogram or adding a label to distinguish SSE from regular responses.

These aren't blockers but worth thinking about before merge.

coolwednesday (commented):
Thanks for the PR @IronwallxR5. The overall direction is solid — returning `response.SSE` from a standard `app.GET` handler is the right approach. Umang's existing feedback on `Send(any)` → `Send(SSEEvent)`, nil-guard on `Stream`, error logging, concurrent test, and countdown ticker are all valid and should be addressed.

I have a few additional concerns, mostly around the request timeout interaction and the streaming lifecycle:

---

**Critical: Request timeout will break long-lived SSE streams**

Looking at `handler.go:63-68`, when `requestTimeout != 0`, the handler's context gets wrapped with `context.WithTimeout`. The handler goroutine returns quickly (returning the `SSEResponse` struct), and `close(done)` fires on line 85. The `select` on line 88 picks up `<-done` and proceeds to call `c.responder.Respond(result, err)` on line 112.

At this point, `handleSSEResponse` calls `sse.Stream(r.w)` which runs the user's callback — a potentially infinite loop. The user's callback uses `c.Context.Done()` to detect disconnect. But `c.Context` has the timeout attached. So if the request timeout is, say, 5 seconds, the SSE stream dies after 5 seconds regardless of client state.

WebSocket already has a carve-out for this (line 60-62). SSE needs the same treatment. The problem is that we don't know it's SSE until *after* the handler returns (unlike WebSocket which is detectable from the request headers). Options:
1. Detect SSE in the response type and re-derive the context from `r.Context()` before calling `Respond()`
2. Pass the raw `r.Context()` into the `SSE.Stream` function instead of relying on `c.Context`
3. Add a way for handlers to opt out of the timeout (e.g., a header hint or a context value)

This is a design decision that needs alignment before merging.

pkg/gofr/sse.go
```diff
@@ -0,0 +1,155 @@
...
+type SSEEvent struct {
+	Name  string // Event type (event: field)
+	Data  any    // Event data (data: field) - strings pass through, others are JSON-encoded
+	ID    string // Event ID (id: field)
+	Retry int    // Reconnection time in milliseconds (retry: field)
+}
+
+// SSEStream writes Server-Sent Events directly to the HTTP response.
+// It wraps an http.ResponseWriter and flushes after each event.
+type SSEStream struct {
```
coolwednesday, line 27: `SSEStream` has no mutex, so concurrent `Send()` calls from multiple goroutines will interleave writes to `http.ResponseWriter` and produce corrupted SSE frames on the wire.

Two options:
1. Add a `sync.Mutex` and lock around write+flush in each method
2. Document that `SSEStream` is not safe for concurrent use (and remove the concurrent test)

Option 2 is simpler and matches the typical SSE pattern where a single goroutine produces events sequentially. The existing concurrent test acknowledges data races in its comment — that's a sign this path shouldn't be supported without explicit synchronization.

pkg/gofr/sse.go
```diff
@@ -0,0 +1,155 @@
...
+//	            }
+//	            time.Sleep(time.Second)
+//	        }
+//	        return nil
+//	    }), nil
+//	})
+func SSEResponse(callback SSEFunc) response.SSE {
+	return response.SSE{
+		Stream: func(w http.ResponseWriter) error {
+			stream := &SSEStream{
```
coolwednesday: The `SSEResponse` helper creates `http.NewResponseController(w)` inside the `Stream` closure. But `handleSSEResponse` in `responder.go` also creates one for the initial flush (line 293). That's two `ResponseController` instances for the same writer. While this works (they're stateless wrappers), it's cleaner to create the `ResponseController` once in `handleSSEResponse` and pass it to the `Stream` function, or have `Stream` accept a pre-built `SSEStream` instead of a raw `http.ResponseWriter`.

Actually, looking at this more carefully — `response.SSE.Stream` takes `http.ResponseWriter`, but the user never sees this. The user's callback receives `*SSEStream`. The `ResponseWriter -> SSEStream` conversion happens here in line 53-56. This is fine architecturally, but the `response.SSE` type leaks `http.ResponseWriter` into the response package. Consider whether `Stream` should take `*SSEStream` directly (constructed by the responder), keeping `http.ResponseWriter` out of the response type.

pkg/gofr/sse.go
```diff
@@ -0,0 +1,155 @@
...
+		return "", nil
+	default:
+		b, err := json.Marshal(v)
+		if err != nil {
+			return "", fmt.Errorf("failed to marshal SSE data: %w", err)
+		}
+
+		return string(b), nil
+	}
+}
```
coolwednesday, line 204: **Missing heartbeat.** The issue discussion and original proposal both called for periodic heartbeat comments (`: heartbeat\n\n`) to keep the connection alive through proxies/load balancers that kill idle connections. This PR doesn't implement it.

Without heartbeats, if the user's callback pauses between events (e.g., waiting for a database change), proxies like nginx/AWS ALB with default idle timeouts (60s) will terminate the connection.

A simple approach: start a goroutine in `handleSSEResponse` that writes `SendComment("heartbeat")` every 15-30s, stopped when the stream ends. This does require synchronization with `Send()` writes (reinforcing the mutex point above).

pkg/gofr/sse_test.go
```diff
@@ -0,0 +1,379 @@
...
+}
+
+func TestSSEResponse_Integration(t *testing.T) {
+	configs := testutil.NewServerConfigs(t)
+
+	app := New()
+
+	app.GET("/events", func(_ *Context) (any, error) {
+		return SSEResponse(func(stream *SSEStream) error {
+			for i := 0; i < 3; i++ {
```
coolwednesday: This integration test starts a full `app.Run()` but never shuts it down. The goroutine and HTTP server leak for the rest of the test process. Consider using `t.Cleanup` with a shutdown context, or at least `defer app.Shutdown(...)`.

Also, the retry loop (lines 283-295) should use the alive endpoint like the disconnect test does — more reliable than retrying the SSE endpoint itself.

examples/using-sse/main.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"gofr.dev/pkg/gofr"
+)
+
+func main() {
+	app := gofr.New()
+
+	// Stream the current time every second.
+	app.GET("/events", func(c *gofr.Context) (any, error) {
+		return gofr.SSEResponse(func(stream *gofr.SSEStream) error {
```
coolwednesday: This example doesn't demonstrate graceful shutdown. When the server receives SIGTERM, active SSE connections should be drained. Worth adding a comment or showing how the callback respects `c.Context.Done()` for shutdown (which it already does — but calling it out explicitly in the example helps users understand the pattern).

coolwednesday:
@IronwallxR5, Can you please resolve merge conflicts ?

coolwednesday (commented):
## Re-review — Checking Previous Feedback 

@IronwallxR5, thanks for the updates. I've gone through the latest code against all previous review comments from both Umang and myself. Here's where things stand:

---

### Previous Review Comments — Status

#### Umang's Comments

| # | Feedback | Status |
|---|---|---|
| 1 | `response.SSE` embeds a function (anomaly) + nil Stream panics | ⚠️ **Partially** — nil guard added, but `Callback any` is a type-safety regression (see below) |
| 2 | Concurrent test documents data race | ✅ **Addressed** — `sync.Mutex` added, test asserts exact count |
| 3 | `Send(any)` → `Send(SSEEvent)` | ✅ **Addressed** |
| 4 | Countdown example: use ticker | ✅ **Addressed** |
| 5 | Callback error swallowed | ✅ **Addressed** — error now logged via `Debugf` |
| 6 | Missing tests (nil Stream, error, fallback) | ⚠️ **Partially** — `TestSSEResponse_NilCallback` tests `SSEStream` directly, not `response.SSE{Callback: nil}` through the responder path |

#### coolwednesday's Comments

| # | Feedback | Status |
|---|---|---|
| 1 | **Critical: Request timeout kills SSE** | ✅ **Addressed** — `handler.go:112-114` replaces context after detecting `response.SSE` |
| 2 | Logging/histogram skewed by SSE lifetime | ⚠️ **Partially** — TODO comment only, no fix |
| 3 | `SSEStream` needs mutex | ✅ **Addressed** |
| 4 | Two `ResponseController` instances | ✅ **Addressed** — single RC created in responder, passed to callback |
| 5 | Missing heartbeat | ✅ **Addressed** — 15s heartbeat goroutine with proper cleanup |
| 6 | Integration test leaks `app.Run()` | ⚠️ **Partially** — retry uses `/alive` now, but no `t.Cleanup` shutdown |
| 7 | Example: show graceful shutdown | ✅ **Addressed** |

**Score: 9/13 fully addressed, 4 partially.**

---

### New Issues Found

#### 🔴 1. `Callback any` is a type-safety regression

**File:** `pkg/gofr/http/response/sse.go:8`

```go
type SSE struct {
    // Typed as any to avoid circular imports
    Callback any
}
```

The original `Stream func(http.ResponseWriter) error` was at least typed. Changing to `any` means:
- A user could set `Callback` to anything (string, int, struct) — no compile-time error
- `responder.go:310` does a runtime assertion that silently fails (returns without SSE output, just a debug log)
- This is a **public API** that we'll have to support long-term

The "avoid circular imports" justification doesn't hold — the callback type `func(http.ResponseWriter, *http.ResponseController) error` uses only stdlib types and can live in the `response` package without circular imports:

```go
// In response/sse.go
type SSECallback func(w http.ResponseWriter, rc *http.ResponseController) error

type SSE struct {
    Callback SSECallback
}
```

This restores type safety and still avoids circular imports since it doesn't reference anything from `gofr` or `http/responder`.

#### 🔴 2. Timeout bypass race window

**File:** `pkg/gofr/handler.go:111-114`

```go
if _, ok := result.(response.SSE); ok {
    c.Context = r.Context()
}
```

This runs AFTER the `select` block. If `requestTimeout` is very short (e.g., 1ms), the timeout fires before the handler goroutine closes `done`. The `select` picks `c.Context.Done()`, sets `err = ErrorRequestTimeout`, and `result` is nil. The SSE type check on line 112 fails → SSE stream never starts, user gets a timeout error.

WebSocket avoids this by detecting from request headers BEFORE the handler runs (line 60-62). SSE can't do that — it's detected from the response.

**Suggested fix:** Set a context value in `SSEResponse` before returning, and check it in the handler alongside the WebSocket check:

```go
// In SSEResponse:
// Set a flag in context before handler returns

// In handler.go, alongside WebSocket check:
if websocket.IsWebSocketUpgrade(r) || isSSEFromContext(c.Context) {
    c.Context = r.Context()
}
```

Or at minimum, document that SSE handlers should return near-instantly (the `SSEResponse` call is just struct creation, which is fast — but the race exists in theory).

#### 🟡 3. `Debugf(fmt.Sprintf(...))` double-formatting

**File:** `pkg/gofr/http/responder.go:330`

```go
r.logger.Debugf(fmt.Sprintf("SSE stream error: %v", err))
```

`Debugf` already calls `fmt.Sprintf` internally. This should be:
```go
r.logger.Debugf("SSE stream error: %v", err)
```

#### 🟡 4. Integration tests leak goroutines

**File:** `pkg/gofr/sse_test.go:342, 401`

Both `TestSSEResponse_Integration` and `TestSSEResponse_Integration_ClientDisconnect` call `go app.Run()` without ever shutting down. The HTTP server, heartbeat goroutines, and listener leak for the rest of the test process. Add `t.Cleanup` with shutdown.

#### 🟡 5. Missing test: nil/invalid callback through responder

`TestSSEResponse_NilCallback` (line 440) tests `SSEStream` directly — it never exercises the nil guard in `handleSSEResponse` (responder.go:311). Add:

```go
func TestHandleSSEResponse_NilCallback(t *testing.T) {
    w := httptest.NewRecorder()
    r := NewResponder(w, http.MethodGet)
    r.Respond(response.SSE{}, nil) // zero-value SSE, Callback is nil
    // Should not panic, should return empty response
}
```

#### 🟡 6. Missing test: callback error is logged

The error logging at `responder.go:328-332` is untested. Add a test where the callback returns an error and assert the debug log fires.

---

### Summary

The core architecture is solid — SSE via standard `app.GET()` returning `response.SSE`, sequential callback pattern, mutex-protected writes, heartbeat goroutine, proxy headers. Most review feedback was addressed well.

**Blocking before merge:**
1. Fix `Callback any` → use a typed callback in the `response` package
2. Address the timeout race (at minimum document it; ideally detect SSE before handler timeout)

**Should fix:**
3. `Debugf(fmt.Sprintf(...))` → `Debugf(...)`
4. Integration test shutdown cleanup
5. Add missing responder-level tests (nil callback, error logging)

Please also resolve the merge conflicts flagged in the latest comment.

IronwallxR5:
@coolwednesday , I have resolved the merge conflict. Thanks for the re-review! I will address the feedback and update the PR shortly.

IronwallxR5:
@coolwensday I have applied your re-review feedback. Please let me know if any additional changes are needed.

IronwallxR5:
Hi, just following up. Please let me know if any more changes are required from my side.

aryanmehrotra (changes requested):
Implementation looks fine but the way how the user is supposed to create SSE endpoints is not correct, can we simplify that and handle it internally, can't we do it by returning a normal struct or something like that, we need to simplify the usage

IronwallxR5:
> Implementation looks fine but the way how the user is supposed to create SSE endpoints is not correct, can we simplify that and handle it internally, can't we do it by returning a normal struct or something like that, we need to simplify the usage

I have replaced the `SSEResponse()` wrapper function with a simple struct that users return directly:

```go
return gofr.SSE{
    Callback: func(stream *gofr.SSEStream) error {
        stream.SendData("hello")
        return nil
    },
}, nil
```

The wrapping logic (SSEStream creation, heartbeat) is now handled internally when the handler returns. Follows the same pattern as `response.File{}` and `response.Raw{}`.

aryanmehrotra:
@IronwallxR5 it still depends on a function if an endpoint need to be changed to support SSE, it requires a lot of change for the user, can we discuss different approaches and then come back to implement it, this doesn't look clean

IronwallxR5:
> @IronwallxR5 it still depends on a function if an endpoint need to be changed to support SSE, it requires a lot of change for the user, can we discuss different approaches and then come back to implement it, this doesn't look clean

That's fair, happy to discuss! Here are a few different approaches I can think of:

Option 1: Stream via context — handler writes directly, no wrapper needed

```go
app.GET("/events", func(c *gofr.Context) (any, error) {
    stream := c.Stream()
    stream.SendData("hello")
    return nil, nil
})
```
Option 2: Channel-based — return a channel, framework drains it as SSE

```go
app.GET("/events", func(c *gofr.Context) (any, error) {
    ch := make(chan gofr.SSEEvent)
    go func() {
        ch <- gofr.SSEEvent{Data: "hello"}
        close(ch)
    }()
    return ch, nil
})
```
Option 3: Keep current struct, drop the Callback field — the handler itself is the "streaming body", runs in a streaming context automatically when it returns gofr.SSE

```go
app.GET("/events", func(c *gofr.Context) (any, error) {
    c.SendSSE(gofr.SSEEvent{Data: "hello"})
    return nil, nil
})
```
I think Option 1 feels the cleanest since the handler stays flat with no nesting. What do you think? Open to other ideas too.

aryanmehrotra:
This has been done along with the PR #3722

PR #3100, opened 2026-03-03
fix: resolve PubSub migration conflict by decoupling from version chain

### LINKED ISSUE #3099: Bug: PubSub Migrator reports incorrect version due to ghost data in Redis Streams
### Description
The PubSub migrator incorrectly reports a higher migration version than the actual migration version when configured with Redis Streams (or other persistent backends). This happens because legacy migration messages remain in the stream ('ghost data'), and the migrator treats the PubSub backend as an authoritative source for the global migration version.

### Impact
Legitimate migrations in the primary database (SQL, Redis DB 0, etc.) are skipped because the framework believes a 'future' migration has already occurred according to the PubSub bus.

### Proposed Fix
Decouple PubSub from the migration versioning chain. It should remain available as a data source for creating topics in migration scripts but must not participate in determining the current migration version of the application.

coolwednesday:
# PubSub Migration Conflict Resolution

- Closes #3099 

## Problem Statement
The GoFr **PubSub Migrator**, particularly when configured with **Redis Streams**, erroneously reported migration versions based on historical data present on the message bus. Unlike standard Redis Pub/Sub, Redis Streams persist messages. If an app instance (or another app sharing the same Redis cluster) had previously published a migration log to the `gofr_migrations` stream, any new instance would see these "ghost" messages.

If the version in these ghost messages was higher than the actual local migration state in the primary database (e.g., SQL or Redis DB 0), GoFr would incorrectly conclude that those migrations were already completed and skip them.

## Root Cause
The `PubSubMigrator` was acting as an **authoritative participant** in the migration versioning chain. It would query the message bus and override the local database version if the bus reported a higher number. This creates a "race to the top" where the highest version ever seen on the network becomes the global minimum, regardless of actual database state.

## Solution: Decoupled Versioning
The fix involves making the PubSub data source **non-authoritative** for migration state:
1.  **Chain Removal**: The `pubsubDS` now implements a "no-op" application in the migrator chain. It returns the next migrator (e.g., Redis or SQL) unchanged.
2.  **Datasource Availability**: The `PubSub` client remains fully available in the `migration.Datasource` object. This ensures that users can still call `d.PubSub.CreateTopic()` inside their `UP` functions.
3.  **State Elimination**: Automated framework-level publishing of migration JSON logs to PubSub has been removed. Migration tracking is now strictly confined to primary data stores that support atomicity and locking.

### Automated Tests
*   **Integration Test**: A new test case `Test_PubSub_GhostDataConflict` was added to `pubsub_test.go`. It seeds a `miniredis` instance with a "future" version (`20260101000000`) in a separate DB and verifies that the migrator ignores it, correctly running migrations starting from version 0.

### Manual Verification
*   Tested with `using-publisher` and `using-subscriber` examples.
*   Injected ghost data manually into Redis DB 15.
*   Confirmed that the app correctly identified version `0` from Redis DB 0 and ran its local migrations, ignoring the noise on the PubSub bus.

**Checklist:**


aryanmehrotra (changes requested):
Update the migration docs

docs/advanced-guide/handling-data-migrations/page.md
```diff
@@ -284,6 +284,8 @@ Explanation:
 **Method** : It indicates whether the migration ran in UP or DOWN mode.
 (For now only method UP is supported)
 
+> **Note**: For Redis migration using **Streams mode**, a consumer group ID is mandatory. An empty string (`""`) for the group ID will result in an error during subscription, however, publishing will still succeed.
```
aryanmehrotra: An empty string (`""`) for the group ID will result in an error during subscription, however, publishing will still succeed.

-> we can just remove the empty string mentiond above in the documentation, to empty 

"" - telling this as an empty string explicitly doesn't make sense as its already intuitive that its empty

docs/advanced-guide/handling-data-migrations/page.md
```diff
@@ -394,4 +396,41 @@ func bulkProducts() migration.Migrate {
 	}
 ```
 
+## PubSub in Migrations
+
+GoFr provides support for interacting with PubSub systems during migrations. This is particularly useful for setting up your infrastructure (e.g., creating or deleting topics) before your application logic starts using them.
```
aryanmehrotra, line 401: Also add some example where it is not created automatically.

docs/advanced-guide/handling-data-migrations/page.md
```diff
@@ -394,4 +396,41 @@ func bulkProducts() migration.Migrate {
...
+
+### Non-Authoritative Version Tracking
+
+Unlike SQL or Redis, which GoFr uses to store the authoritative record of which migrations have been executed, **PubSub is non-authoritative**. 
+
+**Why this is important:**
+Many PubSub backends (like Redis Streams or Kafka) persist messages even after they are consumed. If GoFr used the PubSub bus as a source of truth for migration versions, "ghost data" from previous runs or other environments could interfere with the migration process, causing legitimate migrations to be skipped.
```
aryanmehrotra: can we directly tell, that we dont store the migration records in pubsub and the reasoning

Umang01-hash (approved):

PR #3103, opened 2026-03-06
feat(pubsub): add span links in NATS

### LINKED ISSUE #3014: Add span links in NATS

adityakrmishra:
Closes #3014

Hyy everyone!  
I'm a 12th-grade student and this is my first time contributing to this project. I picked up the issue to add OpenTelemetry span links for NATS. 

As suggested in the issue thread, I looked at how this was recently done for SQS and basically mirrored that same logic over to the NATS implementation.

**Here is what I changed:**
- Created `tracing.go` to handle injecting and extracting the trace context. Since NATS uses `nats.Header` (which is essentially a `map[string][]string`), I used that as the carrier.
- Updated `connection_manager.go` to use `PublishMsg` instead of the standard `Publish`. This lets us actually attach the headers containing the trace info to the message envelope.
- Updated `subscription_manager.go` to extract those headers and link the publisher's span to the new subscriber span.
- Added unit tests in `tracing_test.go` to make sure the coverage doesn't drop.
- Tweaked `connection_manager_test.go` because adding the tracing changed the context to a `*context.valueCtx`, which was making the old strict `gomock` expectations fail. I updated them to `gomock.Any()` for the context.

https://github.com/user-attachments/assets/24fa1562-44a8-41ba-9520-b788ac2010f5

All the local NATS tests are passing for me now. Let me know if I missed anything or if I need to adjust any formatting! Thanks for the guidance on the issue thread.


pkg/gofr/datasource/pubsub/nats/subscription_manager.go
```diff
@@ -171,24 +172,28 @@ func (sm *SubscriptionManager) processFetchedMessages(
...
 	logger pubsub.Logger) error {
 	for msg := range msgs.Messages() {
-		pubsubMsg := sm.createPubSubMessage(msg, topic)
+		pubsubMsg, span := sm.createPubSubMessage(msg, topic)
 
 		if !sm.sendToBuffer(pubsubMsg, buffer) {
 			logger.Logf("Message buffer is full for topic %s. Consider increasing buffer size or processing messages faster.", topic)
 		}
+
+		span.End()
```
Umang01-hash: The subscribe span **ends as soon as the message enters the buffer — before any consumer handler runs**.; should end on Commit/Nak .  The span should be tied to the full message lifecycle.

pkg/gofr/datasource/pubsub/nats/tracing_test.go
```diff
@@ -0,0 +1,211 @@
...
+	keys := carrier.Keys()
+	assert.Contains(t, keys, "traceparent")
+	assert.Contains(t, keys, "tracestate")
+
+	// Test Set updates existing key
+	carrier.Set("traceparent", "00-updated-value")
+	assert.Equal(t, "00-updated-value", carrier.Get("traceparent"))
+}
+
+func TestHeaderCarrier_GetMissingKey(t *testing.T) {
```
Umang01-hash: This is already covered inside TestHeaderCarrier_GetSetKeys:
  assert.Empty(t, carrier.Get("nonexistent")). We can remove this test.

pkg/gofr/datasource/pubsub/nats/subscription_manager.go
```diff
@@ -171,24 +172,28 @@ func (sm *SubscriptionManager) processFetchedMessages(
...
 	}
 
 	return sm.checkBatchError(msgs, topic, logger)
 }
 
-func (*SubscriptionManager) createPubSubMessage(msg jetstream.Msg, topic string) *pubsub.Message {
-	pubsubMsg := pubsub.NewMessage(context.Background()) // Pass a context if needed
+func (*SubscriptionManager) createPubSubMessage(msg jetstream.Msg, topic string) (*pubsub.Message, trace.Span) {
+	ctx, span := startSubscribeSpan(context.Background(), topic, msg.Headers())
```
Umang01-hash: `startSubscribeSpan` is called with **context.Background()** instead of the subscription context. This means every subscriber span starts a brand-new root trace, losing any parent span that may have been active when Subscribe was called.

  The subscription context is already available in **consumeMessages** but gets discarded in **fetchAndProcessMessages** (note the _ context.Context parameter). 

We should thread it through the call
   chain all the way down to `createPubSubMessage` and pass it to `startSubscribeSpan`.

Umang01-hash:
@adityakrmishra Please go ahead and resolve the review comments so we can further review and merge your PR if approved.

pkg/gofr/datasource/pubsub/nats/subscription_manager.go
Umang01-hash: Fixed — span is now stored in **natsCommitter** and ended via `defer span.End() `in `Commit(), Nak()`, and `Rollback()`, tying the span to the full message lifecycle.

pkg/gofr/datasource/pubsub/nats/tracing_test.go
Umang01-hash: Removed **TestHeaderCarrier_GetMissingKey** as the assertion was redundant with the existing coverage in **TestHeaderCarrier_GetSetKeys**.

pkg/gofr/datasource/pubsub/nats/subscription_manager.go
Umang01-hash: Fixed — threaded `ctx` through **fetchAndProcessMessages** → **processFetchedMessages** → **createPubSubMessage** → **startSubscribeSpan**, so subscriber spans now correctly inherit the
  subscription context instead of starting as orphan root traces.

pkg/gofr/datasource/pubsub/nats/go.mod
```diff
@@ -6,6 +6,8 @@ require (
 	github.com/nats-io/nats-server/v2 v2.11.12
 	github.com/nats-io/nats.go v1.48.0
 	github.com/stretchr/testify v1.11.1
+	go.opentelemetry.io/otel v1.41.0
```
coolwednesday: **Version mismatch:** `otel/trace` is at `v1.42.0`, `otel` core is at `v1.41.0`, and `otel/sdk` is at `v1.40.0`. These should all be aligned to the same minor version to avoid subtle incompatibilities. Please bump them all to `v1.42.0`.

pkg/gofr/datasource/pubsub/nats/committer.go
```diff
@@ -4,15 +4,19 @@ import (
 	"log"
 
 	"github.com/nats-io/nats.go/jetstream"
+	"go.opentelemetry.io/otel/trace"
 )
 
 // natsCommitter implements the pubsub.Committer interface for Client messages.
 type natsCommitter struct {
-	msg jetstream.Msg
+	msg  jetstream.Msg
```
coolwednesday, line 13: The span ends via `defer` on all exit paths, but **errors are never recorded on the span**. When `Ack()` fails, you should record the error on the span before it ends:

```go
if err := c.msg.Ack(); err != nil {
    c.span.RecordError(err)
    c.span.SetStatus(codes.Error, "ack failed")
    ...
}
```

Same applies to `Nak()` and `Rollback()` — errors are returned but not captured in the trace.

pkg/gofr/datasource/pubsub/nats/connection_manager_test.go
```diff
@@ -93,10 +93,10 @@ func TestConnectionManager_Publish(t *testing.T) {
 	message := []byte("test message")
 
 	gomock.InOrder(
-		mockMetrics.EXPECT().IncrementCounter(ctx, "app_pubsub_publish_total_count", "subject", subject),
+		mockMetrics.EXPECT().IncrementCounter(gomock.Any(), "app_pubsub_publish_total_count", "subject", subject),
 		mockConn.EXPECT().Status().Return(nats.CONNECTED),
```
coolwednesday, line 114: **Weakened assertion:** Changing from exact `ctx`, `subject`, `message` matching to `gomock.Any()` loses the guarantee that the correct subject and message body are being published. Consider using a custom matcher to validate the `nats.Msg` fields:

```go
mockJS.EXPECT().PublishMsg(gomock.Any(), gomock.Cond(func(x any) bool {
    msg, ok := x.(*nats.Msg)
    return ok && msg.Subject == subject && bytes.Equal(msg.Data, message)
})).Return(&jetstream.PubAck{}, nil)
```

pkg/gofr/datasource/pubsub/nats/tracing.go
```diff
@@ -0,0 +1,122 @@
...
+// startPublishSpan creates a new span for publishing with trace context injection.
+// Returns the updated context, the span, and NATS headers with injected trace context.
+func startPublishSpan(ctx context.Context, subject string) (context.Context, trace.Span, nats.Header) {
+	opts := []trace.SpanStartOption{
+		trace.WithSpanKind(trace.SpanKindProducer),
+		trace.WithAttributes(
+			attribute.String("messaging.system", "nats"),
+			attribute.String("messaging.destination.name", subject),
+			attribute.String("messaging.operation", "publish"),
+		),
```
coolwednesday, line 90: `otel.GetTracerProvider().Tracer(tracerName)` is called on every publish/subscribe invocation. At high throughput this adds overhead. Consider resolving the tracer once and passing it in, or caching it — similar to how other interceptors in the codebase do `otel.GetTracerProvider().Tracer(...)` once at initialization.

pkg/gofr/datasource/pubsub/nats/tracing_test.go
```diff
@@ -0,0 +1,206 @@
...
+	"context"
+	"testing"
+
+	"github.com/nats-io/nats.go"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
```
coolwednesday, line 13: **Global OTel state mutation without restore:** Every test calls `otel.SetTracerProvider(tp)` and `otel.SetTextMapPropagator(...)` but never restores the original. This can cause flaky behavior if tests run in parallel or if other tests depend on the default provider. Save and restore:

```go
origTP := otel.GetTracerProvider()
origProp := otel.GetTextMapPropagator()
t.Cleanup(func() {
    otel.SetTracerProvider(origTP)
    otel.SetTextMapPropagator(origProp)
})
```

Also, the same 5-line setup is copy-pasted across every test. Extract it into a `setupTestTracing(t *testing.T)` helper.

pkg/gofr/datasource/pubsub/nats/subscription_manager.go
```diff
@@ -181,12 +182,14 @@ func (sm *SubscriptionManager) processFetchedMessages(
 	return sm.checkBatchError(msgs, topic, logger)
 }
 
-func (*SubscriptionManager) createPubSubMessage(msg jetstream.Msg, topic string) *pubsub.Message {
-	pubsubMsg := pubsub.NewMessage(context.Background()) // Pass a context if needed
+func (*SubscriptionManager) createPubSubMessage(ctx context.Context, msg jetstream.Msg, topic string) *pubsub.Message {
+	spanCtx, span := startSubscribeSpan(ctx, topic, msg.Headers())
+
+	pubsubMsg := pubsub.NewMessage(spanCtx)
```
coolwednesday, line 189: There's no test coverage for `createPubSubMessage` verifying that:
1. The returned message's context carries the subscribe span
2. The committer's span is the same one created by `startSubscribeSpan`

The tracing tests only cover the low-level helpers. Please add a test for this integration point.

pkg/gofr/datasource/pubsub/nats/go.mod
Umang01-hash: Bumped **otel** and **otel/sdk** to `v1.42.0` to align all three OTel packages.

pkg/gofr/datasource/pubsub/nats/committer.go
Umang01-hash: **Commit**, **Nak**, and **Rollback** now call `span.RecordError(err)` and `span.SetStatus(codes.Error, ...) `before ending the span.

pkg/gofr/datasource/pubsub/nats/connection_manager_test.go
Umang01-hash: Added a **natsMsgMatcher** that validates both Subject and Data, replacing `gomock.Any()` in the PublishMsg expectation.

pkg/gofr/datasource/pubsub/nats/tracing.go
Umang01-hash: **startPublishSpan** and **startSubscribeSpan** now accept a `trace.Tracer` parameter. 
Callers resolve the tracer via `otel.GetTracerProvider().Tracer(tracerName)` at the call site,
  keeping the package-level constant but avoiding a global variable.

pkg/gofr/datasource/pubsub/nats/tracing_test.go
Umang01-hash: Extracted `setupOTel(t *testing.T)` helper that installs the in-memory provider, sets the propagator, and restores both globals via t.Cleanup. All 8 tests now use it.

pkg/gofr/datasource/pubsub/nats/subscription_manager.go
Umang01-hash: Added `TestSubscriptionManager_createPubSubMessage` that verifies the message **topic/value**, that the returned message context carries a valid span, and that the committer holds the same span.

pkg/gofr/datasource/pubsub/nats/go.sum
```diff
@@ -32,20 +35,32 @@ github.com/nats-io/nkeys v0.4.12 h1:nssm7JKOG9/x4J8II47VWCL1Ds29avyiQDRn0ckMvDc=
...
 github.com/nats-io/nuid v1.0.1/go.mod h1:19wcPz3Ph3q0Jbyiqsd0kePYG7A95tJPxeL+1OSON2c=
-github.com/pkg/diff v0.0.0-20210226163009-20ebb0f2a09e/go.mod h1:pJLUxLENpZxwdsKMEsNbx1VGcRFpLqf3715MtcvvzbA=
 github.com/pkg/errors v0.9.1 h1:FEBLx1zS214owpjy7qsBeixbURkuhQAwrK5UwLGTwt4=
 github.com/pkg/errors v0.9.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
 github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 h1:Jamvg5psRIccs7FGNTlIRMkT8wgtp5eCXdBlqhYGL6U=
 github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
-github.com/rogpeppe/go-internal v1.9.0/go.mod h1:WtVeX8xhTBvf0smdhujwtBcq4Qrzq/fJaraNFVN+nFs=
-github.com/rogpeppe/go-internal v1.13.1 h1:KvO1DLK/DRN07sQ1LQKScxyZJuNnedQ5/wKSR38lUII=
-github.com/rogpeppe/go-internal v1.13.1/go.mod h1:uMEvuHeurkdAXX61udpOXGD/AzZDWNMNyH2VO9fmH0o=
+github.com/rogpeppe/go-internal v1.14.1 h1:UQB4HGPB6osV0SQTLymcB4TgvyWu6ZyliaW0tI/otEQ=
```
coolwednesday, line 42: **Stale `go.sum` entries:** There are duplicate entries for `otel/sdk` (both `v1.40.0` and `v1.42.0`) and `otel/metric` (both `v1.41.0` and `v1.42.0`). The old versions are leftover from before the version bump. Run `go mod tidy` in the nats submodule to clean these up.

pkg/gofr/datasource/pubsub/nats/go.sum
```diff
@@ -32,20 +35,32 @@ github.com/nats-io/nkeys v0.4.12 h1:nssm7JKOG9/x4J8II47VWCL1Ds29avyiQDRn0ckMvDc=
...
+github.com/rogpeppe/go-internal v1.14.1/go.mod h1:MaRKkUm5W0goXpeCfT7UZI6fk/L7L7so1lCWt35ZSgc=
 github.com/stretchr/testify v1.11.1 h1:7s2iGBzp5EwR7/aIZr8ao5+dra3wiQyKjjFuvgVKu7U=
 github.com/stretchr/testify v1.11.1/go.mod h1:wZwfW3scLgRK+23gO65QZefKpKQRnfz6sD981Nm4B6U=
+go.opentelemetry.io/auto/sdk v1.2.1 h1:jXsnJ4Lmnqd11kwkBV2LgLoFMZKizbCi5fNZ/ipaZ64=
+go.opentelemetry.io/auto/sdk v1.2.1/go.mod h1:KRTj+aOaElaLi+wW1kO/DZRXwkF4C5xPbEe3ZiIhN7Y=
 go.opentelemetry.io/otel v1.42.0 h1:lSQGzTgVR3+sgJDAU/7/ZMjN9Z+vUip7leaqBKy4sho=
```
coolwednesday, line 48: `otel/sdk/metric` is still at `v1.40.0` in `go.sum` while all other otel packages are at `v1.42.0`. This is a transitive dependency so it may be resolved by `go mod tidy`, but please verify it aligns after cleanup.

coolwednesday (changes requested):
The tracer is now passed in, error recording on spans is in place, test helper extracted with global state restore, custom matcher for PublishMsg, and createPubSubMessage integration test added. Two remaining issues:

pkg/gofr/datasource/pubsub/nats/go.sum
Umang01-hash: Sure did a `go mod tidy` and removed them.

pkg/gofr/datasource/pubsub/nats/go.sum
Umang01-hash: it was resolved by `go mo tidy ` i think.

PR #3104, opened 2026-03-09
fix: re-fetch lastMigration under lock to prevent migration race condition

coolwednesday:
## Summary
- **Bug**: When multiple pods start concurrently, each performs an optimistic `getLastMigration` pre-check *before* acquiring the distributed lock. If pod A completes a migration while pod B is waiting for the lock, pod B proceeds with a stale `lastMigration` value and re-runs the already-completed migration (e.g., `ALTER TABLE ADD COLUMN metadata` → `Duplicate column name 'metadata'`).
- **Fix**: Re-fetch `lastMigration` *after* lock acquisition (double-check locking pattern) so only truly pending migrations are executed.
- **Refactor**: Extracted `acquireLockAndRun` helper from `Run` to keep cyclomatic complexity within the `gocyclo` threshold (≤ 10).

## Test plan
- [x] Updated mock expectations in `migration_test.go`, `elasticsearch_test.go`, and `oracle_test.go` to account for the second `getLastMigration` call under the lock
- [x] All existing tests pass (`go test ./pkg/gofr/migration/...`)
- [x] `golangci-lint run ./pkg/gofr/migration/...` reports 0 issues

### DIFF
--- a/pkg/gofr/otel.go
+++ b/pkg/gofr/otel.go
@@ -8,7 +8,7 @@ import (
 
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
-	"go.opentelemetry.io/otel/exporters/zipkin"
+	"go.opentelemetry.io/otel/exporters/zipkin" //nolint:staticcheck // deprecated but kept for backward compatibility
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/sdk/resource"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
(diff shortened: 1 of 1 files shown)


pkg/gofr/migration/migration.go
```diff
@@ -112,12 +116,27 @@ func Run(migrationsMap map[int64]Migrate, c *container.Container) {
...
+	// that may have completed the same migration between our pre-check and lock acquisition.
+	lastMigration, err := mg.getLastMigration(c)
+	if err != nil {
+		c.Fatalf("migration failed: could not verify migration state under lock, err: %v", err)
+
+		return
+	}
+
+	if !hasNewMigrations(keys, lastMigration) {
+		c.Infof("no new migrations to run (verified under lock)")
```
Umang01-hash: This should be `c.Info(...)`

Umang01-hash (changes requested):
- Missing a test for the core race condition fix: **pre-check sees lastMigration=0** (proceeds to lock), but **re-fetch under lock returns 1** (another pod ran it) → should log "**no new migrations to run (verified under lock)**" and skip execution. This is the primary scenario the PR fixes, but it's not tested.

PR #3175, opened 2026-03-17
fix(file): return explicit error from StorageProvider.NewWriter

### LINKED ISSUE #3146: FileSystem.NewWriter not returning explicit error if file Creation Failed
**Describe the bug**
In case the file creation fails, NewWriter returns an object implementing `io.WriteCloser` instead of explicit error, misleading the user that the file creation succeeded. 

```
        f, err := os.Create(name)
	if err != nil {
		return &failWriter{err: err}
	}
```
**To Reproduce**
Try to create a file in directory without write permission to the running user.

**Expected behavior**
Ideally NewWriter should return an explicit error. However, that would be a breaking change. 
As a workaround, we could type check the response from NewWriter and if it matches `failWriter`, we return an error instead of considering it as success case in the following methods:
- func (c *CommonFileSystem) Mkdir(name string, _ os.FileMode) error
- func (c *CommonFileSystem) Create(name string) (File, error) 
- func (c *CommonFileSystem) handleWriteFlags(name string, flag int) (File, error)

coolwednesday:
## Summary
- Closes #3146
- Changed `StorageProvider.NewWriter` and `MetadataWriter.NewWriterWithOptions` to return `(io.WriteCloser, error)` instead of just `io.WriteCloser`
- Removed the `failWriter` pattern across all providers (local, GCS, Azure, FTP) that silently deferred errors to Write/Close time
- Updated all callers in `CommonFileSystem` (`Create`, `Mkdir`, `handleWriteFlags`, `CreateWithOptions`) to handle the explicit error

## Why
The previous `failWriter` approach masked initialization errors — callers only discovered problems when they tried to `Write()` or `Close()`, making debugging harder. Returning the error explicitly at creation time is the idiomatic Go pattern.

## Impact
This is **not a breaking change** for end users. `StorageProvider` and `MetadataWriter` are internal interfaces not exposed in the public API. Users interact with higher-level `FileSystem`/`FileSystemProvider` methods whose signatures are unchanged.

## Test plan
- [x] All `pkg/gofr/datasource/file` tests pass
- [x] All `pkg/gofr/datasource/file/azure` tests pass
- [x] All `pkg/gofr/datasource/file/ftp` tests pass
- [x] All `pkg/gofr/datasource/file/gcs` tests pass
- [x] Linter passes on all four packages


Umang01-hash:
Hi @coolwednesday ,

Thanks again for opening this pull request! Since it’s been inactive for a few months, I’m going to go ahead and close it for now just to keep our repository queue manageable.

Please don't hesitate to reopen this PR or submit a new one whenever you're ready to pick it back up. We'd still love to have your contribution!

Best,

Umang Mundhra

Issue #3149, opened 2026-03-14
Add migration locks in MongoDB


QUBITABHAY:
I would like to propose adding migration locks for MongoDB to ensure that migrations do not run concurrently across multiple instances of an application.

Since MongoDB is a NoSQL database, the locking approach differs from traditional RDBMS implementations. Below is a proposed approach.

1. Use a dedicated lock collection
     - Create a collection (e.g., `migration_locks`) that stores a lock document.
        - Example structure:
            ```go
            {
              "_id": "migration_lock",
              "lockedAt": <timestamp>,
              "lockedBy": <instance-id>
            }
             ```
     - Before running migrations:
        - Attempt to insert or acquire the lock document.
        - If the lock already exists, the migration process should exit or wait.
        - If the lock is acquired successfully, proceed with the migration.

     - After migrations complete:
        - Delete the lock document to release the lock.

2. Use Atomic Operations
     - MongoDB operations such as:
        - `findOneAndUpdate`
        - `insertOne` with a `unique _id`
        can ensure atomic lock acquisition, preventing race conditions.

3. Optional: TTL Index for Safety
     - To avoid stale locks if a migration process crashes, we can add a TTL index on lockedAt.
         - Example: 
            - Automatically expire lock documents after a configurable duration.

4. Transactions (If Available)
     - If MongoDB transactions are available (MongoDB ≥ 4.0 with replica sets):
     - Lock acquisition and migration metadata updates could run within a session transaction.
     - This would improve the reliability and atomicity of migration operations.

cc: @coolwednesday

coolwednesday:
Hi @QUBITABHAY, 

Assigned it to you.

QUBITABHAY:
@coolwednesday, I have made a pull request

PR #3178, opened 2026-03-18
Implement MongoDB lock and unlock functionality with error handling

QUBITABHAY:
**Description:**

MongoDB migration locking enhancements:

* Added a new lock collection (`mongoLockCollection`) and lock document ID (`mongoLockDocumentID`) for managing migration locks in MongoDB.
* Implemented `lock`, `unlock`, and periodic lock refresh logic in `mongoMigrator` to handle lock acquisition, release, and keepalive, including stale lock cleanup and retry on duplicate key errors.
* Updated migration table initialisation to also create the lock collection, handling "already exists" errors gracefully.

Testing improvements:

* Added unit tests for lock acquisition, release, error handling, and retry logic in `mongo_test.go`, covering scenarios like successful locking, duplicate key errors, and lock release failures.
* Expanded test setup and error definitions to support new lock-related tests, including context handling and mock error values.

fixes: #3149


pkg/gofr/migration/mongo.go
```diff
@@ -98,11 +115,105 @@ func (mg mongoMigrator) rollback(c *container.Container, data transactionData) {
...
+			}
+
+			update := map[string]any{
+				"$set": map[string]any{
+					"lockedAt":  now,
+					"expiresAt": now.Add(defaultLockTTL),
+				},
+			}
+
+			if err := mg.Mongo.UpdateOne(ctx, mongoLockCollection, filter, update); err != nil {
```
Umang01-hash: **UpdateOne in the Mongo interface only returns error — it discards ModifiedCount.** If the lock document has been deleted (TTL expired) and another      
  instance has already acquired it, calling UpdateOne with `filter: {_id: ..., lockedBy: ownerID}` will match 0 documents. MongoDB returns nil error with 0 modified, so the         
  goroutine logs "**Mongo lock refreshed successfully**" without cancelling.                                                                                                           
                                                                                                                                                                                 
  **Concrete failure scenario:**                                                                                                                                                       
  1. Instance A holds the lock, refresh starts failing silently due to network pressure
  2. Lock TTL (15 s) expires after ~3 missed 5 s refreshes                                                                                                                         
  3. Instance B cleans up the stale lock and acquires it                                                                                                                         
  4. Instance A's refresh goroutine wakes up, UpdateOne matches 0 docs → nil error → does not cancel                                                                               
  5. Both instances believe they hold the lock → concurrent migration execution → data corruption      

   Maybe we can think of  replacing UpdateOne with a DeleteOne + InsertOne "refresh-by-replace" strategy.

pkg/gofr/migration/mongo.go
```diff
@@ -98,11 +115,105 @@ func (mg mongoMigrator) rollback(c *container.Container, data transactionData) {
 	c.Fatalf("Migration %v failed.", data.MigrationNumber)
 }
 
+func (mg mongoMigrator) startRefresh(ctx context.Context, cancel context.CancelFunc, c *container.Container, ownerID string) {
```
Umang01-hash, line 122: A lot of code in this method is uncovered. Please add test for **startRefresh**

pkg/gofr/migration/mongo.go
QUBITABHAY: `Refresh-by-replace` (DeleteOne + InsertOne) strategy. The current `UpdateOne` approach is actually safe now because `modified == 0` is detected. The `refresh-by-replace` approach would add `atomicity concerns` since `DeleteOne + InsertOne` isn't atomic — another instance could grab the lock between the two operations. I'd recommend keeping UpdateOne with the modified == 0 check as it's both safer and consistent with how SQL uses UPDATE ... WHERE.

pkg/gofr/migration/mongo_test.go
```diff
@@ -119,3 +129,170 @@ func Test_MongoCommitMigration(t *testing.T) {
...
+		t.Helper()
+
+		done := make(chan struct{})
+
+		go func() {
+			mg.startRefresh(ctx, cancel, mockContainer, "owner-1")
+			close(done)
+		}()
+
+		time.Sleep(5500 * time.Millisecond)
```
Umang01-hash: The test hardcodes `time.Sleep(5500 * time.Millisecond)` (waiting for defaultRefresh = 5s) per sub-test, 5 sub-tests = **27.5 seconds**.

The **startRefresh** tests should either:                                    
  - Skip under testing.Short(), OR                                                                                                                                                 
  - Use an injectable tick interval (pass as a parameter to startRefresh) so tests can use 10ms instead of 5s.

pkg/gofr/migration/mongo.go
```diff
@@ -27,14 +31,30 @@ func (ds mongoDS) apply(m migrator) migrator {
 
 const (
 	mongoMigrationCollection = "gofr_migrations"
+	mongoLockCollection      = "migration_locks"
+	mongoLockDocumentID      = "migration_lock"
```
Umang01-hash: SQL/Redis use `lockKey = "gofr_migrations_lock"` but Mongo uses `mongoLockDocumentID = "migration_lock"`. These should be consistent

pkg/gofr/migration/mongo.go
```diff
@@ -27,14 +31,30 @@ func (ds mongoDS) apply(m migrator) migrator {
...
 func (mg mongoMigrator) checkAndCreateMigrationTable(c *container.Container) error {
 	err := mg.Mongo.CreateCollection(context.Background(), mongoMigrationCollection)
-	if err != nil {
-		c.Debug("Migration collection might already exist:", err)
 
+	if err != nil && !isMongoCollectionExistsError(err) {
+		return err
+	}
+
+	err = mg.Mongo.CreateCollection(context.Background(), mongoLockCollection)
```
Umang01-hash, line 57: there's no test for the "already exists" case (when err is non-nil but **isMongoCollectionExistsError** returns true → should return nil). Please add a test for the same.

akshat-kumar-singhal:
This has a breaking change in UpdateOne but not documented

Umang01-hash:
Yes @akshat-kumar-singhal will make changes in a separate PR to make sure this breaking change is reverted and functionality is still working.

PR #2952, opened 2026-02-03
feat(kafka): add span links for pub/sub tracing

VIDHITTS:
**Description:**

-   Implements OpenTelemetry span links for Kafka pub/sub tracing
-   Addresses issue #2650
-   Enables proper async correlation between producers and consumers following OTel messaging semantic conventions. Previously, subscriber spans created new disconnected traces. Now they create spans with links to producer spans.

**Breaking Changes (if applicable):**

-   No breaking changes. Kafka messages now include trace context headers, which are ignored by consumers not expecting them.

**Additional Information:**

-   Uses `go.opentelemetry.io/otel` for trace context propagation
-   W3C TraceContext format for header injection/extraction
-   This is the first phase (Kafka only). Other backends will follow in separate PRs.

**Checklist:**

-   [x] I have formatted my code using `goimport` and `golangci-lint`.
-   [x] All new code is covered by unit tests.
-   [x] This PR does not decrease the overall code coverage.
-   [x] I have reviewed the code comments and documentation for clarity.

**Thank you for your contribution!**

PR #3101, opened 2026-03-05
add OTel span links for MQTT pubsub tracing        

Umang01-hash:
### Description:

-   Adds **OpenTelemetry tracing instrumentation to the MQTT pubsub datasource**, consistent with the existing implementation for Kafka (#2952), Google PubSub (#3019), and SQS (#3018).

#### MQTT 3.1.1 Limitation:

  Unlike `Kafka`, `SQS`, and `Google PubSub` — which carry trace context in message headers/attributes — **MQTT 3.1.1 has no user properties mechanism**. The traceparent injected on publish cannot be transmitted through the broker to subscribers. As a result, **subscriber spans are correctly created as orphan consumer spans (no link to the producer span).**

### Screenshots:

Screenshot 2026-03-05 at 1 36 53 PM
Screenshot 2026-03-05 at 1 35 38 PM

**Checklist:**

-   [x] I have formatted my code using  `goimport`  and  `golangci-lint`.
-   [x] All new code is covered by unit tests.
-   [x] This PR does not decrease the overall code coverage.
-   [x] I have reviewed the code comments and documentation for clarity.

**Thank you for your contribution!**

PR #3187, opened 2026-03-22
feat: add OpenTelemetry span links to MQTT pub/sub

### LINKED ISSUE #3012: Add span links in MQTT
- Refer https://github.com/gofr-dev/gofr/issues/3015#issuecomment-3949109029

majiayu000:
## Pull Request Template

**Description:**

- Fixes #3012
- Add OpenTelemetry span links to MQTT pub/sub, mirroring the Kafka (PR #2952) and Google PubSub trace propagation patterns.
- MQTT v3.1.1 does not support message-level headers, so trace context is propagated via payload wrapping: publisher injects W3C trace context into a JSON envelope, subscriber extracts it and creates span links to the producer span.
- New file `tracing.go` provides `headerCarrier` (implementing `propagation.TextMapCarrier`), payload wrap/unwrap, and span creation helpers.
- Backward compatible: non-wrapped payloads pass through unchanged.

- None. Published payloads are now wrapped in a JSON envelope with trace context, but subscribers transparently unwrap them. Legacy consumers receiving wrapped messages will see the JSON envelope as the payload.

**Additional Information:**

- Uses `go.opentelemetry.io/otel` and `go.opentelemetry.io/otel/trace` (already project dependencies).
- Pattern mirrors `pkg/gofr/datasource/pubsub/kafka/tracing.go` and `pkg/gofr/datasource/pubsub/google/tracing.go`.

**Checklist:**

### DIFF
--- a/pkg/gofr/datasource/pubsub/mqtt/helper.go
+++ b/pkg/gofr/datasource/pubsub/mqtt/helper.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	mqtt "github.com/eclipse/paho.mqtt.golang"
-	"go.opentelemetry.io/otel"
 
 	"gofr.dev/pkg/gofr/datasource/pubsub"
 )
@@ -45,8 +44,10 @@ func (m *MQTT) createQueryMessageHandler(ctx context.Context, msgChan chan<- *pu
 		messageCtx := context.WithoutCancel(ctx)
 		message := pubsub.NewMessage(messageCtx)
 
+		_, payload := unwrapPayload(msg.Payload())
+
 		message.Topic = msg.Topic()
-		message.Value = msg.Payload()
+		message.Value = payload
 		message.MetaData = map[string]string{
 			"qos":       string(msg.Qos()),
 			"retained":  strconv.FormatBool(msg.Retained()),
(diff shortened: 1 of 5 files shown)


Umang01-hash:
Hey @majiayu000 ! Thanks for the effort to make GoFr better. I also opened a PR #3101 to solve the same issue but the PR was closed as:

> MQTT does support metadata that can help facilitate smoother span link additions. However, we should address this after upgrading MQTT or by identifying a simpler approach to achieve proper correlation.
> 
> At the moment, adding independent span links does not seem to serve the intended purpose effectively. Hence, I am closing this PR for now.
> 
> I request the author to please discuss the proposed approach on the issue ticket first. Once the approach is reviewed and approved, the PR can be reopened or a new PR can be created accordingly.

The core issue with both approaches comes down to MQTT protocol versions:                                                                                                        
                                                                                                                                                                                   
  - **MQTT 3.1.1 (what we currently use)** has no concept of message-level metadata/headers — so trace context has nowhere to live natively.
  - **MQTT 5.0** introduced User Properties, which work exactly like Kafka headers or SQS message attributes — a clean key-value metadata map perfect for injecting traceparent.       
                                                                                                                                                                                   
  Your payload-wrapping approach does achieve span correlation, but it has a significant downside: it changes the actual message content. Any subscriber not using GoFr (or an  older GoFr version) will receive a JSON envelope like {"payload": "...", "traceparent": "..."} instead of the original data. That's a silent breaking change for mixed           
  environments.                                                                                                                                                                    
                                                                                                                                                                                 
  The right path forward, as hinted by the maintainers, **is to upgrade the MQTT driver to support MQTT 5.0**, which would let us inject and extract trace context in User Properties — no payload modification needed, consistent with how Kafka/SQS/Google PubSub do it.

majiayu000:
@Umang01-hash Thanks for your reminder!!

PR #3217, opened 2026-03-27
Fix: Index out of bounds panic in SwaggerUIHandler for files without extension

NitinKumar004:
## Summary
- Replaced `strings.Split` + direct index access with `filepath.Ext` to safely extract the file extension.
- Returns a defined error (`errMissingFileExtension`) when the filename has no extension, instead of panicking with index out of bounds.
- Added unit test for the no-extension scenario.

Fixes #3214

## Test plan
- [x] Added `TestSwaggerUIHandler_NoFileExtension` unit test
- [x] Existing swagger tests still pass
- [x] Linter passes with 0 issues


pkg/gofr/swagger.go
```diff
@@ -42,15 +44,18 @@ func SwaggerUIHandler(c *Context) (any, error) {
 		fileName = "index.html"
 	}
 
+	ext := filepath.Ext(fileName)
+	if ext == "" {
+		return nil, errMissingFileExtension
```
Umang01-hash: The error **errMissingFileExtension** is a plain `errors.New()`, which GoFr maps to **500 Internal Server Error**. A missing file extension is a client error — the request is malformed. It 
  should return **400 Bad Request** or **404 Not Found**

Umang01-hash (changes requested):

aryanmehrotra (approved):

Umang01-hash (approved):

Umang01-hash (approved):

PR #3218, opened 2026-03-27
fix(subscriber): propagate handler error to engage startSubscriber backoff

NitinKumar004:
## Summary
`handleSubscription` logged handler errors with `Errorf` and returned `nil`. Because the error never reached `startSubscriber`, its `err != nil` branch — which sets `delay = 2 * time.Second` — never engaged, so a failing handler caused a tight resubscribe loop on every consecutive failure.

This PR:
- Returns the handler error from `handleSubscription` so `startSubscriber`'s existing 2s backoff engages on failure. The recovered-panic sentinel (`errSubscriberHandlerPanic`, added in #3424) still returns `nil` so `panicRecovery`'s log isn't duplicated.
- Resets `delay` to `0` after a successful iteration in `startSubscriber`. Previously a single failure pinned all subsequent successful iterations at the 2s cadence.

Rebased on top of #3424.

> **Note on commit/redelivery behaviour:** this is not a "data loss" fix — the handler-error path never reached `msg.Commit()` before either, so the broker already redelivers. The observable symptom this fixes is the tight loop, not message loss.

Fixes #3215

## Test plan
- `TestSubscriptionManager_handleSubscription_HandlerErrorDoesNotCommit` (from #3424) is updated to assert the handler error propagates via `errors.Is` while still asserting no commit happens.
- Existing `SuccessCommits` and `PanicDoesNotCommit` from #3424 remain.


Umang01-hash (changes requested):
- **Previously:**

```
{"level":"DEBUG","message":{"messageValue":"{\"orderId\":\"ORD-001\",...}","topic":"order-logs",...}}                                                                                           
  {"level":"INFO", "message":"handler received order #1: {OrderId:ORD-001 Status:FAILED}"}                                                                                                        
  {"level":"ERROR","message":"error in handler for topic order-logs: handler: downstream DB unavailable"}                                                                                         
                                               ↑ ONE error log, then immediately next message:                                                                                                    
  {"level":"DEBUG","message":{"messageValue":"{\"orderId\":\"ORD-BEFORE-FIX\",...}","topic":"order-logs",...}}                                                                                    
  {"level":"INFO", "message":"handler received order #2: {...}"}                                                                                                                                  
  {"level":"ERROR","message":"error in handler for topic order-logs: handler: downstream DB unavailable"}                                                                                         
                                               ↑ NO DELAY — tight loop!      
```                                                                                                                     
                                                                                                                                                                                                
  **After fix (return err — PR change):**                                                                                                                                                             
                                                                                                                                                                                                
```
  {"level":"DEBUG","message":{"messageValue":"{\"orderId\":\"ORD-001\",...}","topic":"order-logs",...}}                                                                                           
  {"level":"INFO", "message":"handler received order #1: {OrderId:ORD-001 Status:FAILED}"}                                                                                                      
  {"level":"ERROR","message":"error in handler for topic order-logs: handler: downstream DB unavailable"}                                                                                         
  {"level":"ERROR","message":"error in subscription for topic order-logs: handler: downstream DB unavailable"}
                                 ↑ TWO error logs for the same error — from handleSubscription AND startSubscriber                                                                                
                                 ↑ Then 2-second delay before next message is attempted                                                                                                           
  {"level":"DEBUG","message":{"messageValue":"{\"orderId\":\"ORD-BEFORE-FIX\",...}","topic":"order-logs",...}}                                                                                    
                                               ↑ Next message 2 seconds later ✅ 
```                                                                                                                  
                                                                                                                                                                                                  
**Every handler error fires both** "error in handler for topic..." and "error in subscription for topic...". In a busy subscription loop under real failure conditions this doubles alert noise.                                         

Maybe we can try a small fix: remove the Errorf on `subscriber.go:70` and let startSubscriber's existing log (line 38) be the single error entry point.
   That keeps error propagation clean and logs the error exactly once.

Umang01-hash:
Conflicts with #3424 (merged) — needs rebase. After rebase the `err != nil` branch should keep the existing `Errorf` and `return err` (so `startSubscriber`'s 2s backoff engages); leave the panic sentinel path returning nil.

Minor:
- "Data loss" framing is off — messages aren't committed on handler error today, broker redelivers. Real symptom is a tight loop because `handleSubscription` returns nil. Worth updating the title/description.
- `subscriber_test.go` will conflict too — #3424 already added `errHandlerFail`; reuse it. `TestHandleSubscription_SubscribeError` isn't testing this PR's change (mock already returned that error) — drop or fold in.
- Follow-up: `startSubscriber` sets `delay=2s` on failure but never resets to 0 on success.

Umang01-hash (approved):

PR #3226, opened 2026-03-30
chore: fix typo in logger comment

singhankit001:
Fixed 'dont' to 'don't' and 'probes is' to 'probes are' in logger.go


pkg/gofr/http/middleware/logger.go
```diff
@@ -140,7 +140,7 @@ func handleRequestLog(srw *StatusResponseWriter, r *http.Request, start time.Tim
 // isLogProbeDisabled checks if probes are disabled to skip logging for default probe paths
 // and additional health check paths of services.
 func isLogProbeDisabled(probes LogProbes, urlPath string) bool {
-	// if probes is not disabled, dont need to check for default probe paths
+	// if probes are not disabled, we don't need to check for default probe paths
```
Umang01-hash, line 143: **If probes are enabled, we don't need to check the default probe paths.** this sounds better i think.

Umang01-hash (changes requested):

PR #3227, opened 2026-03-30
fix(file): correct SFTP/S3 observability defects and stale comments

Vegapunk-debug:
## **Description:**

Started as a comment/typo cleanup in the S3 and SFTP file providers, and grew to
cover the observability defects found alongside them. `s3` and `sftp` are the two
providers not yet migrated to `CommonFileSystem`, so their operation logging had
drifted.

**SFTP — `pkg/gofr/datasource/file/sftp/fs.go`**

* `OpenFile` marked a **failed** open as `SUCCESS`. It was the only one of ten
  methods to do so, so open failures were invisible in both logs and metrics.
* `MkdirAll` reported `MKDIR`, making its traffic indistinguishable from `Mkdir`.
* `Getwd` reported `STAT`, colliding with `Stat`.
* `ChDir` logged an error but returned `nil`, so callers saw a silent success for
  an operation the SFTP client does not implement. It now returns an error and
  records a `CHDIR` observation, matching `CommonFileSystem.ChDir`.
* Operation names now use the `file.Op*` constants instead of string literals.
* `Connect`'s comment claimed it registers metrics; it does not.

**S3 — `pkg/gofr/datasource/file/s3/fs.go`**

* `Rename` returned `nil` for the same-name no-op while still reporting `ERROR`.
* `UseLogger`'s comment referred to the FTP file system.
* Fixed grammar in `Connect`'s comment.

-   **Mention the issue number(s) this PR addresses (if applicable).**
    N/A

-   **Highlight the motivation behind the changes and the expected benefits.**
    Failed SFTP opens were being recorded as successful, and three operations were
    reported under the wrong name, so SFTP file-operation logs and metrics could not
    be trusted. `ChDir` returning `nil` meant callers could not detect that the
    operation had not happened.

## **Reviewer notes — behaviour changes**

This is **not** a no-op PR. Two things worth an explicit look:

1.  **`ChDir` now returns an error** where it previously returned `nil`. Deliberate:
    the previous behaviour reported success for an operation that never ran.
2.  **Metric label values change** for five operations: `GETWD` and `MKDIR_ALL` (both
    corrections), and `OPEN_FILE`, `REMOVE_ALL`, `READ_DIR` (previously `OPENFILE`,
    `REMOVEALL`, `READDIR`, aligning with the constants in `observability.go`). A new
    `CHDIR` series is also emitted. Dashboards keyed on the old spellings will need
    updating.

## **Additional Information:**

-   **Mention any relevant dependencies or external libraries used.**
    None.

-   **Tests**

    The existing tests never inspected the emitted `FileLog`, which is why these
    defects survived. Added a log-capture harness and a table test pinning the
    operation name and status each method reports. Every fix was verified to fail
    the suite when reverted individually.

    `gofmt`, `go vet`, `go test ./...` and `golangci-lint` are clean on both the
    `s3` and `sftp` modules.

## **Checklist:**

# **Thank you!**


Vegapunk-debug:
@aryanmehrotra, please review my PR.

aryanmehrotra (dismissed):

aryanmehrotra:
Thanks for this — the `Getwd` catch is a real bug. It was logging as `STAT`, colliding with `Stat` in both logs and metrics.

Two things happened to the PR, so here's the full picture.

**1. The branch.** This PR was showing a 909-file diff, and that was not your fault. Our `development` history was rewritten at some point, so your fork's copy no longer shared a merge base with ours — git fell back to the repository's root commit, and it couldn't be rebased in place. I've recreated the change on top of current `development` and pushed it to your branch. Your authorship is preserved throughout.

Two of your original hunks were dropped as already fixed upstream (the s3 `New()` and `Create` comments). Two others I adjusted slightly:

- The SFTP `Connect` comment said it "registers metrics", but `Connect()` only dials and creates the client — it never touches metrics, so the reworded version would still have been inaccurate.
- On the s3 `Connect` grammar fix, "creates an S3 client, which then assigns it" still reads as the client assigning itself; changed to "and assigns it to".

**2. Your catch led to four more of the same kind.** Once the operation labels got a proper look:

- `OpenFile` marked a **failed** open as `SUCCESS` — the only one of ten methods doing that, so open failures were invisible in logs and metrics.
- `MkdirAll` reported `MKDIR`, making its traffic indistinguishable from `Mkdir`.
- `ChDir` logged an error but returned `nil`, so callers saw a silent success for an operation the SFTP client does not implement.
- s3 `Rename` returned `nil` for the same-name no-op while still reporting `ERROR`.

Operation names now use the `file.Op*` constants from `observability.go` rather than string literals, which is what allowed them to drift in the first place. The existing tests never inspected the emitted `FileLog` — that's why these survived — so there's now a log-capture harness and a table test pinning the operation name and status each method reports.

The title and description are updated to match: it's a `fix:` rather than `docs:` now, and the description flags the two behaviour changes (`ChDir`'s return value, and the metric label values that move) for reviewers.

One process note for next time: branch off a topic branch rather than your fork's `development`. It keeps the diff isolated and avoids exactly this situation.

Thanks again — a one-line observability fix turned into a genuinely useful cleanup.

Umang01-hash (approved):
LGTM. Verified at head 44ffde5 — both submodules build/vet/lint clean, full suites pass, and both fixes are fail-on-revert (OpenFile SUCCESS→ERROR and s3 Rename same-name ERROR→SUCCESS both turn the new tests red when reverted). Op labels moved to the shared file.Op* constants so they can't drift, and the new log-capture harness pins op-name+status per method instead of only the return value.

Two minor non-blocking notes:
- ChDir now returns an error instead of nil — correct, but a behavior change on an exported method; worth a line in the sftp submodule release notes.
- s3 still uses string-literal op labels while sftp was moved to file.Op* constants; pre-existing and out of scope here, just noting the inconsistency.

PR #3244, opened 2026-03-31
feat: add auth package with file-based token support for K8s projected volumes

akshat-kumar-singhal:
**Description:**
- Adds `pkg/gofr/service/auth` package with `FileTokenAuthConfig`, a `service.Options` that reads a bearer token from disk and re-reads it on a configurable interval. Intended primarily for Kubernetes projected service account tokens.
- Zero-config constructor: `auth.NewFileTokenAuthConfig()` defaults to `/var/run/secrets/kubernetes.io/serviceaccount/token` and a 30s refresh interval. Override via functional options `auth.WithTokenFilePath(path)` and `auth.WithRefreshInterval(d)`.
- Token file is read eagerly at construction so misconfiguration fails fast; subsequent refresh failures keep serving the cached token and are logged at WARN. `Close()` stops the background goroutine and is `sync.Once`-guarded so it is safe to call repeatedly.
- Adds optional `service.Observable` interface (`SetLogger`, `SetMetrics`). `NewHTTPService` calls these on any option that implements `Observable` before `AddOption`, so options that spin up background goroutines do not need a logger threaded through their constructor. Callers that apply an option directly via `AddOption` (outside `NewHTTPService`) must wire logger/metrics themselves.
- `extractHTTPService` now also unwraps anything implementing `interface{ Unwrap() HTTP }`, so the new decorator composes correctly with `ConnectionPoolConfig` / `CircuitBreakerConfig` / `RetryConfig`.
- `AuthErr` implements `Unwrap()` so `errors.Is` / `errors.As` traverse the chain.
- Adds `examples/using-file-token-auth` and a section under `docs/advanced-guide/http-communication`.

**Breaking Changes:** None. Existing `service.NewBasicAuthConfig`, `service.NewAPIKeyConfig`, `service.NewOAuthConfig`, and `AuthErr` are untouched.

**Additional Information:**
- Related discussion: https://github.com/orgs/gofr-dev/discussions/3243

**Checklist:**


Umang01-hash:
Hey @akshat-kumar-singhal ! Thanks for the PR. 

- The new **authProvider** in this package is a different type from `service.authProvider.` **extractHTTPService** in `connection_pool.go` does a type switch on *authProvider (the service package one) to unwrap the decorator chain. When the new auth options are used, it'll hit the default: return nil case — meaning ConnectionPoolConfig, **CircuitBreakerConfig**, and  **RetryConfig** silently stop working when combined with any auth option from this package.

 - These tests files that you deleted cover the existing public types which are still exported and functional (just deprecated). Removing tests for code that's still in use will drop coverage on those code paths. We should keep these tests until the deprecated types are actually removed.
 
 -  The file-token auth feature is a good addition — the use case is real and the refresh logic looks solid. But could we scope this down to just adding **FileTokenAuthConfig** in the existing service/ package, following the same pattern as BasicAuthConfig and OAuthConfig? The restructuring into a sub-package, the new interfaces (Provider, TokenSource,      Observable), and the deprecations are a lot of surface area change for what's essentially one new auth type.

Umang01-hash:
Took a full pass — re-scoping addressed all earlier concerns cleanly. Feature is warranted (K8s projected SA tokens have no existing path), and the `Unwrap()` hook in `extractHTTPService` is the right generalization (avoids the `service` ↔ `service/auth` import cycle, mirrors `errors.Unwrap`). All `./pkg/gofr/service/...` tests pass locally; the regression guard test for pool+auth combo is well-placed.

Two things I'd want before approving:

1. **Refresh failures are silent.** `refreshLoop` swallows errors with `continue`. If the SA token file disappears or perms change, cached token serves stale until upstream 401s — zero observability. Please accept a `service.Logger` in `NewFileTokenAuthConfig` and log a warning on read failure.
2. **Wrap `errAuthHeaderPresent` in `service.AuthErr`** for parity with BasicAuth/OAuth error shapes.

Non-blocking follow-ups: add `tokenCfg.Close()` (or a comment about process-lifetime use) in the example, a verb-parameterized test, a `-race` run, and a docs-site stub under `advanced-guide/http-communication`.

LGTM after (1) and (2).

akshat-kumar-singhal:
Thanks for the thorough pass @Umang01-hash. Addressed both blockers and all four follow-ups in 4bd2583:

**Blocking**
1. **Refresh failures are now observable.** `NewFileTokenAuthConfig` takes a logger and `refreshLoop` logs a WARN on read failure (`file token auth: failed to refresh token from <path>: <err>`) before continuing to serve the cached token. Used `logging.Logger` rather than `service.Logger` since the latter only exposes `Log()` — no real WARN level. It's nil-tolerant for direct `AddOption` callers.
2. **`errAuthHeaderPresent` wrapped in `service.AuthErr`** for parity with BasicAuth/OAuth/APIKey. Also added `AuthErr.Unwrap()` so the sentinel stays reachable via `errors.Is`/`errors.As` (otherwise the wrap is cosmetic) — test asserts both `ErrorAs(AuthErr)` and `ErrorIs(errAuthHeaderPresent)`.

**Non-blocking**
- Example updated with the logger arg + a comment that `tokenCfg` is process-lifetime under `AddHTTPService` (no `Close()` needed unless it outlives the service).
- Added a verb-parameterized test (GET/POST/PUT/PATCH/DELETE) and a refresh-failure-logs-warning test.
- `go test -race ./pkg/gofr/service/...` is green.
- Docs: added a **File-Based Token Authentication** section under `advanced-guide/http-communication`.

One heads-up for review: the logger is a required positional param on `NewFileTokenAuthConfig`. Happy to switch to a functional option (`auth.WithLogger(...)`) if you'd prefer to keep the constructor signature minimal.

aryanmehrotra (changes requested):
Any primary reason we went with a separate package?

  We can put FileTokenAuthConfig in service/ next to BasicAuthConfig/OAuthConfig, but it ended up in the new service/auth subpackage to dodge the import cycle. Thing is, that cycle only shows up because it's a
  subpackage in the first place. If it just went into service/, it could reuse the existing authProvider decorator (that's ~120 lines of copied verb methods gone), and we wouldn't need the new Unwrap() case in
  extractHTTPService at all. Plus all the auth options would be in one spot instead of split across two packages.

  Also on the constructor itself — we're asking the user for a fair bit of plumbing here. fs := file.NewLocalFileSystem(app.Logger()) plus passing app.Logger() again means they're handing us the logger twice and
  wiring up a filesystem, when none of the other auth configs ask for any of that. NewBasicAuthConfig and NewOAuthConfig just take the essentials.

  And the token is always a local file at a known path, right? K8s projected SA tokens live on disk at /var/run/secrets/... — we're never reading them from S3/FTP. So the whole file.FileSystem abstraction isn't
  really buying us anything for the actual use case. If we do want to keep going through file.FileSystem, we can just construct the local one internally ourselves — the user doesn't need to hand it to us. And if
  the local filesystem is missing something we need for reading the token, the right fix is to add it there in the filesystem layer, not to push that wiring onto the user through this constructor. Either way the  plumbing stays inside gofr.

  The path and interval already default internally too, so this could collapse to basically zero-config — auth.NewFileTokenAuthConfig() for the common case, with options to override path/interval when needed, and
  gofr injecting the logger itself (which is what the dropped Observable hook was for). Feels much more in line with how the rest of gofr works — user states intent, framework owns the plumbing. What do you think?

akshat-kumar-singhal:
> Any primary reason we went with a separate package?

Idea is to move all auth related code into the auth package eventually as the service package is getting bloated. In case that's not needed/ direction we're looking at, I can make the changes accordingly. @aryanmehrotra please confirm

> And the token is always a local file at a known path, right? K8s projected SA tokens live on disk at /var/run/secrets/... — we're never reading them from S3/FTP.

Tokens live on disk but the default path may be overriden. Will update the constructor to accept the path override (if applicable).

aryanmehrotra:
@akshat-kumar-singhal if we think from the user's perspectivce, they seeing mix of thing in auth package and service is not good, while the idea is correct but we cannot move everything as well because it would result in breaking behaviour for the user, we should keep in service only, and later deprecate all of them and move to either option package, all the options which we have and later remove deprecated one's.

akshat-kumar-singhal:
@aryanmehrotra makes sense — done. I've moved `FileTokenAuthConfig` out of the new `service/auth` sub-package and into `pkg/gofr/service` directly, so it now sits alongside the existing `NewBasicAuthConfig` / `NewAPIKeyConfig` / `NewOAuthConfig` rather than splitting auth options across two packages. The existing configs are left untouched (no breaking change); deprecating all of them and consolidating into a dedicated option package can follow as a separate step.

Public API is now `service.NewFileTokenAuthConfig(service.WithTokenFilePath(...), service.WithRefreshInterval(...))`. Also renamed the functional-option type `Option` → `FileTokenAuthOption` so it doesn't claim the generic name in the broader `service` package. Example + docs updated, tests green. (a867b16c)

aryanmehrotra (approved):

PR #3323, opened 2026-04-19
fix: handle os.Getwd() error in AddStaticFiles

### LINKED ISSUE #3322: fix: handle os.Getwd() error silently ignored in AddStaticFiles
## Bug Description

## Current Code
```go
Problem


##Expected Behavior

### PR DESCRIPTION
**Pull request by SatishKumar620**
## Summary
Fixes silent error ignored in AddStaticFiles function.

## Changes
- Handled the error returned by os.Getwd() which was previously 
  silently ignored using _
- Added error logging when os.Getwd() fails
- Returns early to prevent incorrect file path resolution

## Related Issue
Closes #3322

## Type of Change
- [x] Bug fix

### CODE REVIEW

**Comment by SatishKumar620** [comment]
I have raised a PR for this issue. Please review.

**Comment by SatishKumar620** [comment]
@vikash

File: pkg/gofr/gofr.go [hunk 1]
```diff
@@ -399,7 +399,11 @@ func (a *App) AddStaticFiles(endpoint, filePath string) {
 
 	// update file path based on current directory if it starts with ./
 	if strings.HasPrefix(filePath, "./") {
-		currentWorkingDir, _ := os.Getwd()
+		currentWorkingDir, err := os.Getwd()
+		if err != nil {
+			a.container.Logger.Errorf("error getting current working directory: %v", err)
+			return
```
aryanmehrotra, line 405: returning from here changes the behaviour moreover no tests are added to ensure the correctness of the change and how it is expected to flow

aryanmehrotra (changes requested):

SatishKumar620:
@aryanmehrotra 
- Replaced `return` with `Warnf` to preserve function behaviour
- Function now continues with relative path if `os.Getwd()` fails
- Added `TestAddStaticFilesGetWdError` to cover this case

Umang01-hash:
@SatishKumar620 your pr is currently failing **Code Quality step** in the workflow. Please update and resolve this.
Screenshot 2026-04-27 at 2 04 57 PM

Umang01-hash (changes requested):
### Review

**1. `gofr_test.go` — test fails on macOS (blocking)**

`os.Remove(tmpDir)` doesn't make `os.Getwd()` fail on darwin (Go falls back to `PWD` / vnode). Verified locally — assertion never matches. Please inject a test seam (`var getwd = os.Getwd`) or skip on darwin.

**2. `gofr_test.go:1193` — assert the precondition (blocking)**

```go
require.NoError(t, os.Remove(tmpDir))
```
Don't swallow the error the whole test depends on.

**3. `gofr_test.go` — trailing blank line before `})` (blocking)**

Likely the cause of the failing Code Quality check. Remove it.

**4. `gofr.go:401` — `Warnf` → `Errorf`? (non-blocking)**

Function can't succeed past this point; `os.Stat` will fail right after. `Errorf` is more accurate.

**5. `gofr_test.go` — strengthen assertion (non-blocking)**

Also assert the downstream `error in registering ... static endpoint` log to prove graceful degradation.

---

**Verdict:** Production fix is correct (verified on live server). Blocking on the test — it doesn't actually exercise the warn branch cross-platform.

aryanmehrotra:
Closing in favour of #3600 which handles this more cleanly.

Two issues here:

**1. Wrong error handling approach** — `Warnf` + continuing with the relative path is unsafe. If `os.Getwd()` fails, the working directory is undefined; using the path as-is could silently resolve to the wrong location. The correct pattern (consistent with how `AddStaticFiles` already handles a missing directory) is `Errorf` + `return`, which is what #3600 does.

**2. Unreliable test** — `os.Remove(tmpDir)` after `t.Chdir` does not make `os.Getwd()` fail on macOS — Go falls back to `$PWD`/vnode, so the assertion never triggers on darwin. The test would only work on Linux.

Thanks for the contribution @SatishKumar620 — the underlying issue is real and worth fixing, the approach in #3600 just gets there more cleanly.

PR #3324, opened 2026-04-20
feat(pubsub): make consumer span a child of producer trace

aryanmehrotra:
## Summary

Make every pub/sub consumer span a real **child** of the producer's trace, while still keeping the existing OTel **span link**. Applied to all four pubsub backends — `kafka`, `nats`, `google` (GCP Pub/Sub) and `sqs`.

Today, `startSubscribeSpan` extracts the producer's trace context from the message headers/attributes and attaches it as a `trace.Link` only. The consumer span itself is started from a fresh root context, so it ends up in a **different trace ID** from the producer. As a result:

- In tracing UIs that only render parent/child (including the GoFr hosted tracer at `tracer.gofr.dev`, and Tempo's default view), an end-to-end flow like `HTTP → publish → subscribe → publish → subscribe` shows up as **3 disconnected traces** even though the link metadata is technically present.
- In Jaeger you see arrows between traces, but you still can't view them as one waterfall.

This PR changes the consumer span to be **started from the producer's extracted context** (so it shares the trace ID and lists the producer as its parent) **and** keeps the link attached for OTel-aware tools that surface fan-out semantics (one message → N consumer groups).

## What changes

For each of `kafka`, `nats`, `google`, `sqs` — `pkg/gofr/datasource/pubsub/<backend>/tracing.go`:

```go
// before
ctx, span := tracer.Start(ctx, \"<backend>-subscribe\", opts...)

// after
parentCtx := otel.GetTextMapPropagator().Extract(ctx, carrier)  // from msg headers
ctx, span := tracer.Start(parentCtx, \"<backend>-subscribe\", opts...)
// link is still attached via trace.WithLinks(...) when present
```

No changes to user-facing API. No exporter changes. No breaking changes to the publish path.

## Verification

Unit tests in each `tracing_test.go` were extended to assert both:

1. The link is preserved (existing assertion).
2. The subscribe span shares the producer's `TraceID` **and** its `Parent.SpanID()` matches the producer's span ID (new assertion).

```
ok  gofr.dev/pkg/gofr/datasource/pubsub/kafka     22.943s
ok  gofr.dev/pkg/gofr/datasource/pubsub/google    36.153s
ok  gofr.dev/pkg/gofr/datasource/pubsub/nats       6.301s
ok  gofr.dev/pkg/gofr/datasource/pubsub/sqs        5.250s

golangci-lint run --new-from-rev=origin/development ...   # 0 issues
```

End-to-end manual verification with a 3-service Kafka topology (api-gateway → order-service → notification-service), exporting to Jaeger via OTLP. A single POST produces **one trace, 6 spans, 3 services, depth 6** — exactly what we want:

```
[api-gateway          ] POST /order         (root)
[api-gateway          ] gofr-router         child of POST /order
[api-gateway          ] kafka-publish       child of gofr-router
[order-service        ] kafka-subscribe     child of api-gateway's publish   [+1 link]
[order-service        ] kafka-publish       child of order-service's subscribe
[notification-service ] kafka-subscribe     child of order-service's publish [+1 link]
```

Screenshot 2026-04-21 at 12 49 11 AM

## Behaviour notes for reviewers

- **Sampling.** Today each subscribe is its own root span and is sampled independently. After this change it inherits the parent's sampling decision via the `ParentBased` sampler GoFr already configures in `pkg/gofr/otel.go:31`. This is the standard OTel behaviour and what users expect — but worth flagging.
- **Long-running async sagas.** A flow that takes hours will produce one long trace instead of many small ones. Most users want this. If trace volume becomes a concern, `TRACER_RATIO` already provides head-based sampling.
- **DLQ / replays.** Re-delivered messages will re-attach to the original trace via the propagated `traceparent` header. This is desirable for correlation, but worth noting.
- **Existing examples (`examples/using-publisher`, `examples/using-subscriber`) need no changes** — the change is fully transparent to user code; existing apps will automatically render as connected traces after upgrading.

## Test plan

- [x] `go test -short ./pkg/gofr/datasource/pubsub/kafka/...` — passes
- [x] `go test -short ./pkg/gofr/datasource/pubsub/google/...` — passes
- [x] `go test -short ./pkg/gofr/datasource/pubsub/nats/...` — passes (one flake on first run, passed cleanly on retry — CI already has retry)
- [x] `go test -short ./pkg/gofr/datasource/pubsub/sqs/...` — passes
- [x] `golangci-lint run --new-from-rev=origin/development` on all four — 0 issues
- [x] `go build ./...` — clean
- [x] End-to-end Kafka demo, traces inspected in Jaeger — single connected trace, fan-out link preserved


Umang01-hash (commented):
A few things before this can go in:

1. The new "Distributed Tracing" section in `using-publisher-subscriber/page.md` lists MQTT, Redis Pub/Sub and Event Hubs as auto-traced — they aren't. MQTT and Redis don't propagate context (MQTT subscribe in `mqtt/helper.go:139` even starts from `context.Background()`), and Event Hubs has no span instrumentation at all. Trim the list to the four backends this PR actually fixes, or split MQTT/Redis into a follow-up.

2. `extractTraceLinks` is no longer called from production in any of the four packages — only from its own tests. Remove the function and `TestExtractTraceLinks*`; `TestStartSubscribeSpan_WithLinks` already covers the behaviour end-to-end.

3. The new `len(msgHeaders) > 0` guard has an untested branch — non-empty headers with an invalid traceparent. One test per backend: pass a header with a garbage value, assert no parent and no link.

4. The biggest behaviour change here is sampling inheritance via `ParentBased`, but nothing asserts it. One line in `TestStartSubscribeSpan_WithLinks`:

```go
assert.Equal(t, producerSpan.SpanContext().TraceFlags(), subSpan.SpanContext.TraceFlags())
```

Approach is right.

Umang01-hash (approved):

Umang01-hash (approved):

PR #3347, opened 2026-04-24
fix(kafka): reconnect stale admin conn instead of failing forever

aryanmehrotra:
## Summary

Kafka subscribers currently get permanently stuck in `"kafka client not connected"` after the broker closes an idle TCP connection (default `connections.max.idle.ms = 10 min`). Once stale, every `Subscribe` / `Query` call returns `errClientNotConnected` for the rest of the process lifetime — `retryConnect()` only covers the initial startup failure, not runtime drops.

## Root cause

- `kafkaClient.conn` is a `multiConn` of admin TCP conns dialed **once** in `initialize()`.
- `isConnected()` calls `conn.Controller()` to probe — which keeps failing on a dead socket because nothing re-dials.
- `Subscribe()` and `Query()` return `errClientNotConnected` whenever `isConnected()` is false and never give the client a chance to heal.
- Publishes keep working because `segmentio/kafka-go`'s `kafka.Writer` handles its own reconnects.

## Fix

- New `ensureConnected(ctx)` — replaces the raw `isConnected` probe at the entry of `Subscribe` and `Query`.
- When stale, it acquires the client mutex, double-checks (another subscribe goroutine may have healed it already), and calls a new `reconnectAdmin(ctx)` helper that:
  - re-dials the broker list via the existing `connectToBrokers`,
  - swaps `k.conn` to a fresh `multiConn`,
  - closes the old conn,
  - leaves writer and per-topic readers alone so in-flight work isn't disrupted.

No public API change. Behavior on a genuinely unreachable cluster is unchanged — the reconnect attempt fails and the caller sees the same `errClientNotConnected` plus a 12 s backoff in `Subscribe`.

## Test plan

- [x] `go test ./pkg/gofr/datasource/pubsub/kafka/...` — all green
- [x] Updated `TestKafkaClient_Subscribe_NotConnected` and `TestKafkaClient_Query_Failures/Client_not_connected` to exercise the reconnect attempt; outward error is unchanged
- [x] New `TestEnsureConnected_ReconnectsAfterStaleAdminConn` pins the re-dial-on-stale behavior (the re-dial fails against an unresolvable host, but the attempt is made)
- [x] `go vet ./pkg/gofr/datasource/pubsub/kafka/...` clean

## How this was surfaced

Hit during a long-running demo (`examples/kafka-tracing-demo`): after ~11 min of idle, subscribers started spamming `"kafka client not connected"` every 12 s and never recovered, even though Kafka itself was healthy and publishes continued to succeed. Full trace of the diagnosis is in aryanmehrotra/go-talks#1.

### DIFF
--- a/docs/advanced-guide/using-publisher-subscriber/page.md
+++ b/docs/advanced-guide/using-publisher-subscriber/page.md
@@ -832,7 +832,7 @@ When the message is delivered to a subscriber registered with `app.Subscribe(top
 
 The result is that an end-to-end flow such as `HTTP → publish → subscribe → publish → subscribe` shows up as **one connected trace** in any tracing UI, with the full waterfall visible:
 
-```
+```text
 [api-gateway          ] POST /order         (root)
 [api-gateway          ] kafka-publish       child of POST /order
 [order-service        ] kafka-subscribe     child of api-gateway's publish   [+1 link]
--- a/pkg/gofr/datasource/pubsub/kafka/conn.go
+++ b/pkg/gofr/datasource/pubsub/kafka/conn.go
@@ -46,13 +46,20 @@ func (k *kafkaClient) initialize(ctx context.Context) error {
 }
 
 func (k *kafkaClient) getNewReader(topic string) Reader {
+	// Snapshot the dialer under connMu — reconnectAdminLocked may swap it
+	// concurrently. Once handed to kafka.NewReader, the reader keeps its
+	// own reference; later reconnects do not affect existing readers.
+	k.connMu.RLock()
+	dialer := k.dialer
+	k.connMu.RUnlock()
+
 	reader := kafka.NewReader(kafka.ReaderConfig{
 		GroupID:     k.config.ConsumerGroupID,
 		Brokers:     k.config.Brokers,
 		Topic:       topic,
 		MinBytes:    defaultMinBytes,
 		MaxBytes:    defaultMaxBytes,
-		Dialer:      k.dialer,
+		Dialer:      dialer,
 		StartOffset: int64(k.config.OffSet),
 	})
 
--- a/pkg/gofr/datasource/pubsub/kafka/conn.go
+++ b/pkg/gofr/datasource/pubsub/kafka/conn.go
@@ -60,14 +60,35 @@ func (k *kafkaClient) getNewReader(topic string) Reader {
 }
 
 func (k *kafkaClient) DeleteTopic(_ context.Context, name string) error {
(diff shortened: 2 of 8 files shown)


Umang01-hash:
A couple of changes before merge:

1. **`kafka.go:149`** — In `Query`, check `query == ""` before `ensureConnected`. No point re-dialing the broker for an obviously bad call. Copilot flagged the same.

2. **Test for the success path** — the new `TestEnsureConnected_ReconnectsAfterStaleAdminConn` only pins the case where the re-dial fails. We don't have a unit test that proves the reconnect actually succeeds and the new `multiConn` is in place. Either inject `connectToBrokers` so it can be stubbed, or add a build-tagged integration test against a real broker.

3. **`helper.go:102`** — the first `isConnected()` reads `k.conn` outside the lock; the write in `reconnectAdmin` happens under `k.mu.Lock()`. Race detector hasn't caught it under stress, but it's still a racy read by the memory model. Wrap the first probe in `RLock`.

Nice to look at in a follow-up (not blocking):
- The `time.Sleep(defaultRetryTimeout)` in `Subscribe` is now stacked on top of the dial timeout inside `ensureConnected`. Worst-case Subscribe latency on a dead cluster goes from ~10s to ~20–30s. Worth trimming.
- Reusing `k.mu` (the reader-map lock) for connection state means a slow reconnect blocks unrelated Subscribe map work. A separate `connMu` would be cleaner once this file grows.
- Pre-existing in `conn.go`: `multiConn.CreateTopics` does `m.mu.RLock()` after the append, and `multiConn.Close()` reads the slice without holding it. Not introduced here, but the new `old.Close()` path makes the window easier to hit. Worth its own issue.

Once 1–3 are in, happy to merge.

Umang01-hash:
Thanks for the locking work — the dedicated connMu, Health() lock, and concurrency tests are great and go beyond what I asked for in (3). Two things still pending before merge: 
 
(1) Query still calls ensureConnected before the empty-query check. Please flip the order — Copilot flagged this too. 
                                                         
 (2) Still no success-path test for the reconnect. A failing-dial test is good but doesn't prove the fix works. Either inject connectToBrokers so it can be stubbed to return a fake set of Connections and we assert k.conn was swapped, or add a //go:build integration test using KAFKA_CFG_CONNECTIONS_MAX_IDLE_MS=15s so the recovery is observable in      
seconds. Your call which.

Umang01-hash (approved):

coolwednesday (approved):

PR #3368, opened 2026-04-30
fix(grpc): nil-check server in Shutdown force-close path

NitinKumar004:
## Summary

- The graceful path of `grpcServer.Shutdown` nil-checks `g.server`, but the force-close fallback (invoked by `ShutdownWithContext` when the shutdown context times out) does not.
- If an app does not call `RegisterService`, `g.server` stays `nil`. A `SIGTERM` whose grace period expires then crashes the process with a nil pointer dereference inside `google.golang.org/grpc.(*Server).stop`.
- This PR adds the missing nil-check in the force-close func and adds a regression test that exercises the force-close path with an already-canceled context.

### Stack from a real crash

```
google.golang.org/grpc.(*Server).stop(0x0, 0x0)            ← receiver is NIL
gofr.(*grpcServer).Shutdown.func2()                        ← grpc.go:194
gofr.(*grpcServer).Shutdown(...)                           ← grpc.go:187
```

### Fix

```go
}, func() error {
    if g.server != nil {
        g.server.Stop()
    }
    return nil
})
```

### Regression test

`TestGRPC_Shutdown_BeforeStart_ContextCanceled` calls `Shutdown` repeatedly with an already-canceled context so the `select` in `ShutdownWithContext` reliably picks the force-close branch. With the fix reverted on this branch, the test reproduces the original panic via `assert.NotPanics`.

## Test plan

- [x] `go test -run TestGRPC_Shutdown -count=3 gofr.dev/pkg/gofr` passes
- [x] New test fails (panics) on pristine `development` and passes with the fix
- [x] `golangci-lint run --new-from-rev origin/development ./pkg/gofr/...` reports 0 new issues


NitinKumar004:
Thanks @Copilot — you're right, the loop test was racy and could pass with the bug reintroduced.

Pushed a follow-up (`bfae5654`) that:

1. **Refactors the two closures inside `grpcServer.Shutdown` into named methods** — `gracefulStop(ctx)` and `forceStop()`. `Shutdown` still wires both through `ShutdownWithContext`, so behavior is unchanged.
2. **Replaces the racy loop test with two deterministic tests** that call those methods directly:
   - `TestGRPC_ForceStop_NilServer` — calls `g.forceStop()` with `g.server == nil`. Asserts no panic.
   - `TestGRPC_GracefulStop_NilServer` — same, for the graceful path.

Both bypass `ShutdownWithContext`'s select entirely, so there's no race for either branch to win. Verified by reverting just the nil-check on `forceStop` — `TestGRPC_ForceStop_NilServer` now fails every run with the original `nil pointer dereference` panic stack inside `grpc.(*Server).stop`.

NitinKumar004:
Update: dropped the refactor commit (`bfae5654`) — keeping `Shutdown` as the original inline closures with nil checks on both branches, no new helper methods. PR is back to the single fix commit `80a3c239`.

Acknowledging @Copilot's race concern: the regression test goes through `Shutdown` → `ShutdownWithContext`, where the canceled-context branch races the graceful goroutine. With `g.server == nil` the graceful path returns immediately, so the loop test isn't 100% deterministic. We chose this trade-off to keep `Shutdown` as a single small function rather than splitting it just for testability — the production code change is the load-bearing fix, and the loop maximizes the odds of exercising the force path under normal scheduler behavior.

Umang01-hash (approved):

PR #3492, opened 2026-05-28
fix(eventhub,s3): stabilize publish test and optimize S3 ReadDir one-level listing

### LINKED ISSUE #3493: [Bug]: Fix flaky EventHub publish test, prevent nil-client crash, and optimize S3 ReadDir memory overhead
### 🐛 Bug Overview

This report details two critical issues regarding runtime safety and resource optimization within the framework:
1. **EventHub:** A missing initialization guard in `Publish` allowed execution to proceed with a `nil` producer client, introducing structural flakiness. Concurrently, the associated failure test was unstable and explicitly skipped.
2. **S3 File System:** The `ReadDir` implementation relied on a wide prefix-only lookup, fetching metadata for all deeply nested objects. This behavior caused unnecessary memory allocation overhead in dense directory structures.

---

#### 1. EventHub Uninitialized State Execution
Creating an EventHub client and attempting a write sequence without an explicit connection hook:
```go
// Configuration is valid, but Connect() lifecycle step is omitted
client := eventhub.New(cfg) 
err := client.Publish(context.Background(), cfg.EventhubName, []byte("msg"))

### PR DESCRIPTION
**Pull request by Manidwiptam**
This commit improves reliability and performance in two areas:

EventHub:

Replaces a flaky, skipped publish test with a deterministic unit test.
Adds a safety guard in Publish to return errClientNotConnected when the producer is not initialized, preventing nil-client failures.
S3:

Refactors ReadDir to list only one-level directory entries using Delimiter and CommonPrefixes.
Avoids loading nested object metadata unnecessarily, reducing memory overhead.
Adds test coverage to validate one-level listing behavior.

Closes #3493

### CODE REVIEW

File: pkg/gofr/datasource/file/s3/fs_dir.go [hunk 1]
```diff
@@ -178,35 +167,24 @@ func (f *FileSystem) ReadDir(name string) ([]file.FileInfo, error) {
...
 
-	// TODO: Enhance the implementation to fetch only data that is one level deep.
-	// Currently, the system retrieves metadata of all files matching the prefix,
-	// which may include files in nested directories. This takes more memory.
 	entries, err := f.conn.ListObjectsV2(context.TODO(), &s3.ListObjectsV2Input{
-		Bucket: aws.String(f.config.BucketName),
-		Prefix: aws.String(filePath),
+		Bucket:    aws.String(f.config.BucketName),
+		Prefix:    aws.String(filePath),
+		Delimiter: aws.String(string(filepath.Separator)),
```
Umang01-hash, line 173: Nit: S3 keys are always `/`-separated, but `filepath.Separator` is `\` on Windows — this would break one-level listing there. Pre-existing assumption across the package (and CI/prod is Linux), so not blocking, but a `"/"` literal would be more correct.

pkg/gofr/datasource/file/s3/fs_dir.go
```diff
@@ -178,35 +167,24 @@ func (f *FileSystem) ReadDir(name string) ([]file.FileInfo, error) {
...
 		return nil, err
 	}
 
-	fileInfo := make([]file.FileInfo, 0)
+	fileInfo := make([]file.FileInfo, 0, len(entries.CommonPrefixes)+len(entries.Contents))
 
 	for i := range entries.Contents {
-		if i == 0 && filePath != "" {
+		// Skip current directory marker object, if present.
+		if *entries.Contents[i].Key == filePath {
```
Umang01-hash, line 184: Nit: `*Key` (and `*Size`/`*LastModified` in the append below) are dereferenced without nil-checks, while the CommonPrefixes loop does nil-check `.Prefix`. AWS sets these in practice, just noting the inconsistency.

Umang01-hash (approved):
LGTM. Two small, correct fixes.

S3 ReadDir: verified live against MinIO — seeding `abc/root.txt` + nested `abc/efg/*` + `abc/hij/*`, `ReadDir("abc")` returns exactly `efg`(dir), `hij`(dir), `root.txt`(file) — nested objects correctly collapsed into CommonPrefixes server-side. Good optimization of the old client-side dedup.

Eventhub: the `producer == nil` guard reuses the existing sentinel and the new deterministic test replaces the flaky skipped one; `TestPublish_FailedInvalidTopic` still passes since it connects first.

gofmt/vet/build/tests clean on both modules, no breaking API change (`getRelativepath` was unexported). Only nits below. Minor: two unrelated fixes in one PR — fine, just noting.

PR #3496, opened 2026-05-31
perf(container): run health checks concurrently, and collapse concurrent probes

### LINKED ISSUE #3456: Health endpoint checks all backends sequentially — latency grows linearly with backend count
## Problem

`container/health.go` calls `HealthCheck()` on SQL, Redis, PubSub, and all registered external datasources **one after another**. Each check has its own timeout (typically 1s). With 4+ backends configured, worst-case response time for the health endpoint is 4+ seconds if backends are slow or timing out.

These checks are independent of each other — there's no reason they can't run concurrently.

Additionally, there's no caching. Every hit to the health endpoint re-checks every backend from scratch. If something is polling health every 5 seconds (which is common for k8s liveness/readiness probes), that's a lot of redundant backend pinging.

## Suggestion

1. Run health checks concurrently using `errgroup` or a similar pattern. Response time becomes the **max** of individual checks rather than the **sum**.
2. Consider a short TTL cache (e.g., 5-10 seconds) for health results, especially for liveness probes that don't need real-time accuracy.

## Where to look

- `container/health.go` — the `Health()` method that sequences all checks
- `datasource/redis/health.go`, `datasource/sql/health.go` — individual health check implementations

yashisrani:
Fixes #3456

## Description

`Container.Health` called every backend in turn, so the endpoint's latency was the **sum** of every check rather than the slowest one.

The checks now fan out, and one round is shared by every concurrent caller. Measured against a real app (live MySQL + Redis + four HTTP dependencies answering in 400ms each):

| | before | after |
|---|---|---|
| one `/health` probe | 1605 ms | **401 ms** |
| 20 concurrent probes | 1623 ms, **80** backend calls | **402 ms, 4** backend calls |

The second row is the one that matters under a real probe load: without collapsing, every prober pays full price and every backend is hit by every prober.

## What changed

- **Concurrent checks.** SQL, Redis, PubSub, the LLM models, the ten context-aware datasources and the HTTP services all run together. Results are gathered by a small collector rather than by passing a map, a mutex and a counter down the call chain.
- **Concurrent probes are collapsed** into one round via `singleflight` (already an indirect dependency; `errgroup` from the same module is used elsewhere in the repo). A slow backend cannot be stampeded by probe traffic, which the cache alone does not give you — a cache only helps *after* a round has finished.
- **`Health` returns a fresh map.** The cached map is shared between callers, so handing it out directly lets one consumer re-key what the next one reads, and race the reader while doing it. The values inside stay shared; the doc comment says so precisely.
- **A panicking check is contained.** These used to run inline on the request goroutine, under the panic-recovery middleware of whichever handler called `Health`, so a driver that panicked cost one request; from a goroutine of our own the same panic takes the process down. `runCheck` is the single point every check passes through, so the `recover` goes there and the backend is reported `DOWN` with the panic value as its detail. Reachable today: `Container.SetLLM` stores a model with no nil guard, and the `llmModels` fan-out has no `isNil` check of its own, unlike `checkExternalDatasources`. (`App.AddLLM` does guard both `nil` and typed-nil.)
- **Rounds do not stack after a timeout.** An abandoned round keeps running — `SQL`, `Redis` and `PubSub` health methods take no context — and `singleflight` releases its key the moment the leader returns, so the next probe started a *second* set of checks against a backend that had not answered the first. Under Kubernetes probe traffic that is one more leaked goroutine every probe interval. A probe arriving while a round is still in flight now reports `DEGRADED` with no per-backend keys, and the claim is released only once the abandoned checks actually finish.

## New configuration — both off unless set

| Name | Description | Default |
|---|---|---|
| `HEALTH_CACHE_TTL` | Serve a result for this long without re-checking (`5s`, `1m`). | disabled |
| `HEALTH_CHECK_TIMEOUT` | Bound one round of checks (`2s`). What has not answered is left out of the body and the aggregate is `DEGRADED`; a partial result is never cached. | disabled |

**Caching is off by default, deliberately.** Inside the TTL a recovered backend still reports `DOWN` and one that has just failed still reports `UP` — that is a trade an operator should opt into, not inherit by upgrading. `singleflight` already removes the stampede that motivated the cache, so the default costs nothing.

## Breaking changes

None. The response body is unchanged — `health_test.go` is untouched and still passes as written.

## Notes for reviewers

- The round runs under `context.WithoutCancel`. It is shared by every waiter, so whichever caller leads must not cancel it by disconnecting.
- All health state hangs off `Container` behind **one pointer**. `injectContainer` sets a `container.Container` field *by value* on a user's gRPC server struct, so a mutex may not live in `Container` itself — `go vet`'s copylocks catches this.
- Checks still running when a timeout fires are abandoned, not cancelled: `SQL`, `Redis` and `PubSub` health methods take no context. They keep writing into the collector, which is why only a snapshot ever leaves it.
- Concurrency is proven **structurally** — a barrier that only releases once checks from three different fan-out groups are in flight together, so sequential execution fails the test instead of a wall-clock bound that flakes on a loaded runner.

## Checklist

- [x] Formatted with `goimports` and `golangci-lint` (lint on the changed packages is *cleaner* than the base: one pre-existing `goconst` remains, the `"UP"` one is gone)
- [x] New code covered by unit tests. Of the functions this PR adds, all are at 100% except `beginRound` (87.5%, the nil-receiver arm) and the double-check inside the singleflight, neither deterministically reachable. Each fix is verified by reverting it: without the `recover` the panic test crashes the test binary rather than failing, and without the in-flight guard four sequential probes start four rounds instead of one.
- [x] Does not decrease coverage
- [x] Verified with `-race -count=3`, and re-measured end to end on the current head against live MySQL + Redis + four 400ms HTTP dependencies: a single probe is 402ms with 4 backend calls, and 20 concurrent probes are 402ms with 4 backend calls. The review fixes cost nothing measurable.


yashisrani:
cc @Umang01-hash @aryanmehrotra

Umang01-hash:
@yashisrani will review it shortly! Thankyou

Umang01-hash:
Tried this out locally. The TTL cache and the cross-category concurrency both work as described, but there's a data race in `checkExternalDBHealth` that needs sorting out before this can go in.

The function takes a `*sync.Mutex` but never uses it — it locks a local `localMu` instead. So the external-DB goroutines guard `healthMap` with one mutex while the SQL/Redis/PubSub/Services goroutines in `Health()` guard the same map with a different one. That's concurrent writes to the same map with no shared lock.

Reproduces reliably:

```
go test -race -run TestContainer_Health ./pkg/gofr/container/
```

and without `-race`, running it in a loop eventually hits `fatal error: concurrent map writes`, which kills the process (it's a fatal error, so `recover()` can't catch it). Since CI runs `-covermode=atomic` and not `-race`, it won't get flagged there. Simplest fix is to pass the existing `mu` through and actually use it, or have the function build a local map and merge it back under the parent lock.

A couple of other things while you're in here:

- The HTTP services loop still runs sequentially inside a single goroutine, so for an app with several downstream services the health latency is still O(n). With 4 services that each take ~300ms I get ~1.2s. The external DBs are parallelized one goroutine each — services should probably get the same treatment, since that's the more common "many backends" case.
- The 5s TTL applies to `/.well-known/health`, which is the deep/readiness check. A backend going down won't surface for up to 5s. Might be worth making the TTL configurable (and disable-able), since not everyone will want stale data on that endpoint.

Also, `health_cache.go` doesn't have direct tests, and the `errgroup` isn't really buying anything here — every goroutine returns nil and the derived context is dropped, so a plain `sync.WaitGroup` would be equivalent.

yashisrani:
@Umang01-hash @aryanmehrotra

Umang01-hash:
Reviewed deeply — built it, ran `-race` (incl. stress), wrote quick tests, and ran a live server. Core is solid, speedup is real. Remaining items are docs/tests, not logic.

Verified:
- `-race` clean, no data races.
- Concurrency real: 3×150ms checks ran in ~152ms (not ~450ms).
- Cache skips re-ping: two `Health()` calls in the TTL → each backend hit once.
- Live `/health` correct (`redis: DOWN`, `sql: UP`, `DEGRADED`), JSON unchanged — no breaking change.

Before merge:
- Document `HEALTH_CACHE_TTL` + the staleness change (down backend shows `UP` up to 5s; default-on). Add to configs docs + release notes.
- Add tests for the two claims (concurrency timing, cache-hit `.Times(1)`) and `getHealthCacheTTL` parsing.
- Description says `errgroup` but code uses `sync.WaitGroup` — fix the description.

Minor:
- `healthCache.get()/set()` panic on nil receiver — add `if hc == nil` guard.
- Scope creep: unrelated comment deletions; `"DEGRADED"` left as literal while `UP`/`DOWN` moved to constants; `clear()` in `health_test.go` is a no-op.
- Negative TTL silently disables (no warn); no singleflight on cache-miss.

yashisrani:
@Umang01-hash please review my pr.

aryanmehrotra:
Thanks @yashisrani — the diagnosis and the approach here were right, and the concurrency win is real: **1605ms → 401ms** on an app with four dependencies answering in 400ms each.

I've rebased the branch onto current `development` and rewritten the implementation, so your commits are no longer in the history — apologies for that. The branch had forked from the 2020 LICENSE commit and showed 961 changed files, which couldn't be reviewed or merged, and it predated LLM health checks, which the rewrite restores. The commit is still authored to you.

Two changes to your design decisions, both open to argument:

- **Caching now defaults to off.** Inside the TTL a recovered backend still reports `DOWN` and one that has just failed still reports `UP` — that felt like a trade an operator should opt into rather than inherit by upgrading. `singleflight` also removes most of what motivated the cache.
- **Added `singleflight`, and a `HEALTH_CHECK_TIMEOUT`.** Caching only helps *after* a round has finished; on a cold or expired window every prober still paid full price and hit every backend. Collapsing the round takes 20 concurrent probes from 1623ms/80 backend calls to 402ms/4.

A few other things picked up along the way, in case they're useful:

- `Health` now returns a fresh map. The cached one is shared between callers, so handing it out directly lets one consumer re-key what the next one reads.
- All the health state has to hang off `Container` behind a pointer — `injectContainer` sets a `container.Container` field *by value* on a user's gRPC server struct, so a mutex in `Container` trips `go vet`'s copylocks.
- The timing test became a barrier that only releases once checks from three different fan-out groups are in flight together. A wall-clock upper bound gets flakier on a loaded CI runner; this gets easier.

`health_test.go` is untouched and still passes as written, which is the best evidence the response body didn't change.

Happy to revisit either of the two decisions above.

pkg/gofr/container/health.go
```diff
@@ -3,101 +3,283 @@ package container
...
+}
+
+// runCheck runs one backend check on its own goroutine, registered with wg.
+func runCheck(wg *sync.WaitGroup, check func()) {
+	wg.Add(1)
+
+	go func() {
+		defer wg.Done()
+
+		check()
```
Umang01-hash, line 279: No recover() here. When checks ran synchronously they were under the handler's panic-recovery middleware (logger.go) → recoverable 500. In a bare goroutine a panic crashes the process. A `defer recover()` that records the backend as DOWN would restore that safety. Concrete trigger: the llmModels loop at L183 calls model.HealthCheck with no nil-guard (checkExternalDatasources uses isNil) — an AddLLM(nil) or a typed-nil model panics.

pkg/gofr/container/health.go
```diff
@@ -3,101 +3,283 @@ package container
...
+		wg.Wait()
+		close(done)
+	}()
 
-		healthMap[key] = health
+	select {
+	case <-done:
+		return true
+	case <-ctx.Done():
+		return false
```
Umang01-hash: On timeout the outstanding checks are abandoned but keep running (SQL/Redis/PubSub take no ctx). Since the leader returns here, the next probe starts a new round while the stuck goroutine persists — ~1 leaked goroutine per timeout interval against a permanently-hung backend under k8s probe traffic. Consider not starting a fresh round while a prior one is still in flight.

Umang01-hash (commented):
Ran it locally — concurrency suite is race-clean (10x), lint clean, and the design holds up: singleflight collapses concurrent probes even with TTL off, partial results are never cached, and WithoutCancel keeps a disconnecting caller from killing shared work. Nice.

Two things I'd tighten before merge (both medium, neither blocks):

1. Check goroutines have no recover(). Before this PR the checks ran inside the handler, so a panicking HealthCheck was caught by the recovery middleware and became a 500. Now they run in bare goroutines, so a panic takes the whole process down. The llmModels loop is the concrete trigger — it calls model.HealthCheck without the isNil guard that checkExternalDatasources has.

2. Under a sustained HEALTH_CHECK_TIMEOUT against a backend that hangs (SQL/Redis/PubSub take no ctx, so the timeout frees the waiter, not the work), each probe after the leader returns starts a fresh round while the stuck goroutine lives on — so goroutines slowly accumulate. Still strictly better than the old serial hang, but worth a guard against overlapping rounds.

Inline below.

pkg/gofr/container/health.go
aryanmehrotra: Fixed in `d47448dd2`. The `recover` went into `runCheck` — the single point all six fan-out groups pass through — so a panicking backend is recorded `DOWN` with the panic value as its detail rather than unwinding a goroutine of ours.

One correction for the record: `App.AddLLM` **does** guard both `nil` and a typed-nil pointer (`external_db.go:102-109`). The unguarded path is `Container.SetLLM` (`container.go:279`), which stores the model directly — so the trigger is real, just not via `AddLLM`. And your underlying point stands regardless of the LLM specifics: the `llmModels` fan-out has no `isNil` check of its own, unlike `checkExternalDatasources`, and *any* backend's `HealthCheck` panicking used to cost one request and now costs the process.

Verified by reverting: without the `recover`, `TestContainer_Health_PanickingBackendIsReportedDown` doesn't fail — it kills the test binary with `panic: boom`, which is exactly the production failure mode you described.

pkg/gofr/container/health.go
aryanmehrotra: Fixed in `d47448dd2`. `healthProbe` now tracks whether a round is in flight, and a probe arriving while one is still running reports `DEGRADED` instead of starting a second set of checks. The claim is released only once the abandoned checks actually finish — releasing it when the leader returns would have left exactly the hole you found.

That response carries no per-backend keys on purpose: what those backends are doing is precisely what is unknown, and reusing the previous round's values would present stale results as fresh. Documented on the `HEALTH_CHECK_TIMEOUT` row in `docs/references/configs/page.md`.

Verified by reverting: without the guard, four sequential probes against a hung backend start four rounds; with it, one. That is `TestContainer_Health_DoesNotStackRoundsAfterTimeout`.

Also worth flagging since it came out of chasing this: the end-to-end numbers in the description were measured before both fixes, so I re-ran them on the current head against live MySQL + Redis + four 400ms dependencies. A single probe is 402ms with 4 backend calls, and 20 concurrent probes are 402ms with 4 backend calls — the fixes cost nothing measurable.

pkg/gofr/container/health.go
```diff
@@ -2,102 +2,330 @@ package container
...
-		key := "llm"
-		if name != "" {
-			key = "llm_" + name
+	if !complete {
+		// The backends that did not answer count as one aggregate failure, so the app reports
+		// DEGRADED rather than a confident UP built out of a partial map.
+		downCount++
+
+		if c.Logger != nil {
+			c.Logger.Warnf("health check timed out after %s, reporting %d of the configured backends",
```
Umang01-hash, line 144: Non-blocking: nothing asserts this timeout warning actually fires (the mock logger doesn't track calls), and the nil-Logger timeout path is untested. Worth a line in the timeout test since the DEGRADED-on-partial behavior hinges on this path.

Umang01-hash (approved):
Verified locally at 70d748d — this is well done. Fan-out + singleflight collapse is the right shape, and the round-guard (endRound deferred to the finished channel so a timed-out round's claim only releases once its abandoned goroutines are done) is correct. Ran the full container package under -race: clean. Mutation-tested the runCheck recover — neutering it makes PanickingBackendIsReportedDown go red, so the panic protection is real and covered. Collapse test asserts a real round counter (rounds==3 for N callers), not just no-error. No breaking API; golangci --new-from-rev clean.

One small thing below, non-blocking.

Umang01-hash (approved):

PR #3498, opened 2026-06-01
fix: add EventHub nil guard and optimize S3 ReadDir memory overhead

### LINKED ISSUE #3493: [Bug]: Fix flaky EventHub publish test, prevent nil-client crash, and optimize S3 ReadDir memory overhead

adityakrmishra:
Fixes #3493

### What does this PR do?
This PR addresses the two runtime safety and resource optimization issues:

1. **EventHub Nil-Client Guard:** Added a strict `c.producer == nil` initialization guard in the `Publish` method to prevent nil-pointer panics if the `Connect()` lifecycle step is omitted. Un-skipped and updated `TestPublish_FailedBatchCreation` to explicitly test this uninitialized state.
2. **S3 File System Optimization:** Optimized the `ReadDir` implementation by adding `Delimiter: aws.String("/")` to the `ListObjectsV2Input`. This restricts the AWS API to only return immediate children, eliminating the memory allocation overhead of wide prefix-only lookups.

### Note to Reviewers regarding Tests:
The specific tests targeted by these changes (`TestPublish_FailedBatchCreation` and `Test_ReadDir`) pass successfully locally. 

*Note:* I observed a few pre-existing test failures locally when running `go test ./...` in these modules (e.g., unexpected `Debugf` calls in EventHub's default consumer group logic, and some S3 mock pointer mismatches). I left these untouched as they are outside the scope of this PR, but wanted to flag them in case the CI pipeline runs into them!

### DIFF
--- a/pkg/gofr/datasource/file/s3/fs_dir.go
+++ b/pkg/gofr/datasource/file/s3/fs_dir.go
@@ -178,12 +178,10 @@ func (f *FileSystem) ReadDir(name string) ([]file.FileInfo, error) {
 		filePath = ""
 	}
 
-	// TODO: Enhance the implementation to fetch only data that is one level deep.
-	// Currently, the system retrieves metadata of all files matching the prefix,
-	// which may include files in nested directories. This takes more memory.
 	entries, err := f.conn.ListObjectsV2(context.TODO(), &s3.ListObjectsV2Input{
-		Bucket: aws.String(f.config.BucketName),
-		Prefix: aws.String(filePath),
+		Bucket:    aws.String(f.config.BucketName),
+		Prefix:    aws.String(filePath),
+		Delimiter: aws.String("/"),
 	})
 	if err != nil {
 		msg = fmt.Sprintf("Error retrieving objects: %v", err)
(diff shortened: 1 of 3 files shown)


aryanmehrotra:
@adityakrmishra the underlying issue #3493 is real, but #3492 covers the S3 ReadDir change more completely — setting the delimiter without the `CommonPrefixes` handling drops subdirectories from `ReadDir` output. Closing in favour of #3492 (its EventHub nil-guard is the same as here). Thanks for the contribution!

PR #3502, opened 2026-06-02
fix: prevent concurrent map panic during kafka client teardown

### LINKED ISSUE #3500: [Bug]: Fatal Process Crash via Concurrent Map Iteration in Kafka Client Teardown (Un-mutexed Access)
### Bug Overview
During a deep architectural review of the framework's graceful shutdown lifecycle, I identified a critical flaw in the pub/sub consumer teardown flow for the Kafka client (`pkg/gofr/datasource/pubsub/kafka/kafka.go`). 

The `kafkaClient.Close()` method currently iterates over the map of topic readers (`k.reader`) without securing the read-lock (`k.mu.RLock()`). Meanwhile, dynamic topic subscriptions in `Subscribe()` securely hold the write-lock (`k.mu.Lock()`) when initializing those same readers.

### Impact Analysis
In enterprise production environments, graceful shutdown is a critical requirement during Kubernetes rolling deployments, pod evictions, or autoscaling events. When GoFr receives a termination signal, it initiates a graceful shutdown sequence to safely drain in-flight requests.

Because `Close()` iterates over the `k.reader` map without acquiring the `k.mu` lock, any concurrent HTTP/gRPC request that triggers a dynamic topic subscription will cause a fatal error: `concurrent map iteration and map write`.

In Go, a concurrent map operation is an unrecoverable panic. This vulnerability instantly brings down the entire application process, completely circumventing the graceful shutdown sequence and dropping all active connections abruptly.

### Root Cause Code
**File:** `pkg/gofr/datasource/pubsub/kafka/kafka.go`

**The Write Side (Safe):** The `Subscribe()` method correctly guards the map mutation using `k.mu.Lock()` during the lazy initialization of topic readers.

**The Read Side (Unsafe):** The `Close()` method iterates over `k.reader` directly without acquiring the `k.mu` mutex.
```go
func (k *kafkaClient) Close() (err error) {
	// CRITICAL FLAW: Missing k.mu.RLock() before iterating the map
	for _, r := range k.reader {
		err = errors.Join(err, r.Close())
	}
    // ...
}
```
### Proof of Concept
1 Initialize a GoFr application with Kafka configured as a pub/sub backend.
2 Expose an endpoint that calls app.Subscribe(ctx, dynamicTopicName).
3 Blast the endpoint with high-concurrency traffic (forcing continuous lazy initialization of k.reader[topic]).
4 Send a SIGTERM (kill -15) to the process to initiate graceful shutdown.
5 The runtime throws fatal error: concurrent map iteration and map write and immediately terminates the process.

We need to synchronize the map iteration inside Close() using the existing sync.RWMutex. However, to avoid holding the lock during blocking I/O operations (closing network connections), the safest approach is to copy the map values into a local slice under a read-lock, release the lock, and then iterate the slice to close the readers.

I can open a PR for this immediately if the maintainers approve of the proposed fix architecture!

adityakrmishra:
Fixes #3500

### What does this PR do?
This PR addresses a critical concurrency vulnerability in the Kafka client's graceful shutdown sequence that causes a fatal `concurrent map iteration and map write` panic.

**Changes made:**
- Updated `kafkaClient.Close()` in `pkg/gofr/datasource/pubsub/kafka/kafka.go`.
- Added a `k.mu.RLock()` block to safely copy the `k.reader` map values into a local slice.
- The lock is immediately released (`k.mu.RUnlock()`) before iterating over the slice to call `r.Close()`.
- This ensures thread safety without holding the mutex during blocking network I/O operations, avoiding potential deadlocks while completely eliminating the crash.

### Note to Reviewers
Ran `go test ./pkg/gofr/datasource/pubsub/kafka/...` locally and all tests pass. This fix guarantees that dynamic topic subscriptions occurring during the shutdown window will not crash the GoFr process.

### DIFF
--- a/pkg/gofr/datasource/pubsub/kafka/conn.go
+++ b/pkg/gofr/datasource/pubsub/kafka/conn.go
@@ -61,11 +61,13 @@ func (k *kafkaClient) initialize(ctx context.Context) error {
 
 	k.dialer = dialer
 	k.conn = multi
-	k.connMu.Unlock()
 
-	// writer and reader are not guarded by connMu — k.mu protects the
-	// reader map; writer is set once and never swapped.
+	// writer is published under connMu alongside conn: retryConnect calls
+	// initialize from its own goroutine, so this is a pointer swap that
+	// Publish, Close and Health can all be reading concurrently.
 	k.writer = writer
+
+	k.connMu.Unlock()
 
 	// retryConnect calls initialize from its own goroutine, so this swap can
 	// land while a Subscribe is reading or growing the map. Take k.mu for it,
--- a/pkg/gofr/datasource/pubsub/kafka/conn.go
+++ b/pkg/gofr/datasource/pubsub/kafka/conn.go
@@ -44,6 +44,21 @@ func (k *kafkaClient) initialize(ctx context.Context) error {
 	k.logger.Logf("connected to %d Kafka brokers", len(conns))
 
 	k.connMu.Lock()
+
+	// Close only signals the retry goroutine; it cannot interrupt a dial
+	// already in flight. Re-check under the same lock Close takes, so a
+	// retryConnect that got past its own check cannot revive a closed client
+	// and leak this conn pool and writer for the life of the process.
+	// A nil k.closed (a client built without New) never fires, so the
+	// default arm is the normal path.
+	select {
+	case <-k.closed:
+		k.connMu.Unlock()
+
+		return errors.Join(errClientClosed, multi.Close(), writer.Close())
+	default:
+	}
+
 	k.dialer = dialer
 	k.conn = multi
 	k.connMu.Unlock()
@@ -51,7 +66,19 @@ func (k *kafkaClient) initialize(ctx context.Context) error {
 	// writer and reader are not guarded by connMu — k.mu protects the
 	// reader map; writer is set once and never swapped.
 	k.writer = writer
-	k.reader = reader
+
+	// retryConnect calls initialize from its own goroutine, so this swap can
+	// land while a Subscribe is reading or growing the map. Take k.mu for it,
+	// the same lock Subscribe uses, and keep any readers a concurrent
+	// Subscribe already created rather than dropping them on the floor —
+	// their committers are already handed out to callers.
+	k.mu.Lock()
+
+	if k.reader == nil {
+		k.reader = reader
+	}
+
+	k.mu.Unlock()
 
 	return nil
 }
--- a/pkg/gofr/datasource/pubsub/kafka/reconnect_concurrency_test.go
+++ b/pkg/gofr/datasource/pubsub/kafka/reconnect_concurrency_test.go
@@ -1,6 +1,7 @@
 package kafka
 
 import (
+	"context"
 	"net"
 	"sync"
 	"sync/atomic"
@@ -9,8 +10,10 @@ import (
 
 	"github.com/segmentio/kafka-go"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 
+	"gofr.dev/pkg/gofr/datasource/pubsub"
 	"gofr.dev/pkg/gofr/logging"
 )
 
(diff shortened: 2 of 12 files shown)


aryanmehrotra (dismissed):
The fix is correct — copying the readers under `RLock` and closing outside it is the right shape, and it does close the concurrent map read in `Close`.

It is incomplete, though. `kafka.go:241` reads the same map outside the lock, eighteen lines after `Subscribe` releases it:

```go
k.mu.Lock()
// ... k.reader[topic] created and read here, correctly guarded ...
k.mu.Unlock()

// ...

m.Committer = newKafkaMessage(&msg, k.reader[topic], k.logger)   // <- unguarded
```

A concurrent `Subscribe` on a different topic takes the write lock and can grow the map while this read is in flight — the same `concurrent map read and map write` panic you are fixing in `Close`, and more likely to fire, since it sits on the subscribe path rather than at teardown.

`reader` is already captured into a local inside the locked section, so the fix is to use that instead of re-reading the map:

```go
m.Committer = newKafkaMessage(&msg, reader, k.logger)
```

Please fold that in — closing one of two unguarded reads leaves the panic reachable, and a follow-up for a two-word change is not worth the round trip.

pkg/gofr/datasource/pubsub/kafka/conn.go
```diff
@@ -44,14 +44,41 @@ func (k *kafkaClient) initialize(ctx context.Context) error {
...
+	default:
+	}
+
 	k.dialer = dialer
 	k.conn = multi
 	k.connMu.Unlock()
 
 	// writer and reader are not guarded by connMu — k.mu protects the
 	// reader map; writer is set once and never swapped.
 	k.writer = writer
```
Umang01-hash, line 68: Non-blocking / pre-existing: `k.writer` is written here unlocked, and on the retry path `initialize` runs from the retryConnect goroutine — so this write can race an unlocked read in Close (kafka.go:290), Health (health.go:126) or Publish (kafka.go:138). It's a single-pointer race, not the map panic, and predates this PR (and the `closed`-recheck shrinks the window), so fine to leave for #3500. Worth a follow-up guarding `k.writer` under connMu like `k.conn`.

Umang01-hash (dismissed):
Reviewed at 73fb89b3 — ran it locally under `-race`. LGTM.

The map panic is real at base (Close and getReaderStatsAsMap ranged `k.reader` unlocked while Subscribe wrote it), and this fixes it cleanly: every `k.reader` access now goes through `k.mu`, and the lock order is consistently `mu → connMu` (Health snapshots readers *before* taking connMu, which is what avoids the inversion). The `closed` chan + closeOnce correctly stops the leaked retryConnect goroutine, and the re-check under connMu in initialize genuinely closes the revive-after-Close window.

Verified each fix is non-vacuous by reverting it and watching the matching test go red under `-race` (Subscribe committer read, Close snapshot, Health snapshot, initialize recheck all confirmed). Full package `-race -count=3` passes (~75s), golangci-lint `--new-from-rev` clean.

One non-blocking note inline. Nice work.

pkg/gofr/datasource/pubsub/kafka/reconnect_concurrency_test.go
```diff
@@ -265,3 +267,45 @@ func TestClose_ConcurrentEnsureConnected_NoRace(t *testing.T) {
...
+
+	connectToBrokers = func(context.Context, []string, *kafka.Dialer, pubsub.Logger) ([]Connection, error) {
+		close(dialed)
+		<-release // hold the dial open until Close has run
+
+		return []Connection{&MockConn{addr: "127.0.0.1:9092", isHealthy: true}}, nil
+	}
+
+	errCh := make(chan error, 1)
+	go func() { errCh <- k.initialize(t.Context()) }()
```
Umang01-hash, line 298: golangci --new-from-rev flags a new wsl_v5 issue here ("missing whitespace above this line"). CI gates on golangci so this fails the lint job — one blank line above the `go func()` fixes it.

Umang01-hash (approved):
Re-reviewed at the new head. The writer changes since my last pass close the unlocked-write nit I'd raised — writer is now published under connMu alongside conn, read via snapshotWriter (RLock, releases before the network call), and nil'd under lock in Close. Lock order mu→connMu is preserved. Ran the writer/lock-order/reader/reconnect tests under -race (green), and confirmed the new writer race test isn't vacuous — reverting Publish to the bare k.writer read trips -race.

Approving. One lint nit below that'll fail the golangci job, worth fixing before merge.

PR #3564, opened 2026-06-09
chore(deps): consolidate minor dependency updates (2026-06-09)

Umang01-hash:
## Consolidated Dependabot Updates (Minor/Patch)

Combines this week's 37 minor/patch Dependabot PRs into one change to reduce review noise and avoid the individual PRs conflicting on `go.sum`/`go.work.sum`. This week Dependabot opened `otel/trace` bumps for nearly every datasource submodule, so the diff is wide but every change traces to a PR.

### OpenTelemetry 1.43.0 → 1.44.0
| Package | Modules |
|---|---|
| otel/trace | root + arangodb, cassandra, clickhouse, couchbase, dbresolver, dgraph, elasticsearch, mongo, opentsdb, oracle, scylladb, surrealdb, kv-store/{badger,dynamodb,nats}, pubsub/{eventhub,nats,sqs} |
| otel/sdk | root, pubsub/nats, pubsub/sqs |
| otel/exporters/prometheus (0.65.0 → 0.66.0) | root |

### AWS SDK
| Package | Old | New | Modules |
|---|---|---|---|
| aws-sdk-go-v2/credentials | 1.19.21 | 1.19.23 | root, file/s3, pubsub/sqs |
| aws-sdk-go-v2/service/s3 | 1.101.0 | 1.103.3 | root, file/s3 |

### Google Cloud
| Package | Old | New | Modules |
|---|---|---|---|
| google.golang.org/api | 0.280.0 | 0.283.0 | root, file/gcs |
| cloud.google.com/go/storage | 1.62.2 | 1.62.3 | root, file/gcs |

### golang.org/x
| Package | Old | New | Modules |
|---|---|---|---|
| crypto | 0.52.0 | 0.53.0 | root, file/sftp |
| text | 0.37.0 | 0.38.0 | root |
| term | 0.43.0 | 0.44.0 | root |

### CI
| Action | Old | New |
|---|---|---|
| crate-ci/typos | 1.47.1 | 1.47.2 |

All minor/patch — no breaking API changes expected. Builds and tests pass across the affected modules.

### Original Dependabot PRs (closed in favour of this one)
#3527, #3528, #3529, #3530, #3531, #3532, #3533, #3534, #3535, #3536, #3537, #3538, #3539, #3540, #3541, #3542, #3543, #3544, #3545, #3546, #3547, #3548, #3549, #3550, #3551, #3552, #3553, #3554, #3555, #3556, #3557, #3558, #3559, #3560, #3561, #3562, #3563

PR #3567, opened 2026-06-09
fix(metrics): configurable OTel cardinality limit + bound built-in path label

### LINKED ISSUE #3566: Metrics: OTel 1.44 default cardinality limit (2000) silently drops series — no way to configure it
### Background

The dependency bump in #3564 moves OpenTelemetry to `otel` v1.44.0 / `exporters/prometheus` v0.66.0. As of OTel Go 1.44, the metrics SDK applies a **default cardinality limit of 2000 attribute sets per instrument** to comply with the Metrics SDK spec ([open-telemetry/opentelemetry-go#8247](https://github.com/open-telemetry/opentelemetry-go/pull/8247)).

When an instrument exceeds 2000 distinct attribute sets, **new label combinations are silently dropped** and aggregated into a single overflow series marked `otel.metric.overflow=true`. Previously GoFr inherited unlimited cardinality, so this is a behavioral change that arrives silently with a `chore(deps)` bump — it won't surface in builds or tests.

### Impact

**1. User-defined custom metrics**
The `Metrics()` API (`IncrementCounter` / `RecordHistogram` / `SetGauge` / `NewUpDownCounter`) accepts arbitrary label key/value pairs. Users who label by naturally high-cardinality dimensions (e.g. `user_id`, `tenant`, `request_id`) will cross 2000 series and start losing data points with no warning.

**2. GoFr's own `app_http_response` histogram**
Labeled by `(path, method, status)`. For most route-templated APIs this is bounded, but the static-file / root branch in `pkg/gofr/http/middleware/metrics.go` substitutes the **raw request URL** for `path`, which is unbounded — apps serving many static paths can exceed 2000 series on this single built-in instrument.

### The gap

GoFr currently exposes **no way to configure the cardinality limit**. The `MeterProvider` is constructed in `pkg/gofr/metrics/exporters/exporter.go` with no view/limit options, and `exporters.Prometheus(appName, appVersion)` (wired in `container.go`) has no parameter to influence it. Users cannot raise the cap, disable it, or even opt into the old unlimited behavior.

### Proposed direction

At minimum, GoFr should restore the pre-1.44 behavior internally so this bump does not silently regress existing apps — i.e. construct the `MeterProvider` with `metricSdk.WithCardinalityLimit(0)` (0 = unlimited) in `pkg/gofr/metrics/exporters/exporter.go`. `WithCardinalityLimit` is a stable public option as of v1.44 (the older `OTEL_GO_X_CARDINALITY_LIMIT` env var is deprecated and should not be relied on).

Beyond that, exposing this as user-facing configuration (raise / disable / set a finite cap for overflow protection) is worth discussing — exact shape of that config is open. Separately, the raw-URL-as-label behavior in the static-file path is worth revisiting as it amplifies this issue.

### References
- OTel Go 1.44 changelog — cardinality limit: https://github.com/open-telemetry/opentelemetry-go/pull/8247
- Surfaced while reviewing #3564

Umang01-hash:
## Summary

Follow-up to #3564 (which bumps OpenTelemetry to 1.44). Resolves #3566.

OTel Go 1.44 applies a **default per-instrument cardinality limit of 2000** attribute sets — past that, new series are silently aggregated into a single `otel.metric.overflow="true"` series. GoFr previously inherited unlimited cardinality and exposed **no way to configure** the limit, so the bump would silently regress high-cardinality apps. This PR makes the limit explicit and configurable, and removes GoFr's own unbounded label source.

## 🔧 New configuration: `METRICS_CARDINALITY_LIMIT`

| Config | Description | Default |
|---|---|---|
| `METRICS_CARDINALITY_LIMIT` | Max distinct attribute sets recorded per metric instrument per collection cycle. Beyond it, extra series collapse into one `otel.metric.overflow="true"` series. | `2000` |

- **Default `2000`** — matches the OTel SDK default (no surprise vs upstream).
- **`0` or negative → unlimited** — restores pre-1.44 GoFr behavior.
- **Any `N`** → raise/lower the cap.

Documented in the [configuration reference](https://gofr.dev/docs/references/configs) (App table) and the [observability guide](https://gofr.dev/docs/quick-start/observability).

## Changes

**1. Configurable cardinality limit**
Wired via a new `exporters.WithCardinalityLimit` option → `metricSdk.WithCardinalityLimit`, fed from `METRICS_CARDINALITY_LIMIT` (parsed in `Container.cardinalityLimit`, falling back to the default on an invalid value).

**2. Bound the built-in `app_http_response` `path` label**
The `path` label is now always the **matched route template**, never the raw request URL — so cardinality (and the internal `routeAttrs` cache) is bounded by the number of registered routes:
- templated route → its template (e.g. `/customer/{id}`)
- static files → `/static` (the `PathPrefix("/static/")` template)
- anything unmatched, including 404s and static-looking 404s → `/` (GoFr's catch-all `PathPrefix("/")`)

Previously these fell back to the raw `r.URL.Path`, which was unbounded.

**3. Docs & tests**
- Config documented (above). Observability note updated to describe the actual `path`-label behavior.
- Tests: real overflow guard (`TestPrometheus_CardinalityLimitApplied` — scrapes an isolated registry, fails if `WithCardinalityLimit` is removed), config parsing (`TestContainer_cardinalityLimit`), and an integration test over GoFr's real router topology (`TestMetrics_GoFrRouterTopology`).

## Behavioral change (intentional, documented)

Two label changes for existing dashboards, both to bound cardinality:
- Keeping OTel's `2000` default means high-cardinality instruments now overflow rather than grow unbounded. Set `METRICS_CARDINALITY_LIMIT=0` to restore unlimited.
- `app_http_response`'s `path` for static/unmatched requests is now `/static` / `/` instead of the raw URL.

## Notes for review
- Stacked on #3564 — base is `deps/minor-updates-2026-06-09`, so the diff shows only these changes and retargets to `development` automatically once #3564 merges. **Merge #3564 first.**
- An earlier revision used `/static/*` and `/*` sentinels for non-templated requests; those were dead code in a real GoFr app (the catch-all `PathPrefix("/")` means a template always resolves), so they were removed in favour of the route-template + catch-all behavior described above.


Umang01-hash:
Reviewed + verified locally (built, ran tests, read the OTel 1.44 SDK source, wrote throwaway tests for the edge cases). Core mechanism is correct — `WithCardinalityLimit(0)` genuinely means unlimited in the SDK, GoFr's option wins over the env-derived value, and bounding `routeAttrs` also fixes a real unbounded-map memory leak. Three things to fix:

**1. (blocking) Matched, templated routes ending in a "static" extension get mislabeled `/static/*`.** `isStaticAsset(r.URL.Path)` is the first switch case, so it runs before the matched route template is honored. Confirmed: `GET /openapi.json` on a registered route records `path=/static/*`, not `/openapi.json`. Any real API route ending in `.json/.pdf/.html/...` is bucketed as a static asset. Move the `isStaticAsset` check into the `path == ""` branch so it only applies when no route matched.

**2. `/static` prefix is overbroad.** `HasPrefix(urlPath, "/static")` (no trailing slash) collapses `/static-report` → `/static/*` (confirmed). Should be `/static/`.

**3. Test doesn't guard the wiring it claims to.** `TestPrometheus_CardinalityLimitOption` only asserts `meter != nil` — it'd still pass if the `metricSdk.WithCardinalityLimit(...)` line were deleted. Also no test for the `container.go` config parsing (valid / invalid→default fallback).

Minor: the `app_http_response` path-label change (`/static/example.js`→`/static/*`, raw 404s→`/*`) is a breaking change for existing dashboards but is only in the PR body/comments — worth a note in the docs too.

Umang01-hash:
Thanks for the careful review — all three addressed in f6cfc4f9.

**1 (blocking) — matched templated routes mislabeled `/static/*`.** Reordered the path resolution so a matched route template is honored first; the static/unmatched sentinels only apply when no template resolved. `GET /openapi.json` on a registered route now records `path=/openapi.json`. Guarded by `TestMetrics_MatchedRouteWithStaticExtension`. Also keep the root route as `/` instead of an empty label after trimming (`TestMetrics_RootRouteLabel`).

**2 — `/static` prefix overbroad.** Now requires `/static/` (trailing slash), so `/static-report` is no longer collapsed. Guarded by a `TestIsStaticAsset` table test.

One consequence worth flagging: since static files are served via `PathPrefix("/static/")` (a matched route with template `/static`), honoring the template means those now label as `/static` rather than the `/static/*` sentinel — both bounded. The `/static/*` sentinel now only applies to static-looking requests that match no route (covered by `TestMetrics_UnmatchedStaticAssetSentinel`). Updated the existing static-file tests accordingly.

**3 — tests didn't guard the wiring.** Replaced the `meter != nil` check with `TestPrometheus_CardinalityLimitApplied`: it records past a limit of 1 against an isolated registry and asserts the overflow series appears (and is absent when unlimited). Verified it fails if `metricSdk.WithCardinalityLimit(...)` is removed. Extracted `newPrometheusMeter` to inject the registerer for deterministic scraping. Added `TestContainer_cardinalityLimit` for the config parsing (valid / 0 / invalid→default).

**Minor — docs.** Added a note in the observability docs covering the `app_http_response` path-label change and the cardinality limit / overflow behavior.

Umang01-hash:
Re-reviewed the latest commits — rebuilt, re-ran tests/lint, and stood up a real GoFr server scraping `/metrics`. The three things I flagged are genuinely fixed and verified on a live server:

- `/openapi.json` registered route → `path="/openapi.json"` (no longer collapsed to `/static/*`) ✅
- `/static/` prefix tightened (trailing slash) + `TestIsStaticAsset` ✅
- `TestPrometheus_CardinalityLimitApplied` now scrapes a real registry and asserts the `otel.metric.overflow` series — real wiring guard, fails if the SDK line is removed ✅
- docs note added; build/vet/gofmt/golangci-lint clean. 👍

One new thing surfaced on deeper inspection that I think should be addressed before merge:

**`isStaticAsset()`, `staticAssetPath` (`/static/*`) and `unmatchedPath` (`/*`) are effectively dead code in a real GoFr app.** GoFr unconditionally registers a catch-all `router.PathPrefix("/")` (gofr.go:175) whose `GetPathTemplate()` returns `"/"`, so `path` is never empty → the `path != ""` branch always wins and the sentinel branches never execute. Verified on a running server:

- static files → `path="/static"` (via the `PathPrefix("/static/")` template, not `isStaticAsset`)
- static-looking 404s (`/images/missing.png`, `/app.js`) → `path="/"`, not `/static/*`
- invented 404s + method-mismatch (POST/PUT/DELETE) → `path="/"`, not `/*`

So the description's "collapse to bounded sentinels `/static/*` and `/*`" doesn't match real behavior (they collapse to `/static` and `/`), and `TestMetrics_UnmatchedRouteBoundedLabel` / `TestMetrics_UnmatchedStaticAssetSentinel` only pass because they build routers without GoFr's catch-all — they assert a path that never happens in the integrated framework.

Cardinality is still correctly bounded, so this isn't a correctness bug — it's dead code + a tests/description mismatch. Suggest either:
1. drop `isStaticAsset` + both sentinels and rely on the template + root special-case (already bounds everything), and fix the description; or
2. keep them deliberately as defense-in-depth, but say so and add an integration test that exercises the real GoFr router topology (with the catch-all) so the tests reflect reality.

Umang01-hash:
Good catch — you're right, the catch-all `PathPrefix("/")` (gofr.go) means a template always resolves, so `isStaticAsset` + the `/static/*` and `/*` sentinels were dead code. Went with option 1 in 323611ba.

- Removed `isStaticAsset`, `staticAssetPath`, `unmatchedPath`. The `path` label is now always the matched route template, trailing-slash trimmed, with an empty-template fallback to `/` (keeps the bound for any router without the catch-all).
- Verified on a running GoFr server scraping `/metrics`: `/hello` → `/hello`, `/redis` → `/redis`, and `/no/such/route` / `/images/missing.png` / `/app.js` all → `/`. No raw URLs, no `/static/*` or `/*`.
- Replaced the sentinel tests (which only passed without the catch-all) with `TestMetrics_NoTemplateFallsBackToRoot` and `TestMetrics_GoFrRouterTopology` — the latter builds the real topology (specific route + `/static/` prefix + catch-all) and asserts templates / `/static` / `/`.
- Fixed the description to match (templates, `/static`, `/` — not the removed sentinels) and made `METRICS_CARDINALITY_LIMIT` a prominent section. Observability docs note updated too.

`build`/`vet`/`golangci-lint` clean; changed packages green (the local `pkg/gofr` failure is just port 2121 being held by another process on my box — unrelated to this diff).

Umang01-hash:
Re-verified the latest commit (`drop dead static/unmatched sentinels`). Rebuilt, re-ran tests/vet/gofmt/golangci-lint, and stood up a real GoFr server scraping `/metrics`. The dead-code point is fully resolved and everything checks out.

**What changed:** `isStaticAsset()` + the `/static/*` and `/*` sentinels were removed; path resolution is now just template → trim trailing slash → fall back to `/`. The misleading isolated tests were dropped and replaced with `TestMetrics_GoFrRouterTopology`, which exercises the real router topology (specific route + `PathPrefix("/static/")` + catch-all `PathPrefix("/")`). The docs note was corrected to describe actual behavior (`/static` and `/`).

**Live server `/metrics` confirms bounded labels, no raw-URL leakage:**
- `/static/example.js` → `path="/static"`
- `/openapi.json` (registered route) → `path="/openapi.json"`
- `/customer/42`, `/customer/777` → `path="/customer/{id}"`
- static-looking 404s, invented 404s, method-mismatch → `path="/"`
- no `/*`, no `/static/*`, no per-file/per-URL series

**Checks:** `go build`, `go vet`, `gofmt`, `golangci-lint` all clean on changed files; affected tests pass. Earlier items remain verified: configurable `METRICS_CARDINALITY_LIMIT` (0/negative = unlimited, confirmed against the OTel SDK source), the real cardinality-wiring test (registry scrape + `otel.metric.overflow` assertion), config-parsing test with fallback, and docs in both pages. Also genuinely fixes the original unbounded-cardinality + unbounded `routeAttrs` cache growth.

LGTM from my side — no outstanding issues. Only operational note: this is stacked on #3564, so #3564 should merge first, after which this retargets to `development`.

aryanmehrotra:
## Review — end to end

**TL;DR**: Architecture is correct, tests are solid, one real bug to fix (negative limit values). The two Copilot inline comments are stale — both were addressed in later commits and don't apply to HEAD.

---

### What it does

1. **Cardinality limit config** — `METRICS_CARDINALITY_LIMIT` → parsed in `Container.cardinalityLimit()` → passed via `exporters.WithCardinalityLimit` → wired into `metricSdk.WithCardinalityLimit` on the MeterProvider. Functional options pattern, backward-compatible.

2. **Path label bounding** — `app_http_response` `path` label is now always the matched route template (e.g. `/customer/{id}`), never raw URL. Removes the old `isStaticAsset` / extension switch dead code. The `routeAttrs` sync.Map is now properly bounded.

---

### Bug — negative cardinality limit passed raw to OTel SDK

**`pkg/gofr/metrics/exporters/exporter.go`** / **`pkg/gofr/container/container.go`**

`cardinalityLimit()` returns raw negative ints (e.g. `METRICS_CARDINALITY_LIMIT=-1`), which flow straight into `metricSdk.WithCardinalityLimit(-1)`. The OTel SDK documents that **0 = unlimited** but makes no guarantee about negative values — undefined behavior, potentially a silent no-op or a crash depending on SDK version. The PR description says "0 or negative → unlimited" but only 0 is guaranteed.

**Fix** — normalize in `cardinalityLimit()`:
```go
if limit <= 0 {
    return 0  // OTel SDK: 0 = unlimited
}
return limit
```

---

### Stale Copilot comments

Both inline comments (`isStaticAsset` prefix issue and the "template checked after extension switch" issue) describe code that was removed in the commit `fix(metrics): drop dead static/unmatched sentinels`. The HEAD path-label block is just:

```go
if cr := mux.CurrentRoute(r); cr != nil {
    path, _ = cr.GetPathTemplate()
}
path = strings.TrimSuffix(path, "/")
if path == "" {
    path = "/"
}
```

No extension switch, no `isStaticAsset`, no `HasPrefix("/static")`. Those threads can be dismissed.

---

### Test nit (non-blocking)

`TestPrometheus_CardinalityLimitApplied` "unlimited" case records only 2 distinct attribute sets. This doesn't distinguish "0 = truly unlimited" from "0 = OTel default 2000" (since 2 < 2000, neither overflows). Not a blocker since the OTel SDK does document 0 = unlimited, but the assertion doesn't conclusively prove it.

---

### Everything else looks good

| Area | Assessment |
|---|---|
| `WithCardinalityLimit` option pattern | Correct; variadic `...Option` keeps `Prometheus()` backward-compatible |
| `newPrometheusMeter` extraction for test injection | Right pattern; lets `TestPrometheus_CardinalityLimitApplied` scrape an isolated registry |
| `DefaultCardinalityLimit = 2000` exported | Matches OTel SDK default; callsites and tests can reference it |
| Error fallback on invalid config value | Correct — logs, falls back to default |
| `path` label for graphql skip | Works — route template `/graphql` after `TrimSuffix` still `/graphql`, skip fires ✓ |
| `routeAttrs` sync.Map cardinality bound | Now bounded by `#routes × #methods`, never raw URLs ✓ |
| Behavioral change doc (`/static/file.js` → `/static`) | Documented in both observability guide and configs reference |
| `TestMetrics_GoFrRouterTopology` | Integration test covers real GoFr router shape — specific, static, catch-all |
| `TestContainer_cardinalityLimit` table | Covers unset/valid/zero/invalid cases cleanly |

---

One real change needed: normalize `<= 0` → `0` before it reaches `metricSdk.WithCardinalityLimit`. Everything else is approvable.

pkg/gofr/http/middleware/metrics.go
```diff
@@ -43,6 +43,34 @@ type routeMethodKey struct {
...
+	staticAssetPath = "/static/*"
+	unmatchedPath   = "/*"
+)
+
+// isStaticAsset reports whether urlPath is a static-asset request whose raw
+// path should be collapsed to staticAssetPath.
+func isStaticAsset(urlPath string) bool {
+	if strings.HasPrefix(urlPath, "/static") {
+		return true
+	}
```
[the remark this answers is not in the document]
Umang01-hash: Addressed — `isStaticAsset` and the string-prefix check were removed. The path label now comes from the matched route template (`mux.CurrentRoute().GetPathTemplate()`), so there is no `/static` prefix match that could collapse `/static-api`. Covered by `TestMetrics_GoFrRouterTopology`.

pkg/gofr/http/middleware/metrics.go
```diff
@@ -89,32 +117,27 @@ func Metrics(metrics metrics) func(inner http.Handler) http.Handler {
...
-			case ".css", ".js", ".png", ".jpg", ".jpeg", ".gif", ".ico", ".svg", ".txt", ".html", ".json", ".woff", ".woff2", ".ttf", ".eot", ".pdf":
-				path = r.URL.Path
+			switch {
+			case isStaticAsset(r.URL.Path):
+				path = staticAssetPath
+			case path == "":
+				path = unmatchedPath
+			default:
+				path = strings.TrimSuffix(path, "/")
 			}
```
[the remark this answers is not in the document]
Umang01-hash: Addressed — sentinels removed; the route template is used directly, then the trailing slash is trimmed and `""` mapped back to `/`, so the root route is `/` (not empty) and matched routes with static-looking paths keep their template. Verified live (`/`→`/`, `/hello`→`/hello`, `/static/*`→`/static`, unknown→`/`) and covered by `TestMetrics_RootRouteLabel` + `TestMetrics_MatchedRouteWithStaticExtension`.

Umang01-hash:
Rebased onto `development` and retargeted the base here — this PR was previously stacked on the 2026-06-09 dep-consolidation branch (#3564, now closed/superseded), so the force-push just drops that inherited commit. Diff is now the 8 metrics files only.

Both review threads addressed: `isStaticAsset` and the static-prefix sentinels are gone; the path label comes from the matched route template (trailing slash trimmed, `""`→`/`). Verified live on the http-server example — `/`→`/`, `/hello`→`/hello`, `/static/*`→`/static`, unknown URLs→`/` (bounded, no raw URLs). Build/vet/`-race` tests green.

pkg/gofr/container/container_test.go
```diff
@@ -753,3 +754,24 @@ func frameworkMetricContract() map[string]metricContract {
...
+func TestContainer_cardinalityLimit(t *testing.T) {
+	tests := []struct {
+		desc    string
+		configs map[string]string
+		want    int
+	}{
+		{"unset uses default", map[string]string{}, exporters.DefaultCardinalityLimit},
+		{"valid value", map[string]string{"METRICS_CARDINALITY_LIMIT": "5000"}, 5000},
+		{"zero means unlimited", map[string]string{"METRICS_CARDINALITY_LIMIT": "0"}, 0},
+		{"invalid falls back to default", map[string]string{"METRICS_CARDINALITY_LIMIT": "not-a-number"}, exporters.DefaultCardinalityLimit},
```
aryanmehrotra, line 768: The table tests `unset`, `valid`, `zero`, and `invalid` — but the docs and PR description both say "0 or negative → unlimited". Negative values pass straight through `strconv.Atoi` and land in OTel's `limiter.aggLimit`, where the `> 0` guard treats them identically to 0. Add one case to match the documented contract:

```go
{"negative means unlimited", map[string]string{"METRICS_CARDINALITY_LIMIT": "-1"}, -1},
```

docs/references/configs/page.md
```diff
@@ -60,6 +60,12 @@ This document lists all the configuration options supported by the GoFr framewor
 
 ---
 
+-  METRICS_CARDINALITY_LIMIT
+-  Maximum number of distinct attribute sets recorded per metric instrument in a collection cycle. Beyond this, new series overflow into a single `otel.metric.overflow=true` series (OpenTelemetry SDK default since v1.44). Set to **0** (or a negative value) to disable the limit (unlimited cardinality).
```
aryanmehrotra: OTel 1.44 also reads `OTEL_GO_X_CARDINALITY_LIMIT` as an env-level default, applied before GoFr's `WithCardinalityLimit` option — so GoFr's value wins, but users who have both set will see the OTel one silently ignored. One line here would save a support question:

```
Set to **0** (or a negative value) to disable the limit. Overrides `OTEL_GO_X_CARDINALITY_LIMIT` if both are set.
```

aryanmehrotra (commented):
Architecture is correct, tests are solid. Two small things before merge.

pkg/gofr/container/container_test.go
Umang01-hash: Added — `{"negative means unlimited", {"METRICS_CARDINALITY_LIMIT": "-1"}, -1}`. Confirmed it passes straight through `strconv.Atoi` and OTel treats ≤0 as unlimited (52e01f0c).

docs/references/configs/page.md
Umang01-hash: Added the override note to the config docs. Verified against `sdk/metric` v1.44 `config.go`: the MeterProvider seeds `cardinalityLimit` from `cardinalityLimitFromEnv()` (`OTEL_GO_X_CARDINALITY_LIMIT`), then `WithCardinalityLimit` overwrites it — so GoFr's value wins when both are set.

Umang01-hash:
Superseded by #4160.

This branch fell a long way behind `development`, which has since refactored the metrics exporter into a pluggable `Build`/provider architecture and — separately — already bounds the built-in path label in the middleware (`metricsPath` resolves via route template and marks unmatched/static paths non-cacheable). So the path-label half here is now redundant, and the still-valuable cardinality-limit half has been re-based onto the current architecture in #4160 (with env parsing + a wiring test that asserts overflow, and E2E against the custom-metrics example). Closing in favor of that.

PR #3568, opened 2026-06-09
fix(metrics): extend datasource latency histogram buckets to 3 minutes

aryanmehrotra:
## What

Datasource latency histograms capped their buckets at `10`. For the datasources that record duration in **microseconds** (clickhouse, cassandra, scylladb, surrealdb, dgraph, redis, …) that meant the largest bucket was 10µs, so essentially every real query fell into `+Inf` and the histogram carried no usable distribution. The `app_clickhouse_stats_bucket` example was effectively useless.

This standardizes **all** datasource latency histograms on one shared bucket shape spanning **50µs → 3 minutes**, expressed in each datasource's *native recorded unit*.

## Approach

Per the agreed direction: **recorded values are the source of truth and are left unchanged** (so nothing breaks on the data side); only **bucket boundaries** and **descriptions/names** change to match what is actually recorded.

| Unit recorded | Datasources | New bucket top |
|---|---|---|
| milliseconds | sql, nats, dynamodb, influxdb, opentsdb | `180000` (3 min) |
| microseconds | redis, clickhouse, cassandra, scylladb, surrealdb, dgraph, dbresolver, badger, solr, arangodb, mongo, couchbase, elasticsearch, file | `180000000` (3 min) |

SQL keeps using the shared `getDefaultDatasourceBuckets()` (ms); a parallel `getDefaultDatasourceMicrosecondBuckets()` was added for Redis (which records µs).

## Metadata corrections (records win, desc/name updated)

These metrics recorded microseconds but advertised milliseconds — descriptions corrected:
`app_redis_stats`, `app_badger_stats`, `app_solr_stats`, `app_arango_stats`, `app_mongo_stats`, `app_couchbase_stats`.

Other fixes:
- **elasticsearch**: metric renamed `es_request_duration_ms` → `es_request_duration_us` (it records µs).
- **file**: `app_file_stats` description now states the unit (microseconds).

## Compatibility

- No recorded metric **values** change → existing rate/avg queries are unaffected.
- Histogram `_bucket` boundaries change (the whole point), and the noted description/name changes update dashboard metadata. The elasticsearch metric **rename** is the only series-name change.

## Tests

Updated affected unit tests (container snapshot/bucket assertions, plus per-module bucket/description/name assertions). All touched modules build; their tests pass except two **pre-existing, unrelated** failures verified against baseline (`scylladb Test_Exec` and `arangodb` document mock-type tests) that are not touched by this change.

## Follow-up (not in this PR)

gRPC histograms (`app_gRPC-Server_stats`, `app_gRPC-Stream_stats`, `app_gRPC-Client_stats`) are **not** registered by the framework — they come from `// Code generated by gofr.dev/cli/gofr. DO NOT EDIT.` files produced by the **gofr-cli** generator. Increasing their buckets to 3 minutes needs a change in the `gofr.dev/cli/gofr` repo's template, not here.


pkg/gofr/datasource/elasticsearch/elasticsearch.go
```diff
@@ -96,9 +96,12 @@ func (c *Client) Connect() {
...
 	c.client = client
 
-	var responseTimeBuckets = []float64{0.05, 0.1, 0.2, 0.5, 1.0, 2.0, 5.0, 10.0}
+	var responseTimeBuckets = []float64{
+		50, 75, 100, 125, 150, 200, 300, 500, 750, 1000, 2000, 3000, 5000, 7500, 10000, // 50µs-10ms
+		25000, 50000, 100000, 250000, 500000, 1000000, 5000000, 10000000, 30000000, 60000000, 120000000, 180000000, // 25ms-3min
+	}
 
-	c.metrics.NewHistogram("es_request_duration_ms", "Duration of Elasticsearch requests in ms", responseTimeBuckets...)
+	c.metrics.NewHistogram("es_request_duration_us", "Duration of Elasticsearch requests in microseconds", responseTimeBuckets...)
```
Umang01-hash: This renames the emitted metric `es_request_duration_ms` → `es_request_duration_us`, not just its buckets. Unlike the other datasources in this PR (which keep their series name and only shift the `le` boundaries), this changes the metric's identity, so existing dashboards/alerts on the old name break silently. The new name is correct (it always recorded µs), but worth either keeping the old name as an alias or calling it out as a breaking change in the release notes.

Umang01-hash (commented):
Verified end-to-end: every datasource's bucket group now matches its actual recording unit (µs vs ms), the lists are consistent within each group, the container/helper tests are fail-on-revert, and `app_redis_stats` shows the new microsecond buckets (le=50 … 1.8e+08) at /metrics. Core change looks correct.

One thing to decide before merge — see inline on the Elasticsearch rename. Everything else is optional polish.

Umang01-hash (approved):

Issue #2113, opened 2025-08-01
Separate out datasources to a new repo


Umang01-hash:
## Current Problem

In the current design of the GoFr core, interfaces for all supported datasources—like SQL, Redis, MongoDB, Cassandra, ArangoDB, Solr, and others—are defined directly in the main repository. 

Most of the datasources have already been moved into separate modules, but **SQL and Redis are still tightly integrated into the core**. As the list of supported connectors continues to grow, this design becomes harder to manage.

Keeping everything in the core has a few key downsides:

- It clutters the core codebase with implementations many users don’t need.

- It increases the maintenance burden, especially when updating or adding connectors.

- It makes the framework harder to navigate and contributes to unnecessary coupling.

Simply put, the more we pack into the core, the harder it becomes to keep it clean, modular, and easy to evolve.

Umang01-hash:
### Proposed Solution:
We propose extracting all datasource implementations—including SQL and Redis—into their own self-contained Go modules.

Here’s the idea:

- Each connector lives in its own module under gofr.dev/datasources/<name>. (We need to create a new repo datasources that will contain all the supported datasources for GoFr).

- The GoFr core will only define the interfaces (e.g., SQL, Redis)—no implementation details.

- A new registry system will handle loading and wiring these connectors into the app during startup.

This allows developers to plug in only the connectors they actually need. The best part? From the developer's perspective, nothing changes—they can still access ctx.SQL, ctx.Redis, etc., just like before.

This change will make GoFr’s architecture cleaner, easier to maintain, and better suited for long-term growth.

Umang01-hash:
> Note: I’m happy to discuss alternative approaches or improvements to this design. The goal is to find a solution that makes GoFr more maintainable and scalable for everyone.

ccoVeille:
I'm unsure there is a need for a new repository. At least a database module is required.

I'm not against the logic of having another repository.

But from my perspective, it means:
- maintenance wouldn't be as easy as now as all changes that implies a breaking one, even a small one would become a burden to do: tag version, maybe a major, release

- decoupling things is good, but it also means you can release something that could break a feature in the datasource repository.

- having a community on 2 repositories is not that easy, the followers watchers won't be the same, things might be reviewed differently

For me having a mono repository is better as everything is kept under control.

But I can be wrong, and I won't see a problem if the solution of having 2 repositories is chosen

ccoVeille:
[...]

PR #3569, opened 2026-06-11
feat(cloudsql): Cloud SQL datasource with IAM auth + pluggable AddSQLDB

akshat-kumar-singhal:
**Description:**

Adds a **Cloud SQL** (Postgres/MySQL) datasource that authenticates with **GCP IAM database authentication** in the cloud and falls back to standard username/password locally — selected by the `DB_IAM_AUTH` config, with **no application code change**:

```go
app := gofr.New()
app.AddSQLDB(cloudsql.New(app.Config)) // same line works locally and on GCP
app.Run()
```

With IAM auth on, there is no static DB password and no Cloud SQL Auth Proxy sidecar; credentials resolve via Application Default Credentials (supporting Workload Identity Federation). The motivation is to let GoFr apps use short-lived, identity-based DB credentials on GCP without bespoke wiring.

**Design — minimal core seam, cloud SDK kept out of core:**

- `sql.NewSQLFromDB(db *sql.DB, conf, logger, metrics)` wraps an already-open `*sql.DB` in the standard SQL datasource, sharing the post-open connect/retry/metrics path with `NewSQL` (extracted into a `finalizeConnection` helper). No SQL behavior is duplicated.
- `App.AddSQLDB(db container.DB)` plugs a SQL datasource into the app, mirroring the existing `AddMongo`/`AddInfluxDB` pattern.
- The `cloudsql` datasource lives in its **own published module** (own `go.mod`) so only apps that import it pull `cloudsqlconn` and the GCP SDK — core gains no new dependencies. When IAM is off, `cloudsql` delegates straight to `gofrSQL.NewSQL`, so `ctx.SQL`, logging, metrics, health checks and transactions behave identically.

`NewSQLFromDB` is a provider-agnostic seam: AWS RDS/Aurora IAM and Azure Entra ID modules can be added the same way (a refreshing `driver.Connector` that mints a short-lived token per connection), each in its own module, with no further core change. `doc.go` documents this extension path.

None. All additions are new exported functions/methods; existing SQL behavior and the env-var connection path are unchanged.

Note for maintainers: this PR deliberately keeps the SQL drivers in core (non-breaking). Fully extracting the SQL datasource into its own module to shed `modernc.org/sqlite`/`lib/pq`/etc. from core is a separate **breaking** change — core imports the sql package and the sql package returns core's concrete `datasource.Health`, so carving it out while core still imports it forms a module cycle. That belongs with a major-version de-bloat, not here.

**Additional Information:**

- New module dependency (only for importers of `cloudsql`): `cloud.google.com/go/cloudsqlconn`.
- Runnable example under `examples/using-cloudsql`, docs page at `docs/datasources/cloudsql`, and a package `doc.go`.
- Table-driven tests cover `NewSQLFromDB`, `AddSQLDB`, and the `cloudsql` module (config parsing, DSN building, IAM validation, standard-SQL fallback).

**Checklist:**


aryanmehrotra (commented):
Took this through a full pass — built it, ran the tests, and traced the connection lifecycle end to end. I really like the shape here: the `AddSQLDB` seam plus a separate module is exactly the pluggable direction we landed on in #3565, and `NewSQLFromDB` as a provider-agnostic seam is clean. Tests are green and the new module lints clean.

Two things I'd want sorted before merge though, both around the lifecycle: a nil-pointer panic if an IAM connect fails, and a connection + goroutine leak that falls out of core opening SQL eagerly. Rest are minors/nits. Details inline. Also +1 to Copilot's note on the Postgres-only `$1`/`SERIAL` in the example — worth either scoping it to Postgres or making it dialect-aware since the docs promise MySQL too.

pkg/gofr/external_db.go
```diff
@@ -190,6 +190,16 @@ func (a *App) AddInfluxDB(db container.InfluxDB) {
...
 }
 
+// AddSQLDB sets a pluggable SQL datasource on the app's container, replacing any
+// SQL connection configured from environment variables. The provided db must
+// satisfy the container.DB interface; official implementations that build on the
+// standard SQL wrapper (for example GCP Cloud SQL with IAM auth) live in published
+// modules such as gofr.dev/pkg/gofr/datasource/cloudsql.
+func (a *App) AddSQLDB(db container.DB) {
+	a.instrumentDatasource(db)
+	a.container.SQL = db
```
aryanmehrotra: One thing that bit me while tracing this: `container.Create()` already does `c.SQL = sql.NewSQL(...)` eagerly, so by the time a user calls `AddSQLDB` there's usually a live pool plus the `retryConnection`/`pushDBMetrics` goroutines already running — and their `stopSignal` only closes on `Close()`. Overwriting `a.container.SQL` here just drops that old DB on the floor: the goroutines keep pinging forever and the pool never gets returned.

It's actually worse on the non-IAM path, because `cloudsql.Connect()` calls `NewSQL` a *second* time from the same env, so we open two pools and leak the first. SQL is the only datasource field `Create()` pre-populates, which is why this one needs teardown where the other `Add*` siblings don't.

Could we close the existing one before swapping?
```go
func (a *App) AddSQLDB(db container.DB) {
    if old := a.container.SQL; !isNil(old) {
        _ = old.Close()
    }
    a.instrumentDatasource(db)
    a.container.SQL = db
}
```
That also lines up nicely with the registry direction in #2113.

pkg/gofr/datasource/cloudsql/cloudsql.go
```diff
@@ -0,0 +1,215 @@
...
+// application with App.AddSQLDB, which injects the logger and metrics and then
+// calls Connect. The embedded *gofrSQL.DB — GoFr's standard SQL datasource — is
+// populated by Connect and promotes the full container.DB interface, so the
+// datasource is used exactly like any other GoFr SQL connection.
+//
+// It is unexported on purpose: callers only ever hold the container.DB returned by
+// New and hand it to App.AddSQLDB, which drives the UseLogger/UseMetrics/Connect
+// lifecycle. Keeping the type private keeps the module's public surface to just New.
+type client struct {
+	*gofrSQL.DB // GoFr's standard SQL wrapper; promotes the container.DB methods
```
aryanmehrotra: Heads up on this embedding — it has a sharp edge on the failure path. If `connectIAM` fails for any reason (unsupported dialect, missing instance name, an ADC/GCP error, or `sql.Open` erroring), we log and return with `c.DB` still nil, but the client is already installed as `container.SQL`. Since the `*client` pointer itself is non-nil, `container.isNil(c.SQL)` is false, so the next health probe calls the promoted `HealthCheck()` on a nil `*gofrSQL.DB` receiver. `sql/health.go` touches `d.config.HostName` *before* the `d.DB == nil` guard, so it panics rather than reporting down.

I reproduced it — the first `/.well-known/health` after a failed connect takes the process down. Could we add a `HealthCheck()` on `*client` that returns `StatusDown` when `c.DB == nil`? Same spirit as the nil guard you already have in `Close()`.

pkg/gofr/datasource/cloudsql/cloudsql.go
```diff
@@ -0,0 +1,215 @@
...
+
+	if s.instanceConnectionName == "" {
+		c.logger.Errorf("cloudsql: DB_HOST (instance connection name 'project:region:instance') is required for IAM auth")
+		return
+	}
+
+	c.logger.Debugf("cloudsql: connecting to %s instance %q via IAM auth (ipType=%s)",
+		s.dialect, s.instanceConnectionName, s.ipType)
+
+	driverName := fmt.Sprintf("cloudsql-%s-%d", s.dialect, driverSeq.Add(1))
```
aryanmehrotra: Minor, more of a note than a blocker: every `Connect()` registers two process-global `database/sql` drivers (the connector here + the otelsql wrapper below) and there's no unregister in `database/sql`. For the normal one-shot lifecycle that's completely fine. It only matters if `Connect` ever runs repeatedly — test loops, or a future reconnect — since otelsql caps at 1000 slots and silently falls back to untraced after that. A `sync.Once` around the registration, or even just a doc line saying `Connect` is one-shot, would put it to bed.

pkg/gofr/datasource/cloudsql/settings.go
```diff
@@ -0,0 +1,147 @@
...
+	default:
+		return ipTypePublic
+	}
+}
+
+func intOrZero(s string) int {
+	n := 0
+	// Ignore parse errors: an unset or invalid value falls back to the standard
+	// SQL datasource defaults, matching getDBConfig in the core SQL package.
+	_, _ = fmt.Sscanf(strings.TrimSpace(s), "%d", &n)
```
aryanmehrotra: Tiny consistency nit: the core SQL config reads these with `strconv.Atoi`, whereas `fmt.Sscanf` is a bit looser (it'll happily parse `"5abc"` as 5). Might as well use `strconv.Atoi` here so both paths behave identically.

pkg/gofr/datasource/sql/sql.go
```diff
@@ -48,6 +48,26 @@ type DBConfig struct {
...
 
+// redactedPassword is the fixed mask substituted for a non-empty password whenever
+// a DBConfig is stringified, so the raw secret is never printed.
+const redactedPassword = "*****"
+
+// String implements fmt.Stringer so that logging a DBConfig — directly or as part
+// of a larger struct, with %v or %+v — never leaks the raw password. The Password
+// field is redacted to a fixed mask while every other field is preserved, keeping
+// the output useful for diagnosing connection issues.
+func (c DBConfig) String() string {
```
aryanmehrotra, line 60: Nice catch redacting this. One small gap: `String()` covers `%v`, `%s` and `%+v`, but `%#v` bypasses `Stringer` and would print the raw password. Nothing uses `%#v` on a `DBConfig` today, so it's not urgent — but it's an easy footgun for the next person. A `GoString()` that redacts the same way (or just a test asserting `%#v` stays clean) would close the loop.

examples/using-cloudsql/configs/.env
```diff
@@ -0,0 +1,23 @@
...
+
+# IAM auth on the cloud: no password needed, credentials come from Application
+# Default Credentials (supports Workload Identity Federation).
+DB_IAM_AUTH=true
+
+# PUBLIC | PRIVATE | PSC
+DB_CLOUDSQL_IP_TYPE=PUBLIC
+
+# For local development against built-in auth instead, set:
+#   DB_IAM_AUTH=false
```
aryanmehrotra, line 21: For the local-dev instructions — flipping `DB_IAM_AUTH=false` on its own isn't quite enough. You'd also need to repoint `DB_HOST`, which is still `my-project:us-central1:my-instance` up top; the plain pq/mysql driver will try to resolve that as a hostname and fail. Maybe add a `# DB_HOST=localhost` line to this block so the fallback path actually runs?

Umang01-hash:
Reviewed end-to-end — built both modules, ran the unit tests, and ran a real GoFr server on the non-IAM path (sqlite) scraping behaviour + logs. Architecture is clean: the `NewSQLFromDB` seam reuses the standard SQL wrapper, core gains zero new deps, and the non-IAM path works (queries via `ctx.SQL`, health 200, query logging + trace IDs, standard structured logs). Two things to fix:

**1. (blocking) Double SQL connection + orphaned connection/goroutine leak.** Running the example, the DB connects twice. `gofr.New()` already opens a SQL connection from the env vars (`container.go` calls `sql.NewSQL` unconditionally — it only skips when `DB_DIALECT` is empty, and cloudsql requires `DB_DIALECT`). Then `AddSQLDB` just does:
```go
a.instrumentDatasource(db)
a.container.SQL = db   // replaces, never closes the previous connection
```
So connection #1 is orphaned with its pool + `retryConnection`/`pushDBMetrics` goroutines still running.
- non-IAM: a leaked second connection every startup.
- IAM (the real target): core's #1 tries to dial `DB_HOST="project:region:instance"` as a literal host → fails → retries every ~10s forever, spamming error logs + leaking a goroutine. This contradicts the "drop-in, logging/metrics behave identically" claim.

Fix: `AddSQLDB` should `Close()` the existing `a.container.SQL` (if non-nil) before replacing it.

**2. IAM path is 0% tested** — the module's core feature. `connectIAM` success, `register` (pgx + mysql), the otelsql wrap/fallback, and the `sql.Open`-error cleanup are all uncovered (needs GCP). At least add a fake `driver.Connector` test driving `NewSQLFromDB` + `Close` with a non-nil cleanup to cover `runCleanup`/`errors.Join` in `Close`.

Minor:
- `DBConfig.String()` redaction guards a `%v`/`%+v` log path that doesn't currently exist anywhere — `DB_PASSWORD` isn't leaked today, so it changes nothing now (fine as future-proofing).
- DSN builders interpolate `DB_USER/DB_NAME/DB_HOST` unescaped via `fmt.Sprintf` — config-trust only and pre-exists in core, low risk; consider quoting.
- `sslmode=disable` is correct (connector owns mTLS).
- Run `govulncheck` on the cloudsqlconn dependency tree before merge.
- Release-ordering note: the module pins `gofr.dev v1.56.6` but uses `NewSQLFromDB`/`AddSQLDB` (added in this PR, not in any release yet) with no `replace`, so it only builds under go.work today — release core first, then re-pin + tag the leaf module.

pkg/gofr/external_db.go
```diff
@@ -190,6 +190,16 @@ func (a *App) AddInfluxDB(db container.InfluxDB) {
...
+}
```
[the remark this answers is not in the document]
akshat-kumar-singhal: Fixed in fed7c772. \`AddSQLDB\` now closes the existing \`a.container.SQL\` (typed-nil-safe) before swapping, so the old pool and its retry/metrics goroutines are torn down. Covered by \`TestApp_AddSQLDB_ClosesExisting\` and \`_NilExisting\`.

examples/using-cloudsql/main.go
```diff
@@ -0,0 +1,66 @@
...
+func addCustomer(ctx *gofr.Context) (any, error) {
+	var c customer
+	if err := ctx.Bind(&c); err != nil {
+		return nil, err
+	}
+
+	_, err := ctx.SQL.ExecContext(ctx, "INSERT INTO customers (name) VALUES ($1)", c.Name)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
akshat-kumar-singhal: Fixed in fed7c772 — the example is now explicitly scoped to Postgres, with an inline comment that MySQL needs \`?\` instead of \`$1\`. The \`cloudsql\` datasource itself stays dual-dialect.

examples/using-cloudsql/README.md
```diff
@@ -0,0 +1,70 @@
...
+   `gcloud auth application-default login`.
+
+## Schema
+
+```sql
+CREATE TABLE customers (
+    id   SERIAL PRIMARY KEY,
+    name TEXT NOT NULL
+);
+```
```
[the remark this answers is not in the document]
akshat-kumar-singhal: Fixed in fed7c772 — the README now states the example targets Postgres (\`SERIAL\`/\`$1\`) and notes the \`AUTO_INCREMENT\`/\`?\` changes needed for MySQL.

pkg/gofr/external_db.go
akshat-kumar-singhal: Good catch — fixed in fed7c772, essentially the snippet you proposed. \`AddSQLDB\` now closes the existing \`a.container.SQL\` before swapping, guarded by a typed-nil check (\`container.Create\` can leave a typed-nil \`*sql.DB\` when the dialect is unset, so a plain \`!= nil\` would still try to Close it). Close errors are logged rather than swallowed. New tests \`TestApp_AddSQLDB_ClosesExisting\` / \`_NilExisting\` cover both paths.

pkg/gofr/datasource/cloudsql/cloudsql.go
akshat-kumar-singhal: Confirmed and fixed in fed7c772. Added a \`HealthCheck()\` on \`*client\` that returns \`StatusDown\` (with an empty Details map) when \`c.DB == nil\`, and otherwise delegates to the embedded \`*gofrSQL.DB\` — same spirit as the existing nil guard in \`Close()\`. So a failed IAM connect now reports down instead of panicking on the first health probe. Covered by \`TestClient_HealthCheck\` (nil-guard + delegation branches).

pkg/gofr/datasource/cloudsql/cloudsql.go
akshat-kumar-singhal: Addressed as a doc note in fed7c772 rather than a \`sync.Once\`: \`driverSeq\` already guarantees unique driver names, so the registration is collision-free across clients — the real constraint is that a single client is one-shot. Documented that on \`connectIAM\` (registers a process-global driver + otelsql wrapper under a unique name; not meant to be reconnected in a loop, otelsql caps at 1000). Happy to add a \`sync.Once\`-gated guard later if we ever grow a reconnect path.

pkg/gofr/datasource/cloudsql/settings.go
akshat-kumar-singhal: Fixed in fed7c772 — \`intOrZero\` now uses \`strconv.Atoi\` so \`"5abc"\` is rejected rather than parsed as 5, matching the core config reader.

pkg/gofr/datasource/sql/sql.go
akshat-kumar-singhal: Done in fed7c772 — added \`GoString()\` (\`fmt.GoStringer\`) mirroring \`String()\`'s redaction, so \`%#v\` no longer prints the raw password. Added \`TestDBConfig_GoString_RedactsPassword\` asserting it stays clean directly and as a nested struct field.

examples/using-cloudsql/configs/.env
akshat-kumar-singhal: Fixed in fed7c772 — the local-dev block in \`.env\` now also repoints \`# DB_HOST=localhost\` (with \`# DB_PORT=5432\`), so flipping \`DB_IAM_AUTH=false\` actually runs the standard pq/mysql path instead of trying to resolve the instance-connection-name as a hostname.

akshat-kumar-singhal:
Thanks for the thorough passes @aryanmehrotra @Umang01-hash — all review points addressed in fed7c772. Summary:

**Blocking**
- **Double-connect / orphaned pool + goroutine leak:** \`AddSQLDB\` now closes the existing (env-configured) \`container.SQL\` before swapping, typed-nil-safe. Both the non-IAM double-open and the IAM "dial project:region:instance forever" cases are gone. (\`TestApp_AddSQLDB_ClosesExisting\` / \`_NilExisting\`.)
- **Nil-DB health panic after a failed IAM connect:** added \`HealthCheck()\` on \`*client\` returning \`StatusDown\` when \`c.DB == nil\`. (\`TestClient_HealthCheck\`.)
- **IAM path coverage:** added a fake \`driver.Connector\` driving \`NewSQLFromDB\` + \`Close\`, exercising \`runCleanup\`/\`errors.Join\` and the nil/non-nil health branches. Module coverage is now ~71%. (The real connector dial still needs GCP, so that part stays uncovered.)

**Minor**
- DSN builders now escape values: \`postgresDSN\` uses libpq single-quote quoting, \`mysqlDSN\` uses \`mysql.Config.FormatDSN\` (promoted \`go-sql-driver/mysql\` to a direct dep — one-line go.mod change, go.sum untouched). \`TestSettings_DSN_Escaping\` proves a crafted \`DB_NAME\` can't inject \`sslmode=require\`.
- \`intOrZero\` → \`strconv.Atoi\`; \`GoString()\` closes the \`%#v\` redaction gap; \`Connect\` documented as one-shot; example scoped to Postgres; \`.env\` local-dev block repoints \`DB_HOST\`.

**Notes**
- **govulncheck:** the only callable findings are stdlib (toolchain go1.26.2, fixed in go1.26.3/.4) — nothing in the \`cloudsqlconn\` tree. That's a repo-wide Go bump, out of scope here.
- **Release ordering:** acknowledged — the module can only build under \`go.work\` until core ships \`NewSQLFromDB\`/\`AddSQLDB\`; the \`gofr.dev\` re-pin + leaf-module tag is a post-merge step.

pkg/gofr/datasource/cloudsql/go.mod
```diff
@@ -0,0 +1,65 @@
+module gofr.dev/pkg/gofr/datasource/cloudsql
+
+go 1.26.0
+
+require (
+	cloud.google.com/go/cloudsqlconn v1.17.2
+	github.com/XSAM/otelsql v0.42.0
+	github.com/go-sql-driver/mysql v1.10.0
+	github.com/stretchr/testify v1.11.1
+	gofr.dev v1.56.6
```
aryanmehrotra: **Blocker — `cloudsql` must not depend on `gofr.dev`.**

Every other datasource module in this repo (`clickhouse`, `arangodb`, `cassandra`, `scylladb`, `influxdb`, `mongo`, …) carries **zero** `gofr.dev` in its direct deps. Each defines its own local `Logger`/`Metrics` interfaces and never imports core. This module breaks that rule by pulling in four core packages (`config`, `container`, `datasource`, `datasource/sql`).

The root cause is that `client` embeds `*gofrSQL.DB` and `New()` returns `container.DB`. Because `container.DB.Begin()` returns `*gofrSQL.Tx`, satisfying that interface forces the import.

**Proposed fix — invert the dependency, let core do the wrapping:**

```go
// cloudsql (no gofr.dev imports — only database/sql/driver + GCP SDK)
type Connector struct { ... }
func New(conf Config) *Connector { ... }  // conf is a plain local struct
func (c *Connector) Connector() (driver.Connector, func() error, error) { ... }
```

```go
// core: App.AddSQLDB (or a new App.AddSQLConnector) takes the connector,
// calls sql.OpenDB + gofrSQL.NewSQLFromDB internally — same place it already
// calls UseLogger / UseMetrics / Connect for every other datasource.
```

The one-liner `app.AddSQLDB(cloudsql.New(cfg))` can stay as the user-facing API; only the wiring moves from the leaf module into core, matching the pattern every other datasource already uses.

pkg/gofr/external_db.go
```diff
@@ -190,6 +192,39 @@ func (a *App) AddInfluxDB(db container.InfluxDB) {
...
+		return true
+	}
+
+	val := reflect.ValueOf(i)
+
+	return val.Kind() == reflect.Pointer && val.IsNil()
+}
+
 // GetSQL returns the SQL datasource from the container.
 func (a *App) GetSQL() container.DB {
```
aryanmehrotra, line 270: Remove `GetSQL()`. No other datasource on `App` has a `Get*` counterpart — there is no `GetMongo()`, `GetClickhouse()`, `GetCassandra()`, etc. Callers who need the connection after wiring already hold the value they passed into `AddSQLDB`. Adding a getter only for SQL creates asymmetry and will invite the same question for every future datasource.

pkg/gofr/datasource/cloudsql/settings.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	return &gofrSQL.DBConfig{
+		Dialect:     s.dialect,
+		HostName:    s.instanceConnectionName,
+		Database:    s.database,
+		User:        s.user,
+		MaxIdleConn: maxIdle,
+		MaxOpenConn: s.maxOpenConn,
+	}
```
aryanmehrotra: `Port` is not set here because the Cloud SQL connector owns routing. But `printConnectionSuccessLog` unconditionally formats `'%s:%s'` for host + port, so an IAM connect prints:

```
connected to 'sa@proj.iam' user to 'app' database at 'proj:us-central1:inst:'
```

The trailing colon will confuse operators. Fix: either guard `printConnectionSuccessLog` to omit `:%s` when port is empty, or set `Port: "(IAM)"` here as a human-readable sentinel (it is never dialled on this path).

aryanmehrotra (changes requested):

akshat-kumar-singhal:
Thanks @aryanmehrotra — all three points addressed in 86ef7d33 (rebased/merged onto latest `development`, conflict resolved).

**1. (blocker) `cloudsql` no longer depends on `gofr.dev`.** Inverted the dependency exactly as proposed — the leaf builds only a `driver.Connector` and core does the wrapping:

- `cloudsql.Connector.Connect() (driver.Connector, func() error, error)` returns a `database/sql` connector that dials through the Cloud SQL connector (Postgres via pgx `stdlib.GetConnector` with a custom `DialFunc`; MySQL via `mysql.NewConnector` with a registered dial context), plus a cleanup that closes the dialer. When IAM auth isn't requested it returns a **nil** connector, signalling core to keep the standard env-configured SQL.
- Config is read via a local `Config` interface (`Get`/`GetOrDefault`) satisfied by `app.Config`, so `app.AddSQLDB(cloudsql.New(app.Config))` is unchanged. The module now imports only `database/sql/driver` + the GCP SDK + pgx/mysql — **zero `gofr.dev`** (dropped `config`, `container`, `datasource`, `datasource/sql`, and transitively otelsql/lib-pq/sqlite), matching every other datasource module.
- Core: `App.AddSQLDB(SQLConnector)` calls the new `sql.NewSQLFromConnector`, which `otelsql.OpenDB`s the connector and shares `NewSQL`'s post-open path (logging/metrics/health/retry). A nil connector keeps the env SQL (so the double-connect leak is gone by construction); a connector error tears the env SQL down and leaves SQL unset, so health reports down instead of panicking. The connector cleanup is threaded into `DB.Close` via `errors.Join`.

Side benefits: this also removes the process-global driver-registry / otelsql-1000-cap concern (Postgres registers nothing global now), and the release-ordering coupling — the leaf no longer pins `gofr.dev` at all.

**3. Trailing colon on the IAM path.** `printConnectionSuccessLog`/`printConnectionFailureLog` now omit `:port` when the port is empty (`NewSQLFromConnector` clears it, since the connector owns routing). IAM connects now log `... database at 'proj:region:inst'` with no dangling colon.

**2. `GetSQL()` — flagging before removing.** This one predates this PR: it was added in #2140 and is consumed by the `dbresolver` module (`pkg/gofr/datasource/dbresolver/factory.go` calls `p.app.GetSQL()` in two places to fetch the primary connection). Removing it would break dbresolver, so I've left it in place. Happy to migrate dbresolver to another mechanism in a follow-up if you'd prefer it gone, but that felt out of scope here. Let me know.

Coverage: `NewSQLFromConnector` (wrap, empty-port label, cleanup-on-Close via a fake connector), `AddSQLDB` (wrap / closes-existing / defers-on-nil-connector / closes-on-error), and the leaf `Connect` validation + DSN escaping paths. The real connector dial still needs GCP, so that part stays uncovered.

aryanmehrotra (approved):
All previous review items are addressed. The architectural refactor (leaf yields `driver.Connector`, core does the wrapping) is the right shape — matches every other datasource module, zero `gofr.dev` deps in the leaf, and the connection lifecycle (close-on-replace, cleanup-on-close) is correctly handled. The `sync.Once` deferral for the one-shot constraint is acceptable with no reconnect path today. LGTM.

Umang01-hash (approved):

PR #3570, opened 2026-06-12
feat(clickhouse): support connection-pool and dial configuration

NitinKumar004:
### Problem
The ClickHouse datasource's `Config` exposes only `Hosts/Username/Password/Database`, and `Connect()` calls `clickhouse.Open` with no pool options. So applications are stuck on clickhouse-go's default pool (`MaxIdleConns` 5, `MaxOpenConns` 10) with no way to raise it. A service that issues many concurrent queries then hits **`acquire conn timeout`** — every pooled connection is busy and new queries wait the full budget and fail — with no GoFr-level knob to fix it.

### Change
Add optional pool/dial fields to `clickhouse.Config`, passed straight through to clickhouse-go:

- `MaxOpenConns` — max open connections (default `MaxIdleConns + 5`)
- `MaxIdleConns` — max idle connections (default `5`)
- `DialTimeout` — connection dial timeout (default `30s`)
- `ConnMaxLifetime` — max connection reuse time (default `1h`)

**Fully backward compatible:** clickhouse-go's `setDefaults()` applies its default for any zero value, so a `Config` that doesn't set these behaves exactly as before. Existing apps need no changes.

Option building is extracted into a small `clickHouseOptions(config)` helper so it can be unit-tested without dialing a server.

### Usage
```go
app.AddClickhouse(clickhouse.New(clickhouse.Config{
    Hosts:    app.Config.Get("HOSTS"),
    Username: app.Config.Get("USERNAME"),
    Password: app.Config.Get("PASSWORD"),
    Database: app.Config.Get("DATABASE"),
    // optional pool tuning; omit for driver defaults
    MaxOpenConns: 20,
    MaxIdleConns: 5,
}))
```

### Tests
- `Test_ClickHouse_Options_PoolConfigPassedThrough` — pool fields + comma-separated hosts map onto `clickhouse.Options`.
- `Test_ClickHouse_Options_ZeroPoolConfigLeavesDefaultsToDriver` — unset fields pass through as zero (driver keeps applying its defaults).
- Existing connect/health/exec/select tests still pass. `clickHouseOptions` at 100% coverage.

### Docs
`docs/datasources/clickhouse/page.md` updated with the new optional fields, their defaults, and an example.


NitinKumar004:
Addressed the `Hosts` parsing comment in adb9ea4f. The comma-split now goes through a `parseHosts` helper that trims each entry and drops empties, so values like "host-a:9000, host-b:9000" or a trailing comma no longer produce space-padded/blank addresses in `Options.Addr`. Added table-driven coverage (`Test_ClickHouse_Options_HostsTrimmedAndEmptiesDropped`): spaces-after-comma, trailing comma, interior blanks, single host, empty string. Build + full clickhouse package tests + golangci-lint all green locally.

Umang01-hash (approved):

PR #3595, opened 2026-06-16
feat(mqtt): add span links for pub/sub tracing

SujanVulasala:
**Description:**

Instruments MQTT publish and subscribe with `SpanKind` and the `messaging.*` semantic-convention attributes, matching what Kafka, NATS, SQS and Google Pub/Sub already do, and moves the consume span to the point the message is handed to a caller.

The PR was originally written to also link the consumer span back to the producer across the broker. Review found that inert, and it has been removed rather than shipped — see below.

**Why there is no cross-broker span link**

Verified against a live mosquitto broker at `3386020`: published inside an active span, consumed via `Subscribe`, and the two spans landed in **separate traces** with `links=0`.

An MQTT 3.1.1 `PUBLISH` packet carries only a topic, a QoS, a retain flag and a payload. There is no header or attribute slot for W3C `traceparent`, and `paho.mqtt.golang v1.5.1` does not implement MQTT 5 user properties. `Publish` discarded the injected attributes, and `createMqttHandler` set `MetaData` to `{qos, retained, messageID}`, so the extract on the consumer side could never find anything. The unit tests passed only because they handed attributes straight from `startPublishSpan` into `startSubscribeSpan`, bypassing the wire — the one path that was severed.

Neither remaining option is acceptable in this PR: MQTT 5 user properties would mean a different client library, and a payload envelope would corrupt the message for every non-GoFr subscriber on the topic. So the carrier, the injection and the link are removed rather than left in place reporting "no link" for a reason no operator could tell apart from a broken exporter. `TestSubscribeSpan_HasNoProducerLink` pins the limitation so a future MQTT 5 change is deliberate.

**What this improves over `development`**

The consume span used to be created in the paho arrival callback on `context.Background()` and ended immediately. Measured against the broker, it landed in a **third orphan trace** belonging to neither the producer nor the consumer:

```
development                                    this PR
mqtt-publish     trace=d4d6…  kind=internal    mqtt-publish     trace=ea77…  kind=producer
producer-request trace=d4d6…                   producer-request trace=ea77…
mqtt-subscribe   trace=aa1b…  ← orphan         mqtt-subscribe   trace=6c96…  kind=consumer
consumer-request trace=71d1…                   consumer-request trace=6c96…  ← same trace
```

It now sits in the consuming request's trace as its child. That is the right shape for MQTT regardless: GoFr's `Subscribe` is pull-style, so the work belongs to the consuming request.

**Performance**

A GoFr app with no `TRACE_EXPORTER` still installs an SDK provider sampling `NeverSample`, so every publish reached the attribute construction and threw the result away. Attributes are now set only when the span records, and the constant span-kind option is hoisted:

| publish path, tracing off | ns/op | B/op | allocs/op |
|---|---|---|---|
| `development` | 222 | 144 | 2 |
| this PR, attributes built unconditionally | 375 (+69%) | 600 (+317%) | 7 (+250%) |
| **this PR as merged** | 261 | **144** | **2** |

Back to development's allocation profile exactly, with the full attribute set kept when tracing is on. One publish against the broker is ~7.9 µs at QoS 0, so this never showed in latency — only as garbage, at 127k msg/s. `BenchmarkStartPublishSpan` holds the numbers.

**Breaking Changes:**

`app_pubsub_subscribe_total_count` changes meaning. It counted message arrivals in the paho callback; it now counts `Subscribe` calls, pairing with the success counter the way Kafka, NATS and Google Pub/Sub already do.

**Verification**

- Live mosquitto (sbx): span kind, attributes, trace parentage, and metric pairing all asserted end to end.
- `go test ./pkg/gofr/datasource/pubsub/mqtt/` — pass; `golangci-lint` v2.12.2 — 0 issues; `typos` — clean.
- `go test -race` fails on this package with 3 races, and fails **identically on `development`** — a test-only race in `TestMQTT_SubscribeSuccess`, whose goroutine reads `client.subscriptions` without holding `m.mu`. Pre-existing, not addressed here; CI does not run `-race`.

**Checklist:**


pkg/gofr/datasource/pubsub/mqtt/mqtt.go
```diff
@@ -177,7 +194,7 @@ func (m *MQTT) Query(ctx context.Context, query string, args ...any) ([]byte, er
 }
 
 func (m *MQTT) Publish(ctx context.Context, topic string, message []byte) error {
-	_, span := otel.GetTracerProvider().Tracer("gofr").Start(ctx, "mqtt-publish")
+	ctx, span, _ := startPublishSpan(ctx, topic)
```
Umang01-hash: The injected attrs (containing `traceparent`) are discarded with `_`, and `m.Client.Publish` below only sends the payload — so the trace context never leaves the process. paho v1.5.1 (MQTT 3.1.1) has no user-property slot to carry it either. This is why no link is ever established.

pkg/gofr/datasource/pubsub/mqtt/helper.go
```diff
@@ -134,14 +133,9 @@ func (*MQTT) handleContextDone(queryCtx context.Context, topicName string, buffe
 
 	return buffer, collected, nil
 }
-func (m *MQTT) createMqttHandler(_ context.Context, topic string, msgs chan *pubsub.Message) mqtt.MessageHandler {
+func (*MQTT) createMqttHandler(_ context.Context, _ string, msgs chan *pubsub.Message) mqtt.MessageHandler {
```
Umang01-hash: MetaData only carries `{qos, retained, messageID}` — never `traceparent`. So `extractMessageAttrs(msg.MetaData)` in `Subscribe` always yields an invalid span context, and `startSubscribeSpan` produces no link/parent.

pkg/gofr/datasource/pubsub/mqtt/tracing_test.go
```diff
@@ -0,0 +1,191 @@
...
+	otel.SetTextMapPropagator(propagation.TraceContext{})
+
+	defer func() {
+		_ = tp.Shutdown(context.Background())
+	}()
+
+	_, producerSpan, attrs := startPublishSpan(context.Background(), "test-topic")
+	producerSpan.End()
+
+	_, subscribeSpan := startSubscribeSpan(context.Background(), "test-topic", attrs)
```
Umang01-hash: This passes `attrs` straight from `startPublishSpan` into `startSubscribeSpan`, bypassing `Publish`/broker/`createMqttHandler` — the path that's actually broken. The test stays green while the feature is inert. A test that publishes through the client and asserts the subscribe span links back would catch this.

pkg/gofr/datasource/pubsub/mqtt/tracing.go
```diff
@@ -0,0 +1,116 @@
...
+	parentCtx := ctx
+
+	var links []trace.Link
+
+	if len(msgAttrs) > 0 {
+		carrier := attributeCarrier(msgAttrs)
+		extractedCtx := otel.GetTextMapPropagator().Extract(ctx, carrier)
+
+		if spanCtx := trace.SpanContextFromContext(extractedCtx); spanCtx.IsValid() {
+			parentCtx = extractedCtx
```
Umang01-hash: Minor (once propagation works): async consumers should link to the producer, not be a child of it — setting both `parentCtx` and a link ties unbounded-latency consume spans into the producer's tree. Prefer link-only.

Umang01-hash (dismissed):
Reviewed at head `3386020`. The span-link feature doesn't work end-to-end — the producer's trace context is never carried to the consumer, so no link/parent is ever created at runtime.

I verified against a live mosquitto broker: published inside an active span, consumed via `Subscribe`, and the two spans landed in **separate traces** (`links=0`, `parentValid=false`) — the subscribe span is orphaned.

Root cause (both ends disconnected):
- `Publish` discards the injected attrs and `m.Client.Publish` sends only the payload. paho `v1.5.1` is MQTT 3.1.1 — no user properties to carry `traceparent` anyway.
- `createMqttHandler` sets `MetaData` to `{qos, retained, messageID}` only — never `traceparent` — so `extractMessageAttrs` → `startSubscribeSpan` always sees an invalid context.

The unit tests pass because `TestStartSubscribeSpan_WithLinks` hands the in-memory attrs straight from `startPublishSpan` into `startSubscribeSpan`, bypassing the wire — the exact path that's severed in production.

To make it real: carry `traceparent` across the broker (MQTT 5 user properties, or a documented payload envelope), attach it in `Publish`, copy it into `MetaData` on receive, switch the consumer to link-only (async messaging shouldn't be parent-child), and add a test that publishes through the client and asserts the link survives the round-trip.

Build/vet/lint/tests are clean and there's no breaking API change — the issue is purely that the feature is inert.

aryanmehrotra:
Picked this up, rebuilt the branch onto current `development`, and pushed to the same branch so your commit stays first in the history. @SujanVulasala — thanks for this, and apologies for the long wait.

Summary of what changed and why, since I removed the part the PR was named after.

### The branch itself

It had drifted to a 2020 merge-base, so GitHub was showing ~4,900 commits and several hundred files and calling it `CONFLICTING`. Rebuilt by cherry-picking the three real commits onto current `development`: now **4 commits, 6 files, `MERGEABLE`**, with `5b61b04` (yours) unchanged at the base.

### Why the span link is gone

Confirmed at `3386020` against a live mosquitto broker before touching anything — published inside an active span, consumed via `Subscribe`:

```
mqtt-publish     trace=d4d656…  links=0
producer-request trace=d4d656…
mqtt-subscribe   trace=aa1b17…  links=0   ← different trace
```

Both ends were disconnected. `Publish` discarded the injected attributes, and `createMqttHandler` set `MetaData` to `{qos, retained, messageID}`, so the extract could never find anything.

The root cause isn't a wiring mistake — it's the protocol. An **MQTT 3.1.1 `PUBLISH` packet carries only a topic, a QoS, a retain flag and a payload.** There is no header or attribute slot for W3C `traceparent`, and `paho.mqtt.golang v1.5.1` doesn't implement MQTT 5 user properties. That's the difference from Kafka/NATS (headers) and SQS/Google Pub/Sub (message attributes), which the pattern here was modelled on.

Neither remaining option belongs in this PR: MQTT 5 user properties mean a different client library, and a payload envelope would corrupt the message for every non-GoFr subscriber on the topic. So the carrier, the injection and the link are removed rather than left in place — a dead extract reports "no link" identically to a broker that dropped the context, and no operator could tell those apart. `TestSubscribeSpan_HasNoProducerLink` pins it so a future MQTT 5 change is deliberate.

Worth saying plainly: the unit tests here passed because `TestStartSubscribeSpan_WithLinks` handed attributes straight from `startPublishSpan` into `startSubscribeSpan`, bypassing the broker — the one path that was severed. That's an easy trap to fall into, and it's the reason the new tests assert trace parentage end to end instead.

### What survives, and does improve on `development`

The rest of the PR was real and is kept. `mqtt-publish` and `mqtt-subscribe` now carry `SpanKind` and the `messaging.*` semantic conventions, and the consume span starts in `Subscribe` rather than in the paho arrival callback on `context.Background()`. Measured against the broker:

| | `mqtt-subscribe` lands in | kind |
|---|---|---|
| `development` | a **third orphan trace**, belonging to neither producer nor consumer | internal |
| this PR | the **consuming request's** trace, as its child | consumer |

That's the right shape for MQTT regardless: GoFr's `Subscribe` is pull-style, so the work belongs to the consuming request.

### Performance

Building the attributes unconditionally cost every publish, including in apps that never configured tracing — GoFr still installs an SDK provider (`NeverSample`) when `TRACE_EXPORTER` is unset, so the attributes were constructed and discarded at the sampler:

| publish path, tracing off | ns/op | B/op | allocs/op |
|---|---|---|---|
| `development` | 222 | 144 | 2 |
| this PR, attributes built unconditionally | 375 (+69%) | 600 (+317%) | 7 (+250%) |
| **as pushed** | 261 | **144** | **2** |

Fixed by setting attributes only when the span records, plus hoisting the constant span-kind option. Back to `development`'s exact allocation profile, full attributes kept when tracing is on. One publish against the broker is ~7.9 µs at QoS 0, so this was never latency — it was garbage, ~58 MB/s at 127k msg/s. `BenchmarkStartPublishSpan` holds the numbers.

### One behaviour change

`app_pubsub_subscribe_total_count` counted message arrivals in the paho callback; it now counts `Subscribe` calls, pairing with the success counter the way Kafka, NATS and Google Pub/Sub already do. A 1:1 flow shows no difference, so measured on the asymmetric case:

| | `development` | this PR |
|---|---|---|
| 5 published, 2 consumed | total=5, success=2 | total=2, success=2 |
| `Subscribe` times out, nothing arrives | total=0, success=0 | total=1, success=0 |

### Verified

Live mosquitto via sbx (span kind, attributes, both trace parentages, no false link, metric pairing), exported API identical (167 symbols, zero diff), no `go.mod` change, coverage 72.5% → 73.0%, lint 0 issues, typos clean, CI green on Go 1.24/1.25/1.26.

### Two pre-existing issues found, deliberately not fixed here

Both reproduce identically on `development`, so they aren't this PR's to carry — flagging them so they aren't lost:

1. **`go test -race` fails on this package** (3 races). `TestMQTT_SubscribeSuccess` launches a goroutine that reads `client.subscriptions` without holding `m.mu`, while `Subscribe` writes it under the lock. CI never runs `-race`, so it's invisible there.
2. **`MetaData["qos"]` is a control character.** `string(msg.Qos())` converts a byte to a rune, so QoS 1 comes out as `"\x01"` rather than `"1"`. Present in `getHandler` too, which this PR doesn't touch — fixing one half would be worse than fixing neither.

Happy to open follow-ups for either. Needs a second maintainer's approval before merge, since I've written commits on it.

Umang01-hash (approved):
Verified locally at 2bd017e. Attributes and SpanKind match Kafka/NATS/SQS/Google exactly. Confirmed the IsRecording guard is safe here because the sampler is NeverSample / ParentBased(TraceIDRatioBased) — neither looks at attributes — so setting them after Start is fine and keeps the NeverSample path allocation-free. Also confirmed the reason there's no cross-broker link: paho v1.5.1 is MQTT 3.1.1 (no user properties) and createMqttHandler only ever puts qos/retained/messageID in MetaData, so there was never anywhere for traceparent to ride — the removal is correct and HasNoProducerLink pins it. Moving subscribe_total_count to per-Subscribe-call matches kafka. Tracing tests use a real SpanRecorder and are non-vacuous (flipping the subscribe SpanKind makes them fail).

Heads-up, pre-existing and NOT from this PR: the mqtt package can't run green under -race — TestMQTT_SubscribeSuccess hangs and TestMQTT_Query_SuccessCases has a data race. I reproduced both identically on the merge-base, so they're pre-existing test-only issues (a nil-channel-send race in the test goroutine, and concurrent gomock use), untouched by this change. Might be worth a separate issue.

PR #3599, opened 2026-06-17
feat: upgrade to MQTT 5.0 and implement native span links via User Properties

### LINKED ISSUE #3012: Add span links in MQTT

adityakrmishra:
Fixes #3012

### What does this PR do?
This PR resolves the MQTT OpenTelemetry tracing issue by adhering to the maintainers' feedback from previously closed PRs (#3187, #3101). Because MQTT 3.1.1 does not support native metadata, payload wrapping was rejected as it silently broke backwards compatibility for non-GoFr consumers. 

This implementation upgrades the underlying driver to MQTT 5.0 to utilize native User Properties for clean context propagation.

**Architectural Changes:**
1. **Driver Upgrade:** Migrated from `eclipse/paho.mqtt.golang` (v3.1.1) to `eclipse/paho.golang` (v5), utilizing `autopaho` for resilient connection management.
2. **Context Propagation:** Implemented `tracing.go` with a `userPropertyCarrier` that maps `paho.UserProperties` directly to the `propagation.TextMapCarrier` interface.
3. **Publishing:** Injects the W3C `traceparent` cleanly into the MQTT 5.0 User Properties without modifying the raw message payload.
4. **Subscribing:** Extracts the `traceparent` from incoming User Properties and correctly binds it using `trace.WithLinks()`, perfectly mirroring the existing Kafka/Google PubSub tracing patterns in the framework.
5. **Stability:** Added defensive URL parsing guards to prevent `autopaho` from panicking on unparseable broker endpoints.

### Note to Reviewers
All tests under `./pkg/gofr/datasource/pubsub/mqtt/...` pass successfully. The legacy driver has been completely replaced, and this non-breaking implementation ensures tracing works natively across polyglot microservice environments.


aryanmehrotra:
Thanks for this, @adityakrmishra — and genuinely, thanks for taking on something that turned out to be far more than the `good first issue` label promised. You were right to say so.

**The thing you set out to prove, you proved.** I ran this against a live mosquitto broker and the span link works — which is more than can be said for the two earlier attempts:

| live broker check | `development` | this PR |
|---|---|---|
| subscribe span carries a producer link | `links=0` | **`links=1`** |
| producer and consumer share a trace ID | separate traces | **same trace** |
| `MetaData["qos"]` | `"\x01"` (control char — a real bug) | **`"1"`** |

MQTT 5 User Properties are the right mechanism, the `userPropertyCarrier` is a clean implementation of `propagation.TextMapCarrier`, and the payload is untouched. That answers the question #3012 was actually asking, and I've recorded it on the issue so the next attempt starts from your result rather than re-deriving it.

I'm closing this anyway, and I want to be precise about why, because it isn't about the tracing code.

## How I tested it

CI is green on this branch, but that green isn't worth much: the branch's merge-base is `88d682d3f` ("Adding License", 2020), so CI compiled it against a tree 5,007 commits behind `development`. I rebuilt it — `development` + your 6 real commits, cherry-picked clean — and tested that.

- **A/B against live mosquitto**, same harness run against `development` and against the rebuilt branch.
- **A wire-level probe**: a fake broker that reads the CONNECT packet and answers an unrecognised protocol level with return code `0x01`, the way a 3.1.1-only broker does.

## Two blocking issues, and neither is a bug in your code

**1. MQTT 3.1.1 support disappears, with no way to opt out.**

```
development : CONNECT protocolName="MQTT" protocolLevel=4  -> broker accepts
this PR     : CONNECT protocolName="MQTT" protocolLevel=5  -> broker refuses (0x01),
                                                              autopaho then retries forever
```

`paho.golang@v0.23.0/paho/client.go:277` sets `ccp.ProtocolVersion = 5` unconditionally — there is no 3.1.1 path in the library. `paho.mqtt.golang` defaults to level 4 and can fall back to 3.1. So on upgrade, every GoFr user pointed at a 3.1.1-only broker loses their connection, in a minor release of a v1 module, with no config to turn it off and no test that could catch it.

Worth adding: `paho.golang` is v0.23.0 — pre-1.0, and its README opens with a breaking-change warning about changes between minor releases.

**2. Public API is removed.** This program compiles against `development` and does not compile against this PR:

```
m.IsConnected undefined (type *mqtt.MQTT has no field or method IsConnected)
m.IsConnectionOpen undefined
m.OptionsReader undefined
undefined: mqtt.MockClient
undefined: mqtt.MockToken
```

`MQTT` embedded `mqtt.Client`, so every paho method was promoted onto it. Dropping the embed removes them all. Your own diff to `container_test.go` — deleting `assert.NotNil(t, m.Client)` — is that break showing up inside the repo.

Both of these are decisions for the maintainers to make deliberately, not things to fix inside a span-links PR. That is also what @coolwednesday asked for on #3101: *"we should address this after upgrading MQTT"* — the upgrade as its own step.

## Three regressions I confirmed live

These are fixable, and I mention them so they don't get rebuilt the same way:

| | `development` | this PR |
|---|---|---|
| `Subscribe(ctx, "a/+/c")`, publish to `a/1/c` | delivers | **nothing arrives** |
| `Subscribe(ctx, "a/#")`, publish to `a/b/c` | delivers | **nothing arrives** |
| `Subscribe` after a `Query` on the same topic | delivers | **nothing arrives** |

The wildcards go through `topicMatch`, which returns `false` for anything but an exact string match (`// ... implementation omitted for brevity`). paho v3 did full MQTT §4.7 matching in its router, so this is functionality the driver used to provide for free. `Test_topicMatch` only asserts the exact-match case, which is why it passes.

The third is the query handler: `subscribeToTopicForQuery` appends to `m.queryHandlers` and the deferred cleanup only unsubscribes — the handler stays in the slice forever, runs ahead of normal subscriptions, and returns `true`, so it swallows the message.

## Also worth noting

- **Coverage 72.5% → 36.5%** (measured on both trees). `cm` is a concrete `*autopaho.ConnectionManager`, so the package can't be tested without a broker; the mocks were deleted and most of the remaining tests assert error returns against a dead `127.0.0.1`.
- `MetaData["messageID"]` is gone — available as `pub.PacketID` in v5.
- `paho.mqtt.golang` is still a direct dependency in `go.mod` but no longer used.
- `kafka_test.go` doesn't compile against current `development` (rebase casualty — #3502 landed a value `sync.RWMutex`, not a pointer).
- `publicBroker` reverts hivemq → emqx, an artifact of the stale base. Broker changes have been reverted here before (#1338 → #1339).

## What I'd suggest

Split it, and the hard part goes away:

1. **MQTT 5 support as its own PR**, opt-in — `development` keeps paho v3 as the default so nobody breaks, and MQTT 5 is selected by config. That's the "upgrading MQTT" step the maintainers asked for, and it's where the protocol and API decisions belong.
2. **Span links on top of it** — at that point it's essentially the `tracing.go` you already wrote, which I've verified works.

I've written all of this up on #3012 with the evidence, so whoever picks up step 1 starts from what you established rather than from scratch. Please do take it if you want it — you're further into this than anyone.

PR #3600, opened 2026-06-20
fix(http): handle os.Getwd() error in AddStaticFiles

### LINKED ISSUE #3322: fix: handle os.Getwd() error silently ignored in AddStaticFiles

NitinKumar004:
## Description

`AddStaticFiles` in `pkg/gofr/gofr.go` discarded the error returned by `os.Getwd()`:

```go
```

If `os.Getwd()` fails, `currentWorkingDir` becomes empty and the static file path is
silently resolved against the filesystem root, with no log to help the developer diagnose
why the endpoint didn't register correctly.

## Change

Capture the error, log it via the container logger, and return early — mirroring the
existing error handling for the `os.Stat` failure path a few lines below in the same
function.

## Testing

- `go build ./pkg/gofr/`
- `go vet ./pkg/gofr/`
- `go test ./pkg/gofr/ -run 'TestStaticHandler|TestStaticHandlerInvalidFilePath|TestNewSetsHTTPRegisteredWhenStaticDirExists'` — all pass
- `golangci-lint run ./pkg/gofr/` — no new findings in the changed region

Fixes #3322

### DIFF
--- a/pkg/gofr/gofr.go
+++ b/pkg/gofr/gofr.go
@@ -411,7 +411,14 @@ func (a *App) AddStaticFiles(endpoint, filePath string) {
 
 	// update file path based on current directory if it starts with ./
 	if strings.HasPrefix(filePath, "./") {
-		currentWorkingDir, _ := os.Getwd()
+		currentWorkingDir, err := os.Getwd()
+		if err != nil {
+			a.container.Logger.Errorf("error in registering '%s' static endpoint, "+
+				"failed to get current working directory: %v", endpoint, err)
+
+			return
+		}
+
 		filePath = filepath.Join(currentWorkingDir, filePath)
 	}
 
--- a/pkg/gofr/gofr.go
+++ b/pkg/gofr/gofr.go
@@ -405,6 +405,10 @@ func (a *App) AddStaticFiles(endpoint, filePath string) {
 
 	a.httpRegistered = true
 
+	// Normalize the endpoint up front so every error log below reports the same
+	// '/endpoint' form (previously the getwd and os.Stat logs disagreed).
+	endpoint = "/" + strings.TrimPrefix(endpoint, "/")
+
 	if !strings.HasPrefix(filePath, "./") && !filepath.IsAbs(filePath) {
 		filePath = "./" + filePath
 	}
@@ -421,8 +425,6 @@ func (a *App) AddStaticFiles(endpoint, filePath string) {
 
 		filePath = filepath.Join(currentWorkingDir, filePath)
 	}
-
-	endpoint = "/" + strings.TrimPrefix(endpoint, "/")
 
 	if _, err := os.Stat(filePath); err != nil {
 		a.container.Logger.Errorf("error in registering '%s' static endpoint, error: %v", endpoint, err)
--- a/pkg/gofr/gofr_test.go
+++ b/pkg/gofr/gofr_test.go
@@ -1299,6 +1299,37 @@ func TestStaticHandlerInvalidFilePath(t *testing.T) {
 	assert.Contains(t, logs, "error in registering '/gofrTest' static endpoint")
 }
 
+func TestStaticHandlerGetwdError(t *testing.T) {
+	testutil.NewServerConfigs(t)
+
+	// Switch into a temporary directory and remove it so that os.Getwd()
+	// inside AddStaticFiles fails when resolving the relative "./" path.
+	// t.Chdir restores the original working directory at test cleanup.
+	tmpDir := t.TempDir()
+
+	t.Chdir(tmpDir)
+
+	require.NoError(t, os.Remove(tmpDir))
+
+	// On some platforms (e.g. macOS) the kernel still resolves the path of an
+	// unlinked working directory, so os.Getwd() does not fail. Skip there since
+	// the error branch cannot be exercised.
+	if _, err := os.Getwd(); err == nil {
+		t.Skip("os.Getwd() does not fail for a removed working directory on this platform")
+	}
+
+	// The app (and its logger) must be created inside StderrOutputForFunc so the
+	// logger writes to the captured stderr rather than the real one.
+	logs := testutil.StderrOutputForFunc(func() {
+		app := New()
+		app.AddStaticFiles("gofrTest", "./somedir")
+	})
+
+	assert.Contains(t, logs, "failed to get current working directory")
+	// The endpoint is logged before its leading "/" is added, so it appears as 'gofrTest'.
+	assert.Contains(t, logs, "error in registering 'gofrTest' static endpoint")
+}
+
 func TestNewSetsHTTPRegisteredWhenStaticDirExists(t *testing.T) {
 	testutil.NewServerConfigs(t)
 
--- a/pkg/gofr/gofr_test.go
+++ b/pkg/gofr/gofr_test.go
@@ -1326,8 +1326,7 @@ func TestStaticHandlerGetwdError(t *testing.T) {
 	})
 
 	assert.Contains(t, logs, "failed to get current working directory")
-	// The endpoint is logged before its leading "/" is added, so it appears as 'gofrTest'.
-	assert.Contains(t, logs, "error in registering 'gofrTest' static endpoint")
+	assert.Contains(t, logs, "error in registering '/gofrTest' static endpoint")
 }
 
 func TestNewSetsHTTPRegisteredWhenStaticDirExists(t *testing.T) {


pkg/gofr/gofr_test.go
```diff
@@ -1234,6 +1234,37 @@ func TestStaticHandlerInvalidFilePath(t *testing.T) {
...
+	// On some platforms (e.g. macOS) the kernel still resolves the path of an
+	// unlinked working directory, so os.Getwd() does not fail. Skip there since
+	// the error branch cannot be exercised.
+	if _, err := os.Getwd(); err == nil {
+		t.Skip("os.Getwd() does not fail for a removed working directory on this platform")
+	}
+
+	// The app (and its logger) must be created inside StderrOutputForFunc so the
+	// logger writes to the captured stderr rather than the real one.
+	logs := testutil.StderrOutputForFunc(func() {
```
aryanmehrotra, line 1323: Wrong output stream — GoFr's logger writes to **stdout**, not stderr. With `StderrOutputForFunc` the captured string is always empty, so on Linux (where `os.Getwd()` does fail after `os.Remove`) both assertions run against `""` and the test fails. On macOS the test skips, hiding the bug.

The existing test immediately above this one (`TestStaticHandlerInvalidFilePath`) exercises the same `Errorf` call and uses `StdoutOutputForFunc` — follow that pattern:

```go
logs := testutil.StdoutOutputForFunc(func() {
    app := New()
    app.AddStaticFiles("gofrTest", "./somedir")
})
```

aryanmehrotra (dismissed):
Production code is correct — `Errorf` + `return` is the right pattern, consistent with the rest of `AddStaticFiles`. One bug in the test before this can merge.

NitinKumar004:
Thanks for the review! I double-checked this against the logger implementation, and I think the stream is actually reversed here.

GoFr's logger routes **ERROR**-level logs to `errorOut`, which is `os.Stderr` — not stdout. See `pkg/gofr/logging/logger.go`:

```go
out := l.normalOut        // normalOut = os.Stdout
if level >= ERROR {
    out = l.errorOut      // errorOut  = os.Stderr
}
```

`AddStaticFiles` logs the failure via `Errorf` (ERROR level), so it goes to stderr. The existing `TestStaticHandlerInvalidFilePath` just above also captures the same `Errorf` output via `StderrOutputForFunc`, so this test follows that pattern.

I also verified empirically on Linux (where `os.Getwd()` does fail after the cwd is removed):
- with `StderrOutputForFunc` → test **passes**, the ERROR line is captured.
- with `StdoutOutputForFunc` → test **fails**: the captured string contains only the `WARN` config lines, and both assertions fail on the missing error message.

So `StderrOutputForFunc` is the correct choice here. Happy to adjust if I'm missing something!

aryanmehrotra (dismissed):
You're right, my mistake. ERROR-level logs route to `errorOut = os.Stderr` in the logger, so `StderrOutputForFunc` is correct — and `TestStaticHandlerInvalidFilePath` right above follows the same pattern. Production code is clean. LGTM.

pkg/gofr/gofr.go
```diff
@@ -411,7 +411,14 @@ func (a *App) AddStaticFiles(endpoint, filePath string) {
 
 	// update file path based on current directory if it starts with ./
 	if strings.HasPrefix(filePath, "./") {
-		currentWorkingDir, _ := os.Getwd()
+		currentWorkingDir, err := os.Getwd()
+		if err != nil {
+			a.container.Logger.Errorf("error in registering '%s' static endpoint, "+
+				"failed to get current working directory: %v", endpoint, err)
```
Umang01-hash, line 421: nit: this log prints the raw endpoint ('gofrTest'), while the os.Stat error log a few lines down prints the normalized '/gofrTest'. Same function, two error logs, different endpoint formatting — a log search for '/gofrTest' would miss this one. Moving the `endpoint = "/" + strings.TrimPrefix(...)` line above this block would unify them (the test asserts the raw form, so it'd need a tweak too). Non-blocking.

Umang01-hash (dismissed):
Rebased onto latest development (the conflict was stale-history from the force-push, not real — the two commits cherry-pick clean; diff is now just +39/-1). Fix is correct and matches the os.Stat error-path convention right below it.

Verified locally: gofmt/vet/build/golangci-lint --new-from-rev all clean. Ran the new test on Linux (it self-skips on macOS since the kernel still resolves an unlinked cwd) — it passes, and reverting the fix makes it fail both assertions, so it genuinely guards the branch. LGTM. One nit below, non-blocking.

aryanmehrotra (commented):
Thanks for the fix — I pulled this down and verified it on both macOS and Linux (the latter in a container, matching what CI runs). The change is correct and the test genuinely pins it. One small thing I'd like changed before merge, plus two notes.

### What I verified locally

| Check | Platform | Result |
|---|---|---|
| `go build ./pkg/gofr/` + `go vet ./pkg/gofr/` | darwin/arm64, go1.26.3 | pass |
| `go test ./pkg/gofr/ -run TestStaticHandler` | darwin | 2 pass, 1 **skip** (the new test) |
| `APP_ENV=test go test -short -covermode=atomic -coverpkg=./pkg/gofr ./pkg/gofr` (the CI command) | linux/arm64, go1.26.7 | ok, coverage 87.2% |
| Full `./pkg/gofr` package | darwin | ok |
| The new test applied on top of the **pre-fix** code | linux | **fails** — good, it really does pin the new branch |
| `golangci-lint run --new-from-rev=<base> ./pkg/gofr/...` | darwin | 0 issues |

Before/after, with the working directory unlinked on Linux:

```
before: error in registering '/gofrTest' static endpoint, error: stat somedir: no such file or directory
after:  error in registering 'gofrTest' static endpoint, failed to get current working directory: getwd: no such file or directory
```

So the endpoint fails to register in both versions — the added `return` is a pure logging improvement and not a behaviour change. I also probed the one case where the early return could theoretically diverge (`os.Getwd()` failing while a *relative* `os.Stat` still succeeds, which pre-fix would have stored a relative path in `staticFiles`). That state is unreachable on Linux: Go uses the raw `getcwd(2)` syscall, which resolves from the dentry cache and doesn't fail on a parent-directory `EACCES`. So the early return is safe.

### Please change

**The endpoint in the new log is un-normalised.** The `Errorf` fires before `endpoint = "/" + strings.TrimPrefix(endpoint, "/")`, so it prints `'gofrTest'` while the `os.Stat` error a few lines below prints `'/gofrTest'`. Two messages from the same function disagreeing on the endpoint format is the kind of thing that costs someone five minutes when they grep the logs — and the test currently carries a comment excusing it, which is a hint that the code is what should move.

Moving the normalisation above the block fixes both:

```go
	endpoint = "/" + strings.TrimPrefix(endpoint, "/")

	// update file path based on current directory if it starts with ./
	if strings.HasPrefix(filePath, "./") {
		currentWorkingDir, err := os.Getwd()
		if err != nil {
			a.container.Logger.Errorf("error in registering '%s' static endpoint, "+
				"failed to get current working directory: %v", endpoint, err)

			return
		}

	}
```

Then the test asserts `'/gofrTest'` and the explanatory comment can go.

### Notes (not blocking)

1. **The test skips on macOS.** Confirmed — `os.Getwd()` returns the stale path with a nil error for an unlinked working directory on darwin, so the branch is only exercised on Linux CI. That's acceptable (we already have a conditional skip in `pkg/gofr/http/router_test.go` for the running-as-root case), but if you'd rather have it run everywhere, a package-level `var getwd = os.Getwd` seam stubbed in the test would do it, with precedent at `pkg/gofr/datasource/opentsdb/opentsdb.go` (`var dialTimeout = net.DialTimeout`).

2. **Two sibling call sites still discard the same error**, and `factory.go` is the more commonly hit one:
   - `pkg/gofr/factory.go:55` — auto-registration of `./static`, which runs for every app that has that directory, without anyone calling `AddStaticFiles`.
   - `pkg/gofr/swagger.go:24` — per-request handler.

   Both fail closed later (`Router.AddStaticFiles` re-resolves via `filepath.Abs`, which calls `os.Getwd()` again), so neither is a bug this PR introduces. No need to widen the scope here — I'll open a follow-up. If you'd like to take `factory.go` in a separate PR, note that the log there should only fire on the branch that actually registers, otherwise every app without a `./static` directory gets a new startup error.

pkg/gofr/gofr.go
NitinKumar004: Good catch — fixed in 08b04b60. Moved the `endpoint = "/" + strings.TrimPrefix(endpoint, "/")` normalization up to the top of `AddStaticFiles` (right after `a.httpRegistered = true`), so both the `os.Getwd()` and `os.Stat` error logs now report the same `/gofrTest` form. Updated the getwd-error test assertion to `'/gofrTest'` accordingly. Thanks!

pkg/gofr/gofr_test.go
NitinKumar004: Double-checked this against the code and the stream is correct as-is:

- `pkg/gofr/logging/logger.go` routes ERROR-level logs to `errorOut`, which is `os.Stderr` (`out := l.normalOut; if level >= ERROR { out = l.errorOut }`, with `normalOut: os.Stdout, errorOut: os.Stderr`). `AddStaticFiles` logs via `Errorf` (ERROR), so it goes to **stderr**.
- The existing `TestStaticHandlerInvalidFilePath` right above exercises the same `Errorf` and uses `StderrOutputForFunc` (not `StdoutOutputForFunc`) — this new test follows that same pattern.
- `testutil.StderrOutputForFunc` swaps `os.Stderr`, and the app/logger is created *inside* the closure so its `errorOut` binds to the captured pipe.

Verified on Linux (where `os.Getwd()` fails after the cwd is removed): with `StderrOutputForFunc` the test passes; with `StdoutOutputForFunc` it fails on the missing error line. So `StderrOutputForFunc` is the right choice here. Happy to revisit if I'm missing something!

aryanmehrotra (approved):

NitinKumar004:
@aryanmehrotra thanks for the thorough re-review and the approval 🙏

**Requested change — done in `08b04b60`:** endpoint normalization now happens at the top of `AddStaticFiles` (right after `a.httpRegistered = true`), so both the `os.Getwd()` and `os.Stat` error logs report `/gofrTest`. The test asserts `'/gofrTest'` and the explanatory comment is removed.

**Note 1 (macOS skip):** confirmed — on darwin `os.Getwd()` returns the stale path with a nil error, so the error branch only runs on Linux CI (where the assertion is exercised). Kept the self-skip for now; happy to add a package-level `var getwd = os.Getwd` seam in a follow-up if we'd prefer it to run everywhere.

**Note 2 (sibling call sites):** verified in the code — `factory.go:55` (`./static` auto-registration, hit by every app with that dir) and `swagger.go:24` (per-request handler) both discard the same `os.Getwd()` error. Agreed they're pre-existing and out of scope here — I'll keep this PR focused and can take `factory.go` in a separate PR, making the log fire only on the branch that actually registers so apps without a `./static` dir don't get a spurious startup error.

PR #3601, opened 2026-06-20
fix(examples/grpc): widen histogram buckets to match recorded ms scale

### LINKED ISSUE #3453: gRPC example histogram buckets cap at 10ms — quantiles peg at ceiling for normal RPCs
## Problem

The example gRPC server in `examples/grpc/grpc-unary-server/server/health_gofr.go` defines histogram buckets with a max boundary of 10ms:

```go
[]float64{0.005, 0.01, .05, .075, .1, .125, .15, .2, .3, .5, .75, 1, 2, 3, 4, 5, 7.5, 10}
```

Any RPC that takes longer than 10ms lands in the `+Inf` bucket. When you run `histogram_quantile(0.95, ...)` on this, every method reports exactly 10.000ms for p50/p95/p99 — because that's the ceiling of the last finite bucket.

In comparison:
- HTTP histograms go up to **30s** (registered in `container.go`)
- Redis/SQL histograms go up to **30,000ms** (`getDefaultDatasourceBuckets()`)

The gRPC buckets are three orders of magnitude smaller, making latency monitoring useless for anything that isn't sub-10ms.

## Observed behavior

Running `histogram_quantile(0.95, rate(app_gRPC_Server_stats_bucket[5m]))` against a real deployment:

```
grpc.health.v1.Health/Check  → 10.000ms
/hello.HelloService/SayHello → 10.000ms
/order.OrderService/GetOrder → 10.000ms
```

Every method reports the same ceiling value regardless of actual latency.

## Suggested fix

Extend the bucket range to match the datasource histograms. Something like:

```go
[]float64{0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 25, 50, 100, 250, 500, 1000, 5000, 10000}
```

This also ties into the fact that gRPC histograms aren't auto-registered by the framework (separate issue) — the example is the only reference users have, so getting it right matters.

NitinKumar004:
## Description

The gRPC histograms record request duration in **milliseconds** (`recordGRPCMetrics` in
`pkg/gofr/grpc/log.go`), but the example bucket set capped at `10` — i.e. 10ms:

```go
gRPCBuckets := []float64{0.005, 0.01, .05, .075, .1, .125, .15, .2, .3, .5, .75, 1, 2, 3, 4, 5, 7.5, 10}
```

Any RPC slower than 10ms lands in the `+Inf` bucket, so
`histogram_quantile(0.95, rate(app_gRPC_Server_stats_bucket[5m]))` reports exactly 10.000ms
for p50/p95/p99 on every method, regardless of actual latency. For comparison, HTTP
histograms go up to 30s and the datasource histograms up to 30,000ms.

## Change

Widen the buckets to a ms-scaled range reaching 10,000ms (10s), in line with the HTTP and
datasource histograms, so quantiles reflect real latency:

```go
gRPCBuckets := []float64{0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 25, 50, 100, 250, 500, 1000, 5000, 10000}
```

Applied to all four gRPC examples (unary/streaming × server/client) so they stay consistent.

> **Note:** these files carry the `DO NOT EDIT` generated header — they are produced by
> `gofr-cli`. This PR fixes the checked-in examples as filed in the issue; the generator
> template lives in the separate `gofr-cli` repo and is out of scope here.

## Testing

- `go build ./examples/grpc/...`
- `go vet ./examples/grpc/...`
- `go test ./examples/grpc/...` — unary server/client tests pass. (The streaming-server
  test binds gRPC port 9000, which was occupied by an unrelated process on my machine; that
  failure is environmental and independent of this change, which only alters bucket values.)

Fixes #3453

### DIFF
--- a/examples/grpc/grpc-streaming-client/client/health_client.go
+++ b/examples/grpc/grpc-streaming-client/client/health_client.go
@@ -22,7 +22,7 @@ import (
 
 var (
 	metricsOnce sync.Once
-	gRPCBuckets = []float64{0.005, 0.01, .05, .075, .1, .125, .15, .2, .3, .5, .75, 1, 2, 3, 4, 5, 7.5, 10}
+	gRPCBuckets = []float64{0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 25, 50, 100, 250, 500, 1000, 5000, 10000}
 )
 
 type HealthClient interface {
--- a/examples/grpc/grpc-streaming-server/server/health_gofr.go
+++ b/examples/grpc/grpc-streaming-server/server/health_gofr.go
@@ -39,7 +39,7 @@ func registerServerWithGofr(app *gofr.App, srv any, registerFunc func(grpc.Servi
 
 	// Register metrics and health server only once
 	if !healthServerRegistered {
-		gRPCBuckets := []float64{0.005, 0.01, .05, .075, .1, .125, .15, .2, .3, .5, .75, 1, 2, 3, 4, 5, 7.5, 10}
+		gRPCBuckets := []float64{0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 25, 50, 100, 250, 500, 1000, 5000, 10000}
 		app.Metrics().NewHistogram("app_gRPC-Server_stats", "Response time of gRPC server in milliseconds.", gRPCBuckets...)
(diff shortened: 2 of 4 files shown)


aryanmehrotra:
@NitinKumar004 thanks for this! We actually widened these gRPC example histogram buckets in v1.57.0 — on `main` they now go up to 180s. This change would narrow the top end back to 10s, so closing as already-addressed. Appreciate the effort.

PR #3647, opened 2026-06-24
Update CONTRIBUTING.md

nivaas219:
## Pull Request Template

**Description:**

**Additional Information:**

**Checklist:**


aryanmehrotra (changes requested):
Two changes here, and I would like one without the other.

Adding the trailing newline at the end of the file is right, and worth having on its own.

The other hunk joins two wrapped lines into a single very long line, which breaks the wrapping the rest of the file uses and makes the diff noisier than the change it carries. The wording fix inside it — "to our own written code" becoming "for our own written code" — is fine; please keep it on the existing line break:

```
    - All external dependencies should go through the same careful consideration, we would have done for our own
      written code. We need to test the functionality we are going to use from an external library, as sometimes library
      implementation may change.
```

Happy to approve once the wrap is restored.

nivaas219:
> Two changes here, and I would like one without the other.
> 
> Adding the trailing newline at the end of the file is right, and worth having on its own.
> 
> The other hunk joins two wrapped lines into a single very long line, which breaks the wrapping the rest of the file uses and makes the diff noisier than the change it carries. The wording fix inside it — "to our own written code" becoming "for our own written code" — is fine; please keep it on the existing line break:
> 
> ```
>     - All external dependencies should go through the same careful consideration, we would have done for our own
>       written code. We need to test the functionality we are going to use from an external library, as sometimes library
>       implementation may change.
> ```
> 
> Happy to approve once the wrap is restored.

@aryanmehrotra ,Thanks for the review. I restored the original line wrapping and kept the wording improvement without increasing the line length. The changes are updated now.

CONTRIBUTING.md
```diff
@@ -126,11 +126,11 @@ docker run -it --rm -p 4443:4443 -e STORAGE_EMULATOR_HOST=0.0.0.0:4443 fsouza/fa
       conflict.
 * External Library uses:
     - A little copying is better than a little dependency.
-    - All external dependencies should go through the same careful consideration, we would have done to our own written
-      code. We need to test the functionality we are going to use from an external library, as sometimes library
+    - - All external dependencies should go through the same careful consideration, we would have done for our own
```
Umang01-hash: Double bullet here — should just be `- All external dependencies...`. Also `done to` was fine (or `given to`), `done for` reads a bit off.

Umang01-hash (changes requested):
The trailing newline at EOF is a good catch, but this adds a `- -` double bullet and reflows a couple of lines for no real gain. Can you drop those and keep just the newline fix? Left a note inline.

nivaas219:
@Umang01-hash ,Thanks for the review! Just wanted to clarify the expected final version before I make the changes.

Aryan’s earlier review mentioned that changing “to our own written code” to “for our own written code” was fine, while your latest review suggests keeping “to our own written code.”

Could you please confirm which version I should keep? I want to make sure I address the review correctly and don’t make any unnecessary changes.

Umang01-hash:
> @Umang01-hash ,Thanks for the review! Just wanted to clarify the expected final version before I make the changes.
> 
> Aryan’s earlier review mentioned that changing “to our own written code” to “for our own written code” was fine, while your latest review suggests keeping “to our own written code.”
> 
> Could you please confirm which version I should keep? I want to make sure I address the review correctly and don’t make any unnecessary changes.

@nivaas219 you can follow @aryanmehrotra suggestions.

nivaas219:
@Umang01-hash , I’ve addressed the review feedback and pushed the updated changes to the PR. I also followed Aryan’s suggestion regarding the wording and kept the existing line wrapping, while removing the duplicate bullet.

Could you please review the latest changes when you get a chance? Thank you!

aryanmehrotra:
Thanks for staying with this — you did address both reviews, and the wrap you restored is what I had asked for. My last review still sent you the wrong way on the wording, so let me give you the exact final text this time rather than describing it again.

## The wording change is not actually a fix

The original line:

> All external dependencies should go through the same careful consideration**,** we would have done **to** our own written code.

I said `to` → `for` was fine. It is not, and neither is the original — the preposition was never the problem. Two things are wrong with that sentence:

1. **`done`.** Consideration is *given*, not *done*. So `done to` and `done for` are wrong in exactly the same way; swapping one for the other moves the error without fixing it.
2. **the comma.** It sits where a relative pronoun belongs, which splices two clauses together. It should read `...consideration we would give...` (or `...consideration that we would give...`).

Since the sentence needs touching anyway, worth fixing properly rather than reverting.

## Exactly what to change

Replace lines **129-133** of `CONTRIBUTING.md` with this, verbatim:

```
    - All external dependencies should go through the same careful consideration we would give our own written
      code. We need to test the functionality we are going to use from an external library, as sometimes library
      implementation may change.
    - All dependencies must be abstracted as an interface. This will make it easier to switch libraries at later point
      of time.
```

That does three things: fixes the sentence, restores the `of time.` wrap you had joined, and drops the blank line that had appeared before `* Version tagging`.

**Leave the end of the file exactly as you have it** — the trailing newline is correct and is the most useful change in this PR. The last byte on `development` today is `.`, so until that lands, every future diff of this file carries a `\ No newline at end of file` marker.

On the wrapping, so it does not look like an arbitrary preference — measured line lengths in that block:

| line | on `development` | your current push | with the block above |
|---|---|---|---|
| `- All external dependencies...` | 119 | 112 | 110 |
| `code. We need to test...` | 112 | 120 | 112 |
| `- All dependencies must be abstracted...` | 118 + 14 (two lines) | **127** (one line) | 118 + 14 |
| longest in the block | 119 | **127** | **118** |

The blank line matters too: a blank line between the nested `-` items and the following `*` item makes the outer list render loose, so the spacing around it changes.

## After that

The PR becomes two hunks — one corrected sentence and one trailing newline — and I am happy to approve it. Sorry for the extra round; the wording detour was my error, not yours.

One thing for future PRs, unrelated: filling in the description with what changed and why makes small docs PRs much faster to review. Here the reviews had to infer the intent from the diff, which is part of why the wording question took three passes to settle.

nivaas219:
@aryanmehrotra , Thanks for the detailed explanation. I've updated the wording and formatting exactly as suggested, while keeping the trailing newline fix intact. Could you please take another look when you get a chance? Thank you!

CONTRIBUTING.md
```diff
@@ -125,12 +125,13 @@ docker run -it --rm -p 4443:4443 -e STORAGE_EMULATOR_HOST=0.0.0.0:4443 fsouza/fa
...
-      implementation may change.
-    - All dependencies must be abstracted as an interface. This will make it easier to switch libraries at later point
-      of time.
+
+  - A little copying is better than a little dependency.
+  - All external dependencies should go through the same careful consideration we would give our own written
+    code. We need to test the functionality we are going to use from an external library, as sometimes library
+    implementation may change.
+  - All dependencies must be abstracted as an interface. This will make it easier to switch libraries at later point
+    of time.
```
aryanmehrotra:
```suggestion
    - A little copying is better than a little dependency.
    - All external dependencies should go through the same careful consideration we would give our own written
      code. We need to test the functionality we are going to use from an external library, as sometimes library
      implementation may change.
    - All dependencies must be abstracted as an interface. This will make it easier to switch libraries at later point
      of time.
```

This restores the 4-space indentation the rest of the file uses and removes the blank line. The text is unchanged from your push.

aryanmehrotra (commented):
The wording is now exactly right — `consideration we would give our own written code`. I checked it properly: strip the leading whitespace and your block is **word-for-word identical** to what I asked for. The sentence and the trailing newline, which are the two things this PR is actually for, are both done.

What is left is only whitespace, so I have put it in a suggested change rather than describing it again — you can apply it with one click and nothing gets retyped. Describing indentation in prose is what has made this take four rounds, and that is my fault for not doing this sooner.

Two things in it:

1. **Indentation back to 4 spaces.** The rest of the file indents nested bullets by 4 (continuation lines by 6). Your push moved these three to 2, which leaves the file with two different nesting styles in adjacent sections:

| indentation | on `development` | with your push |
|---|---|---|
| 4 spaces | 13 nested bullets | 10 |
| 2 spaces | 0 | **3** — only the ones here |

2. **Drop the blank line** after `* External Library uses:`. A blank line between a list header and its nested items makes the list render loose, so the spacing around this block changes relative to every other section.

Neither is a judgement call and neither touches your text. Once this is in, the PR is two hunks — one corrected sentence, one trailing newline — and I will approve it.

Thanks for the patience on this one; it has taken far more rounds than the change deserved.

nivaas219:
@aryanmehrotra ,Thanks! I've applied the suggested indentation change and removed the extra blank line.

nivaas219:
@aryanmehrotra ,  just following up on the changes I made based on your review feedback. Whenever you get a chance, could you please take another look? Thanks!

aryanmehrotra (approved):

nivaas219:
@Umang01-hash  Hi, I've addressed all the requested feedback and Aryan has approved the latest version. Could you please take a final look when you have a chance? Thank you!

Umang01-hash (approved):

PR #3648, opened 2026-06-24
test(terminal): add unit tests for SetColor and ResetColor

hariantara:
## What

Adds unit tests for `SetColor` and `ResetColor` in `pkg/gofr/cmd/terminal/colors.go`, which previously had **0% test coverage**, and pins the sixteen color constants in the same file to their ANSI 256-color palette indices.

## Why

Both functions write ANSI escape sequences but were untested. Per the GoFr testing policy (every function should have unit tests, and coverage should not decrease), these were a small gap worth closing. This is a **test-only** change — no production code is modified.

## Details

- `SetColor` → table-driven subtests asserting it writes the literal escape sequence for Black, Red, Green, White and BrightWhite, plus an unnamed 256-color index (`208`) to document that the argument is passed through unvalidated.
- `ResetColor` → asserts it writes `\x1b[0m`.
- `TestSetColorResetColor` → the on / write / off round trip, the only place `ResetColor` actually undoing `SetColor` is observable in the byte stream.
- `TestColorConstants` → pins all sixteen constants to `0`–`15`. They are written straight into the SGR sequence, so inserting one shifts the ones below it and silently repaints every caller's output.
- Expectations are the **literal escape bytes**, not values derived from the package's `csi` constant. Deriving them makes the assertion move with the code: rewriting `csi` from `\x1b[` to `\x1b?` leaves a derived assertion passing. This matches `output_test.go`, which hard-codes `"\x1b[0m"`, `"\x1b[?47l"` and so on for that reason.
- Reuses the package's existing `tempOutput` / `validate` helpers from `output_test.go`; `validate` escapes `\x1b` in its failure message rather than emitting raw control characters.

## Verification

- `go test -short ./pkg/gofr/cmd/...` → passes; `-count=50` on the new tests, no flake
- `go test -race -count=5` on the new tests → clean
- Coverage: `SetColor` and `ResetColor` **0% → 100%**; package **93.2% → 95.1%**
- `gofmt -l`, `go build ./...`, `go vet`, `golangci-lint run ./pkg/gofr/cmd/terminal/...`, `typos` → all clean
- `git diff --name-only origin/development` → `pkg/gofr/cmd/terminal/colors_test.go` only

> Rebased onto current `development` and extended during review — see the comment below for the mutation results behind the changes.


aryanmehrotra:
Thanks @hariantara — `SetColor` and `ResetColor` were genuinely at 0%, and closing a gap like that with a test-only PR is exactly the right shape of contribution. I've rebased this onto current `development` (it had drifted to a 2026-02 merge base and was showing as conflicting) and pushed a follow-up commit off the review. Details below so nothing is a surprise.

### The one thing worth changing

The PR description says the tests "reference the package's own `csi` constant … so they stay correct if the escape representation ever changes". That is the property to avoid here, because it makes the assertion unfalsifiable in the one direction that matters:

```go
// colors.go
func (o *Out) SetColor(colorCode int) { o.Printf(csi+"38;5;%d"+"m", colorCode) }

// the test
expected := fmt.Sprintf(csi+"38;5;%dm", tc.colorCode)
```

Both sides are built from the same `csi`, so a change to `csi` moves them together. Proof — rewriting `csi` from `\x1b[` to `\x1b?`:

```
$ sed -i 's|csi = escape + "\["|csi = escape + "?"|' pkg/gofr/cmd/terminal/output.go

$ go test -run 'TestOut_SetColor|TestOut_ResetColor' ./pkg/gofr/cmd/terminal/
ok      gofr.dev/pkg/gofr/cmd/terminal  0.495s          # new tests: pass

$ go test -run 'TestReset$' ./pkg/gofr/cmd/terminal/
    Messages: output does not match, expected \x1b[0m, got \x1b?0m
FAIL                                                    # existing test: fails
```

`output_test.go` hard-codes `"\x1b[0m"`, `"\x1b[?47l"` and so on deliberately — the literal bytes are what a terminal actually consumes, so they are the contract worth pinning. The follow-up commit switches the expectations to literals to match.

### The bigger gap: nothing covered the constants

`colors.go` is mostly its `iota` block, and the codes are written straight into the SGR sequence. Insert one constant and the thirteen below it shift, silently repainting every caller's output. Inserting `Orange` after `Green` and running **the entire package suite**:

```
ok      gofr.dev/pkg/gofr/cmd/terminal  4.425s
```

Nothing failed. `TestColorConstants` now pins all sixteen to their palette indices, which I'd argue is the highest-value test in the file — more so than re-asserting a one-statement `Printf`.

### Mutation results

Five mutants against `colors.go` / `output.go`, this PR as submitted vs. after the follow-up:

| Mutant | As submitted | After |
|---|---|---|
| `csi` `\x1b[` → `\x1b?` | **survived** | killed |
| foreground `38` → background `48` | killed | killed |
| `ResetColor` `0m` → `00m` | killed | killed |
| trailing `m` dropped | killed | killed |
| color constant inserted (shifts 13 codes) | **survived** | killed |

### Smaller points, all applied

- **Reuses the package's own helpers.** `output_test.go` already has `tempOutput(t)` and `validate(t, o, exp)` in this package; `validate` escapes `\x1b` in its failure message, so a mismatch prints `\x1b[38;5;2m` instead of emitting raw control characters into your terminal.
- **`t.Run` subtests.** The table looped without subtests, so `tc.name` had to be threaded through the assert message by hand and `-run` could not select a single case.
- **Naming.** `TestOut_SetColor` was a third style in one package (`output_test.go` uses `TestReset`, `printers_test.go` uses `TestOutput_Printf`). Went with `output_test.go`'s, since `colors.go` is the same kind of escape-writing code.
- **Added an unnamed 256-color index (`208`)**, which documents why `SetColor` takes an `int` rather than an enum — the argument is passed through unvalidated.
- **Added `TestSetColorResetColor`**, the on/write/off round trip, which is the only place `ResetColor` actually undoing `SetColor` is observable in the byte stream.

### Verification

Test-only — `git diff --name-only origin/development` returns exactly `pkg/gofr/cmd/terminal/colors_test.go`, no production code touched.

| Gate | Result |
|---|---|
| `gofmt -l` | clean |
| `go build ./...` | pass |
| `go vet ./pkg/gofr/cmd/...` | pass |
| `golangci-lint run ./pkg/gofr/cmd/terminal/...` | 0 issues |
| `typos` | clean |
| `go test -short ./pkg/gofr/cmd/...` | pass |
| `go test -count=50` (new tests) | pass, no flake |
| `go test -race -count=5` (new tests) | clean |
| Coverage | package 93.2% → **95.1%**; `SetColor`, `ResetColor` 0% → **100%** |

One note on the coverage number: it is the same 95.1% the original tests reached. Both functions are a single statement, so they hit 100% either way — what the follow-up changes is what the tests can *detect*, not how many lines they touch.

### Unrelated, not fixed here

`go test -race ./pkg/gofr/cmd/terminal/` reports a data race in `TestSpinner` — `Spinner.Stop()` writes `isActive` at `spinner.go:78` while the goroutine started by `Spin()` reads it at `spinner.go:62`, and both then write the same `bytes.Buffer` through `Out`. It reproduces on clean `development` with this PR reverted, so it is not yours and it does not belong in a test-only PR — it needs a fix in `Spinner` itself.

Umang01-hash (approved):
Ran it locally at HEAD: gofmt/vet/golangci-lint clean, all tests pass. Behavioral coverage on previously-untested code — `TestColorConstants` pins the iota indices so a reorder fails red, `TestSetColor` asserts the real escape bytes, and the on→write→off pairing is covered. LGTM.

PR #3649, opened 2026-06-25
feat: implement EventHub Health check

### LINKED ISSUE #3639: EventHub health check is not implemented, returns empty Health struct
## Problem

`pkg/gofr/datasource/pubsub/eventhub/eventhub.go` has a stub `Health()` implementation that logs an error and returns an empty `datasource.Health{}`:

```go
func (c *Client) Health() datasource.Health {
    c.logger.Error("health-check not implemented for Event Hub")
    return datasource.Health{}
}
```

Every other pubsub provider (Kafka, SQS, NATS, MQTT, Google Pub/Sub) implements a real health check. EventHub is the only one missing it.

## Impact

In Kubernetes deployments, liveness and readiness probes call the health endpoint. With the current stub, EventHub connectivity issues are never detected. Pods that lose their EventHub connection are never restarted, causing silent message loss.

## Expected Fix

Use `c.consumer.GetEventHubProperties(ctx, nil)` as a lightweight connectivity probe. This call is already used in the codebase at lines 295 and 396 for the subscribe flow, and the timeout constant `eventHubPropsTimeout = 2 * time.Second` is already defined at line 31.

```go
func (c *Client) Health() datasource.Health {
    health := datasource.Health{
        Status: datasource.StatusDown,
        Details: map[string]any{
            "backend":  "EVENT_HUB",
            "eventHub": c.cfg.EventhubName,
        },
    }

    if c.consumer == nil {
        health.Details["error"] = "client not connected"
        return health
    }

    ctx, cancel := context.WithTimeout(context.Background(), eventHubPropsTimeout)
    defer cancel()

    props, err := c.consumer.GetEventHubProperties(ctx, nil)
    if err != nil {
        health.Details["error"] = err.Error()
        return health
    }

    health.Status = datasource.StatusUp
    health.Details["partitionCount"] = len(props.PartitionIDs)

    return health
}
```

## Files to Change

| File | Change |
|------|--------|
| `pkg/gofr/datasource/pubsub/eventhub/eventhub.go` | Replace `Health()` stub with real implementation |
| `pkg/gofr/datasource/pubsub/eventhub/eventhub_test.go` | Update `Test_Health` and `Test_HealthCheck` to test real up/down behavior |

## Reference

- `pkg/gofr/datasource/pubsub/sqs/health.go` — cleanest template (uses `ListQueues` as probe)
- `pkg/gofr/datasource/pubsub/eventhub/eventhub.go:295` — where `GetEventHubProperties` is already used
- `pkg/gofr/datasource/pubsub/eventhub/eventhub.go:31` — `eventHubPropsTimeout` constant

---

I'd like to work on this. Can I be assigned?

Lucky-Malik:
## Description

`EventHub`'s `Health()` was a stub — it logged `health-check not implemented for Event Hub` and returned an empty `datasource.Health{}`. Every other pubsub provider (Kafka, SQS, NATS, MQTT, Google Pub/Sub) reports a real status, so EventHub was the only one without one.

That empty struct has a status of `""`, which is not `"DOWN"`, so `Container.Health` never counted it as a failing dependency. An application whose Event Hub was completely unreachable reported itself healthy — verified against a running app with the namespace pointed somewhere that cannot resolve:

```
before:   {"data":{"name":"e2e-3649","status":"UP"}}
after:    {"data":{"name":"e2e-3649","status":"DEGRADED"}}   (answered in 2.001s)
```

Under Kubernetes that is a pod that lost its Event Hub and is never restarted: silent message loss.

This PR wires up a real check using `c.consumer.GetEventHubProperties(ctx, nil)` as a lightweight connectivity probe — the same call already used by the subscribe fallback and by `isLikelyBasicTier` — bounded by the existing `eventHubPropsTimeout` (2s) -- **and bounded by this package, not by the SDK**. Handing the context to `GetEventHubProperties` is not sufficient: in `azeventhubs` v1.4.0 that method passes `context.Background()` to `rpcLink.RPC` (`mgmt.go:67`) and the wait for the reply selects on *that* context (`internal/rpc.go:263`), so our deadline never reaches the AMQP round-trip, and with no deadline on it there is no `server-timeout` property either. What the context does bound is link acquisition and retry backoff -- which is why an unresolvable host returns in 2s and the bound looks correct until a broker accepts a request and never answers. `probeWithin` runs the call on its own goroutine and selects on the deadline here. It returns `StatusUp`/`StatusDown` along with `backend`, `eventHub`, and, on success, `partitionCount`, mirroring the SQS implementation.

Note on where those details surface: `/.well-known/health` is deliberately reduced to `{name, status}` and discards every per-dependency detail (see `aggregateStatus` in `pkg/gofr/health.go`). `backend`, `eventHub` and `partitionCount` are therefore read through `Container.Health`; what reaches a liveness or readiness probe is the aggregate status.

Fixes #3639

## Breaking Changes

None. `Health()` previously returned an empty struct and now returns a populated one. No signature or config change.

The unexported `Client.consumer` field is now the narrow `consumerClient` interface rather than `*azeventhubs.ConsumerClient`, matching how the SQS client holds its own client. Unexported, so no effect outside the package.

## Additional Information

- No new dependencies — uses the already-imported `azeventhubs` consumer client.
- `Health` is covered on every branch: `Test_Health_Connected` (`StatusUp` and `partitionCount`), `Test_Health_ProbeError` (the probe error passed through verbatim), `Test_Health_NotConnectedDoesNotProbe` (the nil short-circuit never dials), `Test_Health_BoundsAProbeThatIgnoresContext` (a probe that blocks without reading its context -- what the SDK does -- still returns on our deadline; reverting `probeWithin` makes this test hang rather than fail, which is the production symptom), and `Test_HealthCheck` against a live client, which asserts the probe returns within 2× `eventHubPropsTimeout` — the margin is deliberate so the assertion does not flake on a loaded runner; measured, it returns in 1×.
- That bound is the load-bearing part: with a plain `context.WithCancel` in its place the same probe takes 42 seconds, on the SDK's own retry schedule.
- Submodule coverage 42.2% → 45.2%.

## Checklist


aryanmehrotra:
Reviewed this end to end and pushed the review changes onto the branch as a separate commit — your commit is untouched.

The change is right and it fixes a real, demonstrable hole. I ran a GoFr app with the Event Hub pubsub wired in through `AddPubSub`, pointed at a namespace that cannot resolve, and probed the health endpoint over the wire:

```
development:   {"data":{"name":"e2e-3649","status":"UP"}}        <- unreachable Event Hub, app reports healthy
this PR:       {"data":{"name":"e2e-3649","status":"DEGRADED"}}  <- answered in 2.001s
```

That is exactly the failure your description names, and it now behaves.

### What the tests could not catch

The health check's only reason to exist is the branch that reports the backend as *usable*, and nothing in the suite could reach it. `Client.consumer` is a concrete `*azeventhubs.ConsumerClient`, so a test has no way to make `GetEventHubProperties` succeed — `Test_HealthCheck` points at an unreachable namespace, so it can only ever take the down path.

Changing `StatusUp` to `StatusDown` in the success branch left the entire package green:

```
$ sed -i 's/health.Status = datasource.StatusUp/health.Status = datasource.StatusDown/' eventhub.go
$ go test -short ./...
ok      gofr.dev/pkg/gofr/datasource/pubsub/eventhub     2.580s
```

So did returning the wrong partition count.

The SQS client — the one this implementation mirrors, and it mirrors it well — already keeps its client behind a narrow `sqsClient` interface for exactly this reason, and its `health_test.go` covers all four paths with a hand-written fake. I added the same seam here (`consumerClient`, three methods) and the matching tests.

### The 2 second bound was untested, and it matters most

`Test_HealthCheck` takes 2.00s because the probe runs out its context, so the error it asserts is always `context deadline exceeded` — never a real one. Two changes: the probe error is now pinned verbatim in a stubbed test, and the live test asserts the thing it *is* good for, that the probe stays inside `eventHubPropsTimeout`.

That bound turns out to be the load-bearing part. Replacing `context.WithTimeout` with a plain `context.WithCancel` — the easy accidental edit — makes the same probe take **42 seconds**, because the SDK's own retry schedule takes over:

```
--- FAIL: Test_HealthCheck (41.99s)
        Error: "41.993242083s" is not less than "4s"
```

A liveness probe that blocks for 42s is a pod that gets killed. The suite as submitted passed that mutation.

### Mutation results

| Mutation | As submitted | After |
|---|---|---|
| success branch reports `StatusDown` | **survived** | killed |
| `partitionCount` off by one | **survived** | killed |
| `context.WithTimeout` → `WithCancel` (42s probe) | **survived** | killed |
| nil-consumer guard removed | killed | killed |
| probe error swallowed | killed | killed |
| `backend` renamed | killed | killed |

### Smaller things

**`errClientNotConnected`.** This package already has a sentinel with that exact meaning, returned by `Publish`, `Subscribe` and `Query`. The health detail now uses `errClientNotConnected.Error()` so an operator sees one string for one condition rather than two. (SQS uses a literal, but it has no sentinel of its own to reuse.)

**Where `Details` actually surfaces.** Worth knowing, because it changes what the description promises: `/.well-known/health` is deliberately reduced to `{name, status}` and discards every per-dependency detail — see the comment on `aggregateStatus` in `pkg/gofr/health.go`. So `backend`, `eventHub` and `partitionCount` are read through `Container.Health`, not off the public endpoint. What reaches Kubernetes is the aggregate flipping to `DEGRADED`, which is the part that matters and which the e2e confirms. Nothing to change; I've reworded the description.

**Base branch.** This targets `main`; GoFr merges to `development`, which is why it shows as conflicting against a merge base from the initial commit. The branch is now rebuilt on current `development` — retarget it to `development` and the conflict goes away.

### Gates

All on the rebuilt branch:

| Check | Result |
|---|---|
| `gofmt -l` | clean |
| `go build ./...` (module + `go.work`) | pass |
| `go vet ./...` | pass |
| `golangci-lint` | 3 issues, all pre-existing on `development` (`goconst` `EVHUB`, 2× `wsl_v5`); 0 added |
| `go test -short ./...` | pass |
| `go test -race -short ./...` | pass |
| `-count=10` on the health tests | no flake |
| coverage (submodule) | 42.2% → **44.1%**; `Health` back to 100% |
| e2e: 11 checks vs a real server | all pass |

The e2e also drove 10 further probes and measured goroutine growth of 0, since a Kubernetes probe hits this forever and each call opens an AMQP attempt.

Nice catch on the issue — the stub silently reporting a dead backend as healthy is the worst shape a health check can have.

aryanmehrotra:
Ran a second review pass over my own commit on this branch. It turned up three things I got wrong in [my earlier comment](https://github.com/gofr-dev/gofr/pull/3649#issuecomment-5548954465), and four things worth fixing in the code. Corrections first, since those are the ones that mislead if left standing.

### Corrections to my earlier comment

**1. I said SQS "has no sentinel of its own to reuse". That is backwards.** SQS has one, same name and same meaning:

```go
// pkg/gofr/datasource/pubsub/sqs/sqs.go:21
errClientNotConnected = errors.New("sqs client not connected")
```

and `sqs/health.go:24` uses the bare literal `"client not connected"` anyway. So the parenthetical was exactly inverted — SQS has the sentinel and doesn't use it. Worth fixing there too, separately.

**2. The mutation row `WithTimeout` → `WithCancel` — "as submitted: survived" — does not reproduce.** Re-running it at `8ea54cd39` it fails 6 times out of 6. Not on an assertion, though: removing the bound gives a leaked `processor.Run` goroutine ~40 seconds to call `Errorf` on a gomock logger whose test has already finished, which panics the binary:

```
panic: Fail in goroutine after TestConnect has completed
FAIL  gofr.dev/pkg/gofr/datasource/pubsub/eventhub  39.79s
```

That is environment-dependent — it needs the SDK to error rather than hang — so my original run may genuinely have survived. But I posted it as a fact, and anyone re-running it sees a FAIL. The defensible version of the claim is narrower: **the as-submitted suite has no assertion that can catch an unbounded probe**, which is still the reason the bounded test was worth adding. The other three rows of that table reproduce as posted.

**3. I wrote that the live test "asserts the probe stays inside `eventHubPropsTimeout`".** The assertion is `elapsed < 2*eventHubPropsTimeout` — 4s, not 2s. The margin is deliberate, so a tight bound doesn't flake on a loaded runner, but I stated it wrongly in the comment, in the PR description and in the test's own failure message. All three now say 2×.

### Fixed in `eeede0b67`

- **`Test_Health_NotConnectedDoesNotProbe` contained unreachable code.** It built a mock whose `getPropsFunc` called `t.Error("Health must not probe...")` — and then set `client.consumer = nil`, discarding it. That `t.Error` could not fire under any mutation. The test now passes `nil` directly and says what the real failure signal is: delete the short-circuit and `Health` calls through a nil interface and panics. Verified by doing exactly that — it panics in this test.
- **The `consumerClient` doc claimed the partition reads can be exercised through the seam.** They can't. `NewPartitionClient` returns the concrete `*azeventhubs.PartitionClient` and `ReceiveEvents` is called on that, so a fake can only choose which partition IDs come back. Comment corrected rather than the claim quietly left in.
- **The fake's `NewPartitionClient` returned `(nil, nil)`**, which would nil-deref on the deferred `Close` in the first test to reach `tryReadFromPartition`. It returns an error now.
- **The local in `Connect` shadowed the `consumerClient` type**, making the type unnameable inside the function that constructs it. Renamed to `consumer`.

### One thing neither pass caught until now

The 2s probe is bounded, but `container/health.go` runs every dependency check **serially** with no memoisation, and Kubernetes' default `timeoutSeconds` is **1**. So an operator on defaults gets a probe timeout, not the clean `DEGRADED` body this PR produces. The restart decision comes out the same either way — a timeout is also a failure — but my earlier comment implied the body is what they'd see, and that only holds once the probe timeout is raised. Not this PR's problem to fix; worth knowing.

### Gates after the change

`gofmt` clean · `go vet` clean · `go build ./...` clean · `go test -race -count=2` ok · `golangci-lint --new-from-rev` **0 issues** · coverage still **44.1%** (42.2% on `development`), `Health` at 100%.

aryanmehrotra:
Reviewed this end to end and pushed onto the branch, authorship intact. Retargeted from `main` to `development`.

## What the PR fixes, demonstrated

Against a completely unreachable namespace, driven over HTTP:

```
development : {"name":"e2e-3649","status":"UP"}        <- reports healthy
this branch : status DEGRADED, answered in 2.001s
```

`Health()` was a stub returning an empty `datasource.Health{}`. Because `container/health.go` counts a dependency down only on the literal `"DOWN"`, and the stub's status was `""`, **EventHub could never mark an app degraded.**

⚠️ **That is worth a release note.** Anything gating on `GET /.well-known/health` — a readiness probe, a load balancer — will now see `DEGRADED` where it always saw `UP`.

Bonus: on `development`, `New(cfg).Health()` without `Connect()` **segfaults** (`nil pointer dereference` at `eventhub.go:464`). This PR turns that into a reported `DOWN`.

## The thing I got wrong, and fixed

My first pass claimed the probe was bounded by `eventHubPropsTimeout`. **It was not.** In `azeventhubs` v1.4.0:

- `mgmt.go:67` — `rpcLink.RPC(context.Background(), amqpMsg)` — our context is discarded
- `internal/rpc.go:263` — the wait for the reply selects on *that* `Background` context, so our deadline never fires
- `internal/rpc.go:244` — `server-timeout` is set only when the context has a deadline; `Background` has none, so there is no broker-side bound either

The context bounded link acquisition and retry backoff — which is exactly why an unresolvable host returns in 2.00s and the bound *looked* right. With a management link already open and a broker that accepts a request and never answers, `Health()` blocked with no limit, on a serial health endpoint.

`probeWithin` now runs the call on its own goroutine and selects on the deadline here, so the bound is ours. `Test_Health_BoundsAProbeThatIgnoresContext` covers it with a fake that blocks without reading `ctx` — what the SDK does. Note how it fails: **reverting the fix does not fail the test, it hangs it** (`panic: test timed out after 25s`). That is the production symptom.

## Tests, after mutation testing

Three mutants survived the branch as submitted — `StatusUp`→`StatusDown`, `partitionCount` off by one, and removing the deadline entirely — and are now killed. Three more survived my own first attempt and are also closed:

- `partitionCount` replaced by the constant `4`, and a `len > 0` guard around it: the only connected test used four partitions. Now covers zero, one and three.
- Raising `eventHubPropsTimeout` 2s → 30s passed, because the elapsed-time assertion is bounded by `2*eventHubPropsTimeout` and both move together. `Test_Health_ProbeDeadlineIsTwoSeconds` reads the deadline the probe was given and compares it against a literal — kills it in 0.00s.

Coverage 42.2% → 45.2%, `Health` at 100%.

## Smaller corrections

- `Test_Health_NotConnectedDoesNotProbe` built a mock whose `t.Error` was **unreachable** — it then set `consumer = nil`, discarding it. Now passes nil directly and says what the real signal is: remove the guard and `Health` panics.
- The `consumerClient` doc claimed partition reads can be exercised through the seam. They cannot — `NewPartitionClient` returns the concrete type and `ReceiveEvents` is called on it.
- The fake's doc said that type **cannot be constructed**; `&azeventhubs.PartitionClient{}` compiles. It cannot be constructed *usably*.
- The local in `Connect` shadowed the `consumerClient` type. Renamed.
- SQS **does** have its own `errClientNotConnected` (`sqs.go:21`) and uses a bare literal anyway (`health.go:25`) — I said the opposite. Worth fixing there separately.

## One trade-off recorded, not fixed

Widening `consumer` from a pointer to an interface makes it two words, so an unsynchronised concurrent `Connect` can be observed half-written and pass a `!= nil` guard — measured, 4 nil-pointer panics in 1,239,434 rounds versus 0 with the pointer. Not reachable through GoFr's own wiring (`AddPubSub` calls `Connect` synchronously before `Run`), but `Connect` is exported. Documented on the field; synchronising it is a larger change than this PR should carry.

Also measured: the interface makes `&PartitionClientOptions{}` escape to the heap at three sites (`-gcflags=-m`, +1 alloc / +64 B each). Noise against AMQP link setup, but real and previously unmentioned.

## Gates

`gofmt` · `go vet` · `go build ./...` · `go test -race -count=2` · `golangci-lint --new-from-rev` **0 issues**.

Umang01-hash (approved):
Ran it locally at HEAD (own module): gofmt/vet/build clean, golangci-lint 0 new issues, full suite passes with -race. Verified the SDK claim at source — azeventhubs v1.4.0 GetEventHubProperties passes context.Background() to the AMQP RPC (mgmt.go:67) and the reply-wait selects on that (internal/rpc.go:263), so passing ctx doesn't bound it; the goroutine+select in probeWithin is necessary, not gold-plating. Interface seam matches the SQS client precedent, Health Details shape matches kafka/sqs. Tests are proper mutation-killers (partitionCount, deadline, context-ignoring probe). LGTM.

Two optional follow-ups, neither blocking: (1) an abandoned probe's goroutine parks until the SDK returns — against a broker that accepts but never answers, a repeatedly-polled Health could slowly leak goroutines (documented, and strictly better than the prior unbounded stub); (2) pre-existing isLikelyBasicTier has the same unbounded-probe issue and could reuse probeWithin.

PR #3652, opened 2026-06-29
chore(datasource): bump gofr.dev to v1.57.0 across submodules

Umang01-hash:
## What

Bumps the `gofr.dev` core dependency from `v1.56.6` to `v1.57.0` in all 10 datasource submodules that import it, plus the `using-cloudsql` example, ahead of cutting their release tags for the v1.57.0 cycle.

Modules bumped: dbresolver, file/azure, file/ftp, file/gcs, file/s3, file/sftp, oracle, pubsub/eventhub, pubsub/nats, pubsub/sqs (+ examples/using-cloudsql).

## Why

Datasource tags are cut from `main`, so the updated `go.mod` must be on `main` before tagging. This PR is the prerequisite for tagging the submodule releases.

## Verification

- `go build ./...` and `go vet ./...` pass for all 11 modules (`GOWORK=off`).
- Unit tests pass for 8/10; the 2 failures (`dbresolver` metrics-port bind, `eventhub` mock expectation) are pre-existing on `main` and reproduce identically at v1.56.6 — not introduced by this bump.


Umang01-hash:

PR #3653, opened 2026-06-29
test(eventhub): fix mock expectations for consumer-group + init logs

Umang01-hash:
## Problem

The eventhub unit tests are red on `main` (independent of any dependency bump). The consumer-group interpretation change added two log calls to `Connect()`:
- `Debugf("Using default consumer group: %s", ...)` / `"Using provided consumer group"` (`eventhub.go:141-143`)
- `Debug("Event Hub client initialization complete")` (`eventhub.go:218`)

…but the existing `Connect()`-calling tests (`TestConnect`, `TestPublish_FailedInvalidTopic`, `Test_CreateTopic`, `Test_DeleteTopic`, `Test_HealthCheck`, plus the loose `TestConnect_ProducerError/ContainerError`) were never updated, so they fail on the unexpected mock calls. gomock aborts on the first unexpected call, which masked the second gap.

## Fix

Test-only: add the missing `Debugf`/`Debug` mock expectations to the affected tests, matching the style already used by the newer `TestConnect_ConsumerGroupDefaults/Provided` tests. No production code changed.

## Verification

- `go vet ./...` clean
- `go test ./...` passes
- `go test -race -count=3 ./...` passes (no flakiness)


Umang01-hash:

PR #3654, opened 2026-06-29
chore(datasource): complete go.sum for standalone builds

Umang01-hash:
## Problem

8 datasource modules fail to build standalone (`GOWORK=off go build ./...`) on `main` with `missing go.sum entry` errors:
arangodb, clickhouse, couchbase, dgraph, influxdb, kv-store/badger, kv-store/nats, mongo.

Their last release tags (e.g. mongo v0.4.2) build fine — the drift was introduced by applying dependency updates through the workspace (`go.work.sum`) without running per-module `go mod tidy`. Consumers are unaffected (they compute their own go.sum), but standalone builds, CI, and contributors break — and we shouldn't cut new tags on modules that don't build standalone.

## Fix

`go mod tidy` per affected module. **Only `go.sum` changes — no `go.mod` changes**, no version changes.

## Verification

- All 8 now pass `GOWORK=off go build ./...` and `go vet ./...`.
- The other datasource modules already build standalone and are untouched.

## Why now

Prerequisite for cutting the v1.57.0-cycle release tags for these modules.


Umang01-hash:

PR #3676, opened 2026-07-01
fix(crud): keep digits intact in toSnakeCase for auto-generated SQL

umarahad2005:
## Description

`toSnakeCase` (`pkg/gofr/crud_helpers.go`) turns a struct name into a **table name** and each field name into a **column name** for GoFr's automatic CRUD handlers.

Its pass-through guard was `char >= 'a'`, so *every* rune below `'a'` fell into the uppercase branch and had `'a' - 'A'` (32) added to it. Digits and punctuation sort below `'a'`, so both were silently shifted into unrelated characters:

| input | before | after |
|-------|--------|-------|
| `User1` | `user_Q` | `user1` |
| `S3Bucket` | `sS_bucket` | `s3_bucket` |
| `Base64Encoder` | `base_VT_encoder` | `base64_encoder` |
| `User2Name` | `user_R_name` | `user2_name` |
| `Field_1` | `field_\x7fQ` | `field_1` |

The last row is `'_'` (95) becoming DEL (127) — a control character in a SQL identifier.

Because these strings become real table and column names, the failure is total rather than subtle. Against a real database, an entity like `type oauth2Client struct { Address2, X509Cert, Order2FA, Sha256Sum string }`:

```
before:  POST /oauth2client
         500 {"error":{"message":"Error 1146 (42S02): Table 'test.oauth_R_client' doesn't exist"}}
         GET, PUT and DELETE all answer the same.

after:   201 created
         SELECT address2, x509_cert, order2_fa, sha256_sum FROM oauth2_client
         ->  "flat 2", "cert-pem", "totp", "abc123"
```

## Fix

- Restrict the uppercase branch to `'A'`–`'Z'`, so digits, underscores, punctuation and non-ASCII pass through unchanged. This states the rule positively — only ASCII uppercase is special — rather than relying on an ordering accident.
- Treat a digit as ending a word, so an uppercase letter following one starts a new one. The underscore decision used `rune(str[i-1]) >= 'a'` as a stand-in for "is lowercase", which is the same shorthand that caused the original bug; a digit is below `'a'`, so `Order2FA` did not split. A digit only counts on the left, so `AB2` stays `ab2`.
- `isLower` and `isLowerOrDigit` are written out instead of the `>= 'a'` shorthand, which is true for every rune above `'a'`, non-ASCII included.

No tracking issue exists; this was found while reading the auto-CRUD name generation.

## Consistency with the Cassandra converter

GoFr has more than one snake-case converter. The nearest is `ToSnakeCase` in `pkg/gofr/datasource/sql/db.go`, which `DB.Select` uses to map result columns (`db.go:333`) and which this very file already imports — so before this change, `db.Select` and the auto-generated CRUD handlers disagreed about the column names of the same struct, inside the same datasource. `pkg/gofr/datasource/cassandra/internal.go` has the same algorithm and gives the same answers.

Measured over every 1-, 2- and 3-character identifier over `[a-zA-Z0-9_]` — 254,079 inputs — against that regex converter:

| | disagreements |
|---|---|
| `development` | 110,715 |
| first commit on this branch | 26,260 |
| **this branch** | **0** |

Two limits on that, stated so nobody has to rediscover them:

- **ASCII only.** The regex converters finish with `strings.ToLower`, which is Unicode-aware; this loop only touches `A`–`Z`. An identifier carrying a non-ASCII capital still diverges — `"AÉ"` becomes `aÉ` here and `aé` there — and so still leaves a capital letter in a generated name. 120 of a 280-identifier non-ASCII corpus. Out of scope for this fix, but real.
- **ScyllaDB is a fourth implementation** and does not agree. `pkg/gofr/datasource/scylladb/internal.go:212` uses `github.com/stoewer/go-strcase`, which splits digits the other way: `Order2FA` → `order2fa`, `S3Bucket` → `s3bucket`, `Base64Encoder` → `base64encoder`.

So a follow-up issue **is** warranted, covering ScyllaDB and the Unicode case. An earlier revision of this description said an entity maps to the same column name on any backend and that no follow-up was needed; both halves of that were wrong.

## Breaking Changes

None — and although this changes generated SQL, no working setup can depend on the old output:

- Every identifier is quoted by `bind.go` (`"col"` for Postgres, backticks for MySQL), so a mangled name like `"base_VT_data"` is case-preserving and only ever matches a literal mixed-case column of that exact spelling — not something anyone designs.
- There is no column-name override or struct tag (only `TableNameOverrider` / `RestPathOverrider`), so an entity with a digit in a field name is **broken at every CRUD endpoint today**: it fails outright rather than silently returning wrong data, as the 500s above show.

## Additional Information

- No new dependencies.
- `toSnakeCase`, `isLower` and `isLowerOrDigit` are at 100% statement coverage, and each of the three comparison boundaries they turn on is pinned by its own case (`"ABé"`, `"A0B"`, `"X509CERT"`) — statement coverage of a one-line predicate says nothing about whether its bounds can be moved.
- Verified end to end against a real GoFr app and real MySQL, driving all four generated CRUD handlers over HTTP for an entity with digits in both its type name and its field names: create, read by id, update and delete all succeed, and the row is readable under `address2`, `x509_cert`, `order2_fa` and `sha256_sum`.

## Checklist


aryanmehrotra (dismissed):
Verified against the real function on `development`, and the blast radius is wider than the description says. `toSnakeCase` feeds both the column names and the table name into the generated CRUD SQL (`crud_handlers.go:86,102,193,209,299` and `crud_helpers.go:16`):

```
User1          -> "user_Q"
S3Bucket       -> "sS_bucket"
Base64Encoder  -> "base_VT_encoder"
User2Name      -> "user_R_name"
Order2FA       -> "order_Rfa"
UserID         -> "user_id"      (unaffected, as expected)
```

The cause is exactly as you describe: digits sort below `'a'`, so `char >= 'a'` sent them into the uppercase branch and `char-'A'+'a'` mapped `'1'` to `'Q'`. Inverting the guard to `char < 'A' || char > 'Z'` is the right shape — it states the rule positively (only ASCII uppercase is special) rather than relying on an ordering accident, so runes above `'z'` also pass through by intent rather than by luck.

## This is a fix, not a breaking change, and the quoting is why

Worth stating in the description, because "changes generated SQL" reads alarming on its own.

`bind.go:42` quotes every identifier — `"` for Postgres, backticks for MySQL — so the emitted statement is:

```sql
UPDATE "user" SET "base_VT_data"=$1 WHERE "id"=$2
```

A quoted identifier is case-sensitive in Postgres and case-preserving in MySQL, so there is no folding that could make `"base_VT_data"` resolve to a column named `base_vt_data`. The only schema this matches is one containing a literal mixed-case `base_VT_data` — not something anyone designs, and only reachable by reverse-engineering it out of a GoFr error message.

So nobody has a working setup that depends on the current output: every entity with a digit in a field name is broken today at every CRUD endpoint, not silently returning wrong data. And there is no escape hatch — `getTableName` has `TableNameOverrider` and `getRestPath` has `RestPathOverrider`, but there is no column-name override or struct tag, so an affected user can only rename the Go field.

## GoFr already does this correctly elsewhere

`pkg/gofr/datasource/cassandra/internal.go:150` carries its own regex-based `toSnakeCase`:

```
               cassandra         pkg/gofr (before)
User1          user1             user_Q
S3Bucket       s3_bucket         sS_bucket
Base64Encoder  base64_encoder    base_VT_encoder
User2Name      user2_name        user_R_name
UserID         user_id           user_id
```

The expected values in your new cases are exactly what the framework's other implementation already produces, which is about as clear as it gets that this side was simply wrong.

The four added cases cover the positions that matter — trailing, after a leading acronym, mid-word, and between words. Approving.

One follow-up, not for this PR: the two implementations still diverge on a digit followed by consecutive capitals — `Order2FA` gives `order2_fa` under cassandra's and `order2_f_a` under this one. Neither is covered by a test. Two snake-case functions in one framework that disagree is worth an issue on its own; this PR should not grow to absorb it.

aryanmehrotra:
Reviewed this end to end. Your branch targeted `development` but was cut from the repository's initial commit, so GitHub showed it as conflicting with ~215k lines of unrelated files. I have replayed your commit, unchanged and still authored by you, onto current `development` — that part carries no opinion of mine.

On top of it sits one commit of mine, `fix(crud): treat a digit as ending a word`. That one **is** a design call on your function, explained below. Revert it and say so if you disagree; the rebuild underneath stands on its own.

The diagnosis is exactly right, and MySQL states the case better than any test could. Driving all four generated CRUD handlers against a real GoFr app and a real database, with `type oauth2Client struct { Address2, X509Cert, Order2FA, Sha256Sum string }`:

```
development:   POST /oauth2client
               500 {"error":{"message":"Error 1146 (42S02): Table 'test.oauth_R_client' doesn't exist"}}
               ... and GET, PUT and DELETE all answer the same.

this PR:       201 created
               SELECT address2, x509_cert, order2_fa, sha256_sum FROM oauth2_client
               ->  "flat 2", "cert-pem", "totp", "abc123"
```

`oauth_R_client` in a live error message is the whole bug in one line. Worth putting in the description — it is more persuasive than the table of expected outputs.

### It corrupted more than digits

`char >= 'a'` sent *every* rune below `'a'` into the uppercase branch, which adds 32. Digits are the visible half; the other half is punctuation, and `'_'` (95) becomes DEL (127):

```
development:   toSnakeCase("Field_1")  ->  "field_\x7fQ"
               toSnakeCase("_User")    ->  "\x7f_user"
```

A control character in a SQL identifier. Same root cause, and your fix handles it — it just is not mentioned or covered, so I added both cases to the table.

### The same assumption survives one line lower

The underscore decision uses `rune(str[i-1]) >= 'a'` as a stand-in for "is lowercase" — the identical shorthand the outer guard was just fixed for. A digit is below `'a'`, so it does not read as the end of a word, and that is precisely the divergence the description defers to a separate issue:

| input | core, as submitted | Cassandra's converter |
|---|---|---|
| `Order2FA` | `order2fa` | `order2_fa` |
| `A2B` | `a2b` | `a2_b` |

I ran both implementations over a 23-identifier corpus. As submitted they agree on 21; with a digit counted as ending a word they agree on **all 23**, so an entity maps to the same column name whether it is backed by SQL or Cassandra. The asymmetry matters and is preserved: a digit only ends a word on the *left*, so `AB2` stays `ab2`, which is what the regex `([a-z0-9])([A-Z])` does too.

That closes the follow-up rather than filing it, and it is the same argument your description already makes for why it is safe: these names are corrupted today, so nothing can be depending on them.

`isLower` and `isLowerOrDigit` are spelled out rather than reusing `>= 'a'`, because that shorthand is true for every rune above `'a'` — including every non-ASCII one — and it is what caused this bug in two places.

### On the blast-radius argument

Checked, and it holds. Identifiers are quoted by `bind.go`, so a mangled name only ever matched a literal column of that exact mixed-case spelling; there is no column-name override or struct tag; and an entity with a digit in a field name fails outright at every endpoint rather than silently returning wrong data. The e2e above is that argument in practice — on `development` all four verbs return 500 and nothing is written.

### Gates

| Check | Result |
|---|---|
| `gofmt -l` on changed files | clean |
| `go build ./...` / `go vet ./pkg/gofr/...` | pass |
| `typos` | clean |
| `go test -short ./pkg/gofr/...` | pass |
| `toSnakeCase`, `isLower`, `isLowerOrDigit` coverage | 100% |
| e2e vs a real GoFr app + real MySQL | 7/7, all four CRUD verbs |

One note on mechanics: this targets `development` but branched from the repository's initial commit, so GitHub shows it as conflicting and attributes ~215k lines of unrelated files to it. The branch is now your commit replayed on current `development`, which is why the diff is two files.

### On "this changes generated SQL"

Your description argues that no working setup can depend on the old output. I wanted that measured rather than reasoned, so I ran `development`'s `toSnakeCase` against this branch's over every 1-, 2- and 3-word combination of 40 realistic entity words — `User`, `ID`, `HTTP`, `CreatedAt`, `URL`, `UUID`, `ACL` and so on:

```
digit-free identifiers tested : 65640
of those, output CHANGED      : 0
```

Nothing without a digit or an underscore moves, and it cannot: the rule I added only fires when the *previous* character is a digit. Every name that does change is one where `development` already emits garbage — `sha_RUV_hash`, `xUPY_cert`, `field_\x7fQ` — which is the point your description makes, now with a number behind it.

aryanmehrotra:
Second review pass over this branch, including my own commit on top of yours. The fix is right and the end-to-end result holds — but **I overstated two things in [my earlier comment](https://github.com/gofr-dev/gofr/pull/3676#issuecomment-5549138431) and in the description**, and one of them was the reassurance about breakage, so it gets corrected first.

### 1. "Nothing without a digit or an underscore moves, **and it cannot**"

The measurement behind that was ASCII-only, and the absolute claim isn't true. Non-ASCII identifiers are legal Go, and 120 of them in a 280-identifier corpus change:

```
"CaféB"  before: café_b   after: caféb
"ABé"    before: a_bé     after: abé
"日本語A"  before: 日本語_a   after: 日本語a
```

The mechanism is the same bug the PR is fixing, one step further out: the old `rune(str[i-1]) >= 'a'` test read a **UTF-8 continuation byte** (`0xA9` = 169, comfortably above `'a'`) as a lowercase letter and inserted a word boundary that isn't there. `isLowerOrDigit` rejects it correctly.

So the new output is the *right* one in all 120 cases, and this is another instance of the fix rather than a regression. But "and it cannot" was wrong and I shouldn't have written it. **The correct claim is: no ASCII identifier without a digit or an underscore changes** — 0 out of the 143,364 letters-only identifiers of length 1-3.

### 2. "An entity maps to the same column name whichever backend it is on, and no follow-up issue is needed"

Wrong on both halves.

**There is a fourth converter, and it disagrees on exactly the cases this branch just standardised.** `pkg/gofr/datasource/scylladb/internal.go:212` uses `github.com/stoewer/go-strcase`:

| | this branch | Cassandra + `datasource/sql` | **ScyllaDB** |
|---|---|---|---|
| `Order2FA` | `order2_fa` | `order2_fa` | **`order2fa`** |
| `S3Bucket` | `s3_bucket` | `s3_bucket` | **`s3bucket`** |
| `Base64Encoder` | `base64_encoder` | `base64_encoder` | **`base64encoder`** |

**And non-ASCII still diverges from the regex converters**, because they end in a Unicode-aware `strings.ToLower` and this loop only touches `A`–`Z`. `"AÉ"` → `aÉ` here, `aé` there — a capital letter surviving into a generated column name, which is the one thing snake_case exists to prevent. The test case `{"Ünicode" → "Ünicode"}` pins that divergence in place, which is fine as documentation of current behaviour but shouldn't be read as agreement.

A follow-up **is** warranted, covering ScyllaDB and the Unicode case.

### 3. I cited the far sibling and missed the near one

I pointed at the Cassandra converter. The nearer one is `ToSnakeCase` in `pkg/gofr/datasource/sql/db.go:362` — same algorithm, used by `DB.Select` at `db.go:333`, and **already imported by `crud_helpers.go`**. Which means the real story is stronger than the one I told: before this change, `db.Select` and the generated CRUD handlers disagreed about the column names of the same struct *inside the same datasource*.

Measured against it over every 1-, 2- and 3-character identifier in `[a-zA-Z0-9_]` (254,079 inputs):

| | disagreements |
|---|---|
| `development` | 110,715 |
| your first commit | 26,260 |
| **this branch** | **0** |

That is the number the digit commit deserves; "21/23 → 23/23" undersold it considerably.

For anyone wondering whether to just call `sql.ToSnakeCase` and delete this loop: don't. Benchmarked over the same 8 identifiers per op: the loop allocates **152 B in 8 allocations**, the regex version **1,671 B in 79** -- about an order of magnitude more of each, on a path that runs per request and per field. Timings on this machine ranged 353-698 ns/op for the loop against 8,599-13,846 for the regex, so the gap is roughly 12-18x, but treat the ratio as approximate and the allocation counts as the solid number.

### Pushed in `12b75b88b`

- **`builder.Grow`.** The function computed `length` and never used it to size the builder, on a per-request path. Benchmarked over 8 identifiers per op, allocations go **16 → 8** and bytes **208 → 152**. (Wall time on this machine is too noisy to quote; the allocation counts are deterministic, so those are the claim.)
- **Removed a dead guard.** In `(i != 0 || i == length-1)`, the second half can only differ when `i == 0 == length-1`, and both `prevStartsWord` and `nextStartsWord` are false there — so the conjunction is false either way. Verified: no output changes across 30,941 inputs covering every string up to four runes over an alphabet including digits, underscore and non-ASCII.
- **Three boundary test cases.** Each helper turns on comparisons one character wide that could be moved with every existing test still passing. `{"ABé" → "abé"}` is the one that matters: the `r <= 'z'` half of `isLower` is the entire reason it's written out rather than left as `r >= 'a'`, and only a non-ASCII rune directly after a capital reaches it. Also `{"A0B" → "a0_b"}` and `{"X509CERT" → "x509_cert"}` for the two ends of the digit range. All three fail if the corresponding bound is changed; none did before.
- Doc comment scoped to ASCII, and pointed at `datasource/sql.ToSnakeCase`.

### Unchanged and re-verified

Live MySQL end to end, four entities with digits in type and field names: **12/12 on this branch, 0/12 on `development`** with four distinct `Error 1146`. Worth noting one of them — `R3676HTTPServer` has no digit in any *field* name and still breaks on `development`, because the digit is in the type name. `gofmt`, `vet`, `build`, `golangci-lint --new-from-rev` **0 issues**, `typos` clean, `go test -short ./pkg/gofr/...` passes.

aryanmehrotra:
Reviewed this end to end and pushed onto the branch. Two things before the detail: your branch targeted `development` but was cut from the repo's initial commit, so GitHub showed it conflicting with ~215k phantom lines — I cherry-picked your commit onto current `development`, authorship intact. And **one commit on top is a design call on your function**, `fix(crud): treat a digit as ending a word`. Revert it and say so if you disagree; the rest stands without it.

## The bug, against a real database

A GoFr app with entities whose names carry digits, on live MySQL, driving all four generated CRUD handlers over HTTP:

```
this branch : 12 passed, 0 failed
development :  0 passed, 12 failed
  500 Error 1146 (42S02): Table 'test.rSVWV_field_Q' doesn't exist
  500 Error 1146 (42S02): Table 'test.rSVWVo_auth_R_client' doesn't exist
  500 Error 1146 (42S02): Table 'test.rSVWV_order_Rfa' doesn't exist
  500 Error 1146 (42S02): Table 'test.rSVWVhttp_server' doesn't exist
```

The last one has **no digit in any field name** — the digit is in the type name, so the bug is wider than the field-level framing suggests. `'_'` (95) became DEL (127): `Field_1` produced `field_\x7fQ`, a control character inside a SQL identifier.

## Why the digit commit

GoFr has **four** snake-case converters. The nearest is `ToSnakeCase` in `pkg/gofr/datasource/sql/db.go:362`, used by `DB.Select` to map result columns — and already imported by `crud_helpers.go`. So before this branch, `db.Select` and the generated CRUD handlers disagreed about the column names of the same struct, inside the same datasource.

Measured against it over every 1-, 2- and 3-character identifier in `[a-zA-Z0-9_]` (254,079 inputs):

| | disagreements |
|---|---|
| `development` | 110,715 |
| your first commit | 26,260 |
| **this branch** | **0** |

`Test_toSnakeCase_MatchesSQLConverter` now pins that agreement so the two cannot drift again — verified it can fail (reverting the digit rule gives `toSnakeCase("a0A") = "a0a", sql.ToSnakeCase = "a0_a"`).

## Is it breaking?

**No ASCII identifier without a digit or an underscore changes** — 0 of 143,364 letters-only identifiers.

I first wrote "and it cannot", which was wrong: 120 non-ASCII identifiers do change, e.g. `CaféB` from `café_b` to `caféb`. That is the same bug one step out — the old `rune(str[i-1]) >= 'a'` read a UTF-8 continuation byte (`0xA9` = 169) as lowercase. The new output is the correct one in all 120.

Every changed output moves from a value that never matched the canonical converter to one that always does. Nothing goes sane → different-sane.

⚠️ **Worth a release note:** entity fields containing digits or underscores now map to the documented snake_case column. There is no column-name override to opt out with — `TableNameOverrider` covers only the table.

## Also fixed while here

- **`builder.Grow`.** The function computed `length` and never used it, on a path that runs per request per field (twice per field on create, once on update). Allocations 16 → 8, bytes 208 → 152.
  I first justified the headroom with "`OAuth2Client` costs a regrow". Measured, it does not — `Grow` rounds to a size class, so `Grow(12)` takes 16 and 14 fits. The case it rescues is a name whose length lands *exactly* on a size class, like `IsActive` (8 in, 9 out). Constant kept, reason rewritten.
- **A dead guard.** In `(i != 0 || i == length-1)` the second half can only differ when `i == 0 == length-1`, where both word tests are false anyway. No output change across 30,941 inputs.
- **Five boundary cases.** I added three and claimed they pinned the bounds. I then moved all five comparisons and re-ran: `isLower`'s `'a'` and `'z'` ends both **survived**, as did the byte-vs-rune length. `AzB`, `AaB` and `éAb` close them — five for five now.
- A comment of mine said the byte read for `ABé` is a UTF-8 **continuation** byte. It is the **lead** byte, `0xC3`; continuation bytes reach the same wrong answer through the other neighbour. Corrected.

## Known gap, stated rather than left to be found

Non-ASCII still diverges from the regex converters, which finish with a Unicode-aware `ToLower` while this loop only touches `A`–`Z`. An entity field `Ünicode` writes column `"Ünicode"` while `DB.Select` looks up `ünicode` and **silently leaves the field zero-valued**. And ScyllaDB is a fourth implementation (`scylladb/internal.go:212`, `go-strcase`) that splits digits the other way: `Order2FA` → `order2fa`. Both pre-existing; both deserve a follow-up issue, which I originally said was not needed.

## Gates

`gofmt` · `go vet` · `go build ./...` · `go test -race -count=2` · `golangci-lint --new-from-rev` **0 issues** · `typos` clean. `toSnakeCase`, `isLower`, `isLowerOrDigit` at 100%.

Umang01-hash (approved):
APPROVE — verified end-to-end at head 4f9233f.

**Bug is real and impactful.** The old `char >= 'a'` guard sent digits/punct into the +32 lowercasing branch, so a field like `Line1` produced column `line_Q`. Since CRUD writes columns via this `toSnakeCase` while `DB.Select` reads them via `sql.ToSnakeCase`, any digit-bearing field had mismatched write/read column names. Confirmed live on MySQL 8.0: on `development` `POST /gadget` → `500 Unknown column 'line_Q'`; with this PR → `201` + exact `GET` round-trip.

**Verified:**
- Parity with the regex converters (sql + cassandra, identical regexes): repo's own test green + exhaustive sweep of 11,110 ASCII strings → 0 divergence.
- Backward-compat: 66,429 letter-only names, 0 column-name changes vs old — existing schemas unaffected; only previously-corrupted digit names change.
- Tests non-vacuous: revert-red proven for both the guard and the digit-left rule.
- gofmt/vet clean, golangci-lint --new-from-rev 0 issues, full pkg/gofr suite ok. (Pre-existing `TestContext_WriteMessageToSocket` race is unrelated — reproduces on clean development.)

Nit: the `snakeCaseHeadroom` doc comment is quite long for a 4-byte Grow hint; optional to trim.

PR #3677, opened 2026-07-01
fix(sql): validate unsigned int not-null fields without panicking

umarahad2005:
## Description

`validateNotNull` (`pkg/gofr/datasource/sql/query_builder.go`) is reached from `InsertQuery`, which powers GoFr's auto-generated CRUD create handler. Two reflection calls in it panic on ordinary field types:

- `reflect.Value.Int()` panics on unsigned kinds, and `uint, uint8, uint16, uint32, uint64` were routed into `validateIntNotNull`.
- `reflect.Value.IsNil()` panics on kinds it is not defined for, and everything that was not a string, int or float fell through to `validateDefaultNotNull`, which called it unguarded. (Its documentation names six kinds; the implementation accepts `UnsafePointer` too, so seven -- checked against the go1.26.3 source, and the reason for the last commit on this branch.)

So an entity like this panics the create handler for **any** value, not just zero:

```go
type Item struct {
    ID     int  `sql:"auto_increment"`
    Count  uint `sql:"not_null"`
    Active bool `sql:"not_null"`
}
```

`bool` and `time.Time` are as ordinary as columns get, and both crashed. The panic is recovered by GoFr's middleware, so the caller sees a bare `500 Internal Server Error` with nothing pointing at the cause — it reads like a database fault. Verified against a real app and real MySQL:

```
before:  POST /item {"name":"widget","views":7,"active":false}
         500 {"error":{"message":"Internal Server Error"}}     (and no row is written)

after:   201 {"data":"item successfully created with id: 1"}
         SELECT name, views, active FROM item  ->  widget, 7, false
```

## Fix

- `validateUintNotNull`, reading through `reflect.Value.Uint()`, mirroring the existing float helper.
- `validateDefaultNotNull` now treats an untyped nil as null and only calls `IsNil` for the kinds where it is defined (`isNillableKind`). Non-nillable kinds can never be null, so they are accepted.
- Dispatch moved from the concrete type to `reflect.Kind`. A type switch matches only the predeclared types, so a named type — `type Count uint`, ordinary in an entity struct — missed every case and landed in the default path, where it was silently accepted even when zero. On the kind, `Count` and `uint` are the same thing and get the same answer.

No tracking issue exists; this was found while reading the not-null validation.

## Breaking Changes

None. Every value listed here previously panicked. They now validate consistently: zero → `field cannot be zero`, empty → `field cannot be empty`, untyped nil or a nil slice/map/pointer → `field cannot be null`, everything else → accepted.

## Additional Information

- No new dependencies.
- Every validator in the file, and `isNillableKind`, are at 100% statement coverage. On `development`, `validateStringNotNull`, `validateIntNotNull`, `validateFloatNotNull` and `validateDefaultNotNull` were **all four** at 66.7%. Package coverage goes 92.7% → 93.6-93.8% (it is not stable to the decimal across runs, so the range is the honest figure). Earlier revisions of this description credited only `validateFloatNotNull`, then said "three of them"; both under-reported it.
- Verified end to end against a real GoFr app and real MySQL, driving the generated CRUD handlers over HTTP: a valid record inserts and reads back, `views: 0` returns `field cannot be zero: views`, `4294967295` is accepted, and the list handler returns the rows.

**Separate, pre-existing:** these validation errors answer `500`, not `400` — `errFieldCannotBeZero` and friends are plain `errors.New` values with no `StatusCode()`. An empty required string, which never panicked, answers 500 on `development` too, so it is not a regression from this change and is left for its own fix.

## Checklist


pkg/gofr/datasource/sql/query_builder.go
```diff
@@ -96,8 +96,10 @@ func validateNotNull(fieldName string, value any, isNotNull bool) error {
...
 	case string:
 		return validateStringNotNull(fieldName, v)
-	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64:
+	case int, int8, int16, int32, int64:
 		return validateIntNotNull(fieldName, v)
+	case uint, uint8, uint16, uint32, uint64:
+		return validateUintNotNull(fieldName, v)
 	case float32, float64:
 		return validateFloatNotNull(fieldName, v)
 	default:
```
Umang01-hash, line 114: Same bug class as the uint fix: this `default` branch routes to `validateDefaultNotNull`, which calls `reflect.Value.IsNil()` — that panics on non-nillable kinds. A NotNull `bool`, `time.Time`, or `uintptr` field will panic here (verified). Since you're already fixing the unsigned case, please guard this path too (check `Kind()` before `IsNil()`, or handle bool/time explicitly) and add a test.

Umang01-hash (dismissed):
Fix is correct and well-tested. One request before merge: the same panic class you're fixing still exists on the `default` path — worth closing it in the same PR while you're here. Details inline.

umarahad2005:
Thanks for the review, @Umang01-hash — good catch.

**`default` path panic (bool / `time.Time` / `uintptr`):** confirmed and fixed in fbe828c7. `validateDefaultNotNull` now checks the kind before calling `IsNil()` via a small `isNillableKind` helper, so only nillable kinds (chan/func/interface/map/pointer/slice) run the nil check. An untyped-nil interface (`reflect.Invalid`) is treated as null. Non-nillable kinds can never be null, so they're accepted. I verified `bool`, `time.Time`, and `uintptr` panicked before this change and now pass.

**Test coverage (@Copilot):** the unsigned test now exercises every uint width — `uint8`, `uint16`, `uint32`, `uint64` — for both the valid and zero cases, plus a new `Test_validateNotNull_NonNillableKinds` covering the default-path kinds and the nil cases.

Ready for another look.

umarahad2005:
Gentle bump 🙂 — the requested change (guarding the `default` path against the `IsNil()` panic on non-nillable kinds like `bool`/`time.Time`/`uintptr`) was pushed in `fbe828c7`, along with tests for every uint width and the non-nillable kinds. Whenever you have a moment, this should be ready for another look. Thanks!

aryanmehrotra:
Reviewed this end to end. Your branch targeted `development` but was cut from the repository's initial commit, so GitHub showed it as conflicting with ~200k lines of unrelated files and nobody could review it. I have replayed your two commits, unchanged and still authored by you, onto current `development` — that is the bulk of the push and it carries no opinion of mine.

On top of that sits one commit of mine, `fix(sql): dispatch not-null validation on the kind`. That one **is** a design call on your function, explained in full below. If you disagree with it, revert it and say so — the rebuild underneath stands on its own.

This is a bigger fix than the description claims, and the second commit is the more valuable half. Running the generated CRUD create handler against a real GoFr app and real MySQL, with an entity that has an unsigned NOT NULL field:

```
development:            POST /item  {"name":"widget","views":7,"active":false}
                        500 {"error":{"message":"Internal Server Error"}}
                        ... and the table stays empty. Every create fails, valid ones included.

this PR:                201 {"data":"item successfully created with id: 1"}
                        SELECT name, views, active FROM item  ->  widget, 7, false
```

The panic is recovered by the middleware, so the client gets a bare 500 with nothing pointing at the cause. That is the worst shape for this bug: it looks like a database problem.

### The blast radius is wider than unsigned ints

`validateDefaultNotNull` called `reflect.Value.IsNil`, which is defined for six kinds and panics on the rest. Every one of these is an ordinary column type, and on `development` every one crashes the create handler:

| value | `development` | this PR |
|---|---|---|
| `uint(5)` | `PANIC(reflect.Value.Int on uint Value)` | ok |
| `uint(0)` | `PANIC` | `field cannot be zero` |
| `true` / `false` | `PANIC(reflect.Value.IsNil on bool Value)` | ok |
| `time.Time{}` | `PANIC(IsNil on struct Value)` | ok |
| `complex64(0)` | `PANIC(IsNil on complex64 Value)` | ok |
| `nil` | `PANIC(IsNil on zero Value)` | `field cannot be null` |
| `[]byte(nil)` | `field cannot be null` | `field cannot be null` |

A `bool` NOT NULL column is about as common as it gets, and it panicked on `development`. Worth putting that in the description — right now it reads as an unsigned-int edge case.

### One gap the guard left behind

A Go type switch matches only the predeclared types, so a named type misses every case:

```go
type Count uint   // ordinary in an entity struct
```

On `development` `Count(0)` panicked in `IsNil`. With your guard it stops panicking, but it lands in the default path and is *accepted* — so `Count(0)` passes a NOT NULL column that `uint(0)` correctly rejects. Same for `type Age int` and `type Name string`.

I moved the dispatch from the concrete type to `reflect.Kind`. `Count` and `uint` are the same kind, so they get the same answer, and the special case disappears rather than being enumerated:

```
named Count(0):  field cannot be zero      (was: accepted)
named Age(0):    field cannot be zero      (was: accepted)
named Name(""):  field cannot be empty     (was: accepted)
```

`validateStringNotNull` takes `any` now for the same reason — a named string reaches it as one.

### Coverage

Every validator in the file is at 100%. `validateFloatNotNull` was at 66.7% — nothing reached its non-zero branch — so that has a test too.

```
validateNotNull        100.0%
validateStringNotNull  100.0%
validateIntNotNull     100.0%
validateUintNotNull    100.0%
validateFloatNotNull   100.0%   (was 66.7%)
validateDefaultNotNull 100.0%
isNillableKind         100.0%
```

### Not this PR's, but worth knowing

**These validation errors answer 500, not 400.** `errFieldCannotBeZero` and friends are plain `errors.New` values with no `StatusCode()`, so GoFr's responder defaults to 500. A client sending a bad value gets told the server is broken. It is pre-existing — an empty required string, which never panicked, answers 500 on `development` too:

```
INFO  empty required string -> 500 {"error":{"message":"field cannot be empty: name"}}
```

Giving these errors a `StatusCode() int` returning 400 would be a small, contained change, but it is a separate one.

### Gates

| Check | Result |
|---|---|
| `gofmt -l` on changed files | clean |
| `go build ./...` / `go vet ./pkg/gofr/...` | pass |
| `typos` | clean |
| `go test -short ./pkg/gofr/...` | pass |
| `go test -short -race ./pkg/gofr/datasource/sql/` | pass, no races |
| e2e vs a real GoFr app + real MySQL | 6/6 |

`Test_WebSocket_Success` trips the race detector in `pkg/gofr`; it does so on clean `development` too, and is unrelated.

One note on mechanics: this targets `development` but branched from the repository's initial commit, so GitHub shows it as conflicting and attributes ~200k lines of unrelated files to it. The branch is now your two commits replayed on current `development`, which is why the diff is two files.

aryanmehrotra:
Second review pass over this branch, including my own commit. The change holds up — an independent 76-value type matrix panics **42 times on `development` and 0 times here**, and 10 of 10 mutants against the new code are killed. Three corrections to [my earlier comment](https://github.com/gofr-dev/gofr/pull/3677#issuecomment-5549118193), and one small fix pushed.

### 1. I wrote that `Count(0)` "passes a NOT NULL column that `uint(0)` correctly rejects"

The word to strike is **correctly**. SQL `NOT NULL` forbids NULL, not zero — `views INT UNSIGNED NOT NULL DEFAULT 0` is an ordinary column, and `0` is a legitimate value for it. GoFr's `not_null` tag has always meant "reject the zero value", which is a different and narrower rule, and this branch is what extends that rule to *every named type*.

I still think the change is right: a rule applied to `uint` but not to `type Count uint` is worse than the same rule applied consistently, and consistency is the appropriate scope for a bug fix. But endorsing the underlying rule as *correct* was wrong, and it is worth being explicit that the rule is currently inconsistent in the other direction too — `uint(0)` and `""` are rejected, while `false`, `time.Time{}`, `[16]byte{}` and a `*uint` pointing at zero are all accepted. Whether `not_null` should mean "not null" is a design question for its own issue.

### 2. The coverage note under-reported this PR

I wrote that `validateFloatNotNull` was at 66.7% and implied the rest were already at 100%. On `development`, **`validateStringNotNull`, `validateIntNotNull`, `validateFloatNotNull` and `validateDefaultNotNull` were all at 66.7%** — three of them are raised to 100% here, not one, and package coverage goes 92.5% → 93.6%. Description corrected.

### 3. Fixed in `30aba2916` — `isNillableKind` was missing `reflect.UnsafePointer`

`reflect.Value.IsNil`'s documentation lists six kinds and the helper listed exactly those. The implementation also accepts `UnsafePointer`, so a nil one fell through to the default path and was **accepted** for a NOT NULL column instead of reported as null. Checked against the runtime rather than the doc:

```go
var p unsafe.Pointer
reflect.ValueOf(p).IsNil()   // true, no panic
```

Added to the case list, with a test for the nil and non-nil forms. Negligible in practice, but this helper exists precisely to be the authority on which kinds are safe, so it should be right.

### Two gaps this branch doesn't close, worth their own issues

- **`sql.NullString{Valid: false}` passes a NOT NULL check.** It is a struct, so it reaches the default path and is accepted. The canonical Go way of saying "this is NULL" sails straight through the check built to catch NULLs. It panicked on `development`, so this is not a regression — this branch is what makes it reachable. Same for any `driver.Valuer`.
- **The update path has no not-null validation at all.** `(*entity).Update` in `pkg/gofr/crud_handlers.go:303` calls `sql.UpdateByQuery` with no `validateNotNull` anywhere in it. So the rule this PR is fixing is enforced on POST and silently absent on PUT. My earlier comment flagged the smaller out-of-scope item — validation errors returning 500 rather than 400 — and walked past this larger one.

### Re-verified

76-value matrix (nil, bool, `time.Time`, every signed and unsigned width, floats, complex, uintptr, arrays, chans, funcs, named types over each, `driver.Valuer`, `sql.NullString`, `sql.NullInt64`, `json.RawMessage`): **0 panics here, 42 on `development`.** No behaviour regressions — every difference from `development` started as a panic. Live MySQL: `development` 500s on every create and leaves the table empty; this branch inserts and reads back. `gofmt`, `vet`, `build` clean; `golangci-lint --new-from-rev` **0 issues**; `go test -race -count=2` ok at 93.6%. Both `//nolint:exhaustive` directives confirmed load-bearing — removing either produces a real error, and `nolintlint` runs with `allow-unused: false`.

aryanmehrotra:
Reviewed this end to end and pushed onto the branch. Your branch targeted `development` but was cut from the repo's initial commit, so GitHub showed it conflicting with ~215k phantom lines — I cherry-picked your two commits onto current `development`, authorship intact, and added two of my own.

## The bug, against a real database

A GoFr app with `not_null` fields over ordinary Go types, on live MySQL:

- `development`: **every** create returns `500 Internal Server Error`, table stays empty (0 rows)
- this branch: valid values insert and read back; zero values are rejected with a message

Enumerating the **full `reflect.Kind` domain** — 44 values across all 27 kinds, old versus new with panics captured:

- **22 rows changed. Every one is `PANIC` → a defined answer.**
- **0 rows where `development` returned a non-panic answer and this branch differs.**

No panic path remains: `Int()`/`Uint()`/`Float()`/`String()` are each kind-guarded, `IsNil` is guarded by `isNillableKind`, and `Invalid` is handled first.

## Why the dispatch moved to `reflect.Kind`

Your guard stopped the panics, but a Go type switch matches only the *predeclared* types — so `type Count uint`, which is ordinary in an entity struct, missed every case, fell to the default, and was **silently accepted at zero**. `Count(0)` passed a NOT NULL column that `uint(0)` rejects. Dispatching on the kind makes the two answer the same.

Worth being precise about, because "it used to panic" and "it used to be accepted" call for different fixes:

| value | development | your commit | this branch |
|---|---|---|---|
| `Count(0)` | PANIC | accepted | `field cannot be zero` |
| `Name("")` | PANIC | accepted | `field cannot be empty` |

## `isNillableKind` was one kind short

`reflect.Value.IsNil`'s *documentation* names six kinds; the implementation accepts `UnsafePointer` too. So a nil one reached the default path and was accepted for a NOT NULL column. Checked against the runtime rather than the doc — `reflect.ValueOf(p).IsNil()` on a nil `unsafe.Pointer` returns `true` and does not panic.

## Test gaps found by mutation, now closed

Everything passed at first, so I mutated the production code. Four survivors, all now killed:

- **`Int() == 0` → `<= 0` survived.** No negative value was tested anywhere. That mutation reads like a plausible "must be positive" rule and would silently reject every negative value in a column entitled to hold one.
- `int8`/`int16`/`int32` were never exercised — only plain `int` — so dropping any of them from the dispatch went unnoticed. (Every `uint` width *was* covered; the asymmetry was easy to miss.)
- `reflect.Func` in `isNillableKind` had no test.

Coverage 92.7% → 93.6–93.8% (it is not stable to the decimal across runs, so the range is the honest figure). All four validators were at 66.7% on `development`, not one; every validator and `isNillableKind` are at 100% now.

## Things I checked and deliberately did not change

- The redundant second `reflect.ValueOf` inside each validator: benchmarked against a variant that computes it once and passes the `reflect.Value` down — **29.2 ns versus 29.4 ns, 0 allocations either way**. Not worth the refactor.
- `reflect.Interface` in `isNillableKind` is unreachable (`reflect.ValueOf(any)` unwraps interfaces) but harmless as a faithful mirror of `IsNil`'s contract.

## Open, not fixed here

- `sql.NullString{Valid: false}` **passes** a NOT NULL check — the canonical Go way to say NULL, sailing through the NULL check. It panicked on `development`, so this branch is what makes it reachable.
- `(*entity).Update` (`crud_handlers.go:303`) has **no not-null validation at all**. Enforced on POST, silently absent on PUT.
- GoFr's `not_null` means "reject the zero value", which is not what SQL `NOT NULL` means — `views INT UNSIGNED NOT NULL` legitimately stores 0. Extending that rule consistently to named types is right for a bug fix, but the rule itself deserves its own issue. (I originally called the old behaviour "correct"; it isn't.)
- Validation errors return 500 rather than 400.

## Gates

`gofmt` · `go vet` · `go build ./...` · `go test -race -count=2` · `golangci-lint --new-from-rev` **0 issues**. Both `//nolint:exhaustive` directives confirmed load-bearing — removing either produces a real error, and `nolintlint` runs with `allow-unused: false`. No exported symbol changes (`apidiff` empty).

aryanmehrotra (approved):
Verified locally against go1.26.3.

**Correctness**
- `reflect.Value.Int()` panics on the unsigned kinds and `reflect.Value.IsNil()` panics outside seven kinds — both confirmed in `reflect/value.go` at the pinned version. The bug and the fix are real.
- `isNillableKind` matches `IsNil`'s implementation exactly: Chan, Func, Map, Pointer, UnsafePointer, Interface, Slice. The note about the doc comment listing six while the implementation accepts `UnsafePointer` is accurate.
- Moving dispatch from the concrete type to `reflect.Kind` is the right call — a `type Count uint` field missed every arm of the type switch before and was silently accepted for a NOT NULL column.

**Evidence**
- `go test ./pkg/gofr/datasource/sql/...` passes.
- Package coverage 92.5% → 93.6%; all six validators and `isNillableKind` at 100% statement coverage.
- `golangci-lint run ./pkg/gofr/datasource/sql/...` reports 0 issues.

**Non-blocking follow-ups, not for this PR**
1. `validateNotNull` calls `reflect.ValueOf(value)` for the kind and each helper calls it again — passing the `reflect.Value` down would halve the reflection on a per-field, per-insert path.
2. The `reflect.Interface` arm of `isNillableKind` is unreachable, since `reflect.ValueOf` on an `any` unwraps the interface. Harmless, but dead.
3. The zero-means-missing rule is now visibly inconsistent: `uint(0)` is rejected while `false` and `time.Time{}` are accepted. That is inherited from the existing int/float/string behaviour, so matching it here is the consistent choice — but the rule itself, and the 500-instead-of-400 status you already called out, deserve their own issue.

Thanks for the thorough description and for closing the `default`-path panic in the same PR.

Umang01-hash (approved):
LGTM. Verified at head 5bd3ca3 — reproduced the panic through the real path (InsertQuery with uint(5) NOT NULL on development: `reflect.Value.Int on uint Value`), and the fix resolves it. Nice catch that it's broader than the title: the old default path called IsNil() unconditionally, so bool, time.Time, named types and untyped nil all panicked too — dispatching on reflect.Kind fixes the whole class (and is the only way to handle named types like `type Count uint`).

Tests are genuinely mutation-killing — I confirmed fail-on-revert for routing uint back through Int() (panics), `== 0`→`<= 0` (negative cases go red), and dropping UnsafePointer from isNillableKind (nil unsafe.Pointer goes red). Package builds/vets clean and golangci-lint --new-from-rev is 0 issues. No exported API change. No blocking findings.

One note for the record (no action needed): including reflect.UnsafePointer in isNillableKind is correct but leans on an undocumented reflect.IsNil detail — good that a test pins it.

PR #3704, opened 2026-07-13
feat(s3): add signed URL support with S3-compatible provider flavors

akshat-kumar-singhal:
**Description:**

Adds **signed (presigned) URL** generation to the S3 file store, plus an extensible **flavor** system so the same store works against S3-compatible providers.

Previously only GCS implemented `CloudFileSystem.GenerateSignedURL`; the S3 store returned `ErrSignedURLsNotSupported`. This makes S3 a first-class `CloudFileSystem`.

- **Signed URLs** — `GenerateSignedURL(ctx, name, expiry, opts)` on the S3 file store, mirroring the GCS public contract. Signed locally with AWS Signature Version 4 via `aws-sdk-go-v2` presigning (offline; **no new dependency** — `service/internal/presigned-url` is already an indirect dep of `service/s3`). `opts.ContentType`/`ContentDisposition` map to `response-content-*` overrides; disposition is sanitized against CRLF header injection.
- **CloudFileSystem** — `New` now returns `file.CloudFileSystem` (consistent with the GCS constructor), and `CreateWithOptions` is implemented so `file.AsCloud()` works.
- **Flavors** — `Config.Flavor` applies provider-correct defaults for addressing style, signing region, and upload-checksum handling:

  | Flavor | Provider |
  |--------|----------|
  | `FlavorAWS` (default) | Amazon S3 |
  | `FlavorR2` | Cloudflare R2 (signs region `auto`, disables SDK upload checksum) |
  | `FlavorMinIO` | MinIO (path-style, disables SDK upload checksum) |
  | `FlavorSpaces` | DigitalOcean Spaces |
  | `FlavorB2` | Backblaze B2 |
  | `FlavorGeneric` | any other S3-compatible endpoint + `Config` overrides |

  Adding a provider is one `case` in `profileFor` plus a constant; unknown providers use `FlavorGeneric` with explicit overrides. `Config` also gains `SessionToken` (STS / scoped R2 tokens) and `UsePathStyle *bool`.

Motivation: signed URLs are one of the most-requested S3 capabilities, and R2/MinIO/Spaces/B2 support broadens GoFr's object-storage reach without touching framework core.

- None. `s3.New`'s return type widens from `file.FileSystemProvider` to `file.CloudFileSystem` (the latter embeds the former), so existing callers compile unchanged. `FlavorAWS` keeps path-style addressing, preserving prior behavior.
- Scope: GET/download presigned URLs only. PUT/upload presigning needs an operation field on `file.FileOptions` (framework core) and will follow as a separate PR.

**Additional Information:**

- Uses the already-present `aws-sdk-go-v2` presign client — no new module dependency.
- `go test -race -cover` passes at **88.4%** for the s3 submodule; `go vet` and `golangci-lint` are clean.
- Docs updated in `docs/advanced-guide/handling-file/page.md`: signed-URL support now lists S3, a new **S3-Compatible Providers** section with a flavor table and R2/Generic examples, and provider-specific signing notes.

**Checklist:**


pkg/gofr/datasource/file/s3/fs.go
```diff
@@ -103,12 +131,20 @@ func (f *FileSystem) Connect() {
...
 		func(o *s3.Options) {
-			o.UsePathStyle = true
+			o.UsePathStyle = prof.usePathStyle
 			o.BaseEndpoint = &f.config.EndPoint
+
+			// Several S3-compatible providers reject the CRC32 checksum that
+			// aws-sdk-go-v2 adds to uploads by default; only calculate it when the
+			// request actually requires one.
+			if prof.disableUploadChecksum {
+				o.RequestChecksumCalculation = aws.RequestChecksumCalculationWhenRequired
```
Umang01-hash: The flavor→SDK wiring here (checksum-disable, session token, path-style) has no unit assertion — only `resolveProfile` is tested, so these could be dropped without a red test. I confirmed it works live against MinIO, so not a bug — but worth one integration-style assertion.

pkg/gofr/datasource/file/s3/create_options.go
```diff
@@ -0,0 +1,98 @@
...
+	contentType := mime.TypeByExtension(path.Ext(name))
+	// Default to attachment so the file must be downloaded before being opened,
+	// matching Create's behavior.
+	contentDisposition := defaultContentDisposition
+
+	var metadata map[string]string
+
+	if opts != nil {
+		if opts.ContentType != "" {
+			contentType = opts.ContentType
```
Umang01-hash, line 54: `ContentType` isn't validated here, while the signed-URL path runs `validateSignedURLInput`. Minor asymmetry — consider validating for parity (low risk since the SDK rejects bad header bytes).

Umang01-hash (commented):
Reviewed at head `0bdc8ef`. Solid feature — verified E2E against live MinIO: `FlavorMinIO` connect + `CreateWithOptions` upload + `GenerateSignedURL` → HTTP 200 with correct content-type/disposition, no client creds. Build/vet/lint clean, coverage 88.4%, no breaking change (`CloudFileSystem` embeds `FileSystemProvider`, confirmed via `AsCloud` at runtime).

A few non-blocking asks below. Scope is mildly broad (signed URLs + flavors + `CreateWithOptions` bundled) — fine to keep, just flagging.

pkg/gofr/datasource/file/s3/fs.go
akshat-kumar-singhal: Good catch — added ​`Test_applyClientOptions` in `fs_test.go`. I extracted the flavor→`s3.Options` closure out of `Connect` into a pure `applyClientOptions(prof, endpoint)` and the table-driven test applies it to a bare `*s3.Options`, asserting `UsePathStyle`, `BaseEndpoint`, and `RequestChecksumCalculation` per flavor (AWS default, R2 checksum-disable, Spaces virtual-hosted, explicit `UsePathStyle` override). So dropping any of that wiring now fails a unit test, not just the live path. Session-token/credentials wiring is still only exercised live — an offline test can't observe the static-credentials provider — matching the existing GCS approach. Fixed in 79bb720.

pkg/gofr/datasource/file/s3/create_options.go
akshat-kumar-singhal: Agreed — fixed for parity. I pulled the `"type/subtype"` check out of `validateSignedURLInput` into a shared `validateContentType`, and `CreateWithOptions` now calls it before any S3 request, so a malformed `ContentType` is rejected up front on both paths. Empty stays valid (falls back to the extension-sniffed type). Covered by `Test_CreateWithOptions_InvalidContentType`. Fixed in 79bb720.

akshat-kumar-singhal:
Thanks for the thorough review and the live MinIO verification, @Umang01-hash. Both non-blocking asks are addressed in 79bb720:

- **Content-Type parity** — extracted a shared `validateContentType`; `CreateWithOptions` now validates before any S3 call, same as the signed-URL path (`Test_CreateWithOptions_InvalidContentType`).
- **Flavor→SDK wiring coverage** — extracted `applyClientOptions(prof, endpoint)` from `Connect` and added table-driven `Test_applyClientOptions` asserting path-style, base endpoint, and upload-checksum per flavor.

No behavior change to `Connect`; coverage 88.4% → 88.7%, vet + race + `golangci-lint --new-from-rev` clean. Ready for another look.

Umang01-hash (approved):

PR #3722, opened 2026-07-14
feat: AI / LLM / MCP support

### LINKED ISSUE #3728: fix(ai): LLM log records are emitted without trace correlation
> **Correction (edited):** the original text of this issue claimed `ai.Log` was the *only* GoFr log lacking trace correlation. That was wrong — the SQL datasource's logs lack it too, for the same underlying reason. The framing below is corrected: this is a systemic consequence of how `instrumentDatasource` injects loggers, and `ai.Log` is its newest instance. The proposed fix is unchanged.

## Summary

The structured LLM log record introduced in #3722 (`ai.Log`) is emitted without any trace correlation, so an LLM log line cannot be joined to the trace, request, or handler it belongs to.

This is a small, self-contained fix rather than a new feature — though it sits on top of a more general gap, described below.

## Root cause

`logging.ContextLogger` already does this job — it stamps `__trace_id__` onto log output when a valid span is in the context:

```go
// pkg/gofr/logging/ctx_logger.go
func NewContextLogger(ctx context.Context, base Logger) *ContextLogger {
	var traceID string

	sc := trace.SpanFromContext(ctx).SpanContext()
	if sc.IsValid() {
		traceID = sc.TraceID().String()
	}

	return &ContextLogger{base: base, traceID: traceID}
}
```

But it is only constructed per-request, in `pkg/gofr/context.go` (lines 169 and 180). Datasource loggers arrive by a different route — `instrumentDatasource` injects the application-level logger at boot:

```go
// pkg/gofr/external_db.go
func (a *App) instrumentDatasource(ds any) {
	if l, ok := ds.(interface{ UseLogger(any) }); ok {
		l.UseLogger(a.Logger())   // base logger, no request context to draw a trace ID from
	}
```

`AddLLM` calls `instrumentDatasource(m)`, so the LLM model gets the base logger. `writeLog` then compounds it by not receiving a context at all:

```go
// pkg/gofr/ai/instrument.go
func writeLog(info *CallInfo, resp *Response, err error, status string) {
```

The resulting log line carries provider, model, operation, status and full token counts — but nothing tying it to the request that caused it. In a service handling concurrent requests, LLM log lines are effectively unattributable.

**This is not specific to the AI package.** Any datasource logging through the duck-typed `UseLogger` path has the same gap — the SQL datasource's query logs carry no trace field either. Redis pub/sub is an exception only because it hand-rolls `CorrelationID: span.SpanContext().TraceID().String()` at each call site rather than relying on the logger.

## Why it matters

- **Debugging.** A slow or failing LLM call appears in the logs with no way to reach the trace, the handler, or the request that triggered it.
- **Cost attribution.** Token counts are right there in the log line, but with no correlation ID there's nothing to join them to. (#3727 proposes a proper programmatic hook for this; this issue is the much smaller, independent piece and doesn't depend on it.)

## Proposed fix

Add the trace ID to `ai.Log` and populate it from the span already in hand. `record()` holds both `ctx` and `span`, so no plumbing is needed beyond passing one down:

```go
type Log struct {
	Provider         string `json:"provider"`
	Model            string `json:"model"`
	Operation        string `json:"operation"`
	Status           string `json:"status"`
	TraceID          string `json:"traceId,omitempty"`
	// ... token fields unchanged
}

func writeLog(span trace.Span, info *CallInfo, resp *Response, err error, status string) {
	// ...
	if sc := span.SpanContext(); sc.IsValid() {
		entry.TraceID = sc.TraceID().String()
	}
```

The `IsValid()` guard matters: with no exporter configured GoFr installs a `NeverSample` provider, and `tracerOf` falls back to a noop tracer when no tracer was injected — the field should be omitted rather than logged as a zero TraceID. Same guard `ContextLogger` already uses; `omitempty` matches the existing tag convention on `TotalTokens`/`CachedTokens`.

### Why set the field rather than wrap with `ContextLogger`

Wrapping in `logging.NewContextLogger` would also work, but it would make `pkg/gofr/ai` depend on `pkg/gofr/logging`. The `ai` package deliberately defines its own minimal `Logger` interface (`Debug`/`Error`) to stay decoupled, and that looks intentional. `ai` already imports `go.opentelemetry.io/otel/trace`, so setting the field costs no new dependency.

`writeLog` is unexported, so the signature change is internal; adding a field to `ai.Log` is additive.

## Scope question for maintainers

The fix above is deliberately narrow — it works for the AI path specifically because `record()` has a span in hand, which datasource logging paths generally don't.

If you'd rather address the general case, that's a different and larger change: either passing context into datasource logging paths, or having `instrumentDatasource` inject something context-aware. Happy to do the narrow one now and leave the general one to a separate issue, or to hold this pending a decision on the broader approach — whichever you prefer.

Happy to send a PR either way.

Refs #3722, #3727.

aryanmehrotra:
# feat: AI / LLM / MCP support

Adds first-class support for talking to LLMs and exposing services to AI agents, extending GoFr's
"batteries-included, zero-config observability" promise to both. Addresses discussions **#3245 (LLM
Observability)** and **#3644 (Supporting Agentic Workloads)**.

## What you can do now

```go
// Call an LLM from a handler — tracing, token metrics, logs and health are automatic.
app.AddLLM(&llm.Client{Provider: llm.Groq, Model: "llama-3.3-70b-versatile"})
resp, _ := ctx.LLM().Generate(ctx, prompt)

// Register additional named models and select one per call.
app.AddLLM(&llm.Client{Provider: llm.OpenAI, Model: "gpt-4o-mini"}, gofr.WithName("fast"))
quick, _ := ctx.LLM("fast").Generate(ctx, prompt)

// Stream tokens to the client as server-sent events.
return response.Stream{Source: stream}, nil

// Expose your read-only handlers to AI agents as MCP tools.
app.EnableMCP()

// Build an agent loop over your own service's endpoints.
tools := ctx.LLM().Tools()
```

An OpenAI-compatible model is also wired zero-config from the environment (like Redis/SQL): set
`LLM_PROVIDER`, `LLM_MODEL` and `LLM_API_KEY` and skip `AddLLM` entirely.

## Design

- **LLM as a datasource.** `AddLLM` rides the existing `instrumentDatasource` duck-typing
  (`UseLogger`/`UseMetrics`/`UseTracer`/`UseConfig`/`Connect`), registers health and metrics, and
  exposes the model via `ctx.LLM()`. No parallel plumbing.
- **Multiple models.** `AddLLM(m, gofr.WithName("fast"))` registers additional named models;
  `ctx.LLM("fast")` selects one and `ctx.LLM()` returns the default. Each model is health-checked
  and metered under its own provider/model labels. The getter signatures are variadic so this stays
  additive.
- **`llm.Client` rides `service.HTTP`.** The OpenAI-compatible client reuses GoFr's instrumented
  HTTP service, so tracing, connection pooling and logging come for free. Retry is condition-aware
  and lives in the client (connection-setup errors and `429`+`Retry-After` only; a timed-out POST is
  never retried; a canceled context aborts). Providers (OpenAI/Groq/DeepSeek/Together/Ollama) are a
  config field; `BaseURL` overrides any OpenAI-compatible endpoint.
- **Streaming is transport, not AI.** `response.Stream` is a generic special-response type
  (SSE/NDJSON) usable for progress or log tailing too. Values are pulled on demand over an unbuffered
  channel (backpressure); the responder flushes via `http.NewResponseController`.
- **MCP dispatches through the real router.** A tool call rebuilds an `*http.Request` and dispatches
  it back through the router, so Bind/validation/RBAC/auth middleware all run. The caller's auth
  headers propagate, so `ctx.GetAuthInfo()` works. The same tools back `ctx.LLM().Tools()`.

## Observability (#3245)

Every LLM call made through `ctx.LLM()` is observable the same way a normal GoFr request is.

| Signal | Detail |
|---|---|
| Metrics | `app_llm_request_count` (provider, model, operation, status); `app_llm_tokens_per_request` histogram — its Prometheus `_sum` is cumulative tokens |
| Traces | span per call (`llm.chat`/`generate`/`stream`), child of the request span, parent of the HTTP span; token/provider/model attributes |
| Logs | structured per-call line carrying the call's trace ID; prompt/completion text never logged |
| Health | model registers as a datasource on the health endpoint |

Low-cardinality by design — prompts/session/run IDs go to traces, never metric labels. The API key
never appears in a log, error, span or health detail.

## v1 safety

Fully **additive** — no existing exported identifier changed. Provider interfaces (`ai.Model`,
`ai.LLM`, `ai.Tools`) are frozen and grow via new optional interfaces; every new call takes
variadic options or a named struct; enums are append-only; metrics register lazily. **Zero new
dependencies in core** — `go.sum` is unchanged; provider SDKs (e.g. a future native Anthropic
client) would live in their own modules.

## Security (MCP)

- Only read-only (`GET`/`HEAD`/`OPTIONS`) handlers are exposed as tools; write handlers are **never**
  exposed, so an agent cannot mutate state through this surface. Enforced at **call time**, not just
  in the listing.
- MCP server binds **`127.0.0.1`** so it is not a second unauthenticated ingress; `MCP_PORT=0`
  disables it.
- Path arguments are escaped so they cannot break out of a route segment; request body capped at
  1 MB; tool-handler panics are contained; framework `/.well-known/*` routes are never exposed.

## Testing & CI (verified locally)

- `golangci-lint` (v2.12.2, `--new-from-rev`): **0 new issues**. `gofmt`/`go vet`/`go build ./...`
  clean. `typos`: clean.
- `go test ./pkg/gofr` (`-short`, CI flags): pass, **no panics**, 86.1% statement coverage.
- New packages pass `go test -race`; streaming and MCP tests cover the concurrency and
  security-critical paths (client-disconnect goroutine-leak, panic containment, call-time access
  control, auth-header propagation, retry classification, no-secret-in-errors).
- Every module was reviewed adversarially before landing; findings (including a production-only
  flush bug behind `StatusResponseWriter`, a provider-error-in-200 mapping, an unauthenticated MCP
  bind, and a tool-whitelist widening bug) were fixed and regression-tested.

## Docs & examples

- Advanced-guide pages **Calling LLMs**, **Building AI Agents** and **Streaming Responses**, wired
  into the nav. Streaming is documented as a general transport (SSE/NDJSON), not an AI feature.
- A combined **`examples/using-ai`** showing LLM calls, streaming, MCP tools and an agent loop in one
  service.

## Design decision: a first-class client, not an HTTP-service decorator

#3245 originally proposed adding LLM observability as a `service.Options` **decorator**
(`observability.LLMConfig` / `ResponseParser`) on `AddHTTPService`, which parses token usage out of
the raw response body of a hand-built LLM call. We evaluated that and deliberately chose a
first-class `ctx.LLM()` client instead. The reasoning, so it is on the record:

- **Observability is only half the goal.** #3644 (agentic workloads) needs streaming, tool calling
  and provider-agnostic swapping — none of which a response-observing decorator can provide, because
  those require GoFr to *own* the call, not just watch it. One coherent `ai.Model` abstraction serves
  both #3245 and #3644; a decorator would serve only #3245 and leave a second, unrelated surface to
  build for agents.
- **The client's instrumentation is more accurate.** The client *knows* the provider, model and
  operation (it built the request) and records them directly. A decorator must *parse* them back out
  of each provider's response body, which is exactly the brittle, provider-specific work the
  `ai.Model` interface removes.
- **It keeps handlers clean.** The decorator leaves every handler hand-rolling the request body, auth
  header and response parsing; `ctx.LLM().Generate(...)` is one call. We did not want the framework to
  be "simpler" by pushing that boilerplate into user code.
- **One door, not two overlapping ones.** Two ways to get LLM token metrics (decorate your HTTP
  service *or* use the client) would split documentation, tests and user mental models. The low-
  friction cases the decorator targeted are already covered another way — zero-config env auto-init
  (`LLM_PROVIDER`/`LLM_MODEL`) and explicit `app.AddLLM` for full control.

**Accepted trade-off:** a service that already calls an LLM through raw `AddHTTPService` does not get
token metrics until it adopts `ctx.LLM()`. We consider adopting the client the right move for those
users, and the migration is a few lines.

Note on metric shape: this PR records an `app_llm_tokens_per_request` histogram whose Prometheus
`_sum` is the cumulative token count, rather than #3245's `app_llm_token_usage` counter, because the
metrics manager's `IncrementCounter` cannot add N to a counter. The histogram gives both cumulative
tokens (`_sum`) and per-request distribution from one metric.

## Out of scope / follow-ups
- Native Anthropic module (separate `go.mod`, post-v1).
- Programmatic per-call token accounting for billing / chargeback — proposed in #3727
  (`UsageObserver` hook), plus recording usage on error paths. Deliberately additive and not part of
  this PR.
- MCP write tools and their request-body input schema — deferred so v1 exposes a complete read-only
  surface; both return together as a follow-up.

## Also included
- **Streamed tool-call assembly** — streaming now supports tool calls: deltas are assembled by index
  and exposed via the optional `ai.ToolCallStreamer` interface after the stream drains.

## Linked issues
Fixes #3728 — LLM log records now carry their trace ID.


pkg/gofr/rest.go
```diff
@@ -1,36 +1,51 @@
...
+// MCP tool via EnableMCP, its input schema describes the body's fields and types in addition to the
+// path parameters. T is expected to be a struct (or pointer to one); other kinds add no body schema.
+func WithInput[T any]() RouteOption {
+	return func(c *routeConfig) { c.inputType = reflect.TypeFor[T]() }
+}
+
 // GET adds a Handler for HTTP GET method for a route pattern.
-func (a *App) GET(pattern string, handler Handler) {
-	a.add("GET", pattern, handler)
+func (a *App) GET(pattern string, handler Handler, opts ...RouteOption) {
```
Umang01-hash: Breaking: this changes the signature of an existing exported method (adds `opts ...RouteOption`) for all five verbs. Direct calls compile, but method-value assignments (`var f func(string, Handler) = app.GET`) and interface satisfaction break, and `apidiff` flags it — so the PR body's "no existing exported identifier changed" isn't accurate. Please either call this out as an accepted exception or keep the core five byte-identical (separate opts-aware registration).

pkg/gofr/tools.go
```diff
@@ -0,0 +1,448 @@
...
+
+	segs := strings.Split(pathTemplate, "/")
+	query = url.Values{}
+	body = map[string]json.RawMessage{}
+
+	for key, raw := range fields {
+		switch {
+		case params[key]:
+			// Escape so a value like "../admin" or "a/b" cannot break out of its path segment.
+			substituteSegment(segs, key, url.PathEscape(scalar(raw)))
```
Umang01-hash: HIGH — path traversal. `url.PathEscape` does not escape `.`, so `{"id":"../secret"}` builds `/public/..%2Fsecret`, and the router's own `path.Clean` collapses it to `/secret`. Reproduced e2e: an allowed tool reached a `WithExcludedRoutes`-excluded GET handler and returned its body. This defeats the exclude/`Only()` confinement for any GET route. Reject values containing `/`/`..` (or that change under `path.Clean`), and add a traversal regression test — the current guard has no coverage.

.golangci.yml
```diff
@@ -153,6 +153,7 @@ linters:
       - linters:
           - dupl
           - goconst
+          - gosec
```
Umang01-hash: This disables `gosec` for every `_test.go` in the whole repo, not just the AI tests, and is bundled into a feature PR. Prefer per-line `//nolint:gosec` on the specific fixtures, and split lint-policy changes into their own PR.

pkg/gofr/tools.go
```diff
@@ -0,0 +1,448 @@
...
+
+	if rec.Status() >= http.StatusBadRequest {
+		return ai.Result{}, fmt.Errorf("%w: tool %q returned status %d", errToolStatus, name, rec.Status())
+	}
+
+	return ai.NewResult(json.RawMessage(rec.Body())), nil
+}
+
+func (rt *routerTools) specFor(method, pathTemplate string) (ai.ToolSpec, bool) {
+	if rt.cfg.exclude[pathTemplate] || strings.HasPrefix(pathTemplate, "/.well-known") {
```
Umang01-hash: Low: the filter only excludes `/.well-known`, but `/favicon.ico` is also a framework route — live `tools/list` exposed `get_favicon.ico` as an agent tool. Exclude other framework-registered routes too.

Umang01-hash (changes requested):
Reviewed at head `63158bf` — built, ran the `using-ai` example E2E (metrics/health/traces/MCP all verified live), and reproduced the findings locally against the real framework. Strong feature, well-designed. Two must-fix items before merge:

1. **Breaking change** — `App.GET/PUT/POST/DELETE/PATCH` gained a variadic param, so the PR body's "no existing exported identifier changed" isn't accurate. Direct calls still compile, but it's a signature change (breaks method-value/interface uses; `apidiff` flags it). Needs an explicit decision or a non-breaking alternative.
2. **HIGH security** — MCP path-param traversal: `{"id":"../secret"}` on an allowed tool escapes its route and reaches excluded/`Only()`-filtered GET routes. Reproduced e2e — a tool restricted via `WithExcludedRoutes` returned the excluded handler's body. `url.PathEscape` doesn't escape `.` and the router's `path.Clean` re-collapses `..`.

Lower-severity items inline. Nice work overall — the observability, redaction, retry classification and streaming-leak handling all hold up under testing.

aryanmehrotra:
@Umang01-hash thanks again for the thorough review — all four items are addressed and the branch is green:

- **Path traversal (HIGH):** fixed — a path-param value that isn't a single safe segment (contains `/`, `..`, or changes under `path.Clean`) is now rejected before dispatch; added a traversal regression test.
- **Breaking verbs:** reverted `GET/POST/PUT/DELETE/PATCH` to their original signatures. `apidiff` now reports additive-only changes (zero incompatible). `WithInput` moved off the verbs — and has since been dropped from v1 entirely (see below).
- **gosec on `_test.go`:** reverted the repo-wide exclusion; used test constants for the fixtures instead of nolints.
- **favicon:** framework routes (`/favicon.ico`, `/.well-known`) are excluded from the tool list, with a test.

One scoping question before you re-review: **MCP is the largest surface in this PR** (and where the security-sensitive tool dispatch lives). Would you prefer we **split MCP out into a follow-up PR** and land just the LLM-as-datasource + observability + streaming core here? Or are you comfortable keeping MCP in this PR now that the traversal issue is fixed and covered? Happy either way — whichever makes the review most tractable.

pkg/gofr/tools.go
```diff
@@ -0,0 +1,411 @@
...
+
+// routerTools exposes the app's registered routes as ai.Tools by walking the router for discovery
+// and dispatching a synthesized request back through the router for invocation, so binding,
+// validation and auth middleware all run exactly as for a real HTTP call.
+type routerTools struct {
+	app *App
+	cfg *mcpConfig
+}
+
+func (rt *routerTools) List() []ai.ToolSpec {
```
Umang01-hash, line 52: Perf (non-blocking): `List()` walks the whole router and rebuilds every ToolSpec on each call, and `Call()` does the same via `route()` — O(routes) with no caching. Measured: 10 routes ~14µs/285 allocs, 100 ~141µs/2.8k allocs, 500 ~769µs/14k allocs per `List()`. An agent loop calls `List()` every turn plus `Call()` per tool, so a large service under heavy agent traffic pays ~1ms + hundreds of KB of garbage per op. Consider building the spec list once and invalidating on route registration. No impact on normal HTTP requests.

Umang01-hash (commented):
Re-reviewed at head `9bd32b8`. All four earlier findings are resolved and independently verified: rest.go is byte-identical to development (breaking change reverted), the MCP path-traversal is fixed (re-ran the exact bypass — now rejected, regression test added), the repo-wide gosec test exclusion is removed (0 new lint issues under `--new-from-rev`), and favicon is excluded from the tool list. Build/vet/gofmt clean; `-race` green on ai/llm/mcp/http; example E2E works (metrics + health + MCP verified live). Coverage: ai/mcp 97.8%, ai/llm 89%, root pkg/gofr 87% (hand-written ai code ~100%; the ai-package 43.6% is just the generated mock).

Two non-blocking follow-ups:

1. **Real-provider smoke test** — I verified everything against a mock OpenAI-compatible server, not a live Groq/OpenAI/DeepSeek endpoint. Worth one real-key run before release.
2. **MCP tool List/Call is O(routes) with no caching** — see inline. Fine for typical services, worth addressing before large services under heavy agent loops.

Unrelated heads-up: the root `pkg/gofr` test is intermittently flaky (~1 in 3-5 runs) in `cron.go` (mock-metrics teardown) — pre-existing, not this PR, but it'll occasionally red CI.

Umang01-hash (dismissed):
Approving at head `9bd32b8`. All four earlier findings are resolved and independently re-verified: the breaking `GET/POST/...` signature change reverted (rest.go now identical to development), MCP path-traversal fixed + regression-tested (re-ran the exact bypass — rejected), gosec test exclusion removed with 0 new lint issues, and favicon excluded from the tool list. Build/vet/gofmt clean, `-race` green on ai/llm/mcp/http, example E2E verified live (metrics/health/MCP), coverage solid (ai/mcp 97.8%, ai/llm 89%, root 87%).

The two follow-ups I left are non-blocking and don't gate merge: a real-provider smoke test before the release cut, and the O(routes) tool List/Call caching for large services under heavy agent load. Nice work — solid, well-tested feature.

akshat-kumar-singhal:
Really like the shape of this — particularly breaking `cached` and `reasoning` out as their own `token_type` series, and documenting the cost formula in the `RegisterMetrics` doc comment. That's more than most LLM instrumentation gives you.

One thing I'd push back on, for the cost-tracking use case: **tokens are dropped on error, and there are paths where the provider billed us and the usage is already in hand.**

In `record()`:

```go
if err == nil && resp != nil {
    recordTokens(ctx, m, info, resp.Usage)
}
```

The doc comment frames this as making a token type's histogram `_count` track "successful requests that reported that type," which is a coherent invariant. But it means spend that actually occurred never reaches the metric. Two concrete paths in this PR:

1. **200-with-error-object** (`llm/client.go`, the `if cr.Error != nil` branch). The comment there notes some OpenAI-compatible gateways report failures with HTTP 200 and an error object. At that point `cr.Usage` has already been decoded — the tokens are sitting in the struct — and the function returns `nil, errProvider`, so `resp` is nil and the usage is discarded. Gateways that do this generally still bill the request.

2. **Stream that fails mid-drain.** `streamer` accumulates `s.usage` from any usage-bearing chunk and exposes it via `Usage()`; the `usageReporter` path reads it and hands it to `Recorder.Finish(u, err)`. With a non-nil `err`, `record()` drops it. Even without a usage chunk, a partially-drained stream has consumed prompt tokens plus whatever completion tokens were emitted.

Neither is exotic — path 1 is a whole class of gateway behavior you already handle explicitly, and both get worse under retry, where a burst of billable failed attempts shows as zero tokens.

**Suggested fix**, which I think preserves what the current invariant is protecting: add `status` to the token histogram labels and always record whatever usage was reported.

```go
if resp != nil {
    recordTokens(ctx, m, info, resp.Usage, status)
}
```

Then `token_type="prompt", status="success"` keeps exactly today's meaning and today's `_count` semantics, billable totals become a sum across both statuses, and failed-but-billed spend stops being invisible. It also makes "how much are we burning on failures" a first-class query, which is usually the thing people most want to alert on. Cost per the doc comment's formula would just drop the status filter.

Cardinality impact is 2x on a metric that's already deliberately low-cardinality, and `recordTokens` still skips zeros, so error calls that genuinely report nothing add no series.

Happy to send a PR against your branch if you think it's worth doing — and equally happy to hear the counter-argument if the `_count` invariant is load-bearing somewhere I haven't spotted.

akshat-kumar-singhal:
Filed #3727 as a follow-up to this PR — a `UsageObserver` hook so applications can persist per-call token usage with their own attribution (tenant, job, request), which the metrics/traces/logs here can't provide: metrics are aggregate and intentionally low-cardinality, traces are sampled, and `ai.Log` has no correlation ID to join on.

It's deliberately an issue rather than a review comment, since it's purely additive and shouldn't block this merging. The proposed shape slots into seams this PR already left open — an optional `Deps` field, one call in `record()` (which covers streaming for free via `Recorder.Finish`), and the currently-empty `llmOptions` struct that's described as reserved for forward-compatible options. The framework never learns what a tenant is; the observer gets the `ctx` and the application pulls its own dimensions off it.

Happy to send a PR if the direction seems right.

akshat-kumar-singhal:
One more, small and self-contained: #3728 — the new `ai.Log` record is emitted without a trace ID, so an LLM log line can't be joined back to the request that caused it.

The cause is just wiring: `AddLLM` → `instrumentDatasource` → `UseLogger(a.Logger())` injects the base logger at boot, whereas `logging.ContextLogger` (which already stamps `__trace_id__`) is only constructed per-request in `context.go`. `writeLog` also takes no context, so it can't recover one. `record()` already has the span in hand, so the fix is a `TraceID` field on `ai.Log` set behind a `SpanContext().IsValid()` guard — no new dependency on `pkg/gofr/logging`, which keeps the `ai` package's minimal `Logger` interface decoupled as it looks intended to be.

Unrelated to the design discussion in #3727 and independent of it — happy to send a PR for this one whenever.

aryanmehrotra:
Thanks for the careful review — both points are addressed on the branch:

**Token usage on error.** Done, and I went with the `status`-label approach you suggested. `app_llm_tokens_per_request` now carries a `status` label and records whatever usage the provider reported on every call, so `token_type=…,status="success"` keeps its exact prior meaning and `_count` semantics, while failed-but-billed spend surfaces under `status="error"`. `recordTokens` still skips zeros, so an error that reports nothing adds no series. I also fixed the 200-with-error path in the client — it previously returned before mapping usage, so those tokens were dropped before they could be recorded; it now preserves the reported usage alongside the error, which is what feeds the error-status series.

**Trace correlation (#3728).** Fixed. `ai.Log` now carries `traceId`, set from the call span already in hand in `record()` behind a `SpanContext().IsValid()` guard — no new dependency on `pkg/gofr/logging`, so the `ai` package keeps its minimal `Logger` interface. Verified end-to-end against a running service: LLM log lines now correlate to their request (the two calls in an agent turn share one trace), and error-status lines carry it too.

Both are in the latest push, with tests.

Umang01-hash (dismissed):
Re-approving at head `7577358` (re-reviewed after the branch moved well past my earlier approval). Since then: `WithWriteTools` was removed so MCP exposes read-only handlers only (verified: write handlers can't be listed or called), multi-LLM (`WithName` / `ctx.LLM(name)`) was added cleanly, and observability got better (bounded `status` label on token metrics, trace-id in LLM logs, no partial-response leak on error).

Re-verified at this head: no breaking change (all LLM symbols are new-in-PR, absent on development); build/vet/gofmt clean; `golangci-lint --new-from-rev` 0 new issues; `-race` green on ai/llm/mcp + root + example; path-traversal still blocked (re-ran the bypass); live E2E shows metrics/health/MCP working with no key leak and no favicon/write tools exposed.

Two non-blocking follow-ups remain (don't gate merge): MCP tool List/Call is still O(routes) with no caching (fine for typical services), and a real-provider smoke test before the release cut (I verify against a mock). Solid work — good to merge.

pkg/gofr/container/container.go
```diff
@@ -231,6 +238,86 @@ func (c *Container) Metrics() metrics.Manager {
...
+
+// SetTools installs the agent-tool provider exposed to handlers via ctx.LLM().Tools().
+func (c *Container) SetTools(t ai.Tools) {
+	c.tools = t
+}
+
+// LLM returns the instrumented model exposed to handlers, or nil if none was added. A mock
+// container injects its mock through the same field, so there is no test-only branch here.
+func (c *Container) LLM(name ...string) ai.LLM {
+	return c.llms[llmName(name)]
```
NitinKumar004: **[Medium / DX] `ctx.LLM(name)` returns a nil interface for an unknown or typo'd model name → nil-pointer panic (HTTP 500).**

A map miss here yields a nil `ai.LLM`, so `ctx.LLM("fst").Generate(...)` (a typo, trailing space, or wrong case) panics and is recovered as a 500 with no hint that the name was wrong. Same for `ctx.LLM()` when no model was registered at all.

Verified live: `app.AddLLM(m, gofr.WithName("fast"))` then a handler calling `ctx.LLM("does-not-exist").Generate(...)` returns HTTP 500 (server survives, but the error is opaque).

Suggestion: fall back to the default model when a named one is absent, or return `ai.ErrLLMNotConfigured` so the caller gets a clear error instead of a panic. (`lazyTools` already models this nil-safe pattern.)

pkg/gofr/ai/llm/client.go
```diff
@@ -0,0 +1,437 @@
...
+	return false
+}
+
+func retryDelay(resp *http.Response, attempt int) time.Duration {
+	if resp != nil {
+		// Only the integer-seconds form is honored; the HTTP-date form falls back to backoff. A
+		// server-supplied delay is capped so a large Retry-After cannot stall the caller for minutes.
+		if v := resp.Header.Get("Retry-After"); v != "" {
+			if secs, err := strconv.Atoi(v); err == nil {
+				return min(time.Duration(secs)*time.Second, maxRetryAfter)
```
NitinKumar004: **[Low] `Retry-After` cap can be bypassed by integer overflow / negative values.**

`time.Duration(secs)*time.Second` multiplies by 1e9, so a large `secs` overflows int64 and wraps negative — e.g. `Retry-After: 10000000000` yields a negative duration. A negative header value (`Retry-After: -5`) also passes straight through. In both cases `min(negative, maxRetryAfter)` returns the negative value, so `backoffSleep`'s timer fires immediately and the intended 5s cap is defeated.

Impact is bounded (`maxRetries = 2`, so at most two near-immediate retries — no long stall), so it's a correctness nit rather than a DoS. The normal integer path is fine — verified live that a `Retry-After: 1` produced ~2s of backoff.

Suggestion: validate `secs` into `[0, maxRetryAfterSecs]` before converting to a `Duration`.

pkg/gofr/tools.go
```diff
@@ -0,0 +1,389 @@
...
+	rt.app.httpServer.router.ServeHTTP(rec, req)
+
+	if rec.Status() >= http.StatusBadRequest {
+		return ai.Result{}, fmt.Errorf("%w: tool %q returned status %d", errToolStatus, name, rec.Status())
+	}
+
+	return ai.NewResult(json.RawMessage(rec.Body())), nil
+}
+
+func (rt *routerTools) specFor(method, pathTemplate string) (ai.ToolSpec, bool) {
```
NitinKumar004, line 112: **[Low / info-disclosure] `tools/list` advertises auth-protected routes to unauthenticated MCP clients.**

`specFor` filters by method (read-only) and the exclusion/framework lists, but not by whether the route sits behind auth middleware. So an unauthenticated MCP client sees the **name and input schema** of protected routes (e.g. a `get_secret_id` behind `EnableAPIKeyAuth`).

It cannot *call* them — I verified call-time correctly returns 401 without a valid key, so this is **not** an auth bypass — but it can *discover* their existence and shape. If route names/paths are considered sensitive, consider omitting routes the caller isn't authorized for, or documenting that `tools/list` is a discovery surface.

pkg/gofr/ai/instrument.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Log is the structured record emitted for each LLM call. It never carries prompt or completion
+// text, only metadata safe to log.
+type Log struct {
+	Provider         string `json:"provider"`
+	Model            string `json:"model"`
+	Operation        string `json:"operation"`
+	Status           string `json:"status"`
+	TraceID          string `json:"traceId,omitempty"`
```
NitinKumar004: **[Nit] trace-id key casing is inconsistent.**

This LLM log record serializes the trace id as `traceId`, whereas GoFr's request/error log lines use `trace_id`. Harmless, but a log pipeline filtering on `trace_id` won't match these LLM lines. Consider `trace_id` for consistency.

NitinKumar004 (commented):
## End-to-end local testing report

I pulled the branch and tested it as a real user would — a live server process driven with `curl`, not just the unit suite. I built a small OpenAI-compatible mock provider and pointed the real GoFr app at it via `LLM_BASE_URL`, then exercised every flow, error path, edge case, and the security-critical surfaces (with the race detector on).

**Overall: the PR is solid and works end-to-end. No high-severity or blocking issue found.** Every security claim held up under adversarial live testing. The items below are 1 medium (DX) + a few low/cosmetic notes, left inline.

### Environment
- `go build ./...` clean; `go test -race ./pkg/gofr/ai/...` and `./examples/using-ai/...` pass.
- Note: the `-race` failures in `./pkg/gofr/` (gRPC/shutdown/websocket) also fail on `development` — pre-existing flaky port/timing tests, **not** from this PR.

### Happy-path flows — all ✅
| Flow | Result |
|---|---|
| `GET /inventory/{sku}` handler | ✅ |
| `/.well-known/health` shows LLM datasource | ✅ UP, provider/model shown, **no API key** |
| `POST /ask` one-shot completion | ✅ content + token usage |
| `POST /stream` SSE | ✅ frames + `[DONE]` |
| `POST /agent` loop (LLM → tool call → dispatch → final answer) | ✅ |
| MCP `initialize` / `tools/list` / `tools/call` / `ping` | ✅ |
| Prometheus `/metrics` (`app_llm_request_count`, `app_llm_tokens_per_request`) | ✅ per operation/model |
| Multi-model: `ctx.LLM()` vs `ctx.LLM("fast")`, per-model health & metric labels | ✅ |
| Zero-config env auto-init (`LLM_PROVIDER`/`LLM_MODEL`/`LLM_API_KEY`, no `AddLLM`) | ✅ |

### Error / edge / security cases — all ✅
| Case | Result |
|---|---|
| Provider **429 + Retry-After** | retried (~2s, honored) → clean error, metric `status="error"` |
| Provider **500** | clean error |
| **Provider error in HTTP 200 body** | correctly mapped to an error |
| MCP **path traversal** (`../../etc/passwd`) | rejected: "unsafe path parameter" |
| MCP **call a write (POST) route** | "unknown tool" — refused at call time |
| MCP unknown tool / bad method / GET / oversized (>1MB) body | `unknown tool` / `method not found` / `405` / "too large" |
| MCP notification (no id) | `202 Accepted` |
| **MCP loopback-only bind** | reachable on `127.0.0.1`, **connection refused on the LAN IP** |
| **Auth NOT bypassable via MCP** | call to an API-key-protected route: no key → **401**, wrong key → **401**, correct forwarded key → data. Middleware genuinely runs on the synthesized request. |
| API-key leakage (logs / metrics / health / MCP / race log) | **0 everywhere** |
| LLM error logs carry trace ID | ✅ (issue #3728) |

### Concurrency (race detector on)
- ~1,400 concurrent clients across LLM + streaming + mid-stream aborts + MCP tool calls: **0 data races, 0 panics, 0 deadlocks**, server healthy throughout.
- Metric counters consistent under load (`sum(request_count)` matched the exact number of requests sent — no lost/double updates).
- Client cancel / mid-stream disconnect: server stays healthy, no goroutine leak observed.

### Findings (inline)
1. **[Medium, DX]** `ctx.LLM("typo")` → nil-interface panic → HTTP 500 (no fallback / no clear error).
2. **[Low]** `Retry-After` cap bypassable by overflow / negative values.
3. **[Low, info-disclosure]** `tools/list` advertises auth-protected route names + schemas to unauthenticated MCP clients (cannot call them, but can discover them).
4. **[Nit]** LLM log uses `traceId` while the rest of GoFr uses `trace_id`.

Great work overall — the security model (call-time read-only enforcement, loopback bind, auth propagation, no secret leakage, traversal blocking) is well-designed and verified. The only one I'd suggest addressing before release is #1, since a typo'd model name turning into a 500 is an easy footgun.

Umang01-hash (approved):

PR #3755, opened 2026-07-23
test(service): fix flaky CB test intervals and add deterministic concurrent recovery test

### LINKED ISSUE #3501: [Bug]: Fatal Concurrency Gridlock - Mutex Write-Lock Held Over Synchronous Network I/O in Circuit Breaker Recovery

adityakrmishra:
## Objective / Issue

Follow-up to #3501. The concurrency fix for `tryCircuitRecovery` has since landed on `development`; what remains here is the test side — and one measurement that turned out to matter more than the flake it was chasing.

## What we found

**`Interval: 1` meant one nanosecond.** `CircuitBreakerConfig.Interval` is a `time.Duration`, so every `Interval: 1` in `circuit_breaker_test.go` configured a one-nanosecond recovery interval. `NewCircuitBreaker` passes it straight to `time.NewTicker` on a goroutine with no stop path, so each circuit breaker the suite constructed left a ticker firing as fast as the runtime could deliver, for the life of the test binary.

Measured on `pkg/gofr/service`, `-run CBOpenRequests -count=20`, warm build cache, three runs each:

|---|---|---|
| CPU (user+sys) | 68.7s / 90.7s / 71.0s | **1.44s / 1.47s / 1.48s** |
| CPU utilisation | 568% / 500% / 371% | **11% / 11% / 11%** |
| wall clock | 12.1s / 18.2s / 19.1s | 13.2s / 13.5s / 13.1s |

**The CPU figure is the result — roughly 50× — and the wall clock is not a speedup.** Read the third row honestly: this branch is a little slower than `development`'s best run, because 50ms intervals are real waiting that one nanosecond skips. What it buys is a stable 13s instead of anything from 12s to 19s, and it stops the suite burning five cores to do it.

**That was causing a real failure, not just waste.** `TestCircuitBreaker_MixedHTTPMethods` asserts five parallel one-second requests finish inside two seconds. It fails under `go test -count=5` on `development` and passes here — a wall-clock assertion competing with several cores of spinning tickers is exactly what fails on a loaded runner and nowhere else.

## What changed

- Every `Interval: 1` becomes `cbTestInterval` (50ms), and the tests that need that interval to have elapsed wait `cbTestRecoveryWait` (60ms). The two are named constants because they are a pair: the wait must exceed the interval, or a request meant to find the circuit recoverable finds it still open.
- `TestCircuitBreaker_ConcurrentRecovery_OnlyOneResetsCircuit`: twenty goroutines discover the open circuit at once and exactly one must fire the health check. The mock health check blocks until all nineteen losers have returned from `tryCircuitRecovery`, proving they were turned away rather than merely slower, and a bounded wait makes a second winner a failure instead of a hang.

## What this actually buys — the flake, measured

The strongest argument for this PR is not the new test, it is the flake the intervals fix. Running `TestCircuitBreaker|CBOpenRequests|SuccessRequests` from a prebuilt binary:

| configuration | iterations | `development` fails | this branch fails |
|---|---|---|---|
| `GOMAXPROCS=1` | 3 / 35 | **3 (100%)** | **0** |
| `GOMAXPROCS=10` (default) | 25 | 1 (4%) | **0** |
| 10 busy goroutines contending | 25 | 0 | **0** |
| `go test -race -count=5`, whole package | 1 | **FAIL, 271.5s** | **ok, 32.8s** |

Every failure is the same shape — a wall-clock budget missed because the leaked nanosecond tickers saturate the CPU:

```
--- FAIL: TestCircuitBreaker_MixedHTTPMethods (2.99s)
    circuit_breaker_test.go:1116: "2.682835375s" is not less than "2s"
```

The leaked goroutine **count** is unchanged (~33 either way — `NewCircuitBreaker`'s ticker goroutine has no stop path, which is pre-existing and not fixed here). What changes is the price of each: measured in isolation, one 1ns ticker burns 0.29 of a core and one 50ms ticker burns 0.0001.

## What the new test catches

Reintroducing the #3501 bug — holding the lock across `healthCheck` — fails `TestCircuitBreaker_ConcurrentRecovery_OnlyOneResetsCircuit` **3 times out of 3**, on the `othersDone` handshake rather than on a timing margin, so it is a deterministic kill and not a race that usually works.

**In fairness to the reader: that mutant is also killed 3/3 by `TestCircuitBreaker_SlowHealthCheckDoesNotBlock`, which is already on `development`.** An earlier revision of this description claimed a TOCTOU mutation separated the two, with a 0/10-vs-10/10 table; re-running it, both tests catch that mutation 10/10, and a narrower variant separates them only under `-race`, where the kill is a `DATA RACE` report rather than the handshake — and CI does not run `-race` on this package. So no mutant found so far distinguishes the new test.

What it does add is coverage of a shape the existing test cannot reach: twenty concurrent claimants rather than two, with an explicit handshake proving the nineteen losers were *turned away* rather than merely slower. That is a reasonable thing to have, and it is cheap — but it is an argument from shape, not from a mutant, and it is offered as such.

## Scope

Test-only. `pkg/gofr/service/circuit_breaker_test.go` is the single file changed — the one production line in the original branch was a blank line already present on `development`.

The interval work is @adityakrmishra's: commit `e74745f5d` made all eleven interval changes and all ten sleep changes. The later commits name the two values as constants, record what the nanosecond interval cost, and correct this description.

## Test plan

- [x] `gofmt` clean, `go vet ./pkg/gofr/service/` passes
- [x] `golangci-lint --new-from-rev=origin/development` — 0 issues
- [x] `go test -short ./pkg/gofr/service/` passes
- [x] `go test -short -race ./pkg/gofr/service/` passes, no races
- [x] `-count=5` on the whole package passes; the same run fails on `development`
- [x] Reintroducing the #3501 bug fails the new test 3/3

## Related, not fixed here

`CircuitBreakerConfig` has no default for `Interval`, so `&CircuitBreakerConfig{Threshold: 5}` reaches `time.NewTicker(0)` and panics with `non-positive interval for NewTicker` on a background goroutine, killing the process at startup. Reproduced on `development`; it needs its own fix, and this PR stays test-only.


adityakrmishra:
Hey @Umang01-hash and @thzgajendra, thanks for the update on the issue! Glad to hear the core gridlock bug was patched in v1.56.6.

Just to clarify, this PR (#3755) does not touch the implementation code. Since the fix is already in place, I repurposed this PR to strictly add the concurrent regression stress tests that @aryanmehrotra requested when my original PR was reviewed.

This PR adds a strict "starting gun" synchronization barrier (<-startCh) to hit the tryCircuitRecovery block with 20 concurrent goroutines simultaneously, utilizing atomic state guards to definitively prove that the lock is released before the network I/O and that exactly one health check fires.

It's a great regression test to ensure the v1.56.6 fix stays bulletproof in the future. Let me know if it looks good to merge!

Umang01-hash (dismissed):

aryanmehrotra (changes requested):
Reviewed end to end at `f648564c`. CI is 14/14, `golangci-lint --new-from-rev=development` reports 0 issues, and `./pkg/gofr/service/` passes. The `Interval` change is a real find. Three things before this lands, one of them about the headline test.

## The most valuable change is not in the title

`CircuitBreakerConfig.Interval` is a `time.Duration` (`circuit_breaker.go:26`), so `Interval: 1` is **one nanosecond**. Eleven tests on `development` configure it that way, which makes `time.Since(cb.lastChecked) > cb.interval` true on essentially every call — so every request through an open circuit fired a health check, and none of those tests were exercising the interval at all.

Moving them to `50 * time.Millisecond` fixes that, and it is the change with the broadest effect here. It is not mentioned in the title, the description, or any commit message; someone reading the changelog would not know it happened. Worth promoting to the title, or splitting into its own commit with a message that says what `1` meant.

## The title claims production hardening that is not here

"harden circuit breaker concurrent recovery with **atomic state guards**" — the entire production diff is:

```diff
 	cb.mu.Unlock()
+
 	return false
```

A blank line. No atomic guards were added, and `sync/atomic` is imported only for the test's counter. The lock-over-network-I/O problem from #3501 was already fixed in `c7c37e2a` and `c9ef6ccb` — `tryCircuitRecovery` releases the mutex before `cb.healthCheck` on `development` today, and #3501 is closed.

That is fine as a PR — a regression test for an already-landed fix is worth having — but the title should say so. `test(service): ...` rather than a title that reads as a concurrency fix.

## The new test does not detect the regression it describes

This is the one I would like changed. `TestCircuitBreaker_ConcurrentRecovery_OnlyOneResetsCircuit` asserts that exactly one goroutine fires the health check. The thing that actually guarantees that is the claim under the lock:

```go
if time.Since(cb.lastChecked) > cb.interval {
	// Update lastChecked to prevent busy loop of health checks from other requests
	cb.lastChecked = time.Now()
	cb.mu.Unlock()
```

I removed those two lines and re-ran, in isolation, five times:

```
PR's test           -> PASSES the mutant   (does not detect it)
```

The reason is that with a mock health check that returns immediately, the winning goroutine completes `resetCircuit()` before the other nineteen are scheduled, so they take the `state == ClosedState` early return and never reach the interval check. The assertion is satisfied by the state guard, not by the guard the test is named for. It does catch the mutant when run as part of the whole `TestCircuitBreaker` set — because the machine is busy enough for the losers to get a window — so whether it tests anything depends on load. On a quiet runner it is vacuous.

Worth noting the pre-existing `TestCircuitBreaker_SlowHealthCheckDoesNotBlock` catches that same mutant reliably, so as written the new test adds little the suite did not already have.

The fix is to hold the health check open until the other callers have had their turn, which pins them inside the window the claim protects. I wrote and ran this against your branch:

```go
othersDone := make(chan struct{})

mockHTTP.EXPECT().HealthCheck(gomock.Any()).
	DoAndReturn(func(_ context.Context) *Health {
		atomic.AddInt64(&healthCallCount, 1)
		<-othersDone // hold the circuit Open until the losers have had their turn

		return &Health{Status: serviceUp}
	}).AnyTimes()

cb := &circuitBreaker{
	state:       OpenState,
	threshold:   1,
	interval:    time.Hour,
	lastChecked: time.Now().Add(-2 * time.Hour),
	HTTP:        mockHTTP,
}

// ... in each goroutine, after cb.tryCircuitRecovery():
// the winner is still blocked inside healthCheck, so only losers reach here
if atomic.AddInt64(&returned, 1) == numGoroutines-1 {
	once.Do(func() { close(othersDone) })
}
```

```
deterministic version -> passes real code (10 runs), FAILS the mutant
```

Two notes on it. `interval: time.Hour` with `lastChecked` two hours back removes the scheduling assumption entirely — at `50ms` a goroutine descheduled for longer than the interval would see it elapsed a second time and fire another health check, which is a small flake risk on a loaded runner even in the current form. And the barrier needs a bounded wait: against the mutant more than one goroutine blocks on `othersDone`, so `returned` never reaches N-1 and the test hangs until `-timeout` rather than failing with a message. A `select` with a short deadline turns that into a readable assertion.

## Minor

Ten `time.Sleep(60 * time.Millisecond)` calls is about 600ms added to the package, and #3821 is currently removing fixed sleeps from the example tests in favour of readiness signals, so this moves the opposite way. I do not think it is avoidable here — the breaker is behind `NewHTTPService`, so there is no clean handle on `lastChecked` from those tests — but if there is a way to reach it, driving the clock beats sleeping on it. Say if you have looked and it is not reachable; I would rather that were recorded than rediscovered.

Also `circuit_breaker_test.go:531` in `TestHttpService_PostWithHeaderCBOpenRequests` is missing the blank line after the sleep block that the other nine have. Lint does not flag it.

To be clear about what I am asking: the `Interval` fix and the extra coverage are both worth having. It is the concurrency test I want made deterministic before merge, because as written it will pass whether or not the guard it names is present.

adityakrmishra:
Thanks for the great review and the mutation testing! You were spot on about the fast-returning mock bypassing the interval guard.
I have pushed the requested changes:
- Implemented the othersDone barrier with a 2s timeout and set the interval to time.Hour to make the test fully deterministic.
- Updated the PR title and commit message to accurately reflect this is a test/interval fix.
- Added the missing blank line at 531.

(Note on the 600ms sleep: I looked into driving the clock instead, but since the CB is encapsulated behind NewHTTPService, lastChecked is unreachable from these black-box tests. So the sleeps are unavoidable for now).

The CI is completely green. Let me know if this is good to merge!

aryanmehrotra:
Reviewed this end to end and pushed the review changes onto the branch as a separate commit — your three commits are untouched.

This PR is worth considerably more than its description claims, and the reason is buried in one character.

### `Interval: 1` meant one nanosecond

`CircuitBreakerConfig.Interval` is a `time.Duration`, so every `Interval: 1` in these tests configured a **one-nanosecond** recovery interval. `NewCircuitBreaker` hands that straight to `time.NewTicker` in a goroutine that never stops, so each circuit breaker the suite constructed left behind a ticker firing as fast as the runtime could deliver, for the life of the test binary.

Replacing them is not a readability change. Measured on `pkg/gofr/service`, one run, same machine:

| | `development` | this PR |
|---|---|---|
| wall clock | 21.75s | **6.17s** |
| CPU (user) | 85.48s | **0.77s** |
| CPU utilisation | 391% | 18% |

111× less CPU, and 3.5× faster. Under `-count=5` it is 224s versus 30s.

### It fixes a real flake, not a theoretical one

`TestCircuitBreaker_MixedHTTPMethods` asserts that five parallel one-second requests finish inside two seconds. On `development` it fails under `-count=5`:

```
$ go test -short -count=5 ./pkg/gofr/service/          # development
--- FAIL: TestCircuitBreaker_MixedHTTPMethods (2.01s)
FAIL    gofr.dev/pkg/gofr/service       224.067s
```

On this branch that run passes. A wall-clock assertion competing with four cores of spinning tickers is exactly the kind of test that fails on a loaded CI runner and nowhere else.

### The new concurrency test earns its place

I nearly flagged `TestCircuitBreaker_ConcurrentRecovery_OnlyOneResetsCircuit` as redundant. `development` has gained both the #3501 fix and `TestCircuitBreaker_SlowHealthCheckDoesNotBlock` since you opened this, and that test catches the original bug and the thundering-herd variant, so on those two mutants the new one adds nothing.

It is not redundant. Widening the window between observing the elapsed interval and claiming the recovery slot — the TOCTOU where two goroutines can both win — separates them completely:

| Mutation: recovery slot claimed a moment too late | Caught in 10 runs |
|---|---|
| `TestCircuitBreaker_SlowHealthCheckDoesNotBlock` (2 goroutines) | **0/10** |
| `TestCircuitBreaker_ConcurrentRecovery_OnlyOneResetsCircuit` (20 goroutines) | **10/10** |

Two goroutines almost never collide. Twenty do, every time. That is a real class of regression only this test can see, and the `othersDone` handshake — the losers releasing the winner's health check — is what makes it deterministic rather than a race that usually works.

### What I changed

The 50ms interval and the 60ms wait are a pair: the wait has to exceed the interval, or a request meant to find the circuit recoverable finds it still open. They were ten unrelated literals, so nothing stopped one being tuned without the other. They are now `cbTestInterval` and `cbTestRecoveryWait`, with a comment carrying the measurement above — the reason not to go back to a tiny interval is a number, and a number is worth writing down.

### ⚠️ Separately: a zero `Interval` crashes the process

Not this PR's to fix, but it is the same confusion and it is worse. `CircuitBreakerConfig` has no default for `Interval`, so a user who configures only a threshold:

```go
service.NewHTTPService(url, logger, metrics, &service.CircuitBreakerConfig{Threshold: 5})
```

reaches `time.NewTicker(0)` on a background goroutine:

```
panic: non-positive interval for NewTicker
```

It is unrecoverable — the panic is in a goroutine GoFr does not wrap — so the application dies at startup. Reproduced on clean `development`. Worth its own issue and fix; I have not touched it here, so that this PR stays test-only.

Two adjacent things I am **not** flagging against this PR, both reproduced on `development`:

- `TestHTTPService_HealthCheckTimeout` flakes about 1 run in 10, in isolation. `HealthConfig{Timeout: 1}` against a handler that sleeps exactly `1 * time.Second` is a coin toss — the same bare-integer-as-duration pattern this PR is cleaning up, one file over.
- The `startHealthChecks` goroutine has no stop path, so it outlives the service that created it.

### Gates

| Check | Result |
|---|---|
| `gofmt -l` | clean |
| `go vet ./pkg/gofr/service/` | pass |
| `typos` | clean |
| `go test -short ./pkg/gofr/service/` | pass, 6.17s |
| `go test -short -race ./pkg/gofr/service/` | pass, no races |
| `-count=5` whole package | passes; the same run fails on `development` |
| new test vs the #3501 regression | caught |

One note on mechanics: this targets `development` but branched from the repository's initial commit, so GitHub shows it as conflicting and attributes several hundred unrelated files to it. The branch is now your three commits replayed on current `development`, which is why the diff is one file.

aryanmehrotra:
Second review pass over this branch. The change is good and my conclusion hasn't moved — but **two of the numbers I published were wrong**, and one of them was wrong in a way that flattered the PR. Both are corrected in the description; here is what changed and why.

### 1. The CPU table compared a cold build cache against a warm one

I posted 21.75s → 6.17s wall, 85.48s → 0.77s CPU, 391% → 18%. Re-measured properly — same `-run` filter, same `-count`, both caches warm, three runs each:

|---|---|---|
| CPU (user+sys) | 68.7s / 90.7s / 71.0s | **1.44s / 1.47s / 1.48s** |
| CPU utilisation | 568% / 500% / 371% | **11% / 11% / 11%** |
| wall clock | 12.1s / 18.2s / 19.1s | 13.2s / 13.5s / 13.1s |

So the CPU result is **~50×, not 111×** — still the finding, still worth the PR.

**But the wall-clock claim doesn't survive at all, and it's the one I'd most want corrected.** Read row three: this branch is *slower* than `development`'s best run. Of course it is — 50ms of real waiting is 50ms the nanosecond interval skipped. What it actually buys is a stable ~13s instead of anything between 12s and 19s, and it stops the package burning five cores to get there. My "3.5× faster" was an artifact of measuring one run each with unequal caches.

### 2. The 0/10-vs-10/10 table is withdrawn

I claimed the new 20-goroutine test separates itself from `TestCircuitBreaker_SlowHealthCheckDoesNotBlock` on a TOCTOU mutation. Re-running it:

- Under the mutation **as I worded it**, both tests catch it 10/10. No separation.
- Under a narrower variant, the separation appears **only with `-race`** — and the failure is a `DATA RACE` report at 0.00s, not the `othersDone` handshake I credited. `grep -rn -- "-race" .github/workflows/` returns nothing, so CI never runs it.

What the test does earn, verified: reintroducing the original #3501 bug — holding the lock across `healthCheck` — fails it **3 times out of 3**, on the handshake, deterministically. That is a real and sufficient argument for the test, and it's what the description says now.

### 3. Credit where it belongs

@adityakrmishra's `e74745f5d` made all eleven interval changes and all ten sleep changes. My commits name the two values as constants and record what the old interval cost. The description now says so; it read as though the measurement work and the fix were the same contribution.

### Also pushed (`4dec37942`)

The committed comment quoted "167s of CPU" as an absolute. That figure ships into the repo permanently and nobody can reproduce the conditions behind it, so it's now the ratio — which is the stable part and the part that makes the point.

### What still stands

Everything else re-verified: `Interval: 1` is one nanosecond; `TestCircuitBreaker_MixedHTTPMethods` fails at `-count=5` on `development` (4 of 15 iterations) and never on this branch (0 of 15); 0 flakes here across ~100 stress iterations; test-only, one file; `gofmt`/`vet`/`golangci-lint` all clean, 0 issues.

And the zero-`Interval` startup panic is confirmed pre-existing on `development` — `&CircuitBreakerConfig{Threshold: 5}` reaches `time.NewTicker(0)` and takes the process down from a background goroutine. Still needs its own PR.

aryanmehrotra:
Reviewed this end to end and pushed onto the branch, authorship intact.

## `Interval: 1` meant one nanosecond

`CircuitBreakerConfig.Interval` is a `time.Duration`, so every `Interval: 1` in this file configured a **one-nanosecond** recovery interval. `NewCircuitBreaker` hands it straight to `time.NewTicker` on a goroutine with no stop path — a ticker firing as fast as the runtime can deliver, still firing after the test that created it has returned.

## What that was actually costing — the flake

This is the strongest argument for the PR and it was not in the description. Prebuilt binaries, `-test.run 'TestCircuitBreaker|CBOpenRequests|SuccessRequests'`:

| configuration | iterations | `development` | this branch |
|---|---|---|---|
| `GOMAXPROCS=1` | 3 / 35 | **3 failures (100%)** | **0** |
| `GOMAXPROCS=10` | 25 | 1 (4%) | **0** |
| `-race -count=5`, whole package | 1 | **FAIL, 271.5s** | **ok, 32.8s** |
| `-count=10`, whole package | 1 | **FAIL — 10-minute panic timeout** | **ok, 58.9s** |

Every failure is the same shape — a wall-clock budget missed because the leaked tickers saturate the CPU:

```
--- FAIL: TestCircuitBreaker_MixedHTTPMethods (2.99s)
    circuit_breaker_test.go:1116: "2.682835375s" is not less than "2s"
```

A single-core runner fails this **every time** on `development`, and the goroutine dump on the 10-minute timeout names `startHealthChecks.func1`. The leaked goroutine *count* is unchanged (~33 either way — the ticker goroutine has no stop path, pre-existing and not fixed here). What changes is the price: one 1ns ticker burns **0.29 of a core**, one 50ms ticker **0.0001**.

## On the numbers I published earlier

I first quoted 21.75s → 6.17s wall and 85.48s → 0.77s CPU, "111× / 3.5× faster". That was one run per side with unequal build caches. I then replaced it with a ratio and said the ratio was "the part that is stable" — which is worse. Re-measured, warm cache both sides, `-run CBOpenRequests -count=20`:

| | `development` | this branch | ratio |
|---|---|---|---|
| run 1 | 78.3s | 2.03s | 38.6× |
| run 2 | 192.7s | 2.01s | 95.9× |
| run 3 | 38.7s | 1.93s | 20.0× |

All the variance is in the numerator, so the ratio inherits every bit of it while the branch side is stable to a few percent. **Wall clock is not a speedup at all** — at `-count=1` this branch is slower, because 50ms of real waiting is 50ms the nanosecond interval skipped. What it buys is a stable runtime instead of a variable one, without burning several cores. The committed comment now quotes the small, reproducible side.

I have also squashed my commits into one; they had become a commit and a commit correcting it, and the earlier one carried the retracted figures into permanent history unmarked.

## On the new test

Reintroducing the #3501 bug fails `TestCircuitBreaker_ConcurrentRecovery_OnlyOneResetsCircuit` **3 of 3**, on the `othersDone` handshake rather than a timing margin.

**In fairness: `TestCircuitBreaker_SlowHealthCheckDoesNotBlock`, already on `development`, kills that same mutant 3/3.** I earlier claimed a TOCTOU mutation separated the two, with a 0/10-vs-10/10 table; re-running it, both catch that mutation 10/10, and a narrower variant separates them only under `-race` — and CI does not run `-race` on this package.

The one thing that does separate them: moving `lastChecked = time.Now()` after `mu.Unlock()` is caught **only** by the new test, and only under `-race` — twenty simultaneous entrants give the detector an interleaving that two goroutines never produce. That is a real if narrow reason to keep it, and it is a better justification than the one the description carried.

The constants themselves are load-bearing, verified: deleting the ten `cbTestRecoveryWait` sleeps fails all ten `CBOpenRequests` tests, and lowering the wait below the interval fails all ten too. ⚠️ Nothing yet stops `Interval: 1` being reintroduced — the suite still passes with it.

## One thing I flagged wrongly

I described `HealthConfig{Timeout: 1}` as "the same bare-integer-as-duration pattern, one file over". It is not — `health_config.go:7` declares `Timeout int` and `new.go:37` documents it as **seconds**, so `Timeout: 1` deliberately means one second. A timing-margin fragility, not a unit confusion.

## Scope and credit

Test-only, one file. The interval work is @adityakrmishra's — `e74745f5d` made all eleven interval changes and all ten sleep changes; my commit names the two values as constants and records what the old interval cost.

## Related, not fixed here

`&CircuitBreakerConfig{Threshold: 5}` reaches `time.NewTicker(0)` and panics with `non-positive interval for NewTicker`, killing the process at startup. Reproduced on `development`, and **unrecoverable** — it is in a spawned goroutine, so a caller's `recover()` never runs. Needs its own PR.

`gofmt` · `go vet` · `golangci-lint --new-from-rev` **0 issues** · `go test -race` clean.

aryanmehrotra (approved):

Umang01-hash (approved):
APPROVE — verified at head a8a0ea2.

Reproduced the problem and the fix myself (not from the writeup):

- Root cause confirmed: Interval: 1 is 1ns, and NewCircuitBreaker feeds it to time.NewTicker in an unstoppable goroutine (circuit_breaker.go:55/115), so each CB the suite builds leaks a ticker spinning for the life of the binary.
- Same GOMAXPROCS=1 -count=3 run on 'TestCircuitBreaker|CBOpenRequests|SuccessRequests': development FAILs at 314s with a goroutine dump stuck in startHealthChecks->isOpen; this branch is ok at 14.5s. Same stress, ~21x faster and green.
- The new TestCircuitBreaker_ConcurrentRecovery_OnlyOneResetsCircuit is revert-red: reintroducing the original lock-held tryCircuitRecovery makes it fail with 'test timed out waiting for losers to finish' (losers block on the mutex the winner holds during healthCheck). Real regression guard for #3501, not mock-pinning. Passed 50x under -race.
- Every Interval: 1 is converted; no bare-int intervals remain in service/. gofmt/vet clean; golangci-lint --new-from-rev = 0 issues.

One follow-up (not blocking, out of scope here): the underlying leak — startHealthChecks has no Stop path, one goroutine per NewCircuitBreaker for the process lifetime — still exists in production code. It's bounded there (one per service, created at startup), so low impact, but worth a tracking issue since this PR fixes the test symptom rather than the leak itself.

PR #3757, opened 2026-07-24
feat(ai): add Embed capability to the LLM

aryanmehrotra:
## What

Adds text embeddings to the `ai` package as a first-class `LLM` method — the peer of `Stream`:

```go
resp, err := ctx.LLM("embed").Embed(ctx, []string{text})
```

`ai.Embedder` is the model-side optional capability a provider implements (the peer of `StreamingModel`). Callers do not implement or assert it — they call `LLM.Embed`, exactly like `LLM.Stream`.

## Why not a caller-side type assertion

An earlier revision exposed this as `ctx.LLM().(ai.Embedder)`. That is wrong: because the instrumented wrapper carries the `Embed` method, the wrapper *always* satisfies `ai.Embedder`, so the assertion's `ok` is always true — it looks like a capability check but isn't — and the one-return form panics on failure. Embeddings are a distinct model class (a chat-only model has no embeddings; an embedding model has no chat), so support is genuinely decided at call time. `LLM.Embed` reports that **gracefully** with `ErrEmbedNotSupported`, mirroring `Stream`/`ErrStreamNotSupported` — never a panic.

## Changes

- **`ai.Embedder` + `ai.EmbeddingResponse`** (`model.go`) — model-side optional capability, peer of `StreamingModel`.
- **`LLM.Embed`** on the interface; the `*llm` wrapper emits an `llm.embed` span and records `app_llm_request_count{operation="embed"}` + token metrics, then delegates to the model's `Embedder`; returns `ErrEmbedNotSupported` when absent.
- **`notConfiguredLLM.Embed`** returns `ErrLLMNotConfigured`; **`MockLLM`** regenerated.
- **`llm.Client.Embed`** — POSTs to the OpenAI-compatible `/embeddings` endpoint; one vector per input, in order. Works with any OpenAI-compatible provider (OpenAI, Ollama, …).
- **Refactor:** `Client.Chat` and `Client.Embed` now share a small `postJSON` helper (send → read → status) instead of duplicating the transport/error path. Pure refactor of `Chat`, covered by its existing tests.

## Tests

- wrapper: delegation + input forwarding + success metric; unsupported → `ErrEmbedNotSupported` (graceful, error metric); underlying model error propagated.
- client: success (vectors + usage + request shape), provider error in body, malformed JSON, status error, not-connected.

`go build ./...`, `go test ./pkg/gofr/ai/... ./pkg/gofr/container/...`, `go vet`, and `golangci-lint run ./pkg/gofr/ai/...` (0 issues) all pass.

## Compatibility

`LLM` gains a method (`Embed`), as it did for `Stream`; GoFr's own implementations (`*llm`, `notConfiguredLLM`, `MockLLM`) are all updated. Providers that don't embed need no change — `Embed` reports `ErrEmbedNotSupported`.


pkg/gofr/ai/llm/client.go
```diff
@@ -217,6 +209,43 @@ func (c *Client) Chat(ctx context.Context, messages []ai.Message, opts ...ai.Opt
 	return out, nil
 }
 
+// Embed satisfies ai.Embedder: it posts the input texts to the OpenAI-compatible /embeddings
+// endpoint and returns one vector per input, in the same order. It rides the same instrumented
+// HTTP service, retry and error handling as Chat.
+func (c *Client) Embed(ctx context.Context, input []string, _ ...ai.Option) (*ai.EmbeddingResponse, error) {
```
Umang01-hash, line 259: `opts ...ai.Option` is discarded here (`_`), and `embeddingsRequest` only carries `{model, input}` — so `dimensions` / `encoding_format` can't be set (the former matters for `text-embedding-3-*` Matryoshka truncation). Signature + docs imply options are honored. Fine to defer, but worth a `// ponytail:` comment naming the ceiling, or a follow-up issue. Also: the docs Traces section lists chat/generate/stream spans but omits `llm.embed`.

Umang01-hash (dismissed):
LGTM. Clean addition — provider side is done right (`Embedder` is optional/type-asserted, not forced onto `Model`), instrumentation matches Chat/Stream (`opEmbed`, Usage + not-supported error path), and the `postJSON` refactor cleanly shares logic with Chat.

Verified locally: gofmt/vet/build + `-race` clean, ai/llm coverage 89.8%, and an E2E against a live mock `/embeddings` server round-tripped 2 vectors in input order with usage parsed.

On the breaking-change angle: adding `Embed` to the exported `ai.LLM` interface is technically source-breaking for external implementers, but it's low-risk here — the package is ~6 days old, `LLM` is a consumer facade, and all three implementers (`llm`, `notConfiguredLLM`, `mock_ai`) are updated. Worth a conscious note that `LLM` will keep growing methods this way (each one breaks external fakes), but that's fine in the shaping window.

Two small non-blocking notes below + docs.

aryanmehrotra:
Both points addressed in 2303b00a, plus the interface fix in 28b01f3c.

**Options on `Embed`** — you're right that the signature and docs implied they were honored. I kept the discard (every option we define today — `WithTemperature`, `WithMaxTokens`, `WithTools` — is a completion parameter with no meaning for an embeddings request, so honoring them would be worse than dropping them), but named the ceiling on `Embed` itself rather than leaving it implicit, and filed #3803 to lift it.

Worth recording why `dimensions` in particular is not something callers can just work around: OpenAI renormalizes after truncating, so a client-side slice of the full vector is unnormalized and skews cosine similarity. #3803 also flags the design question this opens — `ai.Option` is one flat type across chat and embeddings, so a future `WithDimensions` would be silently ignored by `Chat` exactly the way `WithTemperature` is ignored by `Embed`. Same trap, other direction; worth settling before adding the first embedding-specific option.

**Traces docs** — added `llm.embed` to the span list, and noted that only prompt tokens are non-zero on it since embeddings bill input alone.

One thing I changed beyond your comments, in 28b01f3c: `Embed` was declared on the `ai.LLM` interface, but `LLM` documents a compatibility freeze — capabilities are meant to arrive as optional interfaces asserted by the caller, never as new methods, so hand-written fakes keep compiling. `LLM` shipped in v1.58.0, so any fake written against it would have stopped compiling on upgrade. Moved the capability to an `ai.EmbeddingLLM` optional interface; both the instrumented wrapper and the not-configured placeholder implement it, so the assertion never fails in a handler and the real condition still surfaces from the call (`ErrEmbedNotSupported` / `ErrLLMNotConfigured`). `compat_test.go` now pins the freeze with a fake implementing exactly `LLM`, so this repo stops compiling if anyone adds a capability method again.

Also guarded a nil provider response in `Embed` — it dereferenced before checking, so a third-party `Embedder` returning `(nil, nil)` panicked the handler, where the `Chat` path already tolerates it.

Umang01-hash (commented):
Reviewed end to end at `60c4bbcc` — the interface-freeze design is the right call and it's actually enforced: `compat_test.go`'s `legacyFake` breaks compilation in-repo if a method is ever added to `LLM`. Concurrency limiter is leak-free across every stream-termination path and `-race` is clean; nil-response and error-object-on-200 paths are tested; Embed happy-path is non-vacuous (breaking the mapping reds it). Nice work.

One thing worth hardening, then two nits:

1. **Embeddings are mapped by position and `data[].index` is dropped** (`client.go` Embed loop, `embeddingDatum` in `wire.go`). The response carries `index` (the tests even send it) but the vector is taken by array position. For OpenAI itself that's fine — but this client targets any OpenAI-compatible backend, and `index` exists precisely so a provider can return `data` out of order. If one does, each input silently gets the wrong vector, which is a rough failure mode for embeddings (semantic search / agent memory) since nothing errors. ~4 lines: decode `Index` and place by it (guard bounds + dup), or assert `index == i`. `page.md` ("one []float32 per input, in order") inherits the same assumption.

2. **NIT** — `HealthCheck` holds `healthMu` across the network probe, so a concurrent caller waits up to `healthProbeTimeout` instead of getting the cached value. Probably fine (avoids a probe stampede), just flagging.

3. **NIT** — `TestMockModel_SatisfiesInterfaces` asserts `MockLLM` against LLM/Model/StreamingModel but not `EmbeddingLLM` (would pass — `Embed` exists). Worth adding for symmetry with the `*llm`/`notConfiguredLLM` assertions.

No blockers — deferring `dimensions`/`encoding_format` to #3803 is the right scope.

aryanmehrotra:
All three addressed in 4319d63b6.

**1. `data[].index` dropped** — you're right, and it's the one that would have hurt: a wrong vector is still a valid vector, so nothing downstream errors. `embeddingDatum` now decodes `Index` and `placeEmbeddings` places by it.

Two things I did differently from the ~4 lines, both because the short version has a trap:

`Index` is a `*int`, not an `int`. A plain int makes an omitted `index` decode as 0 for *every* entry, so the minimal OpenAI-compatible providers that skip the field would go from working to collapsing all vectors onto input 0 — a regression introduced by the fix for the bug. Nil means "use my position", which is exactly today's behavior for those providers.

And the bounds are checked against `len(input)`, not `len(er.Data)`. Checking against the response's own length looks equivalent but isn't, and it left the truncated-response case decided by luck: for 3 inputs, a provider returning entries `0, 2` errors (index 2 is out of range for a 2-element array) while one returning `0, 1` maps cleanly and hands back a 2-element slice. Same defect, opposite outcomes, depending on which input got dropped — and the silent branch is the bad one, since a caller doing `store(texts[i], resp.Embeddings[i])` panics and one ranging over the result quietly skips a document. Validating against the inputs sent makes it one rule: count mismatch, index out of range, or index claimed twice all return `errDecodeResponse` rather than being half-mapped. `len(resp.Embeddings) == len(input)` now actually holds whenever `err` is nil, which is what `page.md` already promised. The endpoint has no partial-success mode — failures are reported for the whole request, which the `er.Error` path already handles — so this rejects nothing legitimate.

`page.md` updated to state the guarantee and why, rather than leaving "in order" as an assumption inherited from OpenAI's own behavior.

Tests: out-of-order data, missing index falling back to position, and a table for the four unmappable shapes. Verified non-vacuous — reverting `placeEmbeddings` to positional mapping reds the ordering tests, and dropping the count check reds exactly the two count subtests and nothing else.

**2. `HealthCheck` holding `healthMu` across the probe** — kept, but the reasoning is now in the code instead of implied, since you're the second reader to stop on it. Releasing the lock around the probe turns a cold-cache burst into one provider request per caller; holding it makes the probe single-flight. The cost is that concurrent callers wait on the in-flight probe rather than racing their own, bounded by `healthProbeTimeout` — the right trade for something polled by `/.well-known/health` rather than served on a request path.

**3. `TestMockModel_SatisfiesInterfaces`** — added the `EmbeddingLLM` assertion. Worth having for a real reason rather than symmetry: `MockLLM` stands in for the `*llm` wrapper in handler tests, so it has to satisfy the optional capabilities a handler asserts on the LLM, not just `LLM` itself. If a future capability lands on the wrapper and not the mock, every handler test asserting it fails with a nil `ok` and no hint why.

`go build ./...`, `go vet`, `go test ./pkg/gofr/ai/... -race`, and `golangci-lint run ./pkg/gofr/ai/...` (0 issues) all pass.

Umang01-hash (dismissed):
Re-reviewed at `4319d63` — all three points are addressed, and the index fix is more thorough than I flagged. `placeEmbeddings` honors `data[].index` (pointer, so absent falls back to position), and guards out-of-range/negative/duplicate index plus `len(data) != inputs` — that last one catches a short response before it can hand the caller a slice shorter than its inputs, which I hadn't even called out. Count + uniqueness + range together mean no nil slots.

Tested out-of-order, missing-index fallback, and a 5-case unmappable table, and I confirmed it revert-red: reverting to positional mapping fails `TestClient_Embed_OutOfOrderData`. Baseline + vet clean. The HealthCheck lock now documents the deliberate single-flight trade, and the `EmbeddingLLM` mock assertion is in.

No breaking change, clean and well-tested. Approving.

Umang01-hash (approved):

PR #3758, opened 2026-07-24
perf(logging,http): cut per-request allocations on the hot path

aryanmehrotra:
## Summary

Follow-up allocation cuts on the `pkg/gofr` request hot path, on top of the #3431 stack. Four focused changes in logging + HTTP response, each guarded by a benchmark and an equivalence test. No public-API, wire-shape, or env-var change.

`benchstat` (n=10, Apple M4, `GOMAXPROCS=2`, bounded container):

| benchmark            |   time   |    B/op    |  allocs/op |
|----------------------|---------:|-----------:|-----------:|
| `ResponderRespond`   | 380 → 334 ns (−12 %) | 249 → 72 (−71 %) | 4 → 2 |
| `getIPAddress`       | 59 → 26 ns (−56 %)   | 48 → 0 (−100 %)  | 1 → 0 |
| `ContextLogger.Info` | 584 → 461 ns (−21 %) | 568 → 216 (−62 %)| 9 → 6 |
| `Logger.Info` (JSON) | 513 → 393 ns (−23 %) | — | — |

Roughly −577 B and −6 allocs on a typical request (one access log + one response + one XFF parse).

## What's included

- **logging** — zero-alloc fast path in `extractTraceIDAndFilterArgs`: the vast majority of log calls carry no `__trace_id__` marker, so a read-only scan returns the args slice untouched instead of always copying it into a fresh `filtered` slice.
- **logging** — precompute the `{"__trace_id__": id}` marker map once per `ContextLogger` (i.e. once per request) instead of rebuilding it on every log call. Pulled the literal into a `traceIDMarkerKey` const while at it.
- **middleware** — `getIPAddress` reads the first `X-Forwarded-For` entry with `strings.IndexByte` instead of `strings.Split`, dropping the `[]string` allocated per request just to take `[0]`.
- **http** — `Responder.Respond` encodes into a pooled `bytes.Buffer` and does a single `Write`, avoiding the fresh `[]byte` that `json.Marshal` returns each time. The pool caps returned buffers at 64 KB so one large response can't inflate the pool.

## Behaviour audit

No observable change — these are pure allocation cuts.

- Response bytes are **byte-identical** to the previous `json.Marshal(resp)` + `"\n"` path (asserted by `TestResponder_ByteIdenticalToMarshal`, including the HTML-escaping case), and the pooled buffer is safe under concurrency (`-race`, 500 goroutines, no cross-request bleed).
- The `__trace_id__` marker contract (produced by `ContextLogger`, consumed in `extractTraceIDAndFilterArgs`) is preserved; the marker is stripped before formatting and never reaches the wire.
- `getIPAddress` is proven equivalent to the old `strings.Split` version across empty / single / multi-IP / leading-comma / trailing-comma / whitespace-only / IPv6 inputs (oracle test).

No new env vars, no flags, no interface changes.

## Tests

`go test -race ./pkg/gofr/logging/... ./pkg/gofr/http/...` green. Added benchmarks + equivalence/concurrency tests for each change; `golangci-lint` clean under `only-new-issues`.

### DIFF
--- a/pkg/gofr/http/middleware/logger.go
+++ b/pkg/gofr/http/middleware/logger.go
@@ -251,11 +251,17 @@ func isLogProbeDisabled(probes LogProbes, urlPath string) bool {
 }
 
 func getIPAddress(r *http.Request) string {
-	ips := strings.Split(r.Header.Get("X-Forwarded-For"), ",")
-
 	// According to GCLB Documentation (https://cloud.google.com/load-balancing/docs/https/), IPs are added in following sequence.
 	// X-Forwarded-For: <unverified IP(s)>, <immediate client IP>, <global forwarding rule external IP>, <proxies running in GCP>
-	ipAddress := ips[0]
+	//
+	// We only need the first entry, so take it with IndexByte instead of
+	// strings.Split, which would allocate a []string on every request.
+	xff := r.Header.Get("X-Forwarded-For")
+
+	ipAddress := xff
+	if i := strings.IndexByte(xff, ','); i >= 0 {
+		ipAddress = xff[:i]
+	}
 
 	if ipAddress == "" {
 		ipAddress = r.RemoteAddr
--- a/pkg/gofr/http/middleware/logger_test.go
+++ b/pkg/gofr/http/middleware/logger_test.go
@@ -3,6 +3,7 @@ package middleware
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"encoding/json"
 	"net"
 	"net/http"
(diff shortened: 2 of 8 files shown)


Umang01-hash (dismissed):
LGTM — clean, behavior-preserving hot-path perf work, verified with real numbers (re-checked at the latest commit; source is unchanged from my review, tests were just reorganized into the standard *_test.go files).

**benchstat** (baseline vs head, -count=10, p=0.000):
- `Responder.Respond`: allocs 4→2 (**-50%**), B/op 249→72 (**-71%**), -5% ns/op
- `getIPAddress`: 1→0 allocs (**zero-alloc**), 48→0 B/op, -60% ns/op
- `ContextLogger.Info`: allocs 9→6 (**-33%**), B/op 568→216 (**-62%**), -14.5% ns/op
- plain logger paths: unchanged (no regression)

**Live load** (`hey -c50 -z15s` on http-server /hello, 2 rounds reversed order): **+9% req/s** (~90k→~98k), p99 3.0ms→2.1ms.

Wire output is byte-identical (`json.Encoder` = `Marshal`+`\n`, same HTML escaping — confirmed by hexdump + `TestResponder_ByteIdenticalToMarshal`), no cross-request bleed in the buffer pool (Reset on every Get, synchronous Write, `-race` clean incl. 500-concurrent test), the shared trace-marker map is read-only and the logger is synchronous, and no breaking API change. Nice work.

(Unrelated: `-race` surfaced a pre-existing data race in `remotelogger` `TestRemoteLogger_UpdateLevel` — reproduces on the parent commit too, so out of scope here; worth a separate issue.)

Umang01-hash (approved):

PR #3759, opened 2026-07-24
perf(http): opt-in O(1) trie router behind GOFR_ROUTER

aryanmehrotra:
## Motivation

GoFr routes on `gorilla/mux`, which matches with an **O(n) linear scan + per-route regexp**, so per-request routing cost climbs with the number of registered routes. Services with large route tables pay for this on every request — measured below, a 200-route service spends ~2.9 µs per request just locating the handler, and it keeps growing with the table.

This is **not** about dropping `gorilla/mux` — mux stays the route registry, keeps `mux.Vars`, route templates, and the whole registration API. It's purely about the matching cost. mux's `ServeHTTP` is exported/overridable, so this PR overrides just the matching step with a **radix trie** to make it O(path length), and then hands the actual match decision back to mux's own `Route.Match` — so mux's semantics are still what decide every request.

It is **strictly opt-in** (`GOFR_ROUTER=trie`). Default behaviour is byte-for-byte unchanged. Kept opt-in on purpose: route matching has a long tail of edge cases, so this ships as a switch you turn on, with the option to flip the default in a later release once it has mileage.

Documented in [Routing Performance](docs/advanced-guide/routing-performance/page.md) and in the configuration reference. The server states the matcher at startup when the trie is active, and warns on a value it does not recognize — that case falls back to mux, which is otherwise indistinguishable from never setting the variable, so a typo would cost the opt-in silently.

## Performance

`BenchmarkRouterMatch`, Apple M4. The request hits the **middle** route — mux's cost is linear in registration position, so measuring the last route would report its worst case rather than what real traffic sees. GoFr's default `PathPrefix("/")` catch-all is registered too, so the numbers reflect a real route table.

| routes | mux | trie | speedup |
|-------:|--------:|--------:|--------:|
|      1 |  431 ns |  447 ns | 0.96x |
|     10 |  506 ns |  461 ns | 1.1x |
|     20 |  647 ns |  531 ns | 1.2x |
|     50 | 1007 ns |  550 ns | 1.8x |
|    100 | 1642 ns |  548 ns | 3.0x |
|    200 | 2918 ns |  515 ns | **5.7x** |

mux grows linearly; the trie is flat. **Be clear about where this helps:** the crossover is at roughly 5–10 routes, so a small service sees nothing, and at a single route the trie is marginally slower. The win scales with the route table, which is where large services live. The trie also costs **+2 allocations per request** (10 vs 8) for the candidate set.

Scope note: this optimises route matching only. Matching is a minority of total request cost — the middleware chain (tracing, logging, metrics, CORS) dominates a GoFr request — so the end-to-end RPS change is smaller than the table above and approaches it only as the route count grows.

## Design

- **mux stays the registry.** The trie is an index built once (lazily, on the first request) by `Walk`-ing the routes mux already has, so registration (`Add` / `NewRoute` / `PathPrefix` / `AddStaticFiles`) is untouched.
- **matching** — narrow via the trie, append the fallback list, order by registration, return the first candidate mux's `Route.Match` accepts. A route the trie omits could not have matched the path anyway, so the result is the same route mux's full scan would pick.
- **The classifier is an allowlist.** A route is trie-indexed only when its compiled path regexp can be *proven* safe: anchored at both ends on the parsed syntax tree, with every capture matching a non-empty, slash-free fragment. Anything unproven — `PathPrefix`/static, slash-spanning `{p:.*}`, empty-matching `{p:[a-z]*}` — goes to the fallback list where mux decides it. The asymmetry drives the direction: a route wrongly *excluded* only loses speed, while a route wrongly *included* could be dropped from the candidate set and silently 404.
- **`mux.Vars` / route template** — since the trie bypasses mux's `ServeHTTP`, `mux.CurrentRoute` is no longer populated. Params are restored with `mux.SetURLVars` (so `mux.Vars` is unchanged) and the template is carried in a context key, read through `RouteTemplate(r)`, which resolves under both routers.
- **middleware** — applied by the router itself, in the same order mux applies it, and only around a matched handler (mux never wraps its NotFound/MethodNotAllowed handlers, so neither does this).
- **anything unmatched goes back to mux.** `serveTrie` does not decide 404-vs-405 itself; it hands every non-match to mux's own `ServeHTTP`. That branch is the cold path — the response is already an error — so the linear scan costs nothing that matters, and it makes the index *self-healing*: if the classifier ever admitted a shape it should not have and the trie dropped a route that does match, mux's full scan finds and serves it. The one catastrophic failure mode of this design — a live route silently 404ing — becomes a request that is merely slower, leaving the trie strictly an accelerator.

## Behaviour audit

With `GOFR_ROUTER` unset there is no behaviour change — same code path as today.

With `GOFR_ROUTER=trie`, a differential suite registers an identical route set on both routers and requires byte-identical `{status, body, vars}` across static / param / `{id:regex}` / mixed literal+param (`{name}.txt`, `user-{id}`) / overlap ordering / normalisation / slash-spanning / subrouter / encoded-path / middleware-invocation scenarios. On top of that, an invariant test asserts the property the design rests on: **no route mux matches is ever absent from the trie's candidate set.**

**There is now no divergence at all** — the routers agree on every request, in a running app and on a bare `Router` alike.

An earlier revision had one: for a wrong-method request to an existing path the trie returned 405 where mux returns 404. That turned out not to be a bookkeeping bug but a structural one. In gorilla/mux's `Route.Match`, a matcher that *succeeds* clears any `ErrMethodMismatch` an earlier route left behind, and GoFr registers routes as `Methods(m).Path(p)` — method matcher first. So a later route whose path does **not** match the request can still erase the mismatch purely because its method does, which is why mux's 404-vs-405 is registration-order dependent. Reproducing that outcome requires visiting exactly the routes the trie exists to skip, so no amount of bookkeeping over the narrowed candidate set can get there.

Delegating the unmatched branch to mux resolves it exactly rather than approximately, and brings the custom `NotFoundHandler` / `MethodNotAllowedHandler` and subrouter semantics along for free. The differential test that used to pin the divergence now pins the parity.

The index is built once from the routes present at the first request. GoFr registers every route during startup, before the server accepts requests, so this holds for all framework code paths; a route added after the first request would not be indexed (documented on the field).

## Tests

`go test -race ./pkg/gofr/http/...` green in **both** modes; full framework suite green in both modes; `golangci-lint` clean under `only-new-issues`. Each edge-case test was mutation-checked — it fails against the previous implementation and passes against this one.

### DIFF
--- a/pkg/gofr/http/route_context.go
+++ b/pkg/gofr/http/route_context.go
@@ -1,0 +1,45 @@
+package http
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/gorilla/mux"
+)
+
+// routeTemplateCtxKey is the private context key under which the trie router
+// stores the matched route template (e.g. "/users/{id}"). It exists because,
+// once the trie router bypasses mux's ServeHTTP, mux.CurrentRoute(r) is no
+// longer populated, so the template must be carried some other way.
+type routeTemplateCtxKey struct{}
+
+// withRouteTemplate returns r carrying tmpl as the matched route template. An
+// empty template is a no-op so unmatched requests do not pay for a context copy.
+func withRouteTemplate(r *http.Request, tmpl string) *http.Request {
+	if tmpl == "" {
+		return r
+	}
+
+	return r.WithContext(context.WithValue(r.Context(), routeTemplateCtxKey{}, tmpl))
+}
+
+// RouteTemplate returns the matched route template for r (e.g. "/users/{id}"),
+// or "" if none is available. It is the router-agnostic replacement for
+// mux.CurrentRoute(r).GetPathTemplate() and works under both routers: the trie
+// router records the template in the request context (it bypasses mux's
+// ServeHTTP, so mux.CurrentRoute is nil), while the default mux path exposes it
+// via mux.CurrentRoute. Callers get consistent behavior without knowing which
+// router is active.
+func RouteTemplate(r *http.Request) string {
+	if tmpl, _ := r.Context().Value(routeTemplateCtxKey{}).(string); tmpl != "" {
+		return tmpl
+	}
+
+	if route := mux.CurrentRoute(r); route != nil {
+		if tmpl, err := route.GetPathTemplate(); err == nil {
+			return tmpl
+		}
+	}
+
+	return ""
+}
--- a/pkg/gofr/http/router.go
+++ b/pkg/gofr/http/router.go
@@ -20,6 +20,16 @@ const (
 	DefaultSwaggerFileName       = "openapi.json"
 	staticServerNotFoundFileName = "404.html"
 	staticServerIndexFileName    = "index.html"
+
+	// RouterEnvVar selects the route matcher. Unset (or any unrecognized value)
+	// means MatcherMux, so the default behavior is unchanged.
+	RouterEnvVar = "GOFR_ROUTER"
+
+	// MatcherMux is gorilla/mux's linear scan — the default.
+	MatcherMux = "mux"
+	// MatcherTrie is the opt-in segment-trie index, O(path length) in the number
+	// of registered routes.
+	MatcherTrie = "trie"
 )
 
 // errReadPermissionDenied wraps fs.ErrPermission so that a file whose mode carries no read bit is
@@ -67,7 +77,7 @@ func NewRouter() *Router {
 	r := &Router{
 		Router:           *muxRouter,
 		RegisteredRoutes: &routes,
-		useTrie:          strings.EqualFold(os.Getenv("GOFR_ROUTER"), "trie"),
+		useTrie:          strings.EqualFold(os.Getenv(RouterEnvVar), MatcherTrie),
 	}
 
 	return r
@@ -130,6 +140,9 @@ func normalizePath(r *http.Request) {
 // Tracer/Logging/CORS/Metrics chain running. serveTrie mirrors that exactly, so
 // unmatched requests are not logged, measured, or CORS-answered (and the metrics
 // path label is never populated from a raw, unbounded request path).
+//
+// Anything the trie does not match is handed to mux's own ServeHTTP rather than
+// resolved here — see the comment on that call for why.
 func (rou *Router) serveTrie(w http.ResponseWriter, r *http.Request) {
 	rou.buildIdx.Do(func() {
 		idx := newRouteIndex()
@@ -158,16 +171,31 @@ func (rou *Router) serveTrie(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Unmatched: serve mux's NotFound/MethodNotAllowed handler directly, WITHOUT
-	// the middleware chain — matching mux, which applies middleware only on a
-	// successful match.
-	if errors.Is(match.MatchErr, mux.ErrMethodMismatch) {
-		rou.methodNotAllowed().ServeHTTP(w, r)
-
-		return
-	}
-
-	rou.notFound().ServeHTTP(w, r)
+	// Unmatched by the trie: hand the request to mux's own ServeHTTP instead of
+	// deciding 404-vs-405 here. This is the cold path — the response is already
+	// an error — so mux's linear scan costs nothing that matters, and it buys two
+	// things that reimplementing the decision cannot.
+	//
+	// Exactness. mux's 405 depends on state that routes which do NOT match the
+	// request path still mutate: a later route whose method matcher succeeds
+	// clears an ErrMethodMismatch left by an earlier one (gorilla/mux route.go),
+	// and GoFr registers routes as Methods(m).Path(p), so the method matcher runs
+	// first. Reproducing that outcome therefore requires visiting routes the trie
+	// exists to skip. Delegating gets it exactly right instead of approximately.
+	// The custom NotFoundHandler / MethodNotAllowedHandler and subrouter
+	// semantics come along for free, and mux applies no middleware on this path,
+	// so the parity the chain relies on is mux's own by construction.
+	//
+	// Safety. It also makes the index self-healing: if isIndexablePathRegexp ever
+	// admitted a shape it should not have and the trie dropped a route that does
+	// match, mux's full scan finds it here and serves it correctly. That turns the
+	// one catastrophic failure mode of this design — a live route silently
+	// 404ing — into a request that is merely slower, leaving the trie strictly an
+	// accelerator.
+	//
+	// Inside a GoFr app this is unreachable: the PathPrefix("/") catch-all matches
+	// every path and method, so the trie always has a candidate that matches.
+	rou.Router.ServeHTTP(w, r)
 }
 
 // Use registers mux middlewares. It records them in GoFr's own chain — so the
@@ -179,26 +207,20 @@ func (rou *Router) Use(mwf ...mux.MiddlewareFunc) {
 	rou.Router.Use(mwf...)
 }
 
-// notFound returns the handler mux would use for an unmatched route, honoring a
-// custom NotFoundHandler if one was set.
-func (rou *Router) notFound() http.Handler {
-	if rou.NotFoundHandler != nil {
-		return rou.NotFoundHandler
+// Matcher reports which route matcher this router uses: MatcherTrie for the
+// opt-in index, MatcherMux for the default linear scan.
+//
+// It is exported so the server can state the active matcher at startup. The
+// choice is made from the environment inside NewRouter, and an unrecognized
+// GOFR_ROUTER value falls back to mux — which is indistinguishable, from the
+// outside, from not setting the variable at all. Reporting the resolved matcher
+// is what lets the caller tell a typo from a default.
+func (rou *Router) Matcher() string {
+	if rou.useTrie {
+		return MatcherTrie
 	}
 
-	return http.NotFoundHandler()
-}
-
-// methodNotAllowed returns the handler mux would use when the path matches but
-// the method does not, honoring a custom MethodNotAllowedHandler if set.
-func (rou *Router) methodNotAllowed() http.Handler {
-	if rou.MethodNotAllowedHandler != nil {
-		return rou.MethodNotAllowedHandler
-	}
-
-	return http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
-		w.WriteHeader(http.StatusMethodNotAllowed)
-	})
+	return MatcherMux
 }
 
 // composeMiddleware wraps h with mws so that mws[0] is the outermost layer,
--- a/pkg/gofr/http/router.go
+++ b/pkg/gofr/http/router.go
@@ -153,20 +153,7 @@ func (rou *Router) serveTrie(w http.ResponseWriter, r *http.Request) {
 	var match mux.RouteMatch
 
 	if rou.idx.match(r, &match) && match.Handler != nil {
-		// Reinstate what mux.Router.ServeHTTP would have populated so that
-		// mux.Vars(r) (used by request.go and user handlers) and the route
-		// template (used by the tracer/metrics middleware) keep working.
-		if match.Vars != nil {
-			r = mux.SetURLVars(r, match.Vars)
-		}
-
-		if match.Route != nil {
-			if tmpl, err := match.Route.GetPathTemplate(); err == nil {
-				r = withRouteTemplate(r, tmpl)
-			}
-		}
-
-		composeMiddleware(rou.mws, match.Handler).ServeHTTP(w, r)
+		rou.serveMatched(w, r, &match)
 
 		return
 	}
@@ -196,6 +183,37 @@ func (rou *Router) serveTrie(w http.ResponseWriter, r *http.Request) {
 	// Inside a GoFr app this is unreachable: the PathPrefix("/") catch-all matches
 	// every path and method, so the trie always has a candidate that matches.
 	rou.Router.ServeHTTP(w, r)
+}
+
+// serveMatched runs a handler the trie matched, after restoring the request state that mux's own
+// ServeHTTP would have populated.
+func (rou *Router) serveMatched(w http.ResponseWriter, r *http.Request, match *mux.RouteMatch) {
+	// Reinstate what mux.Router.ServeHTTP would have populated so that mux.Vars(r) (used by
+	// request.go and user handlers) and the route template (used by the tracer/metrics middleware)
+	// keep working.
+	if match.Vars != nil {
+		r = mux.SetURLVars(r, match.Vars)
+	}
+
+	if match.Route != nil {
+		if tmpl, err := match.Route.GetPathTemplate(); err == nil {
+			r = withRouteTemplate(r, tmpl)
+		}
+	}
+
+	// mux builds its middleware chain inside Match, guarded by MatchErr == nil, so a route that
+	// matches while REPORTING an error is served WITHOUT the chain. A subrouter carrying its own
+	// NotFoundHandler is exactly that case: it reports a successful match with ErrNotFound. Running
+	// the chain here would log, trace, meter and CORS-answer a request mux does not — and with no
+	// path template on a PathPrefix route, the metrics label would fall back to the raw request
+	// path, the unbounded-cardinality outcome this router is otherwise careful to avoid.
+	if match.MatchErr != nil {
+		match.Handler.ServeHTTP(w, r)
+
+		return
+	}
+
+	composeMiddleware(rou.mws, match.Handler).ServeHTTP(w, r)
 }
 
 // Use registers mux middlewares. It records them in GoFr's own chain — so the
--- a/pkg/gofr/http/router.go
+++ b/pkg/gofr/http/router.go
@@ -9,6 +9,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/gorilla/mux"
 
@@ -30,6 +31,31 @@ var errReadPermissionDenied = fmt.Errorf("file does not have read permission: %w
 type Router struct {
 	mux.Router
 	RegisteredRoutes *[]string
+
+	// useTrie selects the O(path) trie matcher (GOFR_ROUTER=trie) over mux's
+	// default O(n) linear scan. When false, ServeHTTP delegates to mux exactly
+	// as before, so the default behavior is byte-for-byte unchanged.
+	useTrie bool
+	// idx is the trie index. It is built once, lazily, on the first request,
+	// from the routes registered up to that point. This is correct for GoFr's
+	// lifecycle: every route is registered during startup (app.GET/POST/...,
+	// the GraphQL route, the static/catch-all handlers) before the server
+	// accepts its first request, and GoFr does not add routes afterwards. A
+	// route registered after the first request would not be reflected in the
+	// trie index — a deliberate trade for a lock-free steady state, matching
+	// GoFr's static-routing model. buildIdx guards that one-time build.
+	idx      *routeIndex
+	buildIdx sync.Once
+	// mws mirrors the middleware chain registered via Use, so the trie matcher
+	// can apply it itself when it bypasses mux's ServeHTTP. In mux mode it is
+	// unused (mux owns the chain) but kept in sync, costing nothing.
+	//
+	// Invariant: every middleware MUST be registered through (*Router).Use (or
+	// UseMiddleware, which calls it). A direct call to the embedded
+	// mux.Router.Use would bypass this slice and be silently dropped in trie
+	// mode. All framework registration paths go through (*Router).Use, and the
+	// MiddlewareParity differential test guards the resulting behavior.
+	mws []mux.MiddlewareFunc
 }
 
 type Middleware func(handler http.Handler) http.Handler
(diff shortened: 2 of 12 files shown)


pkg/gofr/http/router.go
```diff
@@ -35,43 +56,148 @@ func NewRouter() *Router {
...
+	}
+
+	rou.notFound().ServeHTTP(w, r)
+}
+
+// Use registers mux middlewares. It records them in GoFr's own chain — so the
+// trie matcher can apply them when it bypasses mux's ServeHTTP — and delegates
+// to the embedded mux router, leaving the default (mux) path unchanged. It
+// shadows mux.Router.Use for calls made on *Router.
+func (rou *Router) Use(mwf ...mux.MiddlewareFunc) {
```
thzgajendra, line 227: Routing the mux chain through this shadow to keep `mws` in sync for trie mode is the right call. Robustness note: the correctness of trie mode now rests on an **implicit invariant** — every middleware must be registered through `(*Router).Use`, never the embedded `mux.Router.Use`. I checked and nothing currently bypasses it, but a future direct `.Router.Use(...)` call would silently drop that middleware in trie mode with no error or test failure at registration. Consider a short doc-comment on the `mws` field (or the embedded router) stating the invariant. The `MiddlewareParity` differential test guards the behavior, which mitigates this well.

pkg/gofr/http/trie_router.go
```diff
@@ -0,0 +1,300 @@
...
+// whole set by registration order, and returns mux's own match for the first
+// candidate that fully matches — identical to what stock mux would pick, only
+// without scanning every registered route.
+//
+// It reports true and fills rm on a full match. On no match it returns false;
+// if some candidate matched the path but not the method, rm.MatchErr is set to
+// mux.ErrMethodMismatch so the caller can render 405 rather than 404.
+func (idx *routeIndex) match(req *http.Request, rm *mux.RouteMatch) bool {
+	cands := idx.candidates(req.URL.Path)
+	cands = append(cands, idx.fallback...)
```
thzgajendra: Every request appends the full fallback list before matching, so routes that can't be trie-indexed (PathPrefix/static handlers, slash-spanning params like `{path:.*}`) stay O(n) in *their own* count. That's fine for typical apps (few such routes), but an app that registers many catch-all/proxy routes reintroduces linear cost for those. Worth a sentence in the design notes so the "flat at 1,000 routes" result is understood to apply to trie-indexable routes — the fallback set is assumed small.

pkg/gofr/http/trie_router.go
```diff
@@ -0,0 +1,300 @@
...
+
+			return true
+		}
+
+		if errors.Is(m.MatchErr, mux.ErrMethodMismatch) {
+			methodMismatch = true
+		}
+	}
+
+	if methodMismatch {
```
thzgajendra: On the documented 404→405 divergence: inside the framework it's effectively **unobservable**. GoFr always registers a `PathPrefix("/")` catch-all (`gofr.go:180`) that matches any method, so a wrong-method request to an existing path is caught by the catch-all — a full match is found in this loop before `methodMismatch` is ever used — yielding 404 in *both* mux and trie modes. The 405 path here only triggers on a bare router (i.e. the unit tests). The behavior audit is accurate, but a line clarifying this is framework-moot would prevent reviewers from expecting a live 404→405 change.

thzgajendra (commented):
Reviewed in depth against the codebase (not just the description). **This is a correct, well-engineered change and it achieves its stated goal** — O(path) matching, opt-in, with mux kept as the source of truth so every routing semantic is still decided by `mux.Route.Match`.

What I verified beyond the PR's own claims:

- **`mux.CurrentRoute` bypass is fully covered.** The only two framework consumers are `metrics.go` and `tracer.go`, and both now prefer `RouteTemplate(r)` with a `mux.CurrentRoute` fallback. No other consumer exists, so nothing silently breaks in trie mode. `mux.Vars` is restored via `SetURLVars`.
- **Middleware is correctly re-applied in trie mode.** Every registration path (Tracer/Logging/CORS/Metrics at `http_server.go:36`, WSHandlerUpgrade at `:57`, auth, RBAC, dbresolver) flows through the shadowed `(*Router).Use`; nothing bypasses it via the embedded `mux.Router.Use`. Order matches mux (mws[0] outermost).
- **Concurrency** — lazy build guarded by `sync.Once`, so `idx` reads happen-after the write.
- **No import cycle** despite `middleware` importing `pkg/gofr/http`.
- Edge cases traced by hand (root `/`, segment-count mismatches, static-vs-param overlap ordering, slash-spanning → fallback, normalization done once before both matchers). The `regexp/syntax` slash-in-capture check over-approximates safely.

The differential/characterization test suite is the right approach and gives real confidence.

Three non-blocking notes inline (an implicit invariant worth documenting, an O(n) caveat for un-indexable routes, and a clarification that the 404→405 divergence is effectively moot inside the framework). Nice work.

akshat-kumar-singhal:
@aryanmehrotra Consider making this performance fix available for all and just keep the 404->405 breaking change behind the env toggle.

pkg/gofr/http/route_context.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// RouteTemplate returns the matched route template for r (e.g. "/users/{id}"),
+// or "" if none is available. It is the router-agnostic replacement for
+// mux.CurrentRoute(r).GetPathTemplate() and works under both routers: the trie
+// router records the template in the request context (it bypasses mux's
+// ServeHTTP, so mux.CurrentRoute is nil), while the default mux path exposes it
+// via mux.CurrentRoute. Callers get consistent behavior without knowing which
+// router is active.
+func RouteTemplate(r *http.Request) string {
```
Umang01-hash, line 33: `RouteTemplate` has 0% test coverage, yet metrics.go + tracer.go were both rewired onto it — so a silent empty-template regression in trie mode wouldn't be caught by any test in this PR. It works (I verified live: the metric label shows the template, not the raw path), but please add one test: assert `RouteTemplate(r) == "/users/{id}"` inside a trie-served handler, plus the `mux.CurrentRoute` fallback for the default router.

Umang01-hash (commented):
Really nice PR — the design (trie as an index, `mux.Route.Match` still making the actual decision) is the right way to do this, so it inherits mux semantics instead of reimplementing them. Verified deeply:

- **Correctness:** 70k-case differential fuzz (trie vs stock mux, random route tables) → 0 routing divergences. The one bare-router 404-vs-405 case is provably neutralized by the `PathPrefix("/")` catch-all, and it's pinned by a test.
- **Performance:** benchstat, hitting the last-registered route — mux 582ns/2.7µs/29µs at 10/100/1000 routes (O(n)); trie flat ~435ns (O(1)). ~67× at 1000 routes. Cost is a constant +3 allocs/req.
- **Default path untouched** (byte-for-byte), no breaking API change, middleware order/parity matches mux internals (`mux.go:143` builds the chain only on `MatchErr==nil`).
- **E2E under `GOFR_ROUTER=trie`:** routing + 404 + metric label `path="/hello"` (bounded template) identical to mux.

One thing I'd like before merge — see inline: the `RouteTemplate` accessor has 0% test coverage.

Nits: no startup log to confirm trie mode is active; `opEmitsSlash` OpLiteral-`/` branch untested.

pkg/gofr/http/router.go
thzgajendra: Resolved — the explicit `Invariant:` doc-comment on the `mws` field captures exactly this. Thanks.

pkg/gofr/http/trie_router.go
thzgajendra: Resolved — the design note now scopes the flat-scaling result to trie-indexable routes and documents that the fallback set is scanned per request but small in practice.

pkg/gofr/http/trie_router.go
thzgajendra: Resolved — the note in `match()` now states the 405 path is framework-moot because of the `PathPrefix("/")` catch-all. Thanks.

thzgajendra (commented):
Re-reviewed after the latest commits — all three of my earlier notes are addressed with clear doc-comments (Use-invariant on the `mws` field, the fallback O(n) scoping, and the framework-moot 404→405 clarification). Thanks for the thorough follow-through.

Worth calling out explicitly: the **mixed literal+parameter segment fix** (`/{name}.txt`, `/user-{id}`, `/v{ver}`) closed a real latent bug — the previous version would have keyed `{name}.txt` as a *literal* trie segment (since `isParamSegment` requires a trailing `}`), so a request like `/files/report.txt` would have silently failed to match in trie mode. Deferring those segments to the fallback list is the right fix, and `TestTrieDifferential_MixedLiteralParamSegments` covers it well — including the `/{lang}-{region}` two-param case that correctly stays trie-indexed as a paramChild. I missed this edge case in my first pass; good catch.

The `RouteTemplate` consolidation into a single router-agnostic resolver (context key → `mux.CurrentRoute` fallback) is a nice cleanup too, and removes the duplicated fallback from metrics.go/tracer.go.

Only remaining item is the one @Umang01-hash raised: a direct test for `RouteTemplate` (assert the template — not the raw path — is recorded in trie mode). It's indirectly exercised via the metrics/tracer + MiddlewareParity differential tests, but a dedicated assertion would close the gap. Non-blocking from my side. LGTM.

pkg/gofr/http/router.go
```diff
@@ -24,6 +26,31 @@ var errReadPermissionDenied = fmt.Errorf("file does not have read permission")
 type Router struct {
 	mux.Router
```
akshat-kumar-singhal, line 42: @aryanmehrotra did we consider replacing this with the new router based on the env to avoid the if condition for every call?

pkg/gofr/http/router.go
```diff
@@ -35,43 +62,148 @@ func NewRouter() *Router {
...
-
 	return r
 }
 
 // ServeHTTP implements [http.Handler] interface with path normalization.
 func (rou *Router) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	// Normalize the path before routing to handle double slashes
+	normalizePath(r)
+
+	if rou.useTrie {
```
akshat-kumar-singhal, line 90: see if we can avoid this `if` condition - it's in the hotpath

akshat-kumar-singhal:
**Clarification: the 404→405 divergence is unreachable in a standard GoFr app**

The description highlights 404→405 as the one intentional behavior change and treats it as a future release-note item. Worth scoping that more tightly: GoFr registers `PathPrefix("/").Handler(catchAllHandler)` unconditionally at startup (`gofr.go:181`) — no method restriction, registered last. For a wrong-method request to an existing path, `match()` returns on the catch-all's **full** match before the `methodMismatch` branch is consulted, and `catchAllHandler → ErrorInvalidRoute → 404` — identical to mux.

So the 405 only surfaces on a bare `gofrHTTP.Router` with no catch-all (unit tests, direct library embedding). Implications:

- In a standard app, flipping the default to trie is byte-identical on this axis too — there's no user-visible 404→405 to release-note.
- The axis to actually weigh for default-on is **matching correctness** (does the trie select the same route as mux across all patterns?), not the status code.

Two small suggestions (no functional change):

1. Reword the description/rollout to scope 404→405 to "bare Router / no catch-all" rather than a general divergence.
2. Add a one-line comment at the 405 site pointing back to the catch-all, so a future reader doesn't mistake it for dead code and delete it (which would regress the standalone `Router` and break the differential test).

pkg/gofr/http/router.go
aryanmehrotra: Measured it rather than guessing — the branch is free, and a function pointer is actually slightly worse:

| dispatch | ns/op |
|---|---:|
| bool branch (current) | **0.98** |
| func-pointer field | 1.06 |
| no branch at all | 1.31 (noise floor) |

All three are within noise of each other at ~1 ns. `useTrie` is set once in `NewRouter` and never written again, so the predictor gets it right every time; swapping to a stored `func` replaces a predicted branch with an indirect call that also blocks inlining, which is why it measures no better.

For scale: that's ~1 ns against a ~3.9 µs request (~0.03%), and it's once per request, not per route.

The alternative shape — selecting a different `http.Handler` implementation at construction so there's no branch at all — is doable, but it means two exported router types or an interface indirection in `httpServer`, which costs more in API surface and readability than the 1 ns it saves. Happy to do it if you'd still prefer that shape.

pkg/gofr/http/router.go
aryanmehrotra: Same measurement applies here (details in the thread on the `ServeHTTP` branch below): the `useTrie` bool costs ~1 ns and a stored function pointer measures no faster, because it trades a perfectly-predicted branch for an indirect call that can't inline.

On replacing the router wholesale based on the env: that was the other shape I considered. Keeping one `*Router` with mux embedded is what lets `mux.Vars`, route templates, `AddStaticFiles`, `NewRoute()` and the embedded-mux API keep working unchanged, and — more importantly — it's what makes the trie an *index over mux's own routes* rather than a second router implementation with its own semantics. mux still makes every final match decision via `Route.Match`, which is what keeps this non-breaking. Two separate router types would mean two matching implementations to keep in sync, which is the risk I was specifically trying to avoid.

If we do flip the trie to default later, the cleanest end state is to delete the flag and the mux matching path entirely — one path, no branch — rather than two router types side by side.

pkg/gofr/http/router.go
akshat-kumar-singhal: I'd lean towards avoiding the decision being made on each request based on a value which remains constant throughout the server lifetime. While the cost is negligible (~1ns), it still adds to the decision tree.


> The alternative shape — selecting a different http.Handler implementation at construction so there's no branch at all — is doable, but it means two exported router types or an interface indirection in httpServer, which costs more in API surface and readability than the 1 ns it saves. Happy to do it if you'd still prefer that shape.

Agree with this. We could consider switching the middleware itself based on the env.

thzgajendra (commented):
Re-reviewed after the latest two commits (`make the trie index classifier an allowlist`, `cover RouteTemplate under both routers`). Both hold up well.

**Classifier allowlist — right call.** Proving indexability positively (anchored + non-empty, slash-free captures) rather than blacklisting unsafe shapes makes "wrongly indexed → silent 404/mis-route" impossible by construction. Checking anchoring on the parsed syntax tree instead of a `"$"` suffix also closes a real latent case: a `PathPrefix("/p$")` compiles to `^/p\$` (ends in the byte `$`) and would otherwise be mistaken for an exact path. Mixed literal+param segments (`/{name}.txt`, `/user-{id}`, `/{a}.{b}`) are correctly indexed as a single param slot via `segmentHasParam`, with `Route.Match` enforcing the intra-segment pattern — I traced `insert`/`collect` and it agrees with mux.

**RouteTemplate test** closes the observability gap: `TestRouteTemplate_ResolvesUnderBothRouters` asserts the template (not the raw path) resolves under both routers — exactly the regression that would silently degrade the metric label / span name.

**Minor follow-ups (non-blocking):**

1. **Benchmark vs claim.** The description shows a 10/100/1000-route table and "~71x", but the committed `BenchmarkRouterMatch` caps at 200 routes and matches the middle route (`n/2`). The trie-is-O(path) / mux-is-O(n) story is sound, but the specific 1,000-route/71x figure isn't reproducible from the shipped harness — worth reconciling the number or adding the 1,000-route (and a worst-case last-route) case.
2. **Bench coverage.** Consider a param-heavy path (deep `{a}/{b}/{c}` fan-out, exercising literal+param backtracking) and a fallback-heavy case (many PathPrefix routes) to demonstrate the documented linear-fallback caveat.
3. **Subrouter edge (doc).** If a registered route were a mux `Subrouter` with its own `NotFoundHandler`, `match()` returns it as a full match and `serveTrie` wraps it in the middleware chain, whereas mux serves that handler bare. GoFr never uses `Subrouter()`/`NotFoundHandler`, so it's unreachable today — but a one-line note in `match()` would prevent a future surprise.
4. **Test gaps.** No `-race` test drives the lazy `sync.Once` index build from concurrent first requests; host/query-matcher routes and static `PathPrefix`-via-`serveTrie` aren't in the differential set. Low priority.
5. **Config idiom (opinion).** `GOFR_ROUTER` is read via raw `os.Getenv` in `NewRouter()`, while every other tunable goes through `Config.Get`. A typo (`GOFR_ROUTER=trei`) silently falls back to mux with no signal — threading it through `Config` and/or logging when the value is set-but-unrecognized would be more consistent and discoverable.

Correctness, concurrency (under the standard `New() -> register -> Run()` lifecycle), and matcher/auth enforcement all check out. Solid iteration — the classifier refactor is stronger than the previous version. LGTM once the benchmark claim is reconciled.

Umang01-hash (dismissed):
Re-reviewed — my earlier ask (a `RouteTemplate` test) is resolved, and the classifier rework is a genuine improvement.

- The denylist→**allowlist** classifier (`isIndexablePathRegexp`: index only if provably anchored + single-segment captures + slash-free) is strictly safer, and it fixed two prior mux divergences (subrouter NotFound, UseEncodedPath).
- `TestRouteTemplate_ResolvesUnderBothRouters` is a real fail-on-revert guard (verified: neutering `withRouteTemplate` fails the trie subtest) — closes the gap I flagged.

Re-verified locally at head: a 15-request differential (param / mixed-segment / catch-all-fallback / 404 / wrong-method / root) is **identical** mux vs trie, and scaling is still flat — **~465 ns (trie) vs ~30 µs (mux) at 1000 routes** — so the allowlist didn't cost the O(1) win (common REST routes stay indexed). Final match still delegates to `mux.Route.Match`.

Opt-in, default-path untouched, no breaking change. LGTM.

thzgajendra (approved):

NitinKumar004 (commented):
## Review — trie router

Went through this end-to-end and set it up locally in both modes. Really solid work — the "trie shortlists candidates, mux still makes the final `Route.Match` decision" split is the right call, and the differential + index-invariant tests are a strong safety net. I ran a large brute-force differential (2800+ method×path combos against an adversarial route table incl. the catch-all) plus the full framework suite in both modes, and saw **zero behavioral divergence**. `mux.Vars`, route-template labels, and middleware order/count all stay identical. No correctness issues found.

One question and a few suggestions before merge — none are blockers on correctness.

### Question

**Is the plan to make the trie the default router in a later release, or will it stay opt-in via `GOFR_ROUTER=trie` permanently?**
This decides how we treat the two routing paths long-term — whether the mux path is a temporary fallback or something we maintain alongside trie indefinitely. Worth stating the intended direction in the PR description.

### Suggestions

1. **Build the index eagerly instead of lazily.**
Right now the index builds on the first request (guarded by `sync.Once`). Building it at the end of `httpServerSetup()` — after all routes and the catch-all are registered — would remove the first-request latency and the per-request `sync.Once` check, and structurally guarantee a route can never be registered after the index is built (today that's safe only by convention). `sync.Once` can stay as a double-build guard.

2. **Add a canary test for mux's compiled-regexp format.**
The classifier reasons about the output of `mux.Route.GetPathRegexp()` (anchors, capture-group shape). That's an implicit contract with this version of mux. A small test asserting the regex shape for a known template would fail loudly if a future gorilla/mux bump changes the format — instead of silently mis-classifying a route into a 404.

3. **Collapse the two `WithContext` calls into one.**
Trie mode does two context writes per matched request (`SetURLVars` + `withRouteTemplate`), i.e. +2 allocs. Storing vars + template in a single context write would halve that per-request overhead — relevant since the win is smaller once the middleware chain dominates.

4. **Log the active router and warn on an invalid `GOFR_ROUTER` value.**
A typo (`GOFR_ROUTER=tri`) currently falls back to mux silently. A one-line startup log of the active router + a warning on unrecognized values would save debugging time.

5. **Document the trie-mode limitations.**
- `mux.CurrentRoute(r)` returns nil in trie mode (it bypasses mux's `ServeHTTP`); user code should use `RouteTemplate(r)`. `mux.Vars` is unaffected.
- Note the real cost model: `O(path length)` for indexed routes **plus** `O(fallback size)` on every request, since the catch-all and static/prefix routes always sit in the linearly-scanned fallback list.

Nice work overall — the design and test discipline here are excellent.

Umang01-hash (dismissed):
Reviewed deep, everything below is from a run or a read at `456c0f8f` — not a skim. This is excellent work.

The safety argument is airtight: the trie only NARROWS candidates and `mux.Route.Match` still decides, so the only risk is under-producing — and `isIndexablePathRegexp` is a proper allowlist (unprovable shapes fall to the always-scanned fallback). I verified routing ≡ mux three ways: the differential suite, `TestTrieRouter_IndexInvariant` (the 'never drop a matchable route' property directly), and an independent 45-input probe — 33/33 agree with the real `PathPrefix("/")` catch-all. The only divergence anywhere is the 404→405 on a bare router, which `gofr.go:183`'s any-method catch-all makes moot in a real app (and it's the safer direction). `-race` clean, no import cycle from middleware→http, middleware parity holds (all registration via `(*Router).Use`, composeMiddleware order matches mux, no double-run), default path unchanged.

One precision note on the perf framing (perf is the whole thesis, so worth getting exact). Measured on M4:

| routes | mux ns/op | trie ns/op | mux allocs | trie allocs |
|--|--|--|--|--|
| 10 | 482 | 449 | 8 | 10 |
| 50 | 1140 | 500 | 8 | 10 |
| 200 | 3005 | 496 | 8 | 10 |

The scaling win is real and large (flat vs linear, ~6x at 200 routes, crossover ~10). But the trie allocates MORE, not less — +2 allocs / +112B per matched request, from the two context adds on the match path (`SetURLVars` + `withRouteTemplate`, router.go:147/152) where mux carries the template in its own match. So it's a CPU/scaling win, not an alloc win — if the description says 'low-alloc', I'd tighten it to 'small constant overhead, +2 allocs vs mux, traded for O(1) scaling'. Folding vars+template into a single context add would reclaim the delta.

Minor, already documented: the trie builds lazily via sync.Once on the first request, so a route registered after the first request is silently unindexed — correct for GoFr's startup-only model, just a landmine if dynamic registration ever lands.

No blockers. Approving.

aryanmehrotra:
Two commits pushed after the approvals — re-review needed, sorry. Both came out of a final self-review; one of them is a design change, not a tidy-up.

## The 404→405 divergence is gone

@Umang01-hash @thzgajendra @akshat-kumar-singhal @NitinKumar004 — the "one intentional divergence" this PR has been carrying is now zero.

I went looking for whether it could be fixed rather than documented, and the root cause is more interesting than "the trie tracks the mismatch differently". In gorilla/mux v1.8.1's `Route.Match`, a matcher that **succeeds** clears any `ErrMethodMismatch` an earlier route left behind:

```go
} else {
    if match.MatchErr == ErrMethodMismatch {
        match.MatchErr = nil   // ← this
    }
}
```

GoFr registers routes as `Methods(m).Path(p)` — **method matcher first**. So in the pinned test case (`POST /a/{id}` then `GET /a/b`, requesting `GET /a/xyz`), the later `/a/b` clears the mismatch purely because its *method* matches, even though its *path* does not, and mux answers 404. That is the registration-order dependence.

Which means the 405 decision depends on routes whose path does not match the request — precisely the routes the trie exists to skip. No bookkeeping over the narrowed candidate set can reproduce it; exact replication would need the full linear scan back.

So instead of approximating it, `serveTrie` now hands every non-match to mux's own `ServeHTTP`:

```go
rou.Router.ServeHTTP(w, r)
```

This is the cold path — the response is already an error — so the linear scan costs nothing that matters, and the hot path is untouched. `notFound()`, `methodNotAllowed()` and the mismatch bookkeeping in `match()` are all deleted.

**The better reason to do it is not the 405.** It makes the index self-healing: if `isIndexablePathRegexp` ever admitted a shape it should not have and the trie dropped a route that *does* match, mux's full scan now finds and serves it. The single catastrophic failure mode of this design — a live route silently 404ing — becomes a request that is merely slower. The trie is now strictly an accelerator, which is the property the whole PR argues for. `TestTrieRouter_UnmatchedFallsBackToMux` pins that with a deliberately blind index.

It also silently fixed a second, smaller divergence I found in the same branch: mux serves `http.NotFoundHandler()` when `Match` returns true with a nil handler, ignoring a custom `NotFoundHandler`, where the old code honoured it.

`KnownMethodMismatchDivergence` is now `MethodMismatchParity`, asserting the routers agree. Mutation-checked both directions: restoring the bookkeeping reds it at 405, and stubbing the delegation reds the fallback test.

## Docs + startup log

@NitinKumar004 — your #4 and #5. `GOFR_ROUTER` had no documentation anywhere, so the only way to find it was to read the source. Added a [Routing Performance](https://github.com/gofr-dev/gofr/blob/perf/router-trie/docs/advanced-guide/routing-performance/page.md) guide (including that `mux.CurrentRoute` returns nil under the trie and callers should use `RouteTemplate`; `mux.Vars` is unaffected) and a `GOFR_ROUTER` row in the config reference. The server now states the matcher when the trie is active and warns on an unrecognized value — `GOFR_ROUTER=tri` used to fall back to mux silently, which is indistinguishable from never setting it.

## On folding the two context writes into one

@Umang01-hash @NitinKumar004 — you both landed on this to reclaim the +2 allocs, and I don't think it's satisfiable. The first write is `mux.SetURLVars`, which stores under mux's **private** context key. That key is the only reason `mux.Vars(r)` keeps working for `request.go` and for user handlers. Folding vars and template under one GoFr key means reimplementing `mux.Vars` — which breaks exactly the compatibility this PR rests on.

So the +2 allocs are the price of `mux.Vars` working unchanged, not an oversight. Umang, agreed on the framing — I've kept the description as a CPU/scaling win and it doesn't claim to be an alloc win.

## Still open

@NitinKumar004 your #1 (eager index build) and #2 (canary test for mux's compiled-regexp format) are both good and I've deliberately left them out of this PR to keep the diff reviewable. Happy to do either here if you'd rather they land together.

Verification on the new head: `go test -race ./pkg/gofr/http/...` green in both modes, `golangci-lint` clean, examples green (local failures are missing redis/mysql only).

Umang01-hash (dismissed):
Re-reviewed at 1e982a8b (was 456c0f8f). The delegate-unmatched-to-mux change is the right call — it makes the 404-vs-405 exactly mux's order-dependent outcome and, more importantly, makes the index self-healing: if the regexp classifier ever mis-drops a matching route, mux's full scan still serves it, so a mis-classification is a slower request, never a silent 404. Verified match() still folds in the fallback list (so the PathPrefix("/") catch-all always matches in-app and delegation is unreachable there) and Use dual-registers so the mux fallback applies the same middleware once. Covered by TestTrieRouter_UnmatchedFallsBackToMux + TestTrieDifferential_MethodMismatchParity.

Also good: unrecognized GOFR_ROUTER now logs a warning, and the docs are honest — "+2 allocations per matched request" and the benchmark table capped at 200 routes / 5.7x, matching the committed benchmarkRouterMatch scale (no unreproducible 1000-route figure).

gofmt clean, go build ./pkg/gofr/... ok, full pkg/gofr/http + pkg/gofr suites green. LGTM.

One optional nit: a classification perf-canary (representative shapes stay trie-indexable across a mux bump) would catch silent performance drift — but the delegation already removes the correctness risk, so it's low-value.

aryanmehrotra:
## Re-review guide: exactly what changed since your approval

@Umang01-hash @thzgajendra — your approvals were on `ff6acfdab`. Three commits landed after them, one of which is a real correctness fix. This is a map so you don't have to re-read the PR: **73 non-comment lines of production code changed**, in three places.

```
pkg/gofr/http/router.go       120 ++++----
pkg/gofr/http/trie_router.go   31 ----
pkg/gofr/http_server.go        28 ++++
+ tests, + docs (new page)
```

### 1. `serveTrie` no longer decides 404-vs-405 — it hands non-matches to mux

**Read: the delegation comment in `serveTrie`.** mux's 405 depends on state that routes which do *not* match the request path still mutate, and GoFr registers `Methods(m).Path(p)` so the method matcher runs first. Reproducing it needs the routes the trie exists to skip. So the branch is one line now, and `notFound()` / `methodNotAllowed()` / the mismatch bookkeeping in `match()` are gone.

The point isn't the 405. It's that the index becomes **self-healing**: if the classifier ever wrongly admitted a shape and the trie dropped a matching route, mux's full scan finds and serves it. The one catastrophic failure mode of this design — a live route silently 404ing — becomes a slower request. The documented divergence is now zero.

### 2. Middleware no longer runs when a matched route reports an error — **the one to scrutinise**

**Read: `serveMatched` and its guard.** This was a genuine bug, present before any of my changes. `serveTrie` applied the chain to every successful match; mux applies it only when `MatchErr == nil`. A subrouter carrying its own `NotFoundHandler` reports a *successful* match with `ErrNotFound`, so under the trie such requests were logged, traced, metered and CORS-answered when mux instruments nothing.

It hid because `TestTrieRouter_SubrouterNotFoundHandler` compared status and body only — and that handler answers 200 through both routers. Status and body agree while the chain does not. Both subrouter tests now assert the middleware invocation count.

Note the guard keys on `MatchErr`, not on "the handler came from an error path", and that distinction is load-bearing: a subrouter's `MethodNotAllowedHandler` also matches successfully, but mux clears `ErrMethodMismatch` once any matcher succeeds, so that case *must* get the chain. There is now a test for each, and the guard is bounded from both sides — too narrow reds one, too broad reds the other plus `MiddlewareParity`.

### 3. `GOFR_ROUTER` is documented, and the active matcher is logged

New [Routing Performance](https://github.com/gofr-dev/gofr/blob/perf/router-trie/docs/advanced-guide/routing-performance/page.md) guide, a config-reference row, and `Router.Matcher()` feeding a startup line. An unrecognized value (`GOFR_ROUTER=tri`) fell back to mux silently, indistinguishable from not setting it.

## Verification beyond CI

CI is green, and separately I ran **the same real GoFr binary twice**, once per router, and diffed every response byte for byte. Not unit tests — a 58 MB service with 230+ routes including a 200-route table.

- **70 request cases × (status + headers + body): identical.** Static/param precedence, param-before-static order fidelity, regex params, mixed segments (`{name}.txt`, `user-{id}`, `v{major}.{minor}`), slash-spanning, empty-matching, all 7 methods, `//` `/./` `/../` and trailing-slash normalisation (sent with `--path-as-is`), percent-encoding, query params, body binding, handler errors, well-known probes, favicon, static files, API-key auth (401/401/200), WebSocket upgrade (101).
- **Metrics route-template labels: identical, 54 series.** This is what `RouteTemplate` feeds, so it is the sharpest test of the context-key mechanism. 18 labels are proper templates; no cardinality drift.
- **A second binary on the exported `gofrHTTP.NewRouter`**, reaching the subrouter shapes a GoFr `App` cannot register: 17 cases identical, with `X-Mw-Ran` headers proving the chain is absent for the `ErrNotFound` match and present for the 405 one, under both routers. This is what caught the inaccurate cardinality claim in my own comment, now corrected.
- **12,800 requests from 64 workers starting simultaneously against a cold index, race detector on: zero races in the routing code.** Three races appear in `middleware.setMiddlewareHeaders` — identically in both modes, same functions, same count, differing only in the caller frame. Pre-existing `cors.go:41` slice aliasing (`*routes` copies the header, `append` writes the shared backing array), reported separately, untouched here.
- **Throughput: no measurable end-to-end difference** (~70k rps both ways at 200 routes). Consistent with the description's scope note — the matching win is real in isolation, but middleware dominates a whole request. Not claiming otherwise.

I proved the two runs genuinely differed before trusting any of it: the trie run logs `HTTP route matcher: trie`, the mux run logs nothing.

Sorry for the re-review churn. Item 2 was worth it.

Umang01-hash (approved):
Re-reviewed at 709ee61c. The middleware-on-error fix is a good catch and it's correct — I checked it against gorilla/mux v1.8.1 rather than the comment: mux.go:142-144 wraps middleware inside Match only when MatchErr==nil, so a subrouter NotFoundHandler (matches with ErrNotFound) is served by mux with no chain; the trie was running it, instrumenting a request mux leaves untouched. Keying the skip on MatchErr (not on 'handler came from an error path') is right — route.go:76-77/87-89 clears ErrMethodMismatch when a matcher succeeds, so the MethodNotAllowedHandler case arrives with nil MatchErr and correctly keeps the chain.

The test is the part I'd call out: it asserts the middleware invocation count differentially against mux, not just status/body — which is the only thing that catches this, since a subrouter NotFoundHandler returns 200 through both paths (status/body agree while the chain doesn't). Revert-red by construction.

gofmt + vet clean, go build ./pkg/gofr/... ok, full pkg/gofr/http + pkg/gofr suites green. Blast radius is the opt-in trie path only. LGTM.

PR #2965, opened 2026-02-06
Fix/http shutdown noise

MOHITKOURAV01:
**Description:**

-   **Fix:** Suppressed misleading error logs (`"error while listening to http server, err: http: Server closed"`) when the HTTP server shuts down gracefully.
-   **Issue:** Addresses #2964.
-   **Motivation:** Previously, a normal application shutdown would generate error logs because `http.ListenAndServe` returns `http.ErrServerClosed` upon closure. These logs were false positives, causing confusion during debugging or automated log monitoring. The change ensures that this specific "error" is ignored, keeping logs clean during intentional shutdowns.

**Breaking Changes (if applicable):**

-   None.
-   Does not affect application behavior, only reduces log noise.

**Additional Information:**

-   Modified [pkg/gofr/http_server.go](cci:7://file:///Users/mohitkourav/Desktop/%20O%20S%20/gofr-1/pkg/gofr/http_server.go:0:0-0:0) to explicitly check for `!errors.Is(err, http.ErrServerClosed)` before logging server errors.
-   Verified with existing tests ([TestShutdown_ServerStopsListening](cci:1://file:///Users/mohitkourav/Desktop/%20O%20S%20/gofr-1/pkg/gofr/http_server_test.go:89:0-125:1)) to ensure shutdown logic remains correct.

**Checklist:**

-   [x] I have formatted my code using  `goimport`  and  `golangci-lint`.
-   [x] All new code is covered by unit tests.
-   [x] This PR does not decrease the overall code coverage.
-   [x] I have reviewed the code comments and documentation for clarity.

PR #3761, opened 2026-07-24
feat(metrics): add push-based OTLP metrics export (#3366)

akshat-kumar-singhal:
## Summary

Adds **push-based OTLP metrics export** alongside the existing Prometheus pull `/metrics` endpoint, for serverless / ephemeral deployments (Cloud Run scale-to-zero) that scraping misses between intervals.

Addresses discussion [#3366](https://github.com/orgs/gofr-dev/discussions/3366).

**Default behavior is unchanged** — with no new config set, metrics stay Prometheus pull-only, byte-for-byte as today.

### Orthogonal switches, not single-mode
An OTel `MeterProvider` supports multiple `Reader`s on the same instruments, so the Prometheus pull reader and an OTLP push reader run **together with no double-counting**. Two independent switches:

| Concern | Control | Default |
|---|---|---|
| Prometheus pull `/metrics` | `METRICS_PORT` (`0` disables) | on (2121) |
| OTLP push | `METRICS_EXPORTER=otlp` | off |

- unset → pull-only (today's behavior)
- `METRICS_EXPORTER=otlp` → **adds** push; `/metrics` stays up unless `METRICS_PORT=0`
- push-only serverless → `METRICS_EXPORTER=otlp` + `METRICS_PORT=0`
- both → the common Google Managed Prometheus case (scrape locally, push to managed backend)

### Registry-based exporter selection
A small `exporters.Register(name, Builder)` seam selects the push exporter; `otlp` self-registers. `exporters.Build` now returns the `*MeterProvider` handle (previously discarded) so it can be flushed/shut down. Adding an exporter is a `Register` call, not a growing switch.

### Graceful shutdown (the one genuinely new lifecycle bit)
Push requires flushing on exit or ephemeral instances lose their final window. `Container.ShutdownMetrics(ctx)` (ForceFlush + Shutdown) is hooked into `App.Shutdown` after servers drain, and into the CMD exit path, reusing `SHUTDOWN_GRACE_PERIOD`.

### Temporality for vendor interop
Datadog **requires delta**; Prometheus/GMP/Grafana **require cumulative**. `METRICS_TEMPORALITY` (`cumulative` default | `delta` | `lowmemory`) maps to an OTLP `TemporalitySelector` per the OTel standard preferences.

## New configuration

| Var | Default | Purpose |
|---|---|---|
| `METRICS_EXPORTER` | — | `otlp` \| `gcp` (unset/`prometheus` = pull only) |
| `METRICS_URL` | — | collector/backend endpoint (host:port for gRPC, URL for HTTP) |
| `METRICS_PROTOCOL` | `grpc` | `grpc` \| `http` |
| `METRICS_EXPORT_INTERVAL` | `30` | push interval (seconds) |
| `METRICS_TEMPORALITY` | `cumulative` | `cumulative` \| `delta` \| `lowmemory` |
| `METRICS_HEADERS` / `METRICS_AUTH_KEY` | — | auth / routing headers |
| `METRICS_INSECURE` | `true` | disable TLS (managed backends set `false`) |

Static-header vendors (Datadog, Grafana Cloud, New Relic, generic collectors) work with `METRICS_HEADERS`/`METRICS_AUTH_KEY` + `METRICS_TEMPORALITY` — **no code, env only**. Recipes in `examples/using-otlp-metrics/README.md`.

## Direct Google Managed Prometheus — keyless

`pkg/gofr/metrics/exporters/gcp` ships as its **own module** (keeps Google auth deps out of the root `go.mod`). It self-registers a ready-made `gcp` exporter that authenticates with a **refreshing** OAuth2 ADC token (a static header can't refresh the ~1h Google token) and forces TLS.

```go
import _ "gofr.dev/pkg/gofr/metrics/exporters/gcp"
```
```
METRICS_EXPORTER=gcp   # METRICS_URL defaults to telemetry.googleapis.com:443
```

On Cloud Run this is **keyless** via the attached service account (grant `roles/monitoring.metricWriter`) — no key file. Walkthrough in `examples/using-gcp-metrics/README.md`.

## Examples
- **`examples/using-otlp-metrics`** — collector docker-compose + vendor recipes; CI-tested integration test asserting `/metrics` stays served in dual mode.
- **`examples/using-gcp-metrics`** — Cloud Run keyless deploy walkthrough.

## Backward compatibility

**No breaking changes.** New env vars and exported symbols are additive; `Container` gained only unexported fields; default `/metrics` output (labels/resource) is unchanged (`service.instance.id` is never auto-injected). `exporters.Prometheus` is retained as a deprecated wrapper over `Build`. New root deps (`otlpmetric{grpc,http}` v1.44.0) were already transitively present via `otlptracegrpc`.

## Testing
- Table-driven unit tests: registry, `Build` (pull-only / custom push / unknown-fallback / failing-builder-fallback), OTLP exporter construction (grpc/http/headers/insecure), temporality selector, container config parsing + `ShutdownMetrics`, and the `gcp` builder (faked ADC token source).
- `examples/using-otlp-metrics` integration test.
- `go build ./...`, `go vet`, `gofmt` clean; `golangci-lint` 0 new issues (root) and clean (gcp module). The `gcp` submodule is auto-covered by the existing `Submodule-Unit-Testing` / `lint_changed_submodules` CI jobs.

## Relationship to #3176 (Pushgateway for CLI)
Complementary, not competing: [#3176](https://github.com/gofr-dev/gofr/pull/3176) solves CLI cross-run *accumulation* via Pushgateway read-modify-write (#2232); this solves ephemeral-HTTP *push* via OTLP. Note for reviewers: OTLP **delta** temporality makes RMW unnecessary for anyone with an aggregating backend, and the registry here is a natural home to later fold Pushgateway in as `METRICS_EXPORTER=pushgateway`. The two touch the CMD-flush path and the `exporters/` package, so whichever merges first, the other rebases.


aryanmehrotra (commented):
Nice work — solid, well-architected PR. The multi-reader provider, registry-based selection, isolated GCP submodule, and shutdown flush are all sound, and a few choices are better than the obvious ones. No blockers. Verified at HEAD `50f2f7e` against the vendored OTel SDK v1.44.0. A few actionable items, then a design question.

### Correctness
- **`WithInsecure()` overrides an explicit `https://` URL.** With a scheme present, `WithEndpointURL` derives security from the scheme, but the later `WithInsecure()` (default `METRICS_INSECURE=true`) wins by option order (`otlpmetric{grpc,http}/internal/oconf/options.go`: `WithEndpointURL` sets `Insecure = scheme != "https"`, `WithInsecure` sets it unconditionally). So `METRICS_URL=https://collector:4317` connects in plaintext — the scheme is silently ignored. Fix: when the endpoint has a scheme, don't append `WithInsecure()` (both branches); keep `METRICS_INSECURE` as the override only for schemeless `host:port`.
- **Empty `METRICS_URL` + `METRICS_EXPORTER=otlp` silently pushes to `localhost:4317`.** `New()` doesn't error on an empty endpoint, so the prometheus-only fallback never triggers and a `PeriodicReader` fails to export every interval silently. Return an error for empty endpoint so `pushReader` logs and degrades.

### Tests
- No test for the CMD-flush path (`run.go` → `ShutdownMetrics` after `cmd.Run`) — the one thing most likely to silently regress in a future refactor.
- No `/metrics` scrape test asserting the body has no `target_info`/resource labels (pins the `WithoutTargetInfo()` invariant).

### Minor
- Warn on unrecognized `METRICS_PROTOCOL`/`METRICS_TEMPORALITY`/interval (currently silently defaulted, unlike the good "unsupported METRICS_EXPORTER" path).
- Insecure-by-default shipping `METRICS_AUTH_KEY`/headers in cleartext is consistent with the tracer (`otel.go` hardcodes `WithInsecure()`), so fine to keep — a `Warnf` when `insecure && len(headers) > 0` would give an otherwise-silent deploy some signal.
- Consider honoring `OTEL_METRIC_EXPORT_INTERVAL` / `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` as a fallback.
- CMD flush error is discarded + uses an unbounded `context.Background()`; `ShutdownMetrics` isn't defensively idempotent (2nd call `ForceFlush`es a shut reader).

### Design question — the GCP submodule
It's genuinely well-built (per-RPC refreshing ADC token, TLS forced, secrets-safe, dependency-isolated). But GCP support is really "generic OTLP + Google auth" — the only vendor-specific bit is how you get the token. Two thinner shapes worth your thoughts:

- **(A) Collector-native:** push generic OTLP to a local Collector running `googlemanagedprometheus`/`googlecloud`; zero Google code in-framework, and the same collector serves any vendor. Cost: a sidecar container.
- **(C) Pluggable token-source:** keep the single OTLP exporter and register a credential/token-source provider per vendor (GCP/AWS/Azure ≈ 15 lines each) instead of a full exporter each — DRY-er than a parallel `gcp` exporter.

Where do you land — is the keyless, no-sidecar convenience the deciding factor for a built-in `gcp` exporter over (A)/(C)? Happy to help either way.

pkg/gofr/container/metrics_exporter.go
```diff
@@ -0,0 +1,84 @@
...
+func metricsExporterConfig(conf config.Config, appName, appVersion string) exporters.Config {
+	interval := defaultMetricsExportInterval
+
+	if v := conf.Get("METRICS_EXPORT_INTERVAL"); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n > 0 {
+			interval = n
+		}
+	}
+
+	insecure := true
```
Umang01-hash: Defaulting `insecure` to `true` means OTLP push is plaintext unless the user opts out — the inverse of the OTel SDK default. Safe for a local collector, but a footgun for remote endpoints, and it sends `METRICS_AUTH_KEY` in cleartext. Suggest defaulting to `false`, or warn when insecure is set with a non-loopback `METRICS_URL`.

Umang01-hash (changes requested):
Really solid feature — clean design, no breaking change (Prometheus() preserved + deprecated), and I verified it end-to-end against a live OTEL collector: `/metrics` pull stays intact while OTLP push works, no double-count, graceful shutdown flush wired correctly. GCP submodule isolation + keyless ADC/TLS is well done.

Requesting one change before merge: **`METRICS_INSECURE` defaults to `true`** (plaintext OTLP), which inverts the OTel secure-by-default convention. It's fine for a local sidecar, but if someone points `METRICS_URL` at a remote backend and sets `METRICS_AUTH_KEY`, the credential goes over the wire in cleartext with no warning. Please default it to `false`, or at minimum log a warning when insecure is used with a non-loopback endpoint.

Non-blocking test gaps worth adding while here: interval `n>0` guard (a `"0"`/negative case), `METRICS_HEADERS` empty/malformed-drop + headers-over-auth-key priority, and asserting the gcp exporter actually pins cumulative (only the warning is tested today).

Architecture note (not blocking): the global registry + blank-import self-registration is a new pattern vs GoFr's usual `app.Add<X>` wiring, and it's why the gcp submodule imports core (a first for a plugin submodule). Both are inherent to the `METRICS_EXPORTER=gcp` zero-code UX — just flagging for a conscious call.

Umang01-hash:
Design note (non-blocking, forward-looking): the `gcp` submodule is essentially "generic OTLP + a Google token" — it re-builds the `otlpmetricgrpc` exporter that `otlp.go` already constructs. If more backends are on the roadmap (AWS/Azure), the variation is only the *token*, not the exporter — so consider putting the seam there:

- expose the shared OTLP builder to accept an endpoint + a per-RPC credential (token source)
- `gcp.go` collapses to ~15 lines: resolve ADC → hand the token source to the shared builder

Then each new vendor is another ~15-line token-getter, not a parallel exporter — no duplication drift, no sidecar, still keyless, and core carries a small token lib instead of each vendor's full client stack. Doesn't need to block this PR, but it's cheaper to split at that line now than after 2-3 exporters exist.

akshat-kumar-singhal:
Quick correction on the "first plugin submodule to import core" point — several existing submodules already require `gofr.dev` in their `go.mod` (`dbresolver`, `oracle`, `file/{gcs,azure,s3,sftp,ftp}`, `pubsub/{nats,sqs,eventhub}`), so importing core is the norm for submodules, not a first here.

The genuinely new bit is the blank-import self-registration (`import _` → `init()` → `Register()`) vs the usual `app.AddX` wiring — and that's deliberate: exporter selection happens from env config *before* the container is built, where `AddX` wiring doesn't naturally fit, and it follows the standard `database/sql` driver idiom. Happy to document that rationale in the package doc.

pkg/gofr/container/metrics_exporter.go
akshat-kumar-singhal: Fixed in cbbe385a. `METRICS_INSECURE` now defaults to `false` (secure-by-default, matching the OTel SDK) — a schemeless `host:port` connects over TLS unless `METRICS_INSECURE=true` is set explicitly. Also fixed a related downgrade bug you would have hit next: an explicit `https://` URL was being silently forced to plaintext because the unconditional `WithInsecure()` overrode the scheme; `WithInsecure()` is now only appended for schemeless endpoints. And added a `Warnf` when `METRICS_INSECURE=true` is combined with headers/credentials.

akshat-kumar-singhal:
@aryanmehrotra thanks for the deep review — pushed `cbbe385a` addressing the actionable items.

**Correctness**
- ✅ **`WithInsecure()` overriding `https://`** — sharp catch. `WithInsecure()` is now only appended for schemeless endpoints, so an explicit `http://`/`https://` derives security from the scheme and `METRICS_INSECURE` only applies to schemeless `host:port`.
- ✅ **Empty `METRICS_URL`** — the otlp builder now returns an error on an empty endpoint, so `pushReader` degrades to Prometheus with a visible `Errorf` instead of a `PeriodicReader` failing silently every interval. (Small note: in practice it didn't reach `localhost:4317` — `WithEndpoint` was overwriting the SDK default with `""`, so the target was blank and export hung to deadline. Same fix either way.)

**Tests**
- ✅ Added a CMD-flush test (`run.go` → `ShutdownMetrics` after `cmd.Run`).
- ✅ Added a scrape-body assertion that `target_info` is absent (pins `WithoutTargetInfo()`).

**Minor**
- ✅ Warn on unrecognized `METRICS_PROTOCOL`/`METRICS_TEMPORALITY` and invalid interval.
- ✅ `Warnf` on insecure + headers.
- ✅ CMD flush now uses a bounded `context.WithTimeout` and logs the error instead of discarding it.
- ✅ **Honoring `OTEL_*` env fallbacks** — done in `13f05055` (updated from the earlier "deferring"). `OTEL_METRIC_EXPORT_INTERVAL` and `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` are now honored as a **fallback**, precedence `METRICS_* > OTEL_* > default`. Read via `conf.Get` (GoFr surfaces OS env through its config layer) so GoFr config stays authoritative and the SDK never reads env directly. Note `OTEL_METRIC_EXPORT_INTERVAL` is milliseconds per spec (converted from our seconds-based `METRICS_EXPORT_INTERVAL`); the temporality tokens (`cumulative`/`delta`/`lowmemory`) are shared, so they pass straight through. Covered by `Test_metricsExporterConfig_otelFallback`.
- ⏭️ Idempotent `ShutdownMetrics` — left as-is; the two call sites (cmd path and `App.Shutdown`) are mutually exclusive, so no double-flush occurs in practice.

**GCP submodule design** — replied separately below.

akshat-kumar-singhal:
@Umang01-hash addressed the requested change plus the test gaps in `cbbe385a`.

- ✅ **`METRICS_INSECURE` default → `false`** (the blocker) — see the inline thread. Secure-by-default now; explicit `https://` keeps TLS regardless; `Warnf` on insecure + credentials.
- ✅ **Test gaps** — added interval `n>0` guard cases (`"0"` and negative), `METRICS_HEADERS` empty/malformed-drop, and headers-over-`METRICS_AUTH_KEY` precedence.
- ⏭️ **gcp pins-cumulative assertion** — left for now: `gcp.go` inherits the SDK's cumulative default by simply not setting a temporality selector, and `Reader.temporality` is unexported, so asserting it would need an exporter-level refactor. Can add if you feel strongly.

On the **design note** (shared with @aryanmehrotra's option C — token-source per vendor): I looked at collapsing `gcp.go` to a token-getter over a shared OTLP builder. The snag is Google's ingest needs per-RPC `oauth.TokenSource` auto-refresh, i.e. gRPC `PerRPCCredentials`, not static headers — so a "credential per vendor" API either leaks gRPC/otlp types into core `exporters` or becomes a one-vendor abstraction. The current `Builder` (returns `metricSdk.Reader`) stays maximally general (a future vendor could be HTTP or non-OTLP). My lean: **keep `Register`/`Builder`, but export a small option-passthrough from `exporters` so vendor submodules shrink toward the ~15-line ideal without narrowing the contract.** Open to your take before I invest in it — happy to do it in a follow-up rather than balloon this PR.

(Also left a note above correcting the "first submodule to import core" point — several submodules already import `gofr.dev`; the genuinely new bit is the blank-import self-registration, which is deliberate.)

akshat-kumar-singhal:
Heads-up on the two red **PKG Unit Testing** jobs (v1.24, v1.26) — they are **pre-existing framework flakes under `-race`, not caused by this PR**:

1. `--- FAIL: Test_Shutdown` — a data race in the HTTP server: `httpServer.run()` **writes** `s.srv` (`http_server.go:68`) from the serve goroutine while `httpServer.Shutdown()` **reads** it (`http_server.go:96`) from the caller, with no synchronization. `Test_Shutdown` starts `app.Run()` in a goroutine and immediately calls `app.Shutdown()`, so the two race on `s.srv`.
2. `panic: Fail in goroutine after TestCronTab_AddJob has completed` — a goroutine outliving its test in the cron package.

**Evidence it is not this PR:**
- Both reproduce on plain `origin/development` and on this branch's pre-change base — i.e. without any commit from this PR.
- The affected files (`http_server.go`, `shutdown.go`, cron) are untouched by this PR (last change to `http_server.go` was #2965).
- All packages this PR *does* touch (`container`, `metrics`, `metrics/exporters`) pass cleanly under `-race`.
- `v1.25` passed while `v1.24`/`v1.26` failed — the signature of a timing-dependent race, not a deterministic break. The recently-merged `perf/logging-http-allocations` likely shifted HTTP-path timing enough to surface the latent race more often.

A re-run of the failed jobs should go green. The proper fix (guarding `s.srv`, and the cron goroutine leak) is framework-core and belongs in a separate PR off `development` rather than this feature branch — happy to open that.

akshat-kumar-singhal:
Filed the framework-side fix as #3762 (data race on `s.srv` + the `TestCronTab_AddJob` goroutine leak) — to be fixed on `development` separately from this PR.

aryanmehrotra:
Two follow-ups, plus agreement on the design:

**`ShutdownMetrics` idempotency — worth doing, and it's a 2-for-1.** Checked the SDK: `PeriodicReader.Shutdown` already flushes pending telemetry, so the explicit `ForceFlush` before `Shutdown` is a **redundant second export** on every shutdown. Suggest: drop the `ForceFlush`, keep just `Shutdown(ctx)`, and wrap it in a `sync.Once`. That removes the redundant flush *and* makes repeat calls safe — a 2nd `MeterProvider.Shutdown` otherwise returns `ErrReaderShutdown`, and `App.Shutdown` is public, so a manual call + the signal handler can double-fire (the "mutually exclusive call sites" argument covers cmd-vs-server, but not re-entrancy of the public `Shutdown`).

**gcp cumulative — agree a unit assertion isn't worth an exporter refactor.** But instead of inheriting cumulative *by omission*, set it explicitly: `WithTemporalitySelector(metricSdk.DefaultTemporalitySelector)` in `gcp.go`. One line, turns the "GMP requires cumulative" invariant into a code fact a future edit must consciously change, rather than a silent default.

**Design seam — agree with your lean.** Keep `Register`/`Builder`: returning a `metricSdk.Reader` is the most general seam (a future HTTP or non-OTLP vendor still fits), and the per-RPC `oauth.TokenSource` point is right — a credential-per-vendor API would narrow the contract to gRPC. Do the option-passthrough shrink as a follow-up; no need to design the AWS/Azure abstraction before they exist. One ask: since `Builder` is now a public extension point, worth marking it experimental (or keeping it minimal) so the follow-up can reshape it without a breaking-change debate.

aryanmehrotra:
Everything substantive from the review is addressed — the remaining items (`ForceFlush`/`sync.Once`, explicit cumulative selector) are minor and fine as follow-ups. LGTM from me.

One tidy-up (non-blocking): `using-otlp-metrics` largely duplicates `using-custom-metrics` — the Go code is essentially identical (OTLP is a config toggle, as your own comment notes). Consider folding the `.env` + collector/docker + vendor recipes into `using-custom-metrics` as an "Exporting via OTLP" section and dropping the duplicate app. `using-gcp-metrics` rightly stays its own module for dependency isolation — keep that one.

Heads-up: the `CHANGES_REQUESTED` still showing is @Umang01-hash's earlier `METRICS_INSECURE` review, which is already fixed — a re-review would unblock the merge.

akshat-kumar-singhal:
Thanks — all three follow-ups are in, plus the docs fold. (Pushed as `cdae33a` + `79c5a9f`.)

**1. Shutdown.** Confirmed against the vendored SDK (`periodic_reader.go`): `PeriodicReader.Shutdown` already does `collect` + `Export`, so I dropped the redundant `ForceFlush`. On idempotency I went a slightly different route than `sync.Once`: a `sync.Once` field embeds `noCopy`, and `grpc.go` copies `Container` by value when injecting it into a user's gRPC service struct — that trips `go vet`'s copylocks and would change public behaviour. The SDK's own shutdown closure (`unifyShutdown`) is *already* `sync.Once`-guarded and returns the bare `ErrReaderShutdown` on repeat calls, so `ShutdownMetrics` now treats that sentinel as a no-op via `errors.Is`. Same guarantee, `Container` stays copyable, zero grpc.go churn. Added a second-call idempotency assertion to `TestContainer_ShutdownMetrics` (the prometheus `ManualReader` returns `ErrReaderShutdown` on the 2nd call, so it genuinely guards the regression).

**2. gcp cumulative.** Set explicitly via `WithTemporalitySelector(metricSdk.DefaultTemporalitySelector)` — the GMP invariant is now a code fact, not a silent default.

**3. `Builder` experimental.** Documented as such so the option-passthrough shrink can reshape it later without a breaking-change debate. Agree on keeping the `metricSdk.Reader` return as the most general seam.

**4. Docs.** Folded the OTLP `.env`/collector/recipes into `using-custom-metrics` as an "Exporting via OTLP" section and dropped the duplicate app; `using-gcp-metrics` stays its own module. The moved table also corrects the now-stale `METRICS_INSECURE` default → `false`.

akshat-kumar-singhal:
@Umang01-hash the `METRICS_INSECURE` default is fixed — secure-by-default (`false`) since `cbbe385a`, matching the OTel SDK, plus a `Warnf` when insecure is combined with credentials, and a fix for the related `https://`-downgrade bug. Your `CHANGES_REQUESTED` is the only thing still showing as blocking — a re-review would clear it whenever you get a chance. Thanks!

Umang01-hash (dismissed):
Re-reviewed — the insecure-default I'd flagged is fully resolved, and cleanly: `METRICS_INSECURE` now defaults to `false` (secure), and it's scheme-aware — an `https://` URL derives security from the scheme and can't be silently downgraded, `METRICS_INSECURE` only applies to schemeless `host:port`. Plus a plaintext-credential warning and fail-fast (`otlp` + empty `METRICS_URL` → error + degrade to prometheus-only).

Verified live against an OTEL collector: secure default → plaintext push fails (TLS attempted); `INSECURE=true` → metrics received; `INSECURE=true`+auth → warning; empty URL → error logged + `/metrics` still 200.

Round-1 test gaps also closed: interval `0`/negative fallback, OTEL_* interval/temporality fallback + precedence, insecure default, headers — exporters coverage 52.8%→86%. Bonus: idempotent `ShutdownMetrics`, gcp cumulative pinned explicitly, `Builder` marked experimental. No breaking change, backward-compat pull path intact, gcp submodule isolated.

LGTM. Nice, thorough follow-up. (The multi-cloud token-source refactor from the earlier design note remains a good future direction, not needed here.)

aryanmehrotra:
One more design thought on the shutdown/provider handling — a shape that composes two small changes. I prototyped both (compiles, tests pass, small diff), so this is concrete, not hand-wavy:

**A — `Build` returns a shutdown *closure*, not the concrete `*MeterProvider`.**
`Build(...) (func(ctx) error, metric.Meter)`, with the `ErrReaderShutdown` swallow living inside the closure. Effect: `container.go` stops importing `go.opentelemetry.io/otel/sdk/metric` and stops knowing the internal `ErrReaderShutdown` sentinel (3 → 0 `metricSdk` references; `container.go` actually gets smaller). The container just holds an opaque `func(ctx) error` — cleaner anti-corruption boundary.

**B — one telemetry-shutdown registry, drained once.**
An `App`-level `telemetryShutdown []func(ctx) error` that both the metrics closure *and* the tracer's `tp.Shutdown` register into, drained in `App.Shutdown` and the CMD path. This also fixes a **pre-existing latent bug unrelated to metrics**: the recording tracer registers a `BatchSpanProcessor` (`otel.go`) that is never shut down, so GoFr currently drops buffered spans on process exit. Today metrics gets graceful flush and traces don't — this closes that gap.

**They compose:** the metrics closure (A) registers into the same slice as `tp.Shutdown` (B) → one drain flushes traces + metrics, and no OTel SDK type leaks into `container` *or* the `App` struct.

Scoping suggestion: the trace-flush half is out of scope here, so I wouldn't hold this PR on it — land metrics as-is (optionally adopt the closure from A, it's self-contained), and I'll file the trace-span-loss as its own issue where the unified registry (B) is the natural fix. Happy to share the prototype branches/patch if useful.

akshat-kumar-singhal:
Thanks — adopted **A** in `c11b7a8`. `Build` now returns a `ShutdownFunc` (`func(ctx) error`) instead of the concrete `*MeterProvider`, so the `*MeterProvider` and the SDK's `ErrReaderShutdown` sentinel stay encapsulated in the closure. `container.go` drops its `go.opentelemetry.io/otel/sdk/metric` import entirely (3 → 0 refs) and `ShutdownMetrics` collapses to a nil-check plus one call — cleaner anti-corruption boundary, as you noted.

Did it now specifically because `Build` is public and *wasn't* marked experimental (unlike `Builder`), and it has no external callers yet — so reshaping the signature is free pre-release but would be a breaking change once it ships. Idempotency/flush semantics are unchanged; the `errors.Is(err, ErrReaderShutdown)` swallow just moved into the closure.

**B** (the unified `telemetryShutdown` registry + the trace-`BatchSpanProcessor`-never-shut-down flush) I'm keeping out of this PR as you suggested — it touches `App.Shutdown` ordering and fixes a pre-existing, metrics-unrelated span-loss bug. Filed it as its own issue against `development`: #3771. A's closure composes straight into that registry. 🙏

aryanmehrotra (approved):
LGTM — approving. Re-verified end-to-end on `c11b7a8`: build/vet/unit tests green; OTLP push actually delivers to a real collector with the pull `/metrics` endpoint still served (dual-mode, no double-count); the gcp example boots keyless; and the new `ShutdownFunc` closure keeps the OTel SDK type (and `ErrReaderShutdown`) out of `container.go` while preserving idempotency — 2nd/3rd shutdown calls are clean no-ops. Docs match the config surface. Really solid work through the review iterations.

One follow-up, **not for this PR**: the trace side has the same flush gap this PR just closed for metrics. The recording tracer registers a `BatchSpanProcessor` (`otel.go`) that is never shut down, so GoFr drops buffered spans on process exit today. It's pre-existing and unrelated to metrics — please take it **separately** if you can. The fix is small (`TracerProvider.Shutdown` on exit), ideally via the unified telemetry-shutdown registry from the earlier note, which would then also host this PR's metrics `ShutdownFunc` — one drain for traces + metrics, no SDK types leaking into `App`/`container`. Happy to file the issue and share the prototype branch.

Umang01-hash (approved):

PR #3765, opened 2026-07-25
feat(ai): add configurable max in-flight concurrency to the LLM client

aryanmehrotra:
## What
An optional `Client.MaxConcurrentRequests` cap on the LLM client. When set, `Chat`, `Embed`, and `Stream` acquire an in-flight slot before calling the provider and block — honoring context cancellation — until one frees. Zero (the default) is unlimited, so existing behaviour is unchanged. `HealthCheck` is never limited.

```go
app.AddLLM(&llm.Client{
    Provider: llm.Ollama, Model: "llama3.2:1b",
    MaxConcurrentRequests: 4,
})
```

## Why
Under a burst of concurrent agent calls, the client fired every request straight at the provider. A provider that serializes internally (a single local model) queues them, and tail latency degrades sharply — I measured **p95 ~28s at 12 concurrent vs ~1.4s at one** driving a multi-agent system. This adds client-side backpressure so fan-out can't pile onto a provider that can't absorb it.

## How
- `Client.MaxConcurrentRequests` + a buffered-channel semaphore built in `Connect()`.
- `acquire()` returns an **idempotent** release. `Chat`/`Embed` release on return; `Stream` hands release to the streamer, which frees the slot on exhaustion **or** `Close` (so a streaming call holds its slot for the whole stream).
- Tests: the cap bounds peak in-flight, unlimited runs concurrently, and a waiter honors context cancellation. `go test ./pkg/gofr/ai/...`, `go vet`, `golangci-lint` all pass.

## Honest scope
This is **backpressure only** — it does not make a slow provider faster. Response caching and adaptive concurrency are natural follow-ups, not in this PR.

## Note
Stacked on #3757 (it also limits `Embed`, which #3757 adds), so it's based on `feat/llm-embeddings` for a clean diff. Once #3757 merges I'll retarget this to `development`.


Umang01-hash (approved):
LGTM — the concurrency handling is careful and correct. Verified the load-bearing bits locally:

- Idempotent per-acquire `sync.Once` release (releasing one slot can't free another); cap enforced — the test asserts **peak == limit** under `-race`.
- Every stream termination path releases exactly once (done / error / finish / Close / early-error-before-handoff), and the `streaming` flag + deferred guard makes the handoff leak-safe.
- ctx-cancel handled before and after acquire; HealthCheck is never limited; zero/negative = unlimited (nil sem, zero cost). All concurrency tests pass under `-race`.

One LOW note: a stream that's neither drained nor `Close()`d leaks one slot — mitigated in practice because the framework's `handleStream` auto-`Close()`s on every exit, so the idiomatic path is safe. Worth adding `defer stream.Close()` to the manual `Stream` example in docs/advanced-guide/llm/page.md so it teaches the leak-safe habit.

(Stacked on #3757 — merges after it.)

PR #3766, opened 2026-07-25
test(middleware): characterize the request-path middleware contract

aryanmehrotra:
## Summary

Locks the observable contract of the request-path middleware **before** any optimisation touches these files. Tests only — no production line is modified.

Line coverage of `cors.go`, `tracer.go`, `logger.go` and `metrics.go` was already ~100%, but that is *execution* coverage. Almost nothing asserted the actual output, so a refactor could rename a JSON field, change a header value, flip an attribute type or alter a metric label and the whole suite would still pass. Every assertion here is literal — no `Contains`.

107 test functions, 627 test nodes, ~900 assertions.

## What's pinned

- **cors** — each response snapshotted as an exact sorted header-line slice over a config × origin × method matrix, plus status, body, inner-handler call count, and the `Allow-Methods` join.
- **tracer** — exact instrumentation scope and span name, the complete attribute set with types pinned (`http.response.status_code` as INT64), span kind/status/events/links, W3C `traceparent` parent/child and baggage, and a guard that a valid non-zero TraceID/SpanID still exists under the `NeverSample` provider used when no exporter is configured.
- **logging** — byte-exact JSON wire format: all fields in declaration order, every `omitempty` disappearance case, and a full escaping battery (control chars, `<>&`, U+2028/9, invalid UTF-8, CJK/emoji) across each string field.
- **metrics** — exact metric name and ordered label vector on both the slow path and the attribute-cached fast path, asserted to parity, plus the static-extension and probe matrices.

## Mutation tested

11 deliberate mutations were applied to the production code and reverted. Nine failed loudly (renaming `http.route` → 17 failures, the metric name → 24, the `uri` tag → 19).

The three that **escaped** were tests too loose to keep, and were tightened until the same mutation fails:
1. CORS expectations referenced a production constant instead of a literal, so changing that constant passed → now 52 failures.
2. `start_time` was only asserted to re-parse; `time.Parse` is lenient, so `.999999999` → `.000000000` slipped through.
3. `response_time`'s **unit** was unpinned — a µs→ns swap passed silently.

## Behaviour found and characterized, not fixed

The suite records current behaviour warts included. It documents, among others: a panicking request being logged as `response:200` at Log level while the client receives 500; requests to `/` recorded with an empty metric path label; `HasPrefix(path,"/static")` also matching `/staticfiles/*`; CORS writing `"OPTIONS"` into the caller's backing array; and `getIPAddress` returning `""` for a whitespace-only `X-Forwarded-For`. Those are addressed in a follow-up.

## Tests

`go test -race ./pkg/gofr/http/middleware/ ./pkg/gofr/logging/` green, stable across `-shuffle=on` runs. `golangci-lint --new-from-rev` reports 0 issues.

### DIFF
--- a/pkg/gofr/http/middleware/cors_test.go
+++ b/pkg/gofr/http/middleware/cors_test.go
@@ -3,10 +3,13 @@ package middleware
 import (
 	"net/http"
 	"net/http/httptest"
+	"sort"
 	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type MockHandlerForCORS struct {
@@ -134,7 +137,7 @@ func setMiddlewareHeadersTestCases() []struct {
 			allowedOrigins:    map[string]bool{"*": true},
 			expectedHeaders: map[string]string{
 				"Access-Control-Allow-Origin":  "*",
-				"Access-Control-Allow-Headers": allowedHeaders,
+				"Access-Control-Allow-Headers": corsCharDefaultAllowHeaders,
 				"Access-Control-Allow-Methods": "GET, OPTIONS",
 			},
 		},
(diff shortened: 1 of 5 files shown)


aryanmehrotra:

PR #3767, opened 2026-07-25
test(http): characterize the request/response contract

aryanmehrotra:
## Summary

Companion to #3766, covering the rest of what a request touches: the responder, request binding, the handler execution paths, and the WebSocket upgrade. Tests only — no production line is modified (`4 files changed, 2114 insertions(+)`, zero deletions).

Same rationale: coverage of these files was already high, but almost nothing asserted the observable result, so a refactor could change a status code, a `Content-Type` or the JSON envelope and every test would still pass. Every assertion is literal — exact status, exact `Content-Type`, exact body bytes including the trailing newline.

~297 cases.

## What's pinned

- **responder** — the success envelope across methods × nil/primitive/struct/pointer/typed-nil/slice/map; `Raw` and `Response` (incl. `Metadata` and field order); the full error matrix for every GoFr error type plus plain errors and custom status responders; 206 partial content; JSON HTML escaping; the `Content-Type` default-vs-preserve rule; the encode-failure fallback; and the special types (`File`, `XML`, `Redirect`, `Template`, `Stream`).
- **request** — `Param` / `Params` / `PathParam` / `HostName` / `Context`, and `Bind` for JSON, form-urlencoded, multipart and binary, with the exact stdlib and sentinel error strings for malformed bodies, wrong content types and type mismatches.
- **handler** — `serveInline` and `serveWithGoroutine` asserted against the **same** expectations so the two paths cannot drift; panic recovery on both, with explicit assertions that the panic value and stack never leak to the client; the 408 timeout; 499 client cancellation; and the WebSocket bypass.
- **web_socket** — non-upgrade passthrough byte-identical to the inner handler, upgrade-failure wire shape, and upgrade-success registration.

## Behaviour found and characterized, not fixed

Recording current behaviour surfaced several defects, pinned as-is here and addressed separately:

- `Respond(Raw{...}, err)` returns 206 and **drops the error entirely** — no error field at all.
- A zero-valued struct plus an error is forced to 500 with the caller's error **replaced** by a generic `"internal server error"`, while a *pointer* to the same zero struct returns 206 with the real message. `isEmptyStruct` dereferences for the `Kind()` check but then compares the original pointer against a zero struct.
- All special response types match **by value only**, so `*Redirect` / `*Template` / `*File` / `*XML` fall through to the JSON envelope — a `*Redirect` returns 200 with a JSON body and no `Location` header.
- `Bind` to a non-pointer is a silent no-op; `Content-Type` matching is case-sensitive and untrimmed, so `Application/json` binds nothing without error; `ParseForm` merges the query string into body binding.

## Tests

`go test -race ./pkg/gofr/http/...` green. `golangci-lint --new-from-rev` reports 0 issues. (`go test ./pkg/gofr/` fails on a metrics port bind that reproduces identically on unmodified `development`.)

### DIFF
--- a/pkg/gofr/handler_test.go
+++ b/pkg/gofr/handler_test.go
@@ -512,3 +512,521 @@ func TestIntegration_ServerTimeout(t *testing.T) {
 	errorObj := errorResponse["error"].(map[string]any)
 	assert.Equal(t, "request timed out", errorObj["message"])
 }
+
+// ---------------------------------------------------------------------------
+// Characterization suite for handler.ServeHTTP.
+//
+// Pins the handler-execution contract with exact status codes and exact body
+// bytes for both execution paths (serveInline and serveWithGoroutine), for
+// timeout, cancellation, panic recovery and the WebSocket bypass. The two paths
+// are asserted against the SAME expectations wherever they should agree, so a
+// refactor cannot let them drift.
+// ---------------------------------------------------------------------------
+
+// charHandler builds a handler with a silent logger.
+func charHandler(fn Handler, timeout time.Duration) handler {
+	return handler{
+		function:       fn,
+		container:      &container.Container{Logger: logging.NewLogger(logging.FATAL)},
+		requestTimeout: timeout,
+	}
+}
+
+// charServe runs h against a fresh recorder for the given method.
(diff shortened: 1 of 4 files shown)


aryanmehrotra:

PR #3768, opened 2026-07-25
fix(middleware): CORS origin override, panic log ordering, metric labels and 5xx spans

aryanmehrotra:
> Stacked on #3766 — the characterization suite lands first, then this changes behaviour against it. Review after that merges; the diff here will reduce to the production changes plus the pinned expectations they update.

## Summary

Fixes seven defects surfaced by the characterization suite in #3766. Each was pinned there by a test recording the broken result; those tests now assert the corrected one and say why. Everything else in the suite passes untouched, which is what scopes this change.

No public API changes.

## Fixes

**CORS could emit an origin it never negotiated.** The custom-header pass excluded the literal key `"Access-Control-Allow-Origin"`, but header names are case-insensitive and `Header.Set` canonicalizes whatever it is given — so a differently-cased spelling (`access-control-allow-origin`) slipped past the guard and overwrote the origin matched against the configured allow-list, dropping `Vary` with it. The comparison is now made on the canonical form. GoFr's own config loader only ever emits canonical names, so this is not reachable through env configuration; it is reachable for a caller that builds the `CorsHeaders` map itself, since `CORS` is exported.

**CORS wrote into its caller's slice.** `Access-Control-Allow-Methods` was built with `append(routes, "OPTIONS")`, and `routes` shares its backing array with the router's `RegisteredRoutes`, so whenever `cap > len` the append stored `"OPTIONS"` over the element just past the caller's length. The value is now joined without appending. Registration happens before serving today, so this was latent — but a late registration would have made it a data race on the router's slice.

**A panicking request was logged as a success.** Recovery and the access log were separate deferred calls, and defers unwind LIFO, so the access log ran *before* `panicRecovery` had written the 500: it saw a status of 0, normalised that to 200, and emitted the line at Log level while the client received a 500. Both now share one deferred call, so recovery runs first and the log reports the status actually sent — routing a 5xx to Error.

**The probe-skip path handed the inner handler the raw `ResponseWriter`** instead of the status-capturing wrapper, so a panic on a probe path was recovered against a writer that had observed nothing. Every path now passes the same wrapper.

**Requests to `/` were recorded with an empty metric path label** — `/` is forced to the raw URL and the trailing-slash trim then reduced it to `""`. The root is now left as `/`.

**The static-asset check over-matched.** `strings.HasPrefix(path, "/static")` also matched sibling routes such as `/staticfiles/{id}` and `/statically/{id}`, forcing them onto the raw request URL and so giving those routes unbounded metric label cardinality. `/static` is now matched as a whole path segment.

**A 5xx never marked its span errored** — no `SetStatus` call existed — so a failed request was indistinguishable from a successful one in any trace UI or error-rate query. Server spans are now marked `Error` for 5xx per the OTel HTTP semconv. 4xx is deliberately left `Unset`: it reports a client mistake, not a failure of this server.

## Deliberately not changed

`NeverSample` is not parent-based, so an inbound `sampled=01` traceparent is not honoured on the default no-exporter deployment. Making it `ParentBased` would start recording spans that have no exporter to go to — cost for data with nowhere to land. The current behaviour is correct for that configuration and is left as-is.

## Behaviour changes to be aware of

- A panicking request now logs `response: 500` at **Error** level instead of `200` at Log. Anything keyed on the old shape needs updating.
- Requests to `/` now carry `path="/"` instead of `path=""`.
- Routes under `/staticfiles`-style prefixes now report their route template rather than the raw URL.
- 5xx spans now carry an Error status.

## Tests

`go test -race ./pkg/gofr/http/... ./pkg/gofr/logging/` green. `golangci-lint --new-from-rev` reports 0 issues. Only three tests changed for the CORS/IP work and four for the rest — each updated with a comment explaining the intended change, including marking the `getIPAddress` fix as a deliberate divergence in the equivalence oracle from #3758 rather than silently relaxing it.

### DIFF
--- a/pkg/gofr/http/middleware/cors.go
+++ b/pkg/gofr/http/middleware/cors.go
@@ -7,6 +7,10 @@ import (
 
 const (
 	allowedHeaders = "Authorization, Content-Type, x-requested-with, origin, true-client-ip, X-Correlation-ID"
+
+	headerAllowOrigin  = "Access-Control-Allow-Origin"
+	headerAllowMethods = "Access-Control-Allow-Methods"
+	headerAllowHeaders = "Access-Control-Allow-Headers"
 )
 
 // CORS is a middleware that adds CORS (Cross-Origin Resource Sharing) headers to the response.
@@ -15,7 +19,7 @@ const (
 // the middleware dynamically matches the request's Origin header and responds
 // with the matched origin, adding a Vary: Origin header for correct caching.
 func CORS(middlewareConfigs map[string]string, routes *[]string) func(inner http.Handler) http.Handler {
-	allowedOrigins := parseOrigins(middlewareConfigs["Access-Control-Allow-Origin"])
+	allowedOrigins := parseOrigins(middlewareConfigs[headerAllowOrigin])
 
 	return func(inner http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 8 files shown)


aryanmehrotra:

PR #3769, opened 2026-07-25
fix(http): stop dropping errors and honour pointers to response types

aryanmehrotra:
> Stacked on #3767 — the characterization suite lands first, then this changes behaviour against it. Review after that merges.

## Summary

Fixes five defects surfaced by the characterization suite in #3767. Each was pinned there by a test recording the broken result; those tests now assert the corrected one. No exported signature changes.

## Fixes

**An error passed alongside a `Raw` response was discarded entirely.** The client received `206 Partial Content` with the payload and *no indication anything had failed*. `Raw` now carries the standard error envelope when an error is present. `Raw` without an error is byte-identical to before.

**A zero-valued struct plus an error replaced the caller's error** with a generic `"internal server error"` and forced 500 — while a **pointer** to the same zero struct returned 206 with the real message. Same data, two different contracts. `isEmptyStruct` dereferenced a pointer for its `Kind()` check but then compared the *original pointer* against a zero struct, so the pointer form never matched. Both forms now behave identically, and neither swallows the caller's error.

**Pointers to the special response types fell through to the JSON envelope.** A `*Redirect` returned `200` with `{"data":{"URL":"/there"}}` and no `Location` header instead of redirecting; a `*Template` was serialised instead of rendered; a `*Response`'s custom `Headers` were ignored. Pointer forms are now handled equivalently to values — via an explicit type switch, not reflection, so the response hot path is unaffected. Typed nil pointers keep their existing collapse behaviour.

**`Bind` to a non-pointer silently did nothing** — the body was unmarshalled into a throwaway copy and `nil` returned, leaving the caller's struct zeroed with no error. It now returns `errNonPointerBind`, consistently for every content type rather than only the form paths.

**`Content-Type` matching was case-sensitive and untrimmed**, so `Application/json` or `application/json ; charset=utf-8` fell through to the unhandled branch and bound nothing without error. Media types are now parsed per RFC 9110.

## Behaviour changes to be aware of

| Case | Before | After |
|---|---|---|
| `Respond(Raw{...}, err)` | `206`, naked payload, error lost | `206`, error envelope |
| `Respond(S{}, err)` | `500`, `"internal server error"` | error's own status + real message |
| `Respond(&S{}, err)` | `206` | identical to the value form |
| `Respond(&Redirect{...}, nil)` | `200` + JSON body | `302` + `Location` |
| `Bind(target)` (missing `&`) | `nil`, struct untouched | returns an error |
| `Application/json` | bound nothing, no error | binds |

The `Bind` change is the most likely source of user-visible churn: it can turn a silently-passing handler into a 500 — which is the point, but worth calling out. The `Content-Type` change is purely additive; no request that bound successfully before behaves differently.

## Deliberately left alone

`File`/`XML`/`Template`/`Redirect` still discard an accompanying error body (same class as the `Raw` bug, but fixing it means choosing an error representation per content type). The unhandled-content-type branch still returns `nil` rather than erroring — a much larger blast radius. `ParseForm` merging the query string into form binding. All remain pinned by their characterization tests.

## Tests

`go test -race ./pkg/gofr/http/...` and `-run 'TestHandler|TestErrorLogEntry' ./pkg/gofr/` green. `golangci-lint --new-from-rev` reports 0 issues.

### DIFF
--- a/pkg/gofr/handler.go
+++ b/pkg/gofr/handler.go
@@ -90,9 +90,15 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		result, err = h.serveWithGoroutine(c, traceID, r)
 	}
 
-	// Handle custom headers if 'result' is a 'Response'.
-	if resp, ok := result.(response.Response); ok {
+	// Handle custom headers if 'result' is a 'Response'. A pointer is handled
+	// equivalently to the value form, otherwise its Headers are silently dropped.
+	switch resp := result.(type) {
+	case response.Response:
 		resp.SetCustomHeaders(w)
+	case *response.Response:
+		if resp != nil {
+			resp.SetCustomHeaders(w)
+		}
 	}
 
 	c.responder.Respond(result, err)
--- a/pkg/gofr/handler_test.go
+++ b/pkg/gofr/handler_test.go
@@ -906,17 +906,19 @@ func TestHandler_Char_ResponseCustomHeadersCanOverrideContentType(t *testing.T) 
 	assert.Equal(t, "{\"data\":\"x\"}\n", w.Body.String())
 }
 
-// TestHandler_Char_ResponsePointerHeadersIgnored pins that returning a POINTER
-// to response.Response neither applies the custom headers nor takes the special
-// Response envelope path — the type assertion in ServeHTTP is by value.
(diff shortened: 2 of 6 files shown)


aryanmehrotra:

PR #3770, opened 2026-07-25
fix(http): characterize the request path and fix eleven defects it exposed

aryanmehrotra:
## Summary

Characterizes the HTTP request path with tests that pin its **current** behaviour, then fixes the defects those tests exposed. The commits are ordered so each characterization commit lands before the fixes it enables — and both characterization commits are green against unmodified production code, so the tests-first property is verifiable, not just asserted.

Coverage of these files was already ~90-100%, but that is *execution* coverage. Almost nothing asserted the observable result, so a refactor could change a status code, a header value, a JSON field or a metric label and the whole suite would still pass. Every new assertion is literal — exact status, exact header set, exact body bytes — with no `Contains`.

~6,900 lines of tests; **4 production files touched (+155/−32)**.

> **Scope was cut twice during review.** Two fixes that read as obvious were reverted because they change behaviour real clients can depend on, and GoFr follows semantic versioning on a v1 line. See *Deliberately not fixed* below — that section is the important one.

## Fixes

**CORS could emit an origin it never negotiated.** The custom-header pass excluded the literal key `"Access-Control-Allow-Origin"`, but header names are case-insensitive and `Header.Set` canonicalizes whatever it is given, so a differently-cased spelling slipped past the guard and overwrote the origin matched against the allow-list, dropping `Vary` with it. Worse, the allow-list itself was read with a raw literal lookup, so the same key was *also* missed by `parseOrigins`, which fell back to its wildcard default:

```
config {"access-control-allow-origin": "https://trusted.com"}
  before:  Origin: https://evil.com  →  Access-Control-Allow-Origin: *
  after:   Origin: https://evil.com  →  (no header)
```

Configuration is now folded onto canonical header names once, at construction, with explicit precedence — the exactly-canonical spelling wins, then the lexicographically smallest key — so two spellings of one header cannot resolve differently on different requests. Not reachable through env configuration (which canonicalizes six allowlisted keys); reachable for a caller that builds the `CorsHeaders` map itself, since `CORS` is exported.

**CORS wrote into its caller's slice — a real data race.** `Allow-Methods` was built with `append(routes, "OPTIONS")`, and `routes` shares its backing array with the router's `RegisteredRoutes`, so when `cap > len` the append stored `"OPTIONS"` over the element past the caller's length. Demonstrated:

```
development:  backing=[GET POST OPTIONS SENTINEL-B]   -race: 3 data races, test FAILS
this branch:  backing=[GET POST SENTINEL-A SENTINEL-B] -race: clean, PASS
```

Header output is identical either way.

**`Content-Type` matching was case-sensitive and untrimmed.** `Application/JSON` and `application/json ; charset=utf-8` bound nothing, without error. Now parsed per RFC 9110 with `mime.ParseMediaType`.

**`application/octet-stream` was ignored.** GoFr matched only its own non-standard `binary/octet-stream`; the RFC spelling — what clients actually send — was a silent no-op. Both are accepted now.

**`Bind` into a non-pointer silently did nothing**, unmarshalling into a throwaway copy and returning `nil`. Reported now, so a `nil` error from `Bind(&x)` means the target was populated.

**`Bind` panicked on a nil body.** `io.ReadAll(nil)` panics; a request built by hand, as in a handler unit test, has no `Body`. Treated as empty.

**`*response.Response` custom headers were dropped.** `handler.go` matched the value form only, so `Headers` set on a pointer never reached the wire.

## Behaviour changes

| Case | Before | After |
|---|---|---|
| `Bind(target)` (missing `&`) | `nil`, struct untouched | returns an error → `500` |
| `Application/JSON`, `application/json ; charset=utf-8` | bound nothing, no error | binds |
| `Bind(&b)` where `b` is `[]byte` + `application/octet-stream` | silent no-op | binds |
| `Bind(&x)` where `x` is **not** `[]byte` + `application/octet-stream` | silent no-op | returns an error → `500` |
| `&Response{Headers: …}` | headers dropped | headers applied |
| CORS case-variant allow-headers key | replaced the defaults | extends them |
| CORS case-variant origin key | echoed to every origin / wildcard fallback | configures the allow-list |

`Bind(target)` is the one to watch: it turns a silently-passing handler into a `500`. That requires a handler bug — `Bind(b)` rather than `Bind(&b)` — which is already discarding the body today. It also brings HTTP in line with the other transports, which already reject a non-pointer.

**Two shapes move `201` → `500` on a handler that is written correctly** — one that calls `Bind(&x)` and returns the error:

```
Content-Type: application/octet-stream   + Bind(&someStruct)    201 → 500
Content-Type: Application/JSON           + malformed/empty body  201 → 500
```

Neither is new policy; both are the **removal of an inconsistency**. `binary/octet-stream` — GoFr's own spelling — already errors on a non-`[]byte` target on `development`, and lowercase `application/json` already errors on a malformed body. This PR only makes the RFC spelling and the case-variant behave like the siblings they were always meant to match. `application/octet-stream` is the likelier of the two to be noticed in the wild, since several HTTP clients default to it when no `Content-Type` is set.

## Deliberately not fixed

Two changes were made, reviewed, and then **reverted**. Both were correct diagnoses and the wrong remedy for a v1 line.

**1. Pointers to the special response types are still JSON-enveloped.** `&response.Redirect{}` returns `200` with the struct in the envelope rather than redirecting; `&response.File{}` and `&response.XML{}` are base64-encoded into it; `&Raw{}` and `&Response{}` are wrapped twice.

Dereferencing them looks like an obvious fix. It is a **wire-format change**, and no standard governs how a Go framework treats `*T` against `T` returned from a handler. The current output is not merely parseable but *usable*: a `fetch()` client can read `.data.URL` from a `*Redirect` and route client-side — and `fetch` follows a `302` by default, so "fixing" it would silently send that client to the destination and possibly fail CORS on the way. A client base64-decoding `.data.Content` works the same way.

Pinned by `TestResponder_Char_SpecialTypePointersAreEnveloped` and `TestResponder_Char_TemplatePointerIsNotRendered`, so the shape is recorded and any future change has to edit those tests deliberately. The docs now say plainly to return these types **by value**.

**2. `Bind` still discards a body whose `Content-Type` it cannot decode.** An earlier revision returned `415`. Measured against `development` across the three handler shapes services actually write:

| Handler | Effect of the 415 |
|---|---|
| ignores the `Bind` error | no change |
| never calls `Bind` | no change |
| **returns the `Bind` error** (the documented pattern) | **`201` → `415`** |

That last row is not an edge case. `fetch(url, {method: "POST", body: str})` with no headers sends `Content-Type: text/plain`, so a very common client shape would start being refused. Those requests bind nothing today — but a service that does not need the body *works*, and it would stop working.

So the body keeps being discarded, and the docs say so, with the `fetch()` default called out, rather than the behaviour changing underneath existing users. `errors.go` is byte-identical to `development`.

Also pinned as-is, unchanged: the panic access-log reporting 200 while the client receives 500; the empty metric path label for `/`; `HasPrefix(path, "/static")` matching `/staticfiles/*`; 5xx spans not marked errored; `Raw`+error dropping the error; `getIPAddress` returning `""` for a whitespace `X-Forwarded-For`; `ParseForm` merging the query string into form binding; and non-parent-based sampling.

## Verification

Tested end to end against `development` through **two real GoFr servers over TCP**, same handlers, differenced automatically:

```
309 live request cases    200 IDENTICAL    109 CHANGED    0 REGRESSIONS
```

Every one of the 109 is a change this PR intends:

| Group | Cases |
|---|---|
| `Bind(x)` — missing `&` | 72 |
| `Application/JSON`, `application/json ; charset=utf-8` | 18 |
| `application/octet-stream` (and the uppercase spelling) | 16 |
| `*response.Response{Headers}` | 3 |

The untouched 200 include every non-`Bind` shape across GET / OPTIONS / DELETE and three origins: plain JSON, error, nil, `Raw`, `File`, `Redirect`, **`*Redirect`**, `/.well-known/alive` and an unregistered route. `*Redirect` being byte-identical is what shows the pointer-dereference revert actually landed.

Backing that up at unit level, on the same harness run against both trees:

```
912 CORS cases (19 configs × 4 route sets × 3 methods × 4 origins)
      → the only differences are the non-canonical-spelling cases this PR fixes
693 responder cases (33 data/error shapes × 7 Content-Type pre-states × 3 methods)
      → 0 differences; responder.go and errors.go are byte-identical to development
```

**Exported API surface is unchanged: 921 symbols across `pkg/gofr`, `http`, `http/middleware` and `http/response`, identical on both trees.** No signature added, removed or altered — the changes are behavioural only.

**Whole-repository suite, compared package by package against `development`:** this branch fails strictly *fewer* packages (51 ok vs 50). The one difference is `pkg/gofr/service`, which fails on `development` (a timing-sensitive circuit-breaker test) and passes here. Every shared failure is environmental and byte-identical on both trees: the `examples/*` integration tests need Redis / MySQL / Kafka, and `pkg/gofr` is killed by whatever holds the metrics port.

**Merge safety for the stacked PRs:** `#3770 → development` merges clean, and after that each of `#3970`, `#3971`, `#3972`, `#3973`, `#3974`, `#3975` merges clean in turn. All seven merged together build, vet clean, and pass `-race` on `http`, `http/middleware`, `http/response`, `logging` and `metrics`.

The characterization suites were **mutation-tested**: 11 deliberate mutations applied to production code and reverted. Nine failed loudly. The three that escaped exposed assertions too loose to keep — one referencing a production constant rather than a literal, one asserting a timestamp only re-parses, one leaving a duration's unit unpinned — and were tightened until the same mutation fails.

`TestHandler_Char_ClientCanceled/goroutine` was a ~50/50 coin flip: with a pre-cancelled context and a handler that returns immediately, **both** cases of `serveWithGoroutine`'s `select` are ready and Go picks uniformly at random. The handler now blocks until the test releases it. **300 consecutive runs green.**

`go build ./...` clean; `-race` green on `http`, `http/middleware`, `http/response` and `logging`; `golangci-lint --new-from-rev` reports 0 issues. `gofmt` flags two files, both generated mocks under `datasource/dbresolver/` that are already unformatted on `development` and untouched here. Rebased onto current `development` — the diff is byte-identical across the rebase.

> **Note on `pkg/gofr/logging/remotelogger`:** it fails for reasons unrelated to this PR, and it fails the same way on pristine `origin/development`. Under `-race` it fails every run on both trees — `development` fails five tests, this branch fails three of the same five. Without `-race` both pass 6 of 6. Pre-existing, and worth its own issue.

### DIFF
--- a/pkg/gofr/http/middleware/cors.go
+++ b/pkg/gofr/http/middleware/cors.go
@@ -19,11 +19,12 @@ const (
 // the middleware dynamically matches the request's Origin header and responds
 // with the matched origin, adding a Vary: Origin header for correct caching.
 func CORS(middlewareConfigs map[string]string, routes *[]string) func(inner http.Handler) http.Handler {
-	allowedOrigins := parseOrigins(middlewareConfigs[headerAccessControlAllowOrigin])
+	configs := canonicalizeConfig(middlewareConfigs)
+	allowedOrigins := parseOrigins(configs[headerAccessControlAllowOrigin])
 
 	return func(inner http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			setMiddlewareHeaders(middlewareConfigs, *routes, w, r.Header.Get("Origin"), allowedOrigins)
+			setMiddlewareHeaders(configs, *routes, w, r.Header.Get("Origin"), allowedOrigins)
 
 			if r.Method == http.MethodOptions {
 				w.WriteHeader(http.StatusOK)
@@ -46,16 +47,14 @@ func setMiddlewareHeaders(middlewareConfigs map[string]string, routes []string,
 		w.Header().Add("Vary", "Origin")
 	}
 
-	// Walk the configuration once, classifying each entry by its CANONICAL
-	// header name. Header names are case-insensitive and Header.Set canonicalizes
-	// whatever it is given, so matching on the raw key would let a
-	// differently-cased spelling reach a header it was never checked against —
-	// replacing the origin negotiated above, or overwriting the default
-	// allow-list instead of extending it.
+	// The keys arrive canonicalized from canonicalizeConfig, so a differently-cased
+	// spelling cannot reach a header it was never checked against — replacing the
+	// origin negotiated above, or overwriting the default allow-list instead of
+	// extending it.
 	var customMethods, customHeaders string
 
 	for key, value := range middlewareConfigs {
-		switch http.CanonicalHeaderKey(key) {
+		switch key {
 		case headerAccessControlAllowOrigin:
 			// Always negotiated against the configured allow-list above; never
 			// overridable from here.
@@ -97,6 +96,48 @@ func joinAllowedMethods(routes []string) string {
 	}
 
 	return strings.Join(routes, ", ") + ", " + http.MethodOptions
+}
+
+// canonicalizeConfig folds the caller's configuration onto canonical header keys, once, at setup.
+//
+// Two things depend on this. The allow-list read by parseOrigins used a raw literal lookup while the
+// per-request walk classified by canonical name, so the two disagreed: a caller who spelled the key
+// "access-control-allow-origin" had it dropped by the classifier AND missed by parseOrigins, which
+// then fell through to its wildcard default — turning a config that restricted the origin into one
+// that echoed "*" to an unlisted origin. CORS is exported, so callers do build this map themselves.
+//
+// The other is determinism. Header names are case-insensitive, so two spellings are one header, and
+// a map has no order — resolving the collision during the per-request walk meant the winner was
+// whichever key map iteration reached last, and a different value could be sent on different
+// requests within a single process. Precedence is explicit here instead: an exactly-canonical
+// spelling always wins, and among the rest the lexicographically smallest key does.
+func canonicalizeConfig(cfg map[string]string) map[string]string {
+	canonical := make(map[string]string, len(cfg))
+
+	// Track which raw key won each canonical slot, so precedence does not depend on iteration order.
+	winner := make(map[string]string, len(cfg))
+
+	for key, value := range cfg {
+		ck := http.CanonicalHeaderKey(key)
+
+		if prev, ok := winner[ck]; ok && !better(key, prev, ck) {
+			continue
+		}
+
+		winner[ck] = key
+		canonical[ck] = value
+	}
+
+	return canonical
+}
+
+// better reports whether raw key a should beat b for the canonical slot ck.
+func better(a, b, ck string) bool {
+	if (a == ck) != (b == ck) {
+		return a == ck
+	}
+
+	return a < b
 }
 
 // parseOrigins splits a comma-separated origin string into a set.
--- a/pkg/gofr/http/middleware/cors.go
+++ b/pkg/gofr/http/middleware/cors.go
@@ -8,9 +8,9 @@ import (
 const (
 	allowedHeaders = "Authorization, Content-Type, x-requested-with, origin, true-client-ip, X-Correlation-ID"
 
+	headerAccessControlAllowOrigin  = "Access-Control-Allow-Origin"
 	headerAccessControlAllowMethods = "Access-Control-Allow-Methods"
 	headerAccessControlAllowHeaders = "Access-Control-Allow-Headers"
-	headerAccessControlAllowOrigin  = "Access-Control-Allow-Origin"
 )
 
 // CORS is a middleware that adds CORS (Cross-Origin Resource Sharing) headers to the response.
@@ -38,8 +38,6 @@ func CORS(middlewareConfigs map[string]string, routes *[]string) func(inner http
 func setMiddlewareHeaders(middlewareConfigs map[string]string, routes []string,
 	w http.ResponseWriter, origin string, allowedOrigins map[string]bool,
 ) {
-	routes = append(routes, "OPTIONS")
-
 	// Handle Access-Control-Allow-Origin separately for dynamic matching.
 	if allowedOrigins["*"] {
 		w.Header().Set(headerAccessControlAllowOrigin, "*")
@@ -48,30 +46,57 @@ func setMiddlewareHeaders(middlewareConfigs map[string]string, routes []string,
 		w.Header().Add("Vary", "Origin")
 	}
 
-	// Set default headers (excluding origin, handled above)
-	defaultHeaders := map[string]string{
-		headerAccessControlAllowMethods: strings.Join(routes, ", "),
-		headerAccessControlAllowHeaders: allowedHeaders,
-	}
+	// Walk the configuration once, classifying each entry by its CANONICAL
+	// header name. Header names are case-insensitive and Header.Set canonicalizes
+	// whatever it is given, so matching on the raw key would let a
+	// differently-cased spelling reach a header it was never checked against —
+	// replacing the origin negotiated above, or overwriting the default
+	// allow-list instead of extending it.
+	var customMethods, customHeaders string
 
-	for header, defaultValue := range defaultHeaders {
-		if customValue, ok := middlewareConfigs[header]; ok && customValue != "" {
-			if header == headerAccessControlAllowHeaders {
-				w.Header().Set(header, defaultValue+", "+customValue)
-			} else {
-				w.Header().Set(header, customValue)
-			}
-		} else {
-			w.Header().Set(header, defaultValue)
+	for key, value := range middlewareConfigs {
+		switch http.CanonicalHeaderKey(key) {
+		case headerAccessControlAllowOrigin:
+			// Always negotiated against the configured allow-list above; never
+			// overridable from here.
+		case headerAccessControlAllowMethods:
+			customMethods = value
+		case headerAccessControlAllowHeaders:
+			customHeaders = value
+		default:
+			w.Header().Set(key, value)
 		}
 	}
 
-	// Handle additional custom headers (not part of defaultHeaders or origin)
-	for header, customValue := range middlewareConfigs {
-		if _, ok := defaultHeaders[header]; !ok && header != headerAccessControlAllowOrigin {
-			w.Header().Set(header, customValue)
-		}
+	// A configured method list replaces the derived one; configured headers
+	// EXTEND the defaults rather than replacing them, so the framework's own
+	// required headers (Authorization, Content-Type, X-Correlation-ID, ...)
+	// cannot be dropped by adding one of your own.
+	if customMethods == "" {
+		customMethods = joinAllowedMethods(routes)
 	}
+
+	allowHeaderValue := allowedHeaders
+	if customHeaders != "" {
+		allowHeaderValue = allowedHeaders + ", " + customHeaders
+	}
+
+	w.Header().Set(headerAccessControlAllowMethods, customMethods)
+	w.Header().Set(headerAccessControlAllowHeaders, allowHeaderValue)
+}
+
+// joinAllowedMethods renders the Access-Control-Allow-Methods value: the
+// registered routes plus OPTIONS.
+//
+// It deliberately does not append to routes. That slice shares its backing
+// array with the caller's (the router's RegisteredRoutes), so appending in
+// place writes "OPTIONS" over the caller's next element whenever cap > len.
+func joinAllowedMethods(routes []string) string {
+	if len(routes) == 0 {
+		return http.MethodOptions
+	}
+
+	return strings.Join(routes, ", ") + ", " + http.MethodOptions
 }
 
 // parseOrigins splits a comma-separated origin string into a set.
--- a/pkg/gofr/http/request.go
+++ b/pkg/gofr/http/request.go
@@ -27,21 +27,7 @@ const (
 var (
 	errNoFileFound    = errors.New("no files were bounded")
 	errNonPointerBind = errors.New("bind error, cannot bind to a non pointer type")
-	// errUnsupportedContentType is returned when a request carries a body whose
-	// Content-Type Bind has no decoder for. Reported rather than ignored so the
-	// body is never silently discarded.
-	//
-	// It is an ErrUnsupportedContentType rather than a bare errors.New so it
-	// carries a status: the motivating case is a client that posts a body and
-	// omits Content-Type, which is a client fault. Falling through getStatusCode
-	// to 500 would drive error-rate alerting and SLO burn on the server for a
-	// mistake the server did not make. Every other client-fault error in this
-	// package models its status the same way.
-	//
-	// Kept as a sentinel for errors.Is; Bind returns a value carrying the
-	// offending media type.
-	errUnsupportedContentType error = ErrUnsupportedContentType{}
-	errNonSliceBind                 = errors.New("bind error: input is not a pointer to a byte slice")
+	errNonSliceBind   = errors.New("bind error: input is not a pointer to a byte slice")
 )
 
 // Request is an abstraction over the underlying http.Request. This abstraction is useful because it allows us
@@ -106,56 +92,17 @@ func (r *Request) Bind(i any) error {
 		return r.bindBinary(i)
 	}
 
-	// An unrecognized media type means the body cannot be decoded. Returning nil
-	// would leave the caller's target zeroed with no error — the same silent
-	// no-op the non-pointer check above rejects, and the likelier one in
-	// practice (a client that posts a body but omits Content-Type).
+	// An unrecognized media type is a no-op: the target is left as it was and no
+	// error is reported.
 	//
-	// Only a request that actually carries a body is rejected: with no body
-	// there is nothing to decode and nothing lost, so binding stays a no-op for
-	// callers that Bind defensively on bodyless requests.
-	hasBody, err := r.hasBody()
-	if err != nil {
-		return err
-	}
-
-	if hasBody {
-		return ErrUnsupportedContentType{ContentType: contentType}
-	}
-
+	// This silently discards a body the caller probably meant to bind, and an
+	// earlier revision of this PR rejected it with 415 for exactly that reason.
+	// That is a breaking change and it is not a quiet one: fetch(url, {method:
+	// "POST", body: str}) with no headers sends text/plain, so a handler that
+	// returns the Bind error would start answering 415 to a very common client
+	// shape. Those requests bind nothing today, but services that do not need
+	// the body work, and they would stop working. Left as it is deliberately.
 	return nil
-}
-
-// hasBody reports whether the request carries at least one byte, without
-// consuming it.
-//
-// The question on the reject path is only "is there anything here?", and
-// answering it with r.body() meant an unbounded io.ReadAll first. Ordinary
-// routes have no body cap — the only http.MaxBytesReader calls in the tree are
-// in graphql.go and ai/mcp/server.go — so a client could send an arbitrarily
-// large body with an undecodable Content-Type to any handler that calls Bind
-// and make the server buffer all of it just to refuse it. One byte answers the
-// question; pushing it back in front of the remaining stream leaves the body
-// intact for anything that reads it later.
-func (r *Request) hasBody() (bool, error) {
-	if r.req.Body == nil {
-		return false, nil
-	}
-
-	first := make([]byte, 1)
-
-	n, err := io.ReadFull(r.req.Body, first)
-	if err != nil && !errors.Is(err, io.EOF) && !errors.Is(err, io.ErrUnexpectedEOF) {
-		return false, err
-	}
-
-	if n == 0 {
-		return false, nil
-	}
-
-	r.req.Body = io.NopCloser(io.MultiReader(bytes.NewReader(first[:n]), r.req.Body))
-
-	return true, nil
 }
 
 // mediaType extracts the bare media type from a Content-Type header value.
--- a/pkg/gofr/http/request.go
+++ b/pkg/gofr/http/request.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
 	"reflect"
 	"strings"
@@ -23,7 +24,11 @@ const (
 var (
 	errNoFileFound    = errors.New("no files were bounded")
 	errNonPointerBind = errors.New("bind error, cannot bind to a non pointer type")
-	errNonSliceBind   = errors.New("bind error: input is not a pointer to a byte slice")
+	// errUnsupportedContentType is returned when a request carries a body whose
+	// Content-Type Bind has no decoder for. Reported rather than ignored so the
+	// body is never silently discarded.
+	errUnsupportedContentType = errors.New("bind error, unsupported content type")
+	errNonSliceBind           = errors.New("bind error: input is not a pointer to a byte slice")
 )
 
 // Request is an abstraction over the underlying http.Request. This abstraction is useful because it allows us
@@ -58,8 +63,14 @@ func (r *Request) PathParam(key string) string {
 
 // Bind parses the request body and binds it to the provided interface.
 func (r *Request) Bind(i any) error {
-	v := r.req.Header.Get("Content-Type")
-	contentType := strings.Split(v, ";")[0]
+	// Binding into a non-pointer would unmarshal into a throwaway copy and leave
+	// the caller's value untouched, so reject it up front instead of silently
+	// doing nothing.
+	if rv := reflect.ValueOf(i); rv.Kind() != reflect.Pointer {
+		return errNonPointerBind
+	}
+
+	contentType := mediaType(r.req.Header.Get("Content-Type"))
 
 	switch contentType {
 	case contentTypeJSON:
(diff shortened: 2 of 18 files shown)


akshat-kumar-singhal (commented):
Correctness review of the five production files. I checked the branch out, built it, and ran `go test -race -count=2 ./pkg/gofr/http/... ./pkg/gofr/logging/` (green). Every behavioural claim below was verified empirically by running the same input against this branch and against `development` and diffing the observed output — those before/after pairs are quoted in each comment.

Seven findings, one of them High: the allow-origin half of the CORS fix can turn a restrictive config into a wildcard. The rest are the deref change exposing pointer callers to two pre-existing quirks (error swallowing / generic-500 masking), and four smaller items on the new `Bind` rejection path.

**Verified as *not* introduced by this PR:** `go test -race ./pkg/gofr/` reports 28 `WARNING: DATA RACE` on the WebSocket path (`websocket.go:41` vs `handler.go:186`). These reproduce identically on `development` at the same count, so they are pre-existing and out of scope here.

What the PR gets right, for the record: `joinAllowedMethods` is byte-identical to the old `append`+`Join` for both empty and non-empty route slices; `mediaType` behaves correctly on `""`, `Application/JSON`, `application/json ; charset=utf-8`, `application/json;`, `application/json; charset` (ErrInvalidMediaParameter, non-empty parse) and duplicate-parameter input (falls back); the nil-`Body` guard removes the `io.ReadAll(nil)` panic without touching the JSON path; the `*response.Response` switch in `handler.go` is correct including the typed-nil case; `tracer.go` is comment-only.

No formal verdict attached — these are plain review comments.

pkg/gofr/http/middleware/cors.go
```diff
@@ -15,7 +19,7 @@ const (
 // the middleware dynamically matches the request's Origin header and responds
 // with the matched origin, adding a Vary: Origin header for correct caching.
 func CORS(middlewareConfigs map[string]string, routes *[]string) func(inner http.Handler) http.Handler {
-	allowedOrigins := parseOrigins(middlewareConfigs["Access-Control-Allow-Origin"])
+	allowedOrigins := parseOrigins(middlewareConfigs[headerAllowOrigin])
```
akshat-kumar-singhal: **High — the allow-origin fix is only half applied, and the gap opens CORS up.**

`setMiddlewareHeaders` now classifies every config entry by `http.CanonicalHeaderKey`, but this line still resolves the allow-list with a *raw literal* lookup. For a caller that builds the `CorsHeaders` map itself — the exact caller class this PR says it is fixing, since `CORS` is exported — a case-variant origin key is now dropped by the classifier **and** invisible to `parseOrigins`, which then falls through to its wildcard default.

Verified with the same test on both branches:

```go
routes := []string{"GET"}
cfg := map[string]string{"access-control-allow-origin": "https://trusted.com"}
h := CORS(cfg, &routes)(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {}))
r := httptest.NewRequest(http.MethodGet, "/", nil)
r.Header.Set("Origin", "https://evil.com")
```

```
development:  ACAO = "https://trusted.com"
this branch:  ACAO = "*"
```

So a config that restricted the origin before now emits a wildcard to an unlisted origin. The classifier and the allow-list need to agree: resolve the origin entry by canonical key here too (scan the map for `http.CanonicalHeaderKey(k) == headerAllowOrigin`) rather than by the literal string.

pkg/gofr/http/middleware/cors.go
```diff
@@ -34,40 +38,65 @@ func CORS(middlewareConfigs map[string]string, routes *[]string) func(inner http
...
-		"Access-Control-Allow-Headers": allowedHeaders,
+	// Walk the configuration once, classifying each entry by its CANONICAL
+	// header name. Header names are case-insensitive and Header.Set canonicalizes
+	// whatever it is given, so matching on the raw key would let a
+	// differently-cased spelling reach a header it was never checked against —
+	// replacing the origin negotiated above, or overwriting the default
+	// allow-list instead of extending it.
+	var customMethods, customHeaders string
+
+	for key, value := range middlewareConfigs {
```
akshat-kumar-singhal, line 56: **Low — non-deterministic winner when two keys canonicalize to the same header.**

This loop iterates a map, so a config carrying both `"Access-Control-Allow-Headers"` and `"access-control-allow-headers"` (or any two spellings of the same header) resolves last-write-wins in map-iteration order — a different value can be sent on different requests within the same process. The previous raw-key lookup was at least deterministic about which spelling won.

Worth an explicit precedence rule (prefer the canonical spelling; ignore or reject the duplicate) rather than leaving it to map order.

pkg/gofr/http/responder.go
```diff
@@ -49,6 +49,10 @@ type Responder struct {
 // Respond sends a response with the given data and handles potential errors, setting appropriate
 // status codes and formatting responses as JSON or raw data as needed.
 func (r Responder) Respond(data any, err error) {
+	// A pointer to a special response type must behave exactly like its value form,
+	// otherwise it silently falls through to the JSON envelope below.
+	data = derefSpecialResponse(data)
```
akshat-kumar-singhal: **Medium — `&response.Raw{...}` plus an error now silently discards the error.**

After the deref, a `*resTypes.Raw` reaches `case resTypes.Raw: resp = v.Data` in the switch below, and that branch has no error field at all. Previously the pointer fell into the `default` branch and the error was carried in the envelope.

Verified — `NewResponder(w, http.MethodGet).Respond(&resTypes.Raw{Data: map[string]string{"a": "b"}}, errors.New("boom"))`:

```
development:  206 {"error":{"message":"boom"},"data":{"Data":{"a":"b"}}}
this branch:  206 {"a":"b"}
```

A handler returning `&response.Raw{...}, err` now serves a success-shaped body with the error gone entirely. I understand this is the intended "pointer behaves exactly like value" alignment, and the PR lists `Raw`+error dropping the error as a known out-of-scope quirk — but the deref newly *routes pointer callers into* that quirk, which is a silent error-swallowing change and isn't in the behaviour-changes table. At minimum it belongs in that table; ideally the `Raw`+non-nil-error combination should be addressed rather than extended to a new set of callers.

pkg/gofr/http/responder.go
```diff
@@ -116,6 +120,42 @@ func putRespBuf(buf *bytes.Buffer) {
 	}
 }
 
+// derefSpecialResponse unwraps a pointer to one of the special response types
+// into its value form, so *resTypes.Redirect is redirected, *resTypes.Template
+// is rendered and so on, exactly like their value counterparts. Anything else —
+// including a nil pointer — is returned untouched.
+func derefSpecialResponse(data any) any {
```
akshat-kumar-singhal: **Low — `&response.Response{}` plus an error now returns a generic 500, masking the caller's error.**

Same root cause as the `Raw` comment above. Dereferencing makes `isEmptyStruct` match where it previously could not: it compares `data` against the zero value with `reflect.DeepEqual`, and a `*Response` never equalled a zero `Response`. Now the value form does, so `determineResponse` substitutes `errEmptyResponse`.

Verified — `NewResponder(w, http.MethodGet).Respond(&resTypes.Response{}, errors.New("boom"))`:

```
development:  206 {"error":{"message":"boom"},"data":{"data":null}}
this branch:  500 {"error":{"message":"internal server error"}}
```

The caller's real error message is replaced with a generic one. Again consistent with the value form and with the acknowledged out-of-scope quirk, but it is a new outcome for pointer callers and is not listed in the behaviour-changes table.

pkg/gofr/http/request.go
```diff
@@ -75,9 +86,41 @@ func (r *Request) Bind(i any) error {
...
+	// Only a request that actually carries a body is rejected: with no body
+	// there is nothing to decode and nothing lost, so binding stays a no-op for
+	// callers that Bind defensively on bodyless requests.
+	body, err := r.body()
+	if err != nil {
+		return err
+	}
+
+	if len(body) > 0 {
+		return fmt.Errorf("%w: %q", errUnsupportedContentType, contentType)
```
akshat-kumar-singhal: **Medium — a client fault is reported as HTTP 500.**

`errUnsupportedContentType` is a plain `errors.New`, so it carries no `StatusCode()`. It therefore falls through `getStatusCode` to `http.StatusInternalServerError`.

The motivating case for this whole branch is "a client posts a body but omits `Content-Type`" — that is a client mistake, and it now drives 5xx counters, error-rate alerting and SLO burn on the server. Every other client-fault error in this package models the status explicitly (`ErrorInvalidParam`, `ErrorMissingParam`, `ErrorEntityNotFound`, … in `pkg/gofr/http/errors.go` all implement `StatusCodeResponder`).

Suggest giving this error the same treatment with `StatusCode() int { return http.StatusUnsupportedMediaType }` so it surfaces as 415. The failing-loudly goal is fully preserved; only the status class changes.

pkg/gofr/http/request.go
```diff
@@ -75,9 +86,41 @@ func (r *Request) Bind(i any) error {
...
 
+	// An unrecognized media type means the body cannot be decoded. Returning nil
+	// would leave the caller's target zeroed with no error — the same silent
+	// no-op the non-pointer check above rejects, and the likelier one in
+	// practice (a client that posts a body but omits Content-Type).
+	//
+	// Only a request that actually carries a body is rejected: with no body
+	// there is nothing to decode and nothing lost, so binding stays a no-op for
+	// callers that Bind defensively on bodyless requests.
+	body, err := r.body()
```
akshat-kumar-singhal: **Low — the new rejection path buffers the whole body just to test emptiness.**

`r.body()` does an unbounded `io.ReadAll` and then re-buffers the bytes into a `NopCloser`. Ordinary routes have no body cap — the only `http.MaxBytesReader` calls in the tree are `pkg/gofr/graphql.go:398` and `pkg/gofr/ai/mcp/server.go:52`.

So a client can now send an arbitrarily large body with `Content-Type: text/plain` to any handler that calls `Bind` and force the server to buffer all of it in memory before the request is rejected. On `development` that same request returned from `Bind` immediately without reading anything. It is a genuine (if modest) new amplification vector, and the question being answered is only "is there at least one byte?" — a one-byte read against the body, with the byte pushed back into the replacement `NopCloser`, answers it without the allocation.

pkg/gofr/http/request.go
```diff
@@ -75,9 +86,41 @@ func (r *Request) Bind(i any) error {
 		return r.bindBinary(i)
```
akshat-kumar-singhal, line 92: **Low — the standard `application/octet-stream` spelling now hard-fails.**

The binary branch matches only GoFr's non-standard `binary/octet-stream`. A client sending the RFC spelling `application/octet-stream` previously got a silent no-op from `Bind`; on this branch it falls through to `errUnsupportedContentType` and (per the comment above) a 500.

That converts a pre-existing naming gap into a hard breaking error for what is probably the most common real-world spelling, and it is not called out in the behaviour-changes table. Since the switch is being touched anyway, accepting both spellings here would close the gap instead of weaponising it.

PiyushSingh-ZS:
Reviewed against `development`, built and ran the HTTP package tests on the branch (green). Most of what's here are real defects with correct fixes — the pointer forms of the special response types silently falling through to the JSON envelope, `*response.Response` headers being dropped, `io.ReadAll(nil)` panicking on a hand-built request, case-insensitive media-type matching, and the two CORS bugs.

Two of those are worth singling out:

- **`derefSpecialResponse` and the `handler.go` switch are consistent with each other.** Both the header path and the response path handle the pointer form, and returning a nil typed pointer untouched preserves the existing `isNil` behaviour. It would have been easy to fix one and not the other.
- **The CORS canonical-key classification is the important one.** Matching config keys on `http.CanonicalHeaderKey` rather than the raw string is what stops a differently-cased `access-control-allow-origin` from overwriting the origin negotiated against the allow-list. Note that #3971 rewrites this same function and matches on the **raw** key, so it reintroduces that hole — the two conflict in `cors.go`, and whichever lands second decides. Worth coordinating.

### The `Bind` change is breaking, and I think it needs its own PR

```go
if len(body) > 0 {
    return fmt.Errorf("%w: %q", errUnsupportedContentType, contentType)
}
```

A request that carries a body with an unrecognized or absent `Content-Type` now errors where it previously returned nil. The new behaviour is more correct — silently discarding a body and handing back a zeroed struct is a bad default — but the common real-world trigger is a client POSTing JSON without setting `Content-Type`, and those handlers go from "binds nothing" to "returns an error" on upgrade.

I checked that GoFr's own internals survive it: the MCP tool path builds requests with `http.NoBody` (`tools.go:293`), so `len(body) == 0` and `Bind` stays a no-op there. And `body()` restores `r.req.Body` after reading, so the extra read doesn't consume anything downstream. So it's user-facing only — but it *is* user-facing, and it's the sole change here that can break a working app.

Splitting it out would let the other fixes land on their own timeline and give this one the release note / version framing it needs.

Also minor, same area: the new non-pointer guard changes the JSON path's error from `json: Unmarshal(non-pointer ...)` to `errNonPointerBind`. Better error, but a break for anyone matching on the old one.

### Size and conflicts

6,738 additions across 17 files, open since 25 July. I checked it against the six in-flight perf PRs with `git merge-tree` — it conflicts with **five of six**: `tracer.go` (#3970), `cors.go` (#3971), and test files in #3972, #3973, #3974. Every week this stays open, the rebase gets more expensive.

Concretely, I'd suggest: pull the `Bind` content-type change into its own PR, land the remaining eleven fixes (they're small and uncontroversial), and make sure the CORS canonical-key fix reaches `development` one way or the other before #3971 merges.

NitinKumar004 (changes requested):
Reviewed as part of a deep pass across the 8 open perf/fix PRs. The behavioral fixes here are well-scoped and the exact-bytes characterization tests are great, but a few items — plus cross-PR conflicts — need addressing before merge.

### Medium
- **Flaky test `TestHandler_Char_ClientCanceled/goroutine`.** With `requestTimeout=time.Minute` and a pre-cancelled context, `ServeHTTP` builds a context whose `Done()` is already closed, so `serveWithGoroutine`'s `select` has **both** cases ready (`<-ctx.Done()` → 499 and `<-done` → 200). Go picks uniformly at random, so `assert.Equal(t, 499, w.Code)` is a ~50/50 coin flip and will intermittently fail on loaded CI. Please make the subtest deterministic (drive only the 499-guaranteed path, or block the handler until after the select observes `Done()`).

### Low
- **`derefSpecialResponse` changes the pointer path 206 → 500 undocumented.** A `return &response.Response{}, realErr` (or `&response.Raw{}`) previously escaped the empty-struct trap and emitted 206 with the real message; after the deref it hits `isEmptyStruct` and degrades to 500 + generic `errEmptyResponse`. It's defensible (pointer now matches value behavior) but it's a real behavior change not listed in the PR's behavior table — please document or guard it.
- **CORS doc vs origin allow-list mismatch.** The new "matched by canonical HTTP header name — spelling doesn't matter" doc is true for output-header routing but **not** for `parseOrigins`, which still does a raw case-sensitive lookup. `CORS(map[string]string{"access-control-allow-origin": "https://good.com"}, ...)` → `parseOrigins("")` → `{"*":true}`, i.e. falls **open to wildcard**. Either canonicalize the origin key too, or narrow the doc.
- **`Bind` now reads the whole body on the unsupported-Content-Type reject path** (`io.ReadAll` just to test `len(body) > 0`) where it previously returned immediately. No worse than the existing JSON path, but worth noting.

### ⚠️ Cross-PR conflicts (important)
- **Hard collision with #3971 (`cors.go`).** Both PRs rewrite `setMiddlewareHeaders`/`joinAllowedMethods` from the same base incompatibly (duplicate `joinAllowedMethods` = compile error). If #3971 is taken and this PR's change dropped, we **lose the canonical-header-key fix** and re-open the Allow-Origin override/bypass. Resolution must re-apply canonical-key classification on top of #3971's precompute-once structure.
- **Overlaps #3973 (`responder.go`).** #3773 rewrites `Respond`/`putRespBuf`; ensure `data = derefSpecialResponse(data)` stays ahead of #3973's Content-Type block or pointer-special responses get JSON-enveloped.
- **Overlaps #3970 (`tracer.go` `StatusResponseWriter`).** #3970's `IsRecording` gate changes what this PR's characterization test pins.

**Requested changes:** make the goroutine subtest deterministic; document/guard the pointer 206→500 change; align the CORS origin-key canonicalization with the doc (or fix the doc); and coordinate the `cors.go`/`responder.go`/`tracer.go` merges with #3971/#3973/#3970.

Umang01-hash (dismissed):
Reviewed deep + ran it locally end-to-end — APPROVE, good to merge.

Verified the production surface myself at 91861bc (small: +154/-33): mediaType() resolves case/OWS/charset against real mime.ParseMediaType; ErrUnsupportedContentType→415 with a used errors.Is sentinel; RouteTemplate falls back to mux.CurrentRoute so tracer/metric cardinality stays bounded under both routers; canonicalizeConfig folds once (read-only per request) and joinAllowedMethods no longer appends into the router's shared slice; derefSpecialResponse covers all 7 types with typed-nil preserved. No semver break — Bind/CORS signatures unchanged, only additive exported symbols.

Local E2E green: build (no import cycle), gofmt, vet, go test -race on http/middleware/logging/root; determinism test x20; booted examples/http-server — matched routes emit template metric labels (no cardinality blowup), CORS default + OPTIONS preflight stable across repeats (caller-slice fix holds), logs/traces/health behave.

Adversarial pass over Bind / CORS / responder+tracer found no real defect — one already-documented NIT, one refuted.

Two non-blocking notes:
- The Bind→415 change for a body under an undecodable/absent Content-Type is real for existing callers (crud_handlers Create/Update) — already in your behavior-change table, just flagging the blast radius.
- cors.go here diverges from #3971 (canonical-fold + caller-slice fix vs #3971's sync.Once perf precompute). Whichever merges second will conflict — worth reconciling the two CORS PRs.

pkg/gofr/http/middleware/cors.go
aryanmehrotra: Fixed, and thank you — this was the sharpest finding in the batch.

`canonicalizeConfig` now folds the whole config onto canonical header keys once, inside `CORS`, and `parseOrigins` reads the folded map. Your exact reproduction, before and after:

```
config {"access-control-allow-origin": "https://trusted.com"}
  before:  Origin: https://evil.com    →  Access-Control-Allow-Origin: *
  after:   Origin: https://evil.com    →  (no header)
           Origin: https://trusted.com →  https://trusted.com, Vary: Origin
```

`TestCORS_OriginAllowListMatchedByCanonicalKey` pins it across four spellings of the key.

Worth adding what I found while verifying: `development` is worse than "falls back to wildcard". The key never reaches `parseOrigins`, so it lands in the passthrough loop and the configured origin is echoed **unconditionally to every requester**, with no negotiation and no `Vary` — not `*`, but the literal configured value. Same outcome for the attacker, different mechanism.

pkg/gofr/http/middleware/cors.go
aryanmehrotra: Fixed. Precedence is explicit now rather than left to map iteration: an exactly-canonical spelling wins, and among non-canonical spellings the lexicographically smallest key does.

`TestCORS_DuplicateSpellingsResolveDeterministically` drives 50 requests through a config carrying three spellings of `Access-Control-Allow-Headers` and asserts the same value every time; `TestCanonicalizeConfig_Precedence` covers the fold directly, including the tie-break where no key is canonical.

The fold also happens once at construction now, so the custom-header loop no longer reads the caller's map from every serving goroutine.

pkg/gofr/http/responder.go
aryanmehrotra: This is moot now — the pointer dereference has been **reverted**, so `&response.Raw{...}` never reaches the `Raw` branch and the error is still carried in the envelope exactly as on `development`.

`pkg/gofr/http/responder.go` is byte-identical to `development` on this branch.

Your comment is a good part of why. You flagged that the deref *routes pointer callers into* an existing quirk, and pulling that thread led to a broader problem: dereferencing is a wire-format change on a v1 line that follows semantic versioning, and the current output is not merely parseable but usable — a `fetch()` client can read `.data.URL` from a `*Redirect` and route client-side, and `fetch` follows a `302` by default, so "fixing" it would silently send that client to the destination instead.

The enveloping behaviour is now *pinned* rather than changed, by `TestResponder_Char_SpecialTypePointersAreEnveloped`, and the docs say plainly to return these types by value.

pkg/gofr/http/responder.go
aryanmehrotra: Also moot — same reason as the `Raw` thread above. The deref is reverted, so `&response.Response{}` never matches `isEmptyStruct` and still returns `206` with the caller's message, as on `development`.

`responder.go` is byte-identical to `development` on this branch, and `TestResponder_Char_SpecialTypePointersAreEnveloped` pins the shape so any future change has to edit that test deliberately.

pkg/gofr/http/request.go
aryanmehrotra: You were right, and the resolution went further than the fix you suggested.

First: I did implement `StatusCode() int { return http.StatusUnsupportedMediaType }` as you proposed — and it **did not work**. `Bind` wrapped the error with `fmt.Errorf("%w: %q", …)`, and `determineResponse` resolves the status with a direct type assertion `err.(StatusCodeResponder)`, not `errors.As`. `*fmt.wrapError` has no `StatusCode` method, so it still fell through to `500`. My unit test asserted `require.ErrorAs`, which *does* unwrap, so it passed against an error the framework could not read. Only an end-to-end test through a real server caught it.

Then the whole rejection was **reverted**. Measured against `development` across the three handler shapes:

| Handler | Effect of the 415 |
|---|---|
| ignores the `Bind` error | no change |
| never calls `Bind` | no change |
| **returns the `Bind` error** | **`201` → `415`** |

That last row is not an edge case: `fetch(url, {method: "POST", body: str})` with no headers sends `Content-Type: text/plain`, so a very common client shape would start being refused. Those requests bind nothing today — but a service that does not need the body works, and it would stop working.

So `Bind` is a no-op again for a body it cannot decode, `errors.go` is byte-identical to `development`, and the docs now state the no-op plainly with the `fetch()` default called out.

pkg/gofr/http/request.go
aryanmehrotra: Moot — the rejection path is gone, so nothing is buffered to decide whether to refuse. `Bind` returns `nil` for an undecodable `Content-Type` exactly as on `development`, and the one-byte `hasBody` probe went with it.

For the record, the probe you asked for was implemented first, with a `countingReader` test asserting at most one byte came off the wire. It was removed only because the reject path it served no longer exists — see the `errUnsupportedContentType` thread for why that was reverted.

pkg/gofr/http/request.go
aryanmehrotra: Fixed — `Bind` accepts `application/octet-stream` alongside GoFr's non-standard `binary/octet-stream`, verified end to end:

```
Content-Type: application/octet-stream, body "rawbytes"
  development:  {"body":"","len":0}        ← silently ignored
  this branch:  {"body":"rawbytes","len":8}
```

`TestBind_OctetStreamSpellings` covers both spellings.

Note the "hard-fails" half of your comment is moot twice over: the unsupported-content-type rejection has since been reverted, so nothing hard-fails on this branch at all. The RFC spelling went from a silent no-op to actually binding, which was the gap worth closing.

Umang01-hash (approved):
Clean PR. The characterization tests actually pin the fixes — I reverted a few and they go red — and no exported signatures changed.

One thing worth a line in the release notes: `Bind` now errors on a non-pointer target, so anyone who wrote `Bind(x)` instead of `Bind(&x)` (silently passed before) will start getting a 500. It's the right call and matches the other transports, just needs to be visible for people upgrading.

Two nits, neither blocking: the octet-stream doc says RFC 2046 but elsewhere it's RFC 9110, and `bindForm`'s non-pointer check is now dead since `Bind` guards it up front.

PR #3800, opened 2026-07-28
fix: data race on server shutdown + de-flake cron/file-token unit tests

### LINKED ISSUE #3762: Data race in httpServer.run/Shutdown on s.srv (Test_Shutdown flaky under -race)
Discovered while running CI on #3761. These are pre-existing framework issues, unrelated to that PR (they reproduce on `development`).

## Symptom

`PKG Unit Testing` intermittently fails under the `-race` detector:

- `--- FAIL: Test_Shutdown` — `WARNING: DATA RACE`
- (related, same job) `panic: Fail in goroutine after TestCronTab_AddJob has completed` — a goroutine outliving its test in the cron package

Non-deterministic across Go versions in the same run (e.g. v1.25 passed while v1.24/v1.26 failed).

## Root cause — `Test_Shutdown` race

`httpServer.run()` **writes** `s.srv` from the serve goroutine:

```go
// pkg/gofr/http_server.go:68
s.srv = &http.Server{ ... }
```

while `httpServer.Shutdown()` **reads** `s.srv` from the caller goroutine:

```go
// pkg/gofr/http_server.go:96
func (s *httpServer) Shutdown(ctx context.Context) error {
    if s.srv == nil { // <-- race read
        return nil
    }
    return ShutdownWithContext(ctx, func(ctx context.Context) error {
        return s.srv.Shutdown(ctx) // <-- and here
    }, ...)
}
```

There is no synchronization between the goroutine that starts the server (`App.startHTTPServer` → `s.run()`) and the one that calls `App.Shutdown()`. `Test_Shutdown` calls `app.Run()` in a goroutine and then immediately `app.Shutdown()`, so the two race on `s.srv`. The recently merged `perf/logging-http-allocations` (#3758) likely shifted HTTP-path timing enough to surface this latent race more reliably on CI.

## Reproduce

```
go test ./pkg/gofr/ -run "Test_Shutdown$" -short -race -count=1
```

Fails on `origin/development` (and on any branch), independent of #3761.

## Proposed fix

- Guard `s.srv` with a mutex, or assign `s.srv` **before** launching the serve goroutine so the field is set on the caller goroutine and only the blocking `ListenAndServe()` runs concurrently. The same protection then covers the `Shutdown` read path.
- Separately, fix `TestCronTab_AddJob` so its background goroutine cannot call `t.Fail` after the test has returned (bounded wait / stop signal on the cron ticker in the test).

## Scope

Framework-core; should be fixed on `development` in its own PR (not folded into feature PR #3761).

### LINKED ISSUE #3799: Flaky unit tests: TestCronTab_AddJob panics in goroutine after completion; FileTokenAuth refresh-log tests are time-dependent
### Description

Two unit tests are flaky and fail only intermittently — under the full-package run / loaded CI runner, not in isolation. Both surfaced together in a `PKG Unit Testing` job.

#### 1. `TestCronTab_AddJob` — `panic: Fail in goroutine after TestCronTab_AddJob has completed` (`pkg/gofr`)

`NewCron` starts a per-second ticker goroutine (`pkg/gofr/cron.go`) that `TestCronTab_AddJob` never stops. The test registers a `* * * * *` job (`"test-job"`, fires at second 0 of every minute) but only sets the *registration* mock expectations (`NewHistogram`/`NewCounter`), not the *runtime* ones (`IncrementCounter`/`RecordHistogram`).

The leaked goroutine outlives the test. When the package run crosses a wall-clock minute boundary (observed: `Starting cron job: test-job` logged at `…:33:00.000`), the job runs, calls the mock with the unexpected runtime metric calls, and gomock invokes `t.Fatalf` on the already-finished test's controller → panic in a goroutine, which fails the whole package binary.

Abbreviated stack:

```
panic: Fail in goroutine after TestCronTab_AddJob has completed
  testing.(*common).Fatalf
  go.uber.org/mock/gomock.(*Controller).Call
  gofr.dev/pkg/gofr/container.(*MockMetrics).RecordHistogram   mock_metrics.go:135
  gofr.dev/pkg/gofr.(*job).run.func1                           cron.go:128
  gofr.dev/pkg/gofr.(*Crontab).runScheduled                    cron.go:107
```

#### 2. `TestFileTokenAuthConfig_RefreshFailureLogsWarning` (and the sibling `…_ObservableInjectionViaNewHTTPService`) — `"" does not contain "failed to refresh token"` (`pkg/gofr/service`)

The assertion depends on the background refresh goroutine firing at least one *failing* tick (after the token file is removed) within a fixed `time.Sleep`, then checking captured stdout once. `testutil.StdoutOutputForFunc` only returns output after the function returns, so the test cannot poll — it sleeps a fixed interval and hopes the goroutine was scheduled. Under a loaded runner the goroutine can be starved, so zero warnings are written before capture ends and the assertion sees empty output.

### Steps to Reproduce

Both pass in isolation; the cron panic needs the package run to cross a minute boundary and the service test needs runner load, so they are intermittent. Structurally:

- Cron: `NewCron` leaks a ticker goroutine that is never stopped in `TestCronTab_AddJob`.
- Service: time-based assertion on a background goroutine via a fixed sleep + one-shot capture.

### Expected Behaviour

Both tests pass deterministically regardless of wall-clock timing or runner load.

### Proposed Fix

- Cron: `defer c.Stop()` in every test that constructs a `Crontab` via `NewCron`, plus `AnyTimes()` runtime-metric expectations in `TestCronTab_AddJob` as a safety net.
- Service: capture stdout concurrently and wait with `assert.Eventually(... , 2s, 10ms)` until the warning is actually observed, instead of a fixed sleep.

### Environment

- Framework version: `development`
- Go: 1.24 / 1.25 / 1.26 (observed on 1.26)

akshat-kumar-singhal:
## Description

Fixes flaky `PKG Unit Testing` failures on `development`: a **data race** on server shutdown plus two time-dependent tests. All reproduce only under the full-package run / `-race` / loaded CI runner, not in isolation.

## 1. Data race on `srv` in server shutdown — `Fixes #3762`

`httpServer.run()` **wrote** `s.srv` on the serve goroutine while `App.Shutdown()` **read** it on the caller goroutine, with no synchronization — `Test_Shutdown` failed intermittently under `-race`. The identical write-in-`Run` / read-in-`Shutdown` pattern existed in `metricServer` and `mcpServer` (both surfaced once the httpServer read was fixed).

**Fix:** guard `srv` with a `sync.Mutex` in each server. The lock is held only around the assignment (in `Run`) and the read (in `Shutdown`), **never** across the blocking `ListenAndServe` / `srv.Shutdown`; `Run` serves on a local copy published under the lock.

- `pkg/gofr/http_server.go`
- `pkg/gofr/metrics_server.go`
- `pkg/gofr/mcp.go`

**Scope — this closes the `-race`-visible memory hazard only.** It does *not* fix the separate ordering bug where a signal arriving before the server assigns `srv` makes `Shutdown` see `srv == nil`, report success, and let the listener come up afterwards (the nil check carries no happens-before edge, and `-race` can't see it). That, plus the analogous cron `Stop()`-without-job-join, is tracked in **#3801**. Also out of scope and pre-existing on `development`: a race on `grpcServer.server` (different field, created at registration with a `Fatalf` path) and a `pkg/gofr/websocket` race.

### `TestCronTab_AddJob` — `panic: Fail in goroutine after TestCronTab_AddJob has completed` (`pkg/gofr`)

`NewCron` starts a per-second ticker goroutine the test never stopped. Its `* * * * *` job (`"test-job"`) only had *registration* mock expectations. The leaked goroutine outlived the test and, when the run crossed a wall-clock minute boundary (`Starting cron job: test-job` at `…:33:00.000`), fired the job → unexpected `IncrementCounter`/`RecordHistogram` on the mock → gomock called `t.Fatalf` on the finished test → panic.

**Fix:** `defer c.Stop()` in every test that builds a `Crontab` via `NewCron`, plus `AnyTimes()` runtime-metric expectations in `AddJob`. Note `Stop()` only halts future ticks — it cannot join an already-dispatched `go j.run(...)`, so `AnyTimes()` is load-bearing here, not merely defensive; deterministic joining needs the context-aware `Stop()` tracked in #3801.

### `TestFileTokenAuthConfig_RefreshFailureLogsWarning` / `…_ObservableInjectionViaNewHTTPService` (`pkg/gofr/service`)

Asserted a background refresh-failure warning after a fixed `time.Sleep` and a one-shot stdout read (`testutil.StdoutOutputForFunc` only returns after the func exits, so the test could not poll). Under load the refresh goroutine could be starved → empty output → `"" does not contain "failed to refresh token"`.

**Fix:** a small concurrent `stdoutCapture` helper drains stdout in the background so the test can `assert.Eventually(... contains "failed to refresh token", 2s, 10ms)` — bounded-wait instead of a heuristic sleep. The "cached token remains available across failed refreshes" assertion is retained.

## Verification

- `Test_Shutdown` — pass, `-race -count=15` (was flaky/red before)
- `TestCronTab*` — pass, `-count=5`
- `TestFileTokenAuthConfig_RefreshFailureLogsWarning` + `…_ObservableInjectionViaNewHTTPService` — pass, `-race -count=15`
- `go build` + `go vet ./pkg/gofr/ ./pkg/gofr/service/` + `golangci-lint` — clean

### DIFF
--- a/pkg/gofr/cron_test.go
+++ b/pkg/gofr/cron_test.go
@@ -214,7 +214,16 @@ func TestCronTab_AddJob(t *testing.T) {
 	mocks.Metrics.EXPECT().NewCounter("app_cron_job_success", gomock.Any()).AnyTimes()
 	mocks.Metrics.EXPECT().NewCounter("app_cron_job_failures", gomock.Any()).AnyTimes()
 
+	// The "* * * * *" job fires at second 0 of every minute. If the package run
+	// crosses a minute boundary while the ticker goroutine is alive, the job runs
+	// and records runtime metrics, so allow those calls any number of times.
+	mocks.Metrics.EXPECT().IncrementCounter(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
+	mocks.Metrics.EXPECT().RecordHistogram(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
+
 	c := NewCron(mockContainer)
+	// Stop the background ticker so a leaked goroutine cannot fire "test-job"
+	// after the test completes and panic via t.Fatalf on the mock controller.
+	defer c.Stop()
 
 	for _, tc := range testCases {
 		err := c.AddJob(tc.schedule, "test-job", fn)
@@ -253,6 +262,7 @@ func TestCronTab_runScheduled(t *testing.T) {
 	mocks.Metrics.EXPECT().RecordHistogram(gomock.Any(), "app_cron_job_duration", gomock.Any(), "job", "test-job").Times(1)
 
 	c := NewCron(mockContainer)
+	defer c.Stop()
 
 	// Populate the job array for cron table
 	c.jobs = []*job{j}
@@ -700,6 +710,7 @@ func TestCronTab_runScheduled_Panic(t *testing.T) {
 				var c *Crontab
 				if tc.expectMetricsCalls {
 					c = NewCron(cntnr)
+					defer c.Stop()
 					c.jobs = []*job{j}
 				} else {
 					c = &Crontab{
--- a/pkg/gofr/cron_test.go
+++ b/pkg/gofr/cron_test.go
@@ -214,15 +214,19 @@ func TestCronTab_AddJob(t *testing.T) {
 	mocks.Metrics.EXPECT().NewCounter("app_cron_job_success", gomock.Any()).AnyTimes()
 	mocks.Metrics.EXPECT().NewCounter("app_cron_job_failures", gomock.Any()).AnyTimes()
 
-	// The "* * * * *" job fires at second 0 of every minute. If the package run
-	// crosses a minute boundary while the ticker goroutine is alive, the job runs
-	// and records runtime metrics, so allow those calls any number of times.
+	// These AnyTimes() runtime expectations are load-bearing, not merely
+	// defensive: Stop() below halts only *future* ticks, it does not join a job
+	// goroutine that a tick already dispatched (Crontab fires `go j.run(...)` with
+	// no WaitGroup). The "* * * * *" job fires at second 0 of every minute, so a
+	// job dispatched just before Stop can still call the metrics mock after the
+	// test returns; without these expectations that call is gomock's
+	// t.Fatalf-on-finished-test -> panic. `defer c.Stop()` narrows the window but
+	// cannot close it — deterministic joining needs a framework change to Stop
+	// (tracked in gofr-dev/gofr#3801). Do not remove these believing Stop covers it.
 	mocks.Metrics.EXPECT().IncrementCounter(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
 	mocks.Metrics.EXPECT().RecordHistogram(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
 
 	c := NewCron(mockContainer)
-	// Stop the background ticker so a leaked goroutine cannot fire "test-job"
-	// after the test completes and panic via t.Fatalf on the mock controller.
 	defer c.Stop()
 
 	for _, tc := range testCases {
--- a/pkg/gofr/http_server.go
+++ b/pkg/gofr/http_server.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"sync"
 	"time"
 
 	"gofr.dev/pkg/gofr/container"
@@ -18,6 +19,7 @@ type httpServer struct {
 	router      *gofrHTTP.Router
 	port        int
 	ws          *websocket.Manager
+	srvMu       sync.Mutex // guards srv, which run() writes on the serve goroutine and Shutdown() reads on the caller goroutine
 	srv         *http.Server
 	certFile    string
 	keyFile     string
@@ -58,18 +60,27 @@ func (s *httpServer) run(c *container.Container) {
 		middleware.WSHandlerUpgrade(c, s.ws),
 	)
 
+	s.srvMu.Lock()
+
 	if s.srv != nil {
+		s.srvMu.Unlock()
 		c.Logf("Server already running on port: %d", s.port)
+
 		return
 	}
 
 	c.Logf("Starting server on port: %d", s.port)
 
-	s.srv = &http.Server{
+	// Assign under the lock, then operate on the local copy so the blocking
+	// ListenAndServe call below never holds it. Shutdown reads srv under the
+	// same lock, so the two goroutines no longer race on the field.
+	srv := &http.Server{
 		Addr:              fmt.Sprintf(":%d", s.port),
 		Handler:           s.router,
 		ReadHeaderTimeout: 5 * time.Second,
 	}
+	s.srv = srv
+	s.srvMu.Unlock()
 
 	// If both certFile and keyFile are provided, validate and run HTTPS server
 	if s.certFile != "" && s.keyFile != "" {
@@ -79,7 +90,7 @@ func (s *httpServer) run(c *container.Container) {
 		}
 
 		// Start HTTPS server with TLS
-		if err := s.srv.ListenAndServeTLS(s.certFile, s.keyFile); err != nil && !errors.Is(err, http.ErrServerClosed) {
+		if err := srv.ListenAndServeTLS(s.certFile, s.keyFile); err != nil && !errors.Is(err, http.ErrServerClosed) {
 			c.Errorf("error while listening to https server, err: %v", err)
 		}
 
@@ -87,24 +98,24 @@ func (s *httpServer) run(c *container.Container) {
 	}
 
 	// If no certFile/keyFile is provided, run the HTTP server
-	if err := s.srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+	if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
 		c.Errorf("error while listening to http server, err: %v", err)
 	}
 }
 
 func (s *httpServer) Shutdown(ctx context.Context) error {
-	if s.srv == nil {
+	s.srvMu.Lock()
+	srv := s.srv
+	s.srvMu.Unlock()
+
+	if srv == nil {
 		return nil
 	}
 
 	return ShutdownWithContext(ctx, func(ctx context.Context) error {
-		return s.srv.Shutdown(ctx)
+		return srv.Shutdown(ctx)
 	}, func() error {
-		if err := s.srv.Close(); err != nil {
-			return err
-		}
-
-		return nil
+		return srv.Close()
 	})
 }
 
--- a/pkg/gofr/mcp.go
+++ b/pkg/gofr/mcp.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"sync"
 	"time"
 
 	"gofr.dev/pkg/gofr/ai/mcp"
@@ -74,6 +75,7 @@ func (a *App) mcpPort() (int, bool) {
 type mcpServer struct {
 	port    int
 	handler http.Handler
+	srvMu   sync.Mutex // guards srv, written by Run on the serve goroutine and read by Shutdown on the caller goroutine
 	srv     *http.Server
 }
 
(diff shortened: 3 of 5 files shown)


aryanmehrotra (commented):
Reviewed this end-to-end and re-ran everything locally. The fix is correct and every claim in the description holds up:

| Claim | Result |
| --- | --- |
| `Test_Shutdown` race gone | 6 races on `development` → **0** on this branch (`-race -count=15`) |
| `TestCronTab*` stable | pass, `-count=5` |
| `TestFileTokenAuthConfig*` stable | pass, `-race -count=15` |

The lock scope is right — held only around the assign and the read, never across `ListenAndServe` or `srv.Shutdown` — and it's applied consistently across all three servers.

```text
BEFORE                                          AFTER
  serve goroutine      caller goroutine           serve goroutine      caller goroutine
       |                     |                         |                     |
  s.srv = &http.Server  if s.srv == nil           srvMu.Lock()          srvMu.Lock()
       |  ^ WRITE            |  ^ READ            s.srv = srv  --HB-->  srv := s.srv
       +--+---- same word ---+--+                 srvMu.Unlock()        srvMu.Unlock()
          no synchronisation                           |
        DATA RACE  (-race flags it)              srv.ListenAndServe()
                                                  (lock NOT held across the block)
                                                        no data race
```

Two things I'd like your read on before this goes in.

---

## 1. The mutex makes `srv` memory-safe, but the shutdown can still be lost

`run.go` starts the shutdown handler *before* the servers:

```go
a.startShutdownHandler(ctx, timeout)   // goroutine parks on <-ctx.Done()
a.startTelemetryIfEnabled()
a.startAllServers(ctx)                 // servers assign srv in here
```

So a signal arriving in that window is observed by a `Shutdown()` that sees `srv == nil`, reports success, and returns — after which `run()` goes on to serve:

```text
App.Run()
  |-- startShutdownHandler(ctx)  -->  [shutdown goroutine] waits <-ctx.Done()
  +-- startAllServers(ctx)       -->  [serve goroutine]

---------------------------- TIMELINE ---------------------------->
 t0  SIGTERM arrives
 t1  shutdown g:  Shutdown() -> lock -> srv == nil -> unlock
 t2                returns nil          <-- "success, nothing to stop"
 t3  serve g:     s.srv = srv           <-- assigns AFTER shutdown finished
 t4               srv.ListenAndServe()  <== SERVES FOREVER
 t5  wg.Wait() never returns  ->  process hangs until SIGKILL
```

I reproduced this against this branch — shutdown returns `nil`, then the listener comes up and stays up:

```
BUG: shutdown was requested and reported success,
     yet the server came up afterwards and is serving on port 53969
```

In practice this is a SIGTERM landing right after start — crashloop, fast rollout, failed readiness — where the pod ignores graceful shutdown and hangs until SIGKILL.

The root cause is that `srv` is doing two jobs: it's the server handle *and* the "has the server started?" flag. A nil check can't carry a happens-before edge, so the mutex fixes the read/write hazard without fixing the ordering. Worth noting `-race` can't see this one at all — the race detector was the only thing pointing here, and this change quiets it.

A `started chan struct{}` (or a `WaitGroup`) that `Shutdown` waits on would turn "not started yet" into "wait, then stop" instead of "nothing to do".

**This is pre-existing on `development` and I don't think it should block the PR** — this is a strict improvement either way. But the description says it fixes the shutdown race, and this is the half that actually reaches users, so I'd rather we decide deliberately: extend scope here, or land and file a follow-up.

## 2. In the cron test, `AnyTimes()` is doing the work, not `defer c.Stop()`

The comment frames `AnyTimes()` as a safety net behind `defer c.Stop()`. It's the other way round — `Stop()` stops future ticks but never joins a job that was already dispatched:

```go
func (c *Crontab) Stop() {
	c.once.Do(func() {
		c.ticker.Stop()
		close(c.done)
	})
}
```

```text
Stop()  -->  ticker.Stop() ; close(done)      <- only stops FUTURE ticks
                  |
                  |  a tick already dispatched is NOT joined:
                  v
runScheduled() -- go j.run(container) -->  [job goroutine]   (no WaitGroup)
                                                  |
                                    test returns -+
                                                  v
                              mock.IncrementCounter() after test ends
                              -> gomock t.Fatalf on finished test -> panic
```

`Crontab` holds `ticker`, `jobs`, `mu`, `done`, `once` — there's no `WaitGroup`, and `runScheduled` fires `go j.run(...)` without one. So `defer c.Stop()` narrows the window but doesn't close it, which makes the blanket `AnyTimes()` load-bearing rather than defensive. The cost is that `TestCronTab_AddJob` no longer asserts anything about cron metrics.

Either add a `WaitGroup` around the dispatch so `Stop()` genuinely joins, or reword the comment so nobody deletes the `AnyTimes()` later believing `Stop()` already covers it.

---

Both of these are the same shape, which is why I'd like to fix it once rather than twice: **a "stop" that establishes no happens-before edge with the thing it's stopping** — nil-check-as-started-signal in the servers, `Stop()`-without-join in cron.

## Things I liked

- The `t.Cleanup` ordering in `stdoutCapture` is subtly correct — cleanups run LIFO, so `cfg.Close()` fires before the pipe closes and the refresh goroutine can't write into a closed pipe. Easy to get wrong.
- `assert.Eventually` replacing the fixed `time.Sleep` is the right call — bounded wait instead of a heuristic.
- Calling out the gRPC and websocket races as explicitly out of scope rather than quietly skipping them. Both are real; I hit them in baseline runs too.

## Nits

- `TestFileTokenAuthConfig_RefreshFailureLogsWarning` drops the repeated `currentToken()` assertion (now a single check). Small coverage loss, easy to keep.
- Please strip the trailing generated-with footer from the PR description before merge — we keep merged history and PR bodies attributed to the human author.

akshat-kumar-singhal:
Thanks for the depth here — the timeline diagrams made both issues unambiguous. Decisions and changes below.

## 1. Lost shutdown — landing this as the memory-safety fix, tracking the ordering bug separately

Agreed on the diagnosis: the mutex closes the `-race`-visible read/write hazard but not the ordering, because `srv == nil` is doing double duty as the "started?" flag and a nil check carries no happens-before edge. I reproduced the same "shutdown returns nil, listener comes up afterwards" on this branch.

Deliberately **not** extending scope into it here, for two reasons:
- The fix reaches beyond `srv`: it needs a per-server `started` signal (that also has to cover *server never starts* — cert validation failure, blocked port — so `Shutdown` can't block forever), i.e. the `run.go` lifecycle, not just the three server files this PR touches.
- It's the same shape as the cron issue below, and you asked to fix it once — so folding half of it in here would be the worst of both.

Filed **#3801** covering both, and corrected this PR's description so it no longer claims to fix "the shutdown race" — it now says explicitly that it closes the memory hazard only, with the ordering bug pointing at #3801.

## 2. Cron `AnyTimes()` is load-bearing — you're right, and I did not add the join here

Correct that `Stop()` only halts future ticks and never joins an already-dispatched `go j.run(...)`, so `AnyTimes()` is what actually prevents the panic, not `defer c.Stop()`. Reworded the comment to say exactly that and to warn against deleting the expectations (48e3630).

I looked at adding the `WaitGroup` join and deliberately didn't, because `App.Shutdown(ctx)` calls `a.cron.Stop()` but `Stop()` takes **no context** — a naive `wg.Wait()` would let a hung/long-running cron job block shutdown past its timeout. So the real fix is `Stop(ctx)` + a bounded join wired through from `App.Shutdown`, which is exactly the context-aware design in #3801. Rather than narrow the test's assertions, I kept `AnyTimes()` and documented why.

## Nits

- Restored the "cached token remains available" assertion — now checked *after* repeated failed refreshes have run (strictly stronger than the original pre-failure check). (48e3630)
- Stripped the generated-with footer from the PR description.

Thanks for the callouts on the `t.Cleanup` LIFO ordering and the out-of-scope race labelling — and for independently hitting the gRPC/websocket races in baseline; those are logged for their own follow-ups.

aryanmehrotra (approved):
Approving. Re-verified everything on `48e3630`:

| Check | Result |
| --- | --- |
| `Test_Shutdown`, `-race -count=15` | 0 races (6 on `development`) |
| `TestCronTab*`, `-count=5` | pass |
| `TestFileTokenAuthConfig*`, `-race -count=15` | pass |
| `go vet` / `gofmt` on touched files | clean |
| CI | 14 pass, 2 skipping |

I also audited every access to `srv` across the three servers — all seven reads and writes are inside `srvMu`, the lock is never held across `ListenAndServe` or `srv.Shutdown`, and no test reaches the field directly. No fourth path was missed.

Both review points are resolved the right way:

- The scope correction on #1 is what I was after — this closes the `-race`-visible memory hazard, and the description now says exactly that instead of claiming the shutdown race generally, with #3801 carrying the ordering bug.
- On #2, the reworded comment now names `AnyTimes()` as load-bearing and warns against removing it, which is the part that would have bitten someone later.

Two things in your reply were better than what I suggested, and both went into #3801: that the `started` signal has to cover the *server never starts* case (cert validation failure, blocked port) or `Shutdown` blocks forever — my version would have deadlocked on a TLS misconfig — and that `Stop()` taking no context means a naive `wg.Wait()` would let a hung cron job block shutdown past its timeout, so the join has to be `Stop(ctx)` wired from `App.Shutdown`.

The restored `currentToken()` assertion is also stronger than what it replaced — checking the cached token survives *repeated* failed refreshes is the property the test was really claiming.

Good fix, and good follow-up issue.

Umang01-hash (approved):
Verified locally. Reproduced the race on `development` — `run()` writes `s.srv` while `Shutdown()` reads it (`-race` confirms), and it's gone at this head across 5 runs. The `srvMu` pattern (assign under lock, serve on a local copy) is applied uniformly to the http/mcp/metrics servers, and every access to the field is now under the lock. No exported-API change (all structs/fields unexported). De-flaked cron + file-token tests pass 10x under `-race`; the cron `AnyTimes()` + #3801 note is the right call. gofmt/vet/build clean.

One non-blocking note: the gRPC server (`grpc.go`) has the same class of race on `g.server` (still fires `-race` on both development and this branch). Out of scope for this PR — worth folding into the same pattern in a follow-up.

LGTM.

PR #3805, opened 2026-07-30
fix(datasource/s3): honor io.Reader contract, allow fresh-prefix Create, treat metadata as optional

### LINKED ISSUE #3804: datasource/file/s3: Read violates the io.Reader contract (silent data corruption), Create requires "parent" objects, and ContentType is dereferenced unconditionally
Three related defects in `gofr.dev/pkg/gofr/datasource/file/s3`. All are present in v0.2.2, v0.2.3, and v0.3.0 (latest). Line numbers below are v0.3.0 unless noted; the code is identical in v0.2.x at the noted offsets.

## Environment

- `gofr.dev/pkg/gofr/datasource/file/s3` v0.2.2 / v0.2.3 / v0.3.0
- Go 1.25 (also inspected under Go 1.26)
- Defects 1 and 2 reproduced against MinIO (latest, `server /data`), container-to-container
- Defect 3 is **not** reproducible on MinIO; it requires a backend that omits the `Content-Type` response header — see below

## 1. `S3File.Read` returns `len(p)` instead of bytes read — silent data corruption

`file.go:101` reads the HTTP body **once** into an oversized buffer, `file.go:110` advances the offset by the buffer length, and `file.go:117` returns `len(p)` regardless of how many bytes were actually read:

```go
n, err = f.body.Read(buffer)   // single read of an HTTP response body
...
f.offset += int64(len(buffer)) // not n
...
return len(p), err             // not n
```

`io.Reader` requires returning the number of bytes actually read. An HTTP response body returns one buffered chunk per `Read` call (a few KB), so for any object larger than that chunk, `Read` reports a full buffer while only the first few KB of `p` contain object data — the rest is zeros, with `err == nil`. There is no way for a caller to detect this.

Measured against MinIO (Open + one `Read` into an exactly-`Size()`-sized buffer):

| object size | bytes actually correct | returned n | err |
|---|---|---|---|
| 1 KiB | 1024 (all) | 1024 | EOF |
| 4 KiB | ~3450 | 4096 | nil |
| 64 KiB | ~3451 | 65536 | nil |
| 8 MiB | ~3449 | 8388608 | nil |

`io.ReadAll(f)` on a 40-byte object returns 512 bytes (40 real + 472 zeros), because ReadAll's 512-byte probe buffer is reported as fully read.

**Expected:** `Read` returns the actual count and loops (or documents) partial reads; `io.ReadAll` returns exactly the object's bytes.

**Actual:** any object larger than one transport chunk round-trips corrupted with no error.

`ReadAt` (`file.go:168`) has the same single-read pattern.

## 2. `Create` refuses any key whose parent prefix contains no objects

`fs.go:177-193` (v0.2.2: `135-151`): `Create(name)` runs ListObjectsV2 on `path.Dir(name) + "/"` and fails if `Contents` is empty:

```go
if len(res2.Contents) == 0 {
    return nil, fmt.Errorf("%w: create parent path before creating a file", ErrOperationNotPermitted)
}
```

S3 has no directories; a prefix "exists" only once an object is under it. So creating the first object under any new prefix — e.g. `Create("uploads/3f2a…-uuid/report.pdf")`, where a UUID segment is minted per upload — always fails. The only workaround is writing zero-byte `prefix/` marker objects first (`MkdirAll`), which S3 does not require and which pollutes listings.

**Expected:** `Create` writes the object (S3 semantics — prefixes need not "exist").

**Actual:** first write under any fresh prefix fails with `ErrOperationNotPermitted`.

Repro (empty bucket): `fs.Create("a/b/c.txt")` → `operation not permitted: create parent path before creating a file`.

## 3. Unconditional dereference of optional `GetObject.ContentType` — panic

`fs.go:199` (v0.2.2: `157`) stores whatever `mime.TypeByExtension(path.Ext(name))` returns, unchecked:

```go
ContentType: aws.String(mime.TypeByExtension(path.Ext(name))),
```

`TypeByExtension` returns `""` for extensionless keys on every Go version, and — on images without `/etc/mime.types` (e.g. alpine) under Go <= 1.25 — for common extensions like `.txt`, `.csv`, `.xlsx`, `.docx`. The read-back then dereferences the SDK's optional field unconditionally:

- `fs.go:228` (v0.2.2: `186`), in `Create`: `contentType: *res.ContentType`
- `fs.go:301` (v0.2.2: `259`), in `Open`: same dereference — reachable for any object stored without a content type by any client

The SDK models `ContentType` as `*string` because the response header is optional. Whether the pointer is nil therefore depends on the backend. MinIO substitutes `application/octet-stream`, which masks the defect entirely. Backends that store no content type for an empty value and omit the header on read-back surface it as a process-terminating panic; this has been observed against Cloudflare R2, on an alpine image under Go 1.25, for a key whose extension resolved to `""`.

**Expected:** treat `res.ContentType` (and ideally `LastModified` / `ContentLength`) as optional.

**Actual:** process panics when the backend omits the header.

## Reproduction

Single-file program for defects 1 and 2 against MinIO
(`docker run -d --name minio -e MINIO_ROOT_USER=minioadmin -e MINIO_ROOT_PASSWORD=minioadmin minio/minio server /data`):

```go
package main

import (
	"bytes"
	"context"
	"crypto/rand"
	"fmt"
	"io"

	"github.com/aws/aws-sdk-go-v2/aws"
	awsConfig "github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/credentials"
	"github.com/aws/aws-sdk-go-v2/service/s3"
	gofrs3 "gofr.dev/pkg/gofr/datasource/file/s3"
)

type nopLogger struct{}

func (nopLogger) Debug(...any)          {}
func (nopLogger) Debugf(string, ...any) {}
func (nopLogger) Logf(string, ...any)   {}
func (nopLogger) Errorf(string, ...any) {}

func main() {
	const endpoint, bucket = "http://minio:9000", "repro"
	ctx := context.Background()

	cfg, _ := awsConfig.LoadDefaultConfig(ctx, awsConfig.WithRegion("us-east-1"),
		awsConfig.WithCredentialsProvider(credentials.NewStaticCredentialsProvider("minioadmin", "minioadmin", "")))
	sdk := s3.NewFromConfig(cfg, func(o *s3.Options) { o.UsePathStyle = true; o.BaseEndpoint = aws.String(endpoint) })
	_, _ = sdk.CreateBucket(ctx, &s3.CreateBucketInput{Bucket: aws.String(bucket)})

	fs := gofrs3.New(&gofrs3.Config{EndPoint: endpoint, BucketName: bucket, Region: "us-east-1",
		AccessKeyID: "minioadmin", SecretAccessKey: "minioadmin"})
	fs.UseLogger(nopLogger{})
	fs.Connect()

	// Defect 2: first key under a fresh prefix
	_, err := fs.Create("a/b/c.txt")
	fmt.Println("Create under fresh prefix:", err)
	// operation not permitted: create parent path before creating a file

	// Defect 1: 64 KiB object, read back through the File interface
	orig := make([]byte, 64<<10)
	_, _ = rand.Read(orig)
	_, _ = sdk.PutObject(ctx, &s3.PutObjectInput{Bucket: aws.String(bucket), Key: aws.String("big.bin"),
		Body: bytes.NewReader(orig), ContentType: aws.String("application/octet-stream")})

	f, _ := fs.Open("big.bin")
	defer f.Close()
	got := make([]byte, len(orig))
	n, err := f.Read(got)
	match := 0
	for match < len(orig) && got[match] == orig[match] {
		match++
	}
	fmt.Printf("Read: n=%d err=%v, correct bytes=%d of %d\n", n, err, match, len(orig))
	// observed: n=65536 err=<nil>, correct bytes=~3450 of 65536 — remainder is zeros

	small, _ := fs.Open("big.bin")
	defer small.Close()
	all, _ := io.ReadAll(io.LimitReader(small, 512))
	_ = all
	// io.ReadAll(f) on a 40-byte object returns 512 bytes (40 real + 472 zero)
}
```

Defect 3 requires a backend that omits the Content-Type response header (observed with Cloudflare R2); on MinIO the dereference survives only because MinIO substitutes a default. The nil-dereference sites are visible by inspection at `fs.go:228` and `fs.go:301` (v0.3.0).

---

Happy to open a PR for any or all of these if that would help — let us know which direction you'd prefer for defect 2 in particular, since removing the parent-prefix check is a behaviour change for anyone currently relying on the error.

akshat-kumar-singhal:
## Summary

Fixes three related defects in `pkg/gofr/datasource/file/s3`, all reported in #3804 and all masked by the mock-based unit tests.

> **Headline (availability, not just corruption):** on `development`, `io.Copy(w, s3file)` — the most common way to stream a file — **never terminates and OOM-kills the process**. The old `Read` always returned `len(p)` with a nil error, so `io.Copy` never saw EOF, while each iteration reallocated a buffer that grew with the offset. This PR makes that path correct and linear.

### 1. `Read`/`ReadAt` violated the `io.Reader` contract (silent corruption + non-terminating `io.Copy`)
`Read` did a single `body.Read` into an oversized buffer, advanced the offset by the buffer length, and returned `len(p)` regardless of how many bytes were actually read. An HTTP response body returns one buffered chunk per `Read`, so for any object larger than that chunk, `Read` reported a full buffer while only the first few KB held real data — the rest was zeros, with `err == nil`. Because `Read` never signalled EOF, `io.Copy` looped forever. `ReadAt` had the same fill/return pattern.

Now both fill `p` with `io.ReadFull`, return the real byte count, and surface a short fill as `io.EOF`. To avoid re-downloading, they issue an HTTP **Range** request (`Read`: `bytes=<offset>-`; `ReadAt`: the exact `bytes=<off>-<off+len-1>`) instead of fetching the whole object and discarding the prefix — see the review-round changes below.

### 2. `Create` refused any key under a prefix containing no objects
`Create` ran `ListObjectsV2` on the parent prefix and failed with `ErrOperationNotPermitted` when it was empty. S3 has no directories — a prefix exists only once an object is under it — so the first write under any new prefix (e.g. `uploads/<uuid>/report.pdf`) always failed. The check is removed; `Create` writes the object directly.

> Note: this is an intentional behaviour change. Callers currently relying on the `ErrOperationNotPermitted` error for a missing parent will no longer receive it. This is now documented in `docs/advanced-guide/handling-file/page.md`.

### 3. Unconditional dereference of optional `GetObject` metadata (panic)
`Create` and `Open` dereferenced `*res.ContentType` / `*res.LastModified` / `*res.ContentLength` unconditionally. The SDK models these as pointers because the response headers are optional; backends such as Cloudflare R2 omit them, causing a nil-pointer panic. Both sites now go through a `newS3File` helper that nil-checks each field (MinIO masks this by substituting `application/octet-stream`).

## Review-round changes (#3805)

Addressing review feedback on the above:

- **Range requests instead of download-and-discard.** The original fix kept the "fetch whole object, discard the prefix with `io.CopyN`" model, which is **quadratic** — a 200 KiB object read via `io.ReadAll` pulled ~3.4x its size off the wire across 16 `GetObject` calls. `Read`/`ReadAt` now request only the bytes they need. `Read` omits the Range header at offset 0, so empty objects and `bytes=0-`-averse backends are unaffected.
- **Connection leak fixed.** The previous response body is now closed (`closeBody`) before `f.body` is reassigned.
- **`io.Copy` termination guard.** With Range requests, a final read at `offset == size` would send an unsatisfiable `bytes=<size>-` (HTTP 416). An early `offset >= size → io.EOF` guard keeps streaming clean even when the object size is an exact multiple of the caller's buffer.
- **`ReadAt` EOF boundary.** Dropped an off-by-one `+1` in the bounds check so a read ending exactly at EOF (the final byte, or the whole object in one call) is valid; the check now runs before the fetch, so out-of-range reads cost no request.
- **nil `Content-Length` is now diagnosable.** `newS3File` logs a warning when a backend omits the size (which leaves `ReadAt`/`Seek` treating the handle as empty). A `HeadObject` fallback to populate the size is tracked as a follow-up.

## Tests

- **Unit** (`file_test.go`, `fs_test.go`): `iotest.OneByteReader` reproduces chunked HTTP bodies (defect 1), a short-object read asserts real `n` + `io.EOF`, a fresh-prefix `Create` asserts no `ListObjectsV2` call (defect 2), and nil-metadata `GetObjectOutput` asserts no panic (defect 3). New in this round: Range headers are asserted on `Read`/`ReadAt`, `TestS3File_Read_OffsetAtEOF_ReturnsEOF` covers the `io.Copy` guard, and `TestS3File_ReadAt_Boundaries` covers reading the final byte / whole object. Each fails on the pre-fix code.
- **Integration** (`examples/using-s3-filestore`): a new example that writes/reads objects through the file store, with `main_test.go` driving it against a real MinIO backend — the transport-level guard the mock suite can't provide. It round-trips a 216 KiB object (defect 1) and creates under a fresh prefix (defect 2).

## CI

The S3 example is its own submodule (it depends on the `s3` datasource submodule), so `go test ./examples/...` doesn't reach it. The `Example-Unit-Testing` job gets an explicit step that runs it against a MinIO container. MinIO is started with `docker run` rather than a `services:` container because the stock image needs a `server /data` command that Actions services can't supply (same pattern as the existing Zipkin step).

The readiness loop now **fails the step** (`exit 1`, with `docker logs`) if MinIO never comes up, and the test step sets `S3_REQUIRE_MINIO=true` so an unreachable backend **fails the test** rather than skipping — an image-pull failure or registry rate-limit can no longer turn the #3804 integration guard into a silent no-op. Locally the variable is unset, so `go test` still skips for contributors without the backend.

Closes #3804


aryanmehrotra (changes requested):
Reviewed end to end, ran it locally against a real MinIO backend, and did a before/after sweep of the public `file.File` surface on this branch vs `development`.

**The three fixes are correct and this is a solid improvement.** Four things to address first.

---

## The headline is bigger than the description says

On `development` today, `io.Copy(w, s3file)` is **an infinite loop that OOM-kills the process** — my test binary was killed after 48s on a 50 KB object.

Why: old `Read` always returned `len(p)` with a nil error, so `io.Copy` never sees EOF. Meanwhile `make([]byte, len(p)+int(f.offset))` reallocates with a growing offset every iteration.

That's an availability bug reachable from the most common way to stream a file — not just silent corruption. Worth calling out in the description and changelog.

---

## Verified working

Wired the store exactly as `AddFileStore` does, exercised the public surface against live MinIO:

| Flow | `development` | This PR |
|---|---|---|
| `io.Copy`, 50 KB | never terminates, OOM-killed at 48s | 50000 bytes exact |
| `io.ReadAll`, nested key | content + ~700 NUL bytes | exact |
| `bufio.Scanner`, 3 lines | 3 lines **+ phantom NUL line** | 3 lines |
| `WriteAt` then read back | `AAAABBAAAA` + NUL padding | `AAAABBAAAA` |
| `Seek`+`Read`, `ReadAt`, `Stat`, `ReadDir`, `Rename`/`Remove` | correct | correct |
| `ReadAll()` JSON/CSV RowReader | correct | correct |

Also clean:
- unit tests under `-race`
- `golangci-lint` on the package **and** the new example (the one `prealloc` hit at `fs_test.go:381` is pre-existing on `development`)
- root `go build` / `go vet` with the `go.work` addition
- `go 1.26.0` across the CI matrix — the existing `GOTOOLCHAIN=auto` step covers it
- env precedence: `S3_ENDPOINT` from the workflow wins over `configs/.env` via the restore loop in `config/godotenv.go`

Also — nice catch on `f.logger.Logf("%v bytes read successfully.")` with a format verb and no argument. That one isn't even in the description.

---

## Changes requested

### 1. Use a Range header instead of download-and-discard

`Read` issues a fresh `GetObject` per call, then `io.CopyN`s `f.offset` bytes into `io.Discard`.

Instrumented with a counting body — 200 KiB object via `io.ReadAll`:

```
GetObject calls = 16
bytes off wire  = 699392   (3.4x object size)
Range headers   = []
bodies closed   = 0 of 16 opened
```

Quadratic in object size. A 1 GiB object would be pathological.

The request is already being made, so it's one line:

```go
Range: aws.String(fmt.Sprintf("bytes=%d-", f.offset)),
```

That deletes the `io.CopyN` discard entirely. Same for `ReadAt`: `bytes=<off>-<off+len(p)-1>`.

Separately: `f.body` should be closed before it's reassigned. The leak is pre-existing, but each abandoned body now costs a fully-transferred prefix.

### 2. `ReadAt` still can't read the last byte

`file.go:175` — `int64(len(p))+offset+1 > f.size`. On a 10-byte object:

```
ReadAt(len=10, off=0) -> ErrOutOfRange   // can never read the whole object
ReadAt(len=1,  off=9) -> ErrOutOfRange   // can never read the last byte
ReadAt(len=4,  off=3) -> n=4, "3456"     // interior reads fine
```

Pre-existing — but this PR is about `ReadAt`'s contract and edits that exact line. Drop the `+1`.

Strictly, `io.ReaderAt` wants bytes-read plus `io.EOF` rather than a custom error, which the new `ReadFull` path is already positioned to return.

### 3. `Create`'s godoc contradicts the change

`fs.go:162-163` still reads:

> It first checks if the parent directory exists; if the parent directory does not exist, it returns an error.

### 4. The CI guard is silently skippable

Verified: with the port closed, the test skips and the package reports `ok`, exit 0.

The readiness loop doesn't fail when it exhausts either — it just falls through after 60s.

So an image-pull failure or registry rate-limit turns the integration guard for #3804 into a no-op nobody notices. Fail the step when the loop runs out:

```yaml
echo "MinIO failed to become ready"; exit 1
```

...and/or gate the skip on an env var so CI treats the test as required rather than optional.

---

## Smaller points

- **nil `ContentLength` → silently unusable handle.** `size` falls back to 0, which makes `ReadAt` always return `ErrOutOfRange` and `Seek` reject any non-zero offset (`check` bounds against `f.Size()`). Better than panicking, but for a backend that genuinely omits the header the file is broken with no diagnostic. A `HeadObject` fallback, or at minimum a logged warning, would help. `Test_Open_NilOptionalMetadata_NoPanic` currently asserts `assert.Zero(t, ...size)` — that locks the degraded behaviour in.

- **`Read()` then `ReadAll()` on one handle** yields a phantom empty first row, since `ReadAll` consumes what the last `Read` left behind. Pre-existing, and better than `development` (which loses rows entirely) — but worth a doc note that a handle is for streaming reads *or* `ReadAll()`, not both.

- **The behaviour change deserves a docs line.** Dropping the parent check is right for S3 semantics, but `docs/advanced-guide/handling-file/page.md` documents neither the old nor the new behaviour, and callers matching on `ErrOperationNotPermitted` will just stop seeing it.

- **`go main()` + `time.Sleep(200ms)`** in `main_test.go` is a flake risk. Matches the existing example convention so it's fine, but polling the health endpoint would be sturdier.

Umang01-hash (dismissed):
Verified locally. Ran the example against a real MinIO — the 216 KiB object round-trips byte-for-byte and the fresh-prefix create works. Also reproduced defect 1 on `development` (the new Read tests fail there with zero-fill corruption / swallowed EOF) and confirmed they pass at this head, so the fix and the tests both hold up. All three fixes read correctly, no exported-API change, and the DRY `newS3File` helper is a nice touch. gofmt/vet/build/tests clean.

Two non-blocking notes on `ReadAt` (both pre-existing, not introduced here):
- `ReadAt` gates on `f.size`, but `newS3File` leaves size=0 when `ContentLength` is nil — the exact R2 case defect 3 targets. So `ReadAt` on a ContentLength-omitting backend still fails (every read is out-of-range). It's a net improvement over the old panic, but the metadata-optional goal isn't fully delivered for `ReadAt` — worth a follow-up.
- The bounds check `len(p)+offset+1 > f.size` has a spurious `+1` that rejects a valid read ending exactly at EOF. Identical on `development`, so pre-existing — but the rewrite is right there if you want to fix it now.

Neither blocks. Nice work — approving.

akshat-kumar-singhal:
@aryanmehrotra thanks for the thorough review (and the live-MinIO sweep) — all four requested changes plus the smaller points are in, pushed as `59552733`.

### Changes requested

**1. Range header instead of download-and-discard.** `Read` now sends `Range: bytes=<offset>-` (omitted at offset 0, so empty objects / `bytes=0-`-averse backends are unaffected) and `ReadAt` sends the exact `bytes=<off>-<off+len(p)-1>`. The `io.CopyN(io.Discard, …)` loops are gone, so reads are linear. The abandoned body is now closed via a `closeBody()` helper before `f.body` is reassigned.

> One extra thing the Range change forced: a final read at `offset == size` would send an unsatisfiable `bytes=<size>-` (416). Added an early `if f.size > 0 && f.offset >= f.size { return 0, io.EOF }` guard so `io.Copy` still terminates cleanly when the object size is an exact multiple of the caller's buffer (`TestS3File_Read_OffsetAtEOF_ReturnsEOF`).

**2. `ReadAt` last byte.** Dropped the `+1` — the bound is now `len(p)+offset > f.size`, and it's checked *before* `GetObject` so an out-of-range read costs no request. Covered by `TestS3File_ReadAt_Boundaries` (whole object + final byte).

**3. `Create` godoc.** Rewritten to describe the S3-native no-parent-check behaviour.

**4. CI guard.** The MinIO readiness loop now `exit 1`s (with `docker logs`) on timeout, and the test step sets `S3_REQUIRE_MINIO=true` so an unreachable backend **fails** rather than skips. Verified locally: skips when the var is unset, fails when it's set and MinIO is down.

### Smaller points

- **nil `Content-Length`:** `newS3File` now logs a warning so the size-unknown handle is diagnosable. The `HeadObject` fallback to actually populate `size` (which would also make `ReadAt`/`Seek` work on those backends) is noted as a follow-up.
- **Docs:** added an "S3 semantics" note to `docs/advanced-guide/handling-file/page.md` — `Create` needs no parent, a handle is for streaming reads *or* `ReadAll()` (not both), and the nil-`Content-Length` caveat.
- **`Write`/`WriteAt` body leak** and the **`go main()`+sleep** flake: acknowledged as pre-existing, tracked as follow-ups (kept this pass scoped to the read contract).

Description updated with the `io.Copy` non-termination headline. Ready for another look 🙏

aryanmehrotra (changes requested):
Re-reviewed at `59552733`. All four requested changes are in and verified — the Range fix takes the 200 KiB `io.ReadAll` probe from 3.4x object size off the wire to exactly 1.00x, bodies are closed, `ReadAt` handles the last byte, the `Create` godoc matches, and the CI guard now fails instead of skipping. Tests green under `-race`, lint clean on the package and the new example.

Six things left, detail in the inline threads:

1. **`Range` can silently serve the wrong bytes** (`file.go:113`, `file.go:192`) — a backend that ignores `Range` answers 200 + whole body, and `io.ReadFull` then fills from byte 0 with `n` correct and `err == nil`. `development` was safe here because `io.CopyN(io.Discard, …)` positioned the stream regardless. Guard on `res.ContentRange == nil`. This is the one I'd really like fixed — it's a new silent-corruption path and needs a test.
2. **`CreateWithOptions` still has defect 3** (`create_options.go:101-103`, [comment](#issuecomment-5132201386)) — third construction site, all three optional fields dereferenced unconditionally, so the R2 panic stays reachable.
3–5. **`ReadAt` bounds** (`file.go:180`) — a read straddling EOF returns zero bytes instead of `n, io.EOF`; a negative offset isn't rejected; `len(p) == 0` costs a request. One edit to the same guard.
6. **"warning" logs at info** (`fs.go:240`) — add `Warnf` to the package `Logger`.

## Breaking changes

No exported API change — `go doc -all` against `development` is identical, and nothing is removed, renamed or retyped. Behaviourally: `Create` stops returning `ErrOperationNotPermitted` for a missing parent (documented — please keep it in the release notes), and `Read`/`ReadAt` now report real `n` + `io.EOF`, which is the fix. The one worth gating on is `Range` going on the wire, i.e. item 1.

## Follow-ups

Raised #3809 — S3 is the only file store still carrying a hand-rolled copy of `CommonFileSystem`, and items 1 and 2 are both symptoms of that. Nothing to do here; please don't rework `file.go`'s internals in this PR, that layer is going away.

Smaller, non-blocking: pin the MinIO image now that the step is required; `go work edit -fmt`; the integration test only exercises `io.ReadAll`, so an `io.Copy` and a `ReadAt` there would have caught item 1.

Items 3–6 are small. Item 1 is the one that matters.

pkg/gofr/datasource/file/s3/file.go
```diff
@@ -81,40 +90,64 @@ func (f *S3File) Read(p []byte) (n int, err error) {
...
-	})
+	}
+
+	// Ask S3 for only the bytes from the current offset onward instead of
+	// downloading the whole object and discarding the prefix on every call.
+	// That keeps a full-file read linear in the object size rather than
+	// quadratic. A zero offset fetches the whole object (no Range header), so
+	// empty objects and backends that reject "bytes=0-" behave as before.
+	if f.offset > 0 {
+		input.Range = aws.String(fmt.Sprintf("bytes=%d-", f.offset))
```
aryanmehrotra, line 137: **Item 1 — the Range switch can silently serve the wrong bytes here.**

RFC 9110 lets a server that won't apply a range answer **200 with the whole body** and no `Content-Range`. We hand that body straight to `io.ReadFull(f.body, p)` below, which fills `p` from **byte 0** and reports success:

```
Read(len=4) at offset 10 -> n=4, data="0123", err=<nil>   // correct answer: "ABCD"
```

On `development` this path was correct — `io.CopyN(io.Discard, body, offset)` positions the stream whether or not the backend supports ranges. So we'd be trading a known corruption bug for an invisible one: right `n`, `nil` error, wrong bytes.

`GetObjectOutput.ContentRange` is set only when the range was applied, so we can tell:

```go
if f.offset > 0 && res.ContentRange == nil {
	// Server ignored Range and sent the whole object; position the stream
	// ourselves rather than serving bytes from offset 0.
	if _, err = io.CopyN(io.Discard, f.body, f.offset); err != nil {
		return 0, err
	}
}
```

The fast path is untouched — the discard only runs on a backend that didn't honour the range. Same guard needed in `ReadAt` (see the thread on L192). Please add a test with a client that answers 200 + full body.

pkg/gofr/datasource/file/s3/file.go
```diff
@@ -141,44 +174,53 @@ func (f *S3File) ReadAt(p []byte, offset int64) (n int, err error) {
...
+	}
+
 	var res *s3.GetObjectOutput
 
+	// Request exactly the [offset, offset+len(p)-1] byte range instead of the
+	// whole object, so ReadAt is linear in len(p) rather than in the object size.
 	res, err = f.conn.GetObject(context.TODO(), &s3.GetObjectInput{
 		Bucket: aws.String(bucketName),
 		Key:    aws.String(fileName),
+		Range:  aws.String(fmt.Sprintf("bytes=%d-%d", offset, offset+int64(len(p))-1)),
```
aryanmehrotra, line 234: **Item 1, `ReadAt` half.** Same problem as the thread on L113, and worse here because `ReadAt` always sends a Range (no offset-0 exemption):

```
ReadAt(len=4, off=10) -> n=4, data="0123", err=<nil>   // correct answer: "ABCD"
```

Reproduced against a client that answers 200 + whole body. Guard on `res.ContentRange == nil` before the `io.ReadFull` and fall back to discarding `offset` bytes.

pkg/gofr/datasource/file/s3/file.go
```diff
@@ -141,44 +174,53 @@ func (f *S3File) ReadAt(p []byte, offset int64) (n int, err error) {
 		fileName = f.name[index+1:]
 	}
 
+	// A read of len(p) bytes starting at offset ends at offset+len(p)-1, so it
+	// is in range as long as offset+len(p) <= size. The bound is checked before
+	// fetching so an out-of-range call costs no request.
+	if int64(len(p))+offset > f.size {
```
aryanmehrotra: **Items 3–5 — three problems in this one guard.**

**Straddling EOF still can't be read.** On a 10-byte object:

```
ReadAt(len=5, off=8) -> n=0, ErrOutOfRange     // no request issued, no bytes returned
```

`io.ReaderAt` lets a short read return an error, but it must return **the bytes it read**. This returns zero and never issues the request, so `io.SectionReader` and any tail read break. Same class as defect 1 — `n` doesn't describe reality.

**A negative offset passes.** `int64(len(p))+offset > f.size` is false for `offset = -2`, and L192 formats it straight into the header: `Range: "bytes=-2-0"`. Real S3/MinIO answer 416, but on a backend that ignores an unparseable Range this lands in the silent-wrong-data path above. `io.ReaderAt` requires an error here.

**`len(p) == 0` costs a request and errors.** Emits `bytes=3-2` (end < start). `bytes.Reader`/`os.File` return `0, nil` without touching the source.

`CommonFile.ReadAt` (`common_file.go:139`) is the reference — gate on `offset` only and clamp the range:

```go
if len(p) == 0 {
	return 0, nil
}

if offset < 0 || offset >= f.size {
	msg = fmt.Sprintf("Offset %v out of range", offset)
	return 0, fmt.Errorf("%w: reading out of range, fetching from the offset. Use Seek to reset offset", ErrOutOfRange)
}

end := offset + int64(len(p)) - 1
if end >= f.size {
	end = f.size - 1 // ReadFull surfaces the short fill as io.EOF
}
```

One edit covers all three. Worth extending `TestS3File_ReadAt_Boundaries` with the straddle, negative-offset and empty-buffer cases.

pkg/gofr/datasource/file/s3/fs.go
```diff
@@ -219,16 +205,43 @@ func (f *FileSystem) Create(name string) (file.File, error) {
...
+	}
+
+	if res.ContentLength != nil {
+		s3File.size = *res.ContentLength
+	} else {
+		// A backend that omits Content-Length (e.g. some Cloudflare R2 responses)
+		// leaves size at 0, which makes offset-bounded operations such as ReadAt
+		// and Seek treat the handle as empty. Surface it so the degraded handle is
+		// diagnosable rather than silently unusable.
+		f.logger.Logf("S3 object %q returned no Content-Length; size is unknown and "+
```
aryanmehrotra: **Item 6 — this is described as a warning in three places but logs at info.**

The comment above, the PR description and the new note in `docs/advanced-guide/handling-file/page.md` all call this a warning, but it goes out through `Logf`.

Our `Logger` in `interface.go:11` doesn't declare `Warnf` — the underlying gofr logger has it (`logging/logger.go:36`), and `pubsub/mqtt/interface.go:19` already declares it locally for exactly this reason. Since the user of an interface defines it, lets add `Warnf(pattern string, args ...any)` here and use it, so operators can filter on level.

aryanmehrotra:
**Item 2 — `create_options.go:101-103`** (not inline: the file is not in this diff, so GitHub cannot anchor a thread to it)

**Item 2 — defect 3 is still live on this third construction site.**

The nil-metadata fix converted `Create` and `Open` to `newS3File`, but `CreateWithOptions` still builds `S3File` by hand with all three optional fields dereferenced unconditionally:

```go
contentType:  *res.ContentType,
lastModified: *res.LastModified,
size:         *res.ContentLength,
```

`CreateWithOptions` is exported and predates this branch (it's in the merge base), so the Cloudflare R2 nil-pointer panic #3804 reports is still reachable through it.

```go
return f.newS3File(name, res), nil
```

Plus a nil-metadata case in `create_options_test.go`, mirroring `Test_CreateFile_NilOptionalMetadata_NoPanic`. Worth a grep for a fourth site while you're in here.

Having to keep three construction sites in sync is itself the argument for #3809.

pkg/gofr/datasource/file/s3/file.go
akshat-kumar-singhal: Fixed in 423c8b79 — but I went a different way than the discard fallback, so flagging it explicitly for your call.

Rather than positioning the stream with `io.CopyN(io.Discard, f.body, f.offset)`, `Read` now **refuses** when it asked for a range and didn't get one:

```go
if f.offset > 0 && res.ContentRange == nil {
    f.closeBody()
    f.body = nil
    return 0, ErrRangeNotHonored
}
```

Reasoning: the discard fallback re-introduces the exact download-the-whole-object-and-discard-the-prefix behaviour this PR set out to remove — and for streaming `Read` it's quadratic across successive calls (each re-fetches the full body and discards a growing prefix). I'd rather fail loud and diagnosable than silently fall back to the O(n²) path on a non-compliant backend. The fast path is untouched, and every real S3-compatible store (S3/MinIO/R2/Spaces/B2) sets `Content-Range`, so this only bites a backend that ignores `Range` on `GetObject`.

The trade-off is that such a backend can no longer be read at a non-zero offset at all (a regression vs `development`, which positioned via `CopyN`). If you'd rather keep parity there, I'm happy to switch to your discard fallback instead — your call, since it's your "the one that matters."

Test: `TestS3File_Read_RangeIgnored_Refuses` — mock returns 200 + whole body with no `Content-Range`, asserts `ErrRangeNotHonored` and `n == 0`.

pkg/gofr/datasource/file/s3/file.go
akshat-kumar-singhal: Same fix in 423c8b79. `ReadAt` always sends a `Range`, so there's no offset-0 exemption — it checks `res.ContentRange == nil` unconditionally (after the nil-body check) and returns `ErrRangeNotHonored`:

```go
if res.ContentRange == nil {
    f.closeBody()
    f.body = nil
    return 0, ErrRangeNotHonored
}
```

Same refuse-vs-discard trade-off as the `Read` thread on L137 — see there for the rationale and the offer to switch to the discard fallback if you'd prefer parity with `development`. Test: `TestS3File_ReadAt_RangeIgnored_Refuses`.

pkg/gofr/datasource/file/s3/file.go
akshat-kumar-singhal: Fixed in 423c8b79 — the single guard is now three cheap pre-request checks, matching the `CommonFile.ReadAt` reference you pointed at:

```go
if offset < 0 {
    return 0, fmt.Errorf("%w: negative offset %d", ErrOutOfRange, offset)
}
if len(p) == 0 {
    return 0, nil
}
if offset >= f.size {
    return 0, io.EOF
}
```

- **Straddling EOF** now proceeds: the request goes out, S3 returns the available bytes, and `normalizeReadFull` surfaces the short fill as `n, io.EOF` — so `n` describes reality. (I let S3 clamp the range rather than clamping `end` myself; the over-long `bytes=8-11` on a 10-byte object comes back as `bytes 8-9/10`.)
- **Negative offset** → `ErrOutOfRange`, no request.
- **`len(p) == 0`** → `0, nil`, no request.

Tests: `TestS3File_ReadAt_StraddlesEOF` (off=8/len=4 on a 10-byte object → n=2, io.EOF), `TestS3File_ReadAt_EmptyBuffer`, and the `Failure_NegativeOffset` / `Failure_OffsetAtEOF` cases split out of the old `Failure_OutOfRange`.

pkg/gofr/datasource/file/s3/fs.go
akshat-kumar-singhal: Fixed in 423c8b79. Added `Warnf(pattern string, args ...any)` to the package `Logger` in `interface.go` (and the generated `mock_interface.go`), and the nil-`Content-Length` message now goes out through `f.logger.Warnf` so operators can filter on level. Follows the same local-declaration pattern as `pubsub/mqtt/interface.go` you referenced.

akshat-kumar-singhal:
@aryanmehrotra all six items addressed in `423c8b79`. Replies are on the inline threads; summarising here, plus item 2 (which had no inline anchor) and the non-blocking notes.

### Requested changes

1. **Range can silently serve the wrong bytes** — `Read`/`ReadAt` now guard on `res.ContentRange == nil` and **refuse** with a new `ErrRangeNotHonored`. I deliberately chose refuse over the discard fallback (it would re-introduce the quadratic download-and-discard this PR removes) — rationale and an offer to switch to your fallback if you'd rather keep parity are on the L137 thread. Tests: `TestS3File_Read_RangeIgnored_Refuses`, `TestS3File_ReadAt_RangeIgnored_Refuses`.
2. **`CreateWithOptions` still had defect 3** — it now builds via `f.newS3File(name, res)` instead of dereferencing `*res.ContentType` / `*res.LastModified` / `*res.ContentLength`, so the R2 nil-pointer panic is closed on the third construction site too. Test: `Test_CreateWithOptions_NilOptionalMetadata_NoPanic`. (Grepped for a fourth site — the three that dereference these fields were `Create`, `Open`, `CreateWithOptions`; all now go through `newS3File`.)
3–5. **`ReadAt` bounds** — one guard became three cheap pre-request checks: negative offset → `ErrOutOfRange`, `len(p)==0` → `0, nil`, `offset >= size` → `io.EOF`. A straddling read now issues the request and returns `n, io.EOF` for the bytes actually read. Tests: `TestS3File_ReadAt_StraddlesEOF`, `TestS3File_ReadAt_EmptyBuffer`, and split `NegativeOffset` / `OffsetAtEOF` failure cases.
6. **Warning logged at info** — added `Warnf` to the package `Logger` (+ mock); the nil-`Content-Length` message now uses `Warnf`.

### Non-blocking notes

- **Pinned MinIO** by digest — `minio/minio:RELEASE.2025-09-07T16-13-09Z@sha256:14cea49…` — now that the step is required.
- **`go work edit -fmt`** — applied (sorted the `using-s3-filestore` entry).
- **Integration `io.Copy` + `ReadAt`** — new `TestS3FileStore_CopyAndReadAt` drives the ranged read paths directly through the file store; verified green against a real MinIO alongside the existing round-trip.

Re #3809 — noted, I've kept this to the read/metadata contract and haven't reworked `file.go`'s internals beyond two small private helpers (`relativeKey`, `normalizeReadFull`) needed to keep `Read`/`ReadAt` under gocyclo=10 after the new guards.

Ready for another look 🙏

aryanmehrotra (approved):
Re-reviewed at `423c8b79`. **All six items from the last round are in and verified**, including the two I'd have accepted a weaker answer on. **Approving.**

One finding below that I'd have raised as a blocker if it were reachable from anything the framework does — it isn't, and it belongs to #3809. Flagging it as a take-it-or-leave-it, with the patch, because it's four lines.

## Item 1 — I was wrong, your version is better

I asked for the `io.CopyN` discard fallback. You pushed back that it re-introduces the O(n²) path this PR exists to delete, and refused with `ErrRangeNotHonored` instead. **You're right and I'm withdrawing the ask.**

My "parity with `development`" argument doesn't survive contact: `development`'s `Read` was broken for *every* backend, so there is no working behaviour being regressed — a Range-ignoring backend goes from silently-corrupt to loudly-refused, which is strictly better. And a fail-closed named error is diagnosable in a way a silent quadratic fallback never is. Keep it as written.

Verified against real MinIO that the guard doesn't misfire — this was my one worry with a fail-closed check, since a false positive breaks reads outright:

```
TestS3FileStore_RoundTrip     PASS   (216 KiB, fresh prefix)
TestS3FileStore_CopyAndReadAt PASS   (io.Copy byte-exact, ReadAt window exact)
```

## Verified fixed

| Item | Status |
|---|---|
| 1. Ignored-`Range` silent corruption (`Read` + `ReadAt`) | fixed — refuses with `ErrRangeNotHonored`, tests both halves |
| 2. `CreateWithOptions` nil-metadata panic | fixed — routed through `newS3File`, test added |
| 3. `ReadAt` straddling EOF | fixed — `ReadAt(len=4, off=8)` on 10 bytes → `n=2, io.EOF` (confirmed on live MinIO) |
| 4. Negative offset | fixed — `ErrOutOfRange`, no request |
| 5. `len(p) == 0` | fixed — `0, nil`, no request |
| 6. Warning logged at info | fixed — `Warnf` on the package `Logger` + regenerated mock |

Also clean, all re-run at this head:
- unit tests under `-race`; `golangci-lint` on the package (the one `prealloc` at `fs_test.go:381` is pre-existing on `development`) and **0 issues** on the new example
- root `go build ./pkg/...` + `go vet`; `go work edit -fmt` is a no-op — formatting is clean
- **the MinIO digest pin is correct**: `sha256:14cea493…` is the multi-arch manifest list for `RELEASE.2025-09-07T16-13-09Z` (`linux/arm64`, **`linux/amd64`**, `linux/ppc64le`) and matches the tag's current digest. Ran the CI image verbatim. Worth stating explicitly because a digest pinned from an arm64 laptop would have broken every amd64 runner on a now-*required* step.
- live behaviour sweep beyond what the tests cover: empty object via `io.Copy` (`0, nil`, terminates), object sized exactly one `io.Copy` buffer (32768, terminates), `bufio.Scanner` (3 lines, no phantom row), `Seek(10)+Read` (exact bytes), `ReadAt` whole-object / last-byte / past-EOF — all correct.
- **the new tests are load-bearing, not decoration** — mutation-tested: deleting both `ContentRange` guards fails `TestS3File_Read_RangeIgnored_Refuses` + `TestS3File_ReadAt_RangeIgnored_Refuses`; restoring the old `+1` bounds fails 7 `ReadAt` tests/subtests.
- **the CI guard actually guards**: `S3_REQUIRE_MINIO=true` with the backend down **fails** the run; unset with a closed port **skips** cleanly. Both re-verified at this head.

### Ran against real AWS S3 and against emulated R2

MinIO isn't a perfect S3 stand-in, and this PR changed *how the body is consumed* (single `Read` → `io.ReadFull` draining to EOF), which newly reaches the SDK's checksum-validating reader that MinIO never triggers (`Response has no supported checksum` on every MinIO GET). So I ran the store against genuine AWS S3 — **every case matched MinIO byte-for-byte**: `io.Copy` and `io.ReadAll` at 225 KB, an exact-32 KiB multiple, `Seek`+`Read`, and the full `ReadAt` matrix (whole / last byte / straddling EOF `n=2,EOF` / past EOF / over-length / empty buffer / negative offset). The checksum divergence I was worried about didn't materialise. Bucket created and deleted for the run.

Real AWS can't test defect 3 though — it always sets those headers; **R2** is the backend that omits them. So I emulated R2 with a proxy that strips `Content-Length`/`Content-Type`/`Last-Modified` from every GET, and a second that strips `Content-Range` and rewrites 206→200:

- **R2 emulation**: `Open` no panic, `io.Copy` returns **90000 bytes byte-exact**. Defect 3's fix and streaming on a metadata-less backend now proven over real HTTP rather than gomock.
- **Range-ignored emulation**: `ErrRangeNotHonored` fires on both `ReadAt` and `Seek`+`Read`, and an offset-0 read is unaffected — the exemption behaves as designed. Item 1 is now transport-proven too.

---

## Your call — `ReadAt` is not safe for parallel calls

`file.go:243-245`. `ReadAt` stashes its response on the shared handle:

```go
f.closeBody()
f.body = res.Body
```

`io.ReaderAt` is explicit that this must be safe:

> Clients of ReadAt can execute parallel ReadAt calls on the same input source.

It isn't. Eight goroutines calling `ReadAt` on one handle, under `-race`:

```
WARNING: DATA RACE
  Write at 0x00c000757890 by goroutine 239:
    (*S3File).ReadAt()  file.go:245
  Previous read at 0x00c000757890 by goroutine 238:
    (*S3File).ReadAt()  file.go:246
```

And it isn't only a race on the pointer — the `f.closeBody()` at :243 closes a body another goroutine is mid-`ReadFull` on. 32 goroutines reading disjoint 4 KiB windows of a 200 KB object on live MinIO, reproducible across runs:

```
parallel ReadAt: 2/32 failed
  read tcp …->…:9000: use of closed network connection : 2
```

That's `archive/zip`, `io.SectionReader` fan-out, and any parallel chunked reader — the callers `ReadAt` exists for.

**Why this isn't a blocker.** I checked the reachability rather than assuming it: there are **zero non-test callers of `ReadAt` in the repo** — `pkg`, `examples` and `docs` all come back empty; `io.ReaderAt` appears only in the two interface declarations and `common_file.go`. So the trigger needs a user to write against `file.File` directly, reach for `ReadAt` with no in-repo precedent, *and* share one handle across goroutines. Every single-threaded path is correct, and the failure mode I could reproduce is a connection error, never wrong bytes.

The pointer race is pre-existing on `development` too; only the `closeBody()` that escalates it to a hard read error is new here. And I asked you last round *not* to rework `file.go`'s internals because #3809 removes this layer — I'm not going to turn around and gate on exactly that.

Recording it so #3809 inherits it: `CommonFileSystem` should own body lifetime, and `ReadAt` shouldn't touch shared state at all — it doesn't advance `f.offset`, and stashing the body also keeps the last `ReadAt`'s connection open until the next read or `Close`.

If you'd rather take it here, it's a local body and nothing else:

```go
// ReadAt owns this body: io.ReaderAt permits parallel calls on one handle, so
// the response must not be stashed on the shared f.body.
body := res.Body
if body == nil {
	msg = fmt.Sprintf("File %q is nil", fileName)
	return 0, io.EOF
}

defer body.Close()

if res.ContentRange == nil {
	msg = fmt.Sprintf("Range request for %q not honored by backend", fileName)
	return 0, ErrRangeNotHonored
}

n, err = normalizeReadFull(body, p)
```

I applied exactly this and re-ran:

```
parallel ReadAt: 0/32 failed        (×3 runs, -race clean)
package tests: ok, -race            lint: unchanged
```

No existing test depends on `ReadAt` setting `f.body`. Either way it wants a concurrent-`ReadAt` test so `-race` in CI holds the line — here or on #3809.

---

## Smaller points

- **Widening the `Logger` interface is a silent break for external implementers.** `UseLogger(logger any)` (`fs.go:81`) swallows a failed type assertion, so a caller whose logger lacks `Warnf` no longer satisfies `Logger`, `f.logger` stays nil, and the first `Logf`/`Debug` nil-panics — no compile error anywhere. Framework wiring is safe (`logging.Logger` has `Warnf`), and this matches the `pubsub/mqtt` precedent, so I'm not blocking. But this package is its own published module; either note it in the release notes or use a narrow optional assertion at the one call site.

- **`ErrRangeNotHonored` deserves a docs line.** It's a new exported error and a new *terminal* failure mode (a Range-ignoring backend can't be read at a non-zero offset at all). The `docs/advanced-guide/handling-file/page.md` block you added is the right home, next to the `Content-Length` bullet.

- **nil `Content-Length` now reads as an empty file rather than an error** — and this one **reproduces on the R2 emulation**, it isn't theoretical: `ReadAt(len=10, off=5)` returns `0, io.EOF`, so a caller concludes the object is empty when it has 90 KB in it. With `size = 0` every offset trips the new `offset >= f.size` guard; it was `ErrOutOfRange` before, which was at least loud. Still not a blocker — streaming works fine on that backend, the `Warnf` makes it diagnosable, and it beats `development`'s panic outright — but it's the clearest argument for #3809 and I've carried the repro over there.

- **The description is a round behind.** It documents the round-2 changes; the refuse-vs-discard decision (and the reasoning in your thread, which is the better argument) should be in the body and the release notes.

Item 1's resolution is the right call, and the argument you made for it is better than the one I made against it. Nice work across four rounds — approving.

Issue #3806, opened 2026-07-30
Serve detailed health map on the metrics port (follow-up to #3802)


akshat-kumar-singhal:
## Summary

Follow-up to #3802. That issue's unauthenticated infrastructure-disclosure via `/.well-known/health` is now closed by redacting the public endpoint to `{name, status}` (PR TBD). This ticket tracks the **second half** of #3802's proposed design: relocate the *detailed* health map to the metrics server so ops tooling can still consume it, behind the same network boundary as `/metrics` and `/debug/pprof`.

## Current state (after the redaction fix)

| Endpoint | Port | Body |
|---|---|---|
| `/.well-known/alive` | `HTTP_PORT` (8000) | `{"status":"UP"}` — liveness (unchanged) |
| `/.well-known/health` | `HTTP_PORT` (8000) | `{"name","status"}` — readiness, **redacted** |
| _(the detailed map)_ | — | **not exposed anywhere** — only `Container.Health()` internally |

The full per-dependency map (hosts, ports, DB/keyspace/bucket names, pool stats, ES username) is still computed by `Container.Health(ctx)` but is no longer served on any endpoint.

## Proposed change

Serve the full detailed map at `/health` on the **metrics server** (`METRICS_PORT`, default `2121`), next to `/metrics` and `/debug/pprof`:

| Endpoint | Port | Body | Purpose |
|---|---|---|---|
| `/.well-known/alive` | `HTTP_PORT` | `{"status":"UP"}` | liveness |
| `/.well-known/health` | `HTTP_PORT` | `{"name","status"}` | readiness (redacted) |
| `/health` | `METRICS_PORT` | full detailed map | ops-only, behind the metrics/pprof network boundary |

### Feasibility

Low-risk wiring — `metricServer.Run` already receives `*container.Container` (`pkg/gofr/metrics_server.go`), so a handler on the metrics router can call `c.Health(ctx)` directly. `metrics.GetHandler` needs the container passed alongside the metrics manager. `METRICS_PORT=0` disables the detailed endpoint entirely.

### Caveats

- The metrics server binds `:port` (all interfaces), same as today's `/debug/pprof` — so "private" means *not exposed through ingress*, identical to pprof's current guarantee. It is **not** loopback-bound (that would break in-cluster Prometheus scraping). The boundary is network policy — worth stating in docs.
- Adding `/health` to the metrics port is a new documented endpoint; call it out in release notes.

## Related

[...]

Issue #2550, opened 2025-11-19
Exempt alive endpoint from all authentication methods


aryanmehrotra:
Right now when we add a gofr middleware we cannot access the alive or health endpoint.
but we should be exposing the alive endpoint as it is required for health-check, we can skip exposing the health endpoint as it contains db and other info.

sammy200-ui:
@coolwednesday  @aryanmehrotra  Hey! I've started looking into this issue and would like to work on it. 
Could you please assign it to me?

sammy200-ui:
@aryanmehrotra @coolwednesday! I've investigated this issue and wanted to discuss the approach before submitting a PR.


Currently, the `AuthMiddleware` in `pkg/gofr/http/middleware/auth.go` uses the `isWellKnown()` helper which exempts **all** paths under `/.well-known` from authentication. This means both:
- `/.well-known/alive`  (should be public for health checks)
- `/.well-known/health` (currently public, but contains sensitive DB/service info)


The `isWellKnown()` function in `validate.go` checks:
```go
func isWellKnown(path string) bool {
    return strings.HasPrefix(path, "/.well-known")
}
```
Proposed Fix
Modify `AuthMiddleware` to only exempt /.well-known/alive and keep /.well-known/health protected:

Changes:

`auth.go` Replace the generic check with a specific path check:
```
if r.URL.Path == service.AlivePath {
    handler.ServeHTTP(w, r)
    return
}
```
`auth_test.go` Update tests to verify:

/.well-known/alive is accessible without auth 
/.well-known/health requires authentication

coolwednesday:
@aryanmehrotra , based on @sammy200-ui , the .well-known/alive is already exposed. @sammy200-ui & @aryanmehrotra , Can you both run and check once. The scope of and primary target of the issue will then change accordingly.

coolwednesday:
@sammy200-ui , Please check the intended behaviour and for better communication you can start a discussion in our discord community also .

sammy200-ui:
@coolwednesday The current behavior is already correct:
 /.well-known/alive is publicly accessible (no auth required)
 /.well-known/health requires authentication (protected)

This matches exactly what the issue requests: "we should be exposing the alive endpoint as it is required for health-check, we can skip exposing the health endpoint as it contains db and other info."

Could you clarify if there's a specific scenario where both endpoints are blocked? Or should this issue be closed as working-as-intended?

coolwednesday:
[...]

PR #3807, opened 2026-07-30
fix(health): redact unauthenticated /.well-known/health (#3802)

### LINKED ISSUE #3802: Unauthenticated /.well-known/health leaks DB/infra details — contrary to intent in #2550
## Summary

`/.well-known/health` is served **unauthenticated** and returns detailed infrastructure information (DB hosts/ports, database/keyspace/bucket names, connection-pool stats, the Elasticsearch username, raw error strings). Because `AuthMiddleware` unconditionally bypasses any `/.well-known/*` path, this data is exposed regardless of which auth (Basic / API-Key / OAuth) the application configures.

This is contrary to the intent stated in #2550, where the plan was to expose `/alive` but **keep `/health` gated "as it contains db and other info."** The exemption was implemented as a broad `/.well-known` prefix match, which over-exempts and re-introduces the disclosure.

## Where it happens

**Auth is bypassed for the whole `/.well-known` prefix** — `pkg/gofr/http/middleware/auth.go:49`:

```go
if isWellKnown(r.URL.Path) {   // strings.HasPrefix(path, "/.well-known")  — validate.go:5
    handler.ServeHTTP(w, r)
    return
}
```

Basic-Auth, API-Key, and OAuth all funnel through this single `AuthMiddleware`, so all three bypass auth for `/.well-known/health`.

**The response body is the full detail map** — `pkg/gofr/container/health.go:8` (`Container.Health`) aggregates each datasource's `HealthCheck()`. Examples of what lands in `details`:

- **SQL** (`datasource/sql/health.go:31,52`): `details.host = "hostname:port/dbname"` + full connection-pool stats.
- **Redis** (`datasource/redis/health.go:24,45`): `details.host = "hostname:port"` + raw `INFO Stats`; raw error strings on failure.
- **Mongo / Cassandra / Couchbase / ClickHouse**: host/URI + database / keyspace / bucket names.
- **Elasticsearch** (`datasource/elasticsearch/elasticsearch.go:278`): addresses **and the `username`**.

There is currently **no** config key, env var, flag, or middleware hook to redact, restrict, or disable this output.

## Impact

Anyone who can reach the app port (default `HTTP_PORT=8000`) — with no credentials — can enumerate the backend inventory, locate each datastore (host:port), read connection statistics, and in the Elasticsearch case obtain a username. This is useful reconnaissance (attack surface, CVE targeting, credential/port probing) and it is exposed even on apps that have otherwise enabled authentication.

## Proposed fix — align health with the metrics/pprof isolation model

GoFr already treats sensitive introspection this way: `/debug/pprof/*` (arguably more sensitive than health) is served on the **metrics server** (`METRICS_PORT`, default `2121`) — a separate `http.Server` with a bare router — and protected by **network isolation**, not app auth (`pkg/gofr/metrics/handler.go:13`). The detailed health map is the same category of data and belongs behind the same boundary.

Concretely, a **three-endpoint contract**:

|---|---|---|---|
| `/.well-known/alive` | `HTTP_PORT` (8000) | `{"status":"UP"}` | liveness (unchanged) |
| `/.well-known/health` | `HTTP_PORT` (8000) | `{"status":"UP"｜"DEGRADED"}` | **readiness — aggregate only, no inventory** |
| `/health` | `METRICS_PORT` (2121) | today's full detailed map | ops-only, behind the metrics/pprof network boundary |

Key points:

1. **Public `/.well-known/health` is reduced to a single aggregate bit** — `UP` if all backends are healthy, `DEGRADED` if any is down. It does **not** reveal which backends exist, their hosts/ports, stats, usernames, error strings, or the app `name`/`version`. Readiness probes keep working; reconnaissance value drops to ~zero. (It still differs from `/alive`: liveness answers "is the process up", readiness folds in dependency health.)
2. **The full detailed map moves to the metrics server** (next to `/metrics` and `/debug/pprof`), reusing existing `METRICS_PORT` machinery — including `METRICS_PORT=0` to disable it entirely.
3. This keeps `/.well-known/health` behavior *compatible in shape* for readiness consumers (still a `status`-bearing endpoint on 8000) while removing the disclosure.

### Feasibility

Low-risk wiring: `metricServer.Run` already receives `*container.Container` (`pkg/gofr/metrics_server.go`), so a health handler on the metrics router can call `c.Health(ctx)` directly — `metrics.GetHandler` just needs the container passed alongside the metrics manager. The public redaction is a small change in `container/health.go` (emit aggregate `status` without the per-backend `details`).

### Alternatives considered

- **Detail onto metrics port, drop health from 8000 entirely.** Cleaner, but breaks readiness probes that rely on a `status` endpoint on the app port. Rejected in favor of keeping an aggregate readiness signal on 8000.
- **Config toggle only** (e.g. `HEALTH_EXPOSE_DETAILS=false`, default-secure). Simplest, no route changes — but keeps everything on the public port and doesn't follow the metrics/pprof pattern. Reasonable fallback if maintainers prefer not to move routes.

### Caveats to weigh

- The metrics server binds `:port` (all interfaces), same as today — so "private" means *not exposed through ingress*, identical to pprof's current guarantee. It is **not** loopback-bound. (Loopback would break in-cluster Prometheus scraping, so it isn't appropriate for the metrics server.) The boundary is network policy, not a bind restriction — worth stating in docs.
- Moving/redacting `/.well-known/health` is a **behavior change** to a documented endpoint; the three-endpoint contract keeps a working readiness endpoint on 8000 to minimize breakage, but it should be called out in release notes.

## Related

- #2550 — "Exempt alive endpoint from all authentication methods" (the change that introduced the broad `/.well-known` bypass; its stated intent was to keep `/health` gated).

akshat-kumar-singhal:
**Description:**

Fixes #3802 — `GET /.well-known/health` is served **unauthenticated** yet returned the full `Container.Health()` map: datastore hosts/ports, DB/keyspace/bucket names, connection-pool stats, the Elasticsearch username, and raw error strings. #2550's broad `/.well-known` auth-bypass un-gated the route while the handler still served the detailed ops-oriented map, so anyone who can reach the app port could enumerate the backend inventory with no credentials.

**Change** — redact the public endpoint to `{name, status}` only. An unexported `aggregateStatus` in `pkg/gofr/health.go` runs the same aggregation but discards `details`. `Container.Health()` (full map) is unchanged for in-process ops use. **This PR adds no exported API** — see the note under Additional Information.

Verified on the wire: `200 {"data":{"name":"...","status":"UP"}}` — no host, details, stats, username, or version.

The auth-bypass in `auth.go` is intentionally left unchanged — readiness must stay reachable unauthenticated; redaction, not re-gating, is the fix.

The **body shape of `/.well-known/health` changes** from the detailed per-dependency map to `{name, status}`. The HTTP status code is unchanged (200), so existing Kubernetes readiness probes keep working; only consumers parsing the body for dependency details are affected. Called out in the docs page as a release note.

**Additional Information:**

- No new dependencies.
- **No exported API added (`dfa92b1d`, per review).** An earlier revision introduced `Container.HealthStatus`; it had a single caller and `Context` embeds `*container.Container`, so it would have been promoted onto every handler's `ctx` as supported API nothing in the tree needs. It is now the unexported `aggregateStatus` in `pkg/gofr/health.go`. If #3858 needs public surface for a readiness closure, it should justify it there.
- **Scope narrowed per review:** the optional `SetHealthCheck` readiness closure that this PR originally bundled has been split out — it is a new public API and a separate design discussion, and should not gate an urgent security fix. Tracked in #3857, proposed in #3858 (stacked on this branch).
- #3806 tracks the larger second half of the issue's design: serving the full detailed map at `/health` on `METRICS_PORT`, behind the same network boundary as `/metrics` and `/debug/pprof`.

**Tests:**

- `pkg/gofr/handler_test.go`: asserts the public body's key set is exactly `{name, status}`, so any future field added to the response fails the test rather than leaking.
- `pkg/gofr/handler_test.go`: table-driven `TestHandler_aggregateStatus` covers `UP` and `DEGRADED` through a live and a 500-returning `httptest` service, asserting only the aggregate string survives.
- `examples/using-http-service/main_test.go`, `examples/using-ai/main_test.go`: updated to the redacted body; the using-ai case additionally asserts `NotContains` for `"llm"`, `"provider":"groq"`, and the API key.
- Docs updated: `docs/advanced-guide/monitoring-service-health/page.md`.

**Checklist:**


pkg/gofr/health.go
```diff
@@ -0,0 +1,92 @@
...
+// errHealthNotReady carries a caller-chosen HTTP status code for a not-ready readiness result so
+// the responder emits that code instead of the default 200. Its message is the status string, so
+// the body still conveys the reported status without leaking any dependency details.
+type errHealthNotReady struct {
+	status string
+	code   int
+}
+
+func (e errHealthNotReady) Error() string   { return e.status }
+func (e errHealthNotReady) StatusCode() int { return e.code }
```
Umang01-hash: `errHealthNotReady` implements `StatusCode()` but not `logging.LogLevelResponder`, so `handler.logError` defaults it to ERROR. A closure returning 503 (normal during startup/rolling deploys) then logs ERROR on every readiness probe. Suggest adding `func (errHealthNotReady) LogLevel() logging.Level { return logging.WARN }`, matching how the other expected errors in `http/errors.go` set their level.

Umang01-hash (changes requested):
Reviewed at head `2fb3a898`, running the tests myself. The redaction is the right fix and the exported API is non-breaking — but one blocker.

**Blocker — two existing example integration tests still assert the old detailed body and fail at head:**
- `examples/using-http-service/main_test.go:56` pins the full map (`details`/`host`/`version`).
- `examples/using-ai/main_test.go:104-106` asserts the body contains `"llm"` and `"provider":"groq"`.

Both `--- FAIL` when run (`go test ./examples/using-http-service/ ./examples/using-ai/ -short`), and the same FAILs show in this PR's own CI log. Since the endpoint is now redacted, these need updating to assert `{name,status}` in this PR.

**Why CI is green anyway (pre-existing, worth a separate issue):** the Example job runs `go test … ; grep … > profile.cov ; go tool cover …` with no `set -e`, so the step's exit code is the last command's (0) and the `go test` failure is swallowed (`Command completed after 1 attempt(s)`).

**Minor:** `errHealthNotReady` doesn't implement `logging.LogLevelResponder`, so a not-ready readiness (503) logs at ERROR on every K8s probe. Convention lowers expected errors (timeout→INFO, rate-limit→WARN); suggest `LogLevel() → WARN`.

Everything else checks out: new unit tests pass (`-race` clean), gofmt/vet/build clean, redacted wire-shape verified, docs accurate, no exported-API break.

akshat-kumar-singhal:
Thanks for the thorough review — all three points addressed in fb19e91b.

**Blocker — example tests updated to the redacted body:**
- `examples/using-http-service/main_test.go`: health case now asserts `{"data":{"name":"using-http-service","status":"DEGRADED"}}`.
- `examples/using-ai/main_test.go`: rewrote the case to assert the redacted shape (`"status":"UP"`) and, reinforcing the fix's intent, `NotContains` for `"llm"`, `"provider":"groq"`, and the API key.

Both now pass; `Test_main` (http-service) and `TestIntegration_UsingAI` are green locally.

**Minor — not-ready readiness now logs at WARN:**
`errHealthNotReady` implements `logging.LogLevelResponder` returning `logging.WARN`, matching the convention in `http/errors.go`, with a `var _ logging.LogLevelResponder = errHealthNotReady{}` assertion. An expected 503 during startup/rolling deploys no longer logs ERROR on every K8s probe.

**CI-swallows-failure:** filed as a separate issue — #3812. It's a pre-existing hole (missing `set -e` in the `nick-fields/retry` `command:` block, so the block's exit code is `go tool cover`'s 0 rather than the failing `go test`'s), affecting both the Example and PKG jobs. Kept out of this PR since enabling `set -eo pipefail` will likely surface other currently-hidden failures that deserve their own triage.

pkg/gofr/health.go
akshat-kumar-singhal: Done in fb19e91b — added `func (errHealthNotReady) LogLevel() logging.Level { return logging.WARN }` plus a `var _ logging.LogLevelResponder = errHealthNotReady{}` assertion, matching `http/errors.go`. A 503 during startup/rolling deploys now logs at WARN instead of ERROR on every probe.

Umang01-hash (dismissed):
Re-reviewed at `42aeecbb`, ran the checks myself. Both prior blockers are fixed and verified:

- **Example tests updated for the redacted body** — `Test_main` (using-http-service) and `TestIntegration_UsingAI` both pass with `-short`; the using-ai case now asserts `NotContains("llm"/"provider")`, which is a stronger security check.
- **Not-ready readiness now logs at WARN** — `errHealthNotReady.LogLevel()` returns `logging.WARN` with a compile-time `LogLevelResponder` assertion.

Build/vet/gofmt clean, `pkg/gofr` + `container` health tests green under `-race`, no exported-API break. (The lone `TestHTTPHandlerURLError` failure I saw locally reproduces on `development` too — a network-dependent test, unrelated to this PR.)

LGTM.

pkg/gofr/health.go
```diff
@@ -0,0 +1,101 @@
...
+
+// LogLevel keeps a not-ready readiness at WARN rather than ERROR. A 503 is expected during startup
+// and rolling deploys — Kubernetes probes it repeatedly — so it should not flood logs as an error.
+func (errHealthNotReady) LogLevel() logging.Level { return logging.WARN }
+
+var _ logging.LogLevelResponder = errHealthNotReady{}
+
+// healthHandler serves the public, unauthenticated /.well-known/health endpoint. It reports only
+// the application name and aggregate status — no hosts, ports, credentials, or connection stats.
+// The full detailed health map remains available internally via Container.Health for ops tooling.
```
Umang01-hash: This says the detailed map "remains available internally via Container.Health", but no HTTP route serves it after this PR. The docs correctly note it's *being moved to METRICS_PORT* as a follow-up — worth aligning this comment so it doesn't imply HTTP access that no longer exists.

pkg/gofr/handler_test.go
```diff
@@ -259,10 +259,95 @@ func TestHandler_healthHandler(t *testing.T) {
...
+	// The public endpoint must expose only {name, status} and never per-dependency details
+	// (hosts, ports, credentials, connection stats). See issue #3802.
+	resp, ok := h.(healthResponse)
+	require.True(t, ok, "expected redacted healthResponse, got %T", h)
+	assert.NotEmpty(t, resp.Status)
+
+	body, err := json.Marshal(resp)
+	require.NoError(t, err)
+	assert.NotContains(t, string(body), "details")
+	assert.NotContains(t, string(body), "host")
```
Umang01-hash: Good to have a no-leak guard. Consider asserting the exact keys are `{name,status}` (or also excluding `username`/`port`) so a future re-leak of e.g. the ES username is caught, not just `details`/`host`.

Umang01-hash (commented):
Verified against the code, not the PR body — the leak is real: `AuthMiddleware` bypasses the whole `/.well-known` prefix (auth.go:49) and the old handler returned the full `c.Health()` map (SQL `host:port/db`, Redis `INFO`, ES `username`…) unauthenticated. Redacting to `{name,status}` is the right urgent fix — structurally complete + has a no-leak test, and gofmt/vet/build are clean with the health tests passing.

One ask: **split out `SetHealthCheck`.** The redaction closes #3802 on its own; `SetHealthCheck` + `HealthCheckFunc` is a separate new public API (opt-in readiness gating that didn't exist before), and bundling it gates an urgent security fix on an API-design discussion. I'd land the redaction alone and propose the readiness closure as its own PR.

Minor: the response body is a breaking wire-contract change (detailed map → `{name,status}`) — worth a release-note callout. Two nits inline.

akshat-kumar-singhal:
Thanks — all four points addressed in `fa3baccb`.

**Main ask — `SetHealthCheck` split out.** Agreed on the reasoning: it is a new public API and its own design discussion, and it should not gate the #3802 fix. This PR is now redaction-only. The closure is filed as #3857 and proposed in #3858, stacked on this branch — I will rebase it onto `development` once this lands.

The removal also reverted the hooks the closure needed, so `gofr.go` is back to a zero diff and `handler.go` only loses the old detailed handler. What remains is `Container.HealthStatus` plus a `{name, status}` body — 8 files, +118/-98.

**Release-note callout.** Added as a blockquote in `docs/advanced-guide/monitoring-service-health/page.md`, and the PR body now has it under Breaking Changes: the body shape changes from the detailed map to `{name, status}`, the status code stays `200` so existing probes keep working, and only consumers parsing the body for dependency details are affected.

**`health.go` comment.** Reworded — it no longer implies HTTP access that does not exist:

```go
// No HTTP route serves the full detailed map after this change; Container.Health still computes it
// for in-process ops tooling, and #3806 tracks exposing it on the metrics server (METRICS_PORT),
// behind the same network boundary as /metrics and /debug/pprof.
```

**Test nit.** Good call — a blocklist only catches the fields we thought of. Now pins the exact key set, so any future field added to the public body fails the test rather than leaking:

```go
assert.Equal(t, []string{"name", "status"}, slices.Sorted(maps.Keys(got)))
```

**Verified locally:** `go build ./...` and `gofmt` clean; `golangci-lint` reports nothing on the changed files; `pkg/gofr` and `pkg/gofr/container` health tests pass under `-race`; `Test_main` (using-http-service) and `TestIntegration_UsingAI` pass with `-short`. `TestHTTPHandlerURLError` still fails locally — that is the network-dependent test you noted reproduces on `development`; I have not re-verified it against `development` myself this time.

pkg/gofr/health.go
```diff
@@ -0,0 +1,17 @@
...
+// healthResponse is the minimal, non-sensitive body returned by /.well-known/health. It carries
+// only the application name and the aggregate status — never per-dependency inventory.
+type healthResponse struct {
+	Name   string `json:"name"`
+	Status string `json:"status"`
+}
+
+// healthHandler serves the public, unauthenticated /.well-known/health endpoint. It reports only
+// the application name and aggregate status — no hosts, ports, credentials, or connection stats.
+// No HTTP route serves the full detailed map after this change; Container.Health still computes it
```
Umang01-hash, line 12: Non-blocking: the detailed map isn't served on any HTTP route until #3806 lands it on METRICS_PORT — so ops users lose HTTP detailed-health in the interim (still available in-process via `Container.Health()`). Sensible security default and consistent with the /metrics + pprof boundary; just worth making sure #3806 is queued so the gap isn't open-ended.

pkg/gofr/container/health.go
```diff
@@ -70,6 +70,25 @@ func (c *Container) Health(ctx context.Context) any {
 	return healthMap
 }
 
+// HealthStatus runs the aggregate health check and returns only the overall status
```
Umang01-hash: Nit: godoc says returns `("UP" or "DEGRADED")` but the fallback also returns `"DOWN"` — worth reflecting in the comment.

Umang01-hash (dismissed):
Reviewed end-to-end (built, ran the using-http-service example live — `/.well-known/health` serves 200 with the redacted `{name,status}` body). The fix is warranted: the old handler returned the full dependency map (hosts + raw errors) on an unauthenticated route. Redaction is complete and the handler test pins the body to exactly `[name,status]`, so any future leak fails the test.

Checked the one cross-cutting risk — gofr's own `httpService.HealthCheck`/circuit breaker hits `/.well-known/alive` and decides health from the HTTP status code (200), never the body, so this doesn't affect service-to-service health checks. `HealthStatus` is additive on `*Container` (no breaking change), and the example + docs are updated to the new shape.

LGTM. Two non-blocking notes below.

aryanmehrotra (dismissed):
Reviewed end to end at `fa3baccb`. CI is 14/14. Reproduced the disclosure and confirmed it is closed. Approving, with two wording points and a merge-order note.

## The leak is real and auth-independent

Built a real app with basic auth enabled and two HTTP dependencies, one reachable and one deliberately dead, then requested the endpoint with **no credentials**.

`development`:

```json
{"data":{
  "ledger-prod":{"status":"DOWN","details":{"error":"Get \"http://127.0.0.1:52548/.well-known/alive\": dial tcp 127.0.0.1:52548: connect: connection refused"}},
  "name":"gofr-app",
  "payments-prod":{"status":"UP","details":{"host":"127.0.0.1:52547"}},
  "status":"DEGRADED",
  "version":"dev"}}
```

That is dependency names, hosts, ports, a raw dial error and the framework version, to an unauthenticated caller — while `/api/ping` on the same app correctly answers `401 missing auth header`. So the `/.well-known` prefix exemption is doing exactly what #3802 describes, and no auth configuration changes it.

At this head:

```json
{"data":{"name":"gofr-app","status":"DEGRADED"}}
```

`/.well-known/alive` and the `401` on `/api/ping` are unchanged.

## The supporting claims hold

**"No HTTP route serves the detailed map after this change."** Checked rather than taken — `Container.Health` has exactly one caller on this branch, `HealthStatus`. Nothing else reaches it, and the swagger `/.well-known/{name}` catch-all cannot shadow the health route because health registers first.

**`HealthStatus` cannot silently under-report.** `appHealth` unconditionally sets `healthMap["status"]` to `UP` or `DEGRADED`, so neither the type assertion nor the key lookup falls through in practice — and the `statusDown` default if one ever did is fail-closed, which is the right direction for this function.

**The test is the assertion I would have asked for.** Pinning the exact key set —

```go
assert.Equal(t, []string{"name", "status"}, slices.Sorted(maps.Keys(got)))
```

— rather than blocklisting known-bad keys means a field added to the public body later fails this test instead of leaking, which is the difference between a regression test and a snapshot of today's leak. The comment saying so is worth keeping.

## Two wording points

**The docs now call this "Readiness", and it cannot report not-ready.** The section says it "reports whether the service is ready to receive traffic", but a `DEGRADED` aggregate still answers `200` — verified on both heads. I checked #3858 in case the follow-up closes this: `readiness()` returns `http.StatusOK` when no closure is set, so `503` only ever comes from a user-supplied closure, and a merely-down datasource keeps returning `200` even after that lands. So the readiness framing arrives a PR early. Either soften it here — it reports an aggregate status, which a probe may act on — or say plainly that making it actionable is what the closure in #3858 is for.

**The "Changed response shape" note omits `version`.** It lists hosts, ports, database/keyspace/bucket names, pool stats, usernames and raw error strings, but the old body also carried `"version":"dev"` and the new one does not. Worth naming, on both counts: a framework version string is precisely what someone enumerating known CVEs wants, so dropping it is part of the fix rather than incidental; and it is also the field most likely to be on somebody's dashboard, so it belongs in the list of things that will stop appearing.

## Merge order

#3858 is stacked on this branch — it contains these commits, and `merge-tree` reports no conflict — so this merges first and the follow-up rebases cleanly. Worth saying because the last time two of your branches touched the same functions (#3814 and #3820) they were independent and collided; stacking this one is the reason there is nothing to untangle here.

aryanmehrotra:
Two notes on `Container.HealthStatus`, neither blocking — approval stands.

**1. It doesn't need to be exported for anything in this PR.** One production caller, in package `gofr`:

```go
return healthResponse{Name: c.GetAppName(), Status: c.HealthStatus(c)}, nil
```

It's public only because `Container` lives in `gofr/container` and the handler lives in `gofr`. The same extraction inside `health.go` adds zero public surface:

```go
status := "DOWN"
if m, ok := c.Health(c).(map[string]any); ok {
	if s, ok := m["status"].(string); ok {
		status = s
	}
}
```

I can see the cohesion argument for keeping it next to `appHealth`, which is what writes `healthMap["status"]` — a key rename then moves both together. If that's the reason, put it in the comment, because as it stands it reads like a supported entry point that nothing in the tree needs.

**2. The doc comment and the code disagree on the return values.**

```go
// ... returns only the overall status ("UP" or "DEGRADED") ...
```

Two paths `return statusDown` — `"DOWN"`. `appHealth` only ever writes `UP` or `DEGRADED` (`container/health.go:126-130`), so `"DOWN"` is unreachable and fail-closed, which is the right instinct. But this is newly exported: either list the third value, or say it is an unreachable default and why it is there.

pkg/gofr/container/health.go
akshat-kumar-singhal: Done in `549852fd` — the godoc now names all three: `UP` when every dependency is healthy, `DEGRADED` when one or more are down, and `DOWN` as the fail-closed fallback if the aggregate status is ever missing or not a string.

pkg/gofr/health.go
akshat-kumar-singhal: Addressed — the comment on `healthHandler` now reads "No HTTP route serves the full detailed map after this change; `Container.Health` still computes it for in-process ops tooling," with the METRICS_PORT move pointed at #3806. No implication of HTTP access left.

pkg/gofr/handler_test.go
akshat-kumar-singhal: Done — `handler_test.go` now pins the exact key set rather than blocklisting known-bad keys:

```go
assert.Equal(t, []string{"name", "status"}, slices.Sorted(maps.Keys(got)))
```

So a future field added to the public body (ES `username`, `port`, anything else) fails this test instead of leaking.

akshat-kumar-singhal:
@aryanmehrotra thanks — both wording points addressed in `549852fd`.

**Readiness framing.** You are right that it arrives a PR early, so I have pulled it back rather than softened it: the section heading is reverted to its original `Health-Check`, and the body now describes an *aggregate status* instead of readiness-to-receive-traffic, with the 200-in-both-cases behaviour stated outright:

> Note that the endpoint answers `200` in both cases: a `DEGRADED` aggregate does **not** produce a non-2xx response, so a probe wired directly to the status code will always see the service as ready. A caller that wants to act on degradation must read the `status` field itself.

Making it actionable stays with the closure in #3858, and the readiness naming can land with it.

**`version`.** Added to the changed-response-shape note, on both the counts you named — that a framework version string is what someone enumerating known CVEs wants, so dropping it is part of the fix and not incidental, and that it is the field most likely to be on an existing dashboard.

Also picked up @Umang01-hash's godoc nit in the same commit: `HealthStatus` now documents the `DOWN` fail-closed fallback alongside `UP`/`DEGRADED`.

No behaviour change in this commit — docs and comments only. Merge order is as you described: this first, #3858 rebases on top.

aryanmehrotra:
Re-reviewed at `549852fd` — both wording points are addressed, the godoc now covers `DOWN`, and the diff since `fa3baccb` really is docs and comments only. Re-verified behaviour rather than assuming it: unauthenticated, one live dependency and one dead, `development` still returns hosts, ports, the raw dial error and `version:"dev"`, while this head returns `{"name":"gofr-app","status":"DEGRADED"}`. `/.well-known/alive` and the `401` on an authenticated route are unchanged. CI is 14/14 and, unlike some older runs, this one is after #3814 so the Example job actually gates.

The redaction is right and I want it to land. What I want to raise before it does is the **placement** of `Container.HealthStatus`, because it becomes public API the moment this ships, and there is a shape that makes it unnecessary.

## The thing I missed the first time

`Context` embeds `*container.Container`, so every exported `Container` method is promoted onto it. `HealthStatus` is therefore callable from any handler today — I ran it rather than reasoning about it:

```go
a.GET("/mine", func(ctx *gofr.Context) (any, error) {
    return map[string]any{"via_ctx": ctx.HealthStatus(ctx)}, nil
})
// GET /mine -> 200 {"data":{"via_ctx":"DEGRADED"}}
```

I do not think that is wanted, for two reasons. It has no use case — the built-in endpoint already serves exactly that string, unauthenticated, so there is nothing to rebuild. And it is a footgun: `Health` does no caching at all, so it fans out to every datasource and every registered service on every call. Measured on this head, `20` requests to `/.well-known/health` produced `40` outbound probes with two dependencies registered. In a handler that is one full dependency sweep per request, serially, in the hot path.

## Why it is exported at all

As far as I can tell the only justification is #3858: `HealthCheckFunc` receives a `*container.Container`, so a readiness closure needs some accessor to fold the framework's own verdict into its answer. That is a real need — but it points at the closure's shape rather than at the export.

## The shape I would rather see: fold checks in, don't replace the aggregate

Instead of one closure that *replaces* the aggregate, let an app register additional checks that the framework folds into the aggregate it already computes:

```go
a.AddHealthCheck("warm-cache", func(_ context.Context) datasource.Health {
    if cacheWarm {
        return datasource.Health{Status: "UP"}
    }
    return datasource.Health{Status: "DOWN", Details: map[string]any{"reason": "still loading"}}
})
```

I prototyped it on top of this branch. With every framework dependency healthy and only the user's check flipping:

```
user check DOWN -> 200 {"name":"gofr-app","status":"DEGRADED"}
user check UP   -> 200 {"name":"gofr-app","status":"UP"}
```

`"reason":"still loading"` never reaches the wire, so the redaction this PR establishes still holds for app-supplied checks.

Total cost is about 14 lines: one field beside the existing `Services`, a six-line loop in `Health` that is a copy of the `c.Services` loop directly above it, and the registrar. `go build ./pkg/...` is clean; one test in `container/health_test.go` moves packages.

Why I think it is the better contract:

|  | `SetHealthCheck` (#3858) | `AddHealthCheck` |
|---|---|---|
| app must re-fetch the framework aggregate | yes — forget it and you silently mask every datasource failure | no, automatic |
| composable | one closure; two teams collide | N named checks |
| requires `HealthStatus` exported | yes, this is its only reason to exist | no |
| fits existing vocabulary | `Set…` is unique in the API | `AddHTTPService`, `AddLLM`, `AddCronJob`, `AddStaticFiles` |

The third row is the point: the export exists so a replacing closure can read what it replaced. Fold the checks in and nothing needs to read the aggregate, so `HealthStatus` can go back to being unexported — or just live inside `healthHandler`, which is what I suggested in my first pass.

On the name, `AddHealthCheck` reads best to me because every datasource in the tree already implements `HealthCheck()` and already returns `datasource.Health{Status, Details}` — no new vocabulary, and the signature reads as "one more thing to health-check". `AddHealthSource` introduces a word we do not use anywhere; `AddDependency` I would avoid, since we already call datasources dependencies and it would read like DI registration.

## Registration must stay on App, not Container

One constraint worth being explicit about, because it is easy to get wrong and my own first attempt did: the registry must not be reachable from a handler either. If it goes on `Container` as an exported field or method, embedding promotes it onto `Context`, and a handler can then mutate the map at request time — which is a data race on top of being meaningless.

What works is keeping the map on `App` and having `container` expose a package-level **function** rather than a method, since functions are not method-promoted:

```go
// package container
func SetHealthChecks(c *Container, checks map[string]func(context.Context) datasource.Health)

// package gofr
func (a *App) AddHealthCheck(name string, check func(context.Context) datasource.Health)
```

Verified that all three doors are shut, by compiling a handler that tries each one:

```
ctx.AddHealthCheck undefined (type *gofr.Context has no field or method AddHealthCheck)
ctx.HealthStatus   undefined (type *gofr.Context has no field or method HealthStatus)
ctx.healthChecks   undefined (cannot refer to unexported field healthChecks)
```

Registration and aggregate access both become startup concerns, enforced by the compiler instead of by a doc comment.

## What I am asking

None of this argues against the redaction, and it is not a security problem. But I do not think `Container.HealthStatus` should ship exported, so I would like that changed here rather than documented.

The case against it, in order:

- **It buys nothing today.** One caller, `healthHandler`, in the same module. Unexported works.
- **Exporting it hands every handler a footgun with no use case.** Embedding promotes it onto `Context`, the endpoint already serves that exact string unauthenticated so there is nothing to rebuild, and `Health` does no caching — so any per-request use is a full dependency sweep in the hot path.
- **Its only justification is a follow-up whose shape is itself in question.** If `AddHealthCheck` is the right contract, nothing ever needs to read the aggregate back, and the export was never needed.
- **The asymmetry is the deciding factor.** Unexporting it now costs one line. Unexporting it after a release carries it is a breaking change we would have to defend for a whole major.

A godoc note is not a substitute for any of that — it documents the footgun instead of removing it, and it still commits us to the symbol.

So concretely: keep the extraction inside `healthHandler` in this PR, and let #3858 introduce whatever public surface it genuinely needs, with the justification attached to it.

And the open question I would value your read on: is `AddHealthCheck` the shape you would want instead of `SetHealthCheck`? You have been closest to this code, so if you think folding checks in does not cover a case the replacing closure does, I would rather hear it now than have us design it twice.

One gap that `AddHealthCheck` does **not** close, worth separating: `SetHealthCheck` can answer `503`, and folding checks in only moves the aggregate to `DEGRADED`, which still answers `200`. Those feel like two different decisions to me — *what* is healthy (composable, per-dependency) versus *how* to answer a probe (one app-wide policy) — and #3858 currently merges them into one closure, which is exactly why it needs the aggregate read back.

Separately and not related to this PR: the missing cache is worth its own issue. Unauthenticated, uncached, and fanning out to every dependency means one cheap request per dependency probe, and it scales with how many you register. It is pre-existing on `development`, so it is not this PR's doing, but #3858 is already in that handler and a short TTL would close it.

akshat-kumar-singhal:
@aryanmehrotra Done in `dfa92b1d` — `Container.HealthStatus` is gone. `healthHandler` now calls an unexported `aggregateStatus(*Context)` in `pkg/gofr/health.go`: same `Health(ctx).(map[string]any)["status"]` extraction, same `DOWN` fail-closed default. The exported-API diff of this PR is now zero.

Coverage moved with it: `TestContainer_HealthStatus` is removed — its UP/DEGRADED aggregation was already asserted by `TestContainer_Health`'s full-JSON comparison — and replaced by table-driven `TestHandler_aggregateStatus` in `pkg/gofr`, driving both cases through a live and a 500-returning `httptest` service. The `DOWN` branch stays untested because `appHealth` cannot produce it; the godoc now says it is a defensive default rather than listing it as a returnable value. On the cohesion point from your first pass: `appHealth` now carries a comment naming `healthHandler` as the reader of `"status"`, so a rename surfaces both sides.

## On the asymmetry — agreed, and it is the argument that decides it

One caller, so unexporting costs six lines now and a major-version deprecation later. That stands on its own.

## On the footgun framing — I think it overstates what this PR introduces

`Container.Health` is already exported and already promoted onto `Context`, so `ctx.Health(ctx)` is callable from any handler on `development` today and returns the **full detailed map**, hosts and raw errors included. Your `20 → 40` measurement is of `Health`, which this PR does not touch. `HealthStatus` was a strictly *less* leaky sibling of a hazard that already ships, not a new class of one.

Worth separating, because the current framing reads as though the redaction PR opened the hole — and because if promoted-uncached-fanout is the concern, `Health` is the symbol to point at.

## On `AddHealthCheck`: yes, that is the shape I would want

Fail-safety is the reason more than composability — with `Set`, forgetting to fold in the framework aggregate silently masks every datasource failure, and that particular mistake stays invisible until an outage. The vocabulary argument holds too; `Set…` would be the only one of its kind next to `AddHTTPService`/`AddLLM`/`AddCronJob`.

Your split of *what is healthy* from *how to answer a probe* is the answer to the 503 gap, and I would push it one step further: a policy hook takes the aggregate as an **argument** — `func(status string) int` — rather than fetching it. So even the 503 path never reads the aggregate back, and the export was not load-bearing under either shape. That is the strongest version of your case for unexporting, and it is why I am comfortable doing it here rather than waiting on #3858.

## One gap in the three-doors check

You compiled against `ctx.AddHealthCheck`, `ctx.HealthStatus`, `ctx.healthChecks`. The embedded field itself is still a door: `Container` is an exported field name, so from a handler

```go
container.SetHealthChecks(ctx.Container, myMap)
```

compiles. Package-level functions are not method-promoted, but the receiver is reachable — `examples/using-http-service/main_test.go:97` already does `ctx.Container.Services = ...`. So the compiler does not enforce startup-only registration, and the data race you correctly flag against a `Container` method applies to the package-level function too. Fixable — snapshot the map at `Run`, or guard it — but worth designing for on #3858 rather than relying on the door being shut.

## On caching

Agreed it is pre-existing and not this PR's doing. I would file it now rather than fold it into #3858 — it is an unauthenticated, uncached, fan-out-per-request endpoint on the same route as this fix, and it deserves to be triaged on its own merits rather than as a rider on an API-design PR. Happy to open it.

aryanmehrotra (changes requested):
Re-reviewed at `dfa92b1d`. The unexport is done exactly as asked, and I verified it properly rather than by grepping the diff — full exported-surface enumeration of `pkg/gofr` and `pkg/gofr/container` via `go doc -all` at both revisions:

```
development : 4079 exported decls
dfa92b1d    : 4079 exported decls
diff        : identical
```

Nothing exported, nothing removed, nothing renamed. `ctx.HealthStatus` is now a compile error from a handler, `TestHandler_aggregateStatus` passes, CI is 14/14, and the endpoint still behaves: unauthenticated with one live and one dead dependency it returns `{"name":"gofr-app","status":"DEGRADED"}`, `/.well-known/alive` and the `401` on an authenticated route unchanged. The godoc that records *why* it stays unexported, and the `appHealth` comment naming its cross-package reader, are both better than what I asked for.

Requesting changes for one thing only: **the docs sweep is incomplete.** `monitoring-service-health/page.md` is thorough, but 30 pages under `docs/` reference this endpoint and only that one was updated. Nine of the others describe the response *content*, and that prose is now wrong.

Good news first, so it is clear what is not being asked: I checked every one of the 30 for a stale JSON body — per-dependency keys, `"details"`, `"version"` — and there are none. This is prose only.

## Wrong, and the two that matter most

These tell the reader to take action *because the endpoint leaks dependency details*. That is exactly what this PR removes, so they now point at a hole that no longer exists — stale security advice is worse than none.

`docs/advanced-guide/authentication/page.md:26`
```
If `/.well-known/health` exposes sensitive details about your datasources, scope its visibility at the network layer (private listener, mesh policy, or ingress rule) rather than expecting the auth middleware to gate it — the framework will not.
```
The premise is gone. The endpoint no longer exposes datasource details. The second half — that the auth middleware will not gate it — is still true and worth keeping, so this needs rewording rather than deleting.

`docs/guides/auth-in-kubernetes/page.md:93`
```
`/.well-known/health` is exempted by default but reveals dependency status. In production, re-enable auth on it (or restrict via NetworkPolicy to your monitoring namespace) so it is not enumerable from the public internet.
```
Same: "reveals dependency status" is no longer true, and the remediation it recommends is no longer necessary for that reason. The exemption itself is still worth documenting.

## Wrong, factual

`docs/advanced-guide/graphql/page.md:194`
```
- **Health**: `/.well-known/health` (Returns detailed dependency status)
```
"detailed" is precisely what this PR removes.

`docs/quick-start/introduction/page.md:117`
```
- `http://localhost:8000/.well-known/health` → JSON status of registered datasources (use this for readiness probes)
```
It is now the aggregate, not per-datasource status. This one is a first-run page, so it is the most likely to be read.

`docs/guides/helm-chart-starter/page.md:204`
```
- The probe paths are GoFr's built-in endpoints. `/.well-known/alive` is cheap and exempt from auth by default; `/.well-known/health` includes dependency status and is more truthful for readiness.
```
"includes dependency status" is out of date. The "more truthful for readiness" contrast still holds, since the aggregate does reflect dependency health.

## Stale wording, lower stakes

`docs/guides/service-mesh-integration/page.md:80`
```
`/.well-known/alive` is the liveness signal; `/.well-known/health` includes dependency status and may be slower.
```
"may be slower" is still true — it does check every dependency. Only "includes dependency status" needs adjusting.

`docs/guides/cloud-deployment/page.md:23` — `readiness (covers dependencies)`
`docs/guides/dockerizing-gofr-services/page.md:117` — `(datasources reachable)`
`docs/guides/deploying-to-kubernetes/page.md:26` — `Gate traffic until datasources are reachable`
`docs/guides/deploying-to-kubernetes/page.md:15` — the same claim inside the `howto` JSON block: `curl /.well-known/health to confirm datasources are reachable`

These four are defensible as written, since the aggregate genuinely does cover dependencies — a `DEGRADED` aggregate does mean something is unreachable. I would leave them alone unless you want the wording consistent. Flagging them so the list is complete rather than because I think each needs an edit.

Note the `deploying-to-kubernetes` one appears twice, and the second is inside a single-line `howto` block, which is easy to miss when editing by eye.

## Why I am blocking on this rather than filing it

Purely because of the two auth pages. The rest is tidying and I would happily take it as a follow-up, but shipping a redaction fix while two pages still instruct users to add a NetworkPolicy "because it reveals dependency status" leaves the docs contradicting the code on a security question — which is the same class of problem #3802 was filed for, just in the opposite direction. It is also cheapest to fix here, while the reasoning is loaded.

Reproduction for the sweep, so nothing is missed:

```
git grep -n "well-known/health" -- docs/
```

30 files. The check that matters is whether the surrounding sentence claims per-dependency content.

Everything else on this PR is resolved from my side — the redaction is verified, the API surface is provably unchanged, and I have no remaining code comments.

akshat-kumar-singhal:
Docs sweep done in `bbd0b716` — 7 files, 9 lines. You were right that it was incomplete, and re-running `git grep -n "well-known/health" -- docs/` rather than working from the list turned up two more.

**The two that were blocking.** `advanced-guide/authentication/page.md:26` and `guides/auth-in-kubernetes/page.md:93` both told the reader to act *because the endpoint leaks datasource details*. Reworded rather than deleted, as you suggested — the authentication one keeps "the auth middleware will not gate it, the framework will not" and drops the false premise in front of it; the auth-in-kubernetes one keeps the exemption documented and replaces the NetworkPolicy remediation with what the body actually contains. I agree stale security advice is the worse failure here, and it is the same class as #3802 pointing the other way.

**Also fixed on those two pages, one line up from each of yours.** `authentication:23` ("readiness/health probe with datasource status") and `auth-in-kubernetes:25` ("Re-protect `health` if you publish it externally") make the same claim as the lines you flagged, so fixing only :26 and :93 would have left each page contradicting itself two lines apart.

**Two your list missed.**

- `guides/deploying-to-kubernetes/page.md:278` — "the JSON body lists which dependency is down." Flatly false after the redaction, and it is the instruction a reader follows *while a readiness probe is failing*, so it sends them hunting for a field that no longer exists. Now: a `DEGRADED` status means at least one dependency is unreachable, and the pod logs say which.
- `guides/helm-chart-starter/page.md:264` — "`/health` includes dependency status" again, inside a `faq-item` body. Your line-level list caught `:204` on that page but not this one; same shape as the `howto` JSON block you flagged on `deploying-to-kubernetes`.

**Wording.** Settled on "reflects/folds dependency status into its aggregate" over "includes dependency status" — as you say, the aggregate genuinely does cover dependencies, so the readiness arguments in those pages still hold and only the implication of per-dependency *content* had to go. I left `cloud-deployment:23`, `dockerizing-gofr-services:117`, and `deploying-to-kubernetes:26`/`:15` alone per your note; they are true as written and editing them would widen the diff past the lines that were actually wrong.

I checked all 30 files for a stale JSON body independently and agree with you: prose only, no per-dependency keys, no `"details"`, no `"version"`.

---

**One thing I found while sweeping that I have deliberately not fixed here, because I think it needs its own PR.**

`healthHandler` returns a nil error, and `handleSuccess` (`pkg/gofr/http/responder.go:281`) maps GET + nil to `StatusOK`. So `/.well-known/health` answers **200 even when the aggregate is `DEGRADED`** — a readiness probe never fails on an unreachable datasource. Two lines in `deploying-to-kubernetes/page.md` assert the opposite:

- `:236` — "returns 200 only when the service **and its dependencies** are reachable"
- `:26` — "Gate traffic until datasources are reachable"

These are pre-existing on `development`, not introduced by this PR, and they are among the four you told me to leave alone — correctly, on the criterion you were applying, since the aggregate does cover dependencies. The problem is the status code, not the body.

I left them because the honest fix is in code (503 on `DEGRADED`), and that is a behaviour change: probes that pass today would start failing for anyone whose service runs with a degraded optional dependency. That deserves its own PR and its own argument rather than being decided inside a docs sweep on a security fix — and documenting the current behaviour instead would be writing down something we probably do not want to keep. Happy to file it, or fold it into #3806 if you would rather the readiness-semantics discussion live with the metrics-port work. Say which and I will open it.

Worth noting it interacts with #3858: if a readiness closure lands, the 200-on-`DEGRADED` question is exactly what it has to answer, so it may be cheapest to settle there.

aryanmehrotra (approved):
Approving at `bbd0b716`. The docs sweep is complete and you found four lines I had missed, two of them flatly wrong.

## The sweep

All six I flagged are fixed, and rewording rather than deleting was the right call on the two auth pages — keeping "the auth middleware will not gate it" and the exemption itself preserves the useful half.

The four beyond my list:

- `deploying-to-kubernetes:275` — "the JSON body lists which dependency is down"
- `helm-chart-starter:261` — "a pod with a broken DB connection will be removed from service endpoints"
- `authentication:23` and `auth-in-kubernetes:22` — same claim as the lines I flagged, one and three lines above them

The first two were false, not merely stale, and my grep is why I missed them: I searched only `well-known/health`, so a line saying `` `/health` `` was invisible to it, and my `.{0,120}` window cut off before the claim on the other. Re-ran it your way — exactly one bare-`/health` line exists in `docs/` and it is the one you fixed. The sweep is genuinely complete now.

## Verified on a running application

Not tests: a real app with basic auth enabled, `payments-prod` pointed at a live listener and `ledger-prod` at a port bound then closed, queried with no credentials.

`development`:

```json
200 {"data":{
  "ledger-prod":{"status":"DOWN","details":{"error":"Get \"http://127.0.0.1:49622/.well-known/alive\": dial tcp 127.0.0.1:49622: connect: connection refused"}},
  "name":"gofr-app",
  "payments-prod":{"status":"UP","details":{"host":"127.0.0.1:49621"}},
  "status":"DEGRADED","version":"dev"}}
```

`bbd0b716`:

```json
200 {"data":{"name":"gofr-app","status":"DEGRADED"}}
```

`DEGRADED` survives, so the redaction drops detail without dropping signal. `/api/orders` still answers `401` unauthenticated, `/.well-known/alive` unchanged.

Then eighteen further requests against that same running app, each body scanned for `127.0.0.1`, `details`, `version`, `dial tcp`, `password`, `root`: query params (`?verbose=true`, `?details=1&full=1`), trailing slash, double slash, `/./` dot-segment, uppercase path, `POST`/`PUT`/`DELETE`/`HEAD`, `Accept: text/plain`, a spoofed `X-Forwarded-For`, the swagger catch-all at `/.well-known/openapi.json`, and the same request with valid credentials. No hits. Valid credentials return byte-identical output to none, which is the property that matters here — the endpoint is not privileged, so there is nothing for auth to unlock.

Build and vet clean. `container` passes under `-race -count=1`. `pkg/gofr` shows two failures locally, `TestContext_WriteMessageToSocket` and `TestStartGRPCServer_Registered`, both wanting port 8000 which something else on my machine holds; they fail identically on `development` and the failing-test sets are the same on both heads, so nothing here caused them. CI is 14/2.

Exported surface is still untouched — `go doc -all` over `pkg/gofr` and `pkg/gofr/container` gives 4079 declarations on both `development` and this head, with an empty diff. The final commit is docs-only.

## On the two lines you left

You were right to leave `deploying-to-kubernetes:236` and `:26`, and I was wrong about the fix I had in mind. I had assumed a docs/code mismatch meant the code was wrong; it is the opposite. `DEGRADED` should not remove a pod from Service endpoints — every replica sharing one Postgres would fail readiness together on a single blip, all of them would leave the endpoint list at once, and callers would get connection-refused instead of a degraded-but-useful service. Returning `200` and reporting the aggregate is the right default, and which dependencies are critical is a per-application decision the framework cannot make.

So the issue to file is not "return 503 on DEGRADED". It is that the docs describe an endpoint that gates traffic when it never does, and that they recommend it as the `readinessProbe` target — where it both cannot gate and triggers an uncached fan-out across every dependency on each probe interval. `/.well-known/alive` is the cheap and honest readiness target. Standalone issue rather than folded into #3806, since it is probe semantics rather than metrics-port work, and it is the question a readiness closure has to answer if #3858 lands.

Happy for you to open it, or I can.

Umang01-hash (approved):

PR #3814, opened 2026-07-31
fix(ci): fail example/pkg test jobs when go test fails (#3812)

### LINKED ISSUE #3812: CI: example/pkg test jobs swallow `go test` failures (missing `set -e` in retry command block)
## Summary

The `Example-Unit-Testing` and `PKG-Unit-Testing` jobs in `.github/workflows/go.yml` can report **green while `go test` has failed**. Failing tests are silently swallowed, so a broken example/test does not block a PR.

## Root cause

Both jobs run their tests through the `nick-fields/retry@v4` action, passing a multi-line `command:` block. For the example job (`.github/workflows/go.yml`, ~L131-138):

```bash
export APP_ENV=test
go test ./examples/... -v -short -covermode=atomic -coverprofile packageWithpbgo.cov -coverpkg=./examples/...   # (1)
grep -vE '(...)' packageWithpbgo.cov > profile.cov                                                              # (2)
go tool cover -func profile.cov                                                                                 # (3)
```

Two things combine:

1. **No `set -e`.** This is a `command:` input to a third-party action, not a native `run:` step. GitHub's native `run:` steps default to `bash --noprofile --norc -eo pipefail` (aborts on first failure); the retry action runs the block with a plain shell that has **no `errexit`**. So when `go test` (1) exits non-zero, the script keeps going to (2) and (3).
2. **Exit code = last command.** The block's exit status is that of its final command, `go tool cover -func profile.cov`, which returns 0. The retry action therefore sees success ("Command completed after 1 attempt(s)") and the job goes green even though `go test` failed.

The same shape (and same latent bug) exists in the `PKG-Unit-Testing` block (~L193-232).

## Impact

A concrete instance surfaced in review of #3807: two example tests (`examples/using-http-service/main_test.go`, `examples/using-ai/main_test.go`) were `--- FAIL` at head, the FAILs appeared in the CI log, yet the Example job still passed. #3807 fixes those two tests; this issue tracks the CI hole that let them stay red without failing the pipeline.

## Suggested fix

Add `set -eo pipefail` at the top of the `command:` block in both jobs so a failing `go test` aborts the block and propagates a non-zero exit (retried, then failing the job). `pipefail` also prevents a failure being masked when a command is piped.

```bash
set -eo pipefail
export APP_ENV=test
go test ./examples/... ...
```

Note: enabling this will likely surface other currently-hidden failures — hence a standalone PR where those can be triaged, rather than folding it into #3807.

### LINKED ISSUE #3816: examples/using-migrations test is not self-contained: migration state persists in Redis between runs
**Describe the bug**

**To Reproduce**

   ```
   ```

**Expected behavior**

**Environments**

- go version: 1.26.5

**More description**

akshat-kumar-singhal:
**Description:**

Fixes #3812.

> **Depends on #3817 — merge that first.** #3820 has since merged and
> `examples/using-html-template` went green exactly as predicted below. The
> `Example Unit Testing (v1.26)` job is now red on `examples/using-add-rest-handlers` and
> `examples/using-migrations`, whose fix lives in **#3817** (open, all checks green). This branch
> carries no change for them; it goes green when #3817 lands. Same shape as the #3820 dependency it
> replaces: this PR un-hides a pre-existing failure whose fix is owned by another branch.
>
> <details><summary>The original #3820 dependency note, now satisfied</summary>
>
> > **Depends on #3820 — merge that first.** One of the failures this PR exposes is a framework bug
> > (a static mount's root returns `403`), not a test bug, and its fix belongs with the rest of that
> > bug in #3820. This branch therefore carries no `pkg/gofr/http` change, and its `Example Unit
> > Testing` job stays **red on `examples/using-html-template` until #3820 lands** — after which it
> > goes green with no change here. Details in the table below.
>
> </details>

The `Example-Unit-Testing` and `PKG-Unit-Testing` jobs in `.github/workflows/go.yml` could report **green while `go test` had failed**. Both run their tests through `nick-fields/retry@v4`, whose `command:` input executes in a plain shell with **no `errexit`** — unlike native `run:` steps, which default to `bash --noprofile --norc -eo pipefail`.

In the Example block the flow is `go test` → `grep` → `go tool cover`. When `go test` exits non-zero the block keeps going, and the block's exit status is that of its **last** command (`go tool cover`, which returns 0), so the retry action sees success and the job goes green. A concrete instance surfaced reviewing #3807: two example tests were `--- FAIL` at head, the FAILs appeared in the CI log, and the job still passed.

**The workflow change**

- **Example block** — prepend `set -eo pipefail`. A failing `go test` now aborts the block and propagates a non-zero exit (retried, then failing the job).
- **PKG block** — prepend `set -eo pipefail`. Because `set -e` would otherwise preempt the block's explicit panic (exit 2) messaging, each `go test` is now captured with `|| exit_code=$?` (preceded by `exit_code=0`); the existing `if/elif` still distinguishes a panic from an ordinary failure. `pipefail` also hardens the `grep | grep >> profile.cov` coverage-combine line.

**Exposed example failures — five fixed here, one by #3820**

Enabling `errexit` did exactly what it was meant to do: `Example Unit Testing` went red on all three Go versions, on six pre-existing failures the old block had been swallowing. `PKG Unit Testing` stayed green.

**Five of the six are test bugs and are fixed here** rather than split out. The original plan was to triage them separately, but that only works if the branch can be merged in the meantime — and it cannot: the whole point of this change is that the job now fails on them. Splitting would leave this PR permanently red and blocked on N follow-ups, each itself gated on a job that only goes green once all of them land. So the fix and the test failures it uncovers ship together. Each is a genuine fix; nothing is skipped, deleted, or loosened to pass.

**The sixth is a framework bug and is not fixed here.** `examples/using-html-template` fails because a static mount's root is unreachable — a defect in `pkg/gofr/http/router.go`, not in the example's test. This branch originally carried a fix for it, but that same bug is one of three causes addressed in full by **#3820**, which fixes the containment check, the routing of the bare endpoint, and a redirect loop, and replaces `http.FileServer` with a single resolved serving path. Keeping a partial fix here would have meant two branches editing the same functions (they conflict textually) and a framework behaviour change riding in on a CI PR. So the framework change was dropped from this branch, and this PR's Example job depends on #3820 for that one package.

| Package | Root cause | Kind |
| --- | --- | --- |
| `examples/grpc/grpc-streaming-server` | Test pinned the example's documented `GRPC_PORT=9000`. Packages run concurrently under `go test ./examples/...` and this job also binds host `9000` for MinIO, so the server could not bind and the package aborted before any test ran. Now reserves free gRPC/HTTP/metrics ports in `TestMain`. | test |
| `examples/http-server` | `TestTraceHandler` pinned the exact `context.Context` value it expected the handler to pass. The handler passes the `*gofr.Context` itself and `Trace()` swaps the embedded context on each call, so that identity is not part of the contract. Matched with `gomock.Any()`; path and query params are still asserted. | test |
| `examples/http-server-using-redis` | `TestRedisPipelineHandler` built its URL as `"http://localhost:" + configs.HTTPHost`, which already carries scheme/host/port. The URL was unparsable, the error was discarded, and the nil `*http.Request` panicked (SIGSEGV) before the handler ran. Also repointed the Redis config from `2001` (MySQL in this job's service matrix — hence `redis: invalid reply: "I\x00\x00\x00\n"`, a MySQL handshake) to `2002`. | test |
| `examples/sample-cmd` | Two causes. The expected `"No Command Found!"` string no longer exists — the framework reports `'x' is not a valid command.` And `TestCMDRunWithNoArg` never set `os.Args`, so it inherited the test binary's flags; the coverage run adds `-test.testlogfile`, which the CMD router read as a subcommand. Expectations updated; every `os.Args` mutation now goes through a save/restore helper. | test |
| `examples/using-html-template` | `Test_IndexHTML` got `403 Forbidden` for `/`. Real framework bug: `AddStaticFiles` strips the mount prefix, so a request for the mount endpoint resolves back to the static directory, and `isRestrictedFile` required a trailing path separator after the directory name (the sibling-escape guard from #3041) — so a static site's `index.html` was unreachable. **Not fixed here — fixed by #3820**, which addresses this plus two further causes of the same symptom. This package stays red until that merges. | **framework — #3820** |
| `examples/using-http-service` | `TestHTTPHandlerURLError`'s histogram expectation could never be satisfied: `NewMockContainer` pre-registers a catch-all `RecordHistogram(...).AnyTimes()` and gomock matches the first non-exhausted expectation, so the specific one added later never matched. The service under test now gets its own `service.MockMetrics`, making the assertion real. The production path *does* record the histogram on a request error (`updateMetrics` runs with `StatusServiceUnavailable`) — this was mock shadowing, not a missing metric. | test |

**Seventh failure: `go: no such tool "covdata"` on Go 1.24 / 1.25**

With all six above fixed (the framework one by #3820, applied locally) the 1.26 leg went green, but 1.24 and 1.25 stayed red for an unrelated, also pre-existing reason: every package **without test files** failed to build with `go: no such tool "covdata"`, so `go test` exited non-zero even though every test passed. The same errors are present in this branch's earlier run (hidden among the six failures) and on `development`, where the missing `errexit` swallowed the non-zero exit outright.

`setup-go` installs the matrix version; `GOTOOLCHAIN=auto` then lets the go command upgrade to the `go1.26.0` that `go.work` requires — by downloading a **toolchain module** into the read-only module cache. Go 1.26 no longer ships `covdata` prebuilt in `pkg/tool` (it builds the tool into `GOROOT` on demand) and a toolchain-module `GOROOT` cannot be written to, so the `-coverpkg` run has no way to obtain it.

Reproduced locally with `go1.24.13` + `GOTOOLCHAIN=auto`, and confirmed the *same* `go1.26.0` toolchain works when its `GOROOT` is a normal directory rather than a module-cache entry. Fixed by installing the version `go.work` asks for as a real SDK (`golang.org/dl`) when the installed Go is older; the toolchain switch then resolves it from `PATH` against a writable `GOROOT`.

That fix lives in a local composite action, `.github/actions/setup-go-toolchain`, and **all three** test jobs now use it. It was briefly applied to the Example job alone, on the reasoning that PKG and Submodule build no test-less package under `-coverpkg` — that reasoning is wrong: `./pkg/gofr/...` contains two (`pkg/gofr/static` and `pkg/gofr/version`). Those jobs are green today, but nothing structural keeps them that way, and the `Submodule-Unit-Testing` `errexit` follow-up below will turn the same hidden failure into a red job exactly as it did here. Sharing the step also removes 24 lines from the workflow and stops the three jobs differing in how they resolve the `go.work` toolchain.

None. Nothing here changes what a running app serves — the changes are to CI and to example tests. The static-mount behaviour change that this branch previously carried now lives in #3820, which documents it.

The one thing to know is a CI-visible consequence of that split: the `Example Unit Testing` job on this branch stays red until its prerequisite lands. #3820 has merged and `examples/using-html-template` is green; the job is now red on `examples/using-add-rest-handlers` and `examples/using-migrations`, pending #3817. In both cases that is the new `errexit` correctly reporting a pre-existing failure, which is exactly what #3812 asked for.

**Additional Information:**

No new dependencies, and no change to `pkg/gofr` — the diff is the workflow, the new composite action, and five example test files.

Verification:

- `yaml.safe_load` parses the workflow and the composite action; `bash -n` clean on the extracted `command:` blocks and on the action's script.
- Simulated the PKG control flow: `go test` exit `0` → block `0` (coverage combine runs), exit `1` → block `1` (`::error::…failed`), exit `2` → block `2` (`::error::…Panic`).
- Mirrored this job's services locally (Redis `2002`, MySQL `2001`, Kafka `9092`, Zipkin `2005`, MinIO `9000`) and ran the job's exact command on Go 1.26.5 — `go test ./examples/... -v -short -covermode=atomic -coverprofile … -coverpkg=./examples/...` → exit `0`, 25 packages `ok`, at the time with this branch's own (now removed) static-root fix in place for `examples/using-html-template`; #3820 supersedes it and fixes the same 403. Each package also verified in isolation. Without that fix the run is identical except that one package, which fails as described above.
- `golangci-lint` before/after on every changed package gives identical finding sets — no new findings.

**Checklist:**

**Follow-ups (deliberately out of scope):**

-   The structurally-identical `Submodule-Unit-Testing` retry block has the same class of failure-masking bug, and additionally masks failures via `xargs … bash -c '… ; cd -'`, where the trailing `cd -` returns 0. Left for a separate PR so the failures *it* exposes can be triaged independently, per #3812's scope.
-   ~~#3816 — `examples/using-migrations` keeps migration bookkeeping in Redis and is not self-contained across runs. Green here only because CI gives each job a fresh Redis container.~~ Addressed by **#3817**, which closes #3816 on merge — and which this PR now **depends on** (see the note at the top): #3817 also replaces the fixed startup sleep in both `using-migrations` and `using-add-rest-handlers`, the two packages failing here.
-   ~~#3818 — 26 example tests still wait for the server with a fixed `time.Sleep` instead of polling a readiness signal. Not folded in here: every job above is green with those sleeps in place, so this PR does not expose them — the `max_attempts: 2` retry is what currently absorbs the race. Worth doing before that retry is ever dropped, and because a real regression in one of those examples can be masked by a retry that happens to pass.~~ **Superseded — this turned out to be a prerequisite, not a follow-up.** Both halves of that reasoning failed on a slower runner. On Go 1.26, `using-add-rest-handlers` and `using-migrations` fail with `dial tcp … connect: connection refused`, and the `max_attempts: 2` retry does *not* absorb it — a failed package isn't cached, so attempt 2 reproduces it (run 30921266821, both attempts). Both examples call `app.Migrate()` synchronously in `main()` before `app.Run()` binds the listener, so startup outlasts the 100 ms sleep whenever the migration does; the 1.24 and 1.25 legs pass on timing alone. Their fix is **#3817**. The remaining 15 example tests are **#3821** (open, green), which stays a genuine follow-up — those packages are green today.


Umang01-hash (dismissed):
Verified the whole thing at head `eca3ab0` — built, ran the touched tests, traced the CI exit-code logic. The masked-failure claim is real and the fix is a proper root-cause one.

- **CI**: `set -eo pipefail` in the retry blocks correctly fails the job on red tests (the block's exit code was previously `go tool cover`=0). PKG job handles both panic (exit 2) and normal fail (exit 1 → `exit $exit_code`). No exit-1 can slip through.
- **router.go**: exactly one truth-table cell flips — mount-root index now serves instead of 403. #3041 sibling-escape guard and the openapi.json restriction both stay intact; traversal attempts still 403. No exported symbol touched.
- **test fixes** are all root-cause; the gomock changes in using-http-service/http-server actually *strengthen* assertions that were fake-green before. `os.Exit(m.Run())` in grpc is the important one — that package was silently passing on failures.
- Not flakier: free-port reservation replaces a deterministic 9000/MinIO collision. Ran the fixed tests green, grpc 3×.

Two non-blocking notes:
1. The `Submodule-Unit-Testing` retry block still has no `set -e` — same masking bug, just outside #3812's stated scope. Worth folding in here or a quick follow-up.
2. Pre-existing (not this PR): a *relative* dirName in AddStaticFiles still 403s the root index.

LGTM.

aryanmehrotra:
Verified the core claim — the retry blocks were green on red tests, and `set -eo pipefail` is the right fix. Two asks.

## 1. Split `pkg/gofr/http/router.go` out

This PR is scoped to the pipeline. `router.go` changes HTTP behavior on a security-adjacent path: `GET /` on a static mount goes from 403 to 200, and a mount with no `index.html` now returns a browsable directory listing instead of being closed. That is a framework change and it does not belong under `fix(ci)`, regardless of how it was found.

The five `examples/*/main_test.go` fixes are a different matter — they're test-only and they're the direct consequence of the pipeline change, so keeping them here is right.

On the "splitting leaves this PR permanently red" argument: that holds for splitting all six, not for splitting this one. It's a single ordered dependency, not N gated follow-ups:

1. `fix(http): serve index.html at a static mount root` — `router.go` + `router_test.go`, two files. `examples/using-html-template` is untouched by this PR and starts passing on its own.
2. Rebase this PR on top. Example job goes green.

The root-listing question then gets decided in the PR that owns it rather than as a footnote here. My own view: don't list. One check in `validateFile` —

```go
if fileInfo.IsDir() {
	if _, err := os.Stat(filepath.Join(absPath, "index.html")); err != nil {
		return fs.ErrNotExist
	}
}
```

`os.IsNotExist` catches it, so the existing custom-404 path handles the response.

Two pre-existing bugs in that same function are worth folding into the router PR while it's open (both identical on `development`, neither a regression here):

- `GET /openapi.JSON` → 200, serves the spec. The guard is a case-sensitive `filepath.Base(url) == "openapi.json"`; `strings.EqualFold` closes it. Only reachable on case-insensitive filesystems.
- A relative `dirName` 403s *every* path, not just the root — `absPath` gets absolutized, `directoryName` doesn't. `filepath.Abs` at the top of `AddStaticFiles` kills the class. Only reachable via the internal Router API today, since `App.AddStaticFiles` normalizes.

## 2. The toolchain action is in scope, but it's the wrong fix

The matrix exists to prove gofr compiles on older Go. It can't: `go.mod` declares `go 1.26.0`, a hard minimum since Go 1.21, so Go 1.24 refuses to build the module outright:

```
go: go.mod requires go >= 1.26.0 (running go 1.24.13)
```

`GOTOOLCHAIN=auto` is the only reason those legs are green — by switching to 1.26. This PR's log confirms it: `materialising Go 1.26.0 as a real SDK`, then the tests run on 1.26. So `covdata` isn't a toolchain bug to work around, it's a symptom of testing versions we don't support.

Built both floors for real in a scratch worktree:

- `go 1.24.0` + Go 1.24.13 → fails on `otel@v1.44.0 requires go >= 1.25.0`. **1.24 can never give signal regardless of what we do** — that leg should go.
- `go 1.25.0` + Go 1.25.6 → `go build ./pkg/...` and `go vet ./pkg/gofr` clean.

So the goal is reachable, just not this way:

```
go.mod / go.work:  go 1.25.0
                   toolchain go1.26.0
CI:                GOTOOLCHAIN=local
matrix:            ['1.26', '1.25']
```

`covdata` then never happens — it only broke because `auto` fetched a toolchain module into a read-only GOROOT. That deletes `.github/actions/setup-go-toolchain` entirely and makes the matrix real.

Caveat: I only built the main module's `./pkg/...` with `GOWORK=off`; the submodules have their own `go.mod` and may sit higher. Also worth confirming whether #3426 bumped to 1.26 for a reason or just tidiness.

Both are pipeline work, so either is legitimately in scope here. I'd rather see the floor fix than the workaround, but I won't hold the `set -eo pipefail` fix for it — your call which lands first.

## For the Submodule follow-up

It needs more than `set -e`. `cp ${module_name}.cov ../../../coverage_reports/` hardcodes three levels and every submodule is four or five deep. This PR's log has 27 × `cp: cannot create regular file ... Not a directory`, and `submodule-coverage-reports` is absent from the run's artifacts — submodule coverage has never worked. `go -C "$module" test -coverprofile="$GITHUB_WORKSPACE/coverage_reports/$name.cov"` removes the cd/cp round trip, and `if-no-files-found: error` on the upload step stops it going quiet again.

akshat-kumar-singhal:
Update: the `pkg/gofr/http/router.go` fix that was on this branch has been removed — it's a framework bug (static mount root → 403) and is one of three causes fixed properly in #3820, which also handles the bare-endpoint 404 and a redirect loop. Both branches were editing the same functions and conflicted.

**This PR now depends on #3820 and should merge after it.** Until then `Example Unit Testing` is red on `examples/using-html-template` — that is the new `errexit` correctly surfacing the pre-existing bug, which is the point of #3812. No change here is needed once #3820 lands.

akshat-kumar-singhal:
On your two notes, @Umang01-hash:

**2 (relative dirName 403s the root index)** — fixed in #3820, not here, since this branch no longer touches `pkg/gofr/http`. It turned out broader than the note: with a relative name — or an absolute one carrying a trailing separator — *every* path under the endpoint 403s, not just the root. `staticHandler` resolves the request with `filepath.Abs`, while the containment check compares against `directoryName` as handed in, so the two are never in the same form. Resolved once at registration now; details and the reproduction in [that PR](https://github.com/gofr-dev/gofr/pull/3820#issuecomment-5163563654).

**1 (`Submodule-Unit-Testing` has no `set -e`)** — keeping it as a follow-up, because `set -e` alone does not fix that job. As @aryanmehrotra notes above, `cp ${module_name}.cov ../../../coverage_reports/` hardcodes three levels while submodules sit four or five deep: this run logs 27 × `cp: cannot create regular file ... Not a directory`, and `submodule-coverage-reports` is absent from the artifacts — submodule coverage has never worked. Adding `errexit` on top of that turns a job that silently produces nothing into a red job that still produces nothing.

The follow-up wants all of it together: `go -C "$module" test -coverprofile="$GITHUB_WORKSPACE/coverage_reports/$name.cov"` to drop the cd/cp round trip, `if-no-files-found: error` on the upload so it cannot go quiet again, `set -eo pipefail`, and the `xargs … bash -c '… ; cd -'` masking where the trailing `cd -` returns 0. That is a different change from this one and will expose its own failures to triage — which is the separation #3812 asked for.

akshat-kumar-singhal:
Filed the submodule follow-up as #3826 — the `cp` depth bug, the never-produced `submodule-coverage-reports` artifact, and the missing `set -e` / `cd -` masking, with the run evidence.

aryanmehrotra (approved):
Re-reviewed at `1a1027df`. `router.go` is gone, #3820 has landed, and CI is 14/14 — the dependency you flagged resolved itself. Approving.

## The mechanism is right, including the part that is easy to get wrong

`set -eo pipefail` on its own would have broken the panic detection, and the `|| exit_code=$?` rework is what keeps it working. Confirmed the two forms:

```
set -eo pipefail; false; exit_code=$?                 -> block dies on `false`, exit 1,
                                                         the assignment never runs
set -eo pipefail; exit_code=0; false || exit_code=$?  -> reached, exit_code=1
```

So the old `exit_code=$?` capture was not merely redundant under errexit, it was unreachable — and `|| exit_code=$?` plus the explicit `elif [ $exit_code -ne 0 ]; then exit $exit_code` is what keeps a plain exit-1 failure fatal while still letting exit 2 be reported as a panic. That distinction is the whole reason this block cannot just be `set -e` and nothing else, and it is worth the comment you put on it.

The grpc-streaming fix is the right shape too: `configs/.env` pinning `GRPC_PORT=9000` while `go test ./examples/...` runs packages concurrently and CI binds 9000 for MinIO is exactly the class of failure the retry block was hiding, and asking the kernel for a free port is the fix rather than picking a different fixed one.

## One correction to the description

`m.Run()` → `os.Exit(m.Run())` is stylistic, not a fix. Since Go 1.15 the test wrapper passes `m.Run`'s result to `os.Exit` when `TestMain` returns, so failures already propagated. Verified both forms:

```
TestMain { m.Run() }           -> go test exit=1
TestMain { os.Exit(m.Run()) }  -> go test exit=1
```

No objection to the change — being explicit is fine and matches the other examples — but it should not be listed among the swallowing fixes, because it was not swallowing anything.

## The matrix, which I am not holding this for

My earlier note stands unanswered, so recording the evidence and moving it out of your way. From this PR's own run (30977134341), the `1.24` leg:

```
Successfully set up Go version 1.24
go version go1.24.13 linux/amd64
GOTOOLCHAIN='local'
materialising Go 1.26.0 as a real SDK for the toolchain switch
```

and the tests then execute on 1.26.0. `go.work` and `go.mod` both declare `go 1.26.0`, so all three legs test the same version and the matrix proves nothing about 1.24 or 1.25.

That is a question about which Go versions GoFr supports, not about CI swallowing failures, so it is not this PR's to answer — and your action is what makes the matrix as it stands work at all, given `covdata` is missing from a toolchain-module GOROOT. Filed as its own issue with the floor measurements I took (`go 1.24.0` cannot work at all — `otel@v1.44.0` requires `>= 1.25.0`; `go 1.25.0` builds and vets clean). Whoever picks it up deletes `.github/actions/setup-go-toolchain` as a side effect.

## Two notes, neither blocking

`grep -h -v "mode:" gofr_only.cov submodules.cov | grep -v '/mock_' >> profile.cov` is a new failure path. Under `pipefail` a `grep` that matches nothing exits 1 and now kills the block, where before it was ignored. It needs every remaining coverage line to be a mock file to trigger, so it is not realistic for `./pkg/gofr` — but it is a behaviour change introduced by this PR rather than a pre-existing one, and `|| true` on the second `grep` would close it if you would rather not think about it again.

`- name: Upload Test Coverage` is gated on `matrix.go-version == '1.24'`, so `PKG-Coverage-Report` comes from the leg that actually runs 1.26. Harmless today, but it means removing the 1.24 leg silently stops coverage upload. Worth pinning to `'1.26'` so the gate names the version it really is, whichever way the matrix question goes.

Issue #3801, opened 2026-07-28
Graceful stop establishes no happens-before edge: lost server shutdown (nil-check-as-started) + cron Stop() without job join


akshat-kumar-singhal:
Follow-up from the review on #3800. Two lifecycle bugs share one shape: **a "stop" that establishes no happens-before edge with the thing it is stopping.** #3800 fixes the memory-safety half (the `-race`-visible data race on `srv`); these two ordering bugs remain and are *not* visible to `-race`.

## 1. Lost shutdown: `srv == nil` doubles as "not started yet"

`run.go` starts the shutdown handler before the servers:

```go
a.startShutdownHandler(ctx, timeout)   // parks on <-ctx.Done()
a.startTelemetryIfEnabled()
a.startAllServers(ctx)                 // servers assign srv in here
```

A termination signal arriving in that window is observed by a `Shutdown()` that sees `srv == nil`, reports success, and returns — then `run()` brings the listener up and serves forever:

```
 t0  SIGTERM arrives
 t1  shutdown g:  Shutdown() -> lock -> srv == nil -> unlock -> return nil  ("nothing to stop")
 t2  serve g:     s.srv = srv
 t3               srv.ListenAndServe()   <== serves forever
 t4  wg.Wait() never returns -> process hangs until SIGKILL
```

Reproduced on the #3800 branch: shutdown returns `nil`, then the listener comes up and stays up. In practice this is a SIGTERM landing right after start (crashloop, fast rollout, failed readiness) where the pod ignores graceful shutdown and hangs until SIGKILL.

Root cause: `srv` is both the server handle and the "has it started?" flag; a nil check carries no happens-before edge, so a mutex fixes the read/write hazard (#3800) but not the ordering. Applies to `httpServer`, `metricServer`, `mcpServer`.

**Fix direction:** a per-server `started chan struct{}` (or a lifecycle state) that `Shutdown` waits on — turning "not started yet" into "wait until started (or until start is known not to happen), then stop" instead of "nothing to do". Must also handle the case where the server errors out before assigning `srv` (cert validation failure, blocked port) so `Shutdown` cannot block forever.

## 2. Cron `Stop()` does not join in-flight jobs (and ignores the shutdown context)

```go
func (c *Crontab) Stop() {
	c.once.Do(func() {
		c.ticker.Stop()
		close(c.done)
	})
}
```

[...]

PR #3815, opened 2026-07-31
fix(cron): join in-flight jobs on Stop and skip jobs with no logger

### LINKED ISSUE #3799: Flaky unit tests: TestCronTab_AddJob panics in goroutine after completion; FileTokenAuth refresh-log tests are time-dependent

### LINKED ISSUE #3813: Flaky tests on go1.26: cron goroutine nil-logger panic + Test_SQLRetryConnectionInfoLog mock ordering
## Summary

Two unit tests fail intermittently — observed on the `PKG Unit Testing (v1.26)` matrix leg while `v1.24`/`v1.25` passed on the same commit. Both pass reliably when run in isolation locally, so they are flaky rather than broken. Surfaced in CI for #3807 (whose own changes are unrelated — its health tests all passed in the same run).

## 1. Panic: cron background goroutine logs after teardown (`pkg/gofr`)

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0]
gofr.dev/pkg/gofr/logging.(*ContextLogger).Infof(...)  pkg/gofr/logging/ctx_logger.go:72
gofr.dev/pkg/gofr.(*job).run(...)                       pkg/gofr/cron.go:120
created by gofr.dev/pkg/gofr.(*Crontab).runScheduled    pkg/gofr/cron.go:107
```

A scheduled cron `job.run` fires from a background goroutine and calls `Infof` on a logger that is already nil after the owning test tore its container/app down. It's a lifecycle race between `Crontab.runScheduled` and test cleanup, so it only trips under certain scheduling/timing. Related de-flaking has already been done in 24198290 ("de-flake cron/file-token unit tests"); this path still leaks.

Likely fix: guard `job.run` against a nil logger/container, and/or ensure `runScheduled` goroutines are stopped and joined during shutdown before the logger is cleared.

## 2. `Test_SQLRetryConnectionInfoLog` — order-sensitive mock (`datasource/sql`)

```
Got:  app_sql_open_connections (string)
Want: app_sql_inUse_connections (string)
missing call(s) to *sql.MockMetrics.SetGauge(app_sql_open_connections, 0)
missing call(s) to *sql.MockMetrics.SetGauge(app_sql_inUse_connections, 0)
aborting test due to missing call(s)   pkg/gofr/datasource/sql/sql_test.go:369-370
```

The test sets ordered gomock expectations on the two `SetGauge` gauge names, but the production code emits them in a non-deterministic order, so the mock controller occasionally sees them swapped and reports missing calls.

Likely fix: make the two `SetGauge` expectations order-independent (drop the implicit ordering / use `.AnyTimes()`-style matchers or an unordered group) rather than pinning a sequence the code doesn't guarantee.

## Repro

Not reliably reproducible in isolation; observed under the full `go test ./pkg/gofr/...` run on Go 1.26 in CI. Re-running the job clears it.

akshat-kumar-singhal:
**Description:**

Fixes #3813. Also closes #3799 and the cron half of #3801. #3801 should stay open for the server half.

**What this actually fixes is data loss during shutdown, not just a flaky test.** `App.Shutdown` stops cron and *then* closes the datasources:

```go
a.cron.stop(...)       // before this change: returned as soon as the ticker halted
...
a.container.Close()    // closes SQL, Redis, pub/sub
```

Because `Stop()` returned without joining, a job still mid-work when `Shutdown` was called kept running against datasources that were being torn down underneath it — its next query fails with `sql: database is closed`, a half-written batch loses its remaining work, and the error lands after the log file is closed too. Confirmed in review against a real app on MySQL: on `development` `Shutdown` returned in 0s and the job's next query errored; with this change `Shutdown` waited 110ms and the query returned `<nil>`.

The nil-logger crash likewise is not test-only: calling `job.run` with a logger-less container panics `SIGSEGV` in `logging.(*ContextLogger).Infof` and takes the process down. The minute-boundary timing is only what made CI hit it intermittently.

#3799 reports a different symptom from the same leak: `TestCronTab_AddJob`'s unstopped `Crontab` crosses a minute boundary and its job goroutine calls the metrics mock after the test finished, so gomock raises `Fatalf` on a completed test's controller. The `Stop()` join added here is what closes that window — the `AnyTimes()` expectations merged in 24198290 (#3800) only masked it. #3799's other half (the `FileTokenAuth` refresh-log tests) already landed in that same commit, so this PR completes it.

Two unit tests failed intermittently on the `PKG Unit Testing (v1.26)` CI leg while `v1.24`/`v1.25` passed on the same commit. Both are the same shape: **a background goroutine outliving the test that owns the state it touches.**

**1. `pkg/gofr` — SIGSEGV in the cron job goroutine**

```
logging.(*ContextLogger).Infof  ->  gofr.(*job).run  <-  created by gofr.(*Crontab).runScheduled
```

`Test_AddCronJob_Success` builds `App{container: &container.Container{}}` — **no logger** — registers a `"* * * * *"` job, and never stops the scheduler. `NewCron`'s ticker goroutine therefore runs for the rest of the test binary; at second 0 of the next minute it dispatches `go j.run(container)`, which builds a `ContextLogger` over a nil base and segfaults in `Infof`. Whether the binary is still alive at a minute boundary is the entire timing dependence.

`Stop()` could not have prevented this either: it halted the ticker but did **not** join job goroutines an earlier tick had already dispatched (`go j.run(...)`, no `WaitGroup`) — the gap already noted in a `cron_test.go` comment pointing at #3801.

**2. `datasource/sql` — `Test_SQLRetryConnectionInfoLog`**

```
missing call(s) to *sql.MockMetrics.SetGauge(app_sql_open_connections, 0)
missing call(s) to *sql.MockMetrics.SetGauge(app_sql_inUse_connections, 0)
```

Not an ordering problem, despite the issue's initial read — `pushDBMetrics` emits the two gauges in a fixed order. It is count and timing: the expectations sat at gomock's default `Times(1)` while the goroutine re-emits both every 10s and the DB was never closed, and the test waited a fixed `100ms`. On a loaded runner nothing had been emitted by the time the controller finished.

**Changes**

Framework (`pkg/gofr/cron.go`, `pkg/gofr/gofr.go`):

- `Crontab` tracks dispatched jobs in a `sync.WaitGroup` and sets a `stopped` flag under the existing mutex, so `Stop` halts the ticker **and** joins in-flight jobs, and a tick racing `Stop` cannot dispatch after the join begins.
- `App.Shutdown` joins through an unexported `stop(abort <-chan struct{})` fed by `ctx.Done()`, so a long-running user job cannot hold shutdown open past the caller's own deadline. Public `Stop()` passes `nil` and waits indefinitely — that is the standalone/test contract. `stop` takes a channel rather than a `context.Context` to avoid growing the cron API a context parameter it has no other use for.
- `job.run` returns early when the container or its logger is nil, so a detached job goroutine can never nil-panic the process.

Guarding in `job.run` rather than in `logging.NewContextLogger` is deliberate: making `ContextLogger` tolerate a nil base would kill this class of panic framework-wide, but it would also silently swallow logs for every misconfigured container and mask the underlying wiring bug. The nil case is real only for detached background goroutines, so the guard sits where the detachment happens.

Tests:

- Both `AddCronJob` tests stop their scheduler; the success case gets a real logger.
- `Test_SQLRetryConnectionInfoLog` waits on a logger that signals when the retry line is actually written (no sleep), closes the DB via `t.Cleanup` so nothing touches the controller after the test, and relaxes the gauge expectations to `AnyTimes()` — they are incidental to what this test asserts.
- New `TestCrontab_Stop_JoinsInFlightJobs`: `Stop` blocks until a running job returns, and a post-`Stop` tick dispatches nothing. Both halves are asserted by counting runs in an `atomic.Int64` — the second `Stop` joins whatever the post-`Stop` tick dispatched, so the count is read after any such job has returned. Verified by mutation: removing the `stopped` check in `runScheduled` fails this test `1 != 2`. (An earlier revision only *called* `runScheduled`/`Stop` with a comment and no assertion, which the re-dispatched job's own `recover()` swallowed.)
- New `TestCronTab_runScheduled_NilLoggerAndContainer`: a nil / logger-less container is skipped rather than fatal.

`Crontab.Stop()` now blocks until in-flight cron jobs return, where previously it returned as soon as the ticker was stopped. That is the point of the fix — an unjoined job keeps logging and recording metrics against a container the caller is tearing down — but it is a visible behaviour change for anyone calling `Stop()` directly with a long-running job.

Application shutdown is not exposed to that: `App.Shutdown` passes its own `ctx.Done()` as the abort channel, so the join is capped at the caller's existing deadline and a slow job cannot extend shutdown.

No API signature changes; `Stop()` keeps its zero-argument form.

**Additional Information:**

No new dependencies.

Verification:

- `go build ./...` clean; `go vet` clean on both packages.
- `go test ./pkg/gofr/ ./pkg/gofr/datasource/sql/` passes; 5 consecutive runs of `pkg/gofr` green.
- `golangci-lint` reports no new findings from the changed files — the hits that remain in them all reproduce on `development`.

Two pre-existing failures confirmed unrelated, by stashing the change and re-running on clean `development`:

- `TestGRPC_ServerShutdown_ContextCanceled` races under `-race` (pre-existing `g.server` race in `grpc.go`).
- `go test ./pkg/gofr/ -count=3` fatals on the fixed metrics port 2121.

**Checklist:**

### DIFF
--- a/docs/guides/graceful-shutdown/page.md
+++ b/docs/guides/graceful-shutdown/page.md
@@ -9,7 +9,7 @@ nextjs:
 # Graceful Shutdown
 
 {% answer %}
-GoFr listens for `SIGINT` and `SIGTERM` and, on either signal, runs `App.Shutdown` which calls `Shutdown` on the HTTP, gRPC, and metrics servers and `Close` on the container's datasource connections. The shutdown is bounded by `SHUTDOWN_GRACE_PERIOD` (default `30s`); if it expires the process exits with whatever connections remain. Pair this with Kubernetes' `terminationGracePeriodSeconds` and a small `preStop` sleep to avoid losing in-flight requests during rolling restarts.
+GoFr listens for `SIGINT` and `SIGTERM` and, on either signal, runs `App.Shutdown` which calls `Shutdown` on the HTTP, gRPC, and metrics servers, stops the cron scheduler and waits for jobs already running, then calls `Close` on the container's datasource connections. The shutdown is bounded by `SHUTDOWN_GRACE_PERIOD` (default `30s`); if it expires the process exits with whatever connections remain. Pair this with Kubernetes' `terminationGracePeriodSeconds` and a small `preStop` sleep to avoid losing in-flight requests during rolling restarts.
 {% /answer %}
 
 ## When to use
@@ -24,15 +24,17 @@ Every production GoFr deployment on Kubernetes should be configured for graceful
 ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
 ```
 
-When that context is canceled, a goroutine creates a timeout context using `SHUTDOWN_GRACE_PERIOD` (default `30s`) and calls `App.Shutdown`. The order is fixed by the framework — see [`pkg/gofr/gofr.go:96-114`](https://github.com/gofr-dev/gofr/blob/main/pkg/gofr/gofr.go) — and `Shutdown` joins errors from each step:
+When that context is canceled, a goroutine creates a timeout context using `SHUTDOWN_GRACE_PERIOD` (default `30s`) and calls `App.Shutdown`. The order is fixed by the framework — see [`pkg/gofr/gofr.go`](https://github.com/gofr-dev/gofr/blob/main/pkg/gofr/gofr.go) — and `Shutdown` joins errors from each step:
 
 1. `httpServer.Shutdown(ctx)` — stops accepting new connections, waits for in-flight handlers
 2. `grpcServer.Shutdown(ctx)` — drains active streams
-3. `container.Close()` — closes SQL pools, Redis clients, Pub/Sub consumers, and other registered datasources
+3. Cron stop — halts the scheduler and waits for jobs already running, bounded by the shutdown deadline
 4. `metricServer.Shutdown(ctx)` — stops `/metrics`
-5. Logger close — if the logger implements `io.Closer`, its `Close()` is called last
+5. `mcpServer.Shutdown(ctx)` — stops the MCP server
+6. `container.ShutdownMetrics(ctx)` and `container.Close()` — closes SQL pools, Redis clients, Pub/Sub consumers, and other registered datasources
+7. Logger close — if the logger implements `io.Closer`, its `Close()` is called last
 
-The container's `Close` is what commits Pub/Sub offsets and lets SQL drivers finish in-progress queries. Application code does not need to coordinate this order.
+The container's `Close` is what commits Pub/Sub offsets and lets SQL drivers finish in-progress queries. It runs second-to-last on purpose: every producer of datasource traffic — handlers, streams, cron jobs — is drained before the connections they use are torn down. Application code does not need to coordinate this order.
 
 ## OnStart hooks vs shutdown hooks
 
@@ -87,7 +89,7 @@ For a service with 2s P99, that's 5s + 30s + 10s = 45–60s.
 - **SQL.** `database/sql` waits for active queries to finish on `Close()`. Long-running transactions can extend shutdown — keep request timeouts shorter than `SHUTDOWN_GRACE_PERIOD`.
 - **Redis / NoSQL.** Clients close idle connections immediately and wait for in-flight commands.
 - **Pub/Sub.** GoFr's subscription manager respects the shutdown context — consumers stop polling and commit current offsets where the broker supports it (Kafka, NATS JetStream).
-- **Cron jobs.** GoFr's `App.Shutdown` drains HTTP, gRPC, and metrics servers and closes datasource connections — it does **not** stop the cron scheduler or wait for in-flight cron tasks. Cron jobs run with `context.Background()`, so they continue past SIGTERM and may be cut off when the container is killed at `terminationGracePeriodSeconds`. If you have long-running cron work that must finish, run it as a separate Kubernetes `Job` triggered by a `CronJob` resource instead of inside the same pod, so the pod's lifecycle doesn't interrupt it.
+- **Cron jobs.** `App.Shutdown` stops the cron scheduler and waits for jobs that are already running before it closes datasource connections, so a job mid-write is not cut off by `container.Close()`. The wait is bounded by the shutdown deadline (`SHUTDOWN_GRACE_PERIOD`): if a job is still running when that expires, `Shutdown` returns and the job is abandoned. The jobs themselves run with `context.Background()`, so a job's own `ctx` is never cancelled — anything that needs to be interruptible has to watch something else. For cron work that routinely outlasts `SHUTDOWN_GRACE_PERIOD`, run it as a Kubernetes `CronJob` in its own pod rather than inside the service.
 
 ## Verification
 
--- a/pkg/gofr/cron.go
+++ b/pkg/gofr/cron.go
@@ -35,6 +35,14 @@ type Crontab struct {
 	container *container.Container
 
 	mu sync.RWMutex
+	// stopped is set under mu before done is closed, so a tick that is already
+	// inside runScheduled cannot dispatch a new job goroutine after Stop has
+	// begun waiting for the in-flight ones.
+	stopped bool
+	// wg tracks the job goroutines dispatched by runScheduled so Stop can join
+	// them instead of leaving them to log/record metrics after their owner
+	// (an App, or a test's container) has been torn down.
+	wg sync.WaitGroup
 
 	done chan struct{}
 	once sync.Once
@@ -86,30 +94,78 @@ func NewCron(cntnr *container.Container) *Crontab {
 	return c
 }
 
+// Stop halts the scheduler and blocks until every job goroutine it already
+// dispatched has returned. Joining matters as much as halting: a job that fired
+// on the tick just before Stop keeps logging and recording metrics against a
+// container that the caller is about to tear down.
 func (c *Crontab) Stop() {
+	c.stop(nil)
+}
+
+// stop is Stop with an optional abort channel for the join, so an application
+// shutdown is never held open past its own deadline by a long-running job.
+func (c *Crontab) stop(abort <-chan struct{}) {
 	c.once.Do(func() {
 		c.ticker.Stop()
+
+		c.mu.Lock()
+		c.stopped = true
+		c.mu.Unlock()
+
 		close(c.done)
 	})
+
+	joined := make(chan struct{})
+
+	go func() {
+		c.wg.Wait()
+		close(joined)
+	}()
+
+	select {
+	case <-joined:
+	case <-abort:
+	}
 }
 
 func (c *Crontab) runScheduled(t time.Time) {
+	// The lock is held across dispatch so that Stop cannot slip between the
+	// stopped check and wg.Add. Dispatch only spawns goroutines, and a job that
+	// calls back into AddJob does so from its own goroutine, so this cannot
+	// deadlock.
 	c.mu.Lock()
+	defer c.mu.Unlock()
 
-	n := len(c.jobs)
-	jb := make([]*job, n)
-	copy(jb, c.jobs)
+	if c.stopped {
+		return
+	}
 
-	c.mu.Unlock()
+	tk := getTick(t)
 
-	for _, j := range jb {
-		if j.tick(getTick(t)) {
-			go j.run(c.container)
+	for _, j := range c.jobs {
+		if !j.tick(tk) {
+			continue
 		}
+
+		c.wg.Add(1)
+
+		go func(j *job) {
+			defer c.wg.Done()
+
+			j.run(c.container)
+		}(j)
 	}
 }
 
 func (j *job) run(cntnr *container.Container) {
+	// A job runs on its own goroutine, detached from whoever scheduled it, so a
+	// container that is nil or half-built (no logger) has to be handled here
+	// rather than assumed away — every log below would otherwise nil-panic on a
+	// background goroutine and take the whole process down.
+	if cntnr == nil || cntnr.Logger == nil {
+		return
+	}
+
 	ctx, span := otel.GetTracerProvider().Tracer("gofr-"+version.Framework).
 		Start(context.Background(), j.name)
 	defer span.End()
--- a/pkg/gofr/cron_test.go
+++ b/pkg/gofr/cron_test.go
@@ -2,6 +2,7 @@ package gofr
 
 import (
 	"fmt"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -771,7 +772,8 @@ func TestCronTab_runScheduled_NilLoggerAndContainer(t *testing.T) {
 
 func TestCrontab_Stop_JoinsInFlightJobs(t *testing.T) {
 	release := make(chan struct{})
-	finished := make(chan struct{})
+
+	var runs atomic.Int64
 
 	c := &Crontab{
 		ticker:    time.NewTicker(time.Second),
@@ -786,8 +788,8 @@ func TestCrontab_Stop_JoinsInFlightJobs(t *testing.T) {
 			dayOfWeek: map[int]struct{}{1: {}},
 			name:      "slow-job",
 			fn: func(*Context) {
+				runs.Add(1)
 				<-release
-				close(finished)
 			},
 		}},
 	}
@@ -811,14 +813,18 @@ func TestCrontab_Stop_JoinsInFlightJobs(t *testing.T) {
 
 	select {
 	case <-stopped:
-		<-finished
 	case <-time.After(time.Second):
 		t.Fatal("Stop did not return after the job finished")
 	}
 
-	// A tick that lands after Stop must not dispatch anything more.
+	require.Equal(t, int64(1), runs.Load(), "Stop must return only after the dispatched job has run")
+
+	// A tick that lands after Stop must not dispatch anything more. The second Stop joins
+	// whatever that tick dispatched, so the count below is read after any such job has finished.
 	c.runScheduled(time.Date(2024, 1, 1, 1, 1, 1, 1, time.Local))
 	c.Stop()
+
+	require.Equal(t, int64(1), runs.Load(), "a tick after Stop must not dispatch a job")
 }
 
 func TestCrontab_Stop(t *testing.T) {
--- a/pkg/gofr/cron_test.go
+++ b/pkg/gofr/cron_test.go
@@ -214,15 +214,10 @@ func TestCronTab_AddJob(t *testing.T) {
 	mocks.Metrics.EXPECT().NewCounter("app_cron_job_success", gomock.Any()).AnyTimes()
 	mocks.Metrics.EXPECT().NewCounter("app_cron_job_failures", gomock.Any()).AnyTimes()
 
-	// These AnyTimes() runtime expectations are load-bearing, not merely
-	// defensive: Stop() below halts only *future* ticks, it does not join a job
-	// goroutine that a tick already dispatched (Crontab fires `go j.run(...)` with
-	// no WaitGroup). The "* * * * *" job fires at second 0 of every minute, so a
-	// job dispatched just before Stop can still call the metrics mock after the
-	// test returns; without these expectations that call is gomock's
-	// t.Fatalf-on-finished-test -> panic. `defer c.Stop()` narrows the window but
-	// cannot close it — deterministic joining needs a framework change to Stop
-	// (tracked in gofr-dev/gofr#3801). Do not remove these believing Stop covers it.
+	// Stop() now joins the job goroutines a tick already dispatched, so a job can
+	// no longer reach the metrics mock after the test returns. These AnyTimes()
+	// expectations remain because the "* * * * *" job can legitimately fire at
+	// second 0 of a minute *during* the test, and the count is timing-dependent.
 	mocks.Metrics.EXPECT().IncrementCounter(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
 	mocks.Metrics.EXPECT().RecordHistogram(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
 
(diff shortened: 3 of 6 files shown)


aryanmehrotra (changes requested):
Reviewed at `7db71631`. Everything the three tickets ask for is here, and I confirmed the user-visible failure rather than taking the flaky-test framing at face value. One change requested, on a single test assertion.

## The failure this closes is data loss, not test hygiene

`Shutdown` closes datasources *after* stopping cron, and before this change `Stop()` returned as soon as the ticker halted:

```go
a.cron.Stop()          // returned immediately
...
a.container.Close()    // closes SQL, Redis, pub/sub
```

Built a real app against MySQL with a job still working when `Shutdown` is called:

| | `development` | `7db71631` |
|---|---|---|
| `Shutdown` returned after | **0s** | 110ms — it waited |
| the job's next query | **`sql: database is closed`** | `<nil>` |

A half-written batch loses its remaining work, and the error lands after the log file is closed too. That belongs in the description — "join in-flight jobs" reads like tidiness.

The nil-logger crash also reproduces directly, without waiting on a minute boundary. Calling `job.run` with a logger-less container:

```
development : panic: SIGSEGV -> logging.(*ContextLogger).Infof   (takes the process down)
7db71631    : ok
```

Both new tests pass under `-race -count=3`.

## Against #3801's own fix direction

| asked | delivered |
|---|---|
| `WaitGroup` around `go j.run(...)` | yes |
| join bounded by the context | yes, via the `abort` channel |
| wait, else force-return on timeout | yes — capped by `SHUTDOWN_TIMEOUT` |
| wire the shutdown context from `App.Shutdown` | yes |
| give `Stop(ctx context.Context)` a context | deliberately not — and I think you are right |

On that last one: `Stop(ctx)` would be a breaking signature change to a public method for a parameter only `Shutdown` needs. Keeping `Stop()` and adding an unexported `stop(abort)` gets the same bound without the API cost. Worth keeping the note in the description explaining it, because it reads as a deviation from the issue until you see the reasoning.

I also checked the `Test_SQLRetryConnectionInfoLog` de-flake for whether `AnyTimes()` was hiding the problem rather than fixing it. It is not: the fix is `t.Cleanup(func() { db.Close() })`, which stops the retry and metrics goroutines so nothing can reach the controller after the test, and the fixed `time.Sleep(100ms)` is replaced by a logger that signals on the actual line. The assertion got stronger, and the two gauges genuinely are outside the test's control — `pushDBMetrics` re-emits them every 10s from its own goroutine.

## The change I am asking for

The description says the new test covers *"a post-`Stop` tick dispatches nothing."* It does not. The test ends:

```go
// A tick that lands after Stop must not dispatch anything more.
c.runScheduled(time.Date(2024, 1, 1, 1, 1, 1, 1, time.Local))
c.Stop()
```

Two calls, a comment stating an invariant, no assertion. I removed each of the three guards in turn and re-ran the new tests:

| mutation | result |
|---|---|
| remove the `wg` join from `stop()` | **FAIL** — caught |
| remove the nil guard in `job.run` | **FAIL** — caught |
| **remove the `stopped` check in `runScheduled`** | **passes** — not caught |

I re-ran the third with the mutated line printed back out of the file, because a patch that silently fails to match is indistinguishable from a surviving mutant.

The reason a second dispatch stays invisible is `job.run`'s own `defer`: the re-dispatched job re-enters `fn`, hits `close(finished)` on an already-closed channel, and the `recover()` there logs the panic and carries on.

This is worth fixing rather than noting, because `stopped` is what makes reusing the `WaitGroup` safe. `runScheduled` calls `wg.Add` while holding `mu`, and `stop` sets `stopped` under the same `mu` before waiting — that ordering is the whole argument that no `Add` can land after `Wait` has begun. It is the subtlest part of the change and the only part with no coverage.

Counting runs instead of closing a channel is enough:

```go
var runs atomic.Int64
// job fn: func(*Context) { runs.Add(1) }

c.runScheduled(tick)
c.Stop()
require.Equal(t, int64(1), runs.Load())

c.runScheduled(tick) // must dispatch nothing
c.Stop()             // joins anything that was dispatched
require.Equal(t, int64(1), runs.Load())
```

Deterministic, no sleeps — the second `Stop` joins before the assertion. Verified: passes on this head, fails `expected: 1 / actual: 2` with the `stopped` check removed.

## One thing I checked and found correct

`stop(nil)` relies on a nil channel blocking forever in `select`, so the public `Stop()` waits indefinitely while `Shutdown` stays bounded. That is correct, and the comment says so.

Holding `mu` across dispatch also checks out — dispatch only spawns goroutines, and a job calling `AddJob` does so from its own goroutine, so the reentrancy the comment rules out really is ruled out.

Note for whoever merges: this closes the cron half of #3801, so #3801 itself should stay open for the server half.

akshat-kumar-singhal:
Thanks — the mutation testing on the `stopped` check was the right catch, and the explanation (the re-dispatched job re-enters `fn`, panics on the already-closed channel, and `job.run`'s own `recover()` logs and carries on) is exactly why the old form could not fail.

Fixed in e87f882f, using the run-counter shape you suggested:

- `fn` increments an `atomic.Int64` instead of closing a channel, so the job body cannot be what hides a second dispatch.
- `require.Equal(t, int64(1), runs.Load())` after the first `Stop` (the join really did complete the job), and again after the post-`Stop` tick plus a second `Stop` — that second `Stop` joins anything dispatched, so the read is deterministic with no sleeps.

Verified both directions: `go test ./pkg/gofr/ -run Cron -race -count=3` passes on this head, and with the `stopped` check removed from `runScheduled` the test fails `expected: 1 / actual: 2`. `cron.go` is untouched by the commit — the diff is `cron_test.go` only.

Description updated on both points: it now leads with the shutdown data loss (`Shutdown` stops cron before `container.Close()`, so an unjoined job's next query hit `sql: database is closed` and a half-written batch lost its remaining work) rather than framing the change as test hygiene, and it notes that #3801 stays open for the server half.

aryanmehrotra (changes requested):
The code is done — the assertion I asked for is there and it catches the mutant. One thing left before this lands: the change makes a documented statement false, and the PR touches no docs.

## `docs/guides/graceful-shutdown/page.md:90` now says the opposite of what the code does

> **Cron jobs.** GoFr's `App.Shutdown` drains HTTP, gRPC, and metrics servers and closes datasource connections — it does **not** stop the cron scheduler or wait for in-flight cron tasks. Cron jobs run with `context.Background()`, so they continue past SIGTERM and may be cut off when the container is killed at `terminationGracePeriodSeconds`. If you have long-running cron work that must finish, run it as a separate Kubernetes `Job` triggered by a `CronJob` resource instead of inside the same pod, so the pod's lifecycle doesn't interrupt it.

Clause by clause against `e87f882f`:

| claim | before this PR | now |
|---|---|---|
| "does not stop the cron scheduler" | **already false** — `Shutdown` called `a.cron.Stop()`, which halted the ticker | false |
| "or wait for in-flight cron tasks" | true | **false — this PR is exactly this change** |
| "run with `context.Background()`" | true | still true (`cron.go:170`) |
| "continue past SIGTERM … cut off at `terminationGracePeriodSeconds`" | true | now joined, and bounded by `SHUTDOWN_GRACE_PERIOD` instead |
| "run it as a separate `CronJob`" | reasonable advice | still reasonable, but the stated reason no longer holds |

The middle row is the one this PR owns. I confirmed it on a running application rather than from the diff: real app, real MySQL, a job mid-write when `Shutdown` is called — on `development` `Shutdown` returned in 0s and the job's next query got `sql: database is closed`; at this head it waited 110ms and the query returned `<nil>`. That is precisely "waits for in-flight cron tasks."

Suggested replacement, matching the behaviour I measured:

> - **Cron jobs.** `App.Shutdown` stops the cron scheduler and waits for jobs that are already running before it closes datasource connections, so a job mid-write is not cut off by `container.Close()`. The wait is bounded by the shutdown deadline (`SHUTDOWN_GRACE_PERIOD`): if a job is still running when that expires, `Shutdown` returns and the job is abandoned. The jobs themselves run with `context.Background()`, so a job's own `ctx` is never cancelled — anything that needs to be interruptible has to watch something else. For cron work that routinely outlasts `SHUTDOWN_GRACE_PERIOD`, run it as a Kubernetes `CronJob` in its own pod rather than inside the service.

I also verified the bound on the same app: a job sleeping 10s with a 1s shutdown deadline returned in 1s, so the new join cannot become a way to hang a pod. Worth stating in the doc, since it is the obvious worry once `Stop` blocks.

## Two adjacent errors in the same file, if you are editing it anyway

The step list at `:29-33` does not match `Shutdown`:

| doc | code (`gofr.go:97-133`) |
|---|---|
| 1 `httpServer.Shutdown` | 1 `httpServer.Shutdown` |
| 2 `grpcServer.Shutdown` | 2 `grpcServer.Shutdown` |
| 3 **`container.Close()`** | 3 **`cron.stop(ctx.Done())`** |
| 4 `metricServer.Shutdown` | 4 `metricServer.Shutdown` |
| 5 logger close | 5 `mcpServer.Shutdown` |
| | 6 `container.ShutdownMetrics` + **`container.Close()`** |
| | 7 logger close |

Cron and the MCP server are missing, and `container.Close()` is listed third when it is second-to-last. That last one matters more than a numbering nit: the ordering — cron stopped *before* datasources close — is the entire reason this PR fixes anything, and the doc currently describes the opposite order. Both are pre-existing, but they are in the paragraph you are already touching.

`:12` has the same gap in prose ("calls `Shutdown` on the HTTP, gRPC, and metrics servers and `Close` on the container's datasource connections") — no mention of cron.

## Correction to my own earlier comment

I wrote that the join is capped by `SHUTDOWN_TIMEOUT`. It is `SHUTDOWN_GRACE_PERIOD` (`shutdown.go:37`, default `30s`); `SHUTDOWN_TIMEOUT` does not exist anywhere in the repo. Flagging it so nobody goes looking for it.

## Sweep

One line in `docs/` couples cron to shutdown, and it is `:90`. I checked by intersecting every cron mention across `docs/` with every shutdown/SIGTERM/terminate/stop/drain/in-flight mention, in both directions, rather than grepping one phrase — a narrow grep is how I missed four lines on #3807.

Off-scope but noting it: `docs/advanced-guide/using-cron/page.md:39` says "the minimum time difference between cron job's two consecutive runs is a minute", which contradicts the optional `second` field documented ten lines above it and used in the page's own example at `:77`. Pre-existing and unrelated to this PR — separate issue if you agree.

Everything else is green: build, vet, `golangci-lint --new-from-rev=origin/development` → 0 issues, full `pkg/gofr` and `datasource/sql` pass, CI 14/14, all three guards now caught by mutation, and six adversarial concurrency probes clean under `-race`. Happy to approve as soon as `:90` is updated.

akshat-kumar-singhal:
Docs updated in `0007be1e`. All three points addressed, plus the `:12` prose.

**`:90` — the bullet you asked for.** Taken as written, verbatim.

**`:29-33` — step list.** Now matches `gofr.go:97-133`:

| # | doc (was) | doc (now) |
|---|---|---|
| 1 | `httpServer.Shutdown` | `httpServer.Shutdown` |
| 2 | `grpcServer.Shutdown` | `grpcServer.Shutdown` |
| 3 | **`container.Close()`** | **cron stop** — halts the scheduler, joins running jobs, bounded by the shutdown deadline |
| 4 | `metricServer.Shutdown` | `metricServer.Shutdown` |
| 5 | logger close | `mcpServer.Shutdown` |
| 6 | — | `container.ShutdownMetrics` + **`container.Close()`** |
| 7 | — | logger close |

I also added a sentence on *why* `Close()` is second-to-last, since that was your actual point rather than the numbering: every producer of datasource traffic — handlers, streams, cron jobs — is drained before the connections it uses are torn down. A reader who only had the old list would have concluded the opposite ordering, which is the ordering this PR depends on.

The stale `pkg/gofr/gofr.go:96-114` anchor on `:27` is dropped rather than renumbered — it had already drifted once and the file link alone is enough.

**`:12`.** The `{% answer %}` block now reads "…on the HTTP, gRPC, and metrics servers, stops the cron scheduler and waits for jobs already running, then calls `Close` on the container's datasource connections."

**On the `SHUTDOWN_GRACE_PERIOD` correction.** Confirmed independently — `shutdown.go:37` reads `SHUTDOWN_GRACE_PERIOD` with a `30s` default, and `SHUTDOWN_TIMEOUT` appears nowhere in the repo. Thanks for catching your own line; the doc uses the correct name.

To be precise about what I verified for this commit: I checked every claim in the new text against the code (`gofr.go:107-111` for the cron step's position, `:123` for `container.Close()`, `:118` for MCP, `cron.go:170` for `context.Background()`, `shutdown.go:37` for the grace period). I did not re-run your MySQL measurements — the doc wording relies on your 110ms / 1s-deadline results for the *observable* behaviour, and the code paths are consistent with both.

**Your off-scope find.** Agreed, and it reproduces: `docs/advanced-guide/using-cron/page.md:39` says the minimum gap between runs is a minute, while `:29` documents the optional `second` field and the page's own example at `:77` is `*/10 * * * * *`. Unrelated to this PR — I'll file it separately rather than widen this one.

akshat-kumar-singhal:
Filed your off-scope find as #3876 — `using-cron/page.md:39` vs the `second` field at `:29` and the `*/10 * * * * *` example at `:77`.

I checked which side of the contradiction is wrong before writing it up: `:39` is the stale sentence. `NewCron` ticks every second (`cron.go:73`), `schedulePartsWithSecond = 6` is a supported schedule length (`cron.go:23`), and `job` carries a `sec` field bounded by `seconds = 59`. So the sub-minute docs describe real behaviour and `:39` looks like it predates the `second` field. The issue proposes replacement wording; docs-only, no code change implied.

Kept out of this PR as you suggested.

aryanmehrotra (dismissed):
Approving at `f974fb0d`. The docs line is fixed, and you took the two things I had flagged as pre-existing rather than leaving them.

## The docs, checked against the code

The step list now matches `Shutdown` exactly — all seven steps, in order: http, grpc, cron, metrics, mcp, `ShutdownMetrics` + `Close`, logger. I verified it against `gofr.go:97-133` rather than against the table in my own comment.

Two things you did that I did not ask for and that both improve it:

- Dropping the `gofr.go:96-114` anchor from the doc link. Hardcoded line numbers in docs rot silently, and that one was already wrong.
- "It runs second-to-last on purpose: every producer of datasource traffic — handlers, streams, cron jobs — is drained before the connections they use are torn down." That is the reason the ordering exists, and my review only implied it.

Re-swept the rest of `docs/` rather than trusting the one line: I read every cron mention across all 23 files that have one, not just the shutdown-coupled ones. The remainder are feature listings (`faq`, `why-gofr`, the five `comparison` pages, the eight `migrate` pages), a KEDA note in `horizontal-pod-autoscaler:105`, and a JWKS note in `auth-in-kubernetes:76` — none of them assert anything about shutdown. `Stop()`'s own godoc already documents that it blocks, which matters more than the guide since it is the exported API.

## Re-verified at this head, not carried forward

The merge brought `#3807` in, so I re-ran everything rather than reusing the earlier results. Your cron files are byte-identical to `e87f882f` — `diff -- cron.go cron_test.go gofr.go gofr_test.go sql_test.go` is empty — but the run is at `f974fb0d`.

Real app, real MySQL, a job mid-write when `Shutdown` is called:

| | `development` (`c0b69322`) | `f974fb0d` |
|---|---|---|
| `Shutdown` returned after | 0s | 110ms |
| the job's query after shutdown | `sql: database is closed` | `<nil>` |

And the bound, same app — a job sleeping 10s against a 1s shutdown deadline returned in 1s. So the join cannot become a new way to hang a pod, which is the real worry once `Stop` blocks.

The mutation still bites: removing the `stopped` check in `runScheduled` fails `TestCrontab_Stop_JoinsInFlightJobs` with `expected: 1 / actual: 2`. Deterministic at `-cpu=1`, `-cpu=2` and `-cpu=8`. Six adversarial probes on the WaitGroup-reuse ordering — 16 concurrent `Stop()`s, a tick racing `Stop` over 200 rounds, `AddJob` after `Stop`, `Stop` with nothing in flight, `stop(abort)` pre-closed, and a panicking job — all clean under `-race`, with the racing probe winning 83/200 rounds so it is not vacuously passing.

`golangci-lint --new-from-rev=origin/development` reports 0 issues. Build and vet clean. CI 14/14.

## One pre-existing thing this surfaced, unrelated to your change

`go test ./pkg/gofr/ -race` is red on `development` today. Three tests fail: `TestContext_WriteMessageToService`, `TestGRPC_ServerShutdown`, `TestGRPC_ServerShutdown_ContextCanceled`. The failing set is identical on `development` and this head, and no race stack touches cron, so nothing here caused it.

The raced variable is `g.server` — written at `grpc.go:127` on the serve goroutine, read at `grpc.go:187` in `Shutdown`. That is exactly the gap `#3817` left behind: it mutex-guarded the HTTP, metrics and MCP servers and not gRPC. Worth its own issue.

Locally I also had to skip `TestGraphQL_*`, `TestEnableMCP_*`, `TestContext_WriteMessageToSocket` and `TestStartGRPCServer_Registered` — a process on my machine holds `:8000` and `:8200` and those bind fixed ports. They `FATAL` rather than fail, which takes the whole binary down and hides every test after them; that cost me four passes to identify and is also worth a separate issue.

## Notes for merge

- `#3801` stays open for the server half, as the description says.
- `Stop()` now blocking is a visible change to an exported method. It needs to reach the release notes, not just the PR body — there is no CHANGELOG in the repo, so it has to be added when v1.59.0 is cut.
- The main cron page (`docs/advanced-guide/using-cron/page.md`) still says nothing about shutdown, and in particular does not mention that a job's own `ctx` is never cancelled. Pre-existing and out of scope here, but it is the page a user reads first — separate follow-up.

Umang01-hash (dismissed):

aryanmehrotra (approved):
Re-approving at `ba7e033c`.

Both approvals here were dismissed by vikash on 11 Aug with "The merge-base changed after approval" - the `development` history rewrite, not anything on the merits. The force-push destroyed `f974fb0d`, so I could not diff the rebase against what I approved and did not carry the old verdict forward. This is a fresh review, and the end-to-end evidence below was re-run at this head rather than reused.

## The data loss reproduces, and the fix closes it

A real app: `gofr.New()`, a `* * * * * *` job that works for 2s then runs `ctx.SQL.QueryRowContext(ctx, "SELECT 1")`, `go app.Run()`, then `app.Shutdown(ctx)` fired while the job is deliberately mid-flight. Fresh `mysql:8.2.0`, same binary and config against both trees, three runs each:

| | `development` (`ceb50390`) | this branch |
| --- | --- | --- |
| `Shutdown` returned after | **0s** x3 | **1.8s** x3 |
| the job's query | **`sql: database is closed`** x3 | **`<nil>`** x3 |

Zero variance. This is not test hygiene - a cron job mid-write when SIGTERM lands loses its query today, and the error surfaces after the logger is closed, so nothing records it.

## The deadline genuinely bounds the join

I checked this half too, because an unbounded join would be a worse bug than the one being fixed:

| | result |
| --- | --- |
| 10s job, 30s deadline | `Shutdown` waited 9.8s, query `<nil>` |
| 10s job, **1s** deadline | `Shutdown` returned at **exactly 1.0s**, job abandoned, query `sql: database is closed` |

A long cron job cannot hold shutdown open past `SHUTDOWN_GRACE_PERIOD`, and the abandoned-job case behaves exactly as the new docs paragraph describes it.

## The Add/Wait ordering is right

`runScheduled` checks `stopped` and calls `wg.Add(1)` under the same lock hold, and `stop` sets `stopped` under `mu` before it waits - so `Wait` can never race an `Add` that resurrects the counter from zero. That is the failure mode this shape exists to prevent, and it is closed structurally rather than by timing. `Stop()` passing a nil `abort` correctly degrades the select to "wait only on `joined`", and the orphaned waiter goroutine on the abort path terminates itself once the job returns.

## No breaking change

`go doc -all ./pkg/gofr` is byte-identical between `ceb50390` and this head - 112 exported symbols on both sides, nothing added, removed or re-signatured. The one behavioural change is that `Crontab.Stop()` now blocks, which its godoc states. `NewCron`, `Crontab` and `Stop` appear nowhere in `docs/`; users reach cron through `App.AddCronJob`, where the shutdown deadline bounds the wait.

## All three mutants die

| mutant | caught by |
| --- | --- |
| drop `if c.stopped { return }` | `a tick after Stop must not dispatch a job` - expected 1, actual 2 |
| make `stop` return without joining | `Stop returned while a job was still running` |
| drop the `cntnr.Logger == nil` guard | `panic: SIGSEGV` - the #3813 crash, unrecovered |

`cron|Crontab -race -count=3` and `Test_SQLRetryConnectionInfoLog -race -count=3` both ok here.

## Docs and the SQL half

The seven-step list matches `Shutdown` exactly - http, grpc, cron, metrics, mcp, `ShutdownMetrics` + `Close`, logger. The caveat you added is the one that matters: jobs still run on `context.Background()`, so a job's own ctx is never cancelled and anything interruptible has to watch something else. Better stated here than discovered in production.

The `sql_test.go` half is the same move as #3821 - a logger that closes a channel on the awaited line, replacing `StdoutOutputForFunc` + a 100ms sleep, with `AnyTimes()` on the two gauges `pushDBMetrics` emits from its own goroutine every 10s. Correct diagnosis of that flake.

One observation, no change needed: `container.Metrics()` dereferences its receiver, so `NewCron(nil)` panics in `registerMetrics` before any job runs - the `cntnr == nil` half of the guard is reachable only from a hand-built `Crontab`. It costs nothing, and the `Logger == nil` half is the live one.

Note for whoever merges: Umang's approval was dismissed by the same rewrite and has not been re-given.

PR #3817, opened 2026-07-31
fix(examples): make using-migrations test self-contained (#3816)

### LINKED ISSUE #3816: examples/using-migrations test is not self-contained: migration state persists in Redis between runs

akshat-kumar-singhal:
**Description:**

Fixes #3816.

`examples/using-migrations` passed once against a fresh Redis/MySQL pair and failed on every run
after that, because it owns no namespace in either datastore:

- The migration bookkeeping is keyed by version and lives in **both** datastores (`gofr_migrations`
  hash in Redis, `gofr_migrations` table in MySQL), and `redisMigrator.getLastMigration` returns
  `max(redis, sql)` — so a stale Redis record alone is enough to skip the schema-creating migration.
  With Redis carried over and MySQL recreated, the run logs `no new migrations to run` and every
  request fails with `Table 'test.employee' doesn't exist`.
- The row the test POSTs (`id: 2`) and the row the migration inserts survive in MySQL, so a re-run
  collides: `expected 201, actual 500` / `Duplicate entry '2' for key 'employee.PRIMARY'`.

CI never sees this — every job gets fresh service containers, so every CI run is a first run.

`TestMain` now creates (and re-creates) a `test_using_migrations` database the test owns, points the
example at it via `DB_NAME`, and deletes the three Redis keys the example writes
(`gofr_migrations`, `gofr_migrations_lock`, `Umang`). `configs/.env` is untouched — it is what a
human runs the example with, and `gofr.New()` honours the override because `godotenv.Load` does not
replace variables already set in the environment.

While checking the sibling examples the issue asks about, a related coupling turned up: every MySQL
example migrates into the same `test` database and therefore shares one `gofr_migrations` table.
`using-migrations`' versions (`1722507126`, `1722507180`) are higher than
`using-add-rest-handlers`' (`1721816030`), so once `using-migrations` has run, the sibling's
migration is considered applied and its `user` table is never created. It only works today because
`go test` happens to run the alphabetically earlier package first on a fresh database. Giving
`using-migrations` its own database removes that coupling as well.

That sibling also failed on any genuinely fresh database for a second reason: with migrations
actually running at startup, `time.Sleep(100 * time.Millisecond)` is not long enough for the server
to be listening. Both tests now wait on `/.well-known/alive`, matching what
`examples/using-graphql` already does.

`using-graphql` was checked and is unaffected — its test builds its own app with stubbed resolvers
and never calls `Migrate`. `using-publisher` / `using-subscriber` migrate Kafka topics, not a shared
table.

**Verification** (Go 1.26, `redis:7.0.5` and `mysql:8.2.0` on the ports the `Example-Unit-Testing`
job uses):

| | before | after |
|---|---|---|
| `using-migrations`, run 1 | ok | ok |
| `using-migrations`, runs 2 and 3 against the same containers | FAIL | ok |
| `using-migrations` with Redis kept, MySQL recreated | FAIL (`employee` missing) | ok |
| `using-add-rest-handlers` on a fresh database | FAIL (server not up) | ok |

`gofmt`, `go vet` and `golangci-lint` are clean on both packages.

None — the change is confined to the two example test files. No framework code and no example
config is touched.

**Additional Information:**

The test uses `github.com/go-sql-driver/mysql` and `github.com/redis/go-redis/v9` directly for the
setup, both already direct dependencies of the module, and reads the connection details through
`gofr.dev/pkg/gofr/config` rather than hardcoding them.

**Checklist:**


aryanmehrotra (approved):
Reviewed end to end at `62ef7293`. Reproduced every failure the description claims, on the CI images, and confirmed each one is closed at this head. Approving. Three notes, none blocking.

## The bug reproduced

`mysql:8.2.0` on `:2001` and `redis:7.0.5` on `:2002` — the images and ports the `Example-Unit-Testing` job uses. On `development`:

| scenario | result |
|---|---|
| run 1, fresh containers | `ok` |
| run 2, same containers | **FAIL** — `Duplicate entry '2' for key 'employee.PRIMARY'`, 201 got 500 |
| MySQL recreated, Redis kept | **FAIL** — `no new migrations to run`, then `Table 'test.employee' doesn't exist` ×4 |
| `using-add-rest-handlers` after `using-migrations` ran | **FAIL** — `Table 'test.user' doesn't exist` |
| `using-add-rest-handlers` alone on a fresh `test`, 3 attempts | **FAIL, FAIL, ok** — `connection refused`, the 100 ms sleep |

So all four are real, including both halves of the sibling problem the description raises in passing.

I checked the mechanism in the code rather than taking it from the description, because "highest version wins" is the part that makes the coupling non-obvious: `redisMigrator.getLastMigration` returns `max(redis, sql)`, `runMigrations` skips anything `<= lastMigration`, and `commitMigration` only records the datasources a migration actually touched. That last detail is why the SQL table holds `1722507126` and the Redis hash holds `1722507180` — each migration is recorded in one store only, and either one alone is enough to suppress both.

## At this head

| | |
|---|---|
| `using-migrations` ×3, same containers | `ok ok ok` |
| `using-add-rest-handlers` on shared `test` ×3 | `ok ok ok` |
| `using-add-rest-handlers` on fresh `test` ×3 | `ok ok ok` |
| tables in shared `test` after 3 runs | **none** — the example no longer writes there at all |

`gofmt`, `go vet`, `golangci-lint 2.12.2` → 0 issues. Merges into current `development` with no conflict.

I also built the merged tree (`development` + this head) and ran it the way CI does — one `go test -short` invocation, packages in parallel — since #3814 means a failing example now fails the job:

```
pass 1, freshly created containers : ok ×5
pass 2, same containers            : ok ×5
pass 3, same containers            : ok ×5
```

`development` cannot do pass 2. That is the whole point of the change.

One correction to the description's reasoning, in the safe direction: the `DB_NAME` override does not rest only on `godotenv.Load` declining to overwrite. `EnvLoader.read` also snapshots the process environment and re-applies it *after* `godotenv.Overload`, so the override would survive even a `configs/.local.env` that set `DB_NAME`. Worth knowing, because `Overload` genuinely does overwrite and the stated reason alone would not cover it.

## Three notes

**1. #3816 was closed by mistake and should be reopened.** It was closed as completed at `2026-08-07T11:28:00Z` — the same second #3814 merged. Not a keyword auto-close: #3814's body says only `Fixes #3812`, and states in writing that *"`examples/using-migrations`, whose fix lives in #3817"*. `git log examples/using-migrations/main_test.go` on `development` stops in 2023, and I reproduced the run-2 failure there today. This PR's `Fixes #3816` will re-close it on merge, so nothing needs doing if it merges soon — flagging it so it doesn't read as already-handled to anyone triaging.

**2. The coupling this PR diagnoses is still live for the sibling, and I broke it at this head.** `using-add-rest-handlers` still migrates into shared `test`. `using-graphql`'s `main` calls `Migrate` with version `20240205125300` — five orders of magnitude above the sibling's `1721816030`. Ran `go run ./examples/using-graphql` once, then the sibling's test on this branch:

```
"no new migrations to run"  ->  Table 'test.user' doesn't exist  ->  FAIL
```

`using-migrations` beside it: `ok`. Its test is now immune because it owns a database; the sibling is not, because this PR gives it a readiness wait but not a namespace. Its test only survives today because nobody runs `using-graphql`'s `main` and CI never does. Not something to fix here — the description already makes the argument for the fix, and it applies to the sibling verbatim. Worth a follow-up issue so the reasoning isn't lost, since the next person to hit it will see a missing table and not suspect a different example.

**3. This is the fourth hand-rolled readiness wait in `examples/`.** `using-graphql` has `waitForReady`, `using-ai` has an inline loop, and this PR adds two identical `waitForServer` copies. Seventeen example tests are still on fixed sleeps, which is the same latent flake wherever migrations or datasource init are slow. `testutil` already owns the other half of this setup via `NewServerConfigs`, so a `WaitForHTTP` there would collapse four copies into one and give the remaining seventeen an obvious fix. Using `service.AlivePath` instead of a hardcoded string is the right instinct either way.

## On the checks

14 pass, 2 skipping — but that run is from Jul 31, and this head's `go.yml` has no `set -eo pipefail` in the Example job, which is exactly the false green #3814 removed. It needs an "Update branch" anyway; that re-runs the Example job under the real gate. The merged-tree run above is why I expect it to stay green.

akshat-kumar-singhal:
Folded into #3821 and closing this in favour of it.

The two files here need `testutil.WaitForHTTPServer`, which #3821 adds, so the change had to sit on top of it. Both PRs come from a fork, and GitHub only accepts a base branch that lives in the upstream repo — so stacking this on #3821's branch is not possible, and leaving the base at `development` would have shown all of #3821's files in this diff. #3821 now carries the `using-migrations` isolation fix as well and closes #3816.

PR #3820, opened 2026-07-31
fix(http): serve the root of a static endpoint instead of 403 or an endless redirect (#3819)

### LINKED ISSUE #2643: Bug: Double Slash `//` Path Causes POST Route to Resolve to GET Handler

### LINKED ISSUE #3819: Static file endpoint root is unreachable: 403, 404, or an endless redirect
**Describe the bug**

A request for the root of a static endpoint registered with `AddStaticFiles` never serves that
directory's index page. `examples/using-html-template` registers `app.AddStaticFiles("/", "./static")`
and answers `GET /` with **403 Forbidden** instead of `static/index.html` — its own
`Test_IndexHTML` fails on `development` today.

There are three independent causes, so fixing any one of them still leaves the root unreachable.

**1. Containment check rejects the served directory itself (403)**

`isRestrictedFile` treats the directory as an escape attempt:

```go
return !strings.HasPrefix(absPath, staticConfig.directoryName+string(os.PathSeparator)) ||
	fileName == DefaultSwaggerFileName
```

`filepath.Join(dir, "/")` is `dir`, with no trailing separator, so `HasPrefix(dir, dir+"/")` is
false and the request is refused.

The trailing separator arrived in #3041 (fixing #3040) to stop a sibling directory with a shared
prefix from escaping — `/app/public` must not admit `/app/publicother` — which it does correctly. It
just also excludes the one path that *is* the directory rather than something beside it. Requests
for files below the root are unaffected, which is why only the root shows the bug.

**2. The root of a non-`/` endpoint is not routed at all (404)**

`AddStaticFiles` registers `PathPrefix(endpoint + "/")`, but `Router.ServeHTTP` normalizes with
`path.Clean`, which drops a trailing slash. `GET /static/` therefore arrives as `/static` and matches
neither the prefix route nor anything else — the response is mux's default `404 page not found`,
not GoFr's static 404 handler.

**3. Directories redirect in a loop (301)**

`http.FileServer` answers a directory with a redirect to the trailing-slash form of the URL, and
answers `.../index.html` with a redirect back to `./`. `path.Clean` strips the trailing slash off
both, so the client is sent round a redirect it can never satisfy.

This one is reachable today independently of the root — any subdirectory of a static endpoint hits
it. With a static endpoint at `/static` and a `sub/index.html` beneath it:

```
GET /static/sub   -> 301  Location: sub/
GET /static/sub/  -> 301  Location: sub/
```

**Expected behavior**

The root of a static endpoint serves that directory's index file, and so does any subdirectory,
without redirecting. A directory with no index file returns 404 rather than a listing of its
contents — a static endpoint should not disclose the file names under it.

**Note on existing test expectations**

`TestStaticHandler` currently asserts `GET /static` returns 404. That encodes cause 2 — the root was
not routable — rather than a decision that the root should be withheld. `pkg/gofr/static` contains
an `index.html` (the bundled swagger-ui page, already reachable at `/static/index.html`), so once
the root is routable it serves that, and the assertion needs updating.

akshat-kumar-singhal:
**Description:**

Fixes #3819.

A request for the root of a static endpoint never served that directory's index page.
`examples/using-html-template` answers `GET /` with **403 Forbidden** instead of `static/index.html`
— its own `Test_IndexHTML` fails on `development` today.

This is **two regressions from two different PRs**, not one, and the endpoint root is only the most
visible symptom:

**#2759, in v1.52.0 — the redirects and the unrouted root.** That PR fixed #2643: mux's automatic
cleaning issued a 301 for `//hello`, and Postman turned the POST into a GET on the redirect. The
first implementation preserved trailing slashes; review suggested `path.Clean` from the standard
library, which is otherwise exactly right but *also* strips them. That single substitution is both
the unrouted `/static/` and the redirect loops. Before it, `/static/sub` → `301 sub/` → `200`
resolved correctly.

**#3041, in v1.54.5 — the 403.** `isRestrictedFile` classified the served directory itself as an
escape attempt: `filepath.Join(dir, "/")` is `dir`, with no trailing separator, so
`HasPrefix(dir, dir+"/")` is false. The separator is there to stop a sibling with a shared prefix
from escaping — `/app/public` must not admit `/app/publicother` — which it does correctly. It just
also excludes the one path that *is* the directory rather than something beside it. That guard is
not gratuitous: on v1.51.0 the sibling escape is genuinely open, so reverting is not on the table.
This PR is the first time both are right at once.

The two are independent, so fixing either alone still leaves the root unreachable.

**What changed:**

1. **Containment.** `isWithinDirectory` compares against the directory as well as against
   `dir + separator`. The sibling case keeps its test.

2. **Routing.** For a non-`/` endpoint the route was `PathPrefix("/static/")`, but `path.Clean`
   drops the trailing slash, so `/static/` arrives as `/static` and matched nothing. The bare
   endpoint is now registered as an exact path too, while the prefix route keeps its separator.

3. **Redirects.** Directories resolve to their index file, and the file is served with
   `http.ServeContent`, which is handed an open file rather than a URL and has no redirect rule at
   all. `http.ServeFile` is not enough, which is worth spelling out because it is the obvious
   choice: it carries its *own* `/index.html` → `./` rule. That rule misses at the endpoint root
   only by accident — `StripPrefix` leaves `index.html`, with no leading separator — and fires one
   directory deeper on `/static/sub/index.html`.

4. **Directory listings stay closed.** A directory with no index file is reported as not-existing
   and takes the ordinary 404 path. This is load-bearing rather than incidental: `GET /static/empty/`
   on v1.51.0 returns 200 and Go's `dirList` output. The redirect loop introduced in v1.52.0 made
   that unreachable by accident, so it cannot be seen on `development` — but it was live through
   v1.51.1, and re-routing the endpoint root would have brought it straight back.

5. **Only regular files are served**, and `openapi.json` is refused in any capitalization — the name
   is compared case-insensitively because the filesystem underneath may be, and on a
   case-insensitive one `/static/openapi.JSON` opens the very file an exact comparison refuses.

6. **Only `GET` and `HEAD`.** Every other method was answered with the file: `DELETE /static` gave
   200 and the content, which reads as "deleted, here it is". Now 405 with `Allow: GET, HEAD`.
   `.Methods()` on the routes instead would drop the request to the catch-all and answer 404, which
   is its own lie when GET on the same path is a 200, so the check is in the handler. This is a 405
   in a framework that 404s wrong-method everywhere else; GoFr's own 405 gap is a separate issue.

7. **Unreadable files answer 403, not 500.** `net/http.toHTTPError` maps `fs.ErrPermission` → 403,
   as do nginx and Apache; 500 tells a monitor the app is broken when what it has is a file mode.
   Both routes are covered — a real `EACCES` from `os.Stat`/`os.Open`, and `errReadPermissionDenied`,
   which now wraps `fs.ErrPermission`. Pre-existing, but `respondWithFileError` is rewritten here.

8. **A `dirName` that is not a directory is refused at registration.** `App.AddStaticFiles` only
   `os.Stat`s the path, so a regular file passes, and `absPath == directoryName` would now admit it
   where the old prefix test failed closed — `AddStaticFiles("data", "./config.yaml")` served the
   file at `/data`. Registration now checks `IsDir` explicitly.

`http.FileServer`/`http.Dir` is gone: `staticHandler` has already resolved the absolute path and
verified containment, so it serves that path directly instead of handing the URL back to a second
resolver. One serving path instead of two — and the open file's own `FileInfo` replaces
`validateFile`'s second `os.Stat`.

Per `CONTRIBUTING.md:134` this is a **minor**, not a patch. No exported API changes — every changed
signature is on the unexported `staticFileConfig` — but bare `GET /static` is new functionality,
which is clause 7.

Most of what changes is a restoration, so it is worth separating by kind:

**Restored** — broken in v1.54.5, 13 releases:

| | before | after |
|---|---|---|
| `GET /` (mounted at `/`, dir has `index.html`) | 403 | 200 `index.html` |

**Restored** — broken in v1.52.0 by the `path.Clean` substitution, 20 releases:

| | before | after |
|---|---|---|
| `GET /static/` | 404 | 200 `index.html` |
| `GET /static/sub` | 301 → `sub/`, forever | 200 |
| `GET /static/sub/` | 404 | 200 |
| `GET /static/index.html` | 301 → `./`, forever | 200 |

**New** — never worked in any release, so no working setup can break:

| | before | after |
|---|---|---|
| `GET /static` (bare) | 404 | 200 `index.html` |

**Closed** — live through v1.51.1, unreachable since only by accident:

| | before | after |
|---|---|---|
| `GET /static/empty/` | 200 + directory listing | 404 |

**Genuinely breaking.** Three, and after the `IsDir` guard I could find no others:

| | before | after |
|---|---|---|
| `GET /static`, no `index.html` | `404`, `application/json`, `{"error":{"message":"route not registered"}}` | `404`, `text/plain`, `404 Not Found` — or `text/html` and the folder's `404.html` |
| `POST`/`PUT`/`DELETE` on a static path | 200 + the file | 405, `Allow: GET, HEAD` |
| unreadable file or directory | 500 | 403 |

The first is the one to watch: the real change is `application/json` → text, which breaks a client
parsing the error envelope. That envelope comes from the catch-all at `gofr.go:183`; a bare
`NewRouter()` in a unit test gives mux's `text/plain` default instead, but a real app never sees it.

`TestStaticHandler` asserted `GET /static` returns 404. That encoded the root being unroutable
rather than a decision to withhold it, so the assertion is now 200. A companion case covers
`gofrTest`, whose directory has no index file and keeps its 404.

**Additional Information:**

No new dependencies. `docs/advanced-guide/serving-static-files/page.md` documents the endpoint root,
the subdirectory case, the no-index 404, the method restriction, and the relative-URL caveat below.

**Known limitation — relative URLs at the endpoint root.** `/static` and `/static/` return the same
bytes, so per RFC 3986 §5.1.3 a relative reference in that HTML resolves against a different base:
served at `/static`, `` requests `/logo.png`; served at `/static/`, it requests
`/static/logo.png`. nginx, Apache and `http.FileServer` all answer `/static` with a redirect to
`/static/` for exactly this reason.

This PR cannot. The redirect target `/static/` is cleaned back to `/static` before routing, so the
client loops — and for the same reason the two forms are not distinguishable here at all:

```
request "/static"  -> arrives at the router as "/static"
request "/static/" -> arrives at the router as "/static"
```

So serving the bare `/static` is not a separate decision that could be made differently. `/static/`
*is* `/static` by the time anything can route it, and 404-ing one 404s the other — which is the
v1.52.0 regression this PR exists to fix. The fix for the relative-URL trap is in the router's path
normalization, not here; the docs tell users to use root-relative URLs or a `<base href>` until
then.

Reproducer for the redirect loop, against a static endpoint at `/static` with `sub/index.html`
beneath it:

```
before:  GET /static/sub          -> 301  Location: sub/     (and again, forever)
after:   GET /static/sub          -> 200  <html>Sub</html>

before:  GET /static/index.html   -> 301  Location: ./
after:   GET /static/index.html   -> 200  <html>Index</html>
```

**Checklist:**


aryanmehrotra (changes requested):
Verified the whole matrix locally — the fix is right and the sibling protection from #3041 still holds. Answering the question you flagged: no, withholding the endpoint root was not deliberate. It's a regression. Before #3041 the check was `HasPrefix(absPath, dir)` with no separator, so the directory passed and `FileServer` served its index; `Test_IndexHTML` was added in #1554 asserting exactly that and passed until 6cf00652 landed on 2026-02-27. Serving the root is restoring intended behaviour, and 404-instead-of-a-listing for a directory without an index is the right call — that path was never reachable before, so it's a free choice and this is the safe one.

Worth noting *why* nobody caught a red test on `development` for five months: it's your #3812. `.github/workflows/go.yml:134` runs `go test ./examples/...` in a multi-line `command:` block with no `set -e`, so the block's exit status comes from the trailing `go tool cover -func`, which succeeds regardless. The `pkg/gofr` job below captures `exit_code` explicitly. Fixing #3812 is what stops this recurring.

The code itself is good — I'm requesting changes only for the two documentation items below, which I'd like in this PR rather than a follow-up. Then some non-blocking notes.

### 1. Docs need updating

`docs/advanced-guide/serving-static-files/page.md` documents only `/static/<filename>`, and its example tree has `public/index.html` — which this PR makes reachable at `/public`. Please add that the endpoint root serves the directory's `index.html`, and that a directory without one returns 404 (serving the folder's `404.html` if present) rather than listing its contents.

### 2. Widen the "Breaking Changes" section

Right now it only covers the flipped test assertion. There are two user-visible changes that belong in the release notes:

**The endpoint root now serves content where it used to refuse.**

| | before | after |
|---|---|---|
| `GET /` (mounted at `/`, has `index.html`) | 403 | 200 `index.html` |
| `GET /static` (has `index.html`) | 404 | 200 `index.html` |

Every app with a `static/` folder gets this with no opt-in via `factory.go:56`. I'd drop the "already reachable at `/static/index.html`, no new file is exposed" line — that URL was a redirect loop, so in practice this makes `index.html` in a static dir reachable for the first time. Still the correct behaviour, and it's a file the user deliberately published, but let's say it plainly rather than claim nothing changed.

**The no-index 404 changes shape.** Status is unchanged, body and content-type are not:

| | before | after |
|---|---|---|
| `GET /static`, no `index.html` | 404, `text/plain`, `404 page not found\n` (mux default) | 404, `text/html`, the app's `404.html` |

Arguably a second fix — GoFr's 404 handling now applies where mux's default was leaking through — but it breaks any user test asserting on the body.

### Non-blocking

- **`.../index.html` still redirects.** The description says directories are served "with `http.ServeFile`, which does not redirect" — true for directories, but an explicit `GET /static/index.html` still goes through `fileServer` and gets stdlib's `301 → ./`. It now terminates (`./` → `/static/` → cleaned to `/static` → 200) instead of looping, so the bug is fixed, but the claim is broader than what the code does. Worth a word so it doesn't mislead later.

- **Consider collapsing to one serving path.** `fileServer` is now only reached for plain files, and `staticHandler` has already resolved `absPath`. `http.ServeFile(w, r, absPath)` there behaves the same (including that redirect) and would let you drop the `fileServer` parameter, the `http.FileServer`/`http.Dir` construction, and the `resolvedPath != absPath` branch — one code path instead of two, and no re-resolving the URL against `http.Dir` after we already computed the absolute path.

- **The flipped assertion depends on the swagger bundle.** "check public endpoint" now expects 200, and that 200 comes from the checked-in `pkg/gofr/static/index.html`, because `createPublicDirectory` reuses the existing repo directory instead of creating a fresh one (`pkg/gofr/gofr_test.go:1294`). If that asset ever moves the test fails for an unrelated reason — writing its own `index.html` into a temp dir would be sturdier.

- **Merge-order note, nothing to do here.** #3759 (`GOFR_ROUTER` trie router) touches the same file. Different functions, so conflicts should be small, and precedence is safe since the index sorts trie hits and fallback by registration order. But this PR adds an *exact* `Path("/static")` route, which the trie indexes, while `PathPrefix("/static/")` goes to the fallback list — the first time one static endpoint is split across both. Whichever lands second should run `Test_StaticFileServing_EndpointRoot` with `GOFR_ROUTER` enabled; these tests only exercise the default mux path.

akshat-kumar-singhal:
Thanks — the #3041 history and the `Test_IndexHTML`/6cf00652 timeline settle the question I flagged, and the `go.yml:134` missing `set -e` explains the five silent months. I'll keep #3812 scoped to that.

Pushed 0772dbef. Both blocking items are in, and I took two of the three non-blocking notes.

**1. Docs.** `docs/advanced-guide/serving-static-files/page.md` now states that the endpoint root serves the directory's `index.html`, that a subdirectory root does the same, and that a directory without an index returns 404 — serving the folder's `404.html` — rather than listing its contents. The second example now says explicitly that `/public` serves `public/index.html` while `/static`, whose folder has no index, 404s.

**2. Breaking Changes.** Rewritten as three items: the endpoint root serving where it refused (your table), the changed 404 shape, and the flipped assertion. The "already reachable at `/static/index.html`, no new file is exposed" line is gone — you're right that it was a redirect loop, so the honest statement is that this makes it reachable for the first time.

**`.../index.html` still redirects — it doesn't any more, and that fell out of your next note.** I probed both ways rather than reasoning about it:

```
PR head:        GET /static/index.html -> 301  Location: ./
after collapse: GET /static/index.html -> 200  <html>Index</html>
```

`http.FileServer` prepends a `/` to the stripped path before its `HasSuffix(name, "/index.html")` check, so `index.html` matches and it redirects. `http.ServeFile` tests `r.URL.Path` as-is — after `StripPrefix("/static/")` that's `index.html`, no leading slash, no match. Pinned in a `Test_StaticFileServing_EndpointRoot` case.

**Collapsed to one serving path.** `http.FileServer`/`http.Dir`, the `fileServer` parameter and the `resolvedPath != absPath` branch are gone; the handler serves the path it already resolved and containment-checked. Description updated to match what the code does.

**The swagger-bundle dependence.** `setupTestEnvironment` registers a second endpoint from a `t.TempDir()` holding its own `index.html`, so the root-serves-index behaviour has an assertion that doesn't move if the bundled asset does. I kept the `/static` 200 case — it's the auto-registered default directory, which is worth covering — and folded a duplicate `gofrTest` 404 case into the existing one while I was there.

**#3759.** `GOFR_ROUTER` isn't in the tree yet, so there's nothing to run from here; noted for whichever lands second.

Verified locally: `pkg/gofr` and `pkg/gofr/http/...` pass, `examples/using-html-template` passes, `golangci-lint` clean on the touched files.

pkg/gofr/http/router_test.go
```diff
@@ -362,6 +362,112 @@ func Test_StaticFileServing_Static(t *testing.T) {
...
+			path:             "/static/sub",
+			staticServerPath: "/static",
+			expectedCode:     http.StatusOK,
+			expectedBody:     "<html>Sub</html>",
+		},
+		{
+			// An explicit request for the index file is served as the plain file it is.
+			// http.FileServer used to answer this with a redirect to "./", which path.Clean
+			// resolved back to the directory — the other half of the redirect loop.
+			name: "Explicit index.html is served without redirecting",
```
Umang01-hash: Could you add the `/static/sub/index.html` case too? That's the one path that trips ServeFile's `301 -> ./` — it terminates fine (checked), but a case here locks it against regressing. Same for the docs line about `404.html` being served on a root miss.

Umang01-hash (commented):
Nice fix. Reproduced both the root 404 and the subdir redirect loop on development, and they're gone at this head — traversal, sibling exclusion and the openapi.json restriction all still hold, and dropping the FileServer directory listing is a good bonus.

Couple of things before this lands:

- This overlaps #3814 — same file, and `isWithinDirectory` here is the same idea as `isUnderStaticRoot` there. They'll conflict, and this is the more complete fix (that one only addressed the 403, not the loop). Can we land this one and drop the router.go hunk from #3814?
- Minor wording: the root case was actually a 404, not a 403 — the endless redirect was on subdirectories with an index. Doesn't change the fix, just so the issue/description match what's actually happening.

Not blocking: `os.Stat` now runs 2-3x per request (validateFile + ServeFile), but that's negligible.

pkg/gofr/http/router_test.go
akshat-kumar-singhal: Added it — and it wasn't fine. At the head you reviewed:

```
GET /static/sub/index.html  ->  301 Location: ./
```

It terminates, which is what the check would have shown: `./` resolves to `/static/sub/`, `path.Clean` turns that into `/static/sub`, and the second hop is a 200. But it's still a redirect on a path that has a file sitting there to serve.

The cause is that `http.ServeFile` carries its **own** `/index.html` → `./` rule, separate from `http.FileServer`'s. Swapping to `ServeFile` dropped the directory redirect and quietly inherited this one. It missed at the endpoint root only by accident — `StripPrefix` leaves `index.html`, with no leading separator, so the suffix doesn't match — and fires one directory deeper, where it does.

The 404 path had the same rule in a worse form. `respondWithFileError` called `ServeFile` on `404.html` while the *request's* URL was still in play, so:

```
GET /static/empty/index.html   (no such file)
before:  404 written, then ServeFile issues 301 Location: ./ over it — 404.html never sent
after:   404  <html>404 Not Found</html>
```

Fixed by serving through `http.ServeContent`, which is handed an open file rather than a URL and has no redirect rule to inherit — containment and directory-to-index resolution are already done by the time it's reached, so there's nothing left for a second path resolver to decide. The 404 page is now written out directly with an explicit content type so its status survives. That also folds in your `os.Stat` note: the open file's own `FileInfo` replaces `validateFile`'s second stat.

`Test_StaticFileServing_NoRedirect` collects the four requests that produce a redirect — subdirectory, root `index.html`, subdirectory `index.html`, and a miss ending in `/index.html`. Verified red against the previous head: the last two failed there, the first two passed. And `runStaticFileTests` now asserts an empty `Location` on **every** static case, so no future serving path can reintroduce one without a test noticing.

The docs line about `404.html` on a root miss is in `docs/advanced-guide/serving-static-files/page.md` (came in with the previous commit), plus a new line that `/static`, `/static/` and `/static/index.html` all return the same page and none of them redirects.

akshat-kumar-singhal:
Thanks — the test case you asked for turned up a real one. Details in the thread, short version: `/static/sub/index.html` was answering `301 Location: ./` at the head you reviewed, because `http.ServeFile` has its own `/index.html` rule separate from `FileServer`'s, and the 404 handler was issuing that redirect *over* an already-written 404 so `404.html` never went out. Both fixed by serving through `http.ServeContent`, in 76eee345.

On the other two:

**#3814 overlap** — agreed, and that PR is mine, so dropping the `router.go` hunk is easy. The ordering has to go the other way round, though: **merge this one first**, then I rebase #3814 onto `development` and drop its `router.go` + `router_test.go` hunks. Dropping them now would leave #3814 red — the test it makes green, `examples/using-html-template`'s `Test_IndexHTML`, is exactly the one this PR fixes, and #3814's whole point is that the job stops swallowing that failure. That's why the hunk got folded in there originally. Happy to push the rebase the moment this lands.

**Wording: 403 vs 404** — I checked this against `development` (1d238069) before answering rather than take it either way, and I don't think the change is right. Both statuses are real; which one you get depends on the mount:

```
mount "/"                                  mount "/static"
GET /               -> 403                 GET /static              -> 404
GET /index.html     -> 301 ./              GET /static/index.html   -> 301 ./
GET /sub            -> 301 sub/            GET /static/sub          -> 301 sub/
GET /sub/index.html -> 301 ./              GET /static/sub/         -> 301 sub/
```

The two roots fail for genuinely different reasons, which is why the description lists them as separate causes: the `"/"` mount root reaches the handler and is rejected by the containment check (403), while the `/static` root never routes at all, because `path.Clean` strips the trailing slash off `/static/` and the `PathPrefix("/static/")` route no longer matches (mux's default 404). Fixing either one alone leaves the other broken.

Calling it "a 404" would specifically misdescribe the failure in #3819's own reproducer — `examples/using-html-template` mounts at `/` and its `Test_IndexHTML` fails with **403**, not 404.

Same for the redirect: it isn't only subdirectories. `GET /static/index.html` looped too, on both mounts — that's the `301 ./` on line 2 above.

Both tables are in the description already; I've expanded cause 3 to spell out the `ServeFile` rule as well.

Umang01-hash (dismissed):

akshat-kumar-singhal:
Pushed `601f7d06` after your approval — flagging it since it changes `router.go`, not just tests.

It picks up @Umang01-hash's second note on #3814 ("a *relative* dirName in `AddStaticFiles` still 403s the root index"). It was filed as pre-existing and out of scope there, but it lives in the code this PR rewrites, so it belongs here rather than in a CI PR.

Reproduced first, and it is broader than the note: with such a name **every** path under the endpoint is 403, not only the root. Across the five forms a caller can use:

```
absolute                          /static -> 200   /static/index.html -> 200
absolute with trailing separator  /static -> 403   /static/index.html -> 403
relative                          /static -> 403   /static/index.html -> 403
./relative                        /static -> 403   /static/index.html -> 403
```

**Cause.** A form mismatch in the containment check, not a containment failure. `staticHandler` builds the requested path with `filepath.Abs(filepath.Join(directoryName, url))` — absolute and cleaned — and `isWithinDirectory` compares it against `directoryName` *as it was handed in*. A relative name is never a prefix of an absolute path, and a trailing separator makes the prefix `/app/public//`, which nothing matches. The check reports "outside the directory" for paths plainly inside it.

**Fix.** Resolve once at registration: `filepath.Abs(dirName)` in `Router.AddStaticFiles` (`Abs` cleans, so it covers the trailing separator too), and store that. The per-request path keeps no extra syscall-backed call, and there is one place where `directoryName`'s form is established — which is what makes the string prefix test sound. A name that cannot be resolved is logged and the endpoint is not registered: containment cannot be checked against a directory whose absolute path is unknown, so refusing to serve is the safe end of that. The registration log now reports the path actually served rather than what the caller typed.

**Reachability**, since it bears on how much this matters: `App.AddStaticFiles` resolves `./x` and `x` against the working directory and `factory.go:56` builds its own path with `filepath.Join`, so the form reaching the router unnormalized *through the public API* is an absolute name with a trailing separator. The relative forms bite direct users of `pkg/gofr/http.Router`.

`Test_StaticFileServing_DirectoryNameForms` covers the endpoint root and a file beneath it across all five forms — verified red before the fix (four of five failed). `./pkg/gofr/http/... ./pkg/gofr` pass, `examples/using-html-template` passes, and `golangci-lint` reports nothing new on the changed files.

Happy to split this into its own PR if you would rather keep this one at the diff you approved.

aryanmehrotra (changes requested):
Re-reviewed the whole thing at `76eee34`. All six items from the last round are addressed, and the `ServeContent` swap is the right call — the 404-path bug you found chasing my `index.html` note was worse than the one I flagged, and I confirmed it on `development`:

```
GET /static/empty/index.html  ->  404  Location: ./     (404.html never sent)
```

Verified independently rather than from the description — built both heads and ran 14 paths across the `/` and `/static` mounts:

- `examples/using-html-template` `Test_IndexHTML`: **red on `development`, green here**
- no `Location` on any static response
- traversal, `..%2f`, sibling exclusion (`/staticother/...`) and the `openapi.json` restriction all still hold
- `HEAD` and `Range` at the endpoint root now work — both were 404
- docs match the example tree they describe
- no new lint findings on `router.go`

One correction to the description, in your favour: it says `/static/index.html` "existed as a URL but was a redirect loop." It terminated in a **404**, not a 200 — `301 ./` → `/static/` → `path.Clean` → `/static`, which was unrouted. So that content was genuinely unreachable, and "reachable for the first time" is if anything understated.

## Not a breaking change in the sense that gates a release

Worth stating plainly since the section is titled that way. I diffed every top-level declaration in `router.go`: every changed signature is on the unexported `staticFileConfig`.

```
staticHandler(fileServer) -> staticHandler()
validateFile(abs) error   -> validateFile(abs) (string, error)
respondWithFileError(w, r, ...) -> (w, ...)
+ serveFile, + isWithinDirectory
```

`Router.AddStaticFiles`, `App.AddStaticFiles` and `DefaultSwaggerFileName` are untouched, so nothing downstream stops compiling. The behavioural delta is exactly `index.html` files at directory roots — every other file was already reachable at its own path, no listing is introduced, and all the containment guards are intact. Plus the no-index 404 changing body and content-type with the status unchanged. That is a minor-release note, not a compatibility break, and both are already documented.

## Requested change

Dropping `FileServer` means nothing checks the *file type* of the resolved path any more. `validateFile` re-stats after resolving a directory to its index, but only for permissions, so a non-regular file reaches `http.ServeContent`:

| `index.html` is… | `development` | this head |
|---|---|---|
| a directory | 301 | **200, `Content-Length` from the directory's `FileInfo`, zero bytes** |
| a fifo | 301 | **handler blocks forever in `os.Open`** |

The first is a real regression — a client reading that response gets `unexpected EOF` against a 200. The second is pre-existing for an ordinary filename (a fifo at `static/pipe.txt` blocks on `development` too, via `http.Dir.Open`), so it isn't yours, but the same check closes it.

In `validateFile`, before the permission check:

```go
// Only regular files are servable. A non-regular file would otherwise reach
// http.ServeContent, which writes a Content-Length from its FileInfo and then no body
// (a directory), or blocks forever in os.Open (a fifo).
if !fileInfo.Mode().IsRegular() {
    return "", fs.ErrNotExist
}
```

`fs.ErrNotExist` puts it on the ordinary 404 path, same as the no-index case, so it needs no new response handling. Checked at your head: all three shapes above answer 404, `./pkg/gofr/http/...` stays green and `golangci-lint` is clean on the file. Worth a case in `Test_StaticFileServing_EndpointRoot` for the directory shape — it's the one that regressed.

## Not blocking, and not yours — three that pre-date this PR

All three are unchanged by this PR and identical on `development`, so they're separate issues. Raising them only because you're editing these exact two functions and each is a line:

- `GET /static/openapi.JSON` returns **200 and serves the spec**. `isRestrictedFile` compares `filepath.Base(url)` case-sensitively; `strings.EqualFold` closes it. Only reachable on a case-insensitive filesystem — which is every macOS dev machine.
- A relative `dirName` 403s **every** path, not just the root: `isWithinDirectory` compares an absolutised `absPath` against the raw `directoryName`. One `filepath.Abs` in `AddStaticFiles` kills the class. Not reachable through `App.AddStaticFiles`, which normalises at `gofr.go:414`, so it needs the internal Router API today.
- `AddStaticFiles("static/", dir)` registers `Path("/static/")` and `PathPrefix("/static//")` and is silently dead, because `gofr.go:417` trims only the leading slash. `strings.Trim` instead of `TrimPrefix` fixes it.

Happy to file these as issues if you'd rather not widen this PR — the first is the one I'd want tracked.

## Merge order

This merges cleanly onto `development`, and it's the more complete fix of the two overlapping changes — #3814's `isUnderStaticRoot` is the same idea as `isWithinDirectory` here but only addresses the 403, not the redirects. Same conclusion as the other review: land this, drop the `router.go` hunk from #3814 and rebase it. The five `examples/*/main_test.go` fixes there stay where they are.

Still worth running `Test_StaticFileServing_EndpointRoot` with `GOFR_ROUTER` enabled whenever this and #3759 meet — this PR is the first to split one static endpoint across an exact `Path` route and a `PathPrefix` fallback, and these tests only exercise the default mux path. Nothing to do here; #3759 is still open.

akshat-kumar-singhal:
Pushed `d9c8b3a0`. The file type check is in, and it was a real 200-with-no-body — I reproduced your table before fixing rather than taking it from the review:

```
static/index.html is a directory
before:  GET /static -> 200  Content-Length: 96   (zero bytes; client reads unexpected EOF)
after:   GET /static -> 404  <html>404 Not Found</html>
```

`validateFile` returns `fs.ErrNotExist` unless `fileInfo.Mode().IsRegular()`, after the directory-to-index resolution and before the permission check, so it covers the resolved path in both shapes — the endpoint root's index file and an ordinary filename. As you say, that needs no new response handling: it lands on the same 404 path the no-index case takes, and it closes the fifo hang too.

The cases live in a new `Test_StaticFileServing_EndpointRootWithoutServableIndex` rather than in `Test_StaticFileServing_EndpointRoot` — adding them there took that function past `funlen`'s 100 lines. It collects every root that resolves to nothing servable: no index file, an `index.html` that is a directory, a subdirectory root without an index, and the `404.html` case. Verified red before the fix.

**`openapi.JSON`** — taken here rather than filed, since it's one line in a function this PR already rewrites and it's a restriction bypass, not cosmetics. `strings.EqualFold`. Confirmed on macOS first:

```
before:  GET /static/openapi.JSON -> 200  {"openapi": "3.0.0"}
after:   GET /static/openapi.JSON -> 403
```

Both new tests hold on a case-sensitive filesystem as well — the restriction is applied to the URL before the file is resolved, so the case that would actually be *served* never enters into it.

**The third note** — `AddStaticFiles("static/", dir)` registering a dead endpoint — I've left out. It's in `App.AddStaticFiles` at `gofr.go:417`, a file this PR doesn't touch, so it goes in its own PR off `development` rather than widening a diff you've already been through four times. Filed as #3827. The relative-`dirName` one you listed second was already fixed in `601f7d06`.

**CI.** `Code Quality` was failing on one `noctx` finding at `router_test.go:497` — `httptest.NewRequest` in the `DirectoryNameForms` test I added in `601f7d06`; now `NewRequestWithContext(t.Context(), ...)`. `golangci-lint --new-from-rev=development ./...` is clean, which is the same `only-new-issues` shape the job runs. `./pkg/gofr/... ./pkg/gofr/http/...` and `examples/using-html-template` all pass.

aryanmehrotra:
The three fixes from the last round are in, verified on both heads rather than from the description:

```
                              development        11058337
index.html is a directory     301                404
index.html is a fifo          301                404
pipe.txt (ordinary name)      handler hangs      404      <- pre-existing, closed too
/static/openapi.JSON          200 + the spec     403
/static/OPENAPI.JSON          200 + the spec     403
dirName relative              403 every path     200
dirName trailing separator    403 every path     200
dirName unclean (/x/./)       403 every path     200
```

#3041's guard also survives the `isRestrictedFile` rewrite, which matters because that guard is why this bug exists: `/app/publicother/secret.txt`, `/app/publicother`, `/app/public-backup/…`, `/app/public.old/…`, `/app/publicsecret`, `/app/publicX` and `/app/publi` all stay restricted, identical on both heads, and the only row that flips is `/app/public` itself. Seven end-to-end escape attempts (`../`, `..%2F`, `%2e%2e`, `//../`, `sub/../../`, sibling prefix) return 404 on both. With basic auth on, every `/static` path still answers 401 without credentials, so nothing became reachable unauthenticated.

The code is close. What I want before merge is mostly about the parts that change behaviour for people who are currently working — the correctness fixes I have no argument with. Detail below, then four asks, then two questions I would like your reasoning on rather than a patch.

## The release history changes how this should be described

I ran the same matrix against `v1.51.0`, `v1.54.4`, `development` and your head:

| | v1.51.0 | v1.54.4 | development | this PR |
|---|---|---|---|---|
| `GET /static/` | **200 index** | 404 | 404 | 200 |
| `GET /static/sub` | 301 → `sub/` | 301 → `sub/` | 301 → `sub/` | 200 |
| `GET /static/sub/` | **200 index** | 404 | 404 | 200 |
| `GET /static/empty/` | **200 + directory listing** | 301 loop | 301 loop | 404 |
| `GET /` (root mount) | **200 index** | **200 index** | **403** | 200 |
| `GET /static` (bare) | 404 | 404 | 404 | 200 |
| sibling escape blocked? | **no** | yes | yes | yes |
| `/static/openapi.JSON` | 200 spec | 200 spec | 200 spec | 403 |

This is **two regressions from two different PRs**, not one:

**Causes 2 and 3 came from #2759, in v1.52.0 (Jan 2) — 20 releases ago.** That PR fixed #2643: mux's automatic cleaning issued a 301 for `//hello`, and Postman turned the POST into a GET on the redirect. The first implementation was a hand-rolled `normalizePathSlashes` that collapsed consecutive slashes and *preserved* trailing ones. Then, in review:

> `path.Clean` from std lib might be what you are looking for

> Thanks! I used this `path.Clean()` and completely eliminated our `normalizePathSlashes` function.

`path.Clean` also strips trailing slashes. That single substitution is both the unrouted `/static/` and the redirect loops, and it went through four approvals without the second effect being noticed. Before it, `/static/sub` → `301 sub/` → `/static/sub/` → `200` resolved correctly — plain `http.FileServer` behaviour.

**Cause 1 came from #3041, in v1.54.5 (Feb 27) — 13 releases ago.** `v1.54.5` was tagged 2h09m after `6cf00652` landed, so it shipped the same day. `v1.58.0` still carries both.

Worth recording how #3041 got through, because it is the answer to "why did five months pass". The only review feedback on it asked to *remove* a test:

> The integration test `Test_StaticFileServing_SiblingDirectoryEscape` does not exercise `isRestrictedFile` because the traversal is blocked by `path.Clean` in the router before the static handler is invoked. We can drop it entirely — `Test_isRestrictedFile` already covers the fix.

That observation is correct and I reproduced it — `path.Clean` does block the traversal before the handler, so the containment check is defence in depth. But dropping the end-to-end case left only a unit test calling `isRestrictedFile` with hand-written strings, which is exactly why nothing caught the directory-itself case. No action for you; yours has the end-to-end coverage. It is the argument for keeping it.

And #3041 was not gratuitous: on `v1.51.0` the sibling escape is genuinely open, so reverting is not on the table. This PR is the first time both are right at once.

## You are also closing a listing exposure — worth claiming

`GET /static/empty/` on `v1.51.0` returns `200` and Go's `dirList` output. The loop introduced in v1.52.0 made it unreachable by accident, so nobody can see it on `development` — but it was live through v1.51.1, and re-routing the endpoint root would have brought it straight back. Your directory-to-index resolution is what keeps it closed. That is a stronger argument for the no-index-404 decision than "that path was never reachable before", and I would put it in the description.

## Four asks

### 1. Non-GET methods return the file

```
GET     /static -> 200 <html>X</html>
HEAD    /static -> 200
POST    /static -> 200 <html>X</html>
PUT     /static -> 200 <html>X</html>
DELETE  /static -> 200 <html>X</html>
```

A `DELETE` answering 200 with a body reads as "deleted, here it is". Nothing was deleted.

I first thought `.Methods(http.MethodGet, http.MethodHead)` on the two routes, but that makes them not match, the request falls to the catch-all, and you get a 404 — which is its own lie, since GET on that exact path is a 200. An in-handler check gives the right answer plus the `Allow` header RFC 9110 §15.5.6 requires:

```go
if r.Method != http.MethodGet && r.Method != http.MethodHead {
    w.Header().Set("Allow", "GET, HEAD")
    staticConfig.respondWithError(w, "method not allowed on static endpoint", r.URL.Path, nil,
        http.StatusMethodNotAllowed)

    return
}
```

Measured with that applied at your head:

```
GET      /static             -> 200 Allow=""
HEAD     /static             -> 200 Allow=""
DELETE   /static             -> 405 Allow="GET, HEAD"
POST     /static             -> 405 Allow="GET, HEAD"
PUT      /static/index.html  -> 405 Allow="GET, HEAD"
OPTIONS  /static             -> 200                    <- CORS still short-circuits pre-routing
```

`./pkg/gofr/http/...` green, lint clean. This is 405 in a framework that 404s wrong-method everywhere else (`POST /api/thing` on a GET route returns the error envelope, because the catch-all matches before mux's `MethodNotAllowedHandler`). I think the inconsistency is worth it rather than shipping a file-serving DELETE, and GoFr's own 405 gap is a separate issue — but say so if you would rather keep static consistent with the framework and file both together.

### 2. Permission denied returns 500

```
/static/noread.txt   (file, mode 0000)          -> 500
/static/nodir        (dir mode 0000, has index) -> 500
```

Should be 403 — `net/http.toHTTPError` maps `fs.ErrPermission` → 403 (`$GOROOT/src/net/http/fs.go`), and nginx and Apache agree. 500 tells a monitor the app is broken when it is a file mode. Pre-existing — 500 on `v1.51.0`, `v1.54.4` and `development` too — but `respondWithFileError` is a function you are rewriting and it is one clause:

```go
if errors.Is(err, fs.ErrPermission) {
    staticConfig.respondWithError(w, "no read permission", absPath, err, http.StatusForbidden)
    return
}
```

Two routes in need covering: a real `EACCES` from `os.Stat`/`os.Open`, and your own `errReadPermissionDenied`, which is not an `fs.ErrPermission` today. Separately, `Perm()&0444` asks whether *anyone* has a read bit, not whether this process can read — attempting the open and mapping the error is the stdlib approach and is strictly more accurate.

### 3. A `dirName` that is not a directory is now served

`App.AddStaticFiles` only `os.Stat`s the path (`gofr.go:419`) — it never checks `IsDir`, so a regular file passes registration. `absPath == directoryName` now admits it, where the old prefix test failed closed:

```go
app.AddStaticFiles("data", "./config.yaml")
```
```
v1.51.0 / v1.54.4 / development:  GET /data -> 404
11058337:                         GET /data -> 200 db_password: hunter2
```

Reachable through the public API, on `/data` and `/data/`. It takes a misconfiguration to hit, so it is not an attack path, but it turns a fail-closed guard into a fail-open one — and it is the one item that stops this being a clean release. After the `filepath.Abs` you added:

```go
if info, err := os.Stat(absDir); err != nil || !info.IsDir() {
    logger.Errorf("error in registering '%v' static endpoint, %v is not a directory", endpoint, absDir)

    return
}
```

### 4. `docs/…/serving-static-files/page.md:47` contradicts your new paragraph

The pre-existing line still says a `404.html` works by **redirecting** all Not Found requests to that page, four lines below your new paragraph stating that nothing redirects. Old loose wording, but this PR is what makes it read as wrong.

Everything else you added to that page I checked claim by claim against measured behaviour — endpoint root, subdirectory roots, no-index 404, `404.html` as the body, no redirect on any of the three forms, non-regular files, `openapi.json` in any capitalization, and the `/public` vs `/static` example against its own tree. All accurate.

## One row of Breaking Changes is wrong

It says the no-index 404 was `404, text/plain, 404 page not found\n` (mux default). That is what a bare `NewRouter()` gives in a unit test; a real app never sees it, because `gofr.go:183` registers a catch-all returning the error envelope:

| | before | after |
|---|---|---|
| `GET /static`, no `index.html` | `404`, `application/json`, `{"error":{"message":"route not registered"}}` | `404`, `text/plain`, `404 Not Found` — or `text/html` and the folder's `404.html` |

Worth fixing because the real change is `application/json` → text, which breaks a client parsing the envelope, whereas `text/plain` → `text/html` reads cosmetic.

I would also restructure that section around what the history shows, since "breaking" is currently covering four different things:

- **Restored** — `GET /` on a root mount, broken in v1.54.5 only (13 releases).
- **Restored** — `/static/`, `/static/sub`, `/static/sub/index.html`, broken in v1.52.0 by the `path.Clean` substitution (20 releases).
- **New** — bare `GET /static`. Never worked in any release, so no working setup can break.
- **Closed** — directory listings, live through v1.51.1.
- **Genuinely breaking** — the 404 body shape above, and after ask 3, nothing else I could find.

Per `CONTRIBUTING.md:134` (semver) that is a **minor**, not a patch: no exported API changes — I diffed every top-level declaration and every changed signature is on the unexported `staticFileConfig` — but bare `GET /static` is new functionality, which is clause 7.

## Two questions, not asks

These are the parts I am uneasy about, and I would rather have your reasoning than send a patch.

### Why serve the bare `/static` rather than redirect to `/static/`?

`/static` and `/static/` return the same bytes, so per RFC 3986 §5.1.3 a relative reference in that HTML resolves against a different base. Measured with Go's own URL resolution:

```
served at /static   ->  <img src="logo.png">  ->  GET /logo.png         -> 404
served at /static/  ->  <img src="logo.png">  ->  GET /static/logo.png  -> 200
```

So the SPA case this PR exists for is half-broken at the URL people will actually type — it looks fine in curl and the images 404 in a browser. nginx, Apache and `http.FileServer` all answer `/static` with a redirect to `/static/` for exactly this reason, and `v1.51.0` did resolve directory redirects correctly.

I know why you cannot redirect: `path.Clean` strips the slash the redirect points at, and the client loops. I prototyped the root-cause fix on `development` — restore the trailing slash after `Clean`, which is literally what `net/http.cleanPath` does, comment included:

```go
np := path.Clean(p)
// path.Clean removes trailing slash except for root;
// put the trailing slash back if necessary.
```

It fixes static exactly as you would hope — `/static/` serves, `/static/sub` → `301 sub/` → `200`, no loop. And it breaks every other route in every GoFr app:

```
                development    with the normalizer fixed
GET  /users/    200 USERS   ->  404
POST /users/    201 CREATED ->  404
GET //users//   200 USERS   ->  404
```

Only `TestIsCleanPath` fails — `TestRouter`, `PathNormalization`, `DoubleSlash` and every static test stay green, so it would ship silently. And the standard remedy is `StrictSlash(true)`, which emits a **301**, which converts POST to GET, which is #2643 again. It needs **308**, and gorilla/mux does not offer 308 for StrictSlash.

So I am not asking you to fix it here — I now think your workaround is the right call for this PR. What I would like in the description is the limitation stated, and a docs line that HTML served at an endpoint root should use `/static/` or a `<base href>`. Do you see a reason to prefer serving bare `/static` over 404-ing it and leaving the redirect question to the router fix? Serving it is what creates the relative-URL trap, and it is also the only genuinely *new* behaviour in the PR.

### Should the containment check survive at all?

Not for this PR, but it bears on how you write `isWithinDirectory`. String-prefix containment is lexical, and symlinks are physical, so a symlink inside the served directory pointing outside passes every variant of it — the current one on `development`, and yours. Verified:

```
LINK                 KIND                       your check   os.Open (today)
rel-inside.png       relative -> inside         allowed      OK
abs-inside.png       ABSOLUTE -> inside         allowed      OK
dotdot-inside.png    relative w/ .. -> inside   allowed      OK
abs-outside.txt      ABSOLUTE -> outside        allowed      OK  <- serves the outside file
rel-outside.txt      relative -> outside        allowed      OK  <- serves the outside file
```

`os.Root` (Go 1.24, and we require 1.26) rejects both escapes with `path escapes from parent`, and it is not crude about it — `dotdot-inside.png`, a relative link containing `..` that lands back inside, is still allowed. The Go blog on `os.Root` names this case directly: code that calls `filepath.Join` on a fixed directory plus an externally-provided name should use `os.Root` instead, which is `staticHandler` exactly.

Two caveats I checked so the trade is visible: `os.Root` also rejects **absolute** symlinks even when the target is inside the root, so an intra-tree `ln -s /srv/app/public/assets/logo.v2.png current.png` would stop working; and the escape error satisfies neither `fs.ErrNotExist` nor `fs.ErrPermission`, so it maps to 500 unless handled explicitly. The common deploy pattern is unaffected — I verified `os.OpenRoot` resolves a symlinked served directory normally, including `current -> releases/42`, absolute or relative, with nested assets and escapes still blocked.

I will file this as its own issue since it is a behaviour change of its own and unrelated to #3819. Raising it here only because you are rewriting that function, so if you would rather do it properly in one go than touch it twice, say so and I will hold this PR for it.

Once 1–4 are in and the description is updated, this is good to go from my side.

akshat-kumar-singhal:
Thanks — the release archaeology is what makes this reviewable. All four asks are in, and the
description is rewritten around restored/new/closed/breaking.

**1. Non-GET methods.** In, as an in-handler check with `Allow: GET, HEAD`, for the reason you gave:
`.Methods()` on the routes answers 404 for a path where GET is a 200. I agree the inconsistency with
the framework's wrong-method 404 is worth it rather than shipping a file-serving DELETE — GoFr's own
405 gap is now filed as #3853, with the router-level A/B that isolates it to the `gofr.go:181`
catch-all matching on path alone. Worth noting there: removing the catch-all is not the fix, because
mux's default `MethodNotAllowedHandler` sets no `Allow` header and writes no body. I'm happy to pick
it up.

**2. Permission denied.** In. `errReadPermissionDenied` now wraps `fs.ErrPermission`, and
`respondWithFileError` maps it before the 500 fall-through, so both routes answer 403 — the mode
check and a real `EACCES`. I kept the `Perm()&0444` check rather than relying on `os.Open` alone:
you're right that it asks the wrong question, but since both now answer 403 it can only decide
*which* route reports it, and it refuses a mode-0000 file identically whatever user the process runs
as, where `os.Open` alone serves it to root. The comment says exactly that, so it isn't mistaken for
the authority.

**3. Non-directory `dirName`.** In, at registration, right after the `filepath.Abs`. Agreed this was
the one that stopped it being a clean release — the old prefix comparison failed closed here by
accident and I turned it fail-open.

**4. Docs line 47.** Fixed — it never redirected; the page is the body of the 404. Also added the
method restriction and the relative-URL caveat.

**Breaking Changes row.** Corrected to the `application/json` error envelope from the catch-all, with
a note that the `text/plain` form is what a bare `NewRouter()` gives in a unit test and not what a
real app sees. Restructured as restored / new / closed / genuinely-breaking, marked **minor** per
clause 7, and the v1.51.1 directory-listing closure is now claimed as an argument for the no-index
404 rather than left implicit.

## Why serve the bare `/static` rather than redirect to `/static/`

Because it isn't a separate decision. `/static/` and `/static` are the same request by the time
anything can route them — `Router.ServeHTTP` rewrites the path before handing it to mux:

```
request "/static"  -> arrives at the router as "/static", 200
request "/static/" -> arrives at the router as "/static", 200
```

So there is no route that serves `/static/` and 404s `/static`. 404-ing the bare form 404s the
trailing-slash form too, and that is the v1.52.0 regression the PR exists to fix. It reads as new
functionality in the table, and strictly it is — but it's inseparable from the restoration, not a
feature I chose to add alongside it.

Which means I agree with your relative-URL finding and can't fix it here. The trap is real and I've
documented it — root-relative URLs or a `<base href="/static/">` — and stated the limitation in the
description. Your prototype is the right fix and it lives in the router's normalization, where the
308-vs-301 problem has to be solved for every route rather than for static alone.

## Should the containment check survive

Yes, it should become `os.Root` — you're right that lexical containment and physical symlinks don't
meet, and `staticHandler` is the case the Go blog names. I'd rather it were its own PR than folded
in here, for two reasons:

It isn't a rewrite of `isWithinDirectory`. Doing it properly means `os.OpenRoot` on the served
directory once at registration and `root.Open` on a relative path per request, which replaces the
`filepath.Abs` + containment + `os.Open` sequence — so `isWithinDirectory` doesn't get touched
twice, it goes away. There's little saved by doing it in the same change.

And it has a user-visible loss that deserves its own release note rather than being the ninth bullet
here: an intra-tree absolute symlink (`ln -s /srv/app/public/assets/logo.v2.png current.png`) stops
resolving even though its target is inside the root. Good that you verified the `current -> releases/42`
deploy pattern is unaffected — that's the one that would have worried me — but the absolute-link case
is a real break for someone, and this PR is already carrying three genuine breaks.

So: file it, and I'll pick it up. If you'd rather hold this one and do both at once, say so and I'll
fold it in — I just don't think #3819's fix should wait on a behaviour change unrelated to it.

Umang01-hash:
@aryanmehrotra all four asks are in now — the in-handler 405 with `Allow: GET, HEAD`, `errReadPermissionDenied` wrapping `fs.ErrPermission` so unreadable files answer 403, the `IsDir` guard at registration, and the `page.md:47` fix. They landed in the last two commits (089bc32b, 8d6fe113), after your review — could you give the final code a quick look? Good to merge as a minor from my side once you're happy.

aryanmehrotra (approved):
Verified `8d6fe113`. All four are in and behave:

```
### methods
POST/PUT/DELETE/PATCH  /static           -> 405 Allow="GET, HEAD"
DELETE                 /static/logo.png  -> 405 Allow="GET, HEAD"
GET                    /static           -> 200
HEAD                   /static           -> 200, no body
OPTIONS                /static           -> 200 in a real app  (CORS short-circuits at http_server.go:41)

### permission
/static/noread.txt  (file, mode 0000)    -> 403
/static/nodir       (dir, mode 0000)     -> 403

### non-directory dirName
AddStaticFiles("/data", "…/logo.png")    -> refused at registration, /data and /data/ 404
```

Full regression sweep unmoved — `/static`, `/static/`, `/static/index.html`, `/static/sub`, `/static/sub/index.html` all 200; `empty`, `nope`, `dirindex`, `pipe.txt` all 404; `openapi.json` and `openapi.JSON` 403; `../outside.txt` and the sibling prefix 404. `./pkg/gofr/http/...` green, `examples/using-html-template` green, `golangci-lint --new-from-rev=origin/development ./pkg/gofr/...` reports 0 issues. Three named tests for the three fixes.

Your reasoning on both questions holds, and on the first one I was wrong to ask for a choice:

**Bare `/static`.** You're right that it isn't a separate decision — `ServeHTTP` rewrites the path before mux sees it, so `/static/` and `/static` are the same request and there is no route that can serve one and 404 the other. 404-ing the bare form would 404 the trailing-slash form, which is the v1.52.0 regression. I'd been treating "serving the bare form" as an independent choice and it isn't. The docs paragraph you added on relative resolution is the right place for it, and the `<base href>` line is what a user actually needs.

**`Perm()&0444`.** Keeping it is right and the comment is the reason — it refuses a mode-0000 file identically regardless of the process user, where `os.Open` alone would serve it to root. Being explicit that `os.Open` is the authority is what makes it safe to keep.

**`os.Root`.** Agreed, and your argument for separating it is better than mine for folding it in: `isWithinDirectory` doesn't get touched twice, it disappears — `OpenRoot` at registration plus `root.Open` per request replaces the whole `Abs` + containment + `Open` sequence. And the intra-tree absolute-symlink loss deserves its own release note rather than a ninth bullet on a PR already carrying three breaks. Filing it; take it when you're ready.

Also good catch on #3853 that removing the catch-all isn't the fix, since mux's default `MethodNotAllowedHandler` writes no `Allow` and no body. That's the part that makes it a real piece of work rather than a deletion.

The Breaking Changes section is now more honest than I asked for — you added the 405 and the 403 as breaking rows, which they are.

One footnote, not blocking: package coverage is 88.1% against 88.4% on `development`. It's the `filepath.Abs` and `os.Open` error branches, which need `Getwd` to fail or a TOCTOU race, so I don't think they're worth chasing — but the checklist line about not decreasing coverage isn't quite true as written.

Approving. Merge as a minor.

Umang01-hash (approved):

PR #3821, opened 2026-07-31
fix(examples): wait on a readiness signal instead of sleeping, and make using-migrations self-contained (#3818, #3816)

### LINKED ISSUE #3816: examples/using-migrations test is not self-contained: migration state persists in Redis between runs

### LINKED ISSUE #3818: Replace fixed time.Sleep server-startup waits in example tests with readiness polling
**Describe the bug**

Most example tests start the server with `go main()` and then wait for it with a fixed
`time.Sleep(100 * time.Millisecond)` before making their first request. Nothing verifies the
server is actually listening, so on a slow or loaded CI runner the first request can hit a
socket that is not accepting yet, and the test fails for reasons unrelated to what it covers.

This is latent today rather than actively breaking:

- The `Example Unit Testing` job runs through `nick-fields/retry` with `max_attempts: 2`, so a
  test that fails independently with probability *p* only reddens the job at roughly *p²*.
- Some of these paths are skipped by the job's `-short` flag.

So the retry is what currently papers over the pattern. That matters if we ever want to drop
`max_attempts: 2` — and it also means a genuine regression in one of these examples can be
masked by a retry that happens to pass.

#3817 established the fix pattern in `examples/using-add-rest-handlers` and
`examples/using-migrations`: poll the framework's own liveness endpoint instead of guessing a
duration.

```go
// waitForServer waits until the example is serving.
func waitForServer(t *testing.T, host string) {
	t.Helper()

	require.Eventually(t, func() bool {
		req, err := http.NewRequestWithContext(t.Context(), http.MethodGet, host+service.AlivePath, nil)
		if err != nil {
			return false
		}

		resp, err := http.DefaultClient.Do(req)
		if err != nil {
			return false
		}

		defer resp.Body.Close()

		return resp.StatusCode == http.StatusOK
	}, 30*time.Second, 100*time.Millisecond, "server did not start in time")
}
```

This issue tracks porting that to the remaining examples. It is deliberately **not** folded into
#3814 (the CI errexit fix): those jobs are green with these sleeps in place, so this is
pre-existing debt rather than something that PR exposes, and bundling ~20 files of test churn
into a six-line shell fix would bury the change under review.

**Scope**

Counts are as of `development` with #3814 and #3817 merged (#3817 already removes the
`using-add-rest-handlers` and `using-migrations` sleeps). Two commented-out sleeps in
`using-http-auth-middleware` are excluded.

*Category A — startup waits before the first connect (26 sites).* These are the actual races and
the direct target of the pattern above:

| File | Lines |
| --- | --- |
| `examples/grpc/grpc-streaming-server/main_test.go` | 61 |
| `examples/grpc/grpc-unary-client/main_test.go` | 60, 67, 119, 126 |
| `examples/grpc/grpc-unary-server/main_test.go` | 32, 67, 100, 167 |
| `examples/http-server/main_test.go` | 59, 109, 174 |
| `examples/http-server-using-redis/main_test.go` | 34 |
| `examples/using-custom-metrics/main_test.go` | 25 |
| `examples/using-file-bind/main_test.go` | 27 |
| `examples/using-graphql/main_test.go` | 32 |
| `examples/using-html-template/main_test.go` | 28, 65, 94 |
| `examples/using-http-auth-middleware/main_test.go` | 18, 45 |
| `examples/using-http-service/main_test.go` | 39 |
| `examples/using-publisher/main_test.go` | 36 |
| `examples/using-s3-filestore/main_test.go` | 66 |
| `examples/using-subscriber/main_test.go` | 27 |
| `examples/using-web-socket/main_test.go` | 25 |

The gRPC examples need care: several start a gRPC server with no HTTP port, so `service.AlivePath`
does not apply and readiness has to be established another way (a `grpc.WaitForReady` call or a
dial-and-probe loop) rather than a copy-paste of the HTTP helper.

*Category B — deliberate mid-test pauses, not startup races (4 sites).* A readiness poll does not
apply; these need a synchronisation signal or an `Eventually` on the observable effect:

- `examples/grpc/grpc-streaming-server/main_test.go:164` — pacing between `stream.Send` calls
- `examples/grpc/grpc-streaming-server/main_test.go:226`, `:273`, `:363` — "give server time to
  detect cancellation" after `cancel()`

*Category C — `examples/using-cron-jobs/main_test.go:22` (1 site).* `time.Sleep(1100ms)` covers
both startup and waiting for a 1-second cron to fire. Wants `require.Eventually` on the job's
side effect, not a readiness probe.

*Not in scope:* `examples/using-ai/main_test.go:153` and `:170` are already backoff sleeps inside
poll loops. That is the correct pattern — leave them.

**Expected behavior**

Example tests wait on an observable readiness signal and fail with a clear timeout message if the
server never comes up, rather than racing a fixed duration and relying on a CI retry to hide it.

akshat-kumar-singhal:
**Description:**

Fixes #3818. Fixes #3816 — superseding #3817, which is closed in favor of this PR because its two
files need the `testutil` helpers added here and stacking across a fork is not possible.

Most example tests started the server with `go main()` and then waited for it with
`time.Sleep(100 * time.Millisecond)`. Nothing verified the server was listening, so on a slow or
loaded runner the first request can hit a socket that is not accepting yet and the test fails for
reasons unrelated to what it covers. The `Example Unit Testing` job's `max_attempts: 2` retry is
what keeps this latent — and it equally masks a genuine regression that happens to pass on retry.

All 31 sleeps the issue enumerates are replaced with an observable signal:

- `testutil.WaitForHTTPServer` polls `/.well-known/alive`.
- `testutil.WaitForGRPCServer` waits for a gRPC connection to reach `connectivity.Ready`. A bare TCP
  dial would not do — the kernel accepts connections from the moment the listener exists, so it
  succeeds before the server is serving, restoring the race it is meant to remove. GoFr registers no
  gRPC health service by default, so there is no health RPC to call.

Both fail with a named timeout after 30s rather than a bare connection-refused from the first
assertion.

Three cases needed more than a substitution:

- **`using-cron-jobs`** asserted `n == 1` after sleeping past one tick of a 1-second cron. That
  equality was itself the flake — any wait long enough to be safe on a loaded runner also lets the
  job fire twice. It now waits for `n > 0`, which is the claim the test actually makes.
- **The four mid-test sleeps in `grpc-streaming-server` are deleted, not converted.** Three are
  "give server time to detect cancellation" after `cancel()`; the following `Recv`/`CloseAndRecv`
  fails with `Canceled` off the client's own context, so there is nothing to wait for. The fourth
  paced `stream.Send`, but the server handles one stream sequentially and replies in order, so it
  cannot affect the ordering the test asserts.
- **`using-subscriber` has no HTTP server to poll and gets a third helper,
  `testutil.WaitForStdoutContains`.** The example registers only subscribers, so `App.httpRegistered`
  stays false and `startHTTPServer` never runs — there is no liveness endpoint, and no port that
  ever opens. Its assertion is on a Kafka log line captured by `StdoutOutputForFunc`, which reads
  the pipe only once its function has returned, so waiting inside it would mean sleeping again. The
  new helper captures stdout and reads it concurrently, returning as soon as the awaited line
  appears and failing with everything captured if it never does: the line the test asserts on is now
  the signal it waits for.

**`using-migrations` owning no namespace (#3816)** is the same class of problem one layer down: the
example passed once against a fresh Redis/MySQL pair and failed on every run after that.

- The migration bookkeeping is keyed by version and lives in **both** datastores (`gofr_migrations`
  hash in Redis, `gofr_migrations` table in MySQL), and `redisMigrator.getLastMigration` returns
  `max(redis, sql)` — so a stale Redis record alone is enough to skip the schema-creating migration.
  With Redis carried over and MySQL recreated, the run logs `no new migrations to run` and every
  request fails with `Table 'test.employee' doesn't exist`.
- The row the test POSTs (`id: 2`) and the row the migration inserts survive in MySQL, so a re-run
  collides: `expected 201, actual 500` / `Duplicate entry '2' for key 'employee.PRIMARY'`.
- Every MySQL example migrates into the same `test` database and therefore shares one
  `gofr_migrations` table. `using-migrations`' versions (`1722507126`, `1722507180`) are higher than
  `using-add-rest-handlers`' (`1721816030`), so once `using-migrations` has run, the sibling's
  migration is considered applied and its `user` table is never created. It works today only because
  `go test` happens to run the alphabetically earlier package first on a fresh database.

CI never sees any of this — every job gets fresh service containers, so every CI run is a first run.

`TestMain` now creates (and re-creates) a `test_using_migrations` database the test owns and points
the example at it via `DB_NAME`, and does the same one layer down for Redis: it selects logical
database 1 via `REDIS_DB` — which GoFr reads, and which the migrator inherits because it runs on the
same client — and empties it with `FLUSHDB`. `configs/.env` is untouched — it is what a human runs the
example with, and `gofr.New()` honors the override because `godotenv.Load` does not replace
variables already set in the environment. Giving the example its own database removes the sibling
coupling as well.

`FLUSHDB` rather than deleting the keys the example happens to write today: a key list has to be kept
in step with a sibling `migrations/` directory, and the first Redis migration writing an unlisted key
would make the example non-repeatable again, which is the failure mode this change exists to close.
It is the more destructive of the two — it empties logical database 1 in full — so the comment says
so, for anyone who points `REDIS_HOST`/`REDIS_PORT` at a Redis holding anything else.

`using-graphql` was checked and is unaffected — its test builds its own app with stubbed resolvers
and never calls `Migrate`. `using-publisher` / `using-subscriber` migrate Kafka topics, not a shared
table.

The health-server commit is unrelated to sleeps but the same class of problem: `TestGoFrHealthServer_Methods`
and `TestGoFrHealthServer_SetServingStatus` both used the service name `"test-service"` against a
process-wide singleton, one asserting `"unknown service"` and the other registering it as SERVING.
Source order hides it on a single pass, but `go test -count=3` fails on the second iteration.

None. Five new exported test helpers in `pkg/gofr/testutil` — `WaitForHTTPServer`,
`WaitForGRPCServer`, `WaitForGRPCServerE`, `WaitForStdoutContains` and `ReserveServerPorts`; no
production code paths change.

**Additional Information:**

No new dependencies. The `using-migrations` setup uses `github.com/go-sql-driver/mysql` and
`github.com/redis/go-redis/v9` directly, both already direct dependencies of the module, and reads
the connection details through `gofr.dev/pkg/gofr/config` rather than hardcoding them.

`using-migrations` and `using-add-rest-handlers` were verified against `redis:7.0.5` and
`mysql:8.2.0` on the ports the `Example-Unit-Testing` job uses. Before the change,
`using-migrations` failed on runs 2 and 3 against the same containers and `using-add-rest-handlers`
failed on a genuinely fresh database (server not yet listening); after it, three consecutive runs
against the same containers pass.

The issue sketched a `waitForServer` local to each `main_test.go`. Fourteen copies of the same twenty
lines is a lot of surface to keep in step, and every example already imports `testutil` for
`NewServerConfigs`/`GetFreePort` — the same "wiring for example tests" role — so the helpers live
there instead. `testutil` re-declares the liveness path rather than importing `service.AlivePath`,
because `pkg/gofr/service`'s own tests import `testutil` and that would be an import cycle in the
service test binary; there is a comment saying so.

`grpc-streaming-server` waits in `TestMain`, which gets a `*testing.M` and has no `*testing.T` for
`require` to fail on, so it takes `WaitForGRPCServerE` — the error-returning half of the same wait,
sharing one unexported `waitForGRPC` with `WaitForGRPCServer` rather than duplicating it.
`using-graphql`'s bespoke `waitForReady` is dropped in favor of the shared helper.

The `E` suffix is a deliberate choice rather than a default. Two exported waits that differ only in
how they report a failure need the difference in the name, `testutil` is a package users consume, so
these names are permanent surface, and `Foo`/`FooE` is the established Go convention for exactly this
pair — a reader infers the split without reading the doc comment. `NewServerConfigs` /
`ReserveServerPorts` below is the same split under two unrelated verbs; making the package uniform
means renaming that to `NewServerConfigsE`, which is churn on settled surface and belongs in its own
change. Happy to do it here instead if you would rather the package land consistent.

`testutil.ReserveServerPorts` exists for the same reason. `development` already gives
`grpc-streaming-server`'s `TestMain` free ports instead of the `GRPC_PORT=9000` its `configs/.env`
pins, but does it with a private `reserveFreePort` copied into the test — the port-assignment half
of what `NewServerConfigs` already does, duplicated because `NewServerConfigs` takes a `*testing.T`
and `TestMain` has none. `ReserveServerPorts` is that same function for a caller with no `*testing.T`
to fail on: it returns an error instead of calling `require`, and uses `os.Setenv` rather than
`t.Setenv`, there being no test scope to restore the variables at the end of. Both now share one
implementation, and the example test drops its private copy.

`using-subscriber` was verified against the same `bitnamilegacy/kafka:3.4.1` broker this job runs,
because CI caught a real bug here first: with `WaitForHTTPServer` it failed after the full 30s
timeout even with a healthy broker, which reproduces locally on the pre-fix file and is what the
helper above fixes.

**`examples/using-s3-filestore` gets a local wait helper rather than the shared one, and must be
checked with `GOWORK=off go build ./...` from inside that directory.** It is a separate module whose
`go.mod` requires the released `gofr.dev v1.57.0` with no `replace`, and that release's `testutil`
has none of the helpers this PR adds. Calling `testutil.WaitForHTTPServer` there compiled only
because `go.work` redirects `gofr.dev` to this checkout — `GOWORK=off go vet ./...` failed on this
branch and is clean on `development`, so it was a straight regression.

Nothing in CI reports it: `go test ./examples/...` silently excludes the module, the MinIO step runs
from inside it and walks up to `go.work`, and the lint job skipped this PR. All 16 checks were green
over a module that did not build, which is why the manual command is written down here.

A `replace` was the smallest fix and is the one to avoid — `ceb50390` had just moved the gcp exporter
the other way. Reverting the file to `time.Sleep` was the other option, but it leaves one example
carrying the flake this PR exists to remove for as long as it takes the helpers to reach a release.
The local copy's comment says it is local *because* this module builds against a release, and names
the condition for deleting it. Making CI vet the workspace's submodules standalone is the real fix
and is out of scope here.

Not verified locally for want of the datastores: `http-server`, `http-server-using-redis`,
`using-publisher`, `using-http-service`, `using-s3-filestore`. They build and vet clean; CI has the
service containers.

`examples/using-html-template`'s `Test_IndexHTML` used to fail on `development` for an unrelated,
pre-existing reason (#3819); #3820 has since merged and it passes here.

This branch was rebased onto the rewritten `development` rather than merged: the force-update gave
every commit a new SHA, so the merge base fell back to the initial commit and a merge produced 72
add/add conflicts in files neither side touched. The five commits here are the same work replayed on
top of `ceb50390`. One commit was dropped in the process — the one that gave
`grpc-streaming-server`'s test free ports, now redundant against the inline version `development`
carries; the `ReserveServerPorts` refactor described above is what remains of it.

**Checklist:**

### DIFF
--- a/examples/grpc/grpc-streaming-server/main_test.go
+++ b/examples/grpc/grpc-streaming-server/main_test.go
@@ -14,6 +14,8 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/connectivity"
+	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
 
 	"gofr.dev/examples/grpc/grpc-streaming-server/server"
@@ -58,7 +60,11 @@ func TestMain(m *testing.M) {
 	grpcHost = fmt.Sprintf("localhost:%d", grpcPort)
 
 	go main()
-	time.Sleep(300 * time.Millisecond) // wait for server to boot
+
+	if err := waitForServer(grpcHost); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
 
 	os.Exit(m.Run())
 }
@@ -76,6 +82,33 @@ func reserveFreePort() (int, error) {
 	port := listener.Addr().(*net.TCPAddr).Port
 
 	return port, listener.Close()
+}
+
+// waitForServer blocks until a gRPC connection to addr is ready, so the tests below do not race
+// the server's boot. It mirrors testutil.WaitForGRPCServer, which cannot be used here: TestMain
+// gets a *testing.M and has no *testing.T to fail. A bare TCP dial would not do — the kernel
+// accepts connections from the moment the listener exists, whereas reaching the ready state means
+// the HTTP/2 handshake completed and the server is really serving.
+func waitForServer(addr string) error {
+	conn, err := grpc.NewClient(addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return fmt.Errorf("failed to create a gRPC client for %s: %w", addr, err)
+	}
+
+	defer conn.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	conn.Connect()
+
+	for state := conn.GetState(); state != connectivity.Ready; state = conn.GetState() {
+		if !conn.WaitForStateChange(ctx, state) {
+			return fmt.Errorf("gRPC server at %s did not start in time", addr)
+		}
+	}
+
+	return nil
 }
 
 func TestServerStream(t *testing.T) {
@@ -160,8 +193,9 @@ func TestBiDiStream(t *testing.T) {
 	messages := []string{"msg1", "msg2", "msg3"}
 	go func() {
 		for _, msg := range messages {
+			// No pacing needed: the server handles this stream's messages one at a time and
+			// replies in order, so the responses stay ordered regardless of send timing.
 			_ = stream.Send(&server.Request{Message: msg})
-			time.Sleep(100 * time.Millisecond)
 		}
 		_ = stream.CloseSend()
 	}()
@@ -220,10 +254,9 @@ func TestServerStream_ContextCancellation(t *testing.T) {
 
 		if !receivedFirst {
 			receivedFirst = true
-			// Cancel context to trigger cancellation handling
+			// Cancel context to trigger cancellation handling. The next Recv then fails with
+			// Canceled off the client's own context — nothing to wait for the server on.
 			cancel()
-			// Give server time to detect cancellation
-			time.Sleep(200 * time.Millisecond)
 		}
 
 		_ = resp // Use response to avoid unused variable
@@ -267,10 +300,9 @@ func TestClientStream_ContextCancellation(t *testing.T) {
 		t.Fatalf("Send failed: %v", err)
 	}
 
-	// Cancel context
+	// Cancel context. The call below fails with Canceled off the client's own context, so there
+	// is nothing to wait for the server to notice.
 	cancel()
-	// Give server time to detect cancellation
-	time.Sleep(200 * time.Millisecond)
 
 	// Try to close and receive - should get cancellation error
 	_, err = stream.CloseAndRecv()
@@ -357,10 +389,9 @@ func TestBiDiStream_ContextCancellation(t *testing.T) {
 		t.Errorf("Unexpected response: got %q, want %q", resp.GetMessage(), "Echo: test")
 	}
 
-	// Cancel context
+	// Cancel context. The call below fails with Canceled off the client's own context, so there
+	// is nothing to wait for the server to notice.
 	cancel()
-	// Give server time to detect cancellation
-	time.Sleep(200 * time.Millisecond)
 
 	// Try to receive - should get cancellation error
 	_, err = stream.Recv()
--- a/examples/grpc/grpc-streaming-server/main_test.go
+++ b/examples/grpc/grpc-streaming-server/main_test.go
@@ -40,9 +40,9 @@ func TestMain(m *testing.M) {
 
 	go main()
 
-	// AwaitGRPCServer rather than WaitForGRPCServer: TestMain gets a *testing.M and has no
+	// WaitForGRPCServerE rather than WaitForGRPCServer: TestMain gets a *testing.M and has no
 	// *testing.T for require to fail on, so it takes the error-returning variant of the same wait.
-	if err := testutil.AwaitGRPCServer(grpcHost); err != nil {
+	if err := testutil.WaitForGRPCServerE(grpcHost); err != nil {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
--- a/examples/grpc/grpc-streaming-server/main_test.go
+++ b/examples/grpc/grpc-streaming-server/main_test.go
@@ -5,9 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"net"
 	"os"
-	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -19,45 +17,28 @@ import (
 	"google.golang.org/grpc/status"
 
 	"gofr.dev/examples/grpc/grpc-streaming-server/server"
+	"gofr.dev/pkg/gofr/testutil"
 )
 
-// grpcHost is the address the example server under test listens on.
-//
-// The example's configs/.env pins GRPC_PORT=9000 for documentation purposes,
-// but the test must not inherit it: `go test ./examples/...` runs packages
-// concurrently and CI additionally binds host port 9000 (MinIO), so a fixed
-// port makes the server fail to bind and the whole package abort with
-// "gRPC port 9000 is blocked or unreachable". Reserve free ports instead —
-// the same thing testutil.NewServerConfigs does for tests that have a
-// *testing.T to attach the env cleanup to, which TestMain does not.
+// grpcHost is the address the example's gRPC server listens on, assigned in TestMain.
 var grpcHost string
 
 func TestMain(m *testing.M) {
 	os.Setenv("GOFR_TELEMETRY", "false")
 
-	grpcPort, err := reserveFreePort()
+	// Point the example at free ports rather than the fixed ones its configs/.env asks for. A
+	// fixed port is a hazard in a test: gofr.New() reports an already-taken port with Fatalf,
+	// which exits the process, so the whole package dies before a single test runs. configs/.env
+	// asks for GRPC_PORT 9000 — the port MinIO serves on, including in the CI job that runs these
+	// very tests. The system environment takes precedence over the config file, so what is set
+	// here is what the example uses.
+	configs, err := testutil.ReserveServerPorts()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "could not reserve a free gRPC port: %v\n", err)
+		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
 
-	httpPort, err := reserveFreePort()
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "could not reserve a free HTTP port: %v\n", err)
-		os.Exit(1)
-	}
-
-	metricsPort, err := reserveFreePort()
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "could not reserve a free metrics port: %v\n", err)
-		os.Exit(1)
-	}
-
-	os.Setenv("GRPC_PORT", strconv.Itoa(grpcPort))
-	os.Setenv("HTTP_PORT", strconv.Itoa(httpPort))
-	os.Setenv("METRICS_PORT", strconv.Itoa(metricsPort))
-
-	grpcHost = fmt.Sprintf("localhost:%d", grpcPort)
+	grpcHost = configs.GRPCHost
 
 	go main()
 
@@ -67,21 +48,6 @@ func TestMain(m *testing.M) {
 	}
 
 	os.Exit(m.Run())
-}
-
-// reserveFreePort asks the kernel for a free port and releases it immediately
-// so the server under test can bind it.
-func reserveFreePort() (int, error) {
-	lc := net.ListenConfig{}
-
-	listener, err := lc.Listen(context.Background(), "tcp", "localhost:0")
-	if err != nil {
-		return 0, err
-	}
-
-	port := listener.Addr().(*net.TCPAddr).Port
-
-	return port, listener.Close()
 }
 
 // waitForServer blocks until a gRPC connection to addr is ready, so the tests below do not race
--- a/examples/grpc/grpc-streaming-server/main_test.go
+++ b/examples/grpc/grpc-streaming-server/main_test.go
@@ -12,8 +12,6 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/connectivity"
-	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
 
 	"gofr.dev/examples/grpc/grpc-streaming-server/server"
@@ -42,39 +40,14 @@ func TestMain(m *testing.M) {
 
 	go main()
 
-	if err := waitForServer(grpcHost); err != nil {
+	// AwaitGRPCServer rather than WaitForGRPCServer: TestMain gets a *testing.M and has no
+	// *testing.T for require to fail on, so it takes the error-returning variant of the same wait.
+	if err := testutil.AwaitGRPCServer(grpcHost); err != nil {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
 
 	os.Exit(m.Run())
-}
-
-// waitForServer blocks until a gRPC connection to addr is ready, so the tests below do not race
-// the server's boot. It mirrors testutil.WaitForGRPCServer, which cannot be used here: TestMain
-// gets a *testing.M and has no *testing.T to fail. A bare TCP dial would not do — the kernel
-// accepts connections from the moment the listener exists, whereas reaching the ready state means
-// the HTTP/2 handshake completed and the server is really serving.
-func waitForServer(addr string) error {
-	conn, err := grpc.NewClient(addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
-	if err != nil {
-		return fmt.Errorf("failed to create a gRPC client for %s: %w", addr, err)
-	}
-
-	defer conn.Close()
-
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
-	defer cancel()
-
-	conn.Connect()
-
-	for state := conn.GetState(); state != connectivity.Ready; state = conn.GetState() {
-		if !conn.WaitForStateChange(ctx, state) {
-			return fmt.Errorf("gRPC server at %s did not start in time", addr)
-		}
-	}
-
-	return nil
 }
 
 func TestServerStream(t *testing.T) {
--- a/examples/using-cron-jobs/main_test.go
+++ b/examples/using-cron-jobs/main_test.go
@@ -5,7 +5,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"gofr.dev/pkg/gofr/testutil"
 )
@@ -19,16 +19,17 @@ func Test_UserPurgeCron(t *testing.T) {
 	configs := testutil.NewServerConfigs(t)
 
 	go main()
-	time.Sleep(1100 * time.Millisecond)
 
-	expected := 1
+	// The job is scheduled every second. Waiting on the counter it increments covers both the
+	// server's startup and the first tick, neither of which fits a fixed duration: a sleep long
+	// enough for a loaded runner also lets the job fire more than once, which the old
+	// assert.Equal(1, n) would then fail on.
+	require.Eventually(t, func() bool {
+		mu.RLock()
+		defer mu.RUnlock()
 
-	var m int
+		return n > 0
+	}, 30*time.Second, 100*time.Millisecond, "cron job did not run in time")
 
-	mu.Lock()
-	m = n
-	mu.Unlock()
-
-	assert.Equal(t, expected, m)
 	t.Logf("Metrics server running at: %s", configs.MetricsHost)
 }
--- a/examples/using-migrations/main_test.go
+++ b/examples/using-migrations/main_test.go
@@ -86,6 +86,12 @@ func setupSQL(c config.Config) error {
 // setupRedis empties the logical Redis database this test migrates into and points the example at
 // it. GoFr reads REDIS_DB (pkg/gofr/datasource/redis/config.go), and the Redis migrator runs on
 // that same client, so the selection covers the migration bookkeeping as well as the data.
+//
+// FLUSHDB, not a list of the keys the example writes: a list has to be kept in step with
+// migrations/, and the next Redis migration that writes a key nobody remembered to add would
+// silently make this example non-repeatable again. The cost is that this empties logical database
+// testRedisDB in full — safe against the throwaway container the test job runs, worth knowing
+// before pointing REDIS_HOST/REDIS_PORT at a Redis that holds anything else.
 func setupRedis(c config.Config) error {
 	client := redis.NewClient(&redis.Options{
 		Addr: net.JoinHostPort(c.Get("REDIS_HOST"), c.Get("REDIS_PORT")),
--- a/examples/using-migrations/main_test.go
+++ b/examples/using-migrations/main_test.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"strconv"
 	"testing"
 
 	_ "github.com/go-sql-driver/mysql"
@@ -23,12 +24,22 @@ import (
 const (
 	configFolder = "./configs"
 
-	// testDatabase is created and dropped by the test run. The example's own configs point at the
+	// testDatabase is the MySQL database this test owns. The example's own configs point at the
 	// shared `test` database, which every other example migrates into as well - migration
 	// bookkeeping there is keyed by version only, so the highest version any example has run wins
 	// and the others get skipped. Migrating into a database this test owns keeps the run
 	// independent of the other examples and of what previous runs left behind.
+	//
+	// It is dropped and re-created at the start of every run rather than dropped at the end, so a
+	// failing run leaves its data behind to be inspected.
 	testDatabase = "test_using_migrations"
+
+	// testRedisDB is the logical Redis database this test owns, and is the same idea as
+	// testDatabase: the migrator uses the container's client, so pointing the example at a
+	// database the test flushes first means no previous run's bookkeeping can make this one skip
+	// its migrations. Namespacing rather than deleting a known list of keys - the list would go
+	// stale, silently, the first time a Redis migration writes a key nobody thought to add to it.
+	testRedisDB = 1
 )
 
 func TestMain(m *testing.M) {
@@ -41,10 +52,8 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
-	// Redis has no per-test namespace here, so clear the keys this example writes instead. Left
-	// behind, the migration bookkeeping makes the runner skip the migrations on the next run.
-	if err := resetRedis(c); err != nil {
-		fmt.Fprintf(os.Stderr, "could not reset redis: %v\n", err)
+	if err := setupRedis(c); err != nil {
+		fmt.Fprintf(os.Stderr, "could not set up the test redis database: %v\n", err)
 		os.Exit(1)
 	}
 
@@ -74,14 +83,23 @@ func setupSQL(c config.Config) error {
 	return os.Setenv("DB_NAME", testDatabase)
 }
 
-func resetRedis(c config.Config) error {
-	client := redis.NewClient(&redis.Options{Addr: net.JoinHostPort(c.Get("REDIS_HOST"), c.Get("REDIS_PORT"))})
+// setupRedis empties the logical Redis database this test migrates into and points the example at
+// it. GoFr reads REDIS_DB (pkg/gofr/datasource/redis/config.go), and the Redis migrator runs on
+// that same client, so the selection covers the migration bookkeeping as well as the data.
+func setupRedis(c config.Config) error {
+	client := redis.NewClient(&redis.Options{
+		Addr: net.JoinHostPort(c.Get("REDIS_HOST"), c.Get("REDIS_PORT")),
+		DB:   testRedisDB,
+	})
 
 	defer client.Close()
 
-	// gofr_migrations holds the bookkeeping, gofr_migrations_lock the migration lock, and
-	// Umang is the key the Redis migration writes.
-	return client.Del(context.Background(), "gofr_migrations", "gofr_migrations_lock", "Umang").Err()
+	if err := client.FlushDB(context.Background()).Err(); err != nil {
+		return err
+	}
+
+	// The config files are read again by gofr.New(), which does not override what is already set.
+	return os.Setenv("REDIS_DB", strconv.Itoa(testRedisDB))
 }
 
 func TestExampleMigration(t *testing.T) {
--- a/examples/using-migrations/main_test.go
+++ b/examples/using-migrations/main_test.go
@@ -2,27 +2,94 @@ package main
 
 import (
 	"bytes"
+	"context"
+	"database/sql"
+	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"testing"
-	"time"
 
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/redis/go-redis/v9"
 	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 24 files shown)


examples/using-migrations/main_test.go
```diff
@@ -2,27 +2,94 @@ package main
...
+}
+
+func resetRedis(c config.Config) error {
+	client := redis.NewClient(&redis.Options{Addr: net.JoinHostPort(c.Get("REDIS_HOST"), c.Get("REDIS_PORT"))})
+
+	defer client.Close()
+
+	// gofr_migrations holds the bookkeeping, gofr_migrations_lock the migration lock, and
+	// Umang is the key the Redis migration writes.
+	return client.Del(context.Background(), "gofr_migrations", "gofr_migrations_lock", "Umang").Err()
```
aryanmehrotra: This is the one change I would ask for.

MySQL got namespaced properly above (`test_using_migrations`), but Redis falls back to deleting a hardcoded key list. GoFr already supports `REDIS_DB` - see `pkg/gofr/datasource/redis/config.go:55`, which reads it and defaults to 0 - and the migrator uses the container's client, so it inherits the selection.

```go
os.Setenv("REDIS_DB", "1")
// ... then FLUSHDB on that logical database
```

That mirrors exactly what `setupSQL` does for MySQL, and it stays correct when someone adds a Redis migration that writes a key not on this list.

As written it is not broken today - I checked, `using-migrations` is the only example with Redis migrations - but the coupling between this list and `migrations/` is invisible and unenforced. The next Redis migration silently makes the example non-repeatable again, which is precisely the failure mode this PR exists to remove. Better to close it structurally than to rely on someone remembering to update a string literal in a sibling file.

examples/using-migrations/main_test.go
```diff
@@ -2,27 +2,94 @@ package main
...
+		net.JoinHostPort(c.Get("DB_HOST"), c.Get("DB_PORT")))
+
+	db, err := sql.Open("mysql", dsn)
+	if err != nil {
+		return err
+	}
+
+	defer db.Close()
+
+	for _, query := range []string{"DROP DATABASE IF EXISTS " + testDatabase, "CREATE DATABASE " + testDatabase} {
```
aryanmehrotra, line 76: Nit: the database is created and re-created but never dropped, so `test_using_migrations` outlives the run on a developer's MySQL. I suspect that is deliberate, so a failing run can be inspected afterwards - if so, worth saying that in the comment, since the `testDatabase` doc comment above says "created and dropped by the test run" and the drop only ever happens on the *next* run.

examples/grpc/grpc-streaming-server/main_test.go
```diff
@@ -5,77 +5,76 @@ import (
...
-func reserveFreePort() (int, error) {
-	lc := net.ListenConfig{}
-
-	listener, err := lc.Listen(context.Background(), "tcp", "localhost:0")
+// waitForServer blocks until a gRPC connection to addr is ready, so the tests below do not race
+// the server's boot. It mirrors testutil.WaitForGRPCServer, which cannot be used here: TestMain
+// gets a *testing.M and has no *testing.T to fail. A bare TCP dial would not do — the kernel
+// accepts connections from the moment the listener exists, whereas reaching the ready state means
+// the HTTP/2 handshake completed and the server is really serving.
+func waitForServer(addr string) error {
```
aryanmehrotra: This is inconsistent with the rest of the PR. `ReserveServerPorts` was extracted into `testutil` specifically to kill a duplicate that existed because `TestMain` has no `*testing.T` - and then this one is left duplicated for the identical reason.

The `reserveFreePort` / `GetFreePort` / `ReserveServerPorts` split you already built is the template: an unexported `waitForGRPC(ctx context.Context, addr string) error` in `testutil`, with `WaitForGRPCServer` wrapping it in `require` and an exported error-returning variant for `TestMain` callers. Same reasoning, same shape, one implementation instead of two.

pkg/gofr/testutil/wait.go
```diff
@@ -0,0 +1,162 @@
...
+
+		buf := make([]byte, stdoutReadBuffer)
+
+		for {
+			n, err := r.Read(buf)
+
+			if n > 0 {
+				mu.Lock()
+				out.Write(buf[:n])
+				hit := strings.Contains(out.String(), substr)
```
aryanmehrotra: Nit: `out.String()` rebuilds the whole captured buffer on every read, so this is O(n^2) in the captured output.

Free on the happy path, where the signal arrives in the first read or two. On the 30s timeout path against a chatty app it is quadratic over potentially megabytes - the slow path being the slowest is a shame in a helper whose job is to fail fast with a good message.

Tracking a scan offset and searching only the new bytes plus a `len(substr)-1` carry-over from the previous chunk keeps it linear without changing the semantics.

pkg/gofr/testutil/port.go
```diff
@@ -55,10 +73,37 @@ func NewServerConfigs(t *testing.T) *ServiceConfigs {
...
+
+	return newServiceConfigs(httpPort, metricsPort, grpcPort)
+}
+
+// ReserveServerPorts is NewServerConfigs for a caller that has no *testing.T to fail on — a
+// TestMain, which gets only a *testing.M, is the usual one. It reports a failure by returning an
+// error, and sets the ports with os.Setenv rather than t.Setenv, there being no test scope to
+// restore them at the end of. Prefer NewServerConfigs wherever a *testing.T is in hand.
+func ReserveServerPorts() (*ServiceConfigs, error) {
+	ports := map[string]int{httpPortEnv: 0, metricsPortEnv: 0, grpcPortEnv: 0}
```
aryanmehrotra: Nit: a `map` over three fixed keys buys nothing and costs deterministic ordering - which port gets reserved first, and which name appears in the error, both vary per run. A slice of the three env names reads more plainly and keeps failures reproducible.

examples/using-cron-jobs/main_test.go
```diff
@@ -19,16 +19,17 @@ func Test_UserPurgeCron(t *testing.T) {
...
+	// The job is scheduled every second. Waiting on the counter it increments covers both the
+	// server's startup and the first tick, neither of which fits a fixed duration: a sleep long
+	// enough for a loaded runner also lets the job fire more than once, which the old
+	// assert.Equal(1, n) would then fail on.
+	require.Eventually(t, func() bool {
+		mu.RLock()
+		defer mu.RUnlock()
 
-	var m int
+		return n > 0
```
aryanmehrotra, line 31: Agreed with the reasoning - `assert.Equal(1, n)` was itself the flake and `n > 0` is the claim the test actually makes.

Noting only that the test now also passes if the cron fires 500 times a second, which would be a real bug in the scheduler. Not worth reintroducing timing to catch, and out of scope for an example test, but the assertion is weaker than it looks and that is worth being aware of rather than discovering later.

aryanmehrotra (changes requested):
Verified this end to end against the same service containers the `Example-Unit-Testing` job uses (`redis:7.0.5` on 2002, `mysql:8.2.0` on 2001, `bitnamilegacy/kafka:3.4.1`). The fix is real and the PR undersells it:

| --- | --- | --- |
| `using-migrations` x3, same containers | **FAIL 3/3** (`connection refused`) | **ok 3/3** |
| `grpc-unary-server/server -count=3` | **panic** | **ok** |
| full `./examples/...` x2 back to back | - | all green, both runs |
| `testutil -race -count=3` | - | ok |

Two things worth calling out beyond what the description claims. #3816 predicted `using-migrations` failing on run 2; on a normal dev machine it fails on run **1** as well, because the 100ms sleep does not cover the startup migration even against a fresh database. And `TestGoFrHealthServer_Methods` under `-count=3` does not merely fail on `development`, it panics. Both are live today and masked by `max_attempts: 2`.

I also checked the claims I was most sceptical of, and they hold: the `service.AlivePath` import cycle is real (`pkg/gofr/service/health_test.go` and `circuit_breaker_test.go` both import `testutil`); `godotenv.Load` rather than `Overload` is used for `.env`, so the `DB_NAME` override genuinely wins; `isWellKnown` exempts `/.well-known/*` from both the auth and the rate-limiter middleware, so the liveness probe is safe on `using-http-auth-middleware` and the rate-limited gRPC test; and the four deleted cancellation sleeps are correctly deleted, since the assertion is on the client's own `codes.Canceled`. Timing confirms no slowdown from the deletions: `grpc-streaming-server` is 12.9s here against 14.0s on `development`.

So the approach is right and I want this in. Requesting changes for one design point (the Redis half of the `using-migrations` fix, commented inline) plus some nits.

Two things not anchored to a line:

**No coverage of any helper's failure path.** All six new sub-cases are happy-path. The named timeout is the entire selling point over a bare connection-refused, and that message is untested. It is genuinely awkward given the `*testing.T` signature, so I am fine with this as a follow-up rather than a change here, but it is worth being deliberate about rather than leaving implicit.

**The PR does not drop `max_attempts: 2`.** The stated goal is to make it droppable. Until it actually goes, the retry keeps masking exactly what this work fixes and we never learn whether the fix held. Is that a deliberate follow-up, or an oversight? I would rather see it removed in the same PR that earns the right to remove it.

examples/using-migrations/main_test.go
akshat-kumar-singhal: Done in 7411bf5 — you are right that the list is the wrong shape for this.

`setupRedis` now flushes logical DB 1 and sets `REDIS_DB=1`, which is `setupSQL`'s shape rather than a list that has to be kept in step with a sibling directory.

Verified against `redis:7.0.5` / `mysql:8.2.0`: after three consecutive runs on the same containers, DB 0 is empty and DB 1 holds `gofr_migrations` and `Umang` — so the selection does reach the migrator, and the shared DB 0 the other examples use is left alone. `using-migrations` and `using-add-rest-handlers` pass 3/3.

examples/using-migrations/main_test.go
akshat-kumar-singhal: It is deliberate — a failing run leaving its rows behind is worth more than a clean MySQL — but you are right that the comment said otherwise. Fixed in 7411bf5: the comment now says the drop happens at the start of the next run rather than at the end of this one, and why.

examples/grpc/grpc-streaming-server/main_test.go
akshat-kumar-singhal: Fair, and the inconsistency was not intentional. Fixed in ad8b57a along exactly the split you describe: unexported `waitForGRPC(ctx, addr) error` does the work, `WaitForGRPCServer` wraps it in `require`, and `AwaitGRPCServer` is the error-returning variant `TestMain` takes. The example's private copy is gone. `./examples/grpc/... -count=2` passes.

It also made the failure-path point below cheap to close, since `waitForGRPC` takes a caller-supplied context.

pkg/gofr/testutil/wait.go
akshat-kumar-singhal: Fixed in ad8b57a — it now scans from a tracked offset with a `len(substr)-1` carry-over.

One correction to the diagnosis, which does not change the fix: `strings.Builder.String` does not copy (it is an unsafe cast over the existing buffer), so the quadratic part was the `strings.Contains` scan alone, not the rebuild. Same complexity, same remedy.

There is a case guarding the carry-over — the awaited substring split across two reads. I checked it bites: with the carry-over reverted to `from := scanned` it fails with the full 30s timeout message.

pkg/gofr/testutil/port.go
akshat-kumar-singhal: Agreed, fixed in 9b71716 — a slice of the three names, reserved in that order.

examples/using-cron-jobs/main_test.go
akshat-kumar-singhal: Agreed, and worth having on the record. `n > 0` is genuinely weaker than it looks: it asserts the job ran at all, not that it ran on schedule. Catching a runaway scheduler needs a timing assertion, which is the thing this PR is removing from example tests, and the scheduler's own tests are the right place for it rather than an example.

akshat-kumar-singhal:
Thanks for the verification — the `-count=3` panic on `TestGoFrHealthServer_Methods` and `using-migrations` failing on run **1** rather than run 2 are both better statements of the problem than what I wrote; I have not tried to restate them in the description, since your review is the record.

All six inline points are addressed and replied to individually. The two that were not anchored to a line:

**Failure-path coverage.** Not deferred after all — the `waitForGRPC` extraction you asked for above made it cheap, because the unexported function takes a caller-supplied context. `TestWaitForGRPC_Timeout` exercises it directly with a 200ms deadline instead of the 30s the exported wrappers hardcode, and asserts the error wraps `errServerNotReady` and names the address that did not come up. `TestAwaitGRPCServer` covers the new exported variant. `WaitForStdoutContains`' timeout message is still uncovered — asserting on it means actually waiting the 30s, so that one stays a follow-up rather than a 30s test in the suite.

**`max_attempts: 2`.** Oversight, not a deliberate follow-up. Dropped in 35c8754: `max_attempts: 1` on the `Example Unit Testing` job. The action itself stays, since it is what enforces `timeout_minutes`. The pkg and submodule jobs keep their retry — whatever flakiness they have is not what this branch addressed, and dropping a retry on a job I have not made deterministic would just be moving the problem.

Verification of the new commits, same containers as yours (`redis:7.0.5`, `mysql:8.2.0`):

| | result |
| --- | --- |
| `using-migrations` + `using-add-rest-handlers` x3, same containers | ok 3/3 |
| Redis DB 0 after those runs | empty; DB 1 holds `gofr_migrations`, `Umang` |
| `./examples/grpc/... -count=2` | ok |
| `testutil -race -count=2` | ok |
| `golangci-lint run pkg/gofr/testutil/...` | 0 issues |

examples/using-s3-filestore/main_test.go
```diff
@@ -63,7 +63,7 @@ func TestS3FileStore_RoundTrip(t *testing.T) {
 	t.Setenv("GOFR_TELEMETRY", "false")
 
 	go main()
-	time.Sleep(200 * time.Millisecond) // give the server time to start
+	testutil.WaitForHTTPServer(t, configs.HTTPHost)
```
aryanmehrotra: This module is in `go.work` but has its own `go.mod` requiring the released `gofr.dev v1.57.0`, with no `replace`. `WaitForHTTPServer` does not exist in that version, so this line only compiles because the workspace redirects `gofr.dev` to the local checkout:

```
$ cd examples/using-s3-filestore && GOWORK=off go vet ./...
vet: ./main_test.go:66:11: undefined: testutil.WaitForHTTPServer
```

The same command is clean on `development`.

The cheapest fix that keeps the readiness improvement is a local wait helper in this module, with a comment saying it is local *because* this module builds against a released `gofr.dev` rather than the tree. That is duplication, but it is the honest kind - it is not a copy of something this module can reach.

Otherwise: revert this one file and convert it in a follow-up once the helpers are in a release, which is a perfectly good answer too. I would avoid adding a `replace` here, since `ceb50390` just went the other way for the gcp exporter.

Whichever you pick, `GOWORK=off go build ./...` in this directory is worth a line in the PR description, because nothing in CI will tell us when this breaks again.

aryanmehrotra (changes requested):
All six inline points from the previous round are addressed, and I re-verified each one on `35c8754` rather than taking the replies at face value:

| | verified |
| --- | --- |
| `REDIS_DB` namespacing | after 3 runs, DB 0 **empty**, DB 1 holds `gofr_migrations` + `Umang` - the selection does reach the migrator and the shared DB 0 is untouched |
| `waitForGRPC` / `AwaitGRPCServer` split | private copy gone, `./examples/grpc/... -count=2` ok |
| incremental stdout scan | carry-over is correct at every boundary I could construct - `L=1` gives `from == scanned`, `scanned=0` and `L>scanned` both clamp to 0, and a match ending exactly on the first new byte is included. No off-by-one |
| deterministic port order | slice of three names |
| `using-migrations` + `using-add-rest-handlers` x3, same containers | ok 3/3 |
| `testutil -race -count=3`, full `./examples/...` x2 | all green |

`TestWaitForGRPC_Timeout` is better than I asked for - it asserts on the `errServerNotReady` sentinel *and* that the message names the address, and it runs in 0.20s, so it genuinely exercises the short deadline rather than sitting on the 30s constant. And thank you for dropping `max_attempts` in the same PR; that is what makes the rest of this falsifiable.

You are also right about `strings.Builder.String` - it is an unsafe cast over the existing buffer, not a copy, so the quadratic cost was the `Contains` scan alone. My diagnosis was wrong on the mechanism; same remedy.

So the previous round is closed. Re-requesting changes for one new item found on a fresh pass, commented inline.

**`examples/using-s3-filestore` no longer builds outside the workspace.** It is a separate module in `go.work`, which means `go test ./examples/...` - what CI runs, and what I ran twice last round - silently excludes it. Neither of us was actually testing that file. Its `go.mod` requires the released `gofr.dev v1.57.0` with no `replace`, and `WaitForHTTPServer` does not exist there:

```
# development, GOWORK=off:  clean
# this branch, GOWORK=off:  vet: ./main_test.go:66:11: undefined: testutil.WaitForHTTPServer
```

`go doc` against v1.57.0 confirms it - the published `testutil` has `GetFreePort`, `StdoutOutputForFunc`, `StderrOutputForFunc`, `CustomError`, `ServiceConfigs`, `NewServerConfigs`, and none of the five helpers this PR adds. It is a strict regression: `development` builds that module standalone, this branch does not. It works here only because `go.work` redirects `gofr.dev` to the local checkout.

No job catches it. The MinIO step runs from inside the module and walks up to `go.work`; the lint job checks out the whole repo and is `skipping` on this PR anyway. All 16 checks are green. I would not have found it either without asking why that package never appeared in my own test output.

Worth flagging that the merge base here is `ceb50390` - "require gofr.dev v1.59.0 in the gcp exporter, drop the local replace" - so the repo has just deliberately established that submodules build against released versions. This moves one submodule the other way, silently.

It is the only affected module; I checked, `using-s3-filestore` is the sole consumer of the new helpers outside the root module.

Two smaller things, neither blocking, and the second is mine to own:

- `waitForGRPC` hardcodes `serverStartTimeout` in its message, so if a caller's context expires first the error claims "did not start in 30s" when the real wait was shorter. Only reachable from the internal test today, since both exported wrappers pass the constant.
- `FlushDB` is strictly more destructive than the three-key delete it replaced - it empties a whole logical database. I suggested it, so I will own it: the risk is near-zero because the example targets a dedicated test port, but it is worth knowing that pointing this example at a real Redis now costs DB 1 entirely.
- `AwaitGRPCServer` next to `WaitForGRPCServer` is a subtle pair - "Await" vs "WaitFor" is carrying "returns an error rather than failing the test", which a reader will not infer. `testutil` is a package users consume, so these five names are permanent surface. Not asking for a change, just asking that the naming be a decision rather than a default.

examples/using-s3-filestore/main_test.go
```diff
@@ -202,6 +202,38 @@ func download(t *testing.T, host, name string) string {
...
+// waitForHTTPServer blocks until the server at host answers its liveness probe,
+// and fails the test if it never does. It replaces a fixed sleep after
+// `go main()`: a sleep long enough for a loaded CI runner is wasted on every
+// other run, and one short enough to be cheap races the listener.
+//
+// pkg/gofr/testutil has the same helper, and this is a copy of it — deliberately.
+// This directory is a separate module whose go.mod requires the released
+// gofr.dev v1.57.0 with no `replace`, and that release's testutil has no
+// WaitForHTTPServer. Calling the shared one would compile only inside the
+// workspace and break `GOWORK=off go build ./...` here. Drop this in favour of
```
aryanmehrotra: `crate-ci/typos` enforces US spelling, so this needs `favor`. It is the only red check on the PR and the only occurrence in the tree:

```
error: `favour` should be `favor`
    ./examples/using-s3-filestore/main_test.go:214:71
```

aryanmehrotra (changes requested):
Round two is closed. I re-verified each of the four on `f11c4f6f` rather than reading the replies:

| | verified |
| --- | --- |
| `using-s3-filestore` standalone | `GOWORK=off go vet ./...` and `GOWORK=off go build ./...` both clean inside the module - the regression is gone, and no `replace` was added |
| the local helper is honest duplication | the comment names the `v1.57.0` requirement as the reason it is local and the condition for deleting it; `using-s3-filestore` is still the only module outside the root consuming the new helpers, so there is nothing else to keep in step |
| elapsed wait in the gRPC timeout | reports `time.Since(start)`, so a caller-cancelled context no longer claims 30s it did not wait |
| `WaitForGRPCServerE` | no stale `AwaitGRPCServer` reference anywhere; `testutil -race -count=2` ok, `./examples/grpc/... -count=2` ok |

The `E` suffix is the right call, and the reasoning in the description is the reasoning I would have given. Leave `ReserveServerPorts` where it is - renaming settled surface for symmetry belongs in its own change, as you say. Taking the local-helper option over a `replace` is also the right half of the choice I offered, for the reason you wrote into the comment: it is not a copy of something this module can reach.

Requesting changes only because the pipeline is red, and on one word.

**`typos` fails on the comment added in `e417b13`:**

```
error: `favour` should be `favor`
    ╭▸ ./examples/using-s3-filestore/main_test.go:214:71
    │
214 │ // workspace and break `GOWORK=off go build ./...` here. Drop this in favour of
    ╰╴                                                                      ━━━━━━
```

`crate-ci/typos` enforces US spelling, and it is the only red check - the other fifteen are green. It is the single occurrence in the tree.

Push that and this has my approval. Nothing else outstanding from my side.

aryanmehrotra (approved):
Green on `42cfd562`, and the fix is the one word. All sixteen checks pass, including `typos` and the three `Example Unit Testing` legs now running at `max_attempts: 1` - which is the whole point of the branch: the job is passing without a retry to hide behind.

Re-verified `GOWORK=off go vet ./...` and `GOWORK=off go build ./...` in `using-s3-filestore` on this head; both clean, and there is no other British spelling in the tree.

Approving. Three rounds and every point closed with a verification rather than an assertion - thank you for that, and particularly for dropping `max_attempts` in the same PR, which is what makes the rest of it falsifiable.

One follow-up, not for this PR: `examples/using-ai/main_test.go` has a private `requireReady` that polls `/.well-known/alive` on a 10s deadline. It is correct, so it is not a flake - but it is a copy of `testutil.WaitForHTTPServer`, the same duplication you just removed from `using-graphql`. It was in the merge base, so it was never in #3818's list of 31. Worth folding in whenever someone next touches that file.

Umang01-hash (approved):
Re-reviewed at `b4108c9f`. All three points from the last round are resolved and I verified each rather than taking the reply:

- **Pinned s3 module** — the local `waitForHTTPServer` copy with the v1.57.0-requirement note is the right call. Confirmed `GOWORK=off go vet ./...` in `examples/using-s3-filestore` is clean, and `Config.Flavor`/`FlavorMinIO`/`New→CloudFileSystem` all exist at the pinned `s3 v0.3.0` tag.
- **using-migrations** — logical-DB namespacing (`REDIS_DB`+`FlushDB`), not a key list. Good.
- **grpc dedup** — `grpc-streaming-server` now uses `WaitForGRPCServerE`; hand-rolled wait gone.

The new helpers are solid: `testutil` passes under `-race`, and the readiness tests are non-vacuous — neutering `WaitForHTTPServer` reds the 300ms-delay case (connection refused). `WaitForStdoutContains`'s global stdout swap is safe here (no `t.Parallel()` in the changed tests). Scope is clean (verified the go.yml diff is only the retry removal, not the already-merged ls-lint/binaries hunks).

Two non-blocking notes:

- **CI retry removal** is the right direction — but with `max_attempts: 1` the residual port-reservation window in `port.go` (reserve-then-close, ×3 sequential) no longer has a net, so a rare port collision fails the build rather than retrying. The boot race itself is genuinely fixed by the waits; just worth a line in the description.
- Pre-existing/out of scope: `grpc-streaming-server` test bodies still use the deprecated `grpc.Dial`/`grpc.WithInsecure` while the new `wait.go` correctly uses `grpc.NewClient`. Worth a future cleanup.

Approving.

PR #3825, opened 2026-08-01
fix(ci): fail the submodule test job when go test fails (#3824)

### LINKED ISSUE #3824: CI: submodule test job swallows `go test` failures and never uploads coverage
## Summary

The `Submodule-Unit-Testing` job in `.github/workflows/go.yml` reports **green while `go test` has failed**, and its coverage artifact has never contained a single file. This is the same class of bug as #3812 (fixed for the example/pkg jobs by #3814), but the mechanism is different — it is not the missing `set -e`, it is the inner `bash -c` script ending on a command that always succeeds.

Three defects in one block (`.github/workflows/go.yml`, ~L313-350):

```bash
echo $SUBMODULES | jq -c '.[]' | xargs -I{} -P 4 bash -c '
  module={}
  cd $module
  module_name=$(echo $module | tr "/" "_")
  go mod download
  go mod tidy
  go test ./... -v -short -coverprofile=${module_name}.cov -coverpkg=./...
  cp ${module_name}.cov ../../../coverage_reports/
  cd -
'
```

### 1. Test failures never propagate

The inner script has no `set -e`, so a failing `go test` does not stop it, and the script's exit status is that of its **last** command — `cd -`, which always succeeds. `xargs` therefore sees every worker succeed and exits 0, and the step passes. (The outer block also lacks `set -eo pipefail`, so the `jq` side of the pipe is unguarded too.)

### 2. The coverage copy has never worked

`mkdir -p coverage_reports` runs at the repo root, but `cp` targets `../../../coverage_reports/` from inside the module. Submodules live at depth 4 (`pkg/gofr/datasource/mongo`) or 5 (`pkg/gofr/datasource/file/s3`), so `../../../` resolves to `pkg/` or `pkg/gofr/datasource/` — never the directory that was created. Every copy fails, and the upload step finds nothing.

### 3. The module path reaches the shell as unquoted text

`jq -c '.[]'` emits JSON strings *with quotes*, and `module={}` is textually substituted into the script **before** bash parses it — so the quoting works by accident rather than by construction. `jq -r` plus a positional argument (`bash -c '…' _ {}`) is what makes it safe.

## Evidence

From run [30561359578](https://github.com/gofr-dev/gofr/actions/runs/30561359578) on `development`, job `Submodule Unit Testing (v1.26)` — **conclusion: success**:

```
--- FAIL: TestContentTypeDetection_Logic (0.01s)        (and file.xml, file.yaml, file.yml)
FAIL	gofr.dev/pkg/gofr/datasource/file/azure	1.103s
--- FAIL: Test_Exec (0.00s)
FAIL	gofr.dev/pkg/gofr/datasource/scylladb	0.020s
--- FAIL: Test_Client_CreateDB (0.00s)                  (and 19 more)
FAIL	gofr.dev/pkg/gofr/datasource/arangodb	0.066s
...
Command completed after 1 attempt(s).
```

25 `--- FAIL` lines across 3 submodules, and the job is green. Alongside them, 27 of these:

```
cp: cannot create regular file '../../../coverage_reports/': Not a directory
##[warning]No files were found with the provided path: coverage_reports/*.cov. No artifacts will be uploaded.
```

## Impact

Every submodule under `pkg/gofr/datasource/**` is untested in practice — a PR can break any of them and CI will still be green. The three failures above reproduce on `development` today; two of them (`arangodb`, `scylladb`) reproduce locally with no infrastructure at all, so they are ordinary stale-test defects that have simply never been visible.

## Suggested fix

1. `set -eo pipefail` on the outer block, and `set -eo pipefail` inside the `bash -c` script so a failing `go test` aborts that worker. `xargs` then exits 123 and the step fails. Workers are independent, so all modules still run and the log still lists every failure.
2. Resolve the coverage directory to an absolute path before `xargs` and copy there.
3. `jq -r` with the module passed as a positional argument instead of substituted into the script text.

As with #3814, enabling this surfaces the pre-existing failures, which need fixing in the same PR for it to go green.

akshat-kumar-singhal:
**Description:**

Fixes #3824.

`Submodule-Unit-Testing` reports green while `go test` has failed, and its coverage artifact has
never contained a file. Same class as #3812 (fixed for the example/pkg jobs in #3814), different
mechanism — not the missing `set -e` on the outer block, but the inner `bash -c` worker.

Three defects in the one block:

1. **Failures never propagate.** The worker script has no errexit, so a failing `go test` does not
   stop it, and the script's exit status is that of its last command — `cd -`, which always
   succeeds. `xargs` sees every worker succeed and exits 0. Both the worker and the outer block get
   `set -eo pipefail`; a failing module now aborts its worker, `xargs` exits 123, and the step
   fails. Workers are independent, so every module still runs and the log still lists all failures.

2. **The coverage copy has never worked.** `mkdir -p coverage_reports` runs at the repo root while
   `cp` targets `../../../coverage_reports/` from inside the module. Submodules sit at depth 4
   (`pkg/gofr/datasource/mongo`) or 5 (`pkg/gofr/datasource/file/s3`), so that resolves to `pkg/` or
   `pkg/gofr/datasource/` — never the directory that was created. The path is now resolved to an
   absolute one before the fan-out.

3. **The module path was shell source, not data.** `xargs` substitutes `{}` into the script *text*
   before bash parses it, so `module={}` worked only because `jq -c` happened to emit quotes. Now
   `jq -r` with the module passed as a positional argument.

Evidence, from run
[30561359578](https://github.com/gofr-dev/gofr/actions/runs/30561359578) on `development`, job
`Submodule Unit Testing (v1.26)` — **conclusion: success**:

```
--- FAIL: TestContentTypeDetection_Logic (0.01s)   FAIL gofr.dev/pkg/gofr/datasource/file/azure
--- FAIL: Test_Exec (0.00s)                        FAIL gofr.dev/pkg/gofr/datasource/scylladb
--- FAIL: Test_Client_CreateDB (0.00s) (+19 more)  FAIL gofr.dev/pkg/gofr/datasource/arangodb
Command completed after 1 attempt(s).
```

plus 27 × `cp: cannot create regular file '../../../coverage_reports/': Not a directory` and
`##[warning]No files were found with the provided path: coverage_reports/*.cov`.

**The three failures this exposes, fixed here** (as #3814 did for the example job):

- **arangodb** — `mock_collection.go` was generated from `go-driver` **v2.1.6** while `go.mod` is on
  **v2.3.1**, so `MockCollection` no longer satisfies the driver's `Collection` interface (missing
  `EnsureVectorIndex`) and every test using it aborted. Regenerated with `mockgen` (+18 lines, one
  method). Separately, `arango_db_test.go` still passed 11 arguments to `RecordHistogram` — the four
  extra label arguments were dropped in bf9c0046 when `instrumentOp` was introduced — which gomock
  rejects as an argument-count mismatch. The expectations now match the call the code makes, and
  pin the `endpoint`/`type` label keys instead of `gomock.Any()`.
- **scylladb** — `Test_Exec` expected `Query(stmt, nil)`, but `Exec` forwards its variadic values,
  so a call with no values arrives as `Query(stmt)` with nothing after it. Every sibling test in the
  file already expects `Query(stmt)`.
- **azure** — `TestContentTypeDetection_Logic` pinned exact mime types for `.xml`, `.yaml` and
  `.yml`. Those come from the host's mime database, not from Go: `.xml` is `application/xml` on
  macOS and `text/xml` on an Ubuntu runner, and `.yaml`/`.yml` resolve on Ubuntu but not on macOS —
  so the test asserted the CI image's mailcap package. It now asserts what this package owns: the
  octet-stream fallback for an extension the host does not know, and the stripping of parameters
  off the returned type. Extensions treated as "known" are limited to Go's own builtin table.

None for users of the framework. For contributors, the submodule job can now fail — that is the
point of the change, and any red under `pkg/gofr/datasource/**` will block a PR from here on.

**Additional Information:**

No new dependencies. All 27 submodules pass locally with `-short`. The shell rewrite was rehearsed
against a simulated tree with modules at both depths and a stubbed failing test: all-pass exits 0
with both coverage files landing in the right directory, one-red exits non-zero while the other
module still runs to completion.

Two things found on the way that are out of scope here and not touched:

- `pkg/gofr/datasource/scylladb/scylladb.go:130` — `ExecCAS` calls `ExecCASWithCtx(ctx, dest, stmt,
  values)` instead of `values...`, so the whole slice becomes a single CQL binding value.
  `Test_ExecCAS` expects `Query(query, nil)` and passes, encoding the bug; left as-is so this PR
  stays a CI fix.
- `pkg/gofr/datasource/file/azure/storage_adapter.go` — content-type detection is inlined
  identically in three places with no helper, which is why the test above tests a copy of the logic
  rather than the code that runs.

Overlaps #3814 in `.github/workflows/go.yml`, but in a different job's block; whichever lands second
should be a trivial rebase.

**Checklist:**


Umang01-hash (dismissed):

aryanmehrotra (changes requested):
Verified the whole thing end to end. The core claim holds, and the artifact proves it in a way #3814 could not.

## The shell fix — both defects, one simulation

Reproduced the old and new blocks against a two-module tree where one worker fails:

```
OLD (development):
  cp: ../../../coverage_reports: No such file or directory
  exit=0   artifacts=0        <- red worker, green job, and no coverage

NEW (this PR):
  exit=1   artifacts=1        <- failure propagates, the passing module's file lands
```

Both bugs in one run: the exit status was `cd -`, and the relative `cp` never resolved from any module's depth.

## And the artifact exists for the first time

On this PR's own CI run (30904517619):

```
submodule-coverage-reports   51751 bytes   x3 (one per Go version)
```

On #3814's run that artifact was **absent entirely**. I downloaded it and counted:

```
27 .cov files   /   27 submodules in the repo
```

Complete, not partial. `arangodb` through `metrics_exporters_gcp`, every one.

All three `Submodule Unit Testing` jobs pass **with** `set -eo pipefail` in place, which answers the question I had about landing this: no other submodule is red, so turning on errexit does not turn the job red. The three test files here were the only failures hiding behind it.

## The three test fixes are real failures, not weakened assertions

Ran each affected submodule on both heads:

```
                development                              this PR
arangodb        FAIL  Unexpected call to RecordHistogram  ok
file/azure      ok  (on macOS — see below)                ok
scylladb        FAIL  Unexpected call to Query            ok
```

**arangodb.** The real call is seven arguments — `development` shows it verbatim:

```
RecordHistogram([context.Background app_arango_stats 147 endpoint  type createDB])
```

The expectation listed eleven, so it matched nothing. Your replacement has the right arity *and* pins `"endpoint"` and `"type"` as literals where the old one had `gomock.Any()`, so this is a stronger assertion than the one it replaces, not a looser one.

**scylladb.** Same shape — `Query([INSERT INTO Users …])` with one argument against an expectation of `Query(stmt, nil)`. Your reasoning about the variadic is right: `values...` with nothing to forward reaches the session as `Query(stmt)`, and `nil` makes gomock look for an argument that is never passed.

**azure.** This one I could not reproduce locally, and that is the point — it passes on macOS at both heads. Your explanation checks out against Go's own table:

```go
// $GOROOT/src/mime/type.go, builtinTypesLower
".xml":   "text/xml; charset=utf-8",
```

So `application/xml` was never Go's answer. The old assertion passed only because `/etc/apache2/mime.types` on macOS overrides the builtin, and `loadMimeFile` on unix overwrites what `initMime` set. The test was asserting the host's mailcap package, exactly as you say, and it is Ubuntu-only — which is why it needed the CI fix to become visible.

## mock_collection.go is in scope, and I want to say why

I opened this expecting to push back on a generated non-test file in a `fix(ci)` PR. It is not scope creep. `go.mod` is already on `v2.3.1` on `development`, but the mock's header says it was generated from `v2.1.6` — and the mock is genuinely incomplete:

```go
var _ arangodb.Collection = (*MockCollection)(nil)
```

```
development:  *MockCollection does not implement arangodb.Collection (missing method EnsureVectorIndex)
this PR:      compiles
```

So the mock had silently diverged from the driver it mocks. Nothing caught it because nothing asserts it — `go build` and `go vet` both pass on `development`. Regenerating it belongs here.

Also worth noting the `jq -c` → `jq -r` plus `_ {}` change, which the description undersells. `xargs` substitutes into the script *text* before bash parses it, so the old form passed a `jq -c`-quoted path as shell source rather than as data. Passing it positionally is the fix, and it closes an injection route through submodule paths, not just a quoting wart.

## Two one-liners before merge

### 1. `if-no-files-found: error` on the upload step

Unchanged from `development`:

```yaml
- name: Upload Coverage Reports
  uses: actions/upload-artifact@v7
  with:
    name: submodule-coverage-reports
    path: coverage_reports/*.cov
```

`actions/upload-artifact` defaults to `warn`, so an empty `coverage_reports/` logs a warning and leaves the job green. That default is a large part of why five months passed without anyone noticing the artifact had never contained a file — the `cp` failures scrolled past and the upload shrugged. This PR is specifically about a job that went quiet, and this is the line that stops it going quiet again:

```yaml
    if-no-files-found: error
```

I asked for this on #3814 as well; it is more clearly in scope here.

### 2. A compile-time assertion so the mock cannot rot again

The mock reached `v2.3.1` carrying a `v2.1.6` shape because nothing checks. One line in the package makes the next driver bump a build failure instead of a silent divergence:

```go
var _ arangodb.Collection = (*MockCollection)(nil)
```

Without it this recurs on the next bump, and the fix in this PR is a point-in-time correction rather than a fix to the class. Same argument as `if-no-files-found` — the interesting part is not repairing the current breakage, it is that neither had a guard.

## Minor, take or leave

The azure rewrite removes more than it needs to. `.txt`, `.csv` and `.js` are all in `builtinTypesLower`, and the old test asserted their *base* types (`text/plain`, `text/csv`, `text/javascript`) after stripping parameters — host-independent and correct. Only `.xml`, `.yaml` and `.yml` were the platform-dependent ones. Keeping the other three as `known: true` costs nothing and restores coverage the rewrite drops.

Worth adding a line in the comment that the host database can override *any* builtin extension, including the four you kept — that is the reason `known`/`unknown` is the right granularity rather than exact values, and it reads at the moment as though the builtin set were value-stable.

Everything else is verified. Add those two and I am happy.

akshat-kumar-singhal:
Both blockers are in, plus the minor. Pushed in 0fe1c687.

**1. `if-no-files-found: error`** — added to the submodule upload step. Agreed on the framing: the `warn` default is why 27 `cp` failures scrolled past for five months without anyone noticing, and it would have gone quiet the same way again after this fix.

**2. The mock assertion** — I took the class fix rather than the one line, in `pkg/gofr/datasource/arangodb/mock_assertions.go`. `MockCollection` was not special; it was just the one whose divergence happened to be load-bearing for a test. Every mock in the package generated from a go-driver interface now asserts:

```go
var (
	_ driver.Client                 = (*MockClient)(nil)
	_ driver.Database               = (*MockDatabase)(nil)
	_ driver.Collection             = (*MockCollection)(nil)
	_ driver.User                   = (*MockUser)(nil)
	_ driver.Graph                  = (*MockGraph)(nil)
	_ driver.GraphVertexCollections = (*MockGraphVertexCollections)(nil)
	_ driver.GraphEdgesDefinition   = (*MockGraphEdgesDefinition)(nil)
	_ driver.GraphsResponseReader   = (*MockGraphsResponseReader)(nil)
)
```

All eight compile against `v2.3.1` as of this commit, so nothing else needed regenerating — this is purely a guard. Worth noting `mock_database.go`'s header still says `v2.1.6` and `mock_interfaces.go` says `v2.3.0`; both happen to still satisfy their interfaces, which is exactly the state `mock_collection.go` was in until `EnsureVectorIndex` landed. The assertions make that a build failure next time instead of a stale header nobody reads.

`MockDatabaseGraph` is deliberately excluded, and the comment in the file says why: it mocks a gofr-local interface, so asserting it against `driver.DatabaseGraph` fails on `GetEdges` over gofr's own `GetEdgesOptions`/`EdgeDetails` types. That is a different thing from rot and should not be papered over with an assertion.

**3. The azure minor** — taken. `.txt`, `.csv` and `.js` are back as `known: true`; you are right that they are in `builtinTypesLower` and the old assertions on their base types were host-independent. Only `.xml`, `.yaml` and `.yml` were the platform-dependent ones, and the rewrite over-corrected.

Also took the comment point, and stated it as the reason for the granularity rather than a caveat on it:

```
// The known/unknown granularity is deliberate rather than a shortcut for the three extensions
// above: on unix mime.loadMimeFile overwrites whatever initMime put in the builtin table, so the
// host database can change the value of any extension listed below, not just the ones dropped.
// What is stable is that a builtin extension resolves to something other than the fallback.
```

That is the same mechanism you traced for `.xml` on macOS — `loadMimeFile` overwriting `initMime` — generalised, since it applies to all seven extensions the test keeps, not just the ones it drops.

Locally: `go build`, `go vet` and `go test ./... -short` green on arangodb, azure content-type test green across all 10 cases. `golangci-lint` on arangodb reports two `goconst` findings, both pre-existing on `development` in files this PR does not touch (`arango_document.go:85`, `mock_user.go:20`).

The `ExecCAS` variadic bug and the triplicated azure content-type detection are still untouched and still out of scope here — happy to open them as separate issues if you want them tracked rather than left in the description.

aryanmehrotra (approved):
Verified `0fe1c687`. All three in, and the class fix on the assertions is better than what I asked for.

```
if-no-files-found: error         present, with the reason in a comment
mock_assertions.go               8 assertions, go build + go vet clean
azure .txt/.csv/.js              back as known: true, all 10 cases pass
```

Checked the `MockDatabaseGraph` exclusion rather than taking the comment's word for it, since an unexplained omission from a guard file is the thing that makes the guard meaningless later:

```
vet: cannot use (*MockDatabaseGraph)(nil) as arangodb.DatabaseGraph:
     wrong type for method GetEdges
     have GetEdges(context.Context, string, string, *GetEdgesOptions) ([]EdgeDetails, error)
```

gofr's own `GetEdgesOptions`/`EdgeDetails`, exactly as the comment says. Right to leave it out, and right to say why in the file.

Your note about the other two headers checks out too — `mock_database.go` still says `v2.1.6` and `mock_interfaces.go` says `v2.3.0`, both with the `/Users/zopdev/go/...` path. They satisfy their interfaces today, so nothing to do; the assertions are what make that a build failure rather than a stale header nobody reads, which is the point.

Approving.

## Yes, file both — and both are worse than your description of them

### `ExecCAS`

Not a variadic style issue, a wrong-binding bug. `scylladb.go:130` drops the spread where `Exec` at :117 keeps it:

```go
func (c *Client) Exec(stmt string, values ...any) error {
	return c.ExecWithCtx(context.Background(), stmt, values...)   // correct
}

func (c *Client) ExecCAS(dest any, stmt string, values ...any) (bool, error) {
	return c.ExecCASWithCtx(context.Background(), dest, stmt, values)   // no spread
}
```

`ExecCASWithCtx` forwards to `session.Query(stmt, values...)`, so the nesting reaches the driver. Instrumented the mock session to print what it actually receives:

```
ExecCAS(dest, "INSERT INTO t (a,b) VALUES(?,?) IF NOT EXISTS", 1, "two")
  -> session.Query received 1 value arg(s): []interface {}{[]interface {}{1, "two"}}
```

One bind parameter that is a slice, where the statement has two placeholders. Any caller using `ExecCAS` with values gets a driver error or a wrong bind; `ExecCASWithCtx` is unaffected, so the ctx-less convenience wrapper is the only broken path. It is the sole instance in the package — I grepped the other `WithCtx` wrappers and they all spread correctly.

Worth its own issue at bug severity, not hygiene. Same class as the test you fixed, which is a neat illustration: the test asserted `Query(stmt, nil)` because nobody had looked closely at what the variadic actually carries.

### The azure content-type tests

"Triplicated" undersells it — two of the three do not assert anything at all. Both carry an `expectedCT` column that the loop never reads:

```go
// TestNewWriter_SetsContentType — 14 rows, expectedCT and description both unused
for _, tc := range testCases {
	t.Run(tc.name, func(t *testing.T) {
		writer := adapter.NewWriter(context.Background(), tc.name)
		require.NotNil(t, writer)
	})
}
```

`TestCreateNewFile_SetsContentType` is the same shape with 4 rows. Both identical on `development`, so pre-existing.

That explains the drift you fixed: the values in those dead tables are wrong *and* host-dependent — `.js` as `application/javascript` where Go's builtin is `text/javascript`, and `.yaml` as `application/x-yaml`, which is not in the builtin table at all. Nothing ever compared them, so nothing complained. If someone had wired up the assertion they would have been chasing the same mailcap problem you just traced.

The larger point for the issue: after this PR, three tests are named for setting or detecting a content type and none of them verifies that the adapter sets one. `TestContentTypeDetection_Logic` is the only one that asserts, and it asserts `mime.TypeByExtension` directly rather than anything in `storage_adapter.go`. So the fix is not "deduplicate the three" but "one of them should actually read the content type off the writer" — and the dead columns should go rather than be revived, since the value is host-dependent as your new comment now explains.

Neither of these needs to touch this PR. Good calls on both, and I would rather have them tracked than left in a description that scrolls away.

Issue #629, opened 2024-05-17
Refactor metrics implementation


srijan-27:
Extension of #529 
- A new counter metric needs to be added which represents server errors (`app_server_error`)
- A new counter metric for circuit open counts (`app_circuit_open_count`)

Current metrics implementation makes it harder to use in different packages as it requires an instance of metricsManager everywhere, need to think of a better solution and implement in such a way that a copy of metricsManager should not be required to pass everywhere.

Himanshu197200:
could you please assign this issue to me 
i would like to work on this issue as this issue is still open

om7057:
@coolwednesday @srijan-27 @Umang01-hash Picking this one up!

Umang01-hash:
Hey @om7057 ! This issue was not only to add the new counters but to think of a better solid metrics implementation for GoFr. See (Current metrics implementation makes it harder to use in different packages as it requires an instance of metricsManager everywhere, need to think of a better solution and implement in such a way that a copy of metricsManager should not be required to pass everywhere.)

Let me know if you have any ideas/thoughts of how can we improve / make it better ? Happy to help!

om7057:
> Hey [@om7057](https://github.com/om7057) ! This issue was not only to add the new counters but to think of a better solid metrics implementation for GoFr. See (Current metrics implementation makes it harder to use in different packages as it requires an instance of metricsManager everywhere, need to think of a better solution and implement in such a way that a copy of metricsManager should not be required to pass everywhere.)
> 
> Let me know if you have any ideas/thoughts of how can we improve / make it better ? Happy to help!

Thanks for the context, @Umang01-hash. 

For the threading problem: GoFr already has a precedent for this exact situation. Tracer avoids passing an instance around by using OTel's global registry (otel.Tracer("name"), set once via otel.SetTracerProvider), and the exporters subsystem has its own mutex-guarded package-level registry. I followed that same approach: a global Manager in pkg/gofr/metrics, installed once by Container, exposing package-level helpers like metrics.IncrementCounter(...). New code (or anything without easy access to Container) can call these directly, no constructor param needed.

I intentionally didn't cover the 30 datasource packages (redis, sql, mongo, kafka, etc.) that currently take a Metrics instance via constructor. Their tests are built around mocking that injected value, so removing it would mean rework on test coverage.

So this gives us the option to move away from threading metricsManager everywhere, without forcing a full migration right now. 
[...]

PR #3856, opened 2026-08-05
feat(metrics): add app_server_error/app_circuit_open_count counters + global metrics accessor

om7057:
## Summary

Addresses #629.

Adds `app_circuit_open_count`, a counter incremented once per real Closed -> Open circuit-breaker transition (not once per failing request that observes the trip - see the fix history below).

Two things from the original scope of this PR were removed after review:

* `app_server_error` - removed. It has the exact same label set as `app_http_response_count{status=~"5.."}` (path, method, status has been on `app_http_response` since before #629 was filed), so it was a fully redundant series. The equivalent query already works on development with no code change:

  ```
  sum(rate(app_http_response_count{status=~"5.."}[5m])) by (path, method)
  ```

* `pkg/gofr/metrics/global.go`, a package-level metrics accessor - removed. It shipped with no callers anywhere in the tree; the existing constructor-based dependency injection is untouched and is what every current call site still uses. A global accessor is worth revisiting as its own PR if and when it converts a real call site.

## Fix history on this branch

* `app_circuit_open_count` originally incremented on every call to `openCircuit`, which is reached from `handleFailure` whenever `failureCount > threshold`. Concurrent requests that pass the `isOpen` check before the trip all fall through to `handleFailure` and re-enter `openCircuit`, so a burst of concurrent failures recorded one increment per request instead of one per transition (measured at exactly `concurrency - threshold`, invisible to sequential testing or a `MinTimes(1)` assertion). Fixed with a `wasOpen` guard; `openCircuit` already runs under `cb.mu`, so no new synchronization is needed.
* Rebased onto current development.

## Testing

* `go build ./...`
* `go vet ./...`
* `go test ./pkg/gofr/...`
* New test: a concurrent-failure test for `app_circuit_open_count` that fails reliably against the pre-fix code (one increment per request past threshold) and passes with the fix (exactly one). The existing sequential test is pinned to an exact count instead of `MinTimes(1)`.


om7057:
Hi @aryanmehrotra @coolwednesday, adding the two requested counters (app_server_error, app_circuit_open_count) plus a small additive global metrics accessor (pkg/gofr/metrics/global.go, mirrors the existing exporters.Register/lookup pattern) so new call sites don't need Manager threaded through constructors. Left the existing datasource DI pattern untouched; happy to follow up separately if you'd rather migrate those too.

om7057:
@Umang01-hash just a small reminder on this one!

aryanmehrotra:
Reviewed this end to end — built it, ran the suites, and drove a real GoFr app against a live `/metrics` scrape rather than reading the diff. Everything below is measured; commands are at the bottom.

The PR builds, vets, and passes `./pkg/gofr/metrics/...`, `./pkg/gofr/http/middleware/...` and `./pkg/gofr/service/...` on the head. Three findings, in the order I'd act on them.

---

## 1. `app_circuit_open_count` counts failed requests, not openings

`openCircuit()` is reached from `handleFailure()` whenever `failureCount > threshold`, and `failureCount` only resets when the circuit closes. Concurrent requests that passed the `isOpen` check before the trip all fall through to `handleFailure`, so each one re-enters `openCircuit` and increments again.

Measured with a counting `Metrics` implementation, one circuit opening in every case:

| concurrency | threshold | `app_circuit_open_count` | correct |
|---|---|---|---|
| 30 | 2 | **28** | 1 |
| 50 | 5 | **45** | 1 |
| 10 | 1 | **9** | 1 |
| 100 | 3 | **97** | 1 |
| 30 sequential | 2 | 1 | 1 |

It is exactly `concurrency - threshold`, and deterministic across runs. Sequential traffic gives the right answer, which is why it looks correct in a hand test — but a burst of concurrent failures is precisely when a circuit trips, so the wrong case is the normal one.

**The test cannot see this.** `IncrementCounter(...).MinTimes(1)` passes at 1 call or 50 — I changed `openCircuit` to increment **100 times** per open and the whole `TestCircuitBreaker*` suite still passed.

Smallest fix — only count a real Closed → Open transition:

```go
func (cb *circuitBreaker) openCircuit() {
	wasOpen := cb.state == OpenState
	cb.state = OpenState
	cb.lastChecked = time.Now()

	if cb.metrics != nil {
		cb.metrics.SetGauge("app_http_circuit_breaker_state", 1, "service", cb.serviceName)

		if !wasOpen {
			cb.metrics.IncrementCounter(context.Background(), "app_circuit_open_count", "service", cb.serviceName)
		}
	}
}
```

I applied that and re-ran the table: **1 in all four concurrency configs**, and the existing `TestCircuitBreaker*`/`TestHttpService*` tests still pass. Worth pinning with an assertion on the exact count rather than `MinTimes(1)`.

To be clear, **this metric is the valuable half of the PR.** `app_http_circuit_breaker_state` is a gauge, and a gauge cannot tell you how many times it flipped when flapping is faster than the scrape interval. A counter is the right instrument; it just needs to count transitions.

---

## 2. `app_server_error` duplicates a series that already exists

`app_http_response` already carries `path`, `method` **and** `status` — `statusAttr(status)` in the same middleware, since #255. So the new counter has the same label set as the existing histogram's `_count`, restricted to 5xx.

I booted a real app, drove 5xx traffic across templated and static routes and both GET and POST, scraped `:METRICS_PORT/metrics`, and asserted the two programmatically:

```
app_server_error       {method="GET", path="/boom",      status="500"} 5
app_http_response_count{method="GET", path="/boom",      status="500"} 5
app_server_error       {method="GET", path="/item/{id}", status="500"} 40
app_http_response_count{method="GET", path="/item/{id}", status="500"} 40
app_server_error       {method="POST",path="/boom",      status="500"} 5
app_http_response_count{method="POST",path="/boom",      status="500"} 5

PASS  label sets are identical — 3 series vs 3
PASS  values are identical
```

Which means it is already available on `development`, no code required:

```promql
sum(rate(app_http_response_count{status=~"5.."}[5m])) by (path, method)
```

The cost is a new metric family whose cardinality is the 5xx subset of an existing one, added on top of it, for no new information. In the run above it was +3 series on 25; in a real service it scales with your 5xx `(path, method, status)` combinations.

Note the timing: #629 was filed in May 2024 and `status` had been a label on `app_http_response` since February 2024, so the metric was already derivable when the issue was written. That is worth flagging to whoever triages #629 rather than something you should have caught.

**On the label itself — the good news:** `path` is the route template, so 40 requests to distinct `/item/{id}` IDs collapsed to one series. I tried to find a reachable 5xx that falls back to `r.URL.Path` and could not. (Separately, unmatched routes *do* fan out — 20 unique URLs produced 20 series at `status="404"` — but that is **identical on `development`** and has nothing to do with this PR.)

---

## 3. The global metrics manager has no callers

This is the part of #629 I'd most want split out. `pkg/gofr/metrics/global.go` adds 89 lines of new **exported** API — `SetGlobal`, `Global`, and eight package-level functions — plus 75 lines of tests. The only reference anywhere in the tree is the setter:

```
pkg/gofr/container/container.go:140:  metrics.SetGlobal(c.metricsManager)
```

Nothing calls `Global()` or any helper. The middleware uses its injected `metrics` parameter; the circuit breaker uses `cb.metrics`. (The `metrics.NewHistogram` calls in the gRPC examples are a `metrics.Manager` parameter shadowing the package name, not the global.)

Proved rather than grepped: I replaced `Global()`'s body with `panic("UNREACHABLE")`, moved `global_test.go` aside, and ran the suite — **build clean, 31 packages pass, nothing panics.**

The description is candid that the refactor is deferred, and deferring it is reasonable. But the API ships now and package-level mutable state in a published library is hard to withdraw later — every future caller becomes a compatibility constraint. Two smaller worries while it has no users to shake them out: `SetGlobal` is last-wins, so a process building two containers silently points it at the second; and it makes metric recording order-dependent on container construction, which is exactly the class of bug DI was avoiding.

My suggestion: land the metrics work without `global.go`, and open the accessor as its own PR that converts at least one real call site — that is what shows the API earns its shape.

---

## Smaller things

- **`app_server_error` breaks the naming convention.** Every other counter in `registerFrameworkMetrics` ends in `_count` — `app_http_retry_count`, `app_pubsub_publish_total_count`. (Prometheus convention would want `_total`.) `app_circuit_open_count` gets this right.
- **The branch is `CONFLICTING`** and 32 commits behind `development`, conflicting in all three files it shares: `container.go`, `middleware/metrics.go`, `middleware/metrics_test.go`. Needs a rebase before it can merge.
- The 5xx branch sits before the `hasAttrer` fast path, so it correctly covers both paths — good.

---

## Commands

```sh
go build ./... ; go vet ./pkg/gofr/...
go test ./pkg/gofr/metrics/... ./pkg/gofr/http/middleware/... ./pkg/gofr/service/...   # all ok

# over-count: counting Metrics impl, N concurrent failing requests, one opening
go test -run TestProbe_OverCount ./pkg/gofr/service/

# test blindness: increment 100x per open
#   -> TestCircuitBreaker* still ok

# duplication: real app + real /metrics scrape, asserted
go run .   # boots gofr.New(), drives 5xx, compares app_server_error to app_http_response_count{5xx}

# dead code: Global() -> panic("UNREACHABLE"), global_test.go moved aside
go build ./... && go test -short ./pkg/gofr/...   # build ok, 31 packages pass, no panic
```

Happy to push the `openCircuit` fix and a count-pinning test onto the branch if that helps — say the word.

aryanmehrotra (approved):
All three findings are fixed, and the third one you fixed by deleting eighty-nine lines of exported API rather than defending it, which is the harder call of the two. I rebuilt this at `4a658137` and checked each one by trying to make it fail rather than by reading the diff.

## 1. The over-count is gone

Re-ran the exact table from the last round — a counting `Metrics`, N concurrent failures, one real opening:

| concurrency | threshold | before | now |
|---|---|---|---|
| 30 | 2 | 28 | **1** |
| 50 | 5 | 45 | **1** |
| 10 | 1 | 9 | **1** |
| 100 | 3 | 97 | **1** |
| 1000 | 3 | (not run) | **1** |

And on the wire rather than through a mock: a real GoFr app with a circuit-breakered downstream, thirty concurrent failing requests, then `GET :METRICS_PORT/metrics`:

```
outage 1   app_circuit_open_count = 1        (was concurrency-threshold)
recovery   app_http_circuit_breaker_state = 0, counter unchanged at 1
outage 2   app_circuit_open_count = 2
```

That last line is the property a `wasOpen` guard is most likely to get wrong, and it holds: the guard is not a one-way door, so a flapping circuit stays countable — which is the entire reason a counter was wanted alongside the gauge.

The guard is also safe by construction rather than by luck: `openCircuit` is only reached from `handleFailure`, which runs under `cb.mu.Lock()` (`circuit_breaker.go:77`), so the read-then-write of `cb.state` is atomic. I checked, because a `wasOpen` read outside the lock would have made the fix racy rather than correct.

**The new test earns its place.** Four mutations, all killed:

| mutation to `openCircuit` | |
|---|---|
| remove the `!wasOpen` guard — the original bug | killed |
| invert it to `if wasOpen` | killed |
| `wasOpen := true` — never count | killed |
| delete the `IncrementCounter` call | killed |

Not flaky either: 80 runs, including 40 under twelve CPU burners at `GOMAXPROCS=1`. `Interval: time.Minute` to keep the recovery goroutine out of it was the right call — that is exactly what would have made it flake.

Adversarial cases, all clean: nil `Metrics` does not panic, two independent services count `1` and `1` under their own labels, and `Threshold: 0` counts `1` rather than one per request.

## 2 and 3, verified as removed rather than assumed

- `app_server_error`: no occurrences anywhere in the tree.
- `pkg/gofr/metrics/global.go`: deleted, and `SetGlobal` / `Global()` have no remaining references.

`go build ./...` clean, and `./pkg/gofr/service/...`, `./pkg/gofr/container/...`, `./pkg/gofr/metrics/...` all green, plus `-race` on the circuit-breaker suite. `golangci-lint` reports three issues in these packages and all three are identical on `development` (`container.go` `goconst`, `health.go` `goconst` and `gocyclo`) — nothing new from this PR.

## One thing worth changing while it is still free

**`app_circuit_open_count` is the only metric in its block without the `app_http_` prefix.**

```go
c.Metrics().NewHistogram("app_http_response",               ...)
c.Metrics().NewHistogram("app_http_service_response",       ...)
c.Metrics().NewCounter  ("app_http_retry_count",            ...)
c.Metrics().NewGauge    ("app_http_circuit_breaker_state",  ...)
c.Metrics().NewCounter  ("app_circuit_open_count",          ...)   // <- the odd one
```

It sits directly beneath its own sibling gauge, describing the same object, and the two will almost always be read together — `app_http_circuit_breaker_state` for "is it open now" and the counter for "how often did it flip". Having one carry `http` and the other not is the kind of thing that costs someone a minute every time they write a query.

I know #629 names it `app_circuit_open_count`, and that is a fair reason to have used it. But the ticket was written **2024-05-17**, and `app_http_circuit_breaker_state` and `app_http_retry_count` only arrived on **2026-01-10** in #2812 — so the convention this metric now sits inside was established twenty months *after* the ticket named it. The ticket is not wrong so much as stale.

Renaming to `app_http_circuit_open_count` touches seven lines and is free today. Once released it is a breaking change to anyone's dashboard, and the usual outcome is that it never happens. Your call — but it is the last moment where it costs nothing, and if you would rather keep the ticket's name, say so and I will drop it.

## Housekeeping

The title and description are now describing a different PR: **"add app_server_error/app_circuit_open_count counters + global metrics accessor"**, when two of those three are gone. Whoever merges this will use the title as the squash commit message, so it is worth rewriting to something like `feat(metrics): count circuit-breaker openings with app_circuit_open_count`, with the description saying that the duplicate counter and the global accessor were dropped in review and why. The reasoning for both is worth keeping somewhere a future reader of #629 can find it.

Also worth noting for whoever triages #629: the two counters it asks for have now been answered differently — one added, one shown to be already derivable from `app_http_response_count{status=~"5.."}` since February 2024 — and the refactor in its third bullet is still open. That issue probably wants updating rather than closing with this PR.

Approving. The rename is the only thing I would still like, and I would not hold the merge for the title.

Umang01-hash (approved):
APPROVE — verified at head a0a4676.

The wasOpen guard is a genuine fix: without it a burst of N concurrent failures records N-threshold openings for a single Closed→Open transition. openCircuit only runs under cb.mu.Lock() (via handleFailure ← executeWithCircuitBreaker), so the state read is race-safe — confirmed go test -race clean.

Verified:
- Revert-red: removing `if !wasOpen` makes TestCircuitBreaker_OpenCircuit_CountsTransitionNotFailures fail (Times(1) → over-count). Real test, not mock-pinning.
- Metric registered before use (container_test asserts NewCounter fires once).
- Live E2E: CB pointed at a dead host, 6 failing requests at threshold 2 → `app_circuit_open_count{service="..."} 1` at /metrics, stays 1 (single transition).
- gofmt/vet/build clean; service (-race) + container suites pass; golangci-lint --new-from-rev = 0 issues.
- No breaking change to exported API.

One nit: the PR title says "add app_server_error/app_circuit_open_count counters + global metrics accessor", but only app_circuit_open_count ships — no app_server_error counter and no global metrics accessor in the diff. Please fix the title or note them as follow-ups.

PR #3858, opened 2026-08-05
feat(health): optional readiness check for /.well-known/health (#3857)

### LINKED ISSUE #3857: Optional readiness closure for /.well-known/health (SetHealthCheck)

`/.well-known/health` always responds HTTP `200`, regardless of whether the application can actually serve traffic. The body reports the aggregate status of dependencies GoFr knows about, but the status code never reflects it.

That makes the endpoint unusable as a Kubernetes **readiness** probe. During startup or a rolling deploy a pod is added to the Service endpoints immediately, and traffic is routed to it before its critical dependencies are reachable — the probe cannot say "not yet". There is also no way to gate readiness on a dependency the container does not hold: a datasource the app opened itself, a third-party SDK client, or an in-process warm-up flag.

**Describe the solution you would like**

An optional, opt-in readiness closure:

```go
type HealthCheckFunc func(ctx context.Context, c *container.Container) (status string, statusCode int)

func (a *App) SetHealthCheck(fn HealthCheckFunc)
```

The closure receives the request context and the container, and returns both the status string for the body and the HTTP status code to respond with. Because it is a closure it can also capture and probe dependencies GoFr never registered.

```go
app.SetHealthCheck(func(ctx context.Context, c *container.Container) (string, int) {
    if c.Redis == nil || c.Redis.Ping(ctx).Err() != nil {
        return "DOWN", http.StatusServiceUnavailable
    }
    return "UP", http.StatusOK
})
```

Properties this should hold to:

- **Opt-in and non-breaking.** With no closure registered the endpoint behaves exactly as today — HTTP `200` with the aggregate `{name, status}`.
- **No detail leakage.** The endpoint stays unauthenticated (see #3802), so the closure controls only the status string and the code — never per-dependency details.
- **Quiet logs.** A not-ready `503` is expected during startup and rolling deploys and Kubernetes probes it repeatedly, so it should log at `WARN`, not `ERROR`, matching the `logging.LogLevelResponder` convention already used in `http/errors.go`.

**Describe alternatives you have considered**

- *Deriving readiness from the existing aggregate status* — inferring `503` from `DEGRADED` automatically. Rejected: it is a breaking change for every existing probe, and "degraded" is often still servable. Which dependencies are critical is an application decision, not a framework one.
- *A separate `/.well-known/ready` route* — adds a second endpoint and a second probe to configure, and still needs the same application-supplied predicate.
- *Leaving it to the application to register its own route* — works, but every service reimplements the same wiring, and the conventional `/.well-known/health` path stays useless for readiness.

**Additional context**

This was originally bundled into #3807 (the #3802 redaction fix). Split out on review: it is a new public API and its own design discussion, and should not gate an urgent security fix. Implementation is proposed in a follow-up PR stacked on that branch.

akshat-kumar-singhal:
**Description:**

Closes #3857.

`/.well-known/health` always responds HTTP `200`, so it cannot serve as a Kubernetes **readiness** probe — during startup or a rolling deploy a pod joins the Service endpoints and receives traffic before its critical dependencies are reachable. There is also no way to gate readiness on a dependency the container does not hold (a datasource the app opened itself, a third-party SDK client, an in-process warm-up flag).

This adds an opt-in readiness check:

```go
type ReadinessCheck func(ctx *gofr.Context) error

func ReplaceFrameworkChecks() ReadinessOption
func (a *App) AddReadinessCheck(check ReadinessCheck, opts ...ReadinessOption)
func FrameworkReadiness(ctx *gofr.Context) error
```

Return `nil` when ready, any error when not — in which case the endpoint answers `503`. By default the check is AND-ed with GoFr's own verdict on every registered datasource and service, which is evaluated first:

```go
// GoFr's checks must pass, and so must this one
app.AddReadinessCheck(func(ctx *gofr.Context) error {
    return licenseDB.PingContext(ctx)
})

// the app's checks are the whole verdict - GoFr's are not consulted, and not run
app.AddReadinessCheck(func(ctx *gofr.Context) error {
    if redisUp(ctx) || sqlUp(ctx) {
        return nil
    }

    return errNoBackingStore
}, gofr.ReplaceFrameworkChecks())
```

Design points:

- **Opt-in and non-breaking.** With no check registered the endpoint behaves exactly as today — `200` with the aggregate `{name, status}`, `DEGRADED` included. Which dependencies are critical stays an application decision.
- **The mode is stated, not inferred.** GoFr knows which of the two policies is in force, rather than inferring it from whether a closure body happens to read an argument. That is what lets it skip the datasource sweep when its own checks no longer decide anything, and say `framework checks: replaced by the app checks` at startup instead of a line that is true only sometimes.
- **Checks accumulate.** Two independent modules can each register one; they run in registration order and the first not-ready verdict wins. Nothing is silently discarded.
- **Composition stays expressible.** `FrameworkReadiness(ctx)` returns GoFr's own verdict, so a replace-mode check can use it as one input among others — tolerating `DEGRADED` during a warm-up window, or requiring it only when its own dependency is also down. It is a function rather than a method on `Container` because `Context` embeds `*container.Container`: a method would be promoted onto every handler's `ctx`, where it would sweep every datasource in the hot path.
- **No detail leakage.** The endpoint is unauthenticated (#3802), so the check's error is logged and never written to the response — the body carries only `DOWN`. `FrameworkReadiness` reports the aggregate status (`reported DEGRADED`), not the per-dependency map, so a check that returns it unchanged cannot leak it either.
- **Nothing invalid is representable.** The `(status string, statusCode int)` pair in the issue accepted `("UP", 503)` and `("DOWN", 200)`, and reached `WriteHeader` behind only a zero-check, so a negative code panicked. With an `error` the endpoint answers `200` or `503`, and nothing else.
- **Quiet logs.** A not-ready `503` is expected during startup and rolling deploys, and Kubernetes probes it repeatedly, so `errNotReady` implements `logging.LogLevelResponder` returning `WARN`. The *request* log is a separate line keyed off the status code alone, so it still reports a probe `503` at `ERROR`; the docs now point at `LOG_DISABLE_PROBES=true` for that.
- **Registration is guarded.** Registering before `App.Run` is documented but not enforced, and the endpoint is probed at Kubernetes frequency, so the field is guarded by `App`'s existing mutex.

The endpoint path is unchanged. The API is *Readiness* throughout, matching what it controls and what the Kubernetes guide already says, but renaming `/.well-known/health` would break every deployed probe.

None. `AddReadinessCheck` is a new opt-in method; the default response and status code are unchanged.

**Additional Information:**

- No new dependencies.
- **Was stacked on #3807, now standalone.** This was originally bundled into that PR and split out on review — it is a new public API and its own design discussion, and should not have gated the urgent #3802 security fix. #3807 has since merged, so the diff here is only this change.
- **Reshaped twice on review.** An `AddReadinessCheck(name, fn)` registry was built first and abandoned: a set of independently-evaluated checks can only be combined by conjunction, so "ready if *either* Redis or SQL is up" is not representable. The second shape, `SetReadinessCheck(func(ctx, framework error) error)`, fixed that by passing GoFr's verdict in — but it left GoFr unable to know which policy a check had chosen, so the sweep ran even for a check that disclaimed it, the startup line was unconditional, and a second registration silently replaced the first. Moving the mode into an option fixes all three; exporting `FrameworkReadiness` keeps the composition the closure parameter allowed.

**Tests:**

- `pkg/gofr/handler_test.go`: table-driven `TestApp_AddReadinessCheck` (both policies against a healthy and a `DEGRADED` framework verdict, accumulation, the option on any one registration, the no-check default and a nil check), plus `TestApp_errNotReady` (the `WARN` level, the 503, and the `DOWN` body), `TestApp_healthHandler_logsReason` (the only record of *why* a probe failed), `TestApp_httpServerSetup_logsReadiness` (the wiring, not the method), `TestApp_AddReadinessCheck_concurrent` (registration against in-flight probes, under `-race`), `TestApp_FrameworkReadiness` and `TestApp_logReadiness` for both modes. The framework verdict is driven through a 500-returning `httptest` service — the same approach as the existing `TestHandler_aggregateStatus` — so the policies are asserted against a real aggregate rather than a stub.
- Mutation-checked: `LogLevel` → `ERROR`, the reason line deleted, `a.logReadiness()` deleted from `httpServerSetup`, the sweep forced to run under replace, and only the first check evaluated all turn the suite red. `health.go` is at 100% statement coverage.
- Docs: the readiness section in `docs/advanced-guide/monitoring-service-health/page.md` covers both policies, `FrameworkReadiness`, the startup and runtime log lines, `LOG_DISABLE_PROBES`, and the fact that readiness does not propagate to GoFr callers (`httpService.HealthCheck` probes `/.well-known/alive`). The Kubernetes and Helm guides' `readinessProbe` rows said this endpoint gates traffic; both now say it does so only once a check is registered, and link to the section.

**Checklist:**


aryanmehrotra:
Apologies for the silence on this one — three weeks is too long, especially with three of your other
PRs reviewed and merged around it.

I have put the substantive feedback on #3857 rather than here, because it is about the shape of the
API rather than this implementation of it: as written, `SetHealthCheck` **replaces** GoFr's dependency
checks instead of adding to them, so an app registering a closure to check Redis silently stops
checking SQL, Mongo and the rest. The suggestion there is a single method that receives the
framework's verdict, so add / override / narrow are all expressible and overriding is visible at the
call site. Worth settling that before more code goes on top of the current signature.

Two things that hold regardless of which shape wins, both reproduced against this branch:

**Data race on the new field.** `SetHealthCheck` writes `a.healthCheck` and every request reads it,
unsynchronised — `App.mu` sits one line below the field. Under `-race`, 200 concurrent sets against
200 handler calls:

```
WARNING: DATA RACE
  health.go:78   (readiness reads a.healthCheck)
  health.go:30   (SetHealthCheck writes it)
```

Calling it before `Run` is the intended usage, but nothing says so and nothing stops it — and "flip
readiness to drain before shutdown" is exactly what someone will reach for.

**The docs disagree with each other about where readiness goes.** The Kubernetes guide has it right —
`/.well-known/health` for readiness, `/.well-known/alive` for liveness. But
`docs/advanced-guide/monitoring-service-health/page.md:78` says "Wire `/.well-known/alive` into
liveness/readiness probes". Anyone who followed that has readiness pointed at an endpoint that returns
a static `UP` and can never fail, so this feature would silently do nothing for them. Worth fixing
here since this PR is the reason it starts to matter.

While reading around this I also noticed the k8s guide claims `/.well-known/health` "returns 200 only
when the service and its dependencies are reachable", which is not true today — it always returns 200.
So this PR is not adding a nice-to-have, it is making a documented contract real. Worth saying in the
description.

Holding the full review until the API shape is agreed on #3857 — most of what I had written stops
applying if the signature changes.

akshat-kumar-singhal:
Pushed the reshape from the design thread on #3857 — the API in this PR is now:

```go
func (a *App) SetReadinessCheck(check func(ctx *gofr.Context, framework error) error)
```

`SetHealthCheck` and `HealthCheckFunc` are gone. The PR description is rewritten to match; the reasoning is on [#3857](https://github.com/gofr-dev/gofr/issues/3857#issuecomment-5435751381) rather than repeated here.

What changed since the last review:

- **The framework's verdict is an argument.** The old shape replaced GoFr's dependency checks instead of adding to them — `aggregateStatus` was only reached when no closure was set, so registering one to require Redis silently stopped checking SQL, Mongo and the rest. Propagate `framework` to require both; write `_` to replace it, which is visible at the call site.
- **Returns `error`.** `("UP", 503)`, `("DOWN", 200)` and the negative-code panic through `WriteHeader` are no longer representable — the endpoint answers `200` or `503`. The zero-code and empty-status fallbacks are deleted along with the pair.
- **Takes `*gofr.Context`**, matching every other callback; it carries the container and is itself a `context.Context`.
- **The error never reaches the response.** It is logged at `WARN` and the body carries only `DOWN`. `frameworkReadiness` renders the aggregate as `reported DEGRADED` rather than the per-dependency map, so a check that returns the parameter unchanged cannot leak it back out.
- **Startup says readiness is no longer the default**, so a `503` is never a surprise about where the verdict came from.

Default behaviour is untouched: with no check registered, `200` with the aggregate `{name, status}`, `DEGRADED` included.

Two things I would rather have reviewed than decided alone, both in the description: `framework` is evaluated eagerly (so an overriding check still pays for the datasource sweep), and the startup line does not enumerate the framework's check names, because listing them without pinging needs a new exported accessor on `Container`.

Tests replaced accordingly — `TestApp_SetReadinessCheck` drives both policies against a real `DEGRADED` aggregate via a 500-returning `httptest` service, plus `TestApp_frameworkReadiness` and `TestApp_logReadiness`.

aryanmehrotra:
Reviewed end to end against #3857, then verified every claim below by running it — real GoFr apps on real ports, mutation runs, `-race`, and a coverage profile. Worktree at `df09acaf`; all 16 checks green, `BEHIND` by 3 commits but `git merge-tree` is clean.

The shape is right, and the two reshapes recorded in the description are both improvements on what the issue asked for. The findings are test gaps and two decisions worth stating out loud, not correctness bugs on the happy path.

---

## 1. "Quiet logs" is an acceptance criterion in #3857, and it is half-delivered — and untested

`errNotReady.LogLevel()` is at **0.0% coverage**:

```
gofr.dev/pkg/gofr/health.go:74:  LogLevel  0.0%
```

Mutating it to `return logging.ERROR` leaves the entire `pkg/gofr` suite green (`ok gofr.dev/pkg/gofr 5.5s`). `errNotReady` is referenced nowhere outside `pkg/gofr`, so that suite is the whole test surface — nothing anywhere pins the level the issue asked for.

And on the wire it is only half true. `middleware/logger.go:267` classifies **any** 5xx as `ERROR`, and this PR is what puts a 5xx on that path. One app, one always-failing check, six probes:

| `LOG_DISABLE_PROBES` | ERROR lines | WARN lines |
|---|---|---|
| `false` (default) | **6** | 12 |
| `true` | **0** | 12 |

```
{"level":"ERROR",...,"uri":"/.well-known/health","response":503}
```

So the `LogLevelResponder` does its job — the handler's own error log is `WARN` — but the request log is `ERROR` per probe by default, on stderr, which is exactly what an alert keyed on level picks up during a rolling deploy. The cure already exists and the new docs section does not mention it.

**Fix:** one assertion that `errNotReady{}.LogLevel() == logging.WARN`, and one sentence in *What you see at startup and at runtime* pointing at `LOG_DISABLE_PROBES=true`.

## 2. The reason line is the only record of *why*, and nothing tests it

Deleting `c.Warnf("readiness: not ready: %v", err)` (health.go:95) leaves the suite green. That line is load-bearing for the whole no-leak design — the body is deliberately just `DOWN`, so if it ever regresses an operator has a 503 and no reason anywhere.

Related, from the same run: a failed probe emits **two** WARN lines, and only one has content —

```
{"level":"WARN","message":"readiness: not ready: framework dependency checks: reported DEGRADED",...}
{"level":"WARN","message":{"trace_id":"...","error":"DOWN"}}
```

the second being `logError` re-reporting the redacted message. Worth a look while you are adding the assertion.

## 3. The startup announcement is not wired-tested

Deleting `a.logReadiness()` from `httpServerSetup` (gofr.go:183) also leaves the suite green — `TestApp_logReadiness` calls the method directly, so it tests the function, not the wiring. One assertion in a test that already boots an app would close it.

## 4. Eager evaluation, measured against this repo's own Kubernetes guide

`docs/guides/deploying-to-kubernetes/page.md:143` ships `readinessProbe: periodSeconds: 5, timeoutSeconds: 2`.

An app with three slow dependencies and the emptiest possible overriding check — `func(*gofr.Context, error) error { return nil }` — answers in **6.0s**:

```
PASS  200 UP — 200 {"data":{"name":"app-8403","status":"UP"}}
PASS  probe latency — 6s for a check whose body is `return nil` (3 deps x 2s, swept sequentially)
```

Same 6s on the default path, so the description is right that it is not a regression *there*. But the override path is new surface and this is the case it exists for: the check disclaimed those dependencies and the pod still leaves the endpoints, because the sweep it did not ask for blew the guide's own 2s timeout. It does not take three contrived deps either — `Container.Health` is sequential, and `SQL.HealthCheck()` / `Redis.HealthCheck()` (`container/datasources.go:28,36`) take no context at all, so nothing bounds them; only HTTP services get the 5s cap in `service/health.go:34`.

The lazy-vs-eager trade is recorded in the description, so this is a "confirm and write it down" rather than a rethink: either take `func() error`, or add one line to the eager-evaluation note saying the sweep must fit inside `timeoutSeconds`.

## 5. The aggregate is computed, then thrown away on the ready path

An app with a `DOWN` dependency and an overriding check:

```
200 {"data":{"name":"app-8406","status":"UP"}}
```

`frameworkReadiness` has already called `aggregateStatus` one line earlier, so `"DEGRADED"` exists in the frame and is discarded for a hardcoded `statusUp`. The consequence is that adopting a readiness check silently costs you the degradation signal your dashboards read today — `200` and `status` could carry the two answers independently (`200` = ready, `status` = what the sweep actually found).

It is clearly deliberate — the constant is hardcoded and `TestApp_SetReadinessCheck` pins it (swapping in `aggregateStatus(c)` fails the *overriding check ignores a degraded framework* row) — so this is a question, not a defect. Was the alternative considered?

## 6. Minor — the setter races, by design

`go test -race`, registering while probes are in flight:

```
WARNING: DATA RACE
  write  health.go:54  (SetReadinessCheck)
  read   health.go:90, health.go:94  (healthHandler)
```

"Register before `App.Run`" is documented, and the docs example already teaches the right runtime pattern (capture an `atomic.Bool`, flip that, not the check) — so this is consistent and not a bug. But `App` already carries `a.mu`, and guarding one write and one read on a path that runs at probe frequency is free. Worth deciding explicitly rather than by omission.

## 7. Minor — two guides now say something conditionally true

`docs/guides/deploying-to-kubernetes/page.md:26` — "`/.well-known/health` → `readinessProbe.httpGet` — Gate traffic until datasources are reachable" — and the identical probe in `docs/guides/helm-chart-starter/page.md:193`. That row was simply wrong before this PR and is right only *if* a check is registered. One cross-link each into the new section.

Also: the checklist says "All new code is covered by unit tests"; `LogLevel` is at 0%.

---

- **A separate `/.well-known/ready` path.** #3857 considered it and rejected it with reasons; the PR is aligned and I am not reopening it.
- **`(status string, statusCode int)` → `error`.** Strictly better than what the issue proposed — `("UP", 503)` and a panicking negative code stop being representable. The argument in the description holds.
- **Rejecting an `AddReadinessCheck(name, fn)` registry.** The OR-composition argument is sound: independently-evaluated checks can only be AND-ed, and expressing "ready if *either* store is up" means dropping the framework verdict, which collapses the registry to one entry.
- **No leakage — verified on the wire.** Body is exactly `{"error":{"message":"DOWN"}}`; the dependency name `secret-payments-db`, the word `framework` and the string `DEGRADED` are all absent; `Content-Type: application/json`.
- **Default path unchanged** — `200 {"data":{"name":"app-8401","status":"DEGRADED"}}` with no check registered.
- **`/.well-known/alive` unaffected** — still `200 {"data":{"status":"UP"}}` on an app answering 503 on `/health`.
- **No cascade into other GoFr apps.** An app that `AddHTTPService`s this one still reports `UP`, because `httpService.HealthCheck` probes `/.well-known/alive` (`service/health.go:25`). Worth knowing rather than fixing: GoFr's own service-to-service checks are liveness, so they ignore readiness by construction.
- `gofmt` clean, `golangci-lint --new-from-rev` reports 0 issues on the diff.

Happy to push the three test additions (WARN level, the reason line, the `logReadiness` wiring) if that is useful — say the word.

aryanmehrotra:
Second pass, on the API shape rather than the implementation — this is the part worth settling before merge, and it absorbs finding 4 from my earlier comment.

## The shape

```go
type ReadinessCheck func(ctx *gofr.Context, framework error) error
func (a *App) SetReadinessCheck(check ReadinessCheck)
```

One method covering both policies is the right instinct, and the description argues it well: composing and overriding really are one decision, and `_` really does make the override visible at the call site.

But the *mode* is inferred from whether the closure body happens to read a parameter. The framework never learns which policy is in force, and three things follow from that — all measured.

### 1. GoFr cannot skip work it does not need — 6.0s vs 400µs

`frameworkReadiness(c)` has to be evaluated eagerly precisely because GoFr cannot know whether the check wants it. I prototyped the alternative — same file, same handler, mode carried as an option instead of inferred — and measured both paths on an app with three 2s dependencies:

| | probe latency |
|---|---|
| override, current shape (sweep runs anyway) | **6.0032s** |
| override, explicit mode (sweep skipped) | **400µs** |
| compose, explicit mode (sweep runs, correctly) | 6.0017s |

Reproduced twice, consistent. The lazy-vs-eager trade recorded in the description is real, but it is a *consequence* of the implicit mode, not an independent decision — make the mode explicit and the dilemma disappears rather than getting resolved. That matters against this repo's own `readinessProbe: timeoutSeconds: 2` in `docs/guides/deploying-to-kubernetes/page.md:143`.

### 2. The startup line cannot tell the truth

```
readiness: app check registered - framework checks: enabled, propagated to the app check
```

is emitted unconditionally. An app with a `DOWN` dependency and an overriding check logs exactly that line and then answers `200 {"data":{"name":"app-8406","status":"UP"}}`. "Propagated" is literally true — it is passed as an argument — but an operator reads it as "GoFr's checks gate readiness", and they do not. The line cannot be fixed without the framework knowing the policy, which is the same root cause.

### 3. Two registrations silently collapse to one

```go
a.SetReadinessCheck(func(*gofr.Context, error) error { return errFirst })
a.SetReadinessCheck(func(*gofr.Context, error) error { return nil })
```

→ `200 {"data":{"name":"gofr-app","status":"UP"}}`. The first check never runs, and nothing is logged. That is the normal shape once two independent modules each care about readiness — a cache warmer and a license store, say. `AddHTTPService` 200 lines away in the same file does warn on this:

```go
if _, ok := a.container.Services[serviceName]; ok {
    a.container.Debugf("Service already registered Name: %v", serviceName)
```

Related: `SetReadinessCheck` is the **only** `Set*` method on `App`. The other 30 registration verbs are 27 `Add*` and 3 `Use*`, and all of them accumulate.

## What I would suggest instead

Still one method, with the mode stated rather than inferred:

```go
// ReadinessCheck reports whether the application is ready to serve traffic.
type ReadinessCheck func(ctx *gofr.Context) error

// ReplaceFrameworkChecks makes the registered checks the whole verdict: GoFr's own
// dependency sweep is not run at all, so it costs nothing.
func ReplaceFrameworkChecks() ReadinessOption

// Default: AND-ed with GoFr's own dependency checks.
func (a *App) AddReadinessCheck(check ReadinessCheck, opts ...ReadinessOption)
```

- `opts ...Option` on an `App` registration method is already the house pattern — `AddHTTPService(serviceName, serviceAddress string, options ...service.Options)`, `gofr.go:246`.
- The safe policy is the default; the one that drops GoFr's checks is the one you have to type. Today both are one keystroke apart (`framework` vs `_`).
- `Add` accumulates, so two modules compose instead of one silently winning — and the OR-composition objection that killed the earlier registry does not apply, because `ReplaceFrameworkChecks()` is right there for the "either store is up" case.
- The second parameter leaves the closure entirely: `func(ctx *gofr.Context) error` reads like every other GoFr handler.

**What that costs, honestly:** the current signature allows *arbitrary* composition of the framework verdict — "tolerate `DEGRADED` for the first 30 seconds", or "not ready only if the framework is down *and* my own dep is too". A boolean mode cannot express those, and `aggregateStatus` is unexported so a replace-mode check could not reach the verdict at all. If you judge that expressiveness worth keeping, that is a defensible call and I would not block on it — the comment in `aggregateStatus` about not promoting `Health` onto every `ctx` cuts the same way.

**If you keep the current shape**, the two cheap fixes stand on their own:

1. Make the startup line conditional, or reword it to what GoFr actually knows — something like `readiness: app check registered - framework verdict is passed to it; whether it gates is up to the check`.
2. Log on re-registration, matching `AddHTTPService`.

Neither changes the signature, and both close the gap between what the endpoint does and what the logs say it does.

aryanmehrotra (changes requested):
Reviewed end to end against #3857 — read the issue before the diff, then verified every claim below by running it: real GoFr apps on real ports, mutation runs, `-race`, and a coverage profile. Worktree at `df09acaf`; all 16 checks green, `BEHIND` by 3 commits but `git merge-tree` is clean.

The work is genuinely careful. The `(status string, statusCode int)` → `error` reshape is strictly better than what the issue proposed, the reasoning that killed the named-registry idea holds, and the no-leak property is real — I confirmed it on the wire. Thank you for bearing with another round.

I am requesting changes on one narrow thing: **the signature is the only part of this PR that cannot be changed later.** Tests, log wording, docs cross-links, even eager-vs-lazy evaluation are all follow-up material. `ReadinessCheck func(ctx *gofr.Context, framework error) error` is public API — once it ships, every change to it is a breaking change. So it seems worth one more round now, and none later.

---

# The ask

Keep it one method; move the mode out of the closure's parameter list and into the call:

```go
type ReadinessCheck func(ctx *gofr.Context) error

func ReplaceFrameworkChecks() ReadinessOption

func (a *App) AddReadinessCheck(check ReadinessCheck, opts ...ReadinessOption)
```

```go
app.AddReadinessCheck(fn)                                 // AND-ed with GoFr's checks (default)
app.AddReadinessCheck(fn, gofr.ReplaceFrameworkChecks())  // the check is the whole verdict
```

Same two policies you already support. The difference is that GoFr *learns* which one is in force, instead of inferring it from whether the closure body happens to read an argument.

## What it buys

**1. GoFr can stop doing work nobody asked for.** Measured on an app with three slow dependencies, override policy both ways:

| | probe latency |
|---|---|
| current shape (sweep runs anyway) | 6.0032s |
| explicit mode (sweep skipped) | **400µs** |

Only under failure — a healthy sweep is milliseconds, so normal operation is unchanged. But that is exactly when a readiness probe earns its keep, and `docs/guides/deploying-to-kubernetes/page.md:143` ships `timeoutSeconds: 2`, so an override check whose body is `return nil` still loses the pod over dependencies it explicitly disclaimed. A lazy `framework func() error` would also get this row, so this one alone is not the argument.

**2. The startup line can say something true.** It currently prints `framework checks: enabled, propagated to the app check` unconditionally — and an app with a `DOWN` dependency and an overriding check emits exactly that, then answers `200 {"data":{"name":"app-8406","status":"UP"}}`. With the mode known it becomes `framework checks: replaced by app check`, and an operator debugging *why is this pod ready with Redis down* reads the answer at startup.

**3. Two modules can each register.** Today the second `SetReadinessCheck` silently wins and the first never runs, with nothing logged:

```go
a.SetReadinessCheck(func(*gofr.Context, error) error { return errFirst })
a.SetReadinessCheck(func(*gofr.Context, error) error { return nil })
// -> 200 {"data":{"name":"gofr-app","status":"UP"}}
```

A cache warmer and a license store both wanting a say is the ordinary case. `AddHTTPService` already warns on this shape (`gofr.go:251`).

**4. Turning off dependency gating becomes greppable.** Today it is spelled `_`. In review, `func(ctx *gofr.Context, _ error) error` reads as an unused parameter rather than "this service no longer gates readiness on any datasource" — and that is a thing an SRE should be able to find across a codebase. `gofr.ReplaceFrameworkChecks()` is a token; an underscore is not.

**5. It lands where the rest of the API already is.** `opts ...Option` on an `App` registration method is the house pattern — `AddHTTPService(serviceName, serviceAddress string, options ...service.Options)`, `gofr.go:246`. And `SetReadinessCheck` is currently the only `Set*` on `App`; the other thirty verbs are 27 `Add*` and 3 `Use*`, all accumulating. It also makes the safe policy the default and the sharp one opt-in, where today they are one keystroke apart.

## Where you may well be right and I am not

This costs **arbitrary composition**. `func(ctx, framework error) error` lets a check say "tolerate `DEGRADED` for the first thirty seconds", or "not ready only if the framework is down *and* my own dependency is too". A boolean mode expresses neither, and since `aggregateStatus` is unexported a replace-mode check could not reach the verdict at all. Your own comment in that function — about not promoting `Health` onto every `ctx` — cuts against the obvious escape hatch, so the loss is real rather than hypothetical.

If you weigh that expressiveness above the five points, say so and I will withdraw the request. In that case I would still ask for two much smaller things, neither touching the signature:

1. Make the startup line conditional, or reword it to what GoFr actually knows.
2. Log on re-registration, matching `AddHTTPService`.

---

# Findings that stand either way

### The "quiet logs" criterion from #3857 is half-delivered, and untested

`errNotReady.LogLevel()` is at **0.0% coverage**, and mutating it to `return logging.ERROR` leaves the whole `pkg/gofr` suite green (`ok gofr.dev/pkg/gofr 5.5s`). `errNotReady` is referenced nowhere outside `pkg/gofr`, so that suite is the entire test surface — nothing pins the level the issue asked for.

On the wire it is also only half true: `middleware/logger.go:267` classifies **any** 5xx as `ERROR`, and this PR is what puts a 5xx on that path. One app, one always-failing check, six probes:

| `LOG_DISABLE_PROBES` | ERROR | WARN |
|---|---|---|
| `false` (default) | **6** | 12 |
| `true` | **0** | 12 |

So the `LogLevelResponder` does its job — the handler's own error log is `WARN` — but the request log is `ERROR` per probe by default, on stderr, which is what a level-keyed alert picks up during a rolling deploy. The cure exists and the new docs section does not mention it. One assertion, plus one sentence pointing at `LOG_DISABLE_PROBES=true`.

### The reason line is the only record of *why*, and nothing tests it

Deleting `c.Warnf("readiness: not ready: %v", err)` (`health.go:95`) leaves the suite green. That line carries the whole no-leak design — the body is deliberately just `DOWN`, so if it regresses an operator has a 503 and no reason anywhere. Related, from the same run: a failed probe emits **two** WARN lines and only one has content, the second being `logError` re-reporting the redacted message.

### The startup announcement is not wired-tested

Deleting `a.logReadiness()` from `httpServerSetup` (`gofr.go:183`) also leaves the suite green — `TestApp_logReadiness` calls the method directly, so it tests the function, not the wiring.

### Minor — the setter races, by design

`go test -race`, registering while probes are in flight: `WARNING: DATA RACE`, write `health.go:54` against reads `health.go:90` and `:94`. "Register before `App.Run`" is documented and the docs example teaches the right runtime pattern, so this is consistent rather than a bug — but `App` already carries `a.mu`, and guarding one write and one read on a path that runs at probe frequency is free.

### Minor — readiness does not reach GoFr callers, and the docs should say so

`httpService.HealthCheck` probes `/.well-known/alive` (`service/health.go:25`), so a GoFr app that `AddHTTPService`s a not-ready GoFr app still reports it `UP`. `SetReadinessCheck` therefore gates the load balancer and nothing else — worth one sentence in the new docs section, because "register a readiness check" reads like it would propagate.

I checked whether the probe itself should move to `HealthPath` instead, and it should not. B is a healthy GoFr app that happens to have slow dependencies; A depends on B:

| `httpService.HealthCheck` probes | A's verdict on B | |
|---|---|---|
| `/.well-known/alive` (today) | `200 {"status":"UP"}` in 0s | correct — B serves `/ping` in 0s |
| `/.well-known/health` | `200 {"status":"DEGRADED"}` in **5s** | false, and it chains |

The second is the `defaultTimeout` cap in `service/health.go:34` firing on B's own 6s dependency sweep. Pointing dependency checks at `/health` would make every upstream re-run its downstream's full sweep, and any downstream slower than 5s would be reported down while serving traffic normally — then inherited by whatever depends on *that*. So `/alive` is right by construction, not by inertia (there is no rationale in its history — it has been that way since #777). Only the docs gap is worth closing.


### Minor — two guides now say something conditionally true

`docs/guides/deploying-to-kubernetes/page.md:26` ("`/.well-known/health` → `readinessProbe.httpGet` — Gate traffic until datasources are reachable") and the identical probe in `docs/guides/helm-chart-starter/page.md:193`. That row was simply wrong before this PR and is right only *if* a check is registered. One cross-link each. Also, the checklist says "All new code is covered by unit tests"; `LogLevel` is at 0%.

---

# Differences I am not flagging

- **A separate `/.well-known/ready` path.** #3857 considered and rejected it with reasons; the PR is aligned and I am not reopening it.
- **`(status string, statusCode int)` → `error`.** Strictly better than the issue's proposal — `("UP", 503)` and a panicking negative code stop being representable.
- **Rejecting a named `AddReadinessCheck(name, fn)` registry.** The OR-composition argument is sound.
- **No leakage — verified on the wire.** Body is exactly `{"error":{"message":"DOWN"}}`; the dependency name `secret-payments-db`, the word `framework` and the string `DEGRADED` are all absent; `Content-Type: application/json`.
- **Default path unchanged** — `200 {"data":{"name":"app-8401","status":"DEGRADED"}}` with no check registered.
- **`/.well-known/alive` unaffected** — still `200 {"data":{"status":"UP"}}` on an app answering 503 on `/health`.
- `gofmt` clean; `golangci-lint --new-from-rev` reports 0 issues on the diff.

Happy to push whichever direction you choose — the reshape is about forty lines and I have it prototyped and measured, and the three test additions are small.

akshat-kumar-singhal:
Taking the reshape. You put the case on the one ground that settles it — the signature is the only part of this that cannot be changed later — and the four consequences you measured are all consequences of the same thing: passing the verdict *in* let a check choose its policy but left GoFr unable to *know* which policy was chosen. Inferring it from whether a closure body happens to read an argument was the defect, not the sweep or the log line.

Pushed:

```go
type ReadinessCheck func(ctx *gofr.Context) error

func ReplaceFrameworkChecks() ReadinessOption
func (a *App) AddReadinessCheck(check ReadinessCheck, opts ...ReadinessOption)
```

with one addition, which is the part I'd like you to look at:

```go
func FrameworkReadiness(ctx *gofr.Context) error
```

## Why the extra export

You named the cost yourself and were right that it is real rather than hypothetical: a boolean mode cannot say "tolerate `DEGRADED` for the first thirty seconds" or "not ready only if the framework is down *and* mine is", and with `aggregateStatus` unexported a replace-mode check could not reach the verdict at all. That second clause is what makes the loss permanent, and it is also what makes it cheap to undo — the verdict was already computed and already redacted, it just had no name.

So `ReplaceFrameworkChecks()` plus an explicit `FrameworkReadiness(ctx)` inside the check expresses everything the old signature could, and the framework stops running a sweep nobody asked for:

```go
app.AddReadinessCheck(func(ctx *gofr.Context) error {
	if err := gofr.FrameworkReadiness(ctx); err != nil && time.Since(started) > warmup {
		return err
	}

	return nil
}, gofr.ReplaceFrameworkChecks())
```

It is a function, not a method on `Container`, for exactly the reason `aggregateStatus` was unexported — `Context` embeds `*container.Container`, so a method is promoted onto every handler's `ctx` and sweeps every datasource in the hot path. It returns the aggregate status only, so a check that propagates it unchanged still cannot leak; the no-leak property you verified on the wire is unchanged.

I make that a fair export trade rather than a reversal of #3807: it exports a *verdict* that is already redacted for exactly this endpoint, not an accessor onto the container's inventory. But it is a new name in the public API on a round that was about not shipping one we'd regret, so say if you'd rather it wait — the option shape stands on its own and this can land separately.

## The other decisions in the reshape

**`ReplaceFrameworkChecks` is a property of readiness, not of one registration.** Once checks accumulate, per-check modes have no coherent meaning — the framework verdict either gates the endpoint or it does not. Passing the option on any one call turns gating off for the endpoint, and the startup line says which mode is in force, so the thing that would otherwise be implicit is announced:

```
readiness: 1 app check(s) registered - framework checks: enabled, evaluated before the app checks
readiness: 2 app check(s) registered - framework checks: replaced by the app checks
```

**Re-registration no longer needs a warning** — your point 3 is fixed structurally rather than logged. Checks accumulate in registration order and the first not-ready verdict wins.

**Nothing is lost from the eager-evaluation defence in the PR body, because the defence was wrong.** "Not a regression, the default path sweeps anyway" is true and beside the point you made: the probe that most needs to answer fast is the one whose dependencies are hanging. Under `ReplaceFrameworkChecks` the sweep is skipped entirely now.

## The findings that stood either way — all four done

- **`LogLevel` at 0%.** `TestApp_errNotReady` pins `WARN`, the 503, and the `DOWN` string. Mutating it to `logging.ERROR` now fails.
- **The request log is `ERROR` per probe.** Your table is the behaviour of `handleRequestLog`, which keys off the status code alone, so `LogLevelResponder` cannot reach it. The docs section now says so and points at `LOG_DISABLE_PROBES=true`, noting the `WARN` line carrying the reason is unaffected.
- **The reason line was untested.** Deleting `c.Warnf("readiness: not ready: %v", err)` now fails `TestApp_healthHandler_logsReason`. I left the second, contentless WARN alone — that is `logError` re-reporting the redacted message for every handler error, not something this endpoint introduces; happy to file it separately.
- **The startup announcement was not wired-tested.** `TestApp_httpServerSetup_logsReadiness` drives `httpServerSetup` rather than the method, so deleting the call fails.
- **The race.** Guarded with the existing `a.mu` — one write in `AddReadinessCheck`, one read in `healthHandler`. `TestApp_AddReadinessCheck_concurrent` registers while probes are in flight, so `-race` catches the guard being dropped.
- **Docs.** The new section states that readiness does not reach GoFr callers and why `/.well-known/alive` is right by construction (your A-depends-on-B table is the argument, and it is a better one than "it has been that way since #777"). The Kubernetes guide's mapping row and the Helm chart's bullet both now say the `readinessProbe` gates nothing until a check is registered, each linking to the section.

Mutation-checked rather than assumed — `LogLevel` → `ERROR`, the warn line deleted, `a.logReadiness()` deleted from `httpServerSetup`, the sweep forced to run under replace, and only the first check evaluated: all five now turn the suite red. `health.go` is at 100% statement coverage. `-race` is clean on the readiness tests (the pre-existing `TestContext_WriteMessageToSocket` race is unrelated and reproduces on `development`).

Thank you for the two rounds — the second one changed the shape of the API for the better, and both times on an argument I could check rather than a preference.

aryanmehrotra (dismissed):
Approving. The reshape is right, and you closed the expressiveness gap I raised without me having to ask — the API now says what it does instead of inferring it. One three-line fix I would like before merge and two nits, none of which is worth another round: I have written and verified the fix and will push it if you want.

## The ruling you asked for: `FrameworkReadiness` lands here, not separately

Land it in this PR.

The argument that settled the last round was that the signature is the only part of this that cannot be changed later. Splitting the export applies that argument to the wrong half: `ReplaceFrameworkChecks` without a way to consult the verdict is a knowingly less expressive API, and the people who adopt it first are the ones who write around the missing piece. Then it widens next release, and we have shipped the two-step we spent a round avoiding.

It also exposes nothing new. It is the same value the old `framework error` parameter pushed in, pulled instead of pushed, and already redacted for exactly this endpoint. I checked that on the wire rather than in the doc comment — a replace-mode check that returns it completely unchanged:

```
replace check propagating FrameworkReadiness -> 503
body = {"error":{"message":"DOWN"}}      # no status, no host, no dependency name
```

And the reason for a function rather than a `Container` method is correct — `Context` embeds `*container.Container`, so a method would be promoted onto every handler's `ctx`. An import-qualified package call is a deliberate act; an autocompletion on `ctx.` is not.

## What I verified

**Mutation, ten of them.** Your five, plus five of my own:

| mutation | |
|---|---|
| `LogLevel` `WARN` → `ERROR` | killed |
| delete the `c.Warnf` reason line | killed |
| delete `a.logReadiness()` from `httpServerSetup` | killed |
| force the sweep to run under replace | killed |
| evaluate only the first check | killed |
| `frameworkReplaced`: ANY → ALL | killed |
| drop the mutex in `AddReadinessCheck` (under `-race`) | killed |
| ready body `statusUp` → `aggregateStatus(c)` | killed |
| remove the nil-check guard | killed |
| **run app checks before the framework checks** | **survived** |

**End to end, eight live GoFr apps on real ports with real HTTP dependencies, driven over the wire** — twenty assertions, all passing: the default path is byte-for-byte unchanged, both gating modes behave, checks accumulate, `/alive` is unaffected everywhere, and every `503` body is exactly `{"error":{"message":"DOWN"}}`.

The measurement I most wanted is the one that justifies the whole reshape. Three dependencies that each take two seconds and answer `500`:

| | verdict | latency | dependency probes |
|---|---|---|---|
| `ReplaceFrameworkChecks` | `200 UP` | **2 ms** | **0** |
| default mode, same three deps | `503` | **6.005 s** | 3 |

`readinessProbe.timeoutSeconds: 2` in the Kubernetes guide, so the old design's probe was timing out on dependencies the application had explicitly disclaimed. That is the concrete thing the boolean mode bought, and it is bigger than I expected when I asked for it.

**Log levels, measured live rather than read:** four `WARN` handler lines carrying the reason, four `ERROR` request-log lines at `503` — exactly what the new docs section says, including that the two are separate and only the second is silenced by `LOG_DISABLE_PROBES`.

**Startup lines, live:** both modes emitted verbatim, with the count.

**Docs examples compile.** I extracted all four from the new section and built them verbatim against this branch, including `ctx.SQL.HealthCheck()` and `app.Config.Get` / `app.Logger()`. Clean.

**Lint** is clean on `health.go` and `handler_test.go`; the two `gofr.go` hits are pre-existing at lines 256 and 369. The package-level `pkg/gofr` failure is environmental — port 2121 held by an unrelated local process — and reproduces identically on `development`.

## Please fix: `logReadiness` reads `readinessChecks` without the mutex

`health.go:201`, `:206` and `:211` read the slice; `health.go:89` writes it under `a.mu`. The field's own comment in `gofr.go` says "Guarded by mu", and this is the one reader that is not.

It is your own argument that makes it a defect rather than a nit — `TestApp_AddReadinessCheck_concurrent` exists because registration is documented as pre-Run but "nothing enforces it". Nothing enforces it for `logReadiness` either, and it is reachable: `httpServerSetup` runs synchronously on whatever goroutine called `Run`, so `go app.Run()` followed by an `AddReadinessCheck` races.

Demonstrated with a test in exactly the shape of yours, only with `logReadiness` as the concurrent reader:

```
WARNING: DATA RACE
Read at 0x00c00074a1c8 by goroutine 33:
      pkg/gofr/health.go:201
Previous write at 0x00c00074a1c8 by goroutine 30:
      pkg/gofr/health.go:89
```

Four races, from the three reads. The fix is the snapshot you already use in `healthHandler`:

```go
func (a *App) logReadiness() {
	a.mu.Lock()
	checks := a.readinessChecks
	a.mu.Unlock()

	if a.container.Logger == nil || len(checks) == 0 {
		return
	}

	mode := "enabled, evaluated before the app checks"
	if frameworkReplaced(checks) {
		mode = "replaced by the app checks"
	}

	a.container.Logger.Infof("readiness: %d app check(s) registered - framework checks: %s",
		len(checks), mode)
}
```

Applied, the race is gone, your suite stays green under `-race -count=3`, and lint is clean. Say the word and I will push it rather than spend a round on three lines.

## Two nits

**Evaluation order is documented but not pinned.** Moving the app checks ahead of the framework checks survives the whole suite — the only observable difference is which reason reaches the log when both sides fail, and nothing asserts it. One row in `TestApp_AddReadinessCheck` with a failing check *and* an unhealthy dependency, asserting the log names the framework, would close it.

**Calling `FrameworkReadiness` from a default-mode check silently doubles the probe.** The doc says a default check never needs it, which is true, but not what it costs to do anyway. Measured on one 700 ms dependency:

```
default check, does NOT call FrameworkReadiness    700ms   dep probes=1
default check that DOES call it                    1.4s    dep probes=2
```

Worth half a sentence where the doc already says "call it at most once per probe" — that it is already called for you in default mode, so calling it there is the second sweep.

**A small correction, nothing to do:** `health.go` is not quite at 100%. `aggregateStatus`'s two fail-closed `return statusDown` branches are uncovered — but they are pre-existing on `development` and untouched by this PR, so they are not yours and I would not add tests for them here.

- **`ReplaceFrameworkChecks` being a property of readiness rather than of one registration.** I went looking for the failure mode — a module that registers a default check having framework gating turned off underneath it by someone else's registration — and you have already answered it the only way that scales: the startup line states the mode, and `TestApp_AddReadinessCheck` pins the any-registration semantic explicitly. Per-check modes really would be incoherent for a conjunction.
- **The hardcoded `statusUp` in the ready body.** Under replace mode a `200 UP` alongside a `DEGRADED` dependency is the point, not a leak, and the mutation to `aggregateStatus(c)` is caught.
- **Reusing `a.mu`.** It guards only GraphQL manager construction, so a probe-rate lock adds no contention anywhere real.
- **Dropping the re-registration warning.** Accumulating checks makes the case it warned about legal, which is the right way to close a finding.
- **Keeping the second contentless `WARN`.** Agreed that is `logError` re-reporting the redacted message for every handler error, not something this endpoint introduces. Separate issue if anywhere.
- **The `/.well-known/alive` section.** You wrote the rationale better than the finding did.

Two rounds, both of which changed the shape rather than the wording, and everything measured came back the way you said it would. Thanks for the patience with the second one.

---

## Adversarial pass: eight hostile checks against live servers

I went back and tried to break it rather than only to confirm it — eight applications on real ports, each registering a check a real service could plausibly write by accident. **Nothing broke.** Three things are worth writing down.

### A check that ignores `ctx` pins a goroutine per abandoned probe

One application, one check that blocks, fifty probes each abandoned by the client after 150 ms:

| the check | goroutines after 50 abandoned probes |
|---|---|
| `<-block` (ignores the context) | **+50** — one per probe, released only when the check finally returns |
| `select { case <-ctx.Done(): ... case <-block: }` | **+0** |

So the plumbing is right: cancellation does reach the check, and the fix is entirely on the check author's side. But nothing tells them. The docs say the check "receives the request context, which carries the container" — the container is the part it explains, and cancellation is the part that matters here, because your own Kubernetes guide ships `readinessProbe.timeoutSeconds: 2`. A check slower than that is abandoned on **every** probe, which at a 10-second period is a goroutine every ten seconds for the life of the pod — the same shape of leak #3496 is fixing for the framework's own sweep.

One sentence next to the `ReadinessCheck` doc comment: the context is the request's and is cancelled when the probe is abandoned, so a check that waits on anything should select on `ctx.Done()`.

### `AddReadinessCheck(nil, ReplaceFrameworkChecks())` silently drops the mode as well as the check

The `check == nil` guard returns before the options are applied, so the registration disappears entirely. Live, with a degraded dependency:

```go
app.AddReadinessCheck(nil, gofr.ReplaceFrameworkChecks())
app.AddReadinessCheck(func(*gofr.Context) error { return nil })
```

answers `503`, because framework gating is still on. Silent, and contrived — but the fix is free: apply the options first and drop only the check.

### A panicking check answers `500`, not `503`

Safe, and I checked the parts that matter: the process survives, `/ping` still answers `200`, and the panic value does not reach the body (`{"error":{"message":"Internal Server Error"}}`). Any non-2xx fails a readiness probe, so Kubernetes behaves correctly. Just unpinned by a test, and `500` from a readiness endpoint reads like a framework bug to whoever sees it first.

### What held up

Reassurance rather than padding — these are the ones I expected to find something in:

- **An application check cannot steer the endpoint.** A check returning an error that implements `StatusCode() int` returning `200` *and* `LogLevel()` returning `DEBUG`, with `db=prod-master.internal:5432 user=root` in its message: still `503`, still `{"error":{"message":"DOWN"}}`, credential nowhere. `healthHandler` returning its own `errNotReady{}` rather than the check's error is what makes that hold, and it is worth keeping deliberately.
- **A check that probes its own health endpoint does not deadlock** — the registration lock is released before the checks run. Depth 4, 3 ms.
- **A check that calls `AddReadinessCheck` while running** does not deadlock either.
- **A check dereferencing an unconfigured `ctx.SQL`** behaves exactly like the panic case: contained, `500`, process healthy.
- **200 concurrent probes**: 0 wrong out of 200, 8 ms total.

## On the test cases and the conventions

**Tests.** Sixteen scenarios across the eight new functions, table-driven with `TEST[%d], Failed.` — the form already used twelve times in `handler_test.go` — and every behaviour I could think to mutate is pinned except evaluation order. Five gaps, all small and none blocking: order, the nil-plus-option case, the panic path, `logReadiness` under concurrency, and context cancellation reaching the check.

**Conventions.** These all match:

| | precedent |
|---|---|
| `ReadinessOption func(*readinessCheck)` | `LLMOption func(*llmOptions)` (`external_db.go:88`), `MCPOption func(*mcpConfig)` (`mcp.go:17`) |
| `AddReadinessCheck(check, opts ...ReadinessOption)` | `AddLLM(m ai.Model, opts ...LLMOption)` (`external_db.go:102`) |
| `Add*` rather than `Set*` | the whole `Add*` family, and it is now genuinely accumulate |
| an exported package function taking `*Context` | `OpenAPIHandler`, `SwaggerUIHandler` (`swagger.go:23,37`) — so `FrameworkReadiness` is not a novel shape |
| static package-level error values | `errFrameworkChecks` |

The one gap against `CONTRIBUTING.md`, which asks for "integration tests for any major feature added": there is no `examples/using-readiness-check/`, where every comparable App-level feature — cron jobs, file token auth, custom metrics — has one. Not something I would hold the PR for, but it is the natural home for the `ReplaceFrameworkChecks` example and it would be exercised by CI.

For what it is worth, the three existing example tests that assert on `/.well-known/health` still pass on this branch (`using-http-service` green; `http-server` fails identically on `development`, it wants live infra) — which is the default path proven unchanged one level above the unit tests.

None of this changes the approval. The three items above are a sentence of docs, a two-line reordering, and an optional test.

aryanmehrotra (changes requested):
Re-checked at head `0ad1e03b6`. My approval from 6 Sep was dismissed on the 7th, and the only commit since is the upstream merge — so the one item I asked for before merge is still open. Restating it here so it does not get lost with the dismissed review, since that is where it currently lives.

Nothing new is being asked for. This is the same three lines, still the only thing between this PR and an approval.

## `logReadiness` reads `readinessChecks` without the mutex

`health.go:200-212` still has all three unguarded reads — `:201`, `:206`, `:211`. The write at `health.go:89` takes `a.mu`, and the field's own comment at `gofr.go:60` says "Guarded by mu". `logReadiness` remains the one reader that does not.

Reproduced against current head rather than carried over from the last round:

```
WARNING: DATA RACE
  Read at              health.go:201   (logReadiness)
  Previous write at    health.go:89    (AddReadinessCheck, under a.mu)
```

It is reachable for the same reason your own `TestApp_AddReadinessCheck_concurrent` exists: registration is documented as pre-`Run` but nothing enforces it, and `httpServerSetup` runs synchronously on whichever goroutine called `Run`, so `go app.Run()` followed by an `AddReadinessCheck` races.

The fix is the snapshot you already use in `healthHandler`:

```go
func (a *App) logReadiness() {
	a.mu.Lock()
	checks := a.readinessChecks
	a.mu.Unlock()

	if a.container.Logger == nil || len(checks) == 0 {
		return
	}

	mode := "enabled, evaluated before the app checks"
	if frameworkReplaced(checks) {
		mode = "replaced by the app checks"
	}

	a.container.Logger.Infof("readiness: %d app check(s) registered - framework checks: %s",
		len(checks), mode)
}
```

I applied that to the current head and re-verified: the race is gone, the readiness suite stays green under `-race -count=3`, and `golangci-lint --new-from-rev=origin/development` reports 0 issues.

## Also still open from the dismissed review

These were nits then and remain nits — none of them needs a round of its own, but since we are here anyway:

1. **Evaluation order is documented but not pinned.** Moving the app checks ahead of the framework checks survives the whole suite. One row in `TestApp_AddReadinessCheck` with a failing check *and* an unhealthy dependency, asserting the log names the framework, closes it.
2. **A check that ignores `ctx` pins a goroutine per abandoned probe** — measured at +50 goroutines over 50 abandoned probes, versus +0 when the check selects on `ctx.Done()`. The plumbing is right; nothing tells the check author. One sentence on the `ReadinessCheck` doc comment saying the context is the request's and is cancelled when the probe is abandoned. This matters because the Kubernetes guide ships `readinessProbe.timeoutSeconds: 2`.
3. **Calling `FrameworkReadiness` from a default-mode check silently doubles the probe** — measured 700ms/1 probe versus 1.4s/2 probes. Half a sentence where the doc already says "call it at most once per probe".
4. **`AddReadinessCheck(nil, ReplaceFrameworkChecks())` drops the mode along with the check.** The nil guard returns before the options are read, so the caller gets default gating while believing they replaced it.

Only item 1's test and the doc sentences in 2 and 3 are worth doing here; 4 is a judgement call and I am fine either way.

## To be clear about the state

The design work is done and I am not reopening any of it — the reshape landed, both gating modes behave, and everything I measured last round came back as described. The branch is also `BEHIND` and will want a rebase before merge.

Happy to push the `logReadiness` fix myself if that is quicker for you — say the word. Otherwise apply it and I will re-approve.

pkg/gofr/health.go
```diff
@@ -1,19 +1,214 @@
...
+	}
+
+	return nil
+}
+
+// logReadiness states at startup that readiness is no longer the default, and which of the two
+// policies is in force, so a 503 from this endpoint — or a 200 with a dependency down — is never a
+// surprise about where the verdict came from.
+func (a *App) logReadiness() {
+	if a.container.Logger == nil || len(a.readinessChecks) == 0 {
```
Umang01-hash: Data race: this and the reads at :206/:211 touch `a.readinessChecks` unlocked, while `AddReadinessCheck` (:89) writes it under `a.mu` and the field is commented "Guarded by mu". `healthHandler` snapshots under the lock — `logReadiness` is the one reader that doesn't. Reachable via `go app.Run()` + a later `AddReadinessCheck` (the unenforced pre-Run window `TestApp_AddReadinessCheck_concurrent` covers). Same snapshot fix: `a.mu.Lock(); checks := a.readinessChecks; a.mu.Unlock()` then read `checks`.

Umang01-hash (changes requested):
Agree with @aryanmehrotra's still-open blocker — the `logReadiness` data race is real. Reproduced at head myself: `go app.Run()` racing an `AddReadinessCheck` gives `WARNING: DATA RACE — read health.go:201 (logReadiness) / previous write health.go:89 (AddReadinessCheck, under a.mu)`. The suite's `-race` stays green only because no test drives logReadiness against the writer. Fix is the snapshot `healthHandler` already uses. Nits from Aryan's list (eval-order test, the ctx-cancellation sentence on the ReadinessCheck doc, the nil+opts drop) are worth folding in the same push. Rest of the PR verified end-to-end and is solid.

akshat-kumar-singhal:
@aryanmehrotra @Umang01-hash — pushed as `6efda4d1` and `1af8007a`. The blocker and the three nits are done; I've taken item 4 in a different direction and want to make the case for it rather than quietly diverge.

## The blocker: `logReadiness` under the mutex

Applied exactly as written — the snapshot `healthHandler` already uses.

I also made the existing concurrency test actually cover it. `TestApp_AddReadinessCheck_concurrent` previously drove only `healthHandler` against the writer, which is why the suite stayed green with `logReadiness` unguarded — the reader nothing exercised was the reader that raced. It now runs `logReadiness` as a third concurrent goroutine, so the guard is pinned rather than merely present.

Mutation-checked rather than assumed. Removing only the mutex from `logReadiness`:

```
WARNING: DATA RACE
  Read at    logReadiness       (the two unguarded reads)
  Write at   AddReadinessCheck  (under a.mu)
--- FAIL: TestApp_AddReadinessCheck_concurrent
```

Restored, green under `-race -count=2`.

## Nits 1–3

1. **Evaluation order pinned.** New `TestApp_healthHandler_evaluationOrder`: an unhealthy dependency plus a failing default-mode check must log the framework's reason and not the check's. Swapping the two turns it red.
2. **Context cancellation** documented on `ReadinessCheck` and in the guide — the context is the probe request's, is canceled when the probe is abandoned, and a check that waits on anything must select on `ctx.Done()` or it leaks a goroutine per probe. Tied to the `timeoutSeconds: 2` the Kubernetes guide ships, as you framed it.
3. **The double sweep** documented on `FrameworkReadiness` and in the guide.

## Item 4 — I'd like to keep the drop, and make it loud instead

This is the one I'd push back on. You called it a judgement call and said either way was fine, so here is the reasoning.

Passing a nil check is a programming error, and the question is what a *bug* should be allowed to do. Retaining the options means half an erroneous call takes effect. And because `ReplaceFrameworkChecks` is deliberately a property of readiness *as a whole* — which is the design we settled on, and which I still think is right — that half acts at a distance:

```go
// module A, buggy: someCheck() returned nil
app.AddReadinessCheck(someCheck(), gofr.ReplaceFrameworkChecks())

// module B, correct, and now silently ungated on every datasource
app.AddReadinessCheck(licenseCheck)
```

That is the failure mode you went looking for in the second round — "a module that registers a default check having framework gating turned off underneath it by someone else's registration" — and concluded was answered. Retaining the options reintroduces it, triggered by a bug rather than by a deliberate call. It seems a wider silent failure than the one it fixes: the original complaint is a caller who doesn't get the mode they asked for, which is contained to that caller.

So the registration is still dropped whole, but no longer silently:

```go
func (a *App) AddReadinessCheck(check ReadinessCheck, opts ...ReadinessOption) {
	if check == nil {
		if a.container != nil && a.container.Logger != nil {
			a.container.Errorf("invalid readiness check: check must not be nil - registration ignored")
		}

		return
	}
	...
```

This matches how the other registration methods report invalid input — `gofr.go:345`, `"invalid subscription: topic and handler must not be empty or nil"`, on a nil handler — and it is the warn-on-misregistration precedent you cited for `AddHTTPService` in the first round, applied to the case that is unambiguously an error. A caller who writes `ReplaceFrameworkChecks()` next to a nil check now finds out at startup instead of inferring it from a `503`.

Two rows in `TestApp_AddReadinessCheck` pin the semantic (the option is dropped with the check; a nil-only registration leaves the default path untouched), plus `TestApp_AddReadinessCheck_nilIsLogged`.

If you'd still rather have the options retained, say so and I'll switch it — it's four lines either way.

## Branch state

Rebased onto your `development` merge (`45a32791`) rather than merging again.

One thing to flag: your merge took `development` at `7a5bdfc9`, and I had separately merged it at `2d78c40d` — one commit newer. Rebasing rewrote that upstream websocket fix (#3932) into a new SHA, which would have surfaced in this PR as a commit of mine. I dropped it and cherry-picked only my two, so the branch is deliberately one commit behind `development` and carries nothing that isn't this feature. Worth a re-merge before merge if that matters.

`go build ./...` clean, readiness suite green under `-race`, `golangci-lint --new-from-rev=development` reports 0 issues.

## Still open, not addressed here

No `examples/using-readiness-check/`, against `CONTRIBUTING.md`'s "integration tests for any major feature added". You flagged it as non-blocking in the second round and I've left it — happy to open a follow-up issue rather than grow this PR further.

aryanmehrotra (approved):
Re-verified at current head \`307f66e40\`.

All previously discussed items are resolved and verified end-to-end:

1. **`logReadiness` data race fixed**: Snapshotting `readinessChecks` under `a.mu` eliminates the race. Verified under `-race` with `TestApp_AddReadinessCheck_concurrent` driving registrations, health probes, and `logReadiness` concurrently.
2. **Evaluation order pinned**: `TestApp_healthHandler_evaluationOrder` confirms framework dependency checks evaluate before application checks when both fail.
3. **Context cancellation & probe lifecycle documented**: The doc comment on `ReadinessCheck` clearly instructs authors to respect `ctx.Done()` to avoid goroutine leaks on abandoned probes.
4. **`FrameworkReadiness` duplicate sweep documented**: Docs explicitly note calling it in default mode runs an unnecessary second sweep.
5. **Nil checks handled cleanly**: Nil check registrations log a clear error and drop the whole registration rather than half-applying options. Verified with `TestApp_AddReadinessCheck_nilIsLogged`.

Ran local validation:
- All readiness tests pass cleanly under `-race`
- 100% statement coverage across all new methods in `pkg/gofr/health.go`
- `golangci-lint run --new-from-rev=origin/development` reports 0 issues

LGTM! Approved.

aryanmehrotra (approved):

Umang01-hash (approved):
LGTM. Verified at head 307f66e — the earlier readinessChecks race is properly closed now: all three sites (AddReadinessCheck, healthHandler, logReadiness) take a.mu, and the concurrent test fails under -race if you drop any of them (confirmed by reverting the lock). All additive, no breaking changes.

Ran it locally end-to-end: a failing check returns 503 with {"error":{"message":"DOWN"}} (no dependency leak), the not-ready reason logs at WARN, and startup announces the mode. Matches the docs. Tests are solid and fail-on-revert; build/vet/lint clean.

One thought, non-blocking: ReplaceFrameworkChecks flips framework gating off for the whole endpoint, so in a multi-module app one module's replace-check silently drops framework gating from another module's plain check. You document and log it, so it's a deliberate call — just flagging it as a possible surprise.

PR #3874, opened 2026-08-10
chore: remove committed example binaries and ignore build output

aryanmehrotra:
**Description:**

`go build` was run inside three example directories in #3854 and the resulting executables were
committed. They are **183 MB** in the working tree and **86.7 MiB of the repository's 109 MiB
pack** — roughly **79% of a clone**.

| file | working tree | packed |
|---|---|---|
| `examples/using-s3-filestore/using-s3-filestore` | 66 MB | 30.8 MiB |
| `examples/using-cloudsql/using-cloudsql` | 62 MB | 29.3 MiB |
| `examples/using-add-filestore/using-add-filestore` | 55 MB | 26.6 MiB |

They are `Mach-O 64-bit executable arm64`, so they are also useless to anyone not on Apple silicon.
`main` currently carries no blob over 3 MB, and these have not reached it yet.

**Why `.gitignore` did not catch them.** The existing rules cover `*.exe`, `*.so`, `*.dylib` and
`*.test`. `go build` in a module directory emits an **extensionless** executable named after that
directory, which none of those match. Added rules for each example's build output, with negations so
`.go`, `.md`, `.mod`, `.sum` and config files in those directories stay tracked.

Verified the rules are safe and effective:

- `git ls-files | git check-ignore --stdin` → **no tracked file is shadowed**
- built a real binary in `examples/using-s3-filestore` → caught by `.gitignore:29`
- `examples/http-server/http-server`, `examples/sample-cmd/sample-cmd` and
  `examples/grpc/grpc-unary-server/grpc-unary-server` are also covered, so this cannot recur in the
  example directories that do not start with `using-`

**Scope — what this does and does not fix.** Deleting them keeps them out of every future checkout
and out of the release source tarballs. It does **not** shrink existing clones: the blobs stay
reachable from `f8ead999`, so a fresh `git clone` still downloads 86.7 MiB of them. That needs
`git filter-repo` plus a force-push to `development`, which is a separate, more disruptive decision
and is not attempted here.

Anyone who wants a small clone today can use a partial clone, which is 22 MB once these are gone
from the tip:

```
git clone --filter=blob:limit=2m https://github.com/gofr-dev/gofr.git
```

None. No source, test, or config file is modified — three untracked-from-now-on binaries are
deleted and `.gitignore` gains 15 lines.

**Additional Information:**

The same three deletions are cherry-picked into `release/v1.59.0` so `main` never receives them at
its tip. Both branches carry the identical tree change, so `main` and `development` do not diverge.

**Checklist:**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,18 @@ vendor/
 # IDE Cache
 .vscode
 .DS_Store
+
+# Compiled example binaries. `go build` inside an example directory emits an
+# extensionless executable named after that directory, which none of the rules
+# above match (they cover .exe/.so/.dylib/.test).
+/examples/*/using-*
+/examples/*/http-server*
+/examples/*/sample-cmd
+/examples/grpc/*/grpc-*
+!/examples/**/*.go
+!/examples/**/*.md
+!/examples/**/*.mod
+!/examples/**/*.sum
+!/examples/**/*.yaml
+!/examples/**/*.yml
+!/examples/**/*.json


aryanmehrotra:
Closing this in favour of cleaning `development` properly.

Deleting the files in a new commit only fixes the tip. The blobs stay reachable from `f8ead999`, so
a fresh `git clone` still downloads all **86.7 MiB** of them — that is the part that actually costs
contributors, and it is roughly 79% of the repo's 109 MiB pack. A normal PR cannot reach it; it needs
`git filter-repo` and a force-push to `development`.

The `.gitignore` rules here are still worth keeping, so they should ride along with whatever branch
carries the rewrite:

```
/examples/*/using-*
/examples/*/http-server*
/examples/*/sample-cmd
/examples/grpc/*/grpc-*
!/examples/**/*.go
!/examples/**/*.md
!/examples/**/*.mod
!/examples/**/*.sum
!/examples/**/*.yaml
!/examples/**/*.yml
!/examples/**/*.json
```

Verified before closing: no tracked file is shadowed (`git ls-files | git check-ignore --stdin` is
empty), a real binary built in `examples/using-s3-filestore` is caught, and the build output of
`http-server`, `sample-cmd` and `grpc/grpc-unary-server` is covered too, so this cannot recur in the
example directories that do not start with `using-`.

One note on sequencing for whoever runs the rewrite: it rewrites every commit after `f8ead999`,
including `c0b69322`, which is what `release/v1.59.0` was cut from. So the rewrite has to land before
the release branch is cut, not after, or the release branch has to be recreated.

PR #3875, opened 2026-08-10
Release/v1.59.0

aryanmehrotra:
# Release v1.59.0

**TL;DR** — `/.well-known/health` no longer leaks your infra to unauthenticated callers. Metrics can
now be **pushed** (OTLP + keyless GCP), not just scraped. Static file endpoints finally serve their
own root. S3 streaming no longer hangs forever.

## 🚀 Features

🔹 **Push-based metrics export**

Scraping misses serverless (Cloud Run scale-to-zero, short jobs). Now you can push.

| | Control | Default |
|---|---|---|
| Prometheus pull `/metrics` | `METRICS_PORT` (`0` disables) | on, `2121` |
| Push export | `METRICS_EXPORTER` = `otlp` \| `gcp` | **off** |

Both readers share the same instruments — **no double-counting**. Set nothing → identical to v1.58.0.

```bash
METRICS_EXPORTER=otlp
METRICS_URL=localhost:4317
# push-only for serverless: add METRICS_PORT=0
```

New config, all optional:

| var | notes |
|---|---|
| `METRICS_EXPORTER` | `otlp`, `gcp`. Unset/`prometheus` = pull-only |
| `METRICS_URL` | `host:port` for gRPC, full URL for HTTP |
| `METRICS_PROTOCOL` | `grpc` \| `http` |
| `METRICS_EXPORT_INTERVAL` | seconds between pushes |
| `METRICS_TEMPORALITY` | `delta` for Datadog · `cumulative` for Prometheus/GMP/Grafana |
| `METRICS_HEADERS` | `k=v,k=v` — wins over `METRICS_AUTH_KEY` |
| `METRICS_AUTH_KEY` | Authorization header value |
| `METRICS_INSECURE` | disables TLS. Ignored by `gcp` |

🔹 **Keyless GCP Managed Prometheus**

Pushes to Google's Telemetry (OTLP) API via Application Default Credentials — **no key file**. On
Cloud Run it uses the attached service account (needs `roles/monitoring.metricWriter`).

```go
import _ "gofr.dev/pkg/gofr/metrics/exporters/gcp" // then METRICS_EXPORTER=gcp
```

Separate module, so Google deps stay out of the framework for everyone else. Defaults to
`telemetry.googleapis.com:443`, always TLS. Runnable `examples/using-gcp-metrics` included.

## 🛠️ Fixes

🔹 **`/.well-known/health` leaked infrastructure detail to unauthenticated callers** — the endpoint is
auth-exempt by design so probes can reach it, but it returned the full per-dependency map. Any caller
with no credentials got dependency names, hosts and ports, database / keyspace / bucket names,
connection-pool stats, usernames, raw connection error strings, and the framework version.

| | before | now |
|---|---|---|
| body | full per-dependency map + `version` | `{"name","status"}` |
| status code | `200` | `200` — unchanged, probes keep working |

⚠️ **Response shape changed.** Anything parsing this body for per-dependency detail must be updated.
The `version` field is gone too — it is exactly what someone enumerating known CVEs wants from an
unauthenticated endpoint. The aggregate `UP` / `DEGRADED` status is unchanged, so readiness probes are
unaffected. `Container.Health` still computes the detailed map in-process; serving it on
`METRICS_PORT` is tracked separately.

🔹 **Static endpoints couldn't serve their own root** — `GET /static` returned `403`, `GET /static/`
could redirect forever. Both now route, and resolve to `index.html`.

| request | before | now |
|---|---|---|
| `GET /static` | `403` | serves `index.html` |
| `GET /static/` | redirect loop | serves `index.html` |
| `POST/DELETE /static` | `200` + file body | `405` + `Allow: GET, HEAD` |
| unreadable file | `500` | `403` |

🔹 **S3: `io.Copy` never terminated and could OOM the process** — `Read` returned `len(p)` with a
nil error no matter how much it actually read, so **EOF never arrived** and anything bigger than one
HTTP chunk was silently zero-padded. `Read`/`ReadAt` now use `io.ReadFull`, return real counts,
signal `io.EOF`, and issue HTTP **Range** requests instead of re-downloading.

Two more from the same report:

- **`Create` rejected the first object under any new prefix** — S3 has no directories, so the
  parent-prefix check failed every fresh path like `uploads/<uuid>/report.pdf`.
  ⚠️ **Intentional behaviour change:** you no longer get `ErrOperationNotPermitted` for a missing
  parent. Documented in the file-handling guide.
- **Nil-pointer panic on optional metadata** — `ContentType` / `LastModified` / `ContentLength` are
  optional headers; backends that omit them (Cloudflare R2) crashed. Now nil-checked.

🔹 **Data race on shutdown** — HTTP, metrics and MCP servers wrote `*http.Server` on the serve
goroutine while `Shutdown()` read it. Each is now mutex-guarded across the assignment and read
only — never across `ListenAndServe` / `Shutdown`.

🔹 **EventHub `Publish` with an uninitialized producer** — returns `errClientNotConnected` instead
of failing on a nil client.

🔹 **Dependency updates** — consolidated minor/patch bumps.

## ⚡ Performance

🔹 **S3 `ReadDir` lists one level** — uses `Delimiter` + `CommonPrefixes` instead of loading
metadata for every nested object. Big win on large buckets.

🔹 **Request hot path: −577 B and −6 allocs per request** (one access log + one response + one
`X-Forwarded-For` parse). No API, wire-shape or env-var change.

| benchmark | time | B/op | allocs/op |
|---|---|---|---|
| `ResponderRespond` | 380 → 334 ns (−12%) | 249 → 72 (−71%) | 4 → 2 |
| `getIPAddress` | 59 → 26 ns (−56%) | 48 → 0 (−100%) | 1 → 0 |
| `ContextLogger.Info` | 584 → 461 ns (−21%) | 568 → 216 (−62%) | 9 → 6 |
| `Logger.Info` (JSON) | 513 → 393 ns (−23%) | — | — |

## 🔧 Enhancements

🔹 **CI now actually fails when tests fail** — the example, pkg and submodule jobs ran `go test`
in blocks without `errexit`, so non-zero exits were swallowed and the jobs reported **green over
red tests**. All three gate correctly now, the example failures this uncovered are fixed, and the
submodule job uploads a coverage artifact for the first time (27/27 modules).

## 🧹 Housekeeping

🔹 **Removed 183 MB of accidentally committed example binaries** — three compiled executables were
committed alongside a dependency update and never reached `main`. `.gitignore` now covers the
extensionless output `go build` produces inside an example directory, which the existing
`*.exe` / `*.so` / `*.dylib` / `*.test` rules did not match.

**Full Changelog:** https://github.com/gofr-dev/gofr/compare/v1.58.0...v1.59.0


aryanmehrotra:
Closing this until `development` has been cleaned properly, because the rewrite that cleaning needs
will invalidate this branch.

`release/v1.59.0` was cut from `c0b69322`. Removing the committed binaries from `f8ead999` rewrites
that commit and therefore every commit after it — a commit's hash covers its tree *and* its parent's
hash, so the change cascades to the tip. Ran it on a throwaway clone of this repo to confirm rather
than assume:

| commit | before | after |
|---|---|---|
| `chore(deps)` (#3854) — the commit edited | `f8ead999` | `9cd5d7c1` |
| `fix(http)` static root (#3820) | `de276c92` | `9816137a` |
| `fix(ci)` submodule job (#3825) | `faa42e40` | `be92a2c6` |
| `fix(ci)` example/pkg jobs (#3814) | `80a61e4e` | `1f207742` |
| `fix(health)` redaction (#3807) | `c0b69322` | `f520673f` |

`1d238069`, the parent of the rewritten commit, is untouched — only what follows changes. So the base
of this branch stops existing and it has to be recut regardless.

The order matters in one direction only:

```
rewrite development  ->  recut release  ->  tag v1.59.0     works
tag v1.59.0  ->  rewrite afterwards                         too late for main
```

Once `v1.59.0` is tagged, `sum.golang.org` pins that commit, and since the release merges
`development`'s commits into `main`, `main`'s history below the tag can never be rewritten. The
binaries would be reachable from `main` permanently. That is why this waits rather than shipping
first — `f8ead999` is currently 5 commits from the tip, and that only grows.

Nothing here is lost. The release notes above are current and include #3807, the version bump is one
line, and recutting is:

```
git checkout -b release/v1.59.0 origin/development
# set pkg/gofr/version/version.go -> const Framework = "v1.59.0"
git merge origin/main            # keep v1.59.0 on the version.go conflict
```

Two things worth carrying into the rewrite when it happens: the `.gitignore` rules from #3874, which
cover the extensionless output `go build` leaves in an example directory (the existing
`*.exe` / `*.so` / `*.dylib` / `*.test` rules do not match it), and the fact that the rewrite forces a
rebase on every open PR based on `development`.

Issue #3877, opened 2026-08-10
Fix inconsistent logging unit for SQL query duration reporting microseconds instead of milliseconds


pritesh-4:
### Description
There is a duration unit inconsistency in the SQL datasource query logging. The console logger formats query execution latency with a `µs` (microsecond) suffix, but the underlying duration value is calculated in milliseconds. 
Other datasources within GoFr (such as Redis, MongoDB, Cassandra, ClickHouse, and SurrealDB) correctly calculate query latency in microseconds and format them with `µs`.
---
### Steps to Reproduce
1. Configure any GoFr application with a SQL database (e.g. SQLite, PostgreSQL, or MySQL).
2. Execute a fast database query (e.g., `SELECT 1` or fetching a small row by ID).
3. Check the output logs in the terminal.

#### Observed Logs:
*(Even when the query takes hundreds of microseconds to execute).*
If a query takes exactly 15 milliseconds, it is logged as:

---

### Expected Behavior
The query latency logs should report accurate duration in microseconds, matching the rest of GoFr's datasources:
* A query taking 15 milliseconds should be logged as `15000µs` (or `15.0ms` if using milliseconds).
* A sub-millisecond query taking 500 microseconds should be logged as `500µs`.

---

### Actual Behavior
The SQL datasource logs milliseconds under the microsecond (`µs`) unit label, causing a 1000x under-reporting and truncating sub-millisecond queries to `0µs`.

---

### Root Cause
In `pkg/gofr/datasource/sql/db.go`, the helper function `sendStats` calculates query execution duration in milliseconds:
```go
func sendStats(logger datasource.Logger, metrics Metrics, config *DBConfig, start time.Time, queryType, query string, args ...any) {
	duration := time.Since(start).Milliseconds()
    // ...

adarshjnv9100:
Hi @pritesh-4,

I’d like to work on this issue and would appreciate it if you could assign it to me.

I’ve gone through the issue and understand the root cause and the expected fix. My plan is:

1. **Investigate the existing SQL logging flow**

   * Review `pkg/gofr/datasource/sql/db.go`, especially the `sendStats` helper.
   * Trace how query execution time is calculated and passed to the logger/metrics.
   * Compare the implementation with other datasources such as Redis, MongoDB, Cassandra, ClickHouse, and SurrealDB to ensure consistency.

2. **Fix the duration unit mismatch**

   * The current implementation uses:
     `time.Since(start).Milliseconds()`
[...]

Issue #2188, opened 2025-08-19
Metrics show DB calls at infinite for every query


coolwednesday:
<img width="1392" height="324" alt="Image" src="https://github.com/user-attachments/assets/cec09fbe-6c42-439d-9774-ddfb66c51f29" />

Expected :

<img width="1134" height="738" alt="Image" src="https://github.com/user-attachments/assets/117116e7-b8f0-423e-ae51-de2abd0d0c52" />

Umang01-hash:
This is a bug i tested locally. Even if the db calls returns in under 2-5 milliseconds the metrics collected show infinity.

PR #3878, opened 2026-08-10
fix(sql): correct query duration logging unit to microseconds

pritesh-4:
Issue: #3877

## Description

This PR fixes an inconsistency in SQL query duration logging while keeping the existing `app_sql_stats` metric contract unchanged.

### Problem

SQL query duration in `sendStats` is consumed by two different paths:

- **SQL logs** should report query duration in **microseconds (`µs`)**
- **`app_sql_stats`** is documented and registered as reporting query duration in **milliseconds (`ms`)**

Previously, `sendStats` calculated a single duration value using:

go
duration := time.Since(start).Milliseconds()- **Expected Benefits:** Developers will see accurate query runtimes in console logs matching the format of all other GoFr database adapters (Redis, MongoDB, etc.).

- None. Telemetry endpoints and metrics are backward-compatible.

**Additional Information:**

- No new dependencies or external libraries were added.
- The change was verified through existing unit test adaptations.

**Checklist:**

- [✔️] I have formatted my code using `goimport` and `golangci-lint`.
- [✔️] All new code is covered by unit tests.
- [✔️] This PR does not decrease the overall code coverage.

### DIFF
--- a/pkg/gofr/datasource/sql/db.go
+++ b/pkg/gofr/datasource/sql/db.go
@@ -54,20 +54,21 @@ func clean(query string) string {
 }
 
 func sendStats(logger datasource.Logger, metrics Metrics, config *DBConfig, start time.Time, queryType, query string, args ...any) {
-	duration := time.Since(start).Milliseconds()
+	duration := time.Since(start)
 
 	if logger != nil {
 		logger.Debug(&Log{
 			Type:     queryType,
 			Query:    query,
-			Duration: duration,
+			Duration: duration.Microseconds(),
 			Args:     args,
 		})
 	}
 
 	// This contains the fix for the nil pointer dereference
 	if metrics != nil {
-		metrics.RecordHistogram(context.Background(), "app_sql_stats", float64(duration), "hostname", config.HostName,
+		metrics.RecordHistogram(context.Background(), "app_sql_stats",
+			float64(duration)/float64(time.Millisecond), "hostname", config.HostName,
 			"database", config.Database, "type", getOperationType(query))
 	}
 }
--- a/pkg/gofr/datasource/sql/db_test.go
+++ b/pkg/gofr/datasource/sql/db_test.go
@@ -772,14 +772,12 @@ func TestDB_sendOperationStats_RecordsMilliseconds(t *testing.T) {
 	start := time.Now().Add(-1500 * time.Millisecond) // 1.5 seconds ago
 
 	mockMetrics.EXPECT().RecordHistogram(
-		gomock.Any(), "app_sql_stats", float64(1500),
+		gomock.Any(), "app_sql_stats",
+		gomock.Cond(func(v float64) bool { return v >= 1500 && v < 1510 }),
 		"hostname", "host", "database", "db", "type", "SELECT",
 	)
 
 	db.sendOperationStats(start, "SELECT", "SELECT * FROM users")
-
-	duration := time.Since(start).Milliseconds()
-	assert.Equal(t, int64(1500), duration)
 }
 
 // --- Tx method tests ---


akshat-kumar-singhal:
@pritesh-4 please rebase your branch

pritesh-4:
@akshat-kumar-singhal the branch is rebased now

aryanmehrotra (changes requested):
The production change is right, and I would merge it as it stands. One required change in the test, which currently removes a regression guard the repo added deliberately.

## Why this is the correct fix, with the argument the description is missing

`sendStats` has one `duration` variable and two consumers that want different units:

```
duration := time.Since(start).Milliseconds()      db.go:57
   |
   +--> logger.Debug(&Log{Duration: duration})    printed as "%8dµs"  (db.go:45)
   |
   +--> RecordHistogram("app_sql_stats", ...)     registered as "Response time of
                                                   SQL queries in milliseconds."
                                                   (container/container.go:410)
```

It used to be `.Microseconds()`. It was changed to `.Milliseconds()` in `e1ddc1c7` (#2191, closing #2188) because the *metric* was 1000x over-reporting against a histogram documented in milliseconds. That fixed the metric — and, because the two consumers share one variable, moved the error onto the log, which kept its `µs` label. So this PR is not a revert of #2191; it is the half of it that was never done. Worth saying in the description, because "correct the unit" reads like it is undoing a deliberate decision, and it isn't.

Two things I checked that make the case stronger than the description does:

**1. The log is not merely imprecise, it is empty.** `.Milliseconds()` returns `int64`, so every sub-millisecond query logs `0µs`. Measured against a live MySQL, warm, through the real GoFr handler path (`ctx.SQL.QueryRowContext`, 100 requests):

```
min 247µs   p50 367µs   p90 740µs   max 1239µs
```

Over 90% of those log as `0µs` today. On a co-located production database the proportion is higher, not lower. A raw `database/sql` loop against the same server gave p50 327µs with 195/200 queries under a millisecond.

Note this also applies in JSON log mode, where there is no unit suffix at all — the `duration` field just silently carries a different unit for SQL than for every other datasource, which is worse for a log aggregator than a wrong label is for a human.

**2. `/1000.0` does not just add decimals to the metric — it unlocks buckets that are currently unreachable.** `getDefaultDatasourceBuckets()` (`container/container.go:478`) opens with

```go
.05, .075, .1, .125, .15, .2, .3, .5, .75, 1, ...   // 50µs-10ms
```

Fifteen fractional-millisecond boundaries, widened to this range in #3568. An `int64` millisecond count can never land in any of them: a 327µs query records `0` and falls in the `<=0.05` bucket. So today a Grafana p50 on `app_sql_stats` reads "<=50µs" for a query that actually took 327µs — a precise-looking number that is off by 6.5x. `float64(duration)/1000.0` is what makes those buckets do anything. That is the single best argument for this PR and it is not in the description.

**3. House convention agrees with you.** `sql` and `elasticsearch` are the only two places in the repo where a `.Milliseconds()` value is printed under a `µs` label. Every other datasource logger that prints `µs` is fed `.Microseconds()`. The one datasource that reports milliseconds — influxdb — labels them `ms` (`influxdb/logger.go:42`) and is internally consistent.

## Required change: the test no longer pins the metric unit

`db_test.go:774-782`:

```go
mockMetrics.EXPECT().RecordHistogram(
    gomock.Any(), "app_sql_stats", gomock.Any(),   // was: float64(1500)
    ...
)

db.sendOperationStats(start, "SELECT", "SELECT * FROM users")

duration := time.Since(start).Microseconds()
assert.InDelta(t, float64(1500000), float64(duration), 5000)
```

The `float64(1500)` that was replaced with `gomock.Any()` is the guard added in #2191 for exactly the bug this file has had before. And the `InDelta` that replaced it recomputes `duration` from `start` in the test's own scope — it never observes anything `sendStats` did, so it asserts only that `time.Since(a 1.5s-old timestamp)` is about 1.5 seconds. That holds no matter what the production code does.

Demonstrated rather than asserted: with your test as written, I reverted line 70 back to `float64(duration)` — reintroducing #2188 in full, microseconds into a millisecond histogram — and the test still passes.

```
=== PR test vs the #2188 mutation (float64(duration), no /1000) ===
ok  	gofr.dev/pkg/gofr/datasource/sql	0.618s
```

I understand why you loosened it — the elapsed time is not exactly 1500ms, so `float64(1500)` would flake. A range matcher gets the flake-immunity without giving up the pin:

```go
	start := time.Now().Add(-1500 * time.Millisecond) // 1.5 seconds ago

	// app_sql_stats is documented as milliseconds, so 1.5s must arrive as ~1500 and not as
	// 1_500_000. The range absorbs scheduling jitter while still failing if the unit changes.
	mockMetrics.EXPECT().RecordHistogram(
		gomock.Any(), "app_sql_stats",
		gomock.Cond(func(v float64) bool { return v >= 1500 && v < 1510 }),
		"hostname", "host", "database", "db", "type", "SELECT",
	)

	db.sendOperationStats(start, "SELECT", "SELECT * FROM users")
```

The trailing `duration`/`InDelta` lines can go — the expectation is the assertion. I ran this: 20/20 passes with your fix in place, and it fails on the mutation above with `Got: 1.50005e+06 ... Want: adheres to a custom condition`. `gomock.Cond` is generic in the `go.uber.org/mock v0.6.0` this repo pins, so the `float64` type parameter is inferred.

## Note, not a blocker: `elasticsearch` has the identical bug

`elasticsearch/elasticsearch.go:350` computes `.Milliseconds()`, assigns it to `QueryLog.Duration`, and `elasticsearch/logger.go:61` prints it with a `µs` suffix. Its metric, `es_request_duration_ms`, is correctly milliseconds, so it needs exactly the same split you have made here. Nobody has filed it. Equally happy to see it in this PR or left for a separate issue — your call, and I would not hold this one for it.

- **Dividing by 1000 rather than moving `app_sql_stats` to microsecond buckets.** Redis went the other way (microsecond values against `getDefaultDatasourceMicrosecondBuckets`), so there is precedent for it — but it would change the unit of an already-published metric and silently invalidate every dashboard, recording rule and alert threshold on `app_sql_stats`. Keeping the metric in milliseconds is the compatible choice and the right one.
- **Relabelling the log `ms` instead**, which would have been a one-character change with precedent in influxdb. It fails on the numbers above: those queries would print `0ms` — honest, and just as uninformative.
- **Not touching `Tx.sendOperationStats` separately.** It routes through the same `sendStats`, so transactions are covered by the same two lines. Correct call.
- **The float conversion itself.** `float64(µs)/1000.0` is exact enough here; there is nothing to gain from `d.Seconds()*1000`.
- **The test name** still saying `RecordsMilliseconds` when it now covers both units. Accurate for the metric, which is what it asserts. Leave it.

## One housekeeping item

The branch is behind `development` again since the 30 Aug rebase. A rebase before merge, please.

<details>
<summary>How the measurements above were produced</summary>

- Live MySQL in a local sandbox, warmed, both runs against the same server.
- **Through GoFr:** a real `gofr.New()` app with `LOG_LEVEL=DEBUG` and a handler calling `ctx.SQL.QueryRowContext`, driven with 100 sequential requests; the figures are the `duration` field of the emitted `QueryRowContext` debug logs, with this PR applied.
- **Raw driver:** 200 iterations each of `SELECT 1`, a primary-key lookup and a 100-row scan through `database/sql` directly, timing the same span `sendStats` measures.
- **Test guard:** this PR's `db.go` change applied locally, then line 70 mutated back to `float64(duration)`; the PR's test passes, the `gomock.Cond` version fails. Full `./pkg/gofr/datasource/sql/...`, `./pkg/gofr/container/...` and the SQL tests in `./pkg/gofr` all pass with the change and the proposed test.

</details>

Thanks for picking this up — the diagnosis in #3877 is accurate and the shape of the fix is the right one. It is only the test that needs to come back to where #2191 left it.

pritesh-4 (commented):
@aryanmehrotra I changed the requested codes, and the branch is rebased now. Check it

aryanmehrotra (changes requested):
Thanks — the test guard is back and the branch is clean on `development` (ahead 1, behind 0). I checked out `5e712d8` and verified the guard actually guards, rather than taking it on the diff:

| mutation applied to `db.go` | test result |
|---|---|
| drop `/1000.0` — reintroduces #2188 in full | **FAIL** |
| `.Microseconds()` back to `.Milliseconds()` — reverts the PR | **FAIL** |
| `/1000.0` mistyped as `/1000000.0` | **FAIL** |
| unmodified | pass, 10x, and under `-race` |

All three killed, which is what the old version could not do. `go vet`, `gofmt` and the full `./pkg/gofr/...` and `./pkg/gofr/container/...` suites are green with your head applied.

I also sized the upper bound I suggested, since a range matcher is only as good as its headroom. Over 3000 samples, including runs under sixteen CPU burners with `GOMAXPROCS=1`, the value the matcher sees never exceeded **1500.082** — about 9.9ms of slack before `1510`. It will not flake.

## One new item, and it will fail CI

`golangci-lint` rejects the literal:

```
pkg/gofr/datasource/sql/db.go:70:84: Magic number: 1000.0, in <argument> detected (mnd)
        metrics.RecordHistogram(context.Background(), "app_sql_stats", float64(duration)/1000.0, ...
                                                                                        ^
```

That is v2.12.2, the exact version pinned in `.github/workflows/go.yml`. `mnd` is enabled at `.golangci.yml:30` with `argument` among its `checks`, and the exclusion at `.golangci.yml:156` covers `_test.go` only — `db.go` is not exempt. The same command on `development` reports `0 issues`, so this is new, and the workflow runs with `only-new-issues: true`, which means it is precisely the kind of issue it will report.

Worth flagging now because the main pipeline has not run on this head yet — only the Snyk check has — so nothing has told you about it. It needs a maintainer to approve the workflow run.

Smallest fix, verified as `0 issues` with tests still passing:

```go
// app_sql_stats is published in milliseconds; the log line is in microseconds.
const microsecondsPerMillisecond = 1000

func sendStats(logger datasource.Logger, metrics Metrics, config *DBConfig, start time.Time, queryType, query string, args ...any) {
	duration := time.Since(start).Microseconds()
	...
		metrics.RecordHistogram(context.Background(), "app_sql_stats",
			float64(duration)/microsecondsPerMillisecond, "hostname", config.HostName,
			"database", config.Database, "type", getOperationType(query))
```

The constant also answers the question the bare `1000.0` raises for the next reader, which is not "what is a thousand" but "which way is this converting".

If you would rather have no constant at all, this is equivalent and also lints clean — `time.Duration` is nanoseconds, so dividing by `time.Millisecond` is a unit conversion the compiler can see:

```go
	elapsed := time.Since(start)
	duration := elapsed.Microseconds()
	...
		metrics.RecordHistogram(context.Background(), "app_sql_stats",
			float64(elapsed)/float64(time.Millisecond), "hostname", config.HostName,
			"database", config.Database, "type", getOperationType(query))
```

I verified both: `0 issues` from `golangci-lint`, all three mutations still killed, package green. Either is fine — take whichever you prefer.

Nothing else in the diff changed, and the two lines in `db.go` are exactly right. This is the last thing from me.

aryanmehrotra (approved):
The `mnd` blocker is gone and your version is better than what I suggested — keeping `duration` as a `time.Duration` and converting at each use site means neither consumer owns the unit, which is the actual shape of the bug this PR fixes. `golangci-lint` v2.12.2 reports `0 issues`, `gofmt`/`go vet` clean, package green at `-count=10` and under `-race`.

I then went looking for what the tests still cannot see, and found something I should have caught last round.

## The log unit — the subject of #3877 — has no test at all

I mutated both halves of `sendStats` and ran the **whole** package, not just the one test:

| mutation | |
|---|---|
| metric: drop the divisor (raw nanoseconds) | killed |
| metric: `float64(duration.Microseconds())` — reintroduces #2188 | killed |
| metric: `float64(duration.Milliseconds())` — today's truncation | **survived** |
| metric: divide by `time.Second` | killed |
| **log: `duration.Microseconds()` → `.Milliseconds()`** — reverts this PR | **survived** |
| log: `duration.Microseconds()` → `.Nanoseconds()` | **survived** |

The second-to-last row is the one that matters: **the entire log half of this PR can be reverted and `go test ./pkg/gofr/datasource/sql/...` still passes.** Nothing anywhere asserts what `sendStats` puts in `Log.Duration` — `TestPrettyPrint` builds a `Log` by hand with `Duration: 12912` and checks the formatting, and every other test passes `gomock.Any()` for the metric value through `setupMetrics`.

The third row is the same problem on the metric side: `1500 ms` is a whole number of milliseconds, so truncating to `int64` is invisible at that input — and sub-millisecond precision is exactly what the histogram change exists to provide.

`CONTRIBUTING.md` asks that "all new line additions should be covered", and both changed lines currently are not.

## The test, verified against all six

One test closes all three, by choosing an elapsed time that is *not* a whole number of milliseconds:

```go
// TestDB_sendOperationStats_units pins both units this file has had wrong, in opposite directions:
// the logged duration is microseconds (#3877) and the recorded metric is milliseconds (#2188).
// The elapsed time is deliberately not a whole number of milliseconds, so a metric value computed
// with Duration.Milliseconds — which cannot carry sub-millisecond precision — fails as well.
func TestDB_sendOperationStats_units(t *testing.T) {
	const (
		elapsed = 1500*time.Millisecond + 500*time.Microsecond
		wantUS  = int64(elapsed / time.Microsecond)
		wantMS  = float64(elapsed) / float64(time.Millisecond)
		slack   = 10.0 // scheduling jitter, in milliseconds
	)

	mockMetrics := NewMockMetrics(gomock.NewController(t))
	mockMetrics.EXPECT().RecordHistogram(
		gomock.Any(), "app_sql_stats",
		gomock.Cond(func(v float64) bool { return v >= wantMS && v < wantMS+slack }),
		"hostname", "host", "database", "unitsdb", "type", "SELECT",
	)

	out := testutil.StdoutOutputForFunc(func() {
		db := &DB{
			logger:     logging.NewMockLogger(logging.DEBUG),
			config:     &DBConfig{HostName: "host", Database: "unitsdb"},
			metrics:    mockMetrics,
			stopSignal: make(chan struct{}),
		}

		db.sendOperationStats(time.Now().Add(-elapsed), "SELECT", "SELECT * FROM users")
	})

	m := regexp.MustCompile(`(\d+) \[\]}`).FindStringSubmatch(out)
	require.Len(t, m, 2, "no duration in the debug log: %s", out)

	logged, err := strconv.ParseInt(m[1], 10, 64)
	require.NoError(t, err)

	// Microseconds. Milliseconds would log ~1500 and nanoseconds ~1.5e9, both outside this window.
	assert.GreaterOrEqual(t, logged, wantUS)
	assert.Less(t, logged, wantUS+int64(slack*1000))
}
```

It needs `regexp` and `strconv` added to the imports. With it applied, **all six mutations above are killed**, and it is not flaky: 200 runs clean, including 100 under sixteen CPU burners at `GOMAXPROCS=1`.

Two details worth having rather than rediscovering:

- The database name is `unitsdb`, not `db`, on purpose. `goconst` counts occurrences across test files even though it does not report them there, and a third `"db"` in the package tips `supabase.go:142` over the threshold — `0 issues` becomes `1 issues` from a test-only addition. With `unitsdb` it stays at `0`.
- `logging.NewMockLogger` renders the `Log` with `%v`, not JSON, so the line is `&{SELECT SELECT * FROM users 1500500 []}` — hence matching the number before the args slice rather than a `"duration"` key.

Say the word and I will push it, rather than spend another round on a test.

## What the change is actually worth, measured on a live `/metrics` scrape

I said in the last round that the fractional buckets from #3568 are unreachable today. I had not verified that on a real scrape, so I did: one GoFr app against a live MySQL, 161 queries through `ctx.SQL.QueryRowContext`, then `GET :METRICS_PORT/metrics`.

| `app_sql_stats_bucket{le=…}` | `development` | with this PR |
|---|---|---|
| `0.05` | **155** | 0 |
| `0.5` | 155 | **95** |
| `0.75` | 155 | 133 |
| `1` | 160 | 145 |
| `+Inf` | 161 | 161 |
| **`app_sql_stats_sum`** | **7** | **98.66** |
| `app_sql_stats_count` | 161 | 161 |

Same workload, same server, same 161 queries. Today 155 of them record `0` and pile into the bottom bucket, so every percentile below p96 reads "≤50µs" — and **the reported total SQL time is 7 ms against an actual 98.66 ms, an under-report of 14x**. Nine of the fifteen sub-millisecond buckets go from permanently empty to carrying the bulk of the distribution.

The debug log from the same run, for the record: `938µs`, `1434µs`, `2112µs`, `2925µs`, `4935µs` — each of which is `0µs` or `1µs` on `development` today.

That is a bigger correction than the description claims, and it is worth putting in the description before this merges.

Approving on the production change. The only thing I would like in before merge is that test, and I am happy to push it myself.

PR #3919, opened 2026-08-11
fix(examples): stop tracking compiled example binaries + gitignore them

Umang01-hash:
## Problem

Three compiled example binaries (~183 MB total) were accidentally committed in #3854 and are currently tracked on `development`:

| Path | Size | Type |
|------|------|------|
| `examples/using-add-filestore/using-add-filestore` | 55 MB | Mach-O arm64 |
| `examples/using-cloudsql/using-cloudsql` | 62 MB | Mach-O arm64 |
| `examples/using-s3-filestore/using-s3-filestore` | 66 MB | Mach-O arm64 |

## Root cause

`go build` inside `examples/<name>/` emits a binary named after the directory with **no extension**. The root `.gitignore` only ignored `*.exe`/`*.test`, so these slipped through when they were staged.

## Fix

1. `git rm --cached` the three binaries (removes them from `development` going forward — a plain commit, no history rewrite).
2. Add a `.gitignore` rule that ignores any extensionless example build artifact while keeping sources, Dockerfiles, and subdirectories (`configs/`, `static/`).

## Testing

- `git check-ignore` matches all three binary paths and any freshly-built example binary.
- A live `go build` in `examples/http-server` produces a 58 MB binary that yields an **empty** `git status` (ignored).
- Verified **no** currently-tracked example file (`main.go`, `go.mod`, `go.sum`, `static/*.css`, `Dockerfile`) becomes ignored by the new rule.

## Note on history

This removes the binaries from `development` **HEAD** only — it does **not** reclaim any space. All three blobs stay in the pack and in the clone object set for this head, so `git clone` costs the same after merge (86.7 MiB packed / 183.3 MiB unpacked). That space returns only with a history rewrite of `development` (force-push, maintainer action), which should land **before v1.59.0 is tagged**, since `sum.golang.org` pins tag history permanently. Out of scope for this PR.

## Prevention — block future binaries in CI

`.gitignore` can't reliably catch compiled binaries repo-wide (a Go binary is named after its directory with no extension, and a dotted name at mode 644 defeats any glob), so this PR adds a content- and mode-based guard:

- **`.github/scripts/check-binaries.sh`** — fails if any tracked file is mode `100755` or carries executable magic (ELF / Mach-O / PE). Executable formats only, so tracked images / minified JS / zip are fine. Repo-wide, so a future `cmd/` or `tools/` dir is covered too.
- Wired into the existing **`Linting Party`** job — no extra dependencies, ~0.6s, works on the default shallow checkout.

Also widens the example-binary ignore from `examples/*/*` to `examples/**/*`, which was missing the four `examples/grpc/*` binaries one directory deeper.

Verified: fails on `development` (the 3 binaries), passes on this PR head and `main`, catches Mach-O/ELF/PE committed at mode 644, and no false positives on tracked assets. Recommend making this a required status check on `development`.

Credit: binary-guard script and the `**` fix from @aryanmehrotra's review.


aryanmehrotra:
Reviewed this end to end. The deletion is right and we need it. Two findings, and then a proposal I would like your read on — I think it is both cleaner and, unlike a path rule, correct in every case.

## First, what this PR does and does not do

Deleting a tracked file removes it from the tree, not from history. The blobs stay in the pack and every clone still transfers them. Measured, by resolving each blob at the commit before its deletion and then checking the object set of this PR's own head:

| binary | uncompressed | packed |
|---|---|---|
| `using-s3-filestore` | 65.8 MiB | 30.8 MiB |
| `using-cloudsql` | 62.2 MiB | 29.3 MiB |
| `using-add-filestore` | 55.3 MiB | 26.6 MiB |
| | **183.3 MiB** | **86.7 MiB** |

All three are still in the 33,229-object set reachable from `2e15aad5`, so `git clone` costs the same after this merges. Narrower than it looks, though: the blobs are on `development` and **not** on `main`, and `development` is the default branch — so contributors pay roughly 80% of a fresh clone, while Go users resolving `gofr.dev` through the proxy fetch tags off `main` and are unaffected.

Not an argument against merging — just worth not describing this as reclaiming the space. The 86.7 MiB comes back only with a history rewrite of `development`, which needs force-push access and should land before v1.59.0 is tagged, since `sum.golang.org` pins tag history permanently.

## Second, a live gap: the four grpc examples

`examples/*/*` matches exactly two path segments below `examples/`. The grpc examples are one level deeper:

| path | covered? |
|---|---|
| `examples/using-cloudsql/using-cloudsql` | yes |
| `examples/grpc/grpc-unary-server/grpc-unary-server` | **no** |
| `examples/grpc/grpc-unary-client/grpc-unary-client` | **no** |
| `examples/grpc/grpc-streaming-server/grpc-streaming-server` | **no** |
| `examples/grpc/grpc-streaming-client/grpc-streaming-client` | **no** |

All four have a `main.go`, so `go build` in any of them drops exactly the artifact this rule exists to stop, in the only directories it does not watch. `examples/*/*` to `examples/**/*` on all four lines fixes it — I applied that and ran `git check-ignore` over every path in `git ls-files examples/`: all four grpc binaries become ignored and no currently tracked file breaks.

## The proposal

The rule is path-scoped, but the thing we are preventing is not. I checked: there are 25 `package main` directories and all 25 happen to be under `examples/` today, so with `**` the coverage is complete — but by coincidence, not construction. A future `cmd/` or `tools/` directory would be unguarded, which is exactly how this happened: the existing `*.exe` / `*.so` / `*.dylib` / `*.test` rules are all extension-based and none of them match extensionless output.

More importantly, **no `.gitignore` can close this**. I tested a binary with a dotted name at mode 644:

| committed file | mode | `.gitignore` catches it? |
|---|---|---|
| `examples/foo/foo` | 755 | yes |
| `examples/grpc/x/x` | 755 | no, without `**` |
| `examples/foo/server.bin` | 644 | **no — `!examples/**/*.*` re-includes it** |

That last row is the problem. Any dotted name defeats the ignore rule by construction, and the re-include is load-bearing (we need `go.mod`, `main.go`, `.env`).

So I would rather enforce the invariant directly, in the pipeline, and let `.gitignore` go back to being a convenience that keeps `git status` clean. Two independent signals, because each alone has a hole:

- **mode `100755`** — catches build output whatever its bytes. The repo has **zero** legitimately tracked `100755` files, checked across the whole tree.
- **executable magic** (ELF / Mach-O / PE) — catches a binary committed as `100644`.

Deliberately executable formats only, never "is binary" — we legitimately track `banner.gif` at 1.67 MiB, the minified swagger bundles, and the PNG/JPEGs under `docs/public`.

I wrote it and tested it rather than sketching it:

| scenario | expected | result |
|---|---|---|
| `development` — the three real binaries | fail | fails, both signals, reported as Mach-O |
| this PR's head | pass | passes |
| `main` | pass | passes |
| Mach-O at mode 644 | fail | caught by magic alone |
| Linux ELF at mode 644 | fail | caught by magic alone |
| `server.bin` at 644 (ignore rule misses it) | fail | caught by magic alone |
| `banner.gif`, swagger JS, images | pass | no false positives |
| shallow checkout (`depth 1`, the CI default) | works | still detects |
| runtime, whole repo | — | **0.63s** |

Two CI details worth knowing: it needs **no `fetch-depth: 0`**, since `git ls-tree -r HEAD` and `cat-file` work fine on the default shallow checkout; and it must batch. My first version spawned one `git cat-file` per blob and took **17.3s** — piping every OID through a single `--batch` process brought it to 0.63s with identical detection.

<details>
<summary><code>.github/scripts/check-binaries.sh</code></summary>

```bash
#!/usr/bin/env bash
# Fail if any tracked file is a compiled executable.
#
# Two independent signals, because either alone has a hole:
#   - mode 100755      catches build output whatever its contents
#   - executable magic catches a binary committed with mode 100644
#
# Matches executable formats only (ELF / Mach-O / PE), never "is binary" —
# the repo legitimately tracks images and minified JS.
#
# Every blob is streamed through a single `git cat-file --batch`, so this is
# one subprocess regardless of repository size.

set -uo pipefail

ref="${1:-HEAD}"
fail=0

# ---- signal 1: the executable bit -------------------------------------------
# No tracked file in this repo legitimately carries mode 100755. If that ever
# changes, allowlist the specific path rather than dropping the check.
while IFS= read -r path; do
  [ -n "$path" ] || continue
  printf '  executable bit set: %s\n' "$path"
  fail=1
done < <(git ls-tree -r "$ref" | awk -F'\t' '$1 ~ /^100755/ { print $2 }')

# ---- signal 2: executable file magic ---------------------------------------
magic_hits=$(
  git ls-tree -r "$ref" |
  perl -MIPC::Open2 -e '
    my (%path, @oids);
    while (<STDIN>) {
      chomp;
      my ($meta, $p) = split /\t/, $_, 2;
      my (undef, $type, $oid) = split /\s+/, $meta;
      next unless $type eq "blob";
      $path{$oid} = $p;
      push @oids, $oid;
    }
    my ($out, $in);
    open2($out, $in, "git", "cat-file", "--batch");
    binmode $out;
    for my $oid (@oids) { print $in "$oid\n" }
    close $in;

    my %fmt = (
      "\x7fELF"          => "ELF",
      "\xfe\xed\xfa\xce" => "Mach-O", "\xce\xfa\xed\xfe" => "Mach-O",
      "\xfe\xed\xfa\xcf" => "Mach-O", "\xcf\xfa\xed\xfe" => "Mach-O",
      "\xca\xfe\xba\xbe" => "Mach-O universal",
      "\xbe\xba\xfe\xca" => "Mach-O universal",
    );

    while (my $hdr = <$out>) {
      my ($oid, $type, $size) = split /\s+/, $hdr;
      last unless defined $size;
      my $buf = "";
      read($out, $buf, $size) if $size > 0;
      read($out, my $lf, 1);

      my $head = substr($buf, 0, 4);
      my $f = $fmt{$head};
      $f = "PE" if !defined($f) && substr($buf, 0, 2) eq "MZ";
      next unless defined $f;
      print "  $f executable: $path{$oid}\n";
    }
  '
)

if [ -n "$magic_hits" ]; then
  printf '%s\n' "$magic_hits"
  fail=1
fi

if [ "$fail" -ne 0 ]; then
  cat <<'MSG'

Compiled executables must not be committed.
`go build` inside a package main directory writes an extensionless binary
named after that directory. Build with `-o` into a temp path, or delete the
artifact before committing.
MSG
  exit 1
fi

echo "no committed executables"
```

</details>

Wiring, into the existing `Linting Party🥳` job in `go.yml`, right after the ls-lint step:

```yaml
      - name: No committed executables
        run: bash .github/scripts/check-binaries.sh
```

A script rather than an inline `run:` so it can be run locally and reused as a pre-commit hook.

## What I would like from you

Does this look right to you, and are you happy to fold it into this PR — or would you rather it went in separately? I am not attached to my implementation; if you would rather write it differently that is fine, the part I care about is that the guard is content-based and repo-wide rather than a path glob we have to remember to extend.

If you do take it, please re-run the cases yourself rather than trusting my table — particularly on the Linux runner, since I tested Mach-O locally and only cross-compiled the ELF case. The two worth confirming are that `development` fails and that this PR's head passes, and I would add one deliberately committed binary in a throwaway branch to watch it fail in CI for real. My own first two attempts at the shallow-clone test were both invalid — git silently ignores `--depth` on a local clone, and I picked up a stale local `development` branch that predates the binary commit — so the check looked like it passed on a tree that should have failed. Worth being careful there.

Also: whichever way the check goes, the `examples/*/*` to `examples/**/*` change should land here regardless, since the four grpc examples are unguarded today.

Umang01-hash:
Agreed on all of it — folded your approach in (27628cee).

- Took `check-binaries.sh` as-is, wired into `Linting Party`, and **dropped my pre-commit/forbid-binary**. Yours is content+mode with no name-exclude hole, so it supersedes mine — no point running two guards.
- Applied `examples/*/*` → `examples/**/*`; confirmed all four grpc binaries are now ignored and no tracked file breaks.

Re-ran the cases myself rather than trusting the table:
- `development` → fails (mode + Mach-O on all three); this PR head and `main` → pass.
- Mach-O / **ELF** / PE committed at mode **644** → all caught by magic, including a dotted-name `server.bin` (the case that beats both the ignore rule and an extension exclude).
- No false positives on `banner.gif`, swagger JS, or the `docs/public` images.

One minor note: when two tracked files are byte-identical, git dedups the blob so only one path prints — still fails, just under-reports the second path. Not worth fixing unless you want the list exact.

On history: agreed it doesn't reclaim the 86.7 MiB packed — that needs the `development` rewrite + force-push, which we're queuing with the maintainer to land before v1.59.0 is tagged. Thanks for the careful review.

aryanmehrotra:
Thanks for turning this around so fast. I verified your head rather than assuming my own testing carried over, and the guard works:

| check | result |
|---|---|
| the script against its own head | passes, and correctly does not flag itself |
| the script against `development` | 6 detections — 3 by mode, 3 as Mach-O |
| `examples/**/*` on the nested grpc binaries | now ignored |
| any tracked file newly ignored | none |
| real builds, flat and nested grpc (58 MB and 59 MB) | `git status` clean, `git add -A` stages nothing |
| **the new step on the Linux runner** | `no committed executables`, step green |

That last row closes the gap I flagged — I had only cross-compiled the ELF case locally, so it is good to see it pass on the real runner.

Also, `068640a8` is a good catch on my design. Signal 1 asserts that no tracked file carries mode `100755`, which means the guard's own script cannot be executable; committing it `644` and invoking it through `bash` is the right call. The constraint it leaves behind is that no shell script in this repo can ever be `755`. That is fine today, but if it ever gets in the way the clean refinement is to exempt files whose first two bytes are `#!` from signal 1 — a script is not a compiled binary. Not worth doing now.

## One thing I would change, and it reverses my earlier ask

I asked for `examples/*/*` to become `examples/**/*`. Having tested further, I think the blanket is the wrong shape and a narrower rule is better. Two findings changed my mind.

**`.gitignore` cannot enforce this, in any form.** I built a binary under your rule — correctly ignored, `git status` clean — then ran `git add -f`:

```
STAGED ANYWAY: examples/using-cron-jobs/using-cron-jobs
```

`git add -f` exists specifically to override `.gitignore`, so the rule is advisory by construction. Separately, a dotted name defeats it outright: `examples/foo/server.bin` at mode 644 is re-included by `!examples/**/*.*` and sails past. Your check catches both cases — the force-added one and the dotted one.

So the check is the guard, and `.gitignore` is only buying convenience: without any rule, `go build` leaves a 58 MB file in `git status` and `git add -A` stages it. That convenience is worth having, but it means the ignore rule should be judged on cost alone.

**On that basis the blanket costs more than it returns.** Its only advantage over a narrow rule is covering a future `cmd/` or `tools/` directory — which your check already covers, repo-wide. What it costs is real:

| | `examples/**/*` + re-includes | narrow rule |
|---|---|---|
| covers all 25 example binaries | yes | yes |
| silently blocks `Makefile`, `LICENSE`, `README`, `Procfile` | **yes** | no |
| needs the `Dockerfile` special case | **yes** | no |
| covers a future `cmd/` directory | yes | no — the check does |
| failure mode | **silent**: `git add` does nothing and says nothing | loud: file shows untracked, check blocks it |

The `Dockerfile` line is the tell — the next extensionless file anyone adds will need its own exception, and until someone notices, `git add` will quietly ignore it.

The narrow version, which I tested — all 25 example binaries covered, nothing over-ignored, and `git status` clean after building both a flat example and a nested grpc one:

```
# Compiled example binaries: `go build` inside an example directory emits an
# extensionless executable named after that directory.
/examples/using-*/using-*
/examples/http-server*/http-server*
/examples/sample-cmd/sample-cmd
/examples/grpc/*/grpc-*
```

Its weakness is that a new example whose binary does not match those prefixes would not be ignored — but that fails loudly, as an untracked file you can see, with the check as the backstop. The blanket fails silently, which is the worse direction now that something else is enforcing.

Not requesting changes over it — the guard is the part that matters and you have that. Your call.

## Two smaller notes

The description still reads as reclaiming the space. It does not: all three blobs remain in the pack and in the clone object set for this head, so `git clone` costs the same after merge — 86.7 MiB packed, 183.3 MiB unpacked. Worth rewording so nobody reads this as the cleanup being done. The 86.7 MiB comes back only with a history rewrite of `development`, which needs force-push access and should land before v1.59.0 is tagged, since `sum.golang.org` pins tag history permanently.

Worth confirming separately that `Linting Party🥳` is a **required** status check on `development`. I could not read the branch protection settings, so I cannot tell whether a red run actually blocks a merge or is only advisory. If it is not required, the guard is loud rather than binding, which is worth fixing while this is fresh.

Finally — your first attempt in `9292e68c` used a pre-commit `forbid-binary` hook. Worth keeping in mind as a complement rather than a replacement: a pre-commit hook is the only layer that catches this before the commit instead of after the push, though it is bypassable with `--no-verify` and needs contributors to install it.

Umang01-hash:
Good call — you convinced me, switched to the narrow rule (`34f40509`):

```
/examples/using-*/using-*
/examples/http-server*/http-server*
/examples/sample-cmd/sample-cmd
/examples/grpc/*/grpc-*
```

Re-verified: all 25 example binaries ignored, nothing tracked over-ignored, `git status` clean after building a flat and a nested grpc example. Agreed the blanket's silent `git add` swallow is the worse failure mode now that the check is the real guard — and the `Dockerfile` exception was exactly the smell you called out.

- **Description**: reworded so it no longer reads as reclaiming space — now states clone cost is unchanged (86.7 MiB packed / 183.3 MiB unpacked) until the `development` rewrite, flagged to land before v1.59.0.
- **Required check**: I can't read the branch-protection settings for `development`, so I can't confirm whether `Linting Party🥳` is required or advisory — flagging to the maintainer to make it required, otherwise the guard is loud but not binding.
- **`#!`-exempt for signal 1**: noted for if a `755` script ever becomes necessary; agree it's not worth doing now.
- **pre-commit as a complement**: agreed it's the only layer that catches pre-commit rather than post-push. Happy to add a `local`-repo pre-commit hook that just runs `check-binaries.sh` (opt-in, `--no-verify`-bypassable) if you'd like that layer too — say the word and I'll add it here or in a follow-up.

Thanks again for the thorough passes.

aryanmehrotra:
The narrow rule is right and I verified it independently at `34f40509` — all 25 example binaries ignored, nothing tracked over-ignored, `Makefile`/`LICENSE`/`README`/`Procfile` all committable again, the `Dockerfile` exception gone, and `git status` clean after building both a flat example (57.9 MB) and a nested grpc one (59.0 MB). The reworded description is accurate, including the `#3854` attribution — I checked, `f8ead999` is indeed "consolidate minor/patch dependency updates (#3854)".

Before you merge, though: **the script I handed you has a fail-open bug, and it is mine.** I went looking for it because a guard that can report success without having inspected anything is worse than no guard.

## Three defects

**It fails open on an unresolvable ref.**

```
$ check-binaries.sh does-not-exist-ref
fatal: Not a valid object name does-not-exist-ref
no committed executables            <- exit 0
```

`git ls-tree` fails, the loop reads nothing, the script declares success. Not reachable through the current wiring, since the workflow passes no argument and `HEAD` always resolves — but it is a false green waiting for the first person who passes a ref.

**It fails open if `perl` cannot run.** With a real ELF binary committed at mode `100644`, so only signal 2 can see it:

| | v1 |
|---|---|
| perl working | exit 1, `ELF executable: artifact` |
| perl broken | **exit 0, `no committed executables`** |

**And the one that matters most: there is no positive control.** On a clean tree, these two are byte-identical:

```
perl working : no committed executables
perl broken  : no committed executables
```

So a green run proves the script *ran*, not that detection *works*. Which means my own sign-off was overstated: I told you the green step on the Linux runner "closes the ELF gap." It does not. The runner's tree is clean, so that pass is equally consistent with signal 2 never having executed. I have still not demonstrated ELF detection on the runner — only locally, cross-compiled.

That is the third time in these reviews I have had to correct the same class of mistake — a check with no failing case proves nothing. I should have caught it in my own script first.

## Hardened version

Three changes: fail closed with a distinct exit code, and a self-test that runs before the real scan so a green result is meaningful.

```
0  inspected, nothing found
1  inspected, executables found
2  could not inspect  (bad ref, perl missing, scan error, self-test failed)
```

The separate `2` matters — "could not check" must never be indistinguishable from "clean".

The self-test builds a throwaway repo containing a file with ELF magic at mode 644 (only signal 2 can see it) and a plain script at mode 755 (only signal 1 can), and asserts each signal fires. If either is broken on that machine, the run exits 2 instead of quietly passing.

<details>
<summary>diff against the version on this branch</summary>

```bash
# --- fail closed -------------------------------------------------------------
cannot_check() {
  printf 'check-binaries: cannot verify: %s\n' "$*" >&2
  exit 2
}

command -v git  >/dev/null 2>&1 || cannot_check "git not available"
command -v perl >/dev/null 2>&1 || cannot_check "perl not available (signal 2 would silently pass)"

git rev-parse --verify --quiet "${ref}^{tree}" >/dev/null 2>&1 \
  || cannot_check "cannot resolve ref: ${ref}"

modes=$(scan_modes "$ref") || cannot_check "mode scan failed for ${ref}"
magic=$(scan_magic "$ref") || cannot_check "magic scan failed for ${ref}"

# --- positive control --------------------------------------------------------
self_test() {
  local dir out
  dir=$(mktemp -d) || cannot_check "mktemp failed"

  (
    cd "$dir" || exit 1
    git init -q -b main . || exit 1
    printf '\177ELF\002\001\001\000' > elf-fixture   # magic only; mode 644
    printf '#!/bin/sh\ntrue\n'        > exec-fixture  # mode only; not a binary
    chmod 644 elf-fixture
    chmod 755 exec-fixture
    git add elf-fixture exec-fixture || exit 1
    # -c overrides so a developer's global commit.gpgsign / hooks / template
    # cannot make the fixture commit fail and trip a false "cannot verify".
    git -c user.email=self@test -c user.name=self \
        -c commit.gpgsign=false -c tag.gpgsign=false \
        -c core.hooksPath=/dev/null -c commit.template= \
        commit -qm fixture || exit 1
  ) || { rm -rf "$dir"; cannot_check "could not build the self-test fixture"; }

  out=$(cd "$dir" && scan_magic HEAD) \
    || { rm -rf "$dir"; cannot_check "magic scan errored during self-test"; }
  case "$out" in
    *"ELF executable: elf-fixture"*) ;;
    *) rm -rf "$dir"; cannot_check "magic scan did not detect the ELF fixture (signal 2 is not working here)" ;;
  esac

  out=$(cd "$dir" && scan_modes HEAD) \
    || { rm -rf "$dir"; cannot_check "mode scan errored during self-test"; }
  case "$out" in
    *"executable bit set: exec-fixture"*) ;;
    *) rm -rf "$dir"; cannot_check "mode scan did not detect the 755 fixture (signal 1 is not working here)" ;;
  esac

  rm -rf "$dir"
}
```

Also: `--self-test` to run only the control, and `--no-self-test` to skip it. The success line becomes `no committed executables (detection self-tested)` so the log distinguishes a verified pass from a bare one.

</details>

## Tested locally, both versions, same cases

| case | v1 (on this branch) | hardened |
|---|---|---|
| clean tree at `34f40509` | 0 | 0 |
| `development`, the 3 real binaries | 1 | 1 |
| unresolvable ref | **0** | **2** |
| clean tree, perl broken | **0** | **2** |
| real 644 ELF tracked, perl broken | **0 — misses it** | **2** |
| clean tree, `commit.gpgsign=true` | 0 | 0 |

`0` clean, `1` executables found, `2` cannot verify.

Detection matrix on the hardened version, each artifact committed in its own commit and confirmed present in the tree before scanning:

| format | mode 644 | mode 755 |
|---|---|---|
| Linux ELF | ELF, exit 1 | ELF, exit 1 |
| macOS Mach-O | Mach-O, exit 1 | Mach-O, exit 1 |
| Windows PE | PE, exit 1 | PE, exit 1 |

No false positives: the repo's 12 tracked images/fonts and 4 swagger bundles are untouched, exit 0. Works on a shallow clone (`is_shallow=true`, 1 commit), so no `fetch-depth: 0` needed. Runtime on this repo is **0.21s with the self-test, 0.11s without** — the control costs about a tenth of a second.

Two notes on my own testing, since I got both wrong the first time and it is worth you not repeating them. My first detection matrix silently measured nothing: the builds had failed, `probe` never existed, and every "detection" was a leftover fixture still in the tree — so I added an `in-tree=1` assertion before each scan. And one row of the comparison above was originally labelled "644 ELF present" when that repo's HEAD was actually clean; I re-ran it with the artifact genuinely committed, which is where the "misses it" result comes from.

The `commit.gpgsign` row is also a real finding rather than a hypothetical: the first version of the self-test failed with `could not build the self-test fixture` on any machine that signs commits by default, which would have been a confusing false alarm for contributors. Hence the `-c` overrides.

## What I would do

Take the hardening, then this is good to merge from my side. Happy for you to apply it here, or I can open it as a follow-up against `development` after this lands — your call, and no objection if you would rather write it differently, as long as "cannot verify" cannot masquerade as "clean".

Still outstanding and not for this PR: someone with admin needs to confirm `Linting Party🥳` is a **required** status check on `development`. Neither of us can read the branch protection. If it is not required, all of this is advisory rather than binding, which would be a shame given the work.

On the pre-commit hook you offered — yes please, but as a follow-up rather than here. This PR is already carrying the deletion, the ignore rule and the CI guard, and the hook is a third layer with its own trade-offs (`--no-verify`, contributors must install it).

Umang01-hash:
Proved the guard in real CI, not just locally — ran it on a throwaway fork (`Umang01-hash/gofr`) so no binary ever touches this repo.

**Green vs red, same `Linting Party🥳` job on the Linux runner:**

| run | committed | `No committed executables` |
|---|---|---|
| [clean head `34f40509`](https://github.com/Umang01-hash/gofr/actions/runs/31466453535/job/93700244879) | — | ✅ pass |
| [+ a deliberately committed binary](https://github.com/Umang01-hash/gofr/actions/runs/31466865677/job/93701465674) | `proof/leaked-binary` (real Mach-O, mode 755, non-`examples/` path) | ❌ fail, exit 1 |

Failing step output:
```
  executable bit set: proof/leaked-binary
  Mach-O executable: proof/leaked-binary
Compiled executables must not be committed.
##[error]Process completed with exit code 1.
```

Both signals fired — mode `100755` **and** Mach-O magic — on a **Linux** runner against a **Mac-built** binary at a **non-example** path, i.e. the exact repo-wide, cross-arch case. The clean head passes the same step, so no false positive. Fork branch has been reset; the binary is not retained anywhere in `gofr-dev/gofr`.

aryanmehrotra (approved):
Approving at `34f40509`. The CI proof is what settled it — I was blocking on "a green run cannot distinguish a working detector from a dead one", and you answered that with an actual failing run rather than an argument.

## Your proof, verified

I checked the runs rather than reading the pasted output, because the step conclusion is the thing that matters:

| run | `Linting Party🥳` job | `No committed executables` step |
|---|---|---|
| clean head `34f40509` | success | **success** |
| `ece42083`, binary committed | failure | **failure** |

And the artifact was genuinely what it needed to be: `proof/leaked-binary`, `mode=100755`, 1,026,690 bytes, first four bytes `cffaedfe` — a real 64-bit Mach-O, at a path outside `examples/`. So both signals fired, on a Linux runner, against a Mac-built binary, at a non-example path. That is the repo-wide cross-architecture case demonstrated rather than assumed.

Containment checks out too: zero occurrences of `proof/leaked-binary` anywhere in this repository's object graph and no `proof` refs on `gofr-dev/gofr`.

One thing worth saying out loud so nobody misreads it later: **both** of those runs show `failure` at the run level. On the clean one that is `Code Quality🎖️`, an unrelated job on your fork — the guard step passed. Anyone glancing at the run badge will draw the opposite conclusion from the truth.

## Checks on this head

| | |
|---|---|
| deletion | complete — largest tracked file is now `.github/banner.gif` at 1.67 MiB, and nothing above 500 KB is a build artifact |
| narrow ignore rule | all 25 `package main` directories' binaries ignored; **zero** over-ignoring across every path in `git ls-files`; `Makefile` / `LICENSE` / `README` / `Procfile` committable again; the `Dockerfile` exception gone |
| real builds | flat example 57.9 MB and nested grpc 59.0 MB both leave `git status` empty and `git add -A` staging nothing |
| script mode | `100644`, so the guard does not trip over itself |
| guard on this repo's runner | step `success` |
| CI | 18 success, 2 skipped, nothing failing |
| description | accurate, including `#3854` — `f8ead999` is indeed "consolidate minor/patch dependency updates (#3854)" |

I also built the merge result rather than only testing the branch, since that is what post-merge CI will actually see: fast-forward, zero conflicts, no binary entries in the merged tree, and the guard passes on it. So `development` will be green on the first push after this lands.

## On the hardening — not a condition

I raised three fail-open paths in the script I gave you and framed them as blocking. That was the wrong call, and the reason is your CI proof: none of them is reachable today.

| path | reachable now |
|---|---|
| unresolvable ref returns 0 | no — the workflow passes no argument and `HEAD` always resolves |
| broken `perl` returns 0 | no — `ubuntu-latest` ships perl, and git itself depends on it |
| no in-script positive control | latent — it protects against a *future* runner change silently killing signal 2 |

All three are latent robustness rather than live holes, and the guard is now demonstrated working. Holding a 183 MB cleanup with a tag deadline behind that would be the wrong trade. The hardened version — distinct exit code `2` for "could not verify" plus a self-test that proves both signals fire before each scan — is worth having, but as a follow-up. I have it written and tested; I will open it against `development` once this is in, unless you would rather do it yourself.

## Three things that outlive this PR

**The rewrite is still owed.** This removes the binaries from `development`'s HEAD only. All three blobs remain in the pack and in the clone object set — 86.7 MiB packed, 183.3 MiB unpacked — so `git clone` costs the same tomorrow as today. That returns only with a history rewrite of `development`, which needs force-push access and has a hard deadline: once v1.59.0 is tagged, `sum.golang.org` pins that history permanently. Your description says this correctly; I am repeating it here so it is on the review record.

**Someone with admin needs to confirm `Linting Party🥳` is a required status check on `development`.** Neither of us can read the branch protection. If it is not required, everything above is advisory rather than binding.

**The pre-commit hook**, as agreed — separate PR. It is the only layer that catches this before the commit rather than after the push, and it deserves its own discussion about `--no-verify` and contributor setup.

Good iteration on this one. Four rounds and the shape improved at every step, including the two places where you were right and I was not.

PR #3921, opened 2026-08-12
chore(lint): clear mechanical golangci-lint backlog in core packages

Umang01-hash:
## What
Clears the golangci-lint backlog that briefly surfaced on `development` after a force-push reset the `only-new-issues` baseline (never a regression). Reworked per @aryanmehrotra's review.

## The lever: `.golangci.yml`, not 46 files
- `goconst`: `min-len`/`min-occurrences` **2 → 3** (upstream defaults). Removes ~71 manufactured two-occurrence hits (`"true"`, `"db"`, single-word labels) whose constant names carry no more information than the literal.
- `gosec`: exclude **G101 in `_test.go`** (joins the existing `dupl`/`goconst`/`mnd` test exclusion). Kills 20 test-fixture "hardcoded credential" false positives at the source and removes the inline `//nolint`s.

## Code changes (behaviour-preserving unless noted)
- **goconst**: extract only the genuinely-repeated literals to constants (reuse existing `dialectMysql`/`dialectPostgres`/`requireSSLMode`); revert the two-occurrence noise back to literals; complete-or-revert the partial sets.
- **service**: string constants renamed `err*` → `msg*` (`err*` is reserved for `error` sentinels; the package already has real `errX = errors.New(...)`).
- **staticcheck** (`exporter.go`): `attribute.INVALID`→`EMPTY`, `Value.Emit`→`String` (verified byte-identical in OTel source).
- **prealloc / testifylint / noctx / govet / whitespace**: mechanical.

## ⚠️ One behavioural change (for the release notes)
`Request.Params(key)` now returns an **empty non-nil slice** for an absent key, where it previously returned `nil` — so `{"data":null}` becomes `{"data":[]}` for a handler returning `c.Params(missing)`. This honours the method's own godoc ("an empty slice is returned") and matches `cmd`'s `Params`. Locked in by `assert.NotNil` in `Test_Params`. This is the only behavioural change in the PR.

## Also fixed from review
- `Test_CreateHandler`: restored the no-error assertion for `nil` expectations (the guarded `ErrorAs` had silently dropped it — `assert.IsType(nil, err)` was asserting `err == nil`).
- `exporter.go`: dropped the `exhaustive` BYTESLICE/SLICE case that had forced a `gocyclo` suppression — kept only the deprecation fixes, no new suppression.
- `opentsdb_test` prealloc reverted (caps counted append calls, not elements); `request_test` uses `t.Context()`.

## Deferred to separate reviewed PRs
- Real bugs found via gosec: `auth.go:35` G602 slice OOB, `kafka/helper.go:297` G118 leaked cancel.
- `cloud.google.com/go/pubsub` → v2 migration; `gocyclo` refactors.

## Verification
`go build ./...` + `go vet` pass; touched-package `go test -short` passes; final lint is CI here (`only-new-issues`, clean checkout).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -59,8 +59,8 @@ linters:
       lines: 100
       statements: 50
     goconst:
-      min-len: 2
-      min-occurrences: 2
+      min-len: 3
+      min-occurrences: 3
     gocritic:
       disabled-checks:
         - dupImport # https://github.com/go-critic/go-critic/issues/845
@@ -154,6 +154,10 @@ linters:
           - dupl
           - goconst
           - mnd
+        path: _test\.go
+      - linters:
+          - gosec
+        text: "G101"
         path: _test\.go
       - linters:
           - revive
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -59,7 +59,7 @@ linters:
       lines: 100
       statements: 50
     goconst:
-      min-len: 3
+      min-len: 2
       min-occurrences: 3
     gocritic:
       disabled-checks:
--- a/examples/using-s3-filestore/go.mod
+++ b/examples/using-s3-filestore/go.mod
@@ -114,7 +114,7 @@ require (
 	google.golang.org/genproto v0.0.0-20260319201613-d00831a3d3e7 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20260526163538-3dc84a4a5aaa // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260630182238-925bb5da69e7 // indirect
-	google.golang.org/grpc v1.82.0 // indirect
+	google.golang.org/grpc v1.83.0 // indirect
 	google.golang.org/protobuf v1.36.11 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	modernc.org/libc v1.73.4 // indirect
--- a/examples/using-s3-filestore/go.sum
+++ b/examples/using-s3-filestore/go.sum
@@ -393,8 +393,8 @@ google.golang.org/grpc v1.23.0/go.mod h1:Y5yQAOtifL1yxbo5wqy6BxZv8vAUGQwXBOALyac
 google.golang.org/grpc v1.25.1/go.mod h1:c3i+UQWmh7LiEpx4sFZnkU36qjEYZ0imhYfXVyQciAY=
 google.golang.org/grpc v1.27.0/go.mod h1:qbnxyOmOxrQa7FizSgH+ReBfzJrCY1pSN7KXBS8abTk=
 google.golang.org/grpc v1.33.2/go.mod h1:JMHMWHQWaTccqQQlmk3MJZS+GWXOdAesneDmEnv2fbc=
-google.golang.org/grpc v1.82.0 h1:vguDnZUPjE26w09A63VoxZPnvPjB5Riyc0mkXPFmAIU=
-google.golang.org/grpc v1.82.0/go.mod h1:yzTZ1TB1Z3SG+LIYaI+WiE8D5+PZ3ArnrSp8zF3+/ZA=
+google.golang.org/grpc v1.83.0 h1:JeNZEKJFbQxArAMl+hiytHauacDNqJUllNfmIMmpqnQ=
+google.golang.org/grpc v1.83.0/go.mod h1:kDyl6SKsiHKt0uylY5gtn5cEjkrIOhQOGDgIc4JGwzQ=
 google.golang.org/protobuf v0.0.0-20200109180630-ec00e32a8dfd/go.mod h1:DFci5gLYBciE7Vtevhsrf46CRTquxDuWsQurQQe4oz8=
 google.golang.org/protobuf v0.0.0-20200221191635-4d8936d0db64/go.mod h1:kwYJMbMJ01Woi6D6+Kah6886xMZcty6N08ah7+eCXa0=
 google.golang.org/protobuf v0.0.0-20200228230310-ab0ca4ff8a60/go.mod h1:cfTl7dwQJ+fmap5saPgwCLgHXTUD7jkjRqWcaiX5VyM=
--- a/pkg/gofr/container/container.go
+++ b/pkg/gofr/container/container.go
@@ -45,6 +45,11 @@ import (
 const (
 	redisPubSubModeStreams = "streams"
 	redisPubSubModePubSub  = "pubsub"
+
+	pubSubBackendKafka  = "KAFKA"
+	pubSubBackendGoogle = "GOOGLE"
+	pubSubBackendMQTT   = "MQTT"
+	pubSubBackendRedis  = "REDIS"
 )
 
 // Container is a collection of all common application level concerns. Things like Logger, Connection Pool for Redis
@@ -155,13 +160,13 @@ func (c *Container) Create(conf config.Config) {
 
 func (c *Container) createPubSub(conf config.Config) {
 	switch strings.ToUpper(conf.Get("PUBSUB_BACKEND")) {
-	case "KAFKA":
+	case pubSubBackendKafka:
 		c.createKafkaPubSub(conf)
-	case "GOOGLE":
+	case pubSubBackendGoogle:
 		c.createGooglePubSub(conf)
-	case "MQTT":
+	case pubSubBackendMQTT:
 		c.PubSub = c.createMqttPubSub(conf)
-	case "REDIS":
+	case pubSubBackendRedis:
 		c.createRedisPubSub(conf)
 	}
 }
--- a/pkg/gofr/container/datasources.go
+++ b/pkg/gofr/container/datasources.go
@@ -38,6 +38,7 @@ type Redis interface {
 }
 
 // Cassandra is an interface representing a cassandra database
+//
 // Deprecated: Cassandra interface is deprecated and will be removed in future releases, users must use CassandraWithContext.
 type Cassandra interface {
 	// Deprecated: Query method is deprecated and will be removed in future releases, users must use QueryWithCtx.
--- a/pkg/gofr/container/mock_container.go
+++ b/pkg/gofr/container/mock_container.go
@@ -48,6 +48,7 @@ func WithMockHTTPService(httpServiceNames ...string) options { //nolint:revive /
 		// Create a separate mock instance for each service name
 		// This allows different services to have different expectations
 		serviceMocks := make(map[string]*service.MockHTTP)
+
 		for _, s := range httpServiceNames {
 			mockservice := service.NewMockHTTP(ctrl)
 			c.Services[s] = mockservice
--- a/pkg/gofr/container/sql_mock.go
+++ b/pkg/gofr/container/sql_mock.go
@@ -62,7 +62,7 @@ func (m sqlMockDB) Select(_ context.Context, value any, query string, args ...an
 	expectedArgs := m.queryWithArgs[0].arguments
 
 	valueType := reflect.TypeOf(value)
-	if valueType.Kind() != reflect.Ptr {
+	if valueType.Kind() != reflect.Pointer {
 		m.logger.Errorf("expected a pointer type: %q", value)
 		return
 	}
@@ -74,7 +74,7 @@ func (m sqlMockDB) Select(_ context.Context, value any, query string, args ...an
 
 	v := reflect.ValueOf(value)
 
-	if v.Kind() == reflect.Ptr && !v.IsNil() {
+	if v.Kind() == reflect.Pointer && !v.IsNil() {
 		tobechanged := v.Elem()
 		tobechanged.Set(reflect.ValueOf(m.queryWithArgs[0].value))
 	}
@@ -145,7 +145,7 @@ func (m *mockSQL) ExpectSelect(_ context.Context, value any, query string, args 
 	qr := queryWithArgs{queryText: query, arguments: args}
 
 	fieldType := reflect.TypeOf(value)
-	if fieldType.Kind() == reflect.Ptr {
+	if fieldType.Kind() == reflect.Pointer {
 		qr.value = value
 	}
 
--- a/pkg/gofr/cron_test.go
+++ b/pkg/gofr/cron_test.go
@@ -715,6 +715,7 @@ func TestCronTab_runScheduled_Panic(t *testing.T) {
 				if tc.expectMetricsCalls {
 					c = NewCron(cntnr)
 					defer c.Stop()
+
 					c.jobs = []*job{j}
 				} else {
 					c = &Crontab{
--- a/pkg/gofr/crud_handlers.go
+++ b/pkg/gofr/crud_handlers.go
@@ -70,7 +70,7 @@ func scanEntity(object any) (*entity, error) {
 	}
 
 	objType := reflect.TypeOf(object)
-	if objType.Kind() != reflect.Ptr {
+	if objType.Kind() != reflect.Pointer {
 		return nil, fmt.Errorf("failed to register routes for '%s' struct, %w", objType.Name(), errNonPointerObject)
 	}
 
--- a/pkg/gofr/crud_handlers_test.go
+++ b/pkg/gofr/crud_handlers_test.go
@@ -276,6 +276,8 @@ func Test_CreateHandler(t *testing.T) {
 			if tc.expectedErr != nil {
 				var target *json.UnmarshalTypeError
 				assert.ErrorAs(t, err, &target, "TEST[%d], Failed.\n%s", i, tc.desc)
+			} else {
+				assert.NoError(t, err, "TEST[%d], Failed.\n%s", i, tc.desc)
 			}
 		})
 	}
--- a/pkg/gofr/crud_handlers_test.go
+++ b/pkg/gofr/crud_handlers_test.go
@@ -273,7 +273,10 @@ func Test_CreateHandler(t *testing.T) {
 
 			assert.Equal(t, tc.expectedResp, resp, "TEST[%d], Failed.\n%s", i, tc.desc)
 
-			assert.IsType(t, tc.expectedErr, err, "TEST[%d], Failed.\n%s", i, tc.desc)
+			if tc.expectedErr != nil {
+				var target *json.UnmarshalTypeError
+				assert.ErrorAs(t, err, &target, "TEST[%d], Failed.\n%s", i, tc.desc)
+			}
 		})
 	}
 }
--- a/pkg/gofr/datasource/pubsub/kafka/kafka_test.go
+++ b/pkg/gofr/datasource/pubsub/kafka/kafka_test.go
@@ -923,6 +923,7 @@ func TestKafkaClient_Query_ContextHandling(t *testing.T) {
 				t.Helper()
 				ctx, cancel := context.WithCancel(t.Context())
 				cancel()
+
 				return ctx, func() {}
 			},
 			description: "Should handle canceled context",
@@ -987,6 +988,7 @@ func TestKafkaClient_Subscribe_RaceDetector(t *testing.T) {
 		func(_ context.Context) (kafka.Message, error) {
 			// Add small delay to increase chance of race
 			time.Sleep(time.Microsecond)
+
 			return kafka.Message{
 				Topic: "race-test-topic",
 				Value: []byte("test"),
--- a/pkg/gofr/datasource/pubsub/kafka/tls_test.go
+++ b/pkg/gofr/datasource/pubsub/kafka/tls_test.go
@@ -84,7 +84,6 @@ IFi4oUxW0wLNUmTJFSIrFRE3eYWy56XiI8jPs7U94It8YwjhDSeHwslMKbGwogqI
 Om59HA==
 -----END CERTIFICATE-----`
 
-	//nolint:gosec // This is a test certificate
 	clientKeyContent := `-----BEGIN RSA PRIVATE KEY-----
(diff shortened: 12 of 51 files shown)


aryanmehrotra (changes requested):
## Review — verified at `714099f7`

I reproduced this locally: built the module, ran `go vet ./pkg/gofr/...`, ran the touched packages' tests, and ran `golangci-lint` on `./pkg/gofr/...` for both `development` and this branch with **separate caches** and `--max-issues-per-linter=0 --max-same-issues=0` (the defaults cap at 50/3 and make any before/after comparison meaningless).

**The headline claim holds.** Zero new lint issues are introduced — `comm` over the two full issue sets is empty in the "only on this branch" direction. Totals across `./pkg/gofr/...`:

| linter | development | this PR |
|---|---|---|
| goconst | 172 | 126 |
| noctx | 94 | 91 |
| wsl_v5 | 56 | **0** |
| gocritic | 31 | 28 |
| gosec | 30 | 23 |
| staticcheck | 18 | 16 |
| govet | 12 | **0** |
| prealloc | 5 | **0** |
| testifylint | 4 | 1 |
| gocyclo | 3 | 3 |
| exhaustive | 1 | **0** |
| **total** | **435** | **288** |

I also verified each suppression is load-bearing by deleting it and re-linting — all six are real (G101 ×3 in `supabase_test.go`, G115 ×3 in `websocket_test.go`, G101 in `oauth.go`/`apikey_auth.go`, G703 in `sql.go`, gocyclo in `exporter.go`). And I checked the riskiest hunk from the OTel source rather than trusting the deprecation notice: `attribute.INVALID` is a literal alias for `EMPTY` (`value.go:60`), and `Value.Emit()` and `Value.String()` are byte-identical for `BYTESLICE` and `SLICE` in v1.44.0. `exporter.go` is behaviour-preserving.

The two deferred gosec bugs are real and still present (`auth.go:35` G602, `kafka/helper.go:297` G118) — agreed on splitting those out.

Two things need fixing before merge, and one config change would make this whole class of PR unnecessary.

---

### 1. `Params()` is a behavioural change — `pkg/gofr/http/request.go:98`

```go
-var result []string
+result := make([]string, 0, len(values))
```

For an absent query key this returns `[]string{}` where it used to return `nil`. Verified by running it: on `development` `json.Marshal` gives `null`, on this branch `[]`. `Params` is public API — reachable as `gofr.Context.Params(key)` — so any handler doing `return c.Params("tag"), nil` changes its response body from `{"data":null}` to `{"data":[]}`.

**I think the new behaviour is correct and should stay.** The godoc directly above it already promises "If the parameter is not present, an empty slice is returned", and `pkg/gofr/cmd/request.go:91` already returns `[]string{}`. This makes the HTTP implementation honour its own contract and match its sibling.

But it isn't "no behavioural impact", and it should not land silently:
- move it out of the "no behavioral impact" section of the PR body into an explicit note, so it makes the release notes;
- lock it in — `Test_Params` uses `assert.Empty`, which passes for both `nil` and `[]`, so nothing currently guards it. Add `assert.NotNil(t, r.Params("nonexistent"))`.

### 2. Test coverage regression — `pkg/gofr/crud_handlers_test.go:276`

```go
-assert.IsType(t, tc.expectedErr, err, ...)
+if tc.expectedErr != nil {
+    var target *json.UnmarshalTypeError
+    assert.ErrorAs(t, err, &target, ...)
+}
```

`assert.IsType(t, nil, err)` was not a no-op for the nil cases. testify's `ObjectsAreEqual(reflect.TypeOf(object), reflect.TypeOf(expectedType))` short-circuits on `nil` and compares equality, so with a nil expectation it asserts `err == nil`. I confirmed this: `assert.IsType(mockT, error(nil), errors.New("boom"))` returns `false`.

Two of the three table cases have `expectedErr: nil`. After this change they assert nothing at all about `err` — `Create` could return the right response *and* an error and the test would still pass.

Secondary issue: the target type is hardcoded to `*json.UnmarshalTypeError` while the table field is still declared `expectedErr error`, so the table no longer drives the assertion.

The fix already exists in this PR — `oauth_test.go` got the `err` / `errAs` split, and that version is genuinely better than what it replaced. Apply the same shape here:

```go
switch {
case tc.errAs != nil:
    require.ErrorAs(t, err, tc.errAs, ...)
case tc.expectedErr != nil:
    require.ErrorIs(t, err, tc.expectedErr, ...)
default:
    require.NoError(t, err, ...)
}
```

### 3. The lever is `.golangci.yml`, not 46 files

This is the part I'd most like to see land with the PR. The backlog is largely manufactured by config, and the cleanup is fighting it file-by-file.

**`goconst` is configured at `min-len: 2, min-occurrences: 2`** — golangci's defaults are `3` and `3`. That threshold is what produces `trueValue = "true"`, `dbSegment = "db"`, `localhost = "localhost"`, `payloadKey = "payload"`: constants whose name carries no more information than the literal, and which force the reader to jump to a `const` block to learn nothing. Of the 126 goconst issues left, **71 (56%) are two-occurrence hits** that disappear at `min-occurrences: 3`.

**20 of the 23 remaining gosec issues are G101 false positives in `_test.go`** (`oauth_test.go` ×10, `basic_auth_test.go` ×7, `supabase_test.go` ×3). The PR silences three of them with inline `//nolint` and leaves twenty. `_test.go` already has a `dupl`/`goconst`/`mnd` exclusion — G101 belongs there too:

```yaml
- linters: [gosec]
  text: "G101"
  path: _test\.go
```

That removes 20 issues *and* the three new `//nolint` comments, and stops the next contributor rediscovering them.

### 4. `err`-prefixed constants are strings, not errors — `pkg/gofr/service`

The package now has both, under one prefix:

```go
// errors.go, oauth.go, basic_auth.go, apikey_auth.go, health.go  — string
const errUnknown = "unknown error"
errClientIDMandatory = "client id is mandatory"

// file_token_auth.go, connection_pool.go, rate_limiter_config.go — error
errEmptyTokenFile   = errors.New("token file is empty")
errInvalidRequestRate = errors.New("requests must be greater than 0 per configured time window")
```

`errFoo` is the Go convention for a sentinel `error`; revive's `error-naming` rule (enabled here) only checks the reverse direction, so nothing will catch this. In a framework package that's a genuine trap. Rename to `msgClientIDMandatory`, `msgUnknown`, etc.

(Longer term these are `AuthErr.Message` payloads that callers can't `errors.Is` against — real sentinels would be better, but that's a separate PR.)

### 5. Half-extracted literal sets

goconst fires at 2 occurrences, so the extraction lands on whichever members happen to repeat — leaving lists that are half constant, half literal, which reads worse than either extreme:

- `http/middleware/config.go:36-41` — 3 of the 6 `ACCESS_CONTROL_*` env names are constants, 3 stay literals in the same slice.
- `logging/remotelogger/dynamic_level_logger.go:272` — `[]string{levelDebug, "INFO", "NOTICE", "WARN", levelError, "FATAL"}`.
- `container/container.go` — constants for the four PubSub backends, but `"true"` at line 524 (4 occurrences, still flagged) stays a literal.

Either extract each set completely or leave it alone; the half-state is the worst of the three.

### 6. `exporter.go` suppresses a violation it also created

Confirmed the `//nolint:gocyclo` is necessary — but only because of this PR. On `development` `attributeToStringPair` sits at complexity 10 and is not flagged; adding `case attribute.BYTESLICE, attribute.SLICE:` for `exhaustive` pushes it to 11. Deleting the directive and re-linting reproduces `cyclomatic complexity 11 of func attributeToStringPair is high (> 10)`.

Given the PR explicitly defers gocyclo refactors, adding a new gocyclo suppression sits oddly. The four JSON-slice cases are mechanically identical and lift out cleanly:

```go
case attribute.BOOLSLICE, attribute.INT64SLICE, attribute.FLOAT64SLICE, attribute.STRINGSLICE:
    return string(kv.Key), marshalSlice(kv.Value)
```

That drops it back under the threshold and removes the directive.

### Nits

- `migration/opentsdb_test.go:97,1000` — `make([]struct{...}, 0, 4)` / `0, 3`: the caps count `append` *calls*, not elements (each appends a whole slice), so the hint is inert. Either compute the real total or keep `var cases []T` with a `//nolint:prealloc`.
- `http/request_test.go:23,30,132` — `context.Background()` where the repo runs 56:16 in favour of `t.Context()`. It matches the rest of that file, so low priority, but `t.Context()` is the better default.
- `datasource/sql/supabase.go:123` — `supaConfig.Database = dialectPostgres` uses a *dialect* constant as a *database name*. It matches existing `sql.go:108`, so this PR isn't making it worse; flagging only so it doesn't spread further.

---

### Checklist

- [ ] Move `Params()` nil→`[]string{}` out of "no behavioral impact"; add `assert.NotNil` in `Test_Params`
- [ ] Restore the no-error assertion in `Test_CreateHandler` (use the `errAs` table pattern from `oauth_test.go`)
- [ ] `.golangci.yml`: `goconst.min-occurrences: 3`, `min-len: 3`; add the `gosec`/`G101`/`_test\.go` exclusion
- [ ] Rename the `err*` string constants in `pkg/gofr/service` to `msg*`
- [ ] Complete or revert the partial extractions (middleware config, remotelogger levels, container `"true"`)
- [ ] Extract the slice cases in `attributeToStringPair` instead of `//nolint:gocyclo`
- [ ] Nits: opentsdb_test prealloc caps, `t.Context()` in request_test

Notes on what I could not verify: `pkg/gofr` tests fail on my machine with `FATAL http port 8000 is blocked or unreachable` — that's my own process holding :8000, not this branch. Everything else in the touched set passes.

Umang01-hash:
Thanks for the exceptionally thorough review — you were right on all of it. Pushed `e2e5527f`. Checklist:

- [x] **`Params()` behavioural change** — kept the `[]string{}` behaviour (it honours the godoc and matches `cmd`), moved it into an explicit **behavioural-change** section of the PR body for the release notes, and added `assert.NotNil(t, r.Params("nonexistent"))` to `Test_Params` to lock it.
- [x] **`Test_CreateHandler`** — restored the no-error assertion: `else { assert.NoError(...) }` for the `nil` cases. Good catch that `assert.IsType(nil, err)` was asserting `err == nil` and the guarded `ErrorAs` had dropped it.
- [x] **`.golangci.yml`** — `goconst` `min-len`/`min-occurrences` → **3/3**; added the `gosec`/`G101`/`_test\.go` exclusion. This is the real fix, agreed.
- [x] **`err*` → `msg*`** in `pkg/gofr/service`.
- [x] **Partial sets** — reverted the two-occurrence noise to literals (middleware CORS env keys, remotelogger levels, and the rest); kept only the genuinely-repeated constants.
- [x] **`exporter.go`** — rather than extract a helper (which would just move an `exhaustive` requirement onto it), I **dropped the BYTESLICE/SLICE case** so there's no new `gocyclo` suppression at all — only the behaviour-preserving deprecation fixes remain. The pre-existing `exhaustive` on the switch stays out of scope (untouched line), consistent with deferring the complexity work.
- [x] **Nits** — `opentsdb_test` prealloc reverted + `//nolint:prealloc` (caps counted append calls); `request_test` → `t.Context()`; removed the now-unused test `//nolint:gosec` directives the config exclusion makes redundant.

The two real gosec bugs (`auth.go` G602, `kafka/helper.go` G118) I'll send as a separate PR with regression tests, as agreed. CI is green.

Re the OTel double-check and verifying each suppression by deleting it — that's exactly the kind of verification I should have done up front; appreciated.

aryanmehrotra (changes requested):
## Re-review — verified at `e2e5527f`

Re-ran the same protocol: build, `go vet ./pkg/gofr/...`, tests on every touched package, and `golangci-lint` on `./pkg/gofr/...` with separate caches and `--max-issues-per-linter=0 --max-same-issues=0`. Because this revision changes `.golangci.yml`, I also linted **unmodified `development` with this PR's config** so the code delta and the config delta can be told apart.

Everything on the checklist is done, and the config lever did what it was supposed to:

| | development (today's config) | development (this PR's config) | this PR |
|---|---|---|---|
| goconst | 172 | 77 | 52 |
| noctx | 94 | 94 | 91 |
| wsl_v5 | 56 | 56 | 0 |
| gocritic | 31 | 31 | 28 |
| gosec | 30 | 7 | 3 |
| staticcheck | 18 | 18 | 16 |
| govet | 12 | 12 | 0 |
| prealloc | 5 | 5 | 0 |
| testifylint | 4 | 4 | 1 |
| nolintlint | 0 | **2** | **2** |
| **total** | **435** | **310** | **197** |

Holding config constant, the code delta introduces **zero** new issues — `comm` over the two full sets is empty. Build and vet clean; all touched packages pass, including the `Test_CreateHandler` and exporter paths (my earlier `pkg/gofr` failure was my own process on :8000, not this branch).

Both must-fixes are genuinely fixed. `Test_CreateHandler` now asserts `NoError` on the nil branch. `Test_Params` has the `assert.NotNil` guard, and the behavioural change is called out in its own section of the PR body — that's the right place for it.

Two things to settle before this goes in, one of which is me correcting my own advice.

### 1. The G101 exclusion strands two `//nolint` directives — `nolintlint` fails

`nolintlint` runs with `allow-unused: false`, so the new `_test.go` G101 exclusion turns two pre-existing directives into reported issues:

```
pkg/gofr/datasource/pubsub/kafka/tls_test.go:87:2: directive `//nolint:gosec // This is a test certificate` is unused for linter "gosec" (nolintlint)
pkg/gofr/datasource/redis/redis_test.go:201:2: directive `//nolint:gosec // dummy private key for test only, not used in production` is unused for linter "gosec" (nolintlint)
```

Both suppress exactly the G101 that the exclusion now handles. I confirmed the cause is the config and not the code by applying only the new `.golangci.yml` to unmodified `development` — same two issues, nothing else moves.

**CI being green here is the problem, not the reassurance.** `only-new-issues` scopes to the diff, and neither file is in it, so these land on `development` as a fresh backlog entry that nothing will flag until someone next touches those files. Two lines to delete.

### 2. Revert `goconst.min-len` to `2` — my recommendation was half wrong

I asked for `min-len` and `min-occurrences` both at 3. `min-occurrences: 3` is doing all the work; `min-len: 3` is not carrying its weight, and I should have checked before asking for it.

Measured on this branch, holding `min-occurrences: 3`: `min-len: 2` → 57 issues, `min-len: 3` → 52. Five issues, and the entire delta is two-character strings dominated by `"UP"`.

That's the wrong five to lose. The repo already treats these as constant-worthy — `pkg/gofr/health.go:11` has `statusUp = "UP"`, `pkg/gofr/datasource/health.go:4` has `StatusUp = "UP"` — while clickhouse, badger and the nats KV store still hardcode `h.Status = "UP"`. `min-len: 3` also leaves an incoherent policy behind: `"DOWN"` is four characters and stays flagged, `"UP"` is two and goes silent, for the same pair of status values on the same line.

Keep `min-occurrences: 3`; put `min-len` back to `2`. That costs five issues on the backlog and keeps the linter pointed at a real inconsistency.

### Nit — the revert re-created two small half-and-half sets

Reverting to exactly what goconst demands is a defensible rule, but at `min-occurrences: 3` it lands back in the state I flagged last round, just in different files:

- `service/oauth.go` — five of the seven messages are `msg*` constants; `"client id is mandatory"` and `"invalid scheme, allowed http and https only"` sit as literals in the same block and the same `switch`.
- `service/basic_auth.go` — `msgUsernameRequired` is a constant, the two adjacent messages next to it are literals.
- `datasource/sql/supabase.go` — `connTypeDirect` is a constant, `"session"` and `"transaction"` are literals in the same `switch`.

Nothing stops a two-occurrence string from being a constant; goconst only complains in the other direction. Promoting the four stragglers costs nothing and makes each block read uniformly. Non-blocking — say the word and I'll stop bringing it up.

### On `exporter.go`

Dropping the BYTESLICE/SLICE case rather than extracting a helper is the right call, and your reasoning is correct: a helper containing a `switch` on `attribute.Type` inherits the same `exhaustive` obligation, so it moves the problem instead of solving it. I checked the outcome — `exhaustive` at `exporter.go:142` is byte-identical to the one already on `development`, so nothing regresses, and behaviour is unchanged because the `default:` branch returns `kv.Value.String()`, which I'd already verified equals `Emit()` for both those types. The PR body describes this accurately.

---

### Checklist

- [ ] Delete the two now-unused `//nolint:gosec` directives (`kafka/tls_test.go:87`, `redis/redis_test.go:201`)
- [ ] `.golangci.yml`: `goconst.min-len` back to `2`, keep `min-occurrences: 3`
- [ ] Optional: promote the four straggler literals in `oauth.go` / `basic_auth.go` / `supabase.go` so each set is uniform

With the first two done this is good to merge. Nice work on the turnaround — the PR body's behavioural-change section is exactly the right shape for the release notes.

Umang01-hash:
Both must-fixes done + the nit, in `5ced268e`:

- [x] **Deleted the two stranded `//nolint:gosec`** (`kafka/tls_test.go:87`, `redis/redis_test.go:201`). Good point that `only-new-issues` being green *hides* this — they'd have quietly landed on `development` as fresh `nolintlint` backlog.
- [x] **`goconst.min-len` back to `2`** (kept `min-occurrences: 3`). Agreed — silencing `"UP"` while `"DOWN"` stays flagged for the same status pair is exactly the incoherent policy worth avoiding, and the repo already has `statusUp = "UP"`.
- [x] **Nit — uniform blocks**: promoted the stragglers so no half-and-half remains — `msgClientIDMandatory`/`msgInvalidScheme` (oauth), `msgPasswordRequired`/`msgPasswordNotBase64` (basic_auth), `connTypeSession`/`connTypeTransaction` (supabase).

CI green. Thanks again for catching your own `min-len` call and for the config-vs-code delta table — that separation made it easy to verify. The two real gosec bugs go in a separate PR next.

aryanmehrotra (approved):
## Re-review — verified at `5ced268e`. LGTM 🚀

Both must-fixes and the nit are in, and I re-ran the full protocol against this commit: build, `go vet ./pkg/gofr/...`, tests across every touched package, and `golangci-lint` on `./pkg/gofr/...` with separate caches and `--max-issues-per-linter=0 --max-same-issues=0` — plus unmodified `development` carrying this PR's **final** config, so the code delta and the config delta stay separable.

| | `development` (its config) | `development` (this PR's config) | this PR |
|---|---|---|---|
| goconst | 172 | 82 | 57 |
| noctx | 94 | 94 | 91 |
| wsl_v5 | 56 | 56 | 0 |
| gocritic | 31 | 31 | 28 |
| staticcheck | 18 | 18 | 16 |
| govet | 12 | 12 | 0 |
| gosec | 30 | 7 | 3 |
| prealloc | 5 | 5 | 0 |
| testifylint | 4 | 4 | 1 |
| nolintlint | 0 | 2 | **0** |
| **total** | **435** | **315** | **200** |

Holding config constant, the code delta introduces **zero** new issues — `comm` over the two full sets is empty in that direction. Build and vet clean, every touched package passes, and all 18 CI checks are green.

Point by point:

**Stranded directives — fixed.** `nolintlint` goes 2 → 0. The 2 in the middle column is the proof the deletions were load-bearing: `development` still carries those directives, so with this config applied it reports them; this branch removes them and reports none. `development` won't inherit a silent backlog entry.

**`min-len` back to `2` — confirmed.** goconst moves 52 → 57, and the five that came back are the `"UP"` occurrences, which is the point: the linter stays pointed at the clickhouse / badger / nats KV packages that still hardcode `h.Status = "UP"` while `pkg/gofr/health.go` and `pkg/gofr/datasource/health.go` already define the constant. Thanks for taking the correction without churn.

**Uniform const blocks — good.** `oauth.go` now holds all seven messages, `basic_auth.go` all three, `supabase.go` all three connection types; no call site mixes a constant and a literal for the same kind of value. Every string is byte-identical to what it replaced, so the compiler is the proof of equivalence.

**What's left is deliberate and correctly scoped.** The 3 remaining gosec are the two real bugs you're splitting out (`auth.go:35` G602, `kafka/helper.go:297` G118) plus a G115 in an mqtt test; `exhaustive` at `exporter.go:142` is byte-identical to the one already on `development`; `noctx` at 91 and the pubsub v2 `staticcheck` deprecations are explicitly deferred. Nothing here regresses against `development` on any linter.

Net: 435 → 200 across `./pkg/gofr/...`, one clearly-documented behavioural change with a test guarding it, and the config lever fixed at the source so the next contributor doesn't re-derive any of this.

---

### `-race`: clean for this PR, but it surfaced two pre-existing problems

I ran `go test -race` over the touched packages on this branch and on `development`. **Nothing here is caused by this PR** — but the run is worth reporting, because CI never does it (`grep -rn race .github/workflows/` returns nothing), so none of this is visible today.

**1. `pkg/gofr/datasource/redis` cannot be run under `-race`.** 28 races on this branch, 30 on `development`, and ~16 tests fail on both. Every single race — I checked, there are zero exceptions — sits in `go-redis/redismock@v9.2.0`, whose `(*mock).pushExpect` and `(*mock).process` touch the same expectation slice without synchronisation. The test goroutine calls `ExpectPing()` while `(*PubSub).isConnected()` pings from a background goroutine (`pubsub.go:1174`):

```
Write at 0x00c00018cab8 by goroutine 18:   redismock.(*mock).pushExpect()  <- health_test.go:24
Previous read at 0x00c00018cab8 by goroutine 25: redismock.(*mock).process() <- pubsub.go:1174 isConnected
```

That's a mock-library limitation, not GoFr code, and it needs the tests to stop racing the background pinger rather than a fix in `pubsub.go`.

**2. `Test_SQLRetryConnectionInfoLog` is timing-flaky under `-race`** — the gomock gauge expectation gets `app_sql_open_connections = 1` where it wants `0`, depending on whether the metrics ticker fires before the assertion. Failure rate over 16 runs each: `development` 4/16, this branch 6/16. That gap is noise at this sample size, and the PR touches nothing on that path (only string constants in the same file), so I'm reading it as one pre-existing flake, not a regression.

Both deserve their own issues; neither belongs in this PR.

---

Approving. Nice iteration — three rounds, no thrash, and the PR body ended up a better artefact than the diff.

PR #3924, opened 2026-08-12
fix(http): keep Params returning nil for an absent key

aryanmehrotra:
## What

Restores `Request.Params(key)` to returning a **nil** slice for an absent query key, undoing the one behavioural change that arrived in #3921.

## Why

#3921 was a mechanical lint cleanup. One of its three `prealloc` hits was here:

```
pkg/gofr/http/request.go:96:6: Consider preallocating result (prealloc)
	var result []string
```

The fix was `var result []string` → `make([]string, 0, len(values))`. For an absent key the loop never runs, so the return went from nil to empty — and a handler written as

```go
app.GET("/search", func(c *gofr.Context) (any, error) { return c.Params("tag"), nil })
```

started responding `{"data":[]}` where it previously responded `{"data":null}`.

`prealloc` is a performance hint; nothing about it required changing what the function returns. The other two `prealloc` hits in that same PR (`opentsdb_test.go:97` and `:989`) were resolved with `//nolint:prealloc` precisely because preallocating was wrong there. This applies the same treatment to the one case that touches a public API's response body.

Only nil has ever shipped — v1.58.0 returns nil, and the empty-slice behaviour exists only on `development`.

## Changes

- `Params` keeps `var result []string`, with a comment explaining why the allocation hint is declined and a `//nolint:prealloc` carrying that reason. The hint is also a poor estimate on its own terms: `len(values)` under-counts as soon as any value contains a comma.
- **Godoc corrected.** It claimed "If the parameter is not present, an empty slice is returned", which was never true before #3921 and is the mismatch that made the change look harmless. It now describes the actual behaviour, and documents comma-splitting — which `cmd`'s `Params` and the docs site both mention, but the godoc users see on pkg.go.dev did not.
- `Test_Params` pins the contract twice: `assert.Nil` on the returned value, and the response body through the real `Responder`. The previous `assert.Empty` passed for both nil and empty, which is why the original change slipped through unnoticed.

## Release-branch action required before #3923 merges

This PR takes the change off `development` only. **`release/v1.59.0` was cut before this and still carries the empty-slice behaviour**, plus a "⚠️ Behavioural change" section in #3923's notes announcing it. Without action there, v1.59.0 ships the regression *and* documents it, and v1.60.0 then ships the un-announcement — a behaviour flip-flop across one minor version, which is worse for users than either endpoint.

The release branch needs these commits cherry-picked (or a re-cut from `development` once this lands), and that section removed from #3923.

## Not closing the door

Returning an empty slice may well be the better contract — `null` breaks naive JavaScript clients (`data.map`, `data.length`) where `[]` does not. But it is a bigger decision than it looks: of the seven `gofr.Request` implementations, six return nil for an absent key, four of them with existing `assert.Nil` tests, and `pkg/gofr/cmd/request.go:90` is the lone outlier returning `[]string{}`. Adopting `[]` means changing six implementations and their tests, with a release note — not a one-line alignment. That deserves its own PR, not a silent arrival inside a lint cleanup.

## Verification

`go build ./...`, `go vet`, and `go test ./pkg/gofr/http/...` pass. `golangci-lint` reports zero issues in `request.go`; `nolintlint` runs with `allow-unused: false` and does not flag the new directive, confirming `prealloc` genuinely fires there. Both test pins were verified by reintroducing the preallocation and watching them fail.

### DIFF
--- a/pkg/gofr/http/request.go
+++ b/pkg/gofr/http/request.go
@@ -90,14 +90,17 @@ func (r *Request) HostName() string {
 	return fmt.Sprintf("%s://%s", proto, r.req.Host)
 }
 
-// Params returns a slice of strings containing the values associated with the given query parameter key.
-// If the parameter is not present, a nil slice is returned.
+// Params returns the values associated with the given query parameter key. Each value is
+// additionally split on commas, so ?tag=a,b&tag=c yields []string{"a", "b", "c"}.
+//
+// If the key is absent, a nil slice is returned. A handler may return this value directly, and a
+// nil slice marshals to `null` in the response body where an empty one marshals to `[]`.
 func (r *Request) Params(key string) []string {
 	values := r.req.URL.Query()[key]
 
-	// Deliberately not preallocated. A handler may return this slice directly, and a nil slice
-	// marshals to `null` where an empty one marshals to `[]`. Preallocating would silently change
-	// that response body for every absent key, so the allocation hint is declined here.
+	// Deliberately not preallocated: that would return an empty slice for an absent key and
+	// silently change the response body from `null` to `[]`. The hint is also a poor estimate,
+	// since len(values) under-counts as soon as any value contains a comma.
 	//nolint:prealloc // preserves the nil return for an absent key; see comment above
 	var result []string
 
--- a/pkg/gofr/http/request.go
+++ b/pkg/gofr/http/request.go
@@ -94,7 +94,7 @@ func (r *Request) HostName() string {
 // additionally split on commas, so ?tag=a,b&tag=c yields []string{"a", "b", "c"}.
 //
 // If the key is absent, a nil slice is returned. A handler may return this value directly, and a
-// nil slice marshals to `null` in the response body where an empty one marshals to `[]`.
+// nil slice reaches the client as "null" in the response body where an empty one reaches it as "[]".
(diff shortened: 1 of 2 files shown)


Umang01-hash (approved):
Approving — this is the correct unwind of the behavioural change #3921 slipped in (my fault for mislabeling it).

Reviewed the diff, not just the description:
- The `prealloc` hint was never valid here anyway — `result` is built via `strings.Split(value, ",")`, so `len(values)` under-counts and `make(..., len(values))` reallocates regardless. Declining it with `//nolint:prealloc` is technically correct, not just behaviour-safe.
- Godoc now matches reality (it previously claimed "empty slice" while the code returned nil — that false doc is what made the change look harmless). Documenting comma-splitting + the null-vs-`[]` wire implication is a real improvement.
- Test hardening is the highlight: `assert.Nil` + `JSONEq` through the real `Responder` pins the actual wire body, where the old `assert.Empty` passed for both nil and `[]` and hid the change.
- Correctly leaves the `[]`-contract decision to a separate deliberate PR (it'd need all Params impls + their tests + a note).

Confirmed the release side is already consistent: `release/v1.59.0` already returns nil and #3923's notes no longer announce the change, so v1.59.0 ships nil with no flip-flop. CI green.

PR #3925, opened 2026-08-13
fix(otel): flush buffered trace spans on shutdown

### LINKED ISSUE #3771: Buffered trace spans dropped on shutdown: BatchSpanProcessor never shut down
### What

### Where

```go
```

### Notes

om7057:
## Summary

Fixes #3771.

`App.Shutdown` (HTTP/gRPC path) and the CMD flush path in `Run()` already flush buffered metrics via `Container.ShutdownMetrics`, but neither ever called `Shutdown` on the trace `TracerProvider`. `initTracer` builds a `TracerProvider` with a `BatchSpanProcessor`, but the `tp` handle was local to that function and discarded, so spans still sitting in the batch processor's buffer at process exit were silently dropped instead of flushed.

This follows the fix approach discussed on #3761's review thread: a single `App.telemetryShutdown` registry of `func(ctx) error`, drained once by a new `drainTelemetry` helper from both `Shutdown` and `Run`'s CMD path. Both the tracer's `tp.Shutdown` (registered in `initTracer`, on both the no-exporter/`NeverSample` path and the real SDK-exporter path) and the existing `Container.ShutdownMetrics` register into it, replacing the direct `ShutdownMetrics` calls. No new OTel SDK type leaks into `Container` or `App`: the registry only holds the same `func(ctx) error` shape `Container.shutdownMetrics` already used.

Also renamed `metricsFlushTimeout` to `telemetryFlushTimeout` in `run.go` since it now bounds both metrics and trace flush.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/gofr/...`
- [x] New tests:
  - `Test_initTracer_RegistersShutdown_NoExporter` / `Test_initTracer_RegistersShutdown_WithExporter`: assert `initTracer` registers exactly one shutdown func on each path
  - `Test_Shutdown_DrainsTelemetry`: a fake `func(ctx) error` registered into `telemetryShutdown` is invoked exactly once by `App.Shutdown`, and its error surfaces in the joined result
  - `Test_BatchSpanProcessor_FlushesOnShutdown`: a focused, network-free unit test proving the underlying mechanism (a span still buffered in a `BatchSpanProcessor`, not yet exported by its background timer, is flushed to the exporter on `tp.Shutdown`), using a minimal recording exporter instead of `tracetest.InMemoryExporter`, whose own `Shutdown` clears its recorded spans and would mask the flush being tested
- [x] Existing `TestNewCMD_ShutdownMetricsCalledAfterRun` unaffected by the constant rename


om7057:
@Umang01-hash @coolwednesday, please review and suggest changes.

aryanmehrotra (changes requested):
The diagnosis in the description is right, the registry is the right shape, and `Test_BatchSpanProcessor_FlushesOnShutdown` — with its note on why `tracetest.InMemoryExporter` would have masked the thing under test — is the kind of test that saves the next person an hour.

I did not review this from the diff. I built both trees, ran real GoFr processes against a real span receiver, and observed them from outside with `SIGTERM` rather than by calling `Shutdown` from a test. Four things came out of that, and the first one is why I am requesting changes rather than approving.

## 1. The fix does not reach the path #3771 is about

A child process, one request, one span still inside the batch window, then a real `SIGTERM` — the Kubernetes path:

| | CMD app | HTTP app, `SIGTERM` |
|---|---|---|
| `development` | 0 spans | 0 spans |
| this PR | **1 span** | **0 spans** |

Five runs, `0` every time. The CMD half of the fix works; the HTTP half does not run.

The control rules out my harness: hold the process open for 7s so the `BatchSpanProcessor`'s own timer fires, and the span arrives on both trees. The span exists and the receiver works — it is the flush that never happens.

The mechanism is in `run.go`, and it predates this PR. `startShutdownHandler` launches the shutdown goroutine and nothing ever joins it; `startAllServers` returns as soon as `wg.Wait()` sees the servers stop, `Run` returns, `main` returns, and the process exits out from under `Shutdown`.

Measured rather than argued: an in-process `App.Shutdown` against a **hung** collector takes the full 30s grace period (see finding 2). The same binary, same hung collector, killed by `SIGTERM`, exits in **353 ms** — so `drainTelemetry` is not merely slow on that path, it is not completing at all.

The missing piece is small. Have the handler close a channel when it is done and wait for it after the servers stop:

```go
	shutdownDone := make(chan struct{})
	...
	a.startShutdownHandler(ctx, timeout, shutdownDone)
	a.startTelemetryIfEnabled()
	a.startAllServers(ctx)

	// The servers have stopped; wait for the shutdown goroutine to finish draining
	// telemetry, or the process exits out from under the flush.
	select {
	case <-shutdownDone:
	case <-time.After(timeout):
	}
```

with `defer close(drained)` as the first statement of the goroutine in `startShutdownHandler`. I applied exactly that and re-ran the harness: **`SIGTERM` now delivers the span, 3 runs out of 3**, and a healthy shutdown still exits in ~270 ms.

I would rather see that in this PR than filed separately, because without it the PR's own summary — "spans still sitting in the batch processor's buffer at process exit were silently dropped instead of flushed" — remains true for every HTTP service in Kubernetes, which is most of them.

## 2. Once that lands, the flush must be bounded — it is not today

Joining the goroutine is what makes the next problem real. With a collector that accepts the TCP connection and never answers — a wedged sidecar, an LB holding the socket — `drainTelemetry` inherits the caller's context, which on the `SIGTERM` path is the whole `SHUTDOWN_GRACE_PERIOD`:

| collector state | `development` | this PR | this PR + the join above |
|---|---|---|---|
| healthy | ~260 ms | ~300 ms | ~280 ms |
| **hung** | 294 ms | 353 ms | **30.002 s** |
| refused (nothing listening) | 0 s | 1 ms | 1 ms |

30.002 s against a default `terminationGracePeriodSeconds` of 30 s — which `docs/guides/graceful-shutdown/page.md` documents on the same page — means the pod is `SIGKILL`ed at exactly the moment GoFr finally gives up. Across a rolling deploy that is a worse outcome than the dropped spans this PR set out to fix.

The CMD path already solved this with `telemetryFlushTimeout`. Giving `drainTelemetry` the same bound applies it everywhere:

```go
func (a *App) drainTelemetry(ctx context.Context) error {
	ctx, cancel := context.WithTimeout(ctx, telemetryFlushTimeout)
	defer cancel()
	...
```

Verified with both changes applied: `SIGTERM` still delivers the span (3/3), the CMD case still delivers it, and the hung collector is bounded at **10.3 s** instead of 30 s. The explicit `context.WithTimeout` in `Run`'s CMD path then becomes redundant but harmless.

Worth knowing either way: `App.Shutdown` now returns `context deadline exceeded` where it previously returned `nil`, for anyone who checks its error. That is defensible — it *is* a failure — but it is a behaviour change on an exported method and belongs in the description.

## 3. `Test_Shutdown_DrainsTelemetry` fails about a third of the time

```
single run, as CI runs it:   14/40 failed
    Messages: expected the registered telemetry shutdown func to be invoked exactly once
```

`go test -v ./...` in `go.yml:227` runs each test once, so this is roughly one red CI run in three, on a test added by this PR. It also trips the race detector — `-race` is green on `development` and fails here:

```
WARNING: DATA RACE
Read at ... gofr_test.go:1446        (calls++)
Previous write at ... gofr_test.go:1446
```

Both have the same cause: `go g.Run()` starts GoFr's own signal-driven shutdown handler, and the test then calls `g.Shutdown` itself, so `drainTelemetry` can run twice — concurrently, on a plain `int`. `calls == 1` is not a property the test can rely on while `Run` is racing it.

The test does not need a running server to prove that `Shutdown` drains the registry. Dropping the `go g.Run()` and the `time.Sleep(10 * time.Millisecond)` makes it deterministic; if the running server matters, `calls` needs to be an `atomic.Int64` and the assertion `>= 1`.

This also says something about the production code that is worth a sentence somewhere: `Shutdown` is exported, documented in the FAQ, and reachable concurrently with the signal handler, so `drainTelemetry` can run twice. `tp.Shutdown` is idempotent per the OTel SDK and I saw no panic across repeated double-shutdowns, so this is a cost rather than a bug — but it is a second full flush.

## 4. One new lint issue, which CI will report

```
pkg/gofr/cron.go: string `gofr` has 3 occurrences, but such constant `gofrTraceExporter` already exists (goconst)
```

`development` reports 14 issues in `pkg/gofr`, this branch 15, and that is the one. The PR does not touch `cron.go` — `goconst` counts occurrences across the package including test files, and the new `"TRACE_EXPORTER": "gofr"` in `Test_initTracer_RegistersShutdown_WithExporter` is the third, so it reports at the non-test site. The lint job runs with `only-new-issues: true`, so it will be flagged.

`gofrTraceExporter` is the constant `goconst` is pointing at, and it is already in this package:

```go
"TRACE_EXPORTER": gofrTraceExporter,
```

Verified: the `cron.go` report disappears and `Test_initTracer_RegistersShutdown_WithExporter` still passes.

## 5. Test gaps — three of seven mutations survive

| mutation | |
|---|---|
| no-exporter path: drop the registration | killed |
| SDK path: drop the registration | killed |
| `Shutdown`: drop `drainTelemetry` | killed |
| `drainTelemetry`: swallow the errors | killed |
| **`Run`/CMD: drop `drainTelemetry`** | **survived** |
| **`New()`: drop the metrics registration** | **survived** |
| **`NewCMD()`: drop the metrics registration** | **survived** |

The tracer half is well pinned. The metrics half, which this PR moves behind the new indirection, is not: the whole CMD flush can be deleted, or `New` can stop registering the metrics provider, and the suite stays green.

In fairness the CMD gap is inherited — I mutated `development` the same way and its `TestNewCMD_ShutdownMetricsCalledAfterRun` did not catch it either, despite a doc comment claiming it "pins the CMD flush path". It only asserts the handler ran and `Run` returned. But this PR is rewriting those exact lines, so it is the natural place to close it.

Two changes cover all three, both verified:

```go
// in Test_Shutdown_DrainsTelemetry, before appending the fake:
require.Len(t, g.telemetryShutdown, 2)   // metrics + tracer; kills the New() mutation
```

and a CMD counterpart that asserts `NewCMD` registers both providers and that `Run` drains them exactly once after the subcommand returns. I have it written and it kills the other two; happy to paste or push it.

## 6. The shutdown order is documented, and the doc is now stale

`docs/guides/graceful-shutdown/page.md` enumerates the sequence and says "The order is fixed by the framework — see `pkg/gofr/gofr.go`". Step 6 is still `container.ShutdownMetrics(ctx)` and `container.Close()`, but `ShutdownMetrics` has moved out of that step into the telemetry drain, which now runs before it — and the tracer flush is a new step that page does not mention.

- **No exported API change.** I diffed `go doc ./pkg/gofr` between the two trees: identical. The registry, the helper and the renamed constant are all unexported.
- **No cost on the default deployment.** With no `TRACE_EXPORTER` configured — the common case — mean shutdown over five apps is 0 s on both trees. The `NeverSample` provider has no processor, so registering its `Shutdown` costs nothing.
- **Double `Shutdown` is safe.** Repeated calls return `nil`/`nil` with no panic and no hang.
- **`errors.Join` in `drainTelemetry`** matches what `Shutdown` already does, and dropping the join is caught by a test.
- **Registering in `New`/`NewCMD` rather than inside the provider constructors.** Keeps the OTel SDK types out of `Container`, which was the constraint from #3761's thread. Right call.
- **Renaming `metricsFlushTimeout`.** It bounds both now; the name should say so.

Requesting changes on 1 and 3 — the first because the PR does not yet do what it says on the main path, the second because it will redden CI. 2 becomes load-bearing the moment 1 lands, and 4 is one word.

<details>
<summary>How the measurements were produced</summary>

- **Two trees**: this PR at `de7dcb61` and `development`, built into separate binaries, exercised by the same external harness so nothing in the test process could influence the result.
- **Span receiver**: an HTTP server recording POST bodies, used as `TRACER_URL` with `TRACE_EXPORTER=zipkin`.
- **`SIGTERM` case**: child started as a real OS process, one request issued, then `Process.Signal(syscall.SIGTERM)` and `Wait()`. Control run holds the process open past the 5s `BatchSpanProcessor` delay to prove the span is exportable at all.
- **Hung collector**: a listener that accepts connections and never responds; "refused" is a closed port.
- **Mutation testing**: each mutation applied to the PR head, run against a filter whose unmutated baseline is green (`-run 'CMD|Shutdown|Metric|Telemetry|Tracer|tracer|Flush|otel|Otel' -skip TestInitMetricsServer_DefaultPort`, `ok` before each run).
- **Flake rate**: 40 separate `go test -count=1` invocations of the single test.
- Both proposed fixes were applied locally and re-measured; the numbers in finding 1 and 2 are from those runs.

</details>

aryanmehrotra (changes requested):
Everything from the last round is fixed, and I confirmed each one the same way I found it. One new problem, which my own suggestion caused — the join needs a guard I did not think to include.

## The five findings, re-measured

| | before | now |
|---|---|---|
| span delivered on a real `SIGTERM` | **0 of 5 runs** | **1 of 1, three runs out of three** |
| span delivered from a CMD process | 1 | 1 |
| shutdown against a hung collector | 30.002 s | **10.37 s** |
| `Test_Shutdown_DrainsTelemetry`, 40 single runs | **14 failed** | **0 failed** |
| `-race` on the telemetry tests | **fails** | **passes** |
| `goconst` on `cron.go` | **new issue** | gone — 14 issues in `pkg/gofr`, same as `development` |
| mutations surviving | 3 of 7 | **0 of 7** |

The three test gaps are all closed. `Run`/CMD dropping `drainTelemetry`, and `New()` or `NewCMD()` dropping the metrics registration, are each caught now — I re-ran every mutation against a baseline I verified green first.

The docs read correctly too: the drain is its own numbered step, the renumbering leaves `container.Close` genuinely second-to-last so the paragraph below it stays true, and the note that the step carries its own bound independent of `SHUTDOWN_GRACE_PERIOD` is the part an operator actually needs.

Dropping the running server from `Test_Shutdown_DrainsTelemetry` was the right call rather than reaching for `atomic.Int64` — the test was never about a live server, and the race went away with it instead of being papered over.

## The new problem, and it is mine

The join I asked for waits unconditionally. If `startAllServers` returns without the context having been cancelled, nothing is draining and nothing ever closes `shutdownDone`, so `Run` sits on the `time.After(timeout)` arm for the entire grace period:

```go
	app := gofr.New()   // no routes registered
	// METRICS_PORT=0, which two guides document as the way to disable the metrics server
	app.Run()
```

```
development   Run() returned after 0s
this branch   Run() returned after 30s
```

Same with a cron-only service and `METRICS_PORT=0` — a real application shape, since cron is not in `startAllServers`' WaitGroup:

```
development   Run() returned after 20ms
this branch   Run() returned after 30.01s
```

To be fair to the second case, an immediate return there is arguably already wrong — a cron app that exits at once never fires a job — so `development` is not right either. But turning a 20 ms return into a 30 s one is not the fix for that, and `METRICS_PORT=0` is documented in `docs/guides/load-testing` and `docs/guides/cloud-deployment` as a supported configuration.

The guard is small: only wait if a signal actually started the shutdown goroutine.

```go
	a.startAllServers(ctx)

	// Only wait if a termination signal actually started the shutdown goroutine. If the
	// servers simply stopped on their own, nothing is draining and there is nothing to
	// wait for — blocking here would cost the whole grace period for no reason.
	if ctx.Err() != nil {
		select {
		case <-shutdownDone:
		case <-time.After(timeout):
		}
	}
```

It is safe in the direction that matters: on the `SIGTERM` path the context is cancelled before the servers stop, so by the time `startAllServers` returns `ctx.Err()` is already set. Verified with all three measurements — the span still arrives on `SIGTERM` (3 runs out of 3), the CMD case still delivers, the hung collector is still bounded at 10.35 s, and the no-servers case is back to 0 s.

## Two mutations that now survive, both on the new safety code

Neither is a defect, but both are properties the next person could delete without the suite noticing:

| mutation | |
|---|---|
| remove the `telemetryFlushTimeout` bound from `drainTelemetry` | **survives** |
| remove the `shutdownDone` join from `Run` | **survives** |

The second is the fix for the whole first round: the HTTP flush works now and nothing pins that it does. Both are awkward to unit-test — the honest versions are the process-level ones I ran, which do not belong in `go test`. If you want a cheap approximation, a test that registers a shutdown func blocking longer than `telemetryFlushTimeout` and asserts `drainTelemetry` returns before the caller's much longer context expires would pin the bound. The join I would leave, with a comment saying why it is there, rather than build a signal-driven test for it.

## Not flagging

- **The CMD path's `context.Background()`.** Correct once `drainTelemetry` bounds itself; the previous `WithTimeout` there was doing the same job twice.
- **Bounding inside `drainTelemetry` rather than at each call site.** It means no future caller can forget, which is the right place for a safety bound.
- **`defer close(shutdownDone)` as the goroutine's first statement.** Closes on every path out, including a panic in `Shutdown`.
- **The commit message.** It states what was measured and what changed, which made this re-review considerably faster.

Requesting changes only for the guard. Everything else is done, and the measurements above are on your current head.

<details>
<summary>How these were measured</summary>

Same harness as the last round: this PR at `5167a4a9` and `development` built into separate binaries, driven as real OS processes by an external parent that stands up a span receiver, issues one request, then sends a real `SIGTERM` and waits. Hung collector is a listener that accepts and never answers. Flake rate is 40 separate `go test -count=1` invocations. Every mutation was run against a filter whose unmutated baseline I confirmed `ok` first. The guard above was applied locally and all four measurements re-run.

</details>

aryanmehrotra:
Re-tested the latest head (`df351777`):
- The `if ctx.Err() != nil` guard works as expected — no-routes with `METRICS_PORT=0` exits immediately (~1.5ms) instead of hanging, and real `SIGTERM` still joins `shutdownDone` and delivers buffered spans.
- Shutdown against a hung collector remains bounded at ~10s.
- `TestDrainTelemetry_BoundsItself` passes and pins the `telemetryFlushTimeout` deadline on the context.
- Full test suite, race detector (`-race`), and flake checks (40 runs) are clean.

All review comments are addressed. Please resolve the merge conflicts with `development` (in `pkg/gofr/gofr.go` after #3858 merged) so we can get this merged.

Umang01-hash (changes requested):
Reviewed end to end at `863f4a51`, locally — I traced the shutdown state machine from the ticket myself and ran fail-on-revert on the tests rather than taking the description at face value.

**The fix is right, and I verified it rather than reading it.** One thing to add before merge.

## The fix checks out

Confirmed against `development` that this is a genuine gap, not a phantom: `App.Shutdown` there drains metrics only (`gofr.go:127`, `container.ShutdownMetrics`) and never calls `tp.Shutdown`, and `Run` never waits for the shutdown goroutine — so on `SIGTERM` buffered spans were dropped and even the metrics flush could be truncated by `main` returning.

| What | `development` | This PR |
|---|---|---|
| `tp.Shutdown` registered (no-exporter / `NeverSample` path) | no | `otel.go:59` |
| `tp.Shutdown` registered (real exporter path) | no | `otel.go:85` |
| Trace flush on `App.Shutdown` (HTTP/gRPC) | no | `gofr.go` → `drainTelemetry` |
| Trace flush on CMD path | no | `run.go` → `drainTelemetry` |
| `Run` waits for the drain on `SIGTERM` | no | `run.go:71-78` |

Also clean, all checked locally at the ref:
- **Ordering is correct** — `Shutdown` stops servers → `drainTelemetry` → `container.Close`, so in-flight request spans reach the batch buffer before `tp.Shutdown` flushes it.
- **Timeouts nest** — `drainTelemetry` bounds itself to `telemetryFlushTimeout` (10s) inside the 30s shutdown grace, so a hung collector can't hold shutdown open.
- **No API break** — exported `pkg/gofr` surface is identical to `development` (`startShutdownHandler`'s new param and the `metricsFlushTimeout`→`telemetryFlushTimeout` rename are both unexported, no dangling refs).
- `gofmt` clean, `golangci-lint --new-from-rev` **0 issues**, docs updated (`graceful-shutdown/page.md`).
- **Three of the four tests are properly revert-red** — dropping either `tp.Shutdown` append → `Test_initTracer_RegistersShutdown_*` red; swapping `drainTelemetry` back to `ShutdownMetrics` → `Test_Shutdown_DrainsTelemetry` red.

## Changes requested — the headline behaviour has no test

The whole point of the PR on the `SIGTERM` path is that `Run` now **blocks** until `Shutdown`'s drain finishes, so `main` doesn't exit and drop the flush. That is this block in `run.go:71-78`:

```go
if ctx.Err() != nil {
	select {
	case <-shutdownDone:
	case <-time.After(timeout):
	}
}
```

I deleted the entire block (replaced with `_ = shutdownDone`) and ran `go test ./pkg/gofr/` — **the full suite still passes.** No test delivers a signal to a live `Run`, so the one guarantee the PR exists to make rests on the comment above it. The registration + drain *mechanism* is well covered; it's only the `Run`-level wait that isn't.

**Please add one test for it.** Shape: `go app.Run()` an HTTP app, register a telemetry-shutdown func that records completion (closes a channel / sets an atomic), deliver `SIGTERM` (`syscall.Kill(syscall.Getpid(), syscall.SIGTERM)`), then assert the func has completed *by the time `Run` returns*. Delete the block above and that test should go red.

## Non-blocking — pre-existing, not for this PR

The non-signal exit paths still skip the drain: an HTTP server goroutine returning on a `ListenAndServe` error, or a server-less app with `METRICS_PORT=0` and a trace exporter configured, both leave `run.go:71` with `ctx.Err() == nil`, so the parked shutdown goroutine never runs. This is unchanged from `development` (those paths never drained either) and low-impact — a failed-startup app has ~no buffered spans — so it's a follow-up, not something to fix here.

Add the wait test and I'll approve.

om7057:
@Umang01-hash Added the wait test, and resolved the merge conflict with `development`.

`TestRun_WaitsForTelemetryDrainOnCancellation` pins the guarantee: I split the post-`startShutdownHandler` body of `Run` into `runUntilShutdown` so the wait can be driven with a directly cancelable `context.Context` instead of a real OS `SIGTERM`. A real `SIGTERM` sent from inside `go test` turned out to be process-wide: this package's suite leaks other `Run` goroutines with their own `signal.NotifyContext(SIGTERM)` registrations still alive after their tests complete, so delivering a real signal woke those up too and crashed unrelated, already-finished tests (`panic: Fail in goroutine after Test_APIKeyAuthMiddleware has completed`). Driving `runUntilShutdown` directly avoids that while still exercising the exact same wait logic `Run` uses.

Deleting the `shutdownDone` wait leaves the new test red, confirmed locally by reverting it and running the test 3x.

Also merged `development` in to resolve the conflict in `otel_test.go` (both sides had added independent test functions next to each other).

Full `pkg/gofr` suite is green, including under `-race`.

aryanmehrotra (changes requested):
Re-reviewed end to end at `bbd794a1`, merged locally with current `development` (`23852af7`, clean merge).

**The code is ready. CI is not.** One lint issue in the new test is the only thing blocking this.

## Blocking: `Code Quality` is red on the new test

```
pkg/gofr/gofr_test.go:1585:3: missing whitespace above this line (invalid statement above expr) (wsl_v5)
```

This is the failing check on the PR ([job](https://github.com/gofr-dev/gofr/actions/runs/34991068650/job/104455555897)), and I reproduced it locally with the same `golangci-lint v2.12.2` CI pins. The fix is one blank line in `TestRun_WaitsForTelemetryDrainOnCancellation`:

```go
	go func() {
		defer close(runDone)

		app.runUntilShutdown(ctx, timeout, shutdownDone)
	}()
```

## Every earlier finding is addressed

| Round | Ask | At `bbd794a1` |
|---|---|---|
| 1 | Join the shutdown goroutine so `SIGTERM` flushes | done: `runUntilShutdown` waits on `shutdownDone` |
| 1 | Bound `drainTelemetry` | done: `telemetryFlushTimeout`, pinned by `TestDrainTelemetry_BoundsItself` |
| 1 | Flaky and racy `Test_Shutdown_DrainsTelemetry` | done: green under `-race` |
| 1 | `goconst` | done: uses `gofrTraceExporter` |
| 1 | Mutation gaps, stale docs | done |
| 2 | Only wait when a signal started shutdown | done: `ctx.Err() != nil` guard |
| Umang | Test for the `Run`-level wait | added (see the lint issue above) |

## Verified

| Check | Result |
|---|---|
| `go vet ./pkg/gofr` | clean |
| Shutdown and telemetry tests, `-race` | pass |
| `TestRun_WaitsForTelemetryDrainOnCancellation`, 30 runs | 30/30 pass |
| **Same test with the `shutdownDone` wait deleted** | **20/20 fail**, so it really pins the wait |
| `-race` on the full `pkg/gofr` suite | race in `TestContext_WriteMessageToSocket`, which is **also on `development`** (same result on 3 runs of each), so not from this PR |

Moving the wait into `runUntilShutdown` so it can be tested with a context you cancel yourself, instead of a real `SIGTERM` that reaches every leaked `signal.NotifyContext` in the package, was the right call. Thanks for writing down why.

## Non-blocking

- **No test pins the `ctx.Err() != nil` guard.** I replaced it with `if true` and the full `pkg/gofr` suite still passed. That brings back the 30 s hang from round 2 (no routes, `METRICS_PORT=0`). A cheap test: call `runUntilShutdown` on a server-less app (`METRICS_PORT=0`, no routes) with a context that is never cancelled, and assert it returns well under `timeout`.
- **Stale comment**, `run.go:111`: "see the select on shutdownDone in Run". That select now lives in `runUntilShutdown`.
- `bbd794a1` is titled "remove em dashes from new comments", but some are still there: `gofr.go:77`, `run.go:27`, `run.go:56`, `run.go:111`. Cosmetic, and no lint rule enforces it.
- Pre-existing, and already marked as a follow-up in the last round: non-signal exits (a server failing to start, a server-less app that exits on its own) still skip the drain.

Fix the lint line and I'm happy with this. The guard test and the comment fix would be nice in the same push.

<details>
<summary>How this was checked</summary>

- PR head `bbd794a1` merged with `origin/development` `23852af7` in a separate worktree.
- Wait mutation: the `if ctx.Err() != nil { select { ... } }` block replaced with no-ops, then the new test run in 20 separate `go test -count=1` invocations.
- Guard mutation: `if ctx.Err() != nil` changed to `if true`, then the full `pkg/gofr` suite.
- Lint: `golangci-lint v2.12.2 run --new-from-rev=origin/development` in `pkg/gofr`, matching the CI job's `only-new-issues: true`.
- Race baseline: `-race -run '^TestContext_WriteMessageToSocket$'`, 3 runs each on `development` and on this branch.

</details>

om7057:
@aryanmehrotra All addressed at b9192b3d:

- Fixed the wsl_v5 lint failure: missing blank line above app.runUntilShutdown in the new test.
- Added TestRunUntilShutdown_ReturnsImmediatelyWithoutCancellation for the other half of the guard (the ctx.Err() check itself). Confirmed red against if true, and confirmed it actually hangs without app.httpRegistered = false: New auto-registers pkg/gofr/static as an HTTP server whenever that directory exists in the working directory (it does, for this package's own tests), so a naive version of this test would block on ListenAndServe forever instead of exercising the guard.
- Fixed the stale run.go:111 comment pointing at a select that moved into runUntilShutdown.
- Cleaned up the remaining em dashes in run.go, gofr.go, and gofr_test.go that the last pass missed (plus one more in my own new test comment).

golangci-lint v2.12.2 --new-from-rev=origin/development is clean locally, go test -short -race ./pkg/gofr passes, and the only race in the full pkg/gofr suite is the pre-existing Test_WebSocket_Success/websocket.go:47 one you already confirmed is also on development.

PiyushSingh-ZS (commented):
Re-reviewed at `b9192b3d`. Everything from the three earlier rounds is addressed, and I checked the fixes rather than the summaries: `go build ./...` and `go vet ./pkg/gofr` are clean, and all 29 shutdown/telemetry tests pass locally, including `TestRun_WaitsForTelemetryDrainOnCancellation` and the new `TestRunUntilShutdown_ReturnsImmediatelyWithoutCancellation`.

**Blocking: the same linter as last round, two lines further down.**

```
pkg/gofr/gofr_test.go:1624:2: missing whitespace above this line (no shared variables above expr) (wsl_v5)
pkg/gofr/gofr_test.go:1625:2: missing whitespace above this line (invalid statement above assign) (wsl_v5)
```

Both are in the guard test added to close the last round's non-blocking ask:

```go
	start := time.Now()

	app.runUntilShutdown(t.Context(), timeout, shutdownDone)

	elapsed := time.Since(start)
```

`wsl_v5` wants a blank line above an expression statement that shares no variable with the line above it, and above an assignment whose predecessor is an expression statement. Caveat: I could not run the CI linter locally (mine is v2.5.0, built with go1.25; the repo targets 1.26), so this is from the rule semantics and the CI message rather than a local green run.

Worth doing before the next push rather than after it -- this is the second round blocked on `wsl_v5` in a newly added test, and the previous fix introduced these two. `golangci-lint v2.12.2 run --new-from-rev=origin/development ./pkg/gofr/` reproduces what the job sees.

**One thing worth a comment, because it is load-bearing and invisible.**

`New()` registers `app.container.ShutdownMetrics` as a bound method value before the container is fully populated:

```go
	app.container = container.NewContainer(app.Config)
	app.telemetryShutdown = append(app.telemetryShutdown, app.container.ShutdownMetrics)
```

This is correct, but only because `ShutdownMetrics` has a **pointer** receiver, so the later `c.shutdownMetrics = shutdown` assignment is visible through the captured pointer. Had it been a value receiver, the registry would hold a copy taken before the metrics provider existed and would silently flush nothing. Nothing in the code says the registration order depends on that. One clause at the registration site would stop someone "simplifying" the receiver later and breaking the flush in a way no test would catch.

**Sequencing: this conflicts with #3942.**

That PR rewrites the same region of `Run()` -- it inserts `bindMCPServer(ctx)` between `handleStartupHooks` and the shutdown-timeout lookup, and adds `shutdownAfterFailedStartup` -- while this one splits `Run` into `Run` + `runUntilShutdown` and changes `startShutdownHandler`'s signature. `run.go` is +83/-9 there against +51/-14 here, overlapping. Whichever lands second needs a manual merge.

They are complementary rather than competing: #3942's `shutdownAfterFailedStartup` calls `a.Shutdown`, which after this PR drains telemetry, so it partly closes the non-signal-exit gap @Umang01-hash listed as a follow-up. Worth the two authors agreeing an order rather than discovering it in a rebase.

**Two small ones, neither blocking:**

- `TestRunUntilShutdown_ReturnsImmediatelyWithoutCancellation` sets `timeout := 5 * time.Second` and asserts `elapsed < timeout`, but the failure mode it guards against waits on `time.After(timeout)` -- so the pass/fail boundary is the same duration as the wait, and the margin is the timer's own resolution. It does go red when mutated, so this is about clarity: asserting `elapsed < time.Second` against the 5s timeout states the intent and fails fast instead of after the full wait.
- `telemetryShutdown` sits directly above `readinessChecks`, which documents "Guarded by mu: registration is documented as pre-Run, but probes run concurrently." `telemetryShutdown` says nothing about concurrency. It is safe today -- every append is in `New`/`NewCMD`/`initTracer` and every read is in `drainTelemetry` -- but next to a field that spells out its guard, the silence reads as an omission and invites a late append.

Also confirmed the em dashes from the last round are gone from the Go comments; the only remaining ones are in the markdown prose in `docs/guides/graceful-shutdown/page.md`, which is fine.

Fix the two lint lines and this looks good to me.

aryanmehrotra:
@om7057 can you resolve the conflicts?

om7057:
Closing this as superseded. While this was in review, #4206 (feat(tracing): add a trace exporter registry, resource and shutdown flush) merged into development, and its description explicitly states it closes #3771, the same underlying issue this PR fixes.

The two PRs independently implement the same fix with different designs:

- This PR: a telemetryShutdown []func registry drained by drainTelemetry, with runUntilShutdown waiting on a shutdownDone channel after a termination signal.
- #4206 (merged): a single shutdownTracer field flushed by shutdownTraces, with awaitShutdown providing the identical wait-for-signal-shutdown-to-finish behavior via shutdownWaitMargin.

#4206's description explains the registry approach was considered and deliberately not used: "One ShutdownFunc field rather than #3771's telemetryShutdown []func slice. The slice would re-plumb the metrics flush that has already merged, giving a traces-only PR a metrics blast radius for no gain here."

Since development already has a considered, merged fix for the same bug (including the Run-blocking-on-shutdown guarantee that was the subject of the last two rounds of review here), continuing to resolve the merge conflicts in this PR would just create two competing shutdown mechanisms in the same codebase rather than add anything new. Thank you all for the very thorough review across every round, it's what surfaced the Run-blocking gap in the first place, which is exactly the behavior #4206 also had to get right.

PR #3927, opened 2026-08-13
Release/v1.59.1

aryanmehrotra:
Carries the gcp exporter `go.mod` fix (#3926) from `development` to `main`, so that `pkg/gofr/metrics/exporters/gcp/v0.1.0` can be tagged alongside the rest of the v1.59.0 submodule tags.

### Contents

| File | Change |
|---|---|
| `pkg/gofr/metrics/exporters/gcp/go.mod` | `require gofr.dev v1.59.0` (was v1.57.0), `replace` dropped; grpc → v1.83.0, x/net → v0.57.0 |
| `pkg/gofr/metrics/exporters/gcp/go.sum` | re-tidied against the published v1.59.0 |
| `go.work.sum` | pruned hashes no longer reachable |
| `pkg/gofr/version/version.go` | v1.59.0 → v1.59.1 |

### Why this matters

The gcp exporter is published as its own module, so consumers resolve it through its `go.mod`. It required `gofr.dev v1.57.0` and relied on a `replace` to compile — and a `replace` in a dependency is ignored by the importing module. A consumer therefore resolved v1.57.0, which predates `exporters.Config`, `exporters.Logger` and `exporters.Register`:

```
./gcp.go:37:12: undefined: exporters.Register
./gcp.go:46:54: undefined: exporters.Config
./gcp.go:46:79: undefined: exporters.Logger
```

The module has never been tagged, so nothing is broken in the wild — but tagging `gcp/v0.1.0` off the current `main` would have published an unbuildable module.

### Note on the core version

**No core framework code changes in this release.** Against the `v1.59.0` tag, the only difference to the `gofr.dev` module itself is `version.go`; everything else is confined to the gcp submodule and `go.work.sum`. The bump exists to give the submodule tags a clean anchor commit on `main` and to keep every `main` commit a release, per existing practice. Happy to drop it and merge the fix without a version bump if you'd rather not publish a no-op core version.

### After merge

Tag `v1.59.1`, then cut the submodule tags at that merge commit — 26 patch bumps plus `file/s3` as a minor (its exported `Logger` interface gained `Warnf`), plus `pkg/gofr/metrics/exporters/gcp/v0.1.0` as a first release.


aryanmehrotra:
Superseded by #3928, which carries the same gcp go.mod fix to main without the v1.59.1 version bump. The bump would have published a gofr.dev zip identical to v1.59.0 apart from its own version string, since pkg/gofr/metrics/exporters/gcp is a separate module and is excluded from the parent module's zip.

PR #3928, opened 2026-08-13
fix(metrics): bring the gcp exporter go.mod fix to main

aryanmehrotra:
Brings the gcp exporter `go.mod` fix (#3926, already merged to `development`) onto `main`, so `pkg/gofr/metrics/exporters/gcp/v0.1.0` can be tagged from a `main` commit alongside the other v1.59.0 submodule tags.

### Diff against `main` — three files, no framework code

| File | Change |
|---|---|
| `pkg/gofr/metrics/exporters/gcp/go.mod` | `require gofr.dev v1.59.0` (was v1.57.0); `replace` dropped; grpc → v1.83.0, x/net → v0.57.0 |
| `pkg/gofr/metrics/exporters/gcp/go.sum` | re-tidied against the published v1.59.0 |
| `go.work.sum` | pruned hashes no longer reachable |

**No version bump.** `pkg/gofr/version/version.go` stays at `v1.59.0`, because none of this is in the `gofr.dev` module: `pkg/gofr/metrics/exporters/gcp/` has its own `go.mod`, so Go excludes it from the parent module's zip. Confirmed against the published module:

```
$ ls $(go env GOMODCACHE)/gofr.dev@v1.59.0/pkg/gofr/metrics/exporters/
config.go  consts.go  exporter.go  otlp.go  provider.go  registry.go  telemetry.go ...   # no gcp/
```

A v1.59.1 would therefore have shipped a zip identical to v1.59.0 apart from its own version string, so this merges as a plain fix instead.

### Why the fix is needed

The gcp exporter is published as its own module, so consumers resolve it through its `go.mod`. A `replace` directive in a dependency is ignored by the importing module, so a consumer resolved `gofr.dev v1.57.0` — which predates `exporters.Config`, `exporters.Logger` and `exporters.Register`:

```
./gcp.go:37:12: undefined: exporters.Register
./gcp.go:46:54: undefined: exporters.Config
./gcp.go:46:79: undefined: exporters.Logger
```

The module has never been tagged, so nothing is broken in the wild — but tagging `gcp/v0.1.0` off the current `main` would publish an unbuildable module.

Nothing catches this in CI today: `go.work` `use`s every submodule, so the submodule job compiles them against the local tree rather than the minimum each declares. I checked the other 26 submodules the same way and they all build against their declared minimum; gcp is the only one where the two disagree.

### After merge

Tag the submodules at this merge commit: 26 patch bumps, `file/s3` as a minor (its exported `Logger` interface gained `Warnf`), and `pkg/gofr/metrics/exporters/gcp/v0.1.0` as a first release.

Supersedes #3927.


Umang01-hash (approved):
Verified on main. `GOWORK=off -mod=readonly go build/vet` passes at v1.59.0 and `go mod verify` is clean; pinning v1.57.0 standalone fails with `undefined: exporters.Register/Config/Logger` (gcp.go:37,46), so both the bump and dropping the local `replace` are necessary. main's gcp.go is identical to development's, so it needs the same floor. Pairs with #3926 for development — worth landing both so the branches don't diverge. LGTM.

Issue #3930, opened 2026-08-13
Data race: App.WebSocket writes ctx.Context while serveWithGoroutine reads it (Test_WebSocket_Success fails under -race)


akshat-kumar-singhal:
**Describe the bug**

`App.WebSocket`'s handler writes `ctx.Context` while `serveWithGoroutine` concurrently reads
`c.Context` in its `select`, so every WebSocket request races the framework's own handler wrapper.

This is the websocket race deferred in 050596e8 ("The gRPC server ... and a pre-existing websocket
race are out of scope ... tracked separately"), which never got an issue. The gRPC half is #3929.

**The race**

Write, on the goroutine `serveWithGoroutine` spawned:

```go
// pkg/gofr/websocket.go:41
ctx.Context = context.WithValue(ctx, websocket.WSConnectionKey, conn)
```

Read, on the goroutine that spawned it:

```go
// pkg/gofr/handler.go:180
select {
case <-c.Context.Done():
```

`serveWithGoroutine` was written to avoid exactly this — its own comment says "shared variables, so
the goroutine never writes a memory location the main goroutine reads — `go test -race` stays
clean". That holds for the outcome value, which is passed over a channel, but not for `c.Context`:
the handler is handed the same `*gofr.Context` and is free to reassign the field, which
`App.WebSocket` does on every connection.

**Relationship to #3823**

Same class — an unsynchronised write to `c.Context` that another goroutine reads — but a different
write site and a different reader. #3823 is `Context.Trace()` racing user code in gRPC streaming
handlers; this one is framework code (`App.WebSocket`) racing framework code
(`serveWithGoroutine`), and it reproduces in this repo's own test suite with no user code involved.
Worth deciding together: a fix that makes `c.Context` safe to reassign (or removes the need to
reassign it) would close both. Filing separately rather than burying it in #3823's thread.

**To Reproduce**

```
go test ./pkg/gofr/ -race -count=1 -short -run 'Test_WebSocket_Success'
```

Reproduces on clean `development` (verified at `ceb50390`); `--- FAIL: Test_WebSocket_Success ...
race detected during execution of test`.

Note `go test ./pkg/gofr/websocket/... -race` is green — the race is in the handler wiring in
`pkg/gofr`, not in the websocket package itself.

**Expected behaviour**

A WebSocket handler storing the connection on its context does not race the request wrapper, and
`go test -race ./pkg/gofr/` is clean.

PR #3931, opened 2026-08-13
fix(grpc): guard the server handle so Run and Shutdown stop racing (#3929)

### LINKED ISSUE #3929: Data race on grpcServer.server between Run and Shutdown (three gRPC tests fail under -race)
**Describe the bug**

`grpcServer.server` is written on the serve goroutine and read on the shutdown goroutine with no
synchronisation. This is the same defect as #3762 (`httpServer.srv`), on a different field — and it
was explicitly left out of that fix's scope:

> The gRPC server (races on a different field, `g.server`, created at registration) and a
> pre-existing websocket race are out of scope and reproduce on development independently;
> tracked separately.
> — 050596e8 ("fix(server): guard srv with a mutex to remove HTTP/metrics/MCP shutdown data race")

"Tracked separately" never produced an issue, so this is that issue.

**The race**

`createServer()` writes the field, reached from `Run()` on whichever goroutine serves:

```go
// pkg/gofr/grpc.go:130
g.server = grpc.NewServer(g.options...)
```

`Shutdown()` reads it on the caller's goroutine, in both the graceful and the force-close closure:

```go
// pkg/gofr/grpc.go:189-201
func (g *grpcServer) Shutdown(ctx context.Context) error {
	return ShutdownWithContext(ctx, func(_ context.Context) error {
		if g.server != nil {          // <- race read
			g.server.GracefulStop()   // <- and here
		}
		return nil
	}, func() error {
		if g.server != nil {          // <- race read
			g.server.Stop()           // <- and here
		}
		return nil
	})
}
```

Nothing establishes a happens-before edge between `App.startGRPCServer` → `Run()` and
`App.Shutdown()` → `Shutdown()`.

The detector reports it twice per occurrence: once on the `g.server` field itself, then once inside
grpc-go, where `NewServer`'s `grpcsync.NewEvent` write races `Server.Stop`'s `CompareAndSwapUint32`
(`google.golang.org/grpc@v1.83.0/server.go:1944`) — the downstream consequence of the first.

**This is not test-only.** It is the nil-check-as-started pattern #3762 named. If `Shutdown` reads a
stale `nil`, both closures no-op and `Shutdown` returns success having stopped nothing, so the
process can exit with the gRPC server still serving and in-flight RPCs cut at process death rather
than drained by `GracefulStop`.

**To Reproduce**

```
go test ./pkg/gofr/ -run 'TestGRPC_ServerShutdown_ContextCanceled' -count=1 -race
```

Fails on clean `development` (verified at `ceb50390`). A full `-race` run of the package reports it
across three tests:

- `TestStartGRPCServer_Registered`
- `TestGRPC_ServerShutdown`
- `TestGRPC_ServerShutdown_ContextCanceled`

**Expected behaviour**

`go test -race ./pkg/gofr/` is clean, and `Shutdown` always observes the server the serve goroutine
created, so `GracefulStop`/`Stop` actually run.

**Proposed fix**

Mirror what #3762 did for `httpServer`: guard `server` with a `sync.Mutex` on `grpcServer`, taking a
local copy under the lock in `Run` and `Shutdown`. The lock must never be held across the blocking
`Serve()` / `GracefulStop()` — only around the assignment and the read.

**Scope**

Framework-core (`pkg/gofr/grpc.go`); its own PR off `development`, not folded into an unrelated one.
The websocket race from that same scope note is filed separately.

akshat-kumar-singhal:
**Description:**

Fixes #3929.

`grpcServer.server` was written on the serve goroutine and read on the shutdown goroutine with no synchronisation — the same defect #3762 fixed for `httpServer.srv`, on the field that fix explicitly deferred:

> The gRPC server (races on a different field, `g.server`, created at registration) and a pre-existing websocket race are out of scope and reproduce on development independently; tracked separately.
> — 050596e8

**Root Cause Analysis**

**Symptom.** Three tests fail under `-race` on clean `development` (verified at `ceb50390`): `TestStartGRPCServer_Registered`, `TestGRPC_ServerShutdown`, `TestGRPC_ServerShutdown_ContextCanceled`. The detector reports each occurrence twice — once on `g.server` itself, then once inside grpc-go, where `NewServer`'s `grpcsync.NewEvent` races `Server.Stop`'s `CompareAndSwapUint32` (`google.golang.org/grpc@v1.83.0/server.go:1944`), which is the downstream consequence of the first.

**Root cause.** `createServer()` assigns the field (`grpc.go:130`), reached from `Run()` on whichever goroutine serves; `Shutdown()` dereferences it on the caller's goroutine, in both the graceful and the force-close closure (`grpc.go:189-201`). Nothing orders `App.startGRPCServer → Run()` against `App.Shutdown() → Shutdown()`.

**Impact beyond tests.** This is the nil-check-as-started pattern #3762 named. If `Shutdown` observes a stale `nil`, both closures no-op and `Shutdown` returns success having stopped nothing — the process can exit with the gRPC server still serving, and in-flight RPCs are cut at process death instead of drained by `GracefulStop`.

**Why it wasn't caught.** The race is a latent field race, invisible without `-race`, and CI's race legs are sharded by Go version — the same commit goes green on one shard and red on another, which reads as flake. #3762 found and fixed the identical pattern in the three HTTP-family servers but scoped the gRPC field out for a separate change; that change was never filed until #3929.

**Fix.** A `sync.Mutex` on `grpcServer` guarding `server`, mirroring `httpServer` so the four servers share one shape:
- `createServer` builds into a local, registers reflection on the local, and publishes under the lock **last** — so a concurrent `Shutdown` sees either no server or a fully set-up one, never a half-registered one.
- `Run`, `Shutdown` and `RegisterService` read through `getServer()`. The lock is never held across a blocking call (`Serve`, `GracefulStop`) — each operates on the returned copy.
- `TestStartGRPCServer_Registered` read `app.grpcServer.server` directly and raced the same write; it now stops via `forceStop()`.

**Prevention.** `TestGRPC_ConcurrentRunAndShutdown` races `Shutdown` against `Run`'s publish with no sleep between them. Mutation-verified: with the two `srvMu` Lock/Unlock pairs removed it reports `WARNING: DATA RACE` on every run; with them, clean at `-count=5 -race`.

**On the helper methods**

The natural shape is one guarded read at the top of `Shutdown` into a local both closures capture. `gracefulStop()`/`forceStop()` exist because a local pre-commit hook rejects edits spanning `Shutdown`'s `context.Context` signature line; the methods give the same guarantee at the cost of one extra guarded read on the force path. Since the field never changes once published, both reads observe the same server. Happy to fold them back into a single local read if reviewers prefer that form.

**Why the read side takes the lock too, and mutex vs `atomic.Pointer`**

Guarding only the write would fix nothing. A mutex creates ordering solely between parties that both take it: the release in `createServer` synchronises-with a matching *acquire*, so a lock-free reader pairs with nothing and is left with both a data race on the field and no guarantee that the constructed `Server` is visible to it. That is not theoretical here — with `createServer` and `Shutdown` already guarded, `TestStartGRPCServer_Registered` still raced, because it read `app.grpcServer.server` directly:

```
Read at 0x00c0005f4708 by goroutine 22:
  TestStartGRPCServer_Registered   gofr_test.go:1835
Previous write at 0x00c0005f4708 by goroutine 28:
  (*grpcServer).createServer       grpc.go:146     <- already under the lock
```

One unguarded reader reinstates the whole race, which is why `Run`, `RegisterService` and that test all go through `getServer()`/`forceStop()` rather than touching the field.

Note the lock covers only the handle. The pointee needs no protection from us: `grpc.Server` guards its own state with `s.mu` (`server.go:132`, "guards following"), taken by the `stop()` path behind both `Stop` and `GracefulStop`. Hence the local copy and the release of the lock before any blocking call.

Given that, `atomic.Pointer[grpc.Server]` would give identical publish/acquire semantics with lock-free reads, and arguably suits a write-once field better. I chose the mutex only because #3762 established that shape for `httpServer`, `metricServer` and `mcpServer`, and one shape across all four servers seemed worth more than the read cost on a path taken twice per process lifetime. Happy to switch if reviewers prefer the atomic — it is a small, contained change.

**Out of scope, deliberately**

- **Nil-check-as-started (#3801, server half).** Writing the test surfaced it: a `Shutdown` that runs before `Run` publishes reads nil, no-ops, and `Run` then serves forever — the first draft of the test deadlocked on exactly that. Fixing it needs a `stopped` flag and a `Run` that refuses to start after it, a behaviour change well past a race fix. The test stops whatever `Run` published rather than asserting shutdown wins, with a comment pointing at #3801 so the gap stays visible.
- **The websocket race**, the other half of `050596e8`'s deferred scope — `App.WebSocket` writes `ctx.Context` while `serveWithGoroutine` reads it. Filed as #3930.

None. No API or signature changes; `getServer`, `gracefulStop` and `forceStop` are unexported.

**Additional Information:**

No new dependencies.

Verification:

- `go build ./...` and `go vet ./pkg/gofr/` clean.
- `go test ./pkg/gofr/ -race -short` goes from **28 race warnings / 4 failures** to **1 / 1** on this branch. The remaining one is the websocket race (#3930), which reproduces on `development`.
- `go test ./pkg/gofr/ -run 'GRPC|RegisterService' -count=1 -race` green; `go test ./examples/grpc/...` green.
- `golangci-lint run ./pkg/gofr/...` reports nothing in the three changed files.

**Checklist:**

### DIFF
--- a/pkg/gofr/grpc.go
+++ b/pkg/gofr/grpc.go
@@ -20,7 +20,7 @@ import (
 )
 
 type grpcServer struct {
-	// srvMu guards server, which createServer writes on the serve goroutine and Shutdown reads on the caller goroutine.
+	// srvMu guards server, which ensureServer writes on the serve goroutine and Shutdown reads on the caller goroutine.
 	srvMu              sync.Mutex
 	server             *grpc.Server
 	interceptors       []grpc.UnaryServerInterceptor
@@ -125,14 +125,30 @@ func registerGRPCMetrics(c *container.Container) {
 	c.Metrics().NewCounter("app_grpc_rate_limit_exceeded_total", "Total gRPC requests rejected by rate limiter")
 }
 
-func (g *grpcServer) createServer() error {
+// ensureServer returns the server, creating it on first call. The check and the publish happen
+// under a single hold — mirroring httpServer.run — so two concurrent callers cannot both observe
+// nil and both build one, leaving the loser to Serve a server no later getServer sees and no
+// Shutdown stops. The hold also covers the append to g.options.
+//
+// Building under the lock is safe: grpc.NewServer and reflection.Register do not block. The
+// server is published only once fully set up, so a concurrent Shutdown sees either no server or
+// a ready one, never a partially registered one. Callers use the returned copy for blocking
+// calls such as Serve, which must never hold srvMu.
+func (g *grpcServer) ensureServer() (*grpc.Server, error) {
+	g.srvMu.Lock()
+	defer g.srvMu.Unlock()
+
+	if g.server != nil {
+		return g.server, nil
+	}
+
 	interceptorOption := grpc.ChainUnaryInterceptor(g.interceptors...)
 	streamOpt := grpc.ChainStreamInterceptor(g.streamInterceptors...)
 	g.options = append(g.options, interceptorOption, streamOpt)
 
 	srv := grpc.NewServer(g.options...)
 	if srv == nil {
-		return errFailedCreateServer
+		return nil, errFailedCreateServer
 	}
 
 	enabled := strings.ToLower(g.config.GetOrDefault("GRPC_ENABLE_REFLECTION", "false"))
@@ -140,13 +156,9 @@ func (g *grpcServer) createServer() error {
 		reflection.Register(srv)
 	}
 
-	// Publish under the lock, and only after the server is fully set up, so a concurrent
-	// Shutdown either sees no server at all or one that is ready to be stopped.
-	g.srvMu.Lock()
 	g.server = srv
-	g.srvMu.Unlock()
 
-	return nil
+	return srv, nil
 }
 
 // getServer returns the current server under the lock. Callers operate on the returned
@@ -159,16 +171,12 @@ func (g *grpcServer) getServer() *grpc.Server {
 }
 
 func (g *grpcServer) Run(c *container.Container) {
-	srv := g.getServer()
-	if srv == nil {
-		if err := g.createServer(); err != nil {
-			c.Logger.Fatalf("failed to create gRPC server: %v", err)
-			c.Metrics().IncrementCounter(context.Background(), "grpc_server_errors_total")
+	srv, err := g.ensureServer()
+	if err != nil {
+		c.Logger.Fatalf("failed to create gRPC server: %v", err)
+		c.Metrics().IncrementCounter(context.Background(), "grpc_server_errors_total")
 
-			return
-		}
-
-		srv = g.getServer()
+		return
 	}
 
 	if !isPortAvailable(g.port) {
@@ -236,19 +244,18 @@ func (g *grpcServer) forceStop() {
 
 // RegisterService adds a gRPC service to the GoFr application.
 func (a *App) RegisterService(desc *grpc.ServiceDesc, impl any) {
-	if !a.grpcRegistered {
-		if err := a.grpcServer.createServer(); err != nil {
-			a.container.Logger.Errorf("failed to create gRPC server for service %s: %v", desc.ServiceName, err)
-			return
-		}
+	srv, err := a.grpcServer.ensureServer()
+	if err != nil {
+		a.container.Logger.Errorf("failed to create gRPC server for service %s: %v", desc.ServiceName, err)
+		return
 	}
 
 	a.container.Logger.Infof("registering gRPC Service: %s", desc.ServiceName)
-	a.grpcServer.getServer().RegisterService(desc, impl)
+	srv.RegisterService(desc, impl)
 
 	a.container.Metrics().IncrementCounter(context.Background(), "grpc_services_registered_total")
 
-	err := injectContainer(impl, a.container)
+	err = injectContainer(impl, a.container)
 	if err != nil {
 		a.container.Logger.Fatalf("failed to inject container into gRPC service %s: %v", desc.ServiceName, err)
 	}
--- a/pkg/gofr/grpc.go
+++ b/pkg/gofr/grpc.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 
 	grpc_recovery "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/recovery"
 	"google.golang.org/grpc"
@@ -19,6 +20,8 @@ import (
 )
 
 type grpcServer struct {
+	// srvMu guards server, which createServer writes on the serve goroutine and Shutdown reads on the caller goroutine.
+	srvMu              sync.Mutex
 	server             *grpc.Server
 	interceptors       []grpc.UnaryServerInterceptor
 	streamInterceptors []grpc.StreamServerInterceptor
@@ -127,27 +130,45 @@ func (g *grpcServer) createServer() error {
 	streamOpt := grpc.ChainStreamInterceptor(g.streamInterceptors...)
 	g.options = append(g.options, interceptorOption, streamOpt)
 
-	g.server = grpc.NewServer(g.options...)
-	if g.server == nil {
+	srv := grpc.NewServer(g.options...)
+	if srv == nil {
 		return errFailedCreateServer
 	}
 
 	enabled := strings.ToLower(g.config.GetOrDefault("GRPC_ENABLE_REFLECTION", "false"))
 	if enabled == "true" { //nolint:goconst // standard boolean string
-		reflection.Register(g.server)
+		reflection.Register(srv)
 	}
+
+	// Publish under the lock, and only after the server is fully set up, so a concurrent
+	// Shutdown either sees no server at all or one that is ready to be stopped.
+	g.srvMu.Lock()
+	g.server = srv
+	g.srvMu.Unlock()
 
 	return nil
 }
 
+// getServer returns the current server under the lock. Callers operate on the returned
+// copy so a blocking call such as Serve or GracefulStop never holds srvMu.
+func (g *grpcServer) getServer() *grpc.Server {
+	g.srvMu.Lock()
+	defer g.srvMu.Unlock()
+
+	return g.server
+}
+
 func (g *grpcServer) Run(c *container.Container) {
-	if g.server == nil {
+	srv := g.getServer()
+	if srv == nil {
 		if err := g.createServer(); err != nil {
 			c.Logger.Fatalf("failed to create gRPC server: %v", err)
 			c.Metrics().IncrementCounter(context.Background(), "grpc_server_errors_total")
 
 			return
 		}
+
+		srv = g.getServer()
 	}
 
 	if !isPortAvailable(g.port) {
@@ -174,7 +195,7 @@ func (g *grpcServer) Run(c *container.Container) {
 	c.Metrics().SetGauge("grpc_server_status", 1)
 	c.Logger.Infof("gRPC server started successfully on %s", addr)
 
-	if err := g.server.Serve(listener); err != nil {
+	if err := srv.Serve(listener); err != nil {
 		c.Logger.Errorf("error in starting gRPC server at %s: %s", addr, err)
 		c.Metrics().IncrementCounter(context.Background(), "grpc_server_errors_total")
 		c.Metrics().SetGauge("grpc_server_status", 0)
@@ -188,18 +209,29 @@ func (g *grpcServer) Run(c *container.Container) {
 
 func (g *grpcServer) Shutdown(ctx context.Context) error {
 	return ShutdownWithContext(ctx, func(_ context.Context) error {
-		if g.server != nil {
-			g.server.GracefulStop()
-		}
+		g.gracefulStop()
 
 		return nil
 	}, func() error {
-		if g.server != nil {
-			g.server.Stop()
-		}
+		g.forceStop()
 
 		return nil
 	})
+}
+
+// gracefulStop drains in-flight RPCs and returns when they are done. It is a no-op if no
+// server was ever created, which is the case when no service has been registered.
+func (g *grpcServer) gracefulStop() {
+	if srv := g.getServer(); srv != nil {
+		srv.GracefulStop()
+	}
+}
+
+// forceStop closes the listener and cuts in-flight RPCs. Same no-op contract as gracefulStop.
+func (g *grpcServer) forceStop() {
+	if srv := g.getServer(); srv != nil {
+		srv.Stop()
+	}
 }
 
 // RegisterService adds a gRPC service to the GoFr application.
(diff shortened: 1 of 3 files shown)


pkg/gofr/grpc.go
```diff
@@ -127,27 +130,45 @@ func (g *grpcServer) createServer() error {
...
+func (g *grpcServer) getServer() *grpc.Server {
+	g.srvMu.Lock()
+	defer g.srvMu.Unlock()
+
+	return g.server
+}
+
 func (g *grpcServer) Run(c *container.Container) {
-	if g.server == nil {
+	srv := g.getServer()
```
aryanmehrotra: This is the check-then-act. `httpServer.run` does the check and the publish under one hold and returns early if a server already exists; here `getServer()` -> nil check -> `createServer()` -> `getServer()` spans three separate holds.

Running exactly this sequence from two goroutines, 300 attempts:

```
divergent view of g.server in 17/300 attempts
TOCTOU: 17/300 attempts created two servers, orphaning one
```

The loser calls `Serve` on a server `getServer()` no longer returns, so `Shutdown` stops the survivor and the orphan keeps its listener. An `ensureServer()` that checks and publishes under a single hold closes it, and covers the unsynchronised `g.options` append in `createServer` at the same time.

aryanmehrotra (changes requested):
Verified at `95d635bf`, and I reproduced the defect before checking the fix rather than taking the report at face value.

On clean `development` the three tests you name produce **27 data races** and all three fail; the racing write is `grpc.go:130`, read at `:145` by `Run`'s nil check. On this branch: **0 races**, all three pass at `-count=3`, and every gRPC test passes `-race -count=3`.

The lock scope is right, and it is the part that matters. `srvMu` is held at exactly two points - the publish and the read in `getServer` - and all three blocking calls (`Serve` at `:198`, `GracefulStop` at `:226`, `Stop` at `:233`) run on the returned copy, outside any hold. Holding it across `Serve` would deadlock, and the comment on `getServer` names that reason. Publishing only after reflection registration is the other half: a concurrent `Shutdown` sees either no server or a finished one, never a partial.

`TestGRPC_ConcurrentRunAndShutdown` earns its place - with the guard stripped from both sites it reports 9 races and fails, so it is a real regression guard, not decoration.

`go doc -all ./pkg/gofr` is identical to `development` - 112 exported symbols on both sides; everything added is unexported. No API change.

I was going to approve on that. One thing stopped me.

## The mirror of #3762 is incomplete, and the gap is the same class of bug

Choosing `sync.Mutex` over `atomic.Pointer[grpc.Server]` is the right call - two sibling lifecycle fields guarded two different ways would be worse than both using the heavier primitive, and `httpServer.srv` established the shape. But `httpServer.run` does the check **and** the publish under **one** hold:

```go
s.srvMu.Lock()

if s.srv != nil {
	s.srvMu.Unlock()
	c.Logf("Server already running on port: %d", s.port)

	return
}

srv := &http.Server{...}
s.srv = srv
s.srvMu.Unlock()
```

`grpcServer.Run` splits that across three:

```go
srv := g.getServer()            // hold 1: read
if srv == nil {
	g.createServer()            // hold 2: publish
	srv = g.getServer()         // hold 3: read again
}
```

That is check-then-act, and it is reachable. I ran exactly that sequence from two goroutines, 300 attempts:

```
divergent view of g.server in 17/300 attempts
TOCTOU: 17/300 attempts created two servers, orphaning one
```

Both goroutines observe nil, both create, last write wins. The loser then calls `srv.Serve(listener)` on a server `getServer()` no longer returns - so `Shutdown` stops the survivor and the orphan keeps its listener up. That is a lost shutdown, which is #3801's other half arriving by a new route, inside the PR that exists to close this class of bug.

Being straight about reachability: the framework's own path does not hit it, since `startAllServers` calls `Run` once. It needs concurrent entry - a user calling `RegisterService` (public API) from a goroutine while `Run` starts, or two `Run` calls. So this is not failing in production today. Neither was the field race you are fixing, though; that one was found by `-race`, and this one shows up in 6% of attempts. If we are mirroring a fix, the single-hold check is not incidental to it - it is what makes "already running" safe.

## Suggested shape

Fold the check and the create into one hold. `grpc.NewServer` and `reflection.Register` do not block, so building under the lock is fine:

```go
// ensureServer returns the server, creating and publishing it under a single lock
// hold so two callers cannot both observe nil and both create one.
func (g *grpcServer) ensureServer() (*grpc.Server, error) {
	g.srvMu.Lock()
	defer g.srvMu.Unlock()

	if g.server != nil {
		return g.server, nil
	}

	// build srv: interceptors, options, reflection registration

	g.server = srv

	return srv, nil
}
```

`Run` and `RegisterService` both call it. Two things fall out for free:

- `createServer` appends to `g.options` today with no synchronisation - two concurrent callers race that slice as well. Bringing the build inside the hold covers it.
- `RegisterService` currently does `a.grpcServer.getServer().RegisterService(desc, impl)` with no nil check, while the other three call sites all guard. I traced it and it is unreachable - `createServer` returns early on error and otherwise publishes non-nil, and `grpcRegistered` is set nowhere but after a successful registration - so it is a style asymmetry rather than a defect. Under `ensureServer` it stops being asymmetric at all.

Everything else here is good, and the diagnosis and the `-race` evidence in the description are both accurate. Happy to approve as soon as the create is atomic with its own check.

akshat-kumar-singhal:
Fixed in `78a0f9f1`. You're right that the three-hold shape is check-then-act and that it re-opens the lost-shutdown case from inside the PR meant to close it — the single-hold check isn't incidental to `httpServer.run`'s shape, it's the part that makes "already running" mean anything.

`createServer` is now `ensureServer() (*grpc.Server, error)`, checking and publishing under one hold, essentially the shape you sketched:

```go
func (g *grpcServer) ensureServer() (*grpc.Server, error) {
	g.srvMu.Lock()
	defer g.srvMu.Unlock()

	if g.server != nil {
		return g.server, nil
	}

	// interceptors, options, grpc.NewServer, reflection

	g.server = srv

	return srv, nil
}
```

`Run` and `RegisterService` both go through it, and both use the **returned** server rather than a follow-up `getServer()` — so there is no window between deciding a server exists and getting hold of it. `getServer()` stays for `gracefulStop`/`forceStop`. Building inside the hold is what makes the check and the publish atomic; it costs nothing in hold time, since neither `grpc.NewServer` nor `reflection.Register` blocks, and no blocking call moved under the lock — `Serve`, `GracefulStop` and `Stop` still run on the returned copy.

Both of the follow-ons you named fall out as you predicted:

- **`g.options`** — the append is inside the hold now, so two concurrent callers can't race the slice.
- **`RegisterService`'s missing nil check** — it uses the returned `srv`, so the asymmetry is gone rather than merely justified.

One consequence worth calling out: `grpcRegistered` no longer gates creation. It existed to stop `createServer` running twice; `ensureServer` is idempotent by construction, so the flag is now set-and-read only for `run.go`'s "should a gRPC server start at all" decision, which is what it actually means. No behaviour change — `RegisterService` already ran the create on the first call and skipped it after.

**Regression guard.** `TestGRPC_ConcurrentEnsureServer` releases 8 goroutines together on a closed channel; all must get the same pointer, and `options` must hold exactly the two appended entries. Mutation-verified — restoring the three-hold shape fails it on both counts at `-count=5 -race`:

```
Messages:  caller 5 got a different server: two were created
Error:     "[0xc000102298 ... 0xc0000023a8]" should have 2 item(s), but has 6
```

The options length is the more precise signal of the two: it counts builds directly, so it fires even on a run where every caller happened to land on the same survivor.

**Verification on `78a0f9f1`:** `go build ./...`, `go vet ./pkg/gofr/` and `gofmt` clean; `golangci-lint run ./pkg/gofr/...` reports nothing in the two changed files. `go test ./pkg/gofr/ -run 'GRPC|RegisterService|Reflection' -count=5 -race` green, `./examples/grpc/...` green. `go test ./pkg/gofr/ -race -short` still ends at 1 race / 1 failure — `Test_WebSocket_Success` (#3930), unchanged here and reproducing on `development`.

The branch also picked up a merge of `development` (`358cc6c0`) before this commit; it doesn't touch `grpc.go`, and I re-ran the above after rebasing on it.

aryanmehrotra (approved):
Verified at `78a0f9f1`. The check-then-act point is closed, and I measured it rather than reading the reply.

Both halves reproduce first on `development` @ `bc271ffb`: the three tests you name report **27 data races** and fail 3/3. On this branch, **0 races**, passing at `-count=3`, and 0 across every gRPC test at `-count=2`.

For the TOCTOU itself I rebuilt the probe with a control, because my first attempt found 0/300 on the *old* shape too — which would have made a clean result on the new one meaningless. Once the check→act window was widened and both ran at `-cpu 8` under identical scheduling pressure:

| | two servers built (orphan) | divergent view |
| --- | --- | --- |
| old three-hold shape | **294/300** | 293/300 |
| `ensureServer` | **0/300** | 0/300 |

`ensureServer` has no window to widen, which is exactly the property I was asking for. 64 concurrent callers converge on one server; `Shutdown` racing `ensureServer` orphans nothing in 200 attempts; double stop neither panics nor deadlocks.

The positioning is right, and that is most of why I am happy with it. `srvMu` sits immediately above the field it guards; the shape now mirrors `httpServer.run` from #3762 rather than approximating it; no lock is held across `Serve`, `GracefulStop` or `Stop`, all of which run on the returned copy; and `grpcRegistered` is preserved — only the guard that `ensureServer`'s idempotence made redundant is gone. Cost is a one-off 3µs under the hold at startup, then 12ns/op for a cached `ensureServer` and 9ns/op for `getServer`, neither on a per-RPC path.

One thing I want on the record, because I think the issue overstates it and we should not carry that forward.

I built a real app with the generated `Hello` service, served real RPCs, and put a 3s call in flight before shutting down. It drains correctly — `Shutdown` waits the full 3s, the in-flight RPC completes, the port closes. **It behaves identically on `development`.** Racing `Run` against `Shutdown` against a live server reports 0 races on both sides.

The reason is the gate: `startGRPCServer` runs only `if a.grpcRegistered` (`run.go:166`), and that is set only at the end of `RegisterService` (`grpc.go:263`), which has already published the server on the caller's goroutine via `ensureServer`. Goroutine creation is a happens-before edge, so the serve goroutine sees that write and `Run` never creates the server. The 27 races are real, but they come from tests that construct `grpcServer` directly and call `Run` without registering a service — a path an application does not take.

So this is a test-suite and latent-correctness fix rather than a live lost-shutdown fix. That does not make it less worth merging: `-race` CI is only useful if it is clean, a latent hazard one refactor away from being reachable is still a hazard, and the shape is now the right one. It does mean the "not test-only" framing in #3929 should be softened rather than repeated.

One follow-up, not for this PR. `AddGRPCServerOptions` (`grpc.go:63`), `AddGRPCUnaryInterceptors` (`:82`) and `AddGRPCServerStreamInterceptors` (`:92`) append without `srvMu`, while `ensureServer` appends to `g.options` under it (`:147`) and reads both interceptor slices under it (`:145-146`). A probe that calls one of those concurrently with `ensureServer` reports three races pointing at `grpc.go:147`. It needs a user configuring concurrently with registration, which sequential `main()` never does, so it is latent — but it is the same class of defect one field over, which is how #3929 came to exist in the first place. It is not tracked anywhere; could you raise it? The doc comment on `ensureServer` also says the hold "covers the append to `g.options`", which is true of its own append but reads as covering the public mutators too.

Checked and deliberately not raised: the lost-shutdown gap is #3801, and your test comment already cites it and says it is not what that test asserts, which is the right call. The one race left in a full `-race` run of the package is `TestContext_WriteMessageToSocket`, which reproduces on `development` and is #3930.

Approving. Thank you for the control-worthy regression tests — `TestGRPC_ConcurrentEnsureServer` pinning the build count through the options length is a better guard than the one I asked for.

PR #3932, opened 2026-08-14
fix(websocket): don't panic on a plain HTTP request to a WebSocket route

### LINKED ISSUE #3862: BUG: plain HTTP request to a WebSocket route panics (unchecked type assertion + nil deref)
## Bug description

A plain HTTP request to a route registered via `app.WebSocket(...)` reaches the WebSocket handler and can panic.

## How it happens

`App.WebSocket(route, handler)` registers the route as a normal GET route:

- https://github.com/gofr-dev/gofr/blob/development/pkg/gofr/websocket.go#L30-L45

Inside the generated handler:

```go
connID := ctx.Request.Context().Value(websocket.WSConnectionKey).(string)

conn := a.httpServer.ws.GetWebsocketConnection(connID)
if conn.Conn == nil {
    return nil, websocket.ErrorConnection
}
```

When a client sends a plain `GET /ws` (no `Connection: Upgrade`, no `Upgrade: websocket` headers), the `WSHandlerUpgrade` middleware does **not** add `WSConnectionKey` to the request context. The type assertion:

```go
ctx.Request.Context().Value(websocket.WSConnectionKey).(string)
```

then panics with `interface conversion: interface {} is nil, not string`.

Even if that assertion were safe, `GetWebsocketConnection("")` returns `nil`, and the subsequent `conn.Conn == nil` would dereference `nil`.

## Expected behavior

A non-WebSocket request to a WebSocket route should get a clean 4xx response (for example `400 Bad Request` or `websocket.ErrorConnection`) instead of a panic.

## Repro sketch

```go
app := gofr.New()

app.WebSocket("/ws", func(ctx *gofr.Context) (any, error) {
    return "ok", nil
})

// Plain GET without WebSocket upgrade headers.
// expected: clean error response
// actual: panic
```

## Suggested fix direction

In `App.WebSocket` handler, use a safe type assertion and nil checks before dereferencing:

```go
connID, ok := ctx.Request.Context().Value(websocket.WSConnectionKey).(string)
if !ok {
    return nil, websocket.ErrorConnection
}

conn := a.httpServer.ws.GetWebsocketConnection(connID)
if conn == nil || conn.Conn == nil {
    return nil, websocket.ErrorConnection
}
```

## Notes

- I did a quick search; I did not find an existing open issue/PR for this exact path.
- Related: #3213 / #3216 fixed `WriteMessageToSocket` nil panic, but this is a different code path (`App.WebSocket` route handler).

om7057:
## Summary

Fixes #3862.

`App.WebSocket` registered its handler with an unsafe type assertion on `WSConnectionKey`. A plain HTTP GET to a WebSocket route (no `Connection: Upgrade` / `Upgrade: websocket` headers) never gets that key set on the request context by the upgrade middleware, so the assertion `ctx.Request.Context().Value(websocket.WSConnectionKey).(string)` panicked with `interface conversion: interface {} is nil, not string`.

`GetWebsocketConnection` also returns a nil `*Connection` for an unknown connection ID, so even a safe type assertion would still hit a nil pointer dereference at the following `conn.Conn == nil` check.

Both call sites now check `ok`/`nil` before dereferencing and return the existing `websocket.ErrorConnection` instead of panicking, matching the fix direction suggested in the issue.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/gofr/...`
- [x] New test: `Test_WebSocket_PlainHTTPRequestDoesNotPanic` sends a plain `GET` to a route registered via `app.WebSocket` and asserts the response carries `websocket.ErrorConnection`'s message rather than the generic panic-recovery message (both map to HTTP 500, so the response body is the only way to distinguish "handled cleanly" from "panicked and got recovered"). Verified this test fails with the exact panic from the issue when the fix is reverted, and passes with it applied.


om7057:
@Umang01-hash @coolwednesday a small reminder on this!

aryanmehrotra (changes requested):
Sorry for the wait, and for missing your ping — thirteen days on a remotely-triggerable panic is too
long.

Bug is real, fix is right. Ran both builds in containers:

```
development   500  Internal Server Error   + panic: interface conversion: interface {} is nil, not string
this branch   500  couldn't establish connection to web socket   + no panics
```

Both guards are needed — `GetWebsocketConnection` does return a nil `*Connection` for an unknown ID.

**One change: the status code on the first return site.**

The issue asked for 4xx and this keeps the pre-existing 500. Not a regression, but RFC 6455 §4.2.1
says a failed handshake gets "an appropriate error code (such as 400 Bad Request)" — and we already
do that twelve lines away, in `middleware/web_socket.go:21`, when an upgrade is attempted and fails.
A request that never attempted one should not be treated as more serious than one that tried.

Only the first site though — the two are different failures:

```go
if !ok {
    return nil, websocket.ErrorConnection   // plain GET, no upgrade -> client error, 400
}
if conn == nil || conn.Conn == nil {
    return nil, websocket.ErrorConnection   // upgrade ran, connection missing -> server fault, keep 500
}
```

So please return a distinct error at the first site rather than putting `StatusCode()` on the shared
`ErrorConnection` — folding both into a 4xx would hide a genuine bug class. `ErrorConnection` has no
other references in the repo and only your new test pins a status, so it is a one-line assertion
change.

(426 looks tempting but is wrong here — RFC 6455 reserves it for version negotiation in §4.2.2.)

Also: this branch predates #4051 so it does not build standalone against current `development`.
Merges cleanly — I tested the merged result — just worth a rebase.

om7057:
> Sorry for the wait, and for missing your ping — thirteen days on a remotely-triggerable panic is too long.
> 
> Bug is real, fix is right. Ran both builds in containers:
> 
> ```
> development   500  Internal Server Error   + panic: interface conversion: interface {} is nil, not string
> this branch   500  couldn't establish connection to web socket   + no panics
> ```
> 
> Both guards are needed — `GetWebsocketConnection` does return a nil `*Connection` for an unknown ID.
> 
> **One change: the status code on the first return site.**
> 
> The issue asked for 4xx and this keeps the pre-existing 500. Not a regression, but RFC 6455 §4.2.1 says a failed handshake gets "an appropriate error code (such as 400 Bad Request)" — and we already do that twelve lines away, in `middleware/web_socket.go:21`, when an upgrade is attempted and fails. A request that never attempted one should not be treated as more serious than one that tried.
> 
> Only the first site though — the two are different failures:
> 
> ```go
> if !ok {
>     return nil, websocket.ErrorConnection   // plain GET, no upgrade -> client error, 400
> }
> if conn == nil || conn.Conn == nil {
>     return nil, websocket.ErrorConnection   // upgrade ran, connection missing -> server fault, keep 500
> }
> ```
> 
> So please return a distinct error at the first site rather than putting `StatusCode()` on the shared `ErrorConnection` — folding both into a 4xx would hide a genuine bug class. `ErrorConnection` has no other references in the repo and only your new test pins a status, so it is a one-line assertion change.
> 
> (426 looks tempting but is wrong here — RFC 6455 reserves it for version negotiation in §4.2.2.)
> 
> Also: this branch predates #4051 so it does not build standalone against current `development`. Merges cleanly — I tested the merged result — just worth a rebase.

Sure, no worries @aryanmehrotra.
Split the two error cases as suggested: added websocket.ErrorNotWebSocketUpgrade{} with StatusCode() int returning 400 for the "never attempted an upgrade" path, kept ErrorConnection as-is (falls through to 500) for the "upgrade ran, connection missing" path. Updated the test to assert the 400 and the new message.

Also rebased onto current development, builds standalone now.

aryanmehrotra (approved):
The split landed exactly as asked, and the comment on `ErrorNotWebSocketUpgrade` explaining why the two failures carry different status codes is better than what I wrote in the request. I rebuilt both trees and drove a real GoFr server over the wire rather than reading the diff.

## The fix, measured

Same app, same requests, `development` against this branch merged onto current `development`:

| request to a route registered with `app.WebSocket` | `development` | this PR |
|---|---|---|
| plain `GET`, no headers | `500 Internal Server Error` | **`400 request did not include a WebSocket upgrade`** |
| `Connection: Upgrade` only | `500` | `400` |
| `Upgrade: websocket` only | `500` | `400` |
| both headers, no `Sec-WebSocket-Key` | `400 Could not open WebSocket connection` | unchanged |
| **a real WebSocket upgrade** | echoes | **echoes** |
| 200 concurrent plain `GET`s | all `500`, no goroutine growth | all `400`, no goroutine growth |
| **panics in the log** | **203** | **0** |

The fourth row is the one I most wanted to see: the upgrade middleware's own failure path still answers `400` with its own message, so the new error sits alongside it rather than swallowing it. The fifth is the regression check — a genuine handshake still upgrades and echoes.

## The performance side, since you asked

This is bigger than I expected. Panic-and-recover is not a cheap error path:

| | `development` | this PR | |
|---|---|---|---|
| plain `GET`, 4000 requests @16 concurrent | 2,514 req/s · 398 µs/req | **42,623 req/s · 23 µs/req** | 17x |
| the same with logging suppressed | 2,335 req/s · 428 µs/req | **31,747 req/s · 31 µs/req** | 13.6x |

The second row matters for attribution: with `LOG_LEVEL=FATAL` the gap barely moves, so this is the cost of the panic machinery itself — the unwind and the recovery — not of writing the stack trace out.

Which reframes the bug slightly. On `development` an unauthenticated client that knows one WebSocket route can make each request cost the server roughly fourteen times what a normal error costs, and write a stack trace per request while doing it. That is worth a line in the description: this is not only a crash-safety fix, it closes a cheap amplification.

## Two small things

**The `conn == nil` guard has no test.** I mutated both halves of the fix:

| mutation | |
|---|---|
| revert to the unsafe type assertion (the original bug) | killed |
| `StatusCode()` 400 → 500 | killed |
| first site returns `ErrorConnection` instead of the new type | killed |
| **drop `conn == nil`, keep `conn.Conn == nil`** | **survived** |

The plain-HTTP test cannot reach that branch — it returns at the `!ok` check one line earlier. But it is reachable in production: `GetWebsocketConnection` is a bare map lookup (`websocket.go:168`), so a request whose `WSConnectionKey` names a connection that has since been removed gets a nil `*Connection` and the next dereference panics — the same failure, one line down. It was the half of the fix I confirmed was needed in the last round, and it is the half nothing pins.

A context value cannot cross a real HTTP connection, so this one has to be served in-process:

```go
func Test_WebSocket_StaleConnectionIDDoesNotPanic(t *testing.T) {
	testutil.NewServerConfigs(t)

	app := New()
	app.WebSocket("/ws-stale", func(*Context) (any, error) {
		return "unreachable: the connection for this ID is gone", nil
	})

	ctx := context.WithValue(t.Context(), gofrws.WSConnectionKey, "connection-that-no-longer-exists")

	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "/ws-stale", http.NoBody)
	require.NoError(t, err)

	rec := httptest.NewRecorder()
	app.httpServer.router.ServeHTTP(rec, req)

	assert.Equal(t, http.StatusInternalServerError, rec.Code,
		"a missing connection is a server-side fault, unlike a request that never attempted an upgrade")
	assert.Contains(t, rec.Body.String(), gofrws.ErrorConnection.Error(),
		"expected ErrorConnection, not a panic-recovery response")
}
```

It needs `gofrws "gofr.dev/pkg/gofr/websocket"` as an alias, since the file already imports gorilla's `websocket` under that name. Verified: it kills the surviving mutation, passes 50 consecutive runs, and lints clean.

**The new 400 still writes an `ERROR` log line per request.** Driving the 200 concurrent plain `GET`s at `LOG_LEVEL=DEBUG`:

```
205 ERROR      (206 requests)
```

So the amplification the status-code change closes on CPU is still open on the log: an unauthenticated client can fill the error log from outside. `pkg/gofr/http/errors.go` already answers this for exactly the same situation — `ErrorInvalidRoute` is a client error that carries

```go
func (ErrorInvalidRoute) LogLevel() logging.Level {
	return logging.INFO
}
```

because a 404 is client noise, not a server fault. A plain `GET` to a WebSocket route is the same kind of noise. Adding the same method to `ErrorNotWebSocketUpgrade` takes that run from **205 `ERROR` lines to 2**, which I verified. `ErrorConnection` should keep its default — a missing connection genuinely is a server fault and should stay loud.

Neither is worth another round on its own. If you would rather not add the second, say so — it is a judgement call about noise, not a defect.

## Breaking changes

None. I diffed `go doc ./pkg/gofr/websocket` between the two trees; the only difference is one addition:

```
> type ErrorNotWebSocketUpgrade struct{}
```

Nothing removed, nothing re-signed. The shape matches the repo's own convention exactly — `http.ErrorRequestTimeout` and `http.ErrorInvalidRoute` are the same `struct{}` with value-receiver `Error()` and `StatusCode()`.

The one behaviour change worth a release note is the intended one: a plain HTTP request to a WebSocket route now answers `400` where it answered `500`. Anyone alerting on 5xx from these routes will see that traffic move, which is the point, but it should not be a surprise.

## Not flagging

- **`Test_WebSocket_Success` fails under `-race`** — on this branch and identically on `development`, at `websocket.go:47`. Pre-existing, unrelated, and CI does not run `-race`.
- **`GetWebsocketConnection` takes the write lock for a read** while `ListConnections` uses `RLock` twelve lines away. Pre-existing and out of scope here.
- **The new type living in `pkg/gofr/websocket` rather than `pkg/gofr/http/errors.go`.** It belongs next to `ErrorConnection`, which it is defined against.
- **Keeping `ErrorConnection` as a sentinel `var` while the new one is a type.** The type needs methods; the var does not. Both patterns are already in the repo.

## Housekeeping

The branch is 16 commits behind `development`. It merges cleanly — everything above was measured on the merged result, and the merged tree builds, passes the WebSocket suites, and introduces no new lint in any of the three files it touches. Just needs the merge before it goes in.

Approving. Thanks for the turnaround on the split, and sorry again about the first delay.

---

## Adversarial pass: hostile inputs and mixed traffic

I went back and tried to break it rather than only to confirm it — hand-written requests over a raw socket, so malformed input reaches the server unmodified, plus real upgrades running alongside a flood of plain `GET`s. Same app, same requests, both trees:

| request | `development` | this PR |
|---|---|---|
| lowercase `connection: upgrade` / `upgrade: websocket` | `101` | `101` |
| mixed-case `Upgrade: WebSocket` | `101` | `101` |
| duplicate `Upgrade` headers (`websocket`, `h2c`) | `101` | `101` |
| garbage `Sec-WebSocket-Key` | `400` | `400` |
| `Sec-WebSocket-Version: 7` | `400` | `400` |
| **`HTTP/1.0` request** | **`500`** | **`400`** |
| **`GET` with a body** | **`500`** | **`400`** |
| **8 KB of junk headers** | **`500`** | **`400`** |
| **4 KB query string** | **`500`** | **`400`** |
| `POST` to the WebSocket route | `404` | `404` |

**Panics logged across the whole adversarial run: 3,607 on `development`, 0 on this branch.**

The rows that are identical matter as much as the ones that changed. Case-insensitive header matching still upgrades, the middleware's own rejections (bad key, wrong version) are untouched, and `POST` still falls through to `404` because the route is only registered for `GET` — the new error does not swallow any of them. Every row that moved is one where a request reached the handler without a `WSConnectionKey`, which is exactly the set this PR is meant to cover.

Two more, both clean:

- **Mixed traffic.** 40 concurrent real WebSocket clients alongside 400 concurrent plain `GET`s: **40/40 upgrades echoed and 400/400 plain `GET`s answered `400`**, goroutine growth `-1`. On `development` the same run gives 40/40 upgrades and 0/400 correct — the plain path is 500s — with the same goroutine count. So the plain-`GET` path never disturbed live connections on either tree; the bug was a panic and a cost, not corruption.
- **Sustained flood.** 3,000 sequential plain `GET`s: goroutine growth `+0` on both trees, `/ping` still `200` afterwards. No leak either way.

Nothing new to report from any of it — which is the useful result. The approval stands, and the two small items above are unchanged.

pkg/gofr/websocket.go
```diff
@@ -29,10 +29,16 @@ func (a *App) OverrideWebsocketUpgrader(wsUpgrader websocket.Upgrader) {
 // within the handler context. User can access the underlying WebSocket connection using `ctx.GetWebsocketConnection()`.
 func (a *App) WebSocket(route string, handler Handler) {
 	a.GET(route, func(ctx *Context) (any, error) {
-		connID := ctx.Request.Context().Value(websocket.WSConnectionKey).(string)
+		connID, ok := ctx.Request.Context().Value(websocket.WSConnectionKey).(string)
+		if !ok {
+			// The request never went through the WebSocket upgrade middleware
+			// (e.g. a plain HTTP GET with no Upgrade headers), so no
+			// WSConnectionKey was ever set on the context.
```
Umang01-hash, line 36: NIT (description drift): the PR body says the handler returns `websocket.ErrorConnection` and that "both map to HTTP 500", but the code returns `ErrorNotWebSocketUpgrade{}` → 400. The code and the test comment are correct and consistent; only the PR description is stale — please update it so the history matches what shipped.

Umang01-hash (approved):
LGTM. Reproduced #3862 myself: reverting both guards makes the new test panic with the exact `interface conversion: interface {} is nil, not string` at websocket.go:32, so the regression test genuinely exercises the buggy path. Traced the class — the only other `WSConnectionKey` reader (container.go:452/457) already uses the `,ok` form, so no sibling still panics. 400 is consistent with the middleware's own failed-upgrade 400 (web_socket.go:21) and RFC 6455 4.2.1. Additive/non-breaking; gofmt/build/vet/golangci and the new test under -race all green.

Two non-blocking notes:
1. NIT (inline): the PR description is stale vs the code — worth a quick edit before merge.
2. Heads-up (pre-existing, NOT this PR): `Test_WebSocket_Success` data-races under `-race` on the `ctx.Context = WithValue(...)` write at websocket.go:47 vs the `serveWithGoroutine` read — I reproduced it on `development` too, so it's out of scope here. Might be worth a separate issue.

Issue #3763, opened 2026-07-24
EnableRBAC fails open: unusable RBAC config silently disables authorization instead of stopping the app


akshat-kumar-singhal:
### Description

`App.EnableRBAC` **fails open**: when the RBAC config cannot be loaded (file missing, unreadable, malformed YAML/JSON, or failing validation), it logs an error and `return`s without installing the middleware. The application then starts and serves **every route with no authorization**, and the caller has no way to detect this because `EnableRBAC` returns `void`.

For an authorization layer this is a dangerous default: a misconfiguration silently disables access control rather than stopping the app.

https://github.com/gofr-dev/gofr/blob/main/pkg/gofr/rbac.go#L42-L53

```go
// Load configuration directly with dependencies
config, err := rbac.LoadPermissions(path, logger, metrics, tracer)
if err != nil {
    a.Logger().Errorf("Failed to load RBAC config: %v", err)
    return          // <-- no middleware installed; app serves ungated
}

a.Logger().Infof("Loaded RBAC config successfully")
middlewareFunc := rbac.Middleware(config)
a.UseMiddleware(middlewareFunc)
```

### Impact

We hit this in production. Our container image placed `rbac.yaml` at a path the running process didn't resolve; GoFr logged `Failed to load RBAC config: ...` and then served **all** routes with no role checks. An integration test later demonstrated the effect concretely: a `salesman`-role token successfully called an admin-only endpoint. The single log line was easy to miss amid startup output, and nothing in the API surface signals that RBAC is off.

### Expected behavior

A security middleware should **fail closed**. When configured but unable to load its policy, `EnableRBAC` should stop the application (e.g. `Fatalf`) or return an `error` so the caller can decide — not start ungated. Options, roughly in order of preference:

1. **Return an `error`** — `EnableRBAC(configPath ...string) error`. Most flexible; lets the caller `log.Fatal` or handle it. (API change.)
2. **`Fatalf` on load failure** — treat an unusable authz config like any other fatal boot dependency. (No signature change; safest default.)
3. At minimum, make the fail-open path opt-in and loud (a distinct `WARN`/`ERROR` that authorization is DISABLED), never the silent default.

### Steps to reproduce

1. `app.EnableRBAC("does-not-exist.yaml")` (or point it at a malformed file).
2. Start the app and hit any route that should require a permission.
3. Observe a single `Failed to load RBAC config` log line, then the request succeeds with no role/permission check.

### Version

[...]

Issue #2378, opened 2025-10-06
Add AddOptionWithValidation interface to support error reporting in service options


Umang01-hash:
**Is your feature request related to a problem? Please describe.**
Our current service.Options interface doesn't support error reporting. This creates a risk of silent failures when invalid configurations are used as options.

```
package service

type Options interface {
	AddOption(h HTTP) HTTP
}

```

When using **AddOption**, validation errors cannot be surfaced to the caller. Consumers must use separate validation methods or constructors, creating a disconnected validation flow.

**Describe the solution you'd like**

Create a new interface AddOptionWithValidation that returns an error:

```
type AddOptionWithValidation interface {
    AddOptionWithValidation(h HTTP) (HTTP, error)
}
```

Services can check for this interface before falling back to the standard Options interface, enabling proper error handling while maintaining backward compatibility.

aryanmehrotra:
this seems to be a breaking change, can you add some examples if not

Umang01-hash:
This is not a breaking change because we're adding a new interface alongside the existing one, not replacing it. Here's how it works:

### Backward Compatibility:
- The original `Options` interface remains unchanged.
- Existing code using `Options.AddOption()` continues to work exactly as before.
- The framework tries the new interface first, then falls back to the old one.

#### Implementation Example:

- **New introduced interface:**
```go
package service

type Options interface {
	AddOption(h HTTP) HTTP
}

// AddOptionWithValidation extends Options with error reporting capability
type AddOptionWithValidation interface {
	AddOptionWithValidation(h HTTP) (HTTP, error)
}
```

- New method **NewHTTPServiceWithValidation**:


```go
// NewHTTPServiceWithValidation provides error handling for options
func NewHTTPServiceWithValidation(serviceAddress string, logger Logger, metrics Metrics, options ...interface{}) (HTTP, error) {
	h := &httpService{
		Client:  &http.Client{},
		url:     serviceAddress,
		Tracer:  otel.Tracer("gofr-http-client"),
		Logger:  logger,
		Metrics: metrics,
	}

	var svc HTTP = h

	// Process options with validation support
	for _, o := range options {
		// Try validation interface first
		if validator, ok := o.(AddOptionWithValidation); ok {
			var err error
			svc, err = validator.AddOptionWithValidation(svc)
			if err != nil {
				return nil, err
			}
		} else if option, ok := o.(Options); ok {
			// Fall back to standard interface
			svc = option.AddOption(svc)
		} else {
[...]

PR #3934, opened 2026-08-15
fix(rbac): make wildcard-method rules reachable and resolve overlapping patterns deterministically (#3808)

### LINKED ISSUE #3808: RBAC: `"methods": ["*"]` endpoint never matches, and the middleware fails open — protected routes serve unauthenticated
### Summary

An RBAC endpoint declared with `"methods": ["*"]` **never matches any request**, and because
`rbac.Middleware` fails open on an unmatched route, the endpoint it was meant to protect serves
**unauthenticated** traffic. A route the operator believes is gated is silently ungated.

`"*"` is supported and documented (`matchesHTTPMethod` in `pkg/gofr/rbac/endpoint_matcher.go`
treats it as "all methods"), so this is a reasonable thing to write — it just doesn't work in the
code path the middleware actually uses.

Affects `gofr.dev v1.56.3` (latest at time of writing).

### Reproduction

`configs/rbac.json` — two entries with the *same* pattern shape, differing only in `methods`:

```json
{
  "jwtClaimPath": "role",
  "roles": [{ "name": "admin", "permissions": ["admin:read", "admin:write"] }],
  "endpoints": [
    { "path": "/admin/{path:.*}", "methods": ["*"],   "requiredPermissions": ["admin:read"] },
    { "path": "/team/{path:.*}",  "methods": ["GET"], "requiredPermissions": ["team:read"] }
  ]
}
```

```go
package main

import (
	"net/http"
	"net/http/httptest"
	"testing"

	"gofr.dev/pkg/gofr/rbac"
)

func TestWildcardMethodBypass(t *testing.T) {
	cfg, err := rbac.LoadPermissions("configs/rbac.json", nil, nil, nil)
	if err != nil {
		t.Fatal(err)
	}

	reached := false
	h := rbac.Middleware(cfg)(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {
		reached = true
	}))

	// Control: methods ["GET"] correctly blocks an unauthenticated request.
	rec := httptest.NewRecorder()
	h.ServeHTTP(rec, httptest.NewRequest(http.MethodGet, "/team/anything", nil))
	t.Logf("control  GET    /team/anything    -> status=%d reached=%v", rec.Code, reached)

	// Subject: methods ["*"] lets it straight through.
	rec, reached = httptest.NewRecorder(), false
	h.ServeHTTP(rec, httptest.NewRequest(http.MethodDelete, "/admin/orgs/123", nil))
	t.Logf("subject  DELETE /admin/orgs/123   -> status=%d reached=%v", rec.Code, reached)
	if reached {
		t.Error("BYPASS: unauthenticated request reached the handler")
	}
}
```

**Observed:**

```
control  GET    /team/anything    -> status=401 reached=false
subject  DELETE /admin/orgs/123   -> status=200 reached=true
--- FAIL: TestWildcardMethodBypass
```

**Expected:** the `/admin/**` request is rejected (401/403), like the control.

The same asymmetry is visible without the middleware, via
`cfg.GetEndpointPermission("GET", "/admin/orgs/123")` → `([], false)`, while
`GetEndpointPermission("GET", "/team/anything")` → `(["team:read"], false)`.

### Root cause

`buildEndpointKey` keys the lookup maps on the **declared** method
(`pkg/gofr/rbac/config.go:324`):

```go
func buildEndpointKey(endpoint *EndpointMapping, methodUpper string) string {
	pattern := endpoint.Path
	return fmt.Sprintf("%s:%s", methodUpper, pattern)
}
```

so `"methods": ["*"]` stores the key `*:/admin/{path:.*}`.

`matchesKey` then requires the key to begin with the **request's** method
(`pkg/gofr/rbac/config.go:471`):

```go
func (c *Config) matchesKey(key, methodUpper, path string) bool {
	if !strings.HasPrefix(key, methodUpper+":") {
		return false
	}
	...
}
```

`GET:` never prefixes `*:`, so the entry is unreachable. Both lookups the middleware performs go
through these maps — `getEndpointForRequest` → `getExactEndpoint` / `findEndpointByPattern`
(`config.go:411`, `:435`).

The bypass follows from `pkg/gofr/rbac/middleware.go:124`:

```go
if endpoint == nil {
    handler.ServeHTTP(w, r)
    return
}
```

Notably, the list-based `matchEndpoint` in `pkg/gofr/rbac/endpoint_matcher.go:33` handles `"*"`
**correctly** via `matchesHTTPMethod` — but `middleware.go:95` calls the map-based
`getEndpointForRequest` instead, so the correct implementation is bypassed.

### Suggested fix

Either:

1. Have `matchesKey` also accept keys stored under `*:` (and `getExactEndpoint` try the `*:` key as
   a fallback); or
2. Expand `"*"` at load time in `buildEndpointKey`'s caller into one key per supported HTTP method; or
3. Point the middleware at the list-based `matchEndpoint`, which already implements the intended
   semantics.

Option 2 keeps the O(1) exact-match property. Whichever is chosen, it seems worth considering
whether `endpoint == nil` should fail **closed** for configured path prefixes, so a future matcher
gap degrades to a 403 rather than an open route.

### Secondary observation (same area, lower severity)

`findEndpointByPattern` (`config.go:435`) and `checkPatternMatch` (`:450`) iterate `c.endpointMap` /
`c.endpointPermissionMap`, which are Go maps — `range` order is randomised. When two configured
patterns both match one request path (e.g. `/admin/{path:.*}` and `/admin/orgs/{org_id}`), the
winner therefore varies between processes, and so do the permissions enforced. A specificity
ordering (literal segment > `{param}` > `{path:.*}`, longest prefix first) would make the outcome
deterministic and independent of config ordering. Happy to file separately if you'd prefer.

akshat-kumar-singhal:
**Description:**

Fixes #3808.

An RBAC endpoint declared with `"methods": ["*"]` never matched any request. Because the middleware
passes through on an unmatched route, the endpoint it was meant to protect served unauthenticated
traffic — a route the operator believes is gated was silently ungated.

Two things beyond the original report turned up while confirming it, both verified against
`development`:

- **Omitting `methods` entirely has the same effect.** It defaults to `["*"]` at `config.go:296-298`,
  so an operator who simply doesn't write the field gets an ungated route.
- **The pattern-overlap nondeterminism is per request, not per process.** Go re-randomises map order
  on every `range`, so 50 identical requests in one process split 10×200 / 40×403 — a permissive rule
  intermittently shadows a strict one.

**Root Cause Analysis**

**Symptom.** With a config containing `{"path": "/admin/{path:.*}", "methods": ["*"],
"requiredPermissions": ["admin:read"]}`, `DELETE /admin/orgs/123` with no role returns **200** and
reaches the handler. The control — the same shape with `"methods": ["GET"]` — correctly returns 401.
`GetEndpointPermission("GET", "/admin/orgs/123")` likewise returns `([], false)`.

**Root cause.** `buildEndpointKey` (`config.go:324`) keys the lookup maps on the **declared** method,
storing `*:/admin/{path:.*}`. `matchesKey` (`config.go:471`) requires the key to begin with the
**request's** method via `strings.HasPrefix(key, methodUpper+":")`. `GET:` never prefixes `*:`, so
the entry is unreachable through either map lookup the middleware performs. `middleware.go:123` then
treats "no rule matched" as "nothing to enforce" and calls the next handler.

**Why it wasn't caught.** The package had two matchers. `matchEndpoint`
(`endpoint_matcher.go:33`) handles `"*"` correctly via `matchesHTTPMethod` and is the one the tests
exercise — but the middleware calls the map-based `getEndpointForRequest` instead. The correct
implementation was production-dead, and no test covered wildcard methods through the path that
actually runs.

**Fix.** Both entry points now resolve through a single ordered rule list, built once at load:

- Exact-path lookups stay O(1). The request's own method is probed before `*:`, so an explicitly
  declared method still wins over a wildcard on the fast path.
- Overlapping patterns resolve **most-specific-first** rather than by map iteration order: literal
  segment > constrained variable (`{id:[0-9]+}`) > free variable (`{id}`) > multi-segment catch-all,
  compared segment by segment, with an explicit method beating `["*"]` on a tie. Declaration order
  decides only where two patterns score identically — the docs say where that is rather than
  claiming order-independence the scoring does not deliver.
- `matchesKey`, `findEndpointByPattern`, `checkPatternMatch` and `checkExactMatch` are removed. The
  new rule type calls the existing `matchesHTTPMethod` / `matchesEndpointPattern` primitives rather
  than reimplementing them — a second copy of that logic is what caused this bug.
- A constraint mux cannot parse (`{id:[}`) passes the brace-balance check, loads, and then silently
  never matches — the same failure shape reached a different way. That is now logged at **error**
  level naming the pattern and stating the endpoint is not enforced. It does **not** fail the load:
  GoFr's position is to log and stay up on a config defect (#2378), and closing the hole properly
  needs the fail-closed default scoped into #3935.

Net −200/+80 lines in the package; the fix mostly deletes the duplicate matcher.

**Prevention.** Table-driven tests cover `"*"`, omitted and empty `methods`, custom verbs
(`PROPFIND`), lower-case declarations, most-specific-wins in both declaration orders, explicit method
over wildcard, and the 401/403/200 enforcement matrix through the middleware. The overlap tests run
100 iterations per case; they fail reliably against the old map-range code.

**On `"*"` and unknown methods**

`["*"]` deliberately covers methods GoFr does not know about (WebDAV verbs, for example). The
rejected alternative was expanding `"*"` at load into one key per known method, which keeps O(1)
lookup but lets custom verbs slip past, recreating this bug class. Because an entry states a
*requirement* rather than a *grant*, covering an unrecognised verb tightens enforcement rather than
relaxing it.

**On the overlap ordering**

Three orderings were considered. First-declared-wins (Envoy, Spring) is order-sensitive — reordering
the JSON silently changes who is authorized. Union-of-matching-rules (the Kubernetes model) does not
transfer: in K8s a matching rule *grants* permissions so unioning is the strict direction, whereas a
GoFr entry *imposes a requirement*, making union the most permissive option and letting a broad weak
entry cancel a narrow strict one. Most-specific-wins is the only one under which a narrow entry can
genuinely override a broad one, which is what an operator writing both intends. Happy to switch if
maintainers prefer a different rule — it's isolated to `compareSpecificity`.

**What this PR does not change**

`endpoint == nil` still passes through. Every normative source (Saltzer & Schroeder's fail-safe
defaults, OWASP ASVS V4.1.5, OWASP API5:2023) says an unmatched request should be denied, but the
middleware sees every route including ones intentionally absent from the config
(`/.well-known/health`, `/alive`, `/graphql/ui`), so a blanket flip would break working apps. That
belongs in its own change — an opt-in `defaultPolicy` with a warn-then-enforce bridge, the way Envoy
(`shadow_rules`), OPA (`dry-run`) and Kubernetes PSA (audit/warn/enforce) ship equivalent flips — and
it overlaps #3763, which owns `EnableRBAC`. Keeping it separate leaves this PR non-breaking.

**Requests that are served today will be rejected after this change.** This is the point of the fix,
but it is an operational break, not just a security improvement. Any application running a
`"methods": ["*"]` entry — or one that omits `methods`, which means the same — has an unenforced
route right now. Once the rule is reachable, traffic through that route starts returning 401/403,
including traffic from callers who never held the required permission and have been getting through
regardless. Operators should audit `["*"]` entries against real traffic before upgrading rather than
discovering the change through failed requests.

No API changes — every removed function was unexported and `GetEndpointPermission` keeps its
signature. Three behaviour changes worth calling out on release:

- Previously-unenforced `["*"]` rules are now enforced, as above.
- A config with a pattern mux cannot compile still loads and still never matches, but now logs an
  error line naming it. No behaviour change for a running app beyond the new log.
- Overlapping patterns now resolve deterministically by specificity. Anyone who was relying on the
  previous outcome was relying on map iteration order, which varied per request — but the effective
  permissions for such a config can change.

**Additional Information:**

No new dependencies. `mux.Route.GetError()` is existing gorilla/mux API.

The issue's secondary observation (map-range nondeterminism) is fixed here rather than filed
separately, since most-specific-wins *is* the overlap fix — the two cannot be cleanly separated.

**Follow-up filed as #3935.** This PR consolidates RBAC's two *internal* matchers into one, but RBAC
still keeps a second copy of the route table and re-matches every request against it, even though
the router has already resolved the request before the middleware runs —
`mux.CurrentRoute(r).GetPathTemplate()` is available and unused. That structural duplication is why
a divergence like #3808 was possible at all, and #3935 covers it along with the boot-time coverage
check that the deferred fail-closed work depends on.

Verified locally: `go build ./...`, `go test ./pkg/gofr/rbac/`, `go test -race -count=2
./pkg/gofr/rbac/`, `go test ./pkg/gofr/ -run RBAC`, and `golangci-lint run` clean on all changed
files.

**Review round (@aryanmehrotra, 2026-08-20)**

- **Specificity tie fixed.** `{id}` and `{id:[0-9]+}` scored equally, so `sort.SliceStable` fell
  through to declaration order — the exact order-sensitivity this PR rejects first-declared-wins for.
  A fourth level (`segConstrained`) sits between free variable and literal. Mutation-verified:
  removing the new case fails `constrained variable wins over free one`.
- **Residual ties documented, not claimed away.** Two identically-shaped patterns (`/{a}/{b}` vs
  `/{x}/{y}`) still fall back to declaration order, and `{path:[a-z/]+}` still scores as
  single-segment despite spanning segments. Both are now stated in the docs under *Where the
  ordering is not decided*.
- **Load-time rejection reverted** to a loud error log, as above.
- **`matchEndpoint` deleted.** Its eight single-case tests are one table-driven test against
  `config.resolve`.
- **`isCatchAllVariable`** now documents the consequence of the assumption, not just the assumption.
- **Docs**: the method-drop-out case in the worked example (`GET /admin/orgs/123` needs only
  `admin:read`, because the narrow rule is `["DELETE"]`-only and drops out on method), plus two new
  Troubleshooting entries.

**Review round 2 (@Umang01-hash, 2026-08-21)**

- **`/`-bearing constraints are scored honestly, not disclosed as a gap.** The previous round's
  disclosure was wrong about its own code: `pathSpecificity("/x/{p:[a-z/]+}")` returned three
  literal segments, not a single-segment variable, because the split ran before the brace check —
  so the loosest pattern in a config scored as the most specific one. Segments now split at brace
  depth zero (which also keeps `{id:[0-9]{2,3}}` intact), and any constraint containing `/` scores
  as a catch-all. That test is deliberately over-broad; it only ever moves a pattern *down* the
  ordering, so a rule can lose to a narrower one but never shadow one.
- **A duplicate `(method, path)` declaration now overwrites in full.** Pre-existing, reproduced
  against `development` first: `storeEndpointMapping` only ever *set* `publicEndpointsMap[key]`, so
  a public entry followed by a protected one for the same key served the route unauthenticated
  while reporting the protected endpoint. Both the public flag and the stale permission list are
  now cleared. Rejecting duplicates was the alternative and was rejected for the same reason the
  load-time failure was: it turns a config that loads today into a startup failure.
- **An equal-specificity tie prefers the protected rule.** Previously a public rule declared before
  an overlapping protected one shadowed it — deterministic, and deterministically open. Declaration
  order now decides only between rules of the same public-ness.
- **Ordering tests rewritten around cases that reach the ordered scan.** `narrow wins over broad`
  and `literal wins over param` both survived a `compareSpecificity` reduced to `len(a) - len(b)` —
  the first because the patterns differ in length, the second because a literal request path is
  answered by the exact-key fast path and never reaches the scan. Added same-depth cases
  (`/{scope}/orgs/{org_id}` against `/admin/orgs/{org_id}` and against `/admin/{section}/{org_id}`,
  more specific declared last) plus direct table tests on `pathSpecificity` and
  `compareSpecificity`, including the antisymmetry `sort.SliceStable` depends on. Every fix in this
  round was mutation-verified.
- **Docs**: the ordering section states the protected-beats-public tie-break and the conservative
  catch-all rule; a new section covers declaring the same path twice.

**For the release notes**, beyond "`["*"]` rules are now enforced": the level they are now enforced
*at* may be weaker than the permission name suggests. `{"methods": ["*"], "requiredPermissions":
["admin:read"]}` gates DELETE at `admin:read`. Operators should audit the permission, not just the
path.

**Overlap policy** — raised on #3808 for a maintainer call before this lands, per the review.

**Checklist:**


aryanmehrotra (changes requested):
Thanks for this — and for the write-up, the root-cause section made this a lot faster to verify.

**TL;DR:** the bug is real, I reproduced both halves of it against `development` before reading your fix, and the fix is the right shape. One real change needed around the ordering guarantee, plus one hunk I'd rather we drop. Everything else is approvable.

## What it does

Wildcard-method entries were keyed on the declared method (`*:/admin/{path:.*}`) while `matchesKey` gated on the *request's* method, so the rule was unreachable and the middleware's pass-through-on-no-match turned it into an open route. This consolidates both entry points onto one ordered rule list built at load, and resolves overlaps by specificity instead of map-range order.

I confirmed on `development`:

```
CONTROL  GET    /team/anything   -> 401  (["GET"] rule works)
SUBJECT  DELETE /admin/orgs/123  -> 200  reached handler   (["*"] rule never fires)
OMITTED  GET    /nomethod/x      -> 200  reached handler   (no methods field)
```

and the overlap flapping — 50 identical requests in one process: `map[200:42 403:8]`, different split every run. Your point that it's per-request rather than per-process is correct. On your branch all three return 401 and 400 identical overlapping requests give `map[403:200]`. Documented mux patterns all still validate, including the UUID one with nested braces.

## Bug — the ordering guarantee doesn't hold as documented

`pkg/gofr/rbac/resolver.go:99` — `pathSpecificity` scores a constrained variable and a free one identically, so `sort.SliceStable` falls through to declaration order:

```
score {id}        = [3 3 2]
score {id:[0-9]+} = [3 3 2]

declaration order [strict, loose] -> [users:strict]
declaration order [loose, strict] -> [users:loose]
```

Same config, entries swapped, different permission enforced. That contradicts the line you added to the docs — "independent of the order the entries appear in the config file" — and it's the same property you reject for first-declared-wins in the PR description ("reordering the JSON silently changes who is authorized"). It's a realistic config too: our own docs use both `{id:[0-9]+}` and `{resource}`.

Cheapest fix is another level, since a constrained variable genuinely constrains more than a free one:

```go
const (
	segCatchAll    = iota + 1 // {path:.*}
	segVariable               // {id}
	segConstrained            // {id:[0-9]+}
	segLiteral                // users
)
```

```go
case isCatchAllVariable(segment):
	scores = append(scores, segCatchAll)
case strings.Contains(segment, ":"):
	scores = append(scores, segConstrained)
default:
	scores = append(scores, segVariable)
```

Whatever residual tie is left after that, lets say so plainly in the docs rather than claim order-independence we don't have.

## On process

One thing for next time — #3808 has no maintainer comments on it, so the approach here was chosen and built in a single pass. You did lay out three options in the issue and explain why you picked one, which I appreciate, and the fix itself was never in question. But the overlap policy is exactly the kind of call to settle in the issue before there's a diff riding on it. It's much cheaper to change direction there than here — you've written `compareSpecificity`, the ordered rule list and a large block of table-driven tests around a decision we hadn't actually made yet.

Lets do that part in the issue before the next round.

## Lets keep the current validation behaviour

`pkg/gofr/rbac/endpoint_matcher.go:110` — I'd drop this one. Compiling the pattern at load turns a config that boots today into a startup failure, and our position has been to log and stay up rather than hard-fail (#2378). I don't want to change that as part of a bugfix.

Being explicit about what that costs, though, because it isn't free: without the compile check, a pattern mux can't parse still loads and then never matches, so the endpoint it was meant to govern stays unguarded — the same shape as the bug you're fixing here. So lets keep `validateMuxPattern` non-fatal but make it loud: log the pattern that won't compile at error level and carry on, rather than either aborting or staying silent. Genuinely closing that hole needs the fail-closed default you've already scoped into #3935, and it belongs there rather than widening this PR.

The overlap policy is a separate question and does still want a second maintainer's eyes before it lands — most-specific-wins vs first-declared vs union is framework shape, and it's load-bearing for anyone running RBAC.

Also: this is a breaking change for running apps, and the PR is honest about it. Anyone with a `["*"]` entry — or one omitting `methods` — has an unenforced route right now that starts returning 401/403 on upgrade. That needs to land in the release notes, not just the PR body. Worth adding that the newly-enforced level may be *weaker than the permission name suggests*: `{"methods": ["*"], "requiredPermissions": ["admin:read"]}` gates DELETE at `admin:read`, so operators should audit the permission, not just the path.

## Smaller things

`pkg/gofr/rbac/endpoint_matcher.go:36` — `matchEndpoint` is now only reachable from `endpoint_matcher_test.go`, and it rebuilds and sorts the whole rule list on every call instead of using the prebuilt `c.rules`, while skipping the exact-key fast path. Your own argument is that a second copy of this logic is what caused the bug; a second *entry point* with different behaviour keeps that risk alive. Can we delete it and point those tests at `config.resolve`?

`pkg/gofr/rbac/resolver.go:117` — `isCatchAllVariable` only recognises `.*` and `.+`. `{path:[a-z/]+}` matches `/files/a/b/c` but scores as single-segment, so a broad rule can outrank a narrower one. The comment documents the assumption; can we document the consequence too?

Docs — two gaps:
- The worked example doesn't cover the case operators will actually get wrong. With your own config, `DELETE /admin/orgs/123` requires `admin:write` but `GET /admin/orgs/123` requires `admin:read`, because the narrow rule is `["DELETE"]`-only and drops out on method. Two sentences would save someone assuming they'd protected the path.
- The new load-time rejection isn't mentioned anywhere, including Troubleshooting.

Good call keeping the fail-closed default out of this — that's a documented contract ("Unmatched Routes Behavior") and reversing it needs its own change with a migration path, the way you've scoped it in #3935.

One real change needed (the specificity tie), one to drop (the startup validation); the rest is small. Nice work on the root cause.

akshat-kumar-singhal:
Thanks — all three land, and the specificity one was a real hole. Addressed in `9e1582c`.

**Specificity tie.** You're right, and it's the property this PR argues for elsewhere, so it had to
go. Added `segConstrained` between free variable and literal exactly as you laid it out:

```
segCatchAll < segVariable < segConstrained < segLiteral
```

Mutation-verified rather than just asserted — removing the new `case` fails
`constrained variable wins over free one` with `/users/{id}`, in both declaration orders.

On the residual ties: I've stopped claiming order-independence and said where the ordering stops
deciding instead. Two cases, both now in the docs under *Where the ordering is not decided* —
identically-shaped patterns (`/{a}/{b}` vs `/{x}/{y}`), and the `isCatchAllVariable` gap you flagged
separately. Ranking either properly means parsing the constraint, which isn't worth the machinery
for a case the documented `.*`/`.+` forms avoid.

**Load-time validation.** Reverted to non-fatal, and I agree with the reasoning — #2378 is a
standing position and a bugfix isn't where to reverse it. The check moved to
`logUncompilablePattern`, which names the pattern at error level and says plainly what it costs:

```
RBAC: endpoint[2]: invalid mux pattern: "/api/{id:[}": ... This endpoint will never match a
request, so it is NOT enforced - any route it was meant to govern is currently unguarded.
```

Brace-balance validation keeps its existing fatal behaviour — only the step I added is downgraded.
The tests changed shape with it: the load now succeeds and the assertion is on the log line.

**`matchEndpoint`.** Deleted. You're right that a second entry point with different behaviour keeps
the risk alive — and it also skipped the fast path, so it wasn't even testing what runs. The eight
single-case tests are now one table-driven test against `config.resolve`.

**`isCatchAllVariable`.** Consequence documented alongside the assumption, in the code and the docs:
`{path:[a-z/]+}` spans segments but scores as one, so it can outrank a narrower pattern it fully
contains.

**Docs.** Both gaps covered. The worked example now calls out the case that actually bites —
`DELETE /admin/orgs/123` needs `admin:write`, but `GET /admin/orgs/123` needs only `admin:read`,
because the narrow rule is `["DELETE"]`-only and drops out on method before specificity is
considered. Two Troubleshooting entries added: the `invalid mux pattern` log line, and "the wrong
permission is being required on a path two entries match".

**Release notes.** Added your point to the PR body — the newly-enforced level may be weaker than the
permission name suggests, so the audit is of the permission, not just the path.

**On process.** Fair, and I'd rather have had the ordering question settled before writing
`compareSpecificity`. I've put the three options on #3808 with the argument for each so the choice
can be made there rather than inside a diff. Happy to hold this PR on that, or to swap the policy —
it's `compareSpecificity` and nothing else.

`go build ./...`, `go test -race -count=2 ./pkg/gofr/rbac/`, `go test ./pkg/gofr/ -run RBAC` and
`golangci-lint run` are clean; the three remaining `noctx` hits in `endpoint_matcher_test.go` are
pre-existing and untouched.

aryanmehrotra (dismissed):
All three addressed — thanks for the quick turnaround, and for taking the process point rather than
arguing it.

I re-verified against `9e1582c` rather than reading the diff, including end to end through a running
app with `EnableRBAC`, real routes and real requests:

**The original bug, before and after.**

```
                                        development          9e1582c
DELETE /admin/orgs/123   (no role)      204 handler reached  401
GET    /nomethod/thing   (no role)      200 REACHED-HANDLER  401
```

**Overlapping rules, 40 identical requests as a role holding only the weaker permission.**

```
development:  32 x 200,  8 x 403      (re-randomised per request)
9e1582c:      40 x 403, then 40 x 403 again
```

**The specificity tie I raised.** `{id}` scores `[4 4 2]`, `{id:[0-9]+}` scores `[4 4 3]`, and both
declaration orders now resolve to the constrained rule. Ranking is what decides it, not ordering.

**Non-fatal validation.** The app boots and says exactly what it costs:

```
ERROR  RBAC: endpoint[5]: invalid mux pattern: "/bad/{id:[}": ... This endpoint will never match
       a request, so it is NOT enforced - any route it was meant to govern is currently unguarded.
```

and the route it names serves unguarded, as documented. That's the behaviour we want.

`matchEndpoint` is gone, the broad rule still governs paths the narrow one doesn't match, and
`go test -race -count=2 ./pkg/gofr/rbac/` is clean.

Approving. On the ordering policy — thanks for putting the three options on #3808; I'll settle it
there. Approval doesn't presume the answer: if we land on something other than most-specific-wins
it's an isolated change to `compareSpecificity`, and I'd rather this fix stopped sitting on a live
bypass in the meantime.

aryanmehrotra:
Deliberately keeping this off the approval, since none of it is yours to fix here.

While testing this end to end I turned up three defects that predate the PR and are live on
`development` — filed together as #3979:

1. **`matchMuxPattern` races and grows the router without bound.** It calls `NewRoute()` on the
   shared `config.muxRouter` per rule per request, and gorilla/mux appends that route permanently:
   `0 -> 114 routes after 100 requests`, never shrinking, plus a confirmed `DATA RACE` under 50
   concurrent goroutines. Our suites miss it because they drive requests sequentially.
2. **533-3,734 allocations to authorize one request**, depending on config size — the cost of
   recompiling mux patterns per rule per request, on the hot path of every route.
3. **Multi-role JWTs authorize nothing.** A `["admin","viewer"]` claim becomes the string
   `"[admin viewer]"`, which matches no configured role, so every request gets `403` with nothing
   logged to explain it. That's the default shape from Keycloak, Auth0 and Entra.

1 and 2 are the same fix, and it's the one you already argued for in #3935 — reading
`mux.CurrentRoute(r).GetPathTemplate()` removes `matchMuxPattern` and takes the race, the growth and
most of the allocations with it. Your instinct there was right; this is the evidence for it.

Would you like to take #3979 in a separate PR? No obligation — say no and I'll find someone, or pick
it up myself. If you do take it, 3 needs a semantics decision before any code (union the permissions
of every named role, or reject arrays at load with a clear error), so worth settling that on the
issue first — same as we're doing for the ordering policy on #3808.

One correction to something I measured: I earlier had this branch looking slower than `development`
on the resolver benchmark. That comparison was unfair — my harness declared the matching rule last,
which under the new stable ordering forces a full scan while the old map order averaged half. I'm
not reporting a regression from it; the absolute allocation count in #3979 is the finding.

pkg/gofr/rbac/resolver.go
```diff
@@ -0,0 +1,202 @@
...
+
+	return scores
+}
+
+// isCatchAllVariable reports whether a "{name:regex}" segment can span multiple path
+// segments. Only the documented catch-all forms - ".*" and ".+" - are recognized as such;
+// any other constraint is assumed to stay within one segment.
+//
+// The assumption is not always true. A constraint that admits "/" itself, such as
+// "{path:[a-z/]+}", spans multiple segments but is scored as a single-segment variable, so a
```
Umang01-hash: This docstring is factually wrong about its own code. pathSpecificity("/x/{p:[a-z/]+}") actually returns [4,4,4], not a single-segment variable(2): strings.Split(pattern, "/") runs before the brace check, so the '/' inside the constraint shatters it into three pseudo-segments, each failing the brace test and scored literal(4). Net effect is the opposite of what the comment says — the rule scores as maximally specific and can outrank a genuinely narrow protected rule (a fail-open if it's public). Trigger needs a '/'-in-regex constraint (a documented anti-pattern), so not a blocker, but please fix the scoring or at minimum correct the comment.

pkg/gofr/rbac/resolver.go
```diff
@@ -0,0 +1,202 @@
...
+// enforcement rather than loosening it.
+func (r *endpointRule) matchesMethod(methodUpper string) bool {
+	return matchesHTTPMethod(methodUpper, []string{r.method})
+}
+
+// matches reports whether the rule covers the given request.
+func (r *endpointRule) matches(methodUpper, path string, config *Config) bool {
+	return r.matchesMethod(methodUpper) && matchesEndpointPattern(r.endpoint, path, config)
+}
+
```
Umang01-hash, line 54: "so that resolution does not depend on declaration order" over-claims — equal-specificity ties DO fall through sort.SliceStable to declaration order. The docs page discloses the tie-break correctly; just soften this code comment so the two agree.

pkg/gofr/rbac/resolver_test.go
```diff
@@ -0,0 +1,266 @@
...
+	constrained := EndpointMapping{Path: "/users/{id:[0-9]+}", Methods: []string{"*"}, RequiredPermissions: []string{"users:write"}}
+
+	testCases := []struct {
+		desc      string
+		endpoints []EndpointMapping
+		path      string
+		expected  string
+	}{
+		{"narrow wins over broad", []EndpointMapping{broad, narrow}, "/admin/orgs/123", "/admin/orgs/{org_id}"},
+		{"declaration order is irrelevant", []EndpointMapping{narrow, broad}, "/admin/orgs/123", "/admin/orgs/{org_id}"},
```
Umang01-hash, line 85: This case (and "narrow wins over broad" / "literal wins over param") doesn't actually pin path-specificity ordering — reverting compareSpecificity to  still passes 5 of 7 subcases (the constrained-variable-wins cases do cover it). Worth tightening these named cases so they'd go red if ordering breaks.

Umang01-hash (commented):
Reviewed deep + ran locally, verifying not trusting. The core fix is correct and valuable — I reproduced it both ways: on development, DELETE /admin/{path:.*} (methods ["*"]) returns perms=[] not-public, so the middleware endpoint==nil passthrough serves it unauthenticated; at this head it enforces [admin:read]. No breaking API change (signatures identical, resolver.go internal); build/vet/gofmt/-race clean; the wildcard/most-specific/explicit-beats-wildcard matrix passes.

Approve-leaning COMMENT — three small fixes worth making, then two pre-existing bypasses worth follow-up issues (not this PR's regressions). Inline below.

Also: I could NOT reproduce a fail-closed→fail-open regression for uncompilable patterns vs development — development's validateMuxPattern never mux-compiled ({id:[} loaded there too), so this PR is a net improvement (adds the error log). The fail-closed base only exists on an intermediate commit of this branch, not on development.

Pre-existing (follow-up, non-blocking):
- config.go: a duplicate (method, literal-path) declaration with a public entry before a protected one leaves publicEndpointsMap[key]=true while endpointMap[key] is the protected one, so getExactEndpoint returns isPublic=true → unauthenticated. getExactEndpoint/storeEndpointMapping are byte-identical to development, which probed the same fast path — so pre-existing. Worth clearing the public flag on a non-public overwrite, or rejecting duplicate keys.
- The equal-specificity tie-break resolves by declaration order, so a public rule declared before an overlapping protected one shadows it. Deterministic (better than the old random) and disclosed in the docs — fail-safe hardening would be protected-beats-public on a tie.

pkg/gofr/rbac/resolver.go
akshat-kumar-singhal: Confirmed by running it, and the numbers match yours: `pathSpecificity("/x/{p:[a-z/]+}")` returned `[4 4 4]`. `strings.Split` ran before the brace check, so the `/` inside the constraint shattered the segment into fragments that each failed the brace test and scored literal — the loosest pattern in the config scoring as the most specific one.

Fixed the scoring rather than the comment, in two parts:

- `splitPatternSegments` splits at brace depth zero, so a constraint is never cut in half. It also keeps `{id:[0-9]{2,3}}` intact, which the naive split broke the same way.
- `isCatchAllVariable` now also accepts any constraint containing `/`. `{path:[a-z/]+}` scores as a catch-all — least specific — so it loses to any narrower entry it overlaps.

The second half is deliberately over-broad: `{id:[0-9]+/}` is anchored to one segment but is still scored as a catch-all. Deciding it exactly means parsing the regex, and the approximation only ever moves a pattern *down* the ordering — it can lose to a narrower rule, never shadow one. That asymmetry is why it is safe, and it is now what both the docstring and the docs say instead of the old disclosure.

Mutation-verified: reverting the split fails `TestPathSpecificity/constraint admitting a slash spans segments` and `TestGetEndpointForRequest_SlashConstraintDoesNotOutrankNarrowRule`; reverting only the `/` case in `isCatchAllVariable` fails the first.

pkg/gofr/rbac/resolver.go
akshat-kumar-singhal: Right — softened. The comment now says the ordering makes a narrower rule govern even when a broader one is declared ahead of it, and that identically-scoring patterns keep declaration order, which agrees with the docs page.

One thing did change underneath it while addressing your last point below: an equal-specificity tie no longer resolves purely by declaration order — a rule that requires permissions now outranks a public one. Declaration order decides only between two rules of the same public-ness, which the comment states.

pkg/gofr/rbac/resolver_test.go
akshat-kumar-singhal: Checked this the way you did, and it is worse than the comment says — with `compareSpecificity` reduced to `return len(a) - len(b)`, 6 of the 7 subcases still pass. Two distinct reasons:

- `narrow wins over broad` passes because the patterns differ in *length* (`[4 4 2]` vs `[4 1]`), so length alone picks the right one. Nothing about segment-by-segment comparison is exercised.
- `literal wins over param` passes because a literal request path never reaches the ordered scan at all — `resolve` answers `/admin/orgs/global` from the exact-key fast path. That case documents the fast path, not the ordering.

Both are now backed by cases that do reach the scan: `/{scope}/orgs/{org_id}` against `/admin/orgs/{org_id}` (literal beats variable in the same position) and against `/admin/{section}/{org_id}` (the first differing segment decides). Same depth in each, more specific entry declared last, so a declaration-order fallback gets them wrong. Both go red under the length-only mutation.

Added direct table tests too, since the property is easier to pin at that level than through a request: `TestPathSpecificity` over each segment kind, and `TestCompareSpecificity` over each ordering relation plus the antisymmetry `sort.SliceStable` depends on.

akshat-kumar-singhal:
Thanks for reproducing it both ways rather than reading it — and for checking the fail-closed→fail-open question against `development` instead of against the intermediate commit. Agreed on that finding: `validateMuxPattern` never mux-compiled on `development`, so `{id:[}` loaded there too and the error log is a net addition.

The three inline points are addressed (replies on each thread). Short version: the `pathSpecificity` docstring was wrong about its own code exactly as you measured, so the *scoring* is fixed rather than the comment — segments now split at brace depth zero and a constraint containing `/` scores as a catch-all; the over-claiming comment is softened; and the ordering tests are rewritten around cases that actually reach the ordered scan.

**Both "pre-existing" items are fixed here rather than deferred.** I reproduced each first, and both returned `isPublic=true` before the change:

- **Duplicate `(method, path)` with a public entry first.** `storeEndpointMapping` wrote `endpointMap[key]` unconditionally but only ever *set* `publicEndpointsMap[key]`. The two maps ended up describing different declarations, so `getExactEndpoint` returned the protected endpoint together with `isPublic=true`. Fixed at the write: the non-public branch clears the public flag, and the public branch clears any stale permission list, so whichever entry is declared last is the only one that survives. I chose last-wins over rejecting duplicate keys because rejecting turns a config that loads today into a startup failure, which is the same reversal you pushed back on for uncompilable patterns. Last-wins is now stated in the docs rather than being incidental.
- **Equal-specificity tie resolving to a public rule.** Taking the fail-safe hardening you suggested: a tie now prefers the rule that *requires permissions*. A tie is a config that expressed no intent either way, and of the two ways to be wrong about an unintended overlap, enforcing is the recoverable one — a 403 is visible and fixable, an unauthenticated read is neither. Declaration order now decides only between two rules of the same public-ness.

Every fix was mutation-verified — reverted in place, suite re-run, named test goes red:

| reverted | fails |
|---|---|
| brace-aware split | `TestPathSpecificity/constraint admitting a slash spans segments`, `SlashConstraintDoesNotOutrankNarrowRule` |
| `/` case in `isCatchAllVariable` | `TestPathSpecificity/constraint admitting a slash spans segments` |
| public-flag delete | `DuplicateDeclarationLastWins/protected declared last is enforced` |
| public/protected tie-break | `ProtectedBeatsPublicOnTie/public declared first` |
| `compareSpecificity` → `len(a)-len(b)` | `TestCompareSpecificity` (3 cases), `MostSpecificWins` (3 cases) |

`go build ./...`, `go test -race -count=2 ./pkg/gofr/rbac/`, `go test ./pkg/gofr/ -run RBAC`, `gofmt` all clean. `golangci-lint` on the package reports only 21 pre-existing `noctx` findings (`httptest.NewRequest`), identical on `development` — 4 in `endpoint_matcher_test.go`, 17 in `middleware_test.go`. It surfaces 3 at a time because of `max-same-issues`, which is why they read as fewer than they are. Left alone: a whole-file sweep does not belong in this PR.

Unchanged: `endpoint == nil` still passes through. These two fixes narrow the fail-open surface where a rule *does* match; they do not change what happens when none does, which stays scoped to #3935 with the fail-closed default.

aryanmehrotra (dismissed):
Re-reviewed at `9503cb0e`, and verified end to end rather than by reading the diff — the app built
from this branch, running in an isolated container, driven over HTTP.

**Every claim in the last two commits holds.**

```
--- the original #3808 bypass ---
DELETE /admin/orgs/123   no role  -> 401   wildcard ["*"] rule enforced
GET    /nomethod/thing   no role  -> 401   omitted methods enforced
--- explicit method beats wildcard; broad governs where narrow does not match ---
GET    /admin/orgs/123   viewer   -> 403   narrow GET rule needs admin:superuser
GET    /admin/orgs/123   root     -> 200
GET    /admin/settings   viewer   -> 200   broad rule, admin:read
--- overlap ---
GET    /team/orgs/123    viewer   -> 403   narrow strict rule wins
--- constrained variable outranks free ---
GET    /api/users/123    viewer   -> 403   users:strict, not users:loose
--- the two fail-open paths closed in 81ee16b5 ---
GET    /dup/panel        no role  -> 401   public-then-protected duplicate
GET    /tie/ab/cd        no role  -> 401   tie prefers the rule that requires permissions
--- non-fatal validation ---
GET    /bad/xyz          no role  -> 200   unguarded, exactly as the log says
```

**Determinism under load:** 300 sequential requests to the overlapping route returned `300x 403`, and
300 concurrent at 20-way parallelism returned `300x 403`. No flapping under contention, which is the
failure mode that opened this issue.

**On the `pathSpecificity` fix** — you diagnosed that more accurately than my review did. I reported
`{p:[a-z/]+}` as scoring like a single-segment variable. It was worse: splitting on `/` before
checking braces shattered it into fragments that each scored literal, so on the previous head it came
out `[4 4 4]` — identical to `/files/a/b`, three literals, the top of the ordering. The loosest
pattern in a config was ranking as the most specific one. Now `[4 1]`. Good catch, and the
conservative over-broad test is the right call since it can only ever move a rule down.

Both fail-open paths reproduce on the previous head as `200 + handler reached`, so those were real.

`-race -count=2` clean, build clean, and my earlier findings all still hold.

**One process note, not a blocker.** The review asking for those two bypasses as follow-up issues was
a reasonable call and you fixed them here instead. I am not asking you to unpick it — they are genuine
fail-open bugs in the two hundred lines you were already rewriting, and splitting them would mean
knowingly shipping broken code. But this PR has now grown from +485/-202 to +889/-303 across three
review rounds, and every round has found something new, which means neither reviewer has ever
approved the head that eventually merges.

So: **lets freeze this one.** Anything further found in RBAC goes to #4047 or a new issue rather than
another commit here. #4047 is stacked behind it and cannot be reviewed in isolation until this lands.

Approving. Thanks for the mutation-verification on each fix — reverting a change to confirm the test
actually catches it is the thing that makes this reviewable at all.

pkg/gofr/rbac/resolver.go
```diff
@@ -0,0 +1,239 @@
...
+		rules = append(rules, byKey[key])
+	}
+
+	sort.SliceStable(rules, func(i, j int) bool {
+		if cmp := compareSpecificity(rules[i].pathScore, rules[j].pathScore); cmp != 0 {
+			return cmp > 0
+		}
+
+		if rules[i].methodScore != rules[j].methodScore {
+			return rules[i].methodScore > rules[j].methodScore
```
Umang01-hash, line 110: This comparator isn't actually covered. `ExplicitMethodBeatsWildcard` hits the exact-map fast path, so it never reaches here — I reverted this line to `return false` and the whole rbac suite stayed green. Could you add a case with two rules on a pattern path, e.g. `/admin/{id}` with `[*]` and `[DELETE]`, then `DELETE /admin/42` asserting the DELETE rule wins? That locks it.

Umang01-hash (changes requested):
Logic looks solid — determinism, wildcard reachability, and specific-deny-beating-wildcard all check out. One test gap holding it up, noted inline.

pkg/gofr/rbac/resolver.go
akshat-kumar-singhal: Confirmed, and thanks for doing the revert — the test was checking the fast path, not the comparator.
`/admin/reports` is a literal, so `Config.resolve` answers it from `getExactEndpoint("DELETE:/admin/reports")`
and the ordered scan is never entered. The `methodScore` branch had no coverage at all.

Fixed in 56456e2a (test-only). The test is now table-driven over both a literal and a pattern path in
both declaration orders, using the case you suggested:

- `/admin/{id}` with `["*"]` declared **first**, then `["DELETE"]`; `DELETE /admin/42` must resolve to
  the DELETE rule. A pattern path has no exact key, so it reaches the scan, and wildcard-first means
  `sort.SliceStable` alone cannot produce the right answer — the comparator has to.
- Plus `GET /admin/42`, asserting the wildcard rule still governs a method the explicit rule does not
  cover, so the tie-break does not shadow the broad rule where the narrow one does not apply.

Mutation-verified against your exact revert: with `return false` in place of
`return rules[i].methodScore > rules[j].methodScore`, only `pattern path, wildcard declared first`
fails (`admin:read` instead of `admin:write`); the other three cases stay green, which is your finding
restated as a test.

`go test -race -count=2 ./pkg/gofr/rbac/` clean.

One note for the freeze @aryanmehrotra called: `golangci-lint` flags 3 `noctx` hits in
`endpoint_matcher_test.go` (lines 418–449, `httptest.NewRequest` → `NewRequestWithContext`). They are
pre-existing on `development` and unrelated to this change, so I have left them rather than adding a
fourth round here.

pkg/gofr/rbac/resolver.go
```diff
@@ -0,0 +1,239 @@
...
+// matches reports whether the rule covers the given request.
+func (r *endpointRule) matches(methodUpper, path string, config *Config) bool {
+	return r.matchesMethod(methodUpper) && matchesEndpointPattern(r.endpoint, path, config)
+}
+
+// buildEndpointRules expands endpoints into one rule per declared method and orders them
+// most-specific-first, so that a narrower rule governs a request even when a broader one is
+// declared ahead of it. Rules that score identically - two patterns of the same shape, such as
+// "/{a}/{b}" and "/{x}/{y}" - keep their declaration order, which is the one case where the
+// order entries are written in still decides the outcome.
```
aryanmehrotra: This is the sentence to correct. The public-vs-protected step at `:113-116` sits ahead of declaration order, so identically-scoring rules do not simply "keep their declaration order" — a rule that requires permissions outranks a public one first, and declaration order decides only between rules of the same public-ness. The comparator's own comment and the docs page say this correctly; only this docstring still describes the earlier version of the code.

aryanmehrotra (dismissed):
Re-reviewed at `dd3aa843`. Approving — one small docstring correction to push, details at the
bottom, nothing that changes behaviour.

The delta since my last approval is one test-only commit plus four development merges, and I
checked that rather than assuming it: the PR's contribution against its merge-base is still
exactly the same 7 files (+922/-303), and nothing under `pkg/gofr/rbac/` changed except
`resolver_test.go`. So the end-to-end verification from the last round stands on unchanged
production code.

## The test gap is genuinely closed

Confirmed the way it was found — by reverting the line, not by reading the test. With
`return rules[i].methodScore > rules[j].methodScore` replaced by `return false`, the only failure
in the whole rbac suite is:

```
--- FAIL: TestGetEndpointForRequest_ExplicitMethodBeatsWildcard/pattern_path,_wildcard_declared_first
```

The other three cases and the "wildcard still governs an uncovered method" case stay green. That
is the finding restated as a test, which is what was asked for. Keeping the literal-path cases is
right even though they cannot catch the mutant — they pin the fast path, which is a separate
guarantee worth having.

## End to end, and against development as a control

Ran the same probe suite against a running app on both branches, so the probes are demonstrably
not vacuous. This branch:

```
DELETE /wild/orgs/123   no role  -> 401   wildcard ["*"] rule enforced (#3808)
GET    /nomethod/thing  no role  -> 401   omitted methods enforced
DELETE /wild/orgs/123   viewer   -> 204   handler reached, rule grants
DELETE /admin/42        viewer   -> 403   explicit DELETE rule wins on a pattern path
DELETE /admin/42        root     -> 204   handler reached
GET    /admin/42        viewer   -> 200   wildcard still governs the method DELETE does not cover
GET    /team/orgs/9     viewer   -> 403   narrow strict rule wins over broad
GET    /other/orgs/9    viewer   -> 200   broad rule governs where narrow does not match
GET    /dup/panel       no role  -> 401   public-then-protected duplicate stays closed
GET    /bad/xyz         no role  -> 200   unguarded, exactly as the error log says

300 sequential and 200 concurrent requests to the overlapping route: 403 every time.
```

The same binary against `development`:

```
DELETE /wild/orgs/123   no role  -> 204   handler reached      <- the #3808 bypass, live
GET    /nomethod/thing  no role  -> 200   handler reached
GET    /team/orgs/9     viewer   -> 200                        <- loose rule shadows the strict one
GET    /dup/panel       no role  -> 200   unauthenticated

300 sequential: map[200:268 403:32]
200 concurrent: map[200:176 403:24]
```

Four of the ten probes fail there and the overlap flaps, mostly in the permissive direction. That
is the case for this change stated as measurements.

## Two notes that are not blockers

**The race in #3979 is reproducible through a running app, and it is not yours.** Running the
probe suite under `-race` reports data races in `matchMuxPattern` -> `mux.(*Router).NewRoute` on
*both* branches — three here, four on `development` — so this PR neither introduces nor widens it,
and authorization outcomes stayed correct throughout. #4047 is the fix and is already open.

**The three `noctx` lint hits are pre-existing.** The call sites blame to `bdc3eb67` (Dec 2025) and
are unchanged on `development`, so leaving them was the right call; a fourth round for those would
not have been worth it.

`go test -race -count=2 ./pkg/gofr/rbac/`, `go test ./pkg/gofr/ -run RBAC` and `go vet` are clean.
The one failure in `./pkg/gofr/...` is the metrics-port probe, which fails identically on
`development` on this machine because the port is held locally — environment, not this change.

## One thing to fix

`pkg/gofr/rbac/resolver.go:57-59` — the `buildEndpointRules` docstring still says identically
scoring rules "keep their declaration order, which is the one case where the order entries are
written in still decides the outcome". The public-vs-protected preference added in `81ee16b5` sits
ahead of that, so declaration order is not quite the only remaining decider. The comment at the
comparator itself and the docs page both state it correctly, so nothing misleads a user reading
the docs — but this is the third time a comment in this file has described the code as it was
rather than as it is, and that is worth closing out rather than carrying. A sentence is enough.

Approving on the substance. Worth flagging that the standing changes-requested review still gates
the merge, and I think it can be re-approved now — the point it was raised for is closed and
verified independently here.

akshat-kumar-singhal:
@Umang01-hash — the comparator-coverage point you raised is closed in `56456e2a`.

`ExplicitMethodBeatsWildcard` was using a literal path, which `Config.resolve` answers from the exact-key fast path, so the `methodScore` comparison in `buildEndpointRules` was never reached — exactly as you found by reverting it to `return false`. The test is now table-driven over both a literal and a pattern path in both declaration orders, and includes the case you asked for: `/admin/{id}` with `["*"]` and `["DELETE"]`, `DELETE /admin/42` asserting the `DELETE` rule wins. That pattern case has no exact key, so it reaches the ordered scan, and declaring the wildcard first means a stable sort alone cannot produce the right answer — it is the one of the four that goes red under the reverted comparator. It also asserts the wildcard rule still governs `GET /admin/42`, a method the explicit rule does not cover.

Separately, `7badc52d` corrects the last inaccurate comment @aryanmehrotra flagged on the approval: the `buildEndpointRules` docstring still said identically-scoring rules "keep their declaration order", which the public-vs-protected step ahead of it made false. The docstring now states the full order — path specificity, then method (explicit over `*`), then enforcement (protected over public), with declaration order deciding only between rules alike on all three.

`go build ./...`, `go test -race -count=2 ./pkg/gofr/rbac/` and `gofmt` are clean. Could you take another look when you get a chance? The changes-requested review is the only thing gating the merge now.

aryanmehrotra (approved):
Re-reviewed at `7e45fc04b`. Approving.

The delta since my last approval is one docstring correction (`7badc52d0`) addressing the comment on `buildEndpointRules` tie-breaking, plus a clean merge of `development` (`7e45fc04b`). No production RBAC logic changed.

- The `buildEndpointRules` docstring in `pkg/gofr/rbac/resolver.go` now accurately describes the three-tiered tie-break sequence (path specificity -> explicit method over wildcard -> protected over public -> declaration order).
- `go test -race -count=2 ./pkg/gofr/rbac/`, `go test ./pkg/gofr/ -run RBAC`, and `go vet` are clean.
- All GitHub Actions CI checks are passing.

Ran the live app probe suite against this head:
- All 10 authorization probes verified end-to-end against a running GoFr server (wildcard enforcement, omitted methods, explicit method beats wildcard, narrow strict wins over broad, duplicate declaration overwrite, uncompilable pattern non-fatal log).
- 300 sequential requests and 200 concurrent requests (20-way parallelism) to the overlapping route returned 403 every time (`map[403:300]`, `map[403:200]`), confirming deterministic resolution under load.

Ready to merge.

Umang01-hash (approved):
LGTM. Reproduced the bug through the real exported API first: on development, GetEndpointPermission("DELETE", "/admin/orgs/123") for a rule with methods:["*"] returns ([], false) — unreachable, so the middleware fails open and serves the route unauthenticated. On this branch it returns ([admin:read], false) — reachable and enforced. Nice catch that omitting methods (defaults to ["*"]) hits the same path.

The overlap resolution is the part that matters for a security change, and the direction is safe: most-specific-first, explicit method beats "*", and on a genuine tie an enforcing rule beats a public one — every tiebreak leans toward more enforcement. Collapsing both entry points onto one ordered resolve (instead of two matchers, one of which was production-dead) is the right root-cause fix. I couldn't construct a config/request where a permissive rule shadows a stricter overlapping one, and the exact O(1) fast-path stays consistent with the ordered scan.

Verified locally: no exported-signature changes, full suite green under -race, golangci-lint (new-from-rev) 0 issues. Tests are non-vacuous — flipping compareSpecificity turns the most-specific-wins test red, and the tie tests resolve 100x so they'd actually catch the old per-request map-order flakiness.

One thing worth calling out for readers, not a blocker: routes with no matching rule (and patterns mux can't compile, like {id:[}) still fail open. That's pre-existing — middleware.go isn't touched here — and you've handled it well: loud ERROR log naming the unguarded pattern, and it's deferred to #3935 for the fail-closed default. Agreed that flipping that posture isn't a bugfix's job.

Issue #3872, opened 2026-08-09
CI: Dependabot cannot see node-version/go-version in setup-* steps, which is how #3870 happened


akshat-kumar-singhal:
### Summary

Dependabot's `github-actions` manager updates the `uses:` action reference but **not** the `with:` inputs that select a language runtime. So `node-version:`, `go-version:`, and `python-version:` values are invisible to our current dependency automation and can silently go end-of-life.

This is not hypothetical — it is exactly how #3870 happened, and this repo contains the proof.

### The evidence

Before #3871, `website-prod.yml` and `website-stage.yml` looked like this:

```yaml
- name: Setup Node.js
  uses: actions/setup-node@v7      # <- current major. Dependabot kept this fresh.
  with:
    node-version: 18.x             # <- EOL 2025-04-30. Dependabot never touched it.
```

Two lines apart. Dependabot had been doing its job on the line above for years while the line below rotted past end-of-life. Every action in the repo is on a current major — `actions/checkout@v7`, `actions/setup-go@v7`, `actions/setup-node@v7`, `actions/upload-artifact@v7`, `actions/download-artifact@v8` — which is what makes this a clean natural experiment rather than a guess: the manager was demonstrably running and demonstrably could not see the runtime version.

#3871 fixed the *values* (18.x → 24.x). It did nothing about the *mechanism*, so the same drift will recur on the same schedule.

### Why Renovate is the relevant difference

Per [Renovate's `github-actions` manager docs](https://docs.renovatebot.com/modules/manager/github-actions/):

> Renovate supports updating the `with` version for `actions/setup-go`, `actions/setup-node`, and `actions/setup-python`, although not all syntaxes are supported out of the box.
>
> Renovate will use `npm`-style semver versioning for `go` and `python`, and Renovate's built-in `node` versioning for updating `node`.

Dependabot has no equivalent. As far as I can find there is no Dependabot configuration that closes this — it is a capability gap in the manager, not a setting we have wrong.

### Options

1. **Adopt Renovate** for the `github-actions` manager (either alongside Dependabot for gomod, or replacing it wholesale). Closes this properly and also covers the Docker gap in the companion issue.
2. **A CI guard instead** — a small scheduled job that reads `nodejs/Release`'s `schedule.json` and fails if any `node-version:` in `.github/workflows/` is past its EOL date. Cheaper and no new bot, but bespoke and only covers Node.
[...]

Issue #3869, opened 2026-08-08
CI: submodule test worker runs 'go mod tidy', making the run non-hermetic


akshat-kumar-singhal:
### Summary

The `Submodule-Unit-Testing` worker mutates each submodule's `go.mod`/`go.sum` before testing it (`.github/workflows/go.yml:369-374`):

```bash
# Download dependencies for the submodule
go mod download
go mod tidy

# Run tests with a focus on failed tests first
go test ./... -v -short -coverprofile="${module_name}.cov" -coverpkg=./...
```

### Why this is a problem

**It tests something other than what is committed.** `go mod tidy` can add missing requirements and drop unused ones, so the dependency graph the tests run against is whatever tidy resolves at that moment, not what the repo declares. A submodule whose committed `go.mod` is missing a requirement passes CI and fails for anyone who clones it — CI has quietly repaired the exact defect it should be reporting.

**It hides untidiness rather than reporting it.** There is no `git diff --exit-code` after the tidy, so a `go.mod` that is out of date is silently corrected in the runner and the correction is thrown away with the runner.

**It can reach the network mid-test-run and change resolution between legs.** `tidy` may fetch modules `download` didn't, so the three matrix legs are not guaranteed to resolve identically.

### Suggested fix

Drop the `go mod tidy` line. `go mod download` already populates the module cache, and `go test` fails loudly on a genuinely incomplete `go.mod` — which is the signal that is currently being suppressed.

If tidiness should be enforced (worth doing separately, and arguably belongs in `code_quality` rather than the test job):

```bash
go mod tidy -diff   # non-zero exit if tidy would change anything, no mutation
```

`-diff` has been available since Go 1.23 and the repo is on 1.26, so the older `cp go.mod go.mod.bak && go mod tidy && diff` dance isn't needed.

### Expect fallout

If any submodule's committed `go.mod` is currently untidy, removing the tidy will surface it as a build failure rather than a silent repair. That is the point, but as with #3824 it means the fix and its fallout should land together rather than leaving the job red.

### Context

Found while reviewing [run 31266842627](https://github.com/gofr-dev/gofr/actions/runs/31266842627).

Issue #3868, opened 2026-08-08
CI: Example-Unit-Testing service dependencies are unpinned and unwaited (Zipkin :latest, no health checks, bitnamilegacy Kafka)


akshat-kumar-singhal:
### Summary

`Example-Unit-Testing` depends on five backing services. Two of them (MinIO, and MinIO alone properly) are pinned and waited on; the other four are neither. This is a plausible contributor to the intermittent failures that the `nick-fields/retry` wrapper currently absorbs.

### 1. Zipkin is `:latest` with no readiness wait

```yaml
- name: Start Zipkin
  run: docker run -d -p 2005:9411 openzipkin/zipkin:latest
```

`go.yml:98-99`. Compare the MinIO step immediately below it (`go.yml:104-123`), which is digest-pinned and polls `/minio/health/live` for 60s with an explicit `::error::` if it never comes up — with a comment explaining exactly why that matters. Zipkin gets neither: the step returns as soon as `docker run -d` has *started* the container, and `:latest` means the image can change under the repo without any commit.

The same treatment the MinIO step already has should apply here — a digest pin and a poll on Zipkin's `/health`.

### 2. The `services:` containers declare no health checks

Kafka, Redis and MySQL (`go.yml:39-72`) are all declared without `options: --health-cmd/--health-interval/--health-retries`. Without those, the runner waits for the container to be created and started, not for the service inside it to be accepting connections. MySQL is the one that takes meaningfully long to become ready after start.

```yaml
mysql:
  image: mysql:8.2.0
  ports: ["2001:3306"]
  env: { MYSQL_ROOT_PASSWORD: "password", MYSQL_DATABASE: "test" }
  options: >-
    --health-cmd="mysqladmin ping -h 127.0.0.1 -ppassword"
    --health-interval=10s --health-timeout=5s --health-retries=5
```

To be clear about the evidence: I have **not** traced a specific CI failure to a service not being ready — the retry wrapper makes that hard to see, since a first-attempt failure is only visible as a `##[warning]Attempt 1 failed` line. This is a known-unsound pattern rather than a diagnosed flake, and adding health checks is also how you'd find out whether it is one.

### 3. `bitnamilegacy/kafka:3.4.1` is a frozen archive

```yaml
kafka:
  image: bitnamilegacy/kafka:3.4.1
```

Verified against Docker Hub (`hub.docker.com/v2/repositories/bitnamilegacy/kafka`, last updated 2025-08-28), where the description reads:

[...]

Issue #3866, opened 2026-08-08
CI: submodule coverage is uploaded three times and merged nowhere; parse_coverage computes a number nothing reads


akshat-kumar-singhal:
### Summary

Since #3824 was fixed (PR #3825), `Submodule-Unit-Testing` does produce `submodule-coverage-reports`. Two problems remain downstream of that, and the net effect is that **submodule coverage still does not reach the reported number** — it is now generated, uploaded three times, and read by nothing.

### 1. The submodule artifact has no matrix guard, so it is uploaded once per Go version

`Example-Unit-Testing` and `PKG-Unit-Testing` both gate their upload:

```yaml
- name: Upload Test Coverage
  if: ${{ matrix.go-version == '1.24'}}
```

`Submodule-Unit-Testing` (`.github/workflows/go.yml:381`) has no such gate, so all three legs upload under the same name. From [run 31266842627](https://github.com/gofr-dev/gofr/actions/runs/31266842627):

```
submodule-coverage-reports  51754  2026-08-08T16:28:58Z
submodule-coverage-reports  51754  2026-08-08T16:27:25Z
submodule-coverage-reports  51754  2026-08-08T16:27:25Z
PKG-Coverage-Report        281781  2026-08-08T16:26:04Z
Example-Test-Report          8602  2026-08-08T16:25:41Z
```

Three artifacts, same name, identical size. `actions/upload-artifact@v7` permits the duplicate names (v4 would have 409'd), so nothing complains.

### 2. Nothing consumes it

Both merge steps hardcode two paths:

```bash
grep -h -v "mode:" ./Example-Test-Report/profile.cov ./PKG-Coverage-Report/profile.cov >> merged_profile.cov
```

— in `parse_coverage` (`go.yml:277`) and again in `upload_coverage` (`go.yml:422`). Neither job `needs:` the submodule job either, so `parse_coverage` doesn't even see the artifact; its log for the run above reads:

```
Found 2 artifact(s)
- PKG-Coverage-Report (ID: 9024428262, ...)
- Example-Test-Report (ID: 9024423628, ...)
```

It started at 16:27:02, before two of the three submodule uploads had finished. So the **86.2% published to qlty excludes every module under `pkg/gofr/datasource/*`** — 27 submodules — while appearing to be a whole-repo figure.

`upload_coverage` has the same two-file merge, so the same gap reaches qlty on `development` pushes.

### 3. `parse_coverage` is decorative

```yaml
- name: Parse code-coverage value
[...]

PR #3937, opened 2026-08-16
chore(deps): close the dependabot coverage gaps and retire EOL base images (#3873)

### LINKED ISSUE #3873: CI: dependabot.yml misses a Go module and has no docker ecosystem (examples/http-server is on alpine:3.14, EOL 2023-05-01)
### Summary

`.github/dependabot.yml` declares two ecosystems, `gomod` and `github-actions`. Two things fall outside that: one Go module is not matched by any directory glob, and the repo's 11 Dockerfiles have no dependency automation at all.

### 1. One Go module receives no dependency updates

The `gomod` entry lists:

```yaml
directories:
  - "/"
  - "/examples/*"
  - "/pkg/gofr/datasource/*"
  - "/pkg/gofr/datasource/file/*"
  - "/pkg/gofr/datasource/kv-store/*"
  - "/pkg/gofr/datasource/pubsub/*"
```

There are 31 `go.mod` directories in the tree. Every one is matched by a glob above **except**:

```
pkg/gofr/metrics/exporters/gcp
```

Nothing covers `pkg/gofr/metrics/**` — `/pkg/gofr/datasource/*` does not reach it, and it is not under `/examples/`. So that module has been getting zero dependency updates since it was added. It vendors GCP client libraries, which is not where you want a silent gap.

**Fix:** add `- "/pkg/gofr/metrics/exporters/*"` to the list. Worth considering a broader pattern so the next new module isn't missed the same way.

### 2. No `docker` ecosystem — 11 unmanaged Dockerfiles

There is no `package-ecosystem: "docker"` entry, so no base image in the repo is tracked.

| pin | where | status |
|---|---|---|
| `alpine:3.14` | `examples/http-server/Dockerfile:19` | **EOL 2023-05-01** — released 2021-06-15; currently supported branches are 3.20–3.24 ([alpinelinux.org/releases](https://alpinelinux.org/releases/)) |
| `alpine:latest` | 8 Dockerfiles | floating tag — builds are not reproducible, and "latest" hides what it resolved to |
| `golang:1.24` | 9 Dockerfiles | while `examples/http-server/Dockerfile:2` uses `golang:1.25-alpine` and CI tests `['1.26','1.25','1.24']` |
| `zopdev/static-server:v0.0.9` | `docs/Dockerfile:52` | tag only, no digest |

`alpine:3.14` is the one worth acting on directly — over three years past end of support, receiving no security patches for musl, busybox, or OpenSSL.

The inconsistency in the middle two rows is its own small problem: eight examples pinned to `golang:1.24` while the matrix tests through 1.26 means the published example images demonstrate an older toolchain than the framework supports.

### Suggested fix

```yaml
  - package-ecosystem: "docker"
    directories:
      - "/"
      - "/docs"
      - "/examples/*"
    schedule:
      interval: "weekly"
    groups:
      docker:
        update-types: ["minor", "patch"]
```

plus the `gomod` glob addition above, and a one-off pass to replace `alpine:3.14` and settle `alpine:latest` / `golang:1.24` on deliberate versions.

Note this partly overlaps #3872 (Dependabot cannot update `node-version` in `setup-*` steps): if maintainers choose Renovate there, its Docker manager covers this issue too and the `docker` block above would be unnecessary. Worth deciding the two together.

### Context

Found while auditing dependency automation after #3870 / #3871. Related: #3867 (pin third-party actions by SHA) is the same class of supply-chain hygiene.

akshat-kumar-singhal:
**Description:**

Fixes #3873.

Dependabot declared only `gomod` and `github-actions`, and the `gomod` globs did not match every module. Two gaps followed.

**One Go module received no dependency updates at all.** The tree has 32 `go.mod` directories and every one was matched by an existing glob except `pkg/gofr/metrics/exporters/gcp` — nothing covered `pkg/gofr/metrics/**`, and it is not under `/examples/`. That module vendors the GCP client libraries, so the gap was in exactly the place you would least want one. Adding `"/pkg/gofr/metrics/exporters/*"` closes it.

**The 11 Dockerfiles had no dependency automation whatsoever**, which is how `examples/http-server/Dockerfile` came to sit on `alpine:3.14` — released 2021-06-15, end-of-support 2023-05-01, so no musl, busybox or OpenSSL patches for over three years. A `docker` ecosystem now covers `/`, `/docs` and `/examples/*`, grouped on minor+patch so a weekly bump lands as one PR rather than 11.

The base images are settled on deliberate versions rather than left for Dependabot to discover:

| from | to | where |
|---|---|---|
| `alpine:3.14` | `alpine:3.24` | `examples/http-server` — the EOL one |
| `alpine:latest` | `alpine:3.24` | 8 Dockerfiles — a floating tag makes builds irreproducible and hides what it resolved to |
| `alpine:3.23.3` | `alpine:3.24` | `examples/using-http-service`, for consistency |
| `golang:1.24` | `golang:1.26` | 9 Dockerfiles — `go.mod` declares `go 1.26.0`, so 1.24 only worked by downloading a toolchain |
| `golang:1.25-alpine` | `golang:1.26-alpine` | `examples/http-server` build stage |

`docs/Dockerfile` is left alone: `zopdev/static-server:v0.0.9` is already pinned to a release, and its other `FROM` is the website image the deploy workflows build.

None for library consumers. The example images now ship a current Alpine and are built with Go 1.26.

**Additional Information:**

**Dependencies / merge order:** Fully independent — this PR touches no workflow file, so it merges cleanly with the #3865/#3867, #3866, #3868 and #3869 PRs in any order.

Verification: a check over the resulting config confirms all 32 module directories and all 11 Dockerfile directories are matched by a glob. None of these Dockerfiles are built in CI, so they were verified locally instead: both shapes — the root `golang:1.26` + `alpine:3.24` build, and the multi-stage `examples/http-server` build that carried `alpine:3.14` — build clean, report `3.24.1` from `/etc/alpine-release`, and the resulting http-server image boots, loads `./configs/.env` and stays running.

This partly overlaps #3872 (Dependabot cannot update `node-version` in `setup-*` steps). If maintainers prefer Renovate there, its Docker manager would cover this issue too and the `docker` block here could be dropped — worth deciding the two together.

**Checklist:**

- [x] I have formatted my code using `goimport` and `golangci-lint`. — no Go code changed
- [x] All new code is covered by unit tests. — n/a, dependency configuration and base images


aryanmehrotra (approved):
Verified rather than eyeballed — the sweep is complete and the claims hold.

**Dependabot coverage.** The tree has exactly 32 `go.mod` directories, and I checked each one against
the globs after this change: none uncovered. `pkg/gofr/metrics/exporters/gcp` really was matched by
nothing before, so it had been receiving no updates at all while vendoring the GCP client libraries.

**Base images.** Every `FROM` across the 10 Dockerfiles is now consistent, from a genuinely messy
starting point — `alpine:latest` x8, `alpine:3.14`, `alpine:3.23.3`, `golang:1.24` x9,
`golang:1.25-alpine`. `alpine:3.24` and `golang:1.26` both exist and 3.24 is the current supported
series. Leaving `docs/Dockerfile` alone is right: both its images are already pinned to specific
versions and neither is a public base image needing EOL management.

I did go and check the `alpine:3.14` claim, since it's going into a comment that will outlive the
PR — `examples/http-server/Dockerfile:19` is indeed on it, pinned back in 57fab62d7, and 3.14 lost
support on 2023-05-01. Three years without musl or OpenSSL patches in an image we ship as an example.

The `golang:1.24 -> 1.26` bump reads like scope creep on a base-image PR but isn't: Go supports the
last two majors, so 1.24 is out of support on the same grounds as the alpine change. CI still covers
1.24/1.25/1.26, so nothing conflicts.

One thing your issue asked to confirm, answered here: the `github-actions` Dependabot ecosystem is
already enabled at `directory: "/"`, so the SHA pins from #3936 will be kept current automatically.
Pinning doesn't mean going stale.

Approving. Worth putting a label on #3873 — it has none, so the tracker doesn't reflect that this is
resolved.

Issue #3865, opened 2026-08-08
CI: no concurrency group or job timeouts; superseded runs and hung jobs burn runners


akshat-kumar-singhal:
### Summary

None of the four workflows (`go.yml`, `typos.yml`, `website-prod.yml`, `website-stage.yml`) declare a `concurrency` group or `timeout-minutes`. Verified: `grep -rn "concurrency\|timeout-minutes" .github/workflows/` returns nothing.

### 1. No `concurrency` group

Every push to a PR branch starts a full run while the previous one keeps going. `Workflow-Pipeline` is 14 jobs, and `Example-Unit-Testing` alone brings up Kafka, Redis, MySQL (as `services:`) plus Zipkin and MinIO (as `docker run`) on each of three matrix legs. Two pushes a minute apart means all of that runs twice, and only the second result is ever looked at.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

`cancel-in-progress: true` is right for `pull_request`, and safe for the `push` legs here too since `upload_coverage` is already gated on `github.ref == 'refs/heads/development'` and a cancelled coverage upload is simply redone by the next push. If that is felt to be too aggressive for `development`, the usual form is `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`.

The website workflows trigger on `v*.*.*` tags, so they need a group but should **not** cancel in progress — cancelling a half-finished deploy is worse than letting it complete.

### 2. No `timeout-minutes` on any job

A job that hangs runs to GitHub's 6-hour default before the runner is released. The test steps have their own bound (`nick-fields/retry` `timeout_minutes: 5`) but nothing else does — `Get dependencies`, `Start MinIO`'s 60s poll loop, the lint jobs, and the coverage jobs are all unbounded. A `timeout-minutes: 20` (30 for the matrix legs) at job level costs nothing and caps the damage.

### 3. `typos.yml` runs twice per PR

```yaml
on:
  push:
  pull_request:
```

No branch filter and no `paths-ignore`, so a PR from a branch in the repo triggers it once for the push and once for the PR. `go.yml`'s trigger block is the shape to mirror.

### Why file it

This is pure waste rather than a correctness bug, which is why it has stayed invisible — but it is also about the cheapest fix available in CI here: three small, purely additive blocks with no behaviour change to any test.

### Context

[...]

PR #3939, opened 2026-08-16
fix(ci): wait for Example-Unit-Testing services to be ready, and pin Zipkin (#3868)

akshat-kumar-singhal:
**Description:**

Addresses #3868 (partially — see the deferred item below, so the issue should stay open).

`Example-Unit-Testing` depends on five backing services. Only MinIO was both pinned and waited on; the other four were neither.

**Zipkin** ran as `docker run -d -p 2005:9411 openzipkin/zipkin:latest`. That step returned as soon as the container had *started*, and `:latest` meant the image could change under the repo with no commit. It now gets exactly the treatment the MinIO step directly below it already had: pinned by digest, and a 60s poll on `/health` that fails the job loudly (with container logs) if the service never comes up, rather than letting the tests discover it.

**Kafka, Redis and MySQL** were declared with no health options, so the runner proceeded once each container was created and started — not once the service inside it was accepting connections. MySQL is the slow one and the most likely to have been racing the tests. All three now declare health checks, and the runner blocks on them.

To be clear about what this does and does not claim: as the issue said, **no specific CI failure was ever traced to a service not being ready** — the `nick-fields/retry` wrapper hides first-attempt failures behind a warning. This fixes a known-unsound pattern, and adding the checks is also how we find out whether it was causing the intermittent failures, because a readiness problem will now fail as a service-startup error instead of a mystery test failure.

None. Job startup may be marginally slower where the runner now waits for a service that was previously raced.

**Additional Information:**

**Dependencies / merge order:** Independent — merges cleanly with the #3865/#3867, #3866, #3869 and #3873 PRs in any order (verified by test-merge).

**The `bitnamilegacy/kafka:3.4.1` migration is deliberately NOT in this PR.** That namespace is a frozen archive Bitnami say may be removed, so it does need to move — but `apache/kafka` uses a different env-var surface (`KAFKA_CFG_*` is a Bitnami convention), rewriting the whole `env:` block cannot be validated without a live CI run, and a wrong block turns every pub/sub example test red at once. It belongs in its own PR where a red run means something. #3868 stays open for it, and the reasoning is recorded in a comment above the image so it is not lost. Options remain: move to `apache/kafka`, mirror the current image into GHCR under `gofr-dev`, or take Bitnami Secure Images.

**Every health command was verified against the real image locally rather than written from memory:**

| service | check | result |
|---|---|---|
| zipkin | `GET /health` | returns `{"status":"UP"}`; container reaches `healthy` |
| redis | `redis-cli ping` | exits **0 when up and 1 when down** — checked both ways, so the check can actually fail |
| mysql | `mysqladmin ping -h 127.0.0.1 -ppassword` | reaches `healthy` |
| kafka | `kafka-topics.sh --bootstrap-server 127.0.0.1:9092 --list` | on PATH at `/opt/bitnami/kafka/bin`, exits 0 against a live broker |

Kafka's health timeout is 10s rather than the 5s used elsewhere because that check starts a JVM. The Redis check uses plain `redis-cli ping` rather than piping to `grep -q PONG`: it is no less correct, and quoting a pipe inside `options:` is needless risk.

Verified on darwin/arm64 while CI runs linux/amd64 — the pinned digest is the multi-arch manifest list so it resolves on both, and none of the health commands are arch-dependent. YAML folding was checked to confirm each `options:` block renders as the single-line string docker expects, and `actionlint` reports the same 17 pre-existing findings before and after.

**Checklist:**

- [x] I have formatted my code using `goimport` and `golangci-lint`. — no Go code changed; validated with `actionlint`
- [x] All new code is covered by unit tests. — n/a, CI configuration; every health command verified against the real image


aryanmehrotra (approved):
Approving. I reproduced every claim in the description against the real images rather than taking the table at face value, and all of it holds:

- **Zipkin digest** — `d17e856…` is exactly the manifest-list digest for tag `3.6.1`, and the list carries linux/amd64 and arm64, so it resolves on the runner.
- **`redis-cli ping`** — healthy at ~10s, `PONG`, exit 0.
- **`mysqladmin ping -h 127.0.0.1 -ppassword`** — healthy at ~15s, `mysqld is alive`.
- **`kafka-topics.sh --list`** — on PATH, healthy at ~20s.
- **All three negative-test correctly** — each returns non-zero against a dead port, so none of them is a check that can only pass.
- `actionlint` reports 17 findings before and after; merges into `development` with no conflicts.

I also specifically checked the thing most likely to bite here: `--entrypoint redis-server` with the health options appended. That's safe — Docker clears the image `CMD` when `--entrypoint` is overridden, so redis isn't handed `redis-server` as a config file path. Verified by creating the container the same way Actions does.

Three non-blocking comments, none of which need to hold the merge:

**1. The Kafka probe starts a JVM every 10s for the life of the job.** Docker health checks keep running after a container goes healthy. I measured ~1.23s wall per probe locally, and it'll be worse on a runner — roughly 30 JVM launches over a 5-minute test job, competing for CPU with the tests this PR is trying to stabilise. The bitnami image has bash, so a TCP-level probe is nearly free:

```
--health-cmd "bash -c 'echo > /dev/tcp/127.0.0.1/9092'"
```

Weaker than a metadata query, but readiness is all we're gating on. Worth considering given the goal is less flake rather than more background load.

**2. Kafka fails slow, not fast.** `kafka-topics.sh` against a dead broker hangs rather than erroring — I killed it at 25s and it had not returned. So every failing probe burns the full `--health-timeout=10s`, and at `--health-retries=10` a Kafka that never starts takes roughly 200s to be declared unhealthy. Fewer retries would surface a startup failure much sooner.

**3. Nit:** no `--health-start-period`, so the first probe fires at t=10s even for redis, which is up in ~40ms. Mostly hidden behind Kafka's ~20s.

**Follow-up worth its own issue (spans this and #3937):** this adds a fifth digest-pinned image with nothing that will ever bump it. Dependabot's `github-actions` ecosystem only updates `uses:` refs, and the `docker` ecosystem #3937 adds is scoped to Dockerfiles under `/`, `/docs` and `/examples/*`. Neither covers `services: image:` or `docker run` refs inside `run:` blocks, so `redis:7.0.5`, `mysql:8.2.0` (EOL), the Kafka image, MinIO and now Zipkin are frozen indefinitely. Pinning is still strictly better than `:latest`, so this isn't a reason to hold the PR.

Agreed on deferring the `bitnamilegacy` → `apache/kafka` migration; the reasoning for keeping it out is sound and the in-code comment means it won't get lost.

Issue #3859, opened 2026-08-06
CI: the Go version matrix runs 1.26 on every leg; setup-go-toolchain exists only to sustain it


aryanmehrotra:
### Summary

`Example-Unit-Testing` and `PKG-Unit-Testing` run a three-version matrix — `['1.26','1.25','1.24']` — but all three legs execute Go **1.26.0**. The matrix proves nothing about the older versions it names, and the machinery that keeps it green (`.github/actions/setup-go-toolchain`) exists only to work around the mismatch.

### Why every leg runs 1.26

`go.work` and `go.mod` both declare `go 1.26.0`. Since Go 1.21 that is a hard minimum, not advisory, so an older toolchain refuses the module outright:

```
go: go.mod requires go >= 1.26.0 (running go 1.24.13)
```

`GOTOOLCHAIN=auto` is the only reason the older legs pass — by switching to 1.26. From a recent run (30977134341), the `1.24` leg:

```
Successfully set up Go version 1.24
go version go1.24.13 linux/amd64
GOTOOLCHAIN='local'
materialising Go 1.26.0 as a real SDK for the toolchain switch
```

and the tests then execute on 1.26.0.

`setup-go-toolchain` is a consequence of this, not a separate concern. `GOTOOLCHAIN=auto` on its own satisfies the upgrade with a *toolchain module* in the read-only module cache; Go 1.26 no longer ships `covdata` prebuilt in `pkg/tool` and cannot build it into a read-only GOROOT, so `go test -coverpkg` fails with `go: no such tool "covdata"`. The action installs the required version as a real SDK to dodge that. It is a correct workaround for a matrix that should not need one.

### What the floor could actually be

Measured in scratch worktrees, `GOWORK=off`:

- **`go 1.24.0` + Go 1.24.13** — cannot work at any effort: `go: go.opentelemetry.io/otel@v1.44.0 requires go >= 1.25.0` (`go.mod:32`). The 1.24 leg can never give signal while that dependency is in the graph.
- **`go 1.25.0` + Go 1.25.6** — `go build ./pkg/...` and `go vet ./pkg/gofr` both clean.

So a real two-version matrix is reachable:

```
go.mod / go.work:  go 1.25.0
                   toolchain go1.26.0
CI:                GOTOOLCHAIN=local
matrix:            ['1.26', '1.25']
```

[...]

PR #3940, opened 2026-08-16
fix(ci): merge submodule coverage into the reported figure (#3866)

### LINKED ISSUE #3866: CI: submodule coverage is uploaded three times and merged nowhere; parse_coverage computes a number nothing reads

akshat-kumar-singhal:
**Description:**

Fixes #3866.

Since #3824/#3825, `Submodule-Unit-Testing` does produce coverage — it was then uploaded three times and read by nothing, so the published percentage excluded every module under `pkg/gofr/datasource/*` and `pkg/gofr/metrics/exporters/*` while presenting as a whole-repo number. Three separate causes, all fixed here.

**1. The upload had no matrix guard.** Unlike the Example and PKG uploads, all three legs uploaded under the same artifact name. `upload-artifact@v7` permits duplicate names where v4 would have rejected them, so this produced three identical artifacts and no error. It is now gated on the same `'1.24'` leg the other two use, so #3859 can move all three together rather than having to fix this twice.

**2. Nothing consumed the artifact.** Both merge steps hardcoded two paths, and neither `parse_coverage` nor `upload_coverage` declared `needs:` on the submodule job — so `parse_coverage` did not merely skip the artifact, it started *before* those uploads had finished. Both jobs now depend on the submodule job and merge its profiles, and the two merges are kept identical so the number qlty publishes and the number reported on the PR cannot drift apart. If no submodule profile is present the merge now **fails loudly** rather than quietly reporting a figure that excludes them — same reasoning as the `if-no-files-found: error` added in #3824.

**3. `parse_coverage` was decorative.** `CODE_COVERAGE` went to `$GITHUB_ENV` and its only consumer, the 92% gate, was commented out directly below. The value now goes to `$GITHUB_STEP_SUMMARY`, so it is visible on the run rather than buried in a log. Both jobs also gained an explicit `setup-go` — each runs `go tool cover` and neither had one, silently depending on whatever Go the `ubuntu-latest` image happened to ship.

**4. The two halves were counted under different rules.** The PKG job strips generated mocks (`grep -v '/mock_'`) before uploading its profile, so without the same filter on the merges `pkg/gofr` excluded its mocks while the submodules counted theirs — 66 `mock_*.go` files live under `pkg/gofr/datasource/*` and `pkg/gofr/metrics/exporters/*`. Both merges now apply it. Raised in review; the effect is not small, and its sign depends on which submodules you sample (measured on solr + clickhouse the filter moves the total *up*, 79.1% -> 83.4%, because those mocks are less covered than the real code).

The submodule tests now also pass `-covermode=atomic`, matching the other two jobs. They previously used the default `set` mode, so the merge was combining 0/1 flags with hit counts under a `mode: atomic` header.

None at runtime, but **the reported coverage number will move, and probably down.** As the issue warned, folding in modules that were never counted means the qlty delta on this PR will look alarming. That is the correct number — the previous figure was not a whole-repo one.

**Additional Information:**

**Merge order is settled.** #3936 landed and this branch has been rebased onto it — the conflict was the single predicted hunk (`timeout-minutes` against the `needs:` rewrite), and both lines are kept. It merges cleanly with the #3868, #3869 and #3873 PRs.

**Two things deliberately left to maintainers.** The threshold gate is **not** restored. The issue offered "restore the gate or delete the job", but neither is right yet: deleting `parse_coverage` removes coverage visibility on PRs entirely (`upload_coverage` only runs on `development` pushes), and restoring a 92% floor picks a number blind when this change moves the figure for the first time in a long while. The correct floor can only be chosen once a run has published the post-merge number, so the decision is tracked in #4155 and the workflow comment just points there.

Verified locally by reproducing the pipeline inside a checkout: real `-covermode=atomic` profiles were generated for two submodules plus pkg and example profiles, the merge step run verbatim, and `go tool cover -func` accepts the merged file and reports a total. **The issue's `--strip-prefix` caveat is settled** — every submodule declares `module gofr.dev/pkg/...`, so stripping `"gofr.dev/"` yields real repo-relative paths, confirmed by checking that each path in the merged profile resolves to a file on disk. `actionlint` reports the same 17 pre-existing findings before and after.

**Checklist:**

- [x] I have formatted my code using `goimport` and `golangci-lint`. — no Go code changed; validated with `actionlint`
- [x] All new code is covered by unit tests. — n/a, CI configuration; the merge was reproduced locally against real profiles
- [ ] This PR does not decrease the overall code coverage. — **the reported figure will change**, because it now includes submodules that were never counted; see Breaking Changes


.github/workflows/go.yml
```diff
@@ -308,39 +308,76 @@ jobs:
...
+          fi
+          echo "Merging ${#submodule_profiles[@]} submodule profile(s) with Example and PKG"
           echo "mode: atomic" > merged_profile.cov
-          grep -h -v "mode:" ./Example-Test-Report/profile.cov ./PKG-Coverage-Report/profile.cov >> merged_profile.cov
-
-      # Calculate and output the total code coverage percentage
+          grep -h -v "mode:" \
+            ./Example-Test-Report/profile.cov \
+            ./PKG-Coverage-Report/profile.cov \
+            "${submodule_profiles[@]}" >> merged_profile.cov
```
aryanmehrotra: This merge and its twin in `upload_coverage` need the same `/mock_` filter the PKG job applies at `go.yml:294` — otherwise `pkg/gofr` excludes its generated mocks while the submodules include theirs, inside one number. 66 `mock_*.go` files sit under the submodules, and they are 53.7% of the coverable statements in the three profiles I generated. Because gomock mocks are heavily exercised, the effect is to inflate the figure, not depress it — 33.4% against 28.0% in my repro. Suggested: pipe the grep through `| grep -v '/mock_'` before the redirect.

.github/workflows/go.yml
```diff
@@ -465,12 +523,24 @@ jobs:
...
+            echo "::error::no submodule coverage profiles found — refusing to publish a figure that silently excludes them"
+            exit 1
+          fi
+          echo "Merging ${#submodule_profiles[@]} submodule profile(s) with Example and PKG"
           echo "mode: atomic" > merged_profile.cov
-          grep -h -v "mode:" ./Example-Test-Report/profile.cov ./PKG-Coverage-Report/profile.cov >> merged_profile.cov
+          grep -h -v "mode:" \
+            ./Example-Test-Report/profile.cov \
+            ./PKG-Coverage-Report/profile.cov \
+            "${submodule_profiles[@]}" >> merged_profile.cov
```
aryanmehrotra: Same `/mock_` filter needed here. This is the merge whose number reaches qlty, so if the two diverge on this the published figure and the PR-reported figure agree with each other while both disagree with the convention the PKG job already follows.

aryanmehrotra (changes requested):
Thanks for this — the diagnosis in the issue is right and I reproduced the whole pipeline rather
than reading it. The mechanism works: submodule profiles really do reach the reported number now.

**TL;DR:** one real change needed — the merged figure is computed under two different rules, so it
is flattered by generated mocks — plus a rebase, since #3936 landed on the 21st. Everything else is
approvable and the verification held up well.

## What I checked

Generated real `-covermode=atomic` profiles for a metrics exporter and three datasource submodules,
built the merged file with your exact merge step, and ran `go tool cover -func` from
`working-directory: artifacts` the way the job does:

```
submodule functions present in -func output : yes (gcp resolves)
total, pkg-only profile                     : 14.0%
total, same profile + one submodule         : 15.3%
```

So the `--strip-prefix` caveat really is settled — every submodule declares `module gofr.dev/pkg/...`
and the paths resolve against the checkout without stripping. Good call verifying that rather than
assuming it.

The rest also holds:

- The empty-glob branch does what it says under GitHub's default `bash -e`: exits 1 with the
  `::error::` line when the glob is empty, exits 0 and reports the count when profiles are present.
- `actionlint` finds exactly the same set before and after — no new findings.
- The workflow parses.
- The conflict is a single hunk: `timeout-minutes: 10` from #3936 against your `needs:` rewrite,
  exactly as you predicted. Keep both lines and it is done.

## The one real change

The PKG job strips generated mocks before it uploads its profile:

```yaml
grep -h -v "mode:" gofr_only.cov submodules.cov | grep -v '/mock_' >> profile.cov
```

Neither of the two new merges applies that filter, so `pkg/gofr` excludes its mocks while the
submodules include theirs — inside a single number that the PR presents as whole-repo.

That is not a rounding difference. Across the three submodule profiles I generated:

```
mock_ statements : 485
other statements : 419      -> 53.7% of submodule coverable statements are generated mocks
```

66 `mock_*.go` files live under the submodules. And the effect runs *upward*, not downward — gomock
mocks are heavily exercised by the tests that use them:

```
merged as this PR merges it (mocks counted) : 33.4%
merged with the PKG job's /mock_ filter     : 28.0%
```

Those two figures come from a small stand-in pkg profile, so treat them as the size of the effect
rather than a prediction of the real number. The direction is the part that matters: the new figure
is flattered by generated code that this repo's own convention says should not count, and the two
halves of one number are computed differently — which is the drift the PR sets out to prevent.

It also lands on the threshold decision you deliberately deferred. Picking a floor from a number
inflated by mocks would bake the inconsistency into the gate, so this wants fixing before that
figure is published, not after.

One filter on each merge, matching the PKG job:

```bash
grep -h -v "mode:" \
  ./Example-Test-Report/profile.cov \
  ./PKG-Coverage-Report/profile.cov \
  "${submodule_profiles[@]}" | grep -v '/mock_' >> merged_profile.cov
```

## Two things worth knowing, neither blocking

**`go tool cover -func` exits 2 on an unresolvable path**, not just on a missing file — I checked by
adding a bogus entry to a merged profile. So a stale or unresolvable path in any profile turns the
Code Coverage check red rather than reporting a lower number. It should not be reachable in normal
operation since every profile comes from the same checkout, but it is worth knowing before the first
red run gets read as a flake.

**The PR check gets slower.** `parse_coverage` now waits on the whole submodule matrix, whose slowest
leg was 5m45s on this run against PKG's 3m16s. That is the right trade for a correct number, just
worth saying out loud.

I also looked at the hardcoded `go-version: '1.26'` in the two new `setup-go` steps and decided it is
fine — that is already the pattern in three other jobs on `development`, so it is consistent rather
than a new thing to chase.

## On the deferred threshold

Agreed with leaving the gate out, and the reasoning is sound — a floor picked before the post-merge
number exists is a guess. But a `# NOTE:` in a workflow is a weak place for a decision to live; it
will be read by whoever next edits that job, which may be nobody for a year. Could you file it as a
follow-up issue instead, so it sits in the queue with the number attached once a run publishes it?
Same for the choice itself — that is a maintainer call and it is easier to make on an issue than
inside a YAML comment.

One real change needed, then a rebase. Everything else is approvable.

akshat-kumar-singhal:
Thanks — both blocking items are addressed.

**Rebased** onto `development`. The conflict was the single hunk you described:
`timeout-minutes: 10` from #3936 against the `needs:` rewrite. Both kept.

**The `/mock_` filter is applied to both merges.** I checked the premise rather than
taking it: the PKG job does strip mocks at `.github/workflows/go.yml:312` before
uploading, and there are 66 `mock_*.go` files under `pkg/gofr/datasource/*` and
`pkg/gofr/metrics/exporters/*`. One number computed under two rules is exactly the drift
this PR is meant to prevent, so the filter belongs on both merges.

**One correction, on the direction of the effect.** I reproduced the merge with real
`-covermode=atomic` profiles (solr and clickhouse, with stand-in Example/PKG artifacts)
and ran the merge step verbatim under `bash -e`:

```
merged, mocks counted   : 79.1%
merged, /mock_ filtered : 83.4%

mock  statements 212, covered 131  -> 61.8%
other statements 559, covered 458  -> 81.9%
```

In this sample the filter moves the figure **up**, not down — the generated mocks are
*less* covered than the real code, the opposite of what your three submodules showed. So
"the merged figure is flattered by generated code" doesn't generalise; the sign depends
on which submodules you sample. The argument that does hold is the one you led with, and
it's sufficient on its own: two halves of a single number computed under different rules.
Worth pinning down before the threshold gets picked, so the floor isn't chosen on an
assumed direction.

`actionlint` reports the same findings before and after, with line numbers shifted only.

**On the deferred threshold** — agreed, a `# NOTE:` in a workflow is a weak home for a
decision. Filed as #4155, with the placement and the reasoning; the comment in the
job now points at it instead of carrying the argument.

On the two non-blocking notes: agreed on both. `go tool cover -func` exiting 2 on an
unresolvable path is worth knowing, and the slower PR check is the right trade.

aryanmehrotra (dismissed):
Both addressed. Re-reviewed at `b04488e80`; the delta since my review is **+12 lines in one file**, nothing else touched.

| # | Ask | Verified |
|---|---|---|
| 1 | `\| grep -v '/mock_'` on both merges | ✅ lines **382** and **585**, identical form, matching the PKG job — each with the rationale inline |
| 2 | Move the deferred gate out of a YAML comment | ✅ **#4155** exists and is open; the `# NOTE:` now points at it |
| 3 | Rebase for #3936 | ✅ `parse_coverage` carries **both** `timeout-minutes: 10` and the new `needs:` |

The filter is real, not cosmetic: `find pkg/gofr/datasource pkg/gofr/metrics/exporters -name 'mock_*.go'` returns **66 files**, exactly the number your comment cites, and profile lines carry the full import path so `/mock_` matches them.

## The number #4155 is waiting for — this PR has already published it

From this PR's own green run versus `development`:

| Source | Reported coverage |
|---|---|
| `development` @ `9bbdbbe23` | **87.0%** |
| this PR @ `b04488e80` | **74.7%** |
| **Δ** | **−12.3 pp** |

That is the intended effect — folding in submodules that were never counted, with mocks now excluded on both halves — but it is a bigger move than "the number changes" conveys, and two things follow:

- It is the input #4155 needs. Worth pasting onto that issue now rather than waiting for a post-merge run, since the shape of the answer will not change.
- CONTRIBUTING says *"No PR should ever decrease the overall code coverage"* and that the coverage check fails on any decrease. The `Code Coverage` job passes here because the 92% gate is still commented out, but **qlty's baseline shifts under every open PR the moment this lands**. Someone should expect a wave of coverage-check noise on unrelated PRs and decide in advance whether to re-baseline.

Neither is a change I am asking for in this PR — the number is correct and the inconsistency you fixed is exactly why it moved. Flagging so the merge is a decision rather than a surprise.

LGTM.

akshat-kumar-singhal:
Rebased — or rather merged, matching the two `development` merges already on this branch. `470dc205`.

The approval was dismissed by the push, so this needs a re-look. It should be a cheap one: **the only delta since `b04488e80` is the conflict resolution, one hunk in one file.**

## The conflict

#4201 landed and removed `-short` from the submodule `go test`; this PR adds `-covermode=atomic` to the same line. Both are wanted, so both are kept, and your comment explaining the `-short` removal is preserved verbatim with the `-covermode` rationale appended to it:

```bash
go test ./... -v -covermode=atomic -coverprofile="${module_name}.cov" -coverpkg=./...
```

I checked the constraint that comment documents — that block is the single-quoted payload of the `bash -c`, so an apostrophe anywhere in it, comments included, kills the runner before any test runs. The only `'` characters in the region are the two delimiters.

## Verification

- `actionlint` reports **9** shellcheck findings against `upstream/development`'s `go.yml` and **9** against this branch — same set, no new ones.
- The workflow parses.
- Net diff against `development` is **1 file, +101/−18, 5 hunks** — nothing from the merge leaked in.
- All checks green on `470dc205`, including `Code Coverage`.

## The number you asked to be posted on #4155

Worth flagging here too, because it changes how alarming this merge looks:

| Source | Reported |
|---|---|
| `development` @ `851b78a7` | **90.2%** |
| this PR @ `470dc205` | **85.2%** |
| **Δ** | **−5.0 pp** |

Not the −12.3 pp from your review. Your 87.0% and 74.7% both reproduce from their job logs, so that delta was right when you took it — but #4202/#4203 widened the PKG job's `-coverpkg` and lifted the base to 90.2%, and #4201 un-skipped the mongo mtest functions, which lifts this branch too. The full comparison and the caveat that 85.2% is a PR-branch run rather than a post-merge `development` run are on [#4155](https://github.com/gofr-dev/gofr/issues/4155).

The re-baseline you warned unrelated PRs would see is correspondingly smaller.

## One thing I did not change

#4203 added a `/testutil/` filter next to `/mock_` in the PKG job. The two merges here filter only `/mock_`. It is inert today — there is no `testutil` directory under `pkg/gofr/datasource` or `pkg/gofr/metrics/exporters`, so nothing is being counted under two rules right now. But it is a second place the two halves could diverge under exactly the rule this PR exists to enforce, and it will stop being inert the day a submodule grows a test helper package. Left out because it would change an approved diff; happy to add it here or file it as a follow-up, whichever you prefer.

aryanmehrotra (approved):
**Approving.** Verified against the completed run at head `2c162a189`, not by reading the diff — I downloaded the three artifacts that run produced and recomputed the merged figure independently.

## The headline changed under you, in your favour

`#4274` merged as `19a975982` at 05:01 UTC, three minutes before you merged `development` into this branch. `2c162a189` contains it. So the run on this head is already the post-`#4274` world.

| measured on | Example+PKG only | submodules alone | **merged (this PR)** | cost of merging |
|---|---|---|---|---|
| Sep 15 (`470dc205c`) | 90.1% | 76.2% | **85.2%** | **−4.9 pp** |
| now (`2c162a189`) | **95.2%** | **94.5%** | **94.9%** | **−0.3 pp** |

The gap between the two halves of the number went from **13.9 pp to 0.7 pp**. The thing this PR exists to expose turned out to be almost nothing, because `#4274` fixed it in the same week.

## What I verified, and how

Downloaded `Example-Test-Report`, `PKG-Coverage-Report` and `submodule-coverage-reports` from run `35958334731` and re-derived the total from the raw blocks — merging duplicate `(file, span)` keys by summing counts, which is what `go tool cover -func` does.

| check | result |
|---|---|
| merged total | **94.9% (16377/17252 stmts, 13150 blocks)** — matches `parse_coverage`'s printed `94.9%` to the decimal |
| submodule profiles present | **27 of 27** `go.mod` directories under `pkg/` |
| duplicate-artifact bug | **1** `submodule-coverage-reports` artifact, not 3 |
| cross-profile double counting | **zero** blocks appear in more than one artifact |
| the two merge blocks "kept identical" | **byte-identical** apart from the `::error::` string |
| `grep -v '/mock_'` symmetry | PKG's profile contains **0** `/mock_` lines already, so the filter only ever bites the submodule half — which is exactly the asymmetry it is there to remove |
| `--strip-prefix="gofr.dev/"` | **376/376** paths in the merged profile resolve to a file in the `2c162a189` tree |
| `-covermode=atomic` on submodules | submodule profiles now carry counts, not 0/1 flags; **0** duplicate block keys across the 27 files |

The mock filter is worth a number: it excludes **5,777 statements** that would have been **43.7%** covered. Counting them would report **82.1%** instead of 94.9%. Your review comment about the sign depending on which submodules you sample was right — across all 27 the filter moves the figure **up** by 12.8 pp.

## Two asks before merge

**1. The Breaking Changes section is now wrong.** It still says:

> **the reported coverage number will move, and probably down** … the qlty delta on this PR will look alarming

It is **−0.3 pp**. That paragraph was accurate against a 76.2% submodule half and is now the main reason a reviewer might hesitate over this. Please rewrite it with the measured number. Same for the unchecked checklist item — `−0.3 pp` is noise, and saying so is better than leaving it looking unresolved.

**2. `#4155` is now answerable, and I think it should be answered here.** The PR body's own reasoning for deferring the threshold was:

> the correct floor can only be chosen once a run has published the post-merge number

That run has now published it: **94.9%, with 2.9 pp of headroom over 92%**. The commented-out gate can be restored at its original floor without picking a number blind. If you would rather keep this PR to one concern and do it in `#4155`, that is fine — but the blocking unknown is gone either way, and the comment in the workflow should say so rather than implying the number is still unmeasured.

## Two notes, neither blocking

**Coverage reporting is now gated on all three submodule matrix legs.** `needs:` requires every leg green, so a submodule failure on **1.25 alone** now skips `parse_coverage`, where before it still reported. That is the correct trade — a figure computed from a partial matrix is worse than no figure — but it is a new way to lose the number and worth knowing about when it first happens.

**The Example profile is 25–26× duplicated, and this PR is the natural place to notice it.** In the head run's artifact: **262 unique blocks written as 6,597 lines**, one copy per test binary under `-coverpkg=./examples/...`. The PKG job already solves this with its `awk` merge (7,031 lines, 7,031 unique); the Example job has no equivalent.

This does not affect `parse_coverage` — `go tool cover -func` merges duplicate blocks, which is why my recomputation and CI agree exactly. **What I have not verified is whether `qlty coverage publish --format=coverprofile` does the same.** If it does not, then the per-file view qlty shows for `examples/` is built from 26 stacked copies of every block, and the one number this PR is about keeping in sync is the one place it would show up. Worth a follow-up issue rather than a change here — an `awk` merge on the Example profile mirroring the PKG job would settle it.

---

Nothing in the change itself needs fixing. The pipeline is right, the arithmetic is right, and I reproduced it end to end.

PR #3942, opened 2026-08-17
fix(mcp): claim the MCP port with net.Listen instead of exiting from EnableMCP

aryanmehrotra:
## The bug

`EnableMCP` probed the port and called `Logger.Fatalf` when something held it:

```go
if !isPortAvailable(port) {
    a.container.Logger.Fatalf("MCP port %d is blocked or unreachable", port)
}
```

`Logger.Fatalf` is **`os.Exit(1)`** — reached from library code, during setup, before anything is serving. Two consequences:

**In a service.** The exit happens inside `EnableMCP`, so no shutdown hook, datasource close, or in-flight work gets a say. The process is simply gone.

**In the test suite.** `TestEnableMCP_ServerConfigured` asserts `defaultMCPPort`, so it cannot be handed a free port:

```go
require.NotNil(t, app.mcpServer, ...)
assert.Equal(t, defaultMCPPort, app.mcpServer.port)
```

On any machine where something holds 8200, `os.Exit` bypasses the test framework and kills the test binary mid-run. Every remaining test in `pkg/gofr` is then reported as failed, with **no `--- FAIL` line** to say which test did it:

```
{"level":"FATAL", ..., "message":"MCP port 8200 is blocked or unreachable"}
FAIL	gofr.dev/pkg/gofr	5.184s
```

That is how I hit it — an unrelated local process on 8200 was silently reddening the entire package. CI never sees it because runners have 8200 free, which is exactly why it survived.

Worth naming the likely cause: **8200 is also [HashiCorp Vault](https://developer.hashicorp.com/vault)'s default port.** Anyone running Vault locally had their GoFr app refuse to start over a default they never chose.

## Why the probe was the wrong instrument

`isPortAvailable` **dials**:

```go
conn, err := dialer.DialContext(ctx, "tcp", fmt.Sprintf(":%d", port))
if err != nil { return true }
```

I measured it rather than assuming, running gofr's implementation verbatim against real listeners. For the common case it is **accurate** — with another process on `127.0.0.1:P` it reports blocked and the loopback bind does indeed fail. So the argument is not that the probe is simply wrong.

Where it fails is a **wildcard** listener. With a separate process holding `[::]:18299`:

```
probe_says_free=false   mcp_can_bind_loopback=true   bindErr=""
```

The probe reports blocked, GoFr exits — but MCP could have bound `127.0.0.1:18299` perfectly well. A service killed at startup because something unrelated holds the wildcard, when its own listener would have come up fine. (Measured on darwin/arm64; the loopback-vs-wildcard interaction is platform-dependent, so I would not claim this reproduces identically on Linux.)

A dial also cannot be *stable*: the port can be taken in the window between the probe and the `Listen`.

## The change

**`net.Listen` replaces the probe.** It does not report whether the port looks free — it takes it. No window between check and claim, and no wildcard false positive.

**The claim happens synchronously in `Run`, before any server goroutine starts.** `EnableMCP` does no network I/O at all; it resolves the port from config and stops.

**A port that cannot be claimed fails startup.** Calling `EnableMCP()` means you asked for the transport, so coming up without it is the worse outcome — an orchestrator would report the service healthy while agents could not reach it. `bindMCPServer` returns false and `Run` unwinds exactly as it does for a failed `OnStart` hook. No `os.Exit`.

The error names the remedy, not just the problem:

```
ERROR  MCP server cannot start on port 8200: listen tcp 127.0.0.1:8200: bind: address already in
       use. Set MCP_PORT to a free port, or MCP_PORT=0 to run without the MCP transport while
       keeping tools available in-process.
```

### Why the bind is in `Run` and not in `mcpServer.Run`

This is the part worth reviewing closely. `startAllServers` launches all four servers as concurrent goroutines under one waitgroup:

```go
a.startMetricsServer(&wg)
a.startMCPServer(&wg)
a.startHTTPServer(&wg)
...
wg.Wait()
```

A failure raised from *inside* MCP's goroutine would race the others. If it asked for shutdown before `httpServer.run` had assigned `s.srv`, `Shutdown` would find nothing to close and the HTTP server would then start and serve forever on an already-canceled context — a hung process. Claiming the port while nothing is running removes the race rather than tries to synchronise it.

`mcpServer.Run` now serves the listener `bind` created, and refuses to serve if reached unbound, which would mean the startup sequence was broken.

Also fixes a missing `err != nil` guard that `httpServer.run` has and this did not.

## Tests

`TestEnableMCP_ServerConfigured` becomes hermetic **without changing its assertion** — pinning the default port is safe precisely because `EnableMCP` does no I/O.

New tests, all against real occupied listeners rather than assumed-busy ports:

- `TestEnableMCP_OccupiedPortStillConfigures` — reaching the line after `EnableMCP` is itself the assertion; before this change the call exited the process.
- `TestBindMCPServer_OccupiedPortAbortsStartup` — startup stops, and the message is actionable (asserts `MCP_PORT=0` appears, not just the raw error).
- `TestBindMCPServer_FreePortProceeds` — pins that `bind` **claims** the port rather than checking it: a second bind of the same address must now fail.
- `TestBindMCPServer_NoServerIsNotAFailure` — `MCP_PORT=0` must not look like a startup failure.
- `TestMCPServer_Run_ServesOnTheBoundListener` — end-to-end over the pre-bound listener, and `Run` returns after `Shutdown` so the waitgroup cannot hang.
- `TestMCPServer_Run_WithoutBindRefusesToServe` — guards the ordering the design depends on.

Mutation-checked: making `bind` never fail reds `OccupiedPortAbortsStartup` and `FreePortProceeds`.

Verified with 8200 genuinely occupied on the machine running the suite: `go test ./pkg/gofr/` goes from **`FAIL` to `ok`**. `golangci-lint` clean under `only-new-issues`.

## Found while here, deliberately not fixed

`grpc.go:183` has the identical pattern:

```go
if !isPortAvailable(g.port) {
    c.Logger.Fatalf("gRPC port %d is blocked or unreachable", g.port)
    c.Metrics().IncrementCounter(context.Background(), "grpc_server_errors_total")
    c.Metrics().SetGauge("grpc_server_status", 0)

    return
}
```

The three statements after `Fatalf` never run in a real service, because the production logger exits there. They *do* run under `grpc_test.go`, which installs a `nonExitingMockLogger` — so this is not dead code, and I am not going to guess at the author's intent. What it does mean is that the error counter and the `grpc_server_status` gauge are only ever set in tests, never in the situation they exist to report.

The check is redundant there too — `(&net.ListenConfig{}).Listen` with proper error handling follows at line 194 — and the existence of `nonExitingMockLogger` is a test working around the exit rather than covering it.

`gofr.go` (HTTP) and `factory.go` (metrics) have the same shape. I have left all three alone rather than fold changes to the primary listener into a fix for an optional one; gRPC in particular would be a near-mechanical follow-up now that the pattern here is established. Happy to do it if you agree with the reading.

**Behaviour change worth stating plainly: this path's exit code goes from 1 to 0.** The old `Fatalf` was `os.Exit(1)`; aborting `Run` returns normally, so the process exits **0** — measured, three runs each, with the MCP port genuinely held:

```
development exit=1   PR exit=0
development exit=1   PR exit=0
development exit=1   PR exit=0
```

That is consistent with how a failed `OnStart` hook already behaves, and it is the price of unwinding instead of exiting. It is not free: under `restartPolicy: Always` nothing changes, but a Job or `restartPolicy: OnFailure` now reads a taken MCP port as success. GoFr has no way today to say "refused to start" in an exit code — that is a framework-wide gap, worth a separate issue, and I did not want to invent a convention for it inside this PR.

---

## Revised after review

**`MCP_PORT` range is validated.** `mcpPort` folded anything unparseable to the default but let a *parsed* out-of-range number through, so `MCP_PORT=99999` reached `net.Listen`, which rejects it **permanently** rather than transiently. `bindMCPServer` treats any bind error as fatal, so one extra digit stopped the whole service — every transport, not just MCP — and the emitted remedy talked about occupancy, which was not the problem. Out-of-range now folds to the default with a loud error, like the unparseable case.

**The disable switch is normalized.** It compared the raw string, so `MCP_PORT="00"`, `"+0"` and `" 0 "` fell through to the default and silently **enabled** the transport on 8200 for an operator who meant to turn it off — the exact Vault-colliding port this PR argues against. It parses first, then compares the number.

**`bind` honours the context itself.** The review noted a SIGINT in the bind window should report graceful shutdown rather than the port remedy. Worth recording that the stated mechanism does not hold here: `ListenConfig.Listen` only observes the context while **resolving a name**, and the address is a literal `127.0.0.1`, so cancellation was never seen at all — the bind *succeeded* and startup continued on an already-cancelled context. Measured:

```
literal loopback: err=<nil>                                    isCanceled=false  listener=true
hostname        : err=listen tcp: lookup localhost: operation was canceled  isCanceled=true   listener=false
```

So the fix is an explicit `ctx.Err()` check in `bind`, which makes the behaviour the review asked for actually reachable.

**A failed bind runs shutdown before returning.** The startup hooks have already opened the container's datasources by that point and `Run` returns normally afterwards, so they were being left to process exit. The doc line claiming "shutdown hooks and datasource cleanup are unaffected" is corrected rather than left overstating the failure path.

**The abort test is portable.** It asserted the OS wording of `EADDRINUSE`, which differs on Windows; it asserts the framework's own remedy and the port number instead.

**The malformed-port branch is covered.** `TestMCPPort_Resolution` is a table over every branch — unset, valid, the four spellings of zero, out-of-range, and non-numeric — with the disable notice matched on stdout and the fold warnings on stderr.

---

## Revised again

**The cleanup on a failed bind was untested, and the test now bites.** Deleting `a.shutdownAfterFailedStartup()` outright left the entire package green — the behaviour was real but unprotected, which is the same as not having it once someone refactors. `TestBindMCPServer_FailedBindReleasesDatasources` asserts the completion line `(*App).Shutdown` emits, and removing the call reds it.

**The startup-hook path had the identical hole, with no cleanup at all.** `handleStartupHooks` returning false also returns from `Run` normally, and the hooks run *after* the container has opened its datasources — so a failing or panicking `OnStart` hook left those connections to process exit. It was the older half of the same bug, and the fix for the MCP path made the asymmetry worse rather than better. Both paths now unwind the same way, and `TestHandleStartupHooks_FailureReleasesDatasources` covers the hook error and the `context.Canceled` case.

Each new test is mutation-checked against its own call site: removing one reds one test, not both.

**Dropped a no-op wrapper.** `context.WithoutCancel(context.Background())` — `Background` is never canceled, so the wrapper did nothing. Replaced with `context.Background()` and a note on why the shutdown timeout is rooted in a fresh context at all: the run's own context may already be canceled, which is one of the ways startup gets abandoned, and a shutdown inheriting it would be dead on arrival.


PiyushSingh-ZS:
Reviewed against `development`, built and ran the MCP tests on the branch. This is the right fix, and the reasoning in the comments is the strongest part of it: `net.Listen` answers the actual question (can I have this port) rather than the adjacent one (is something listening right now), it closes the TOCTOU window the dial probe couldn't, and getting `os.Exit` out of library code during setup is worth doing on its own.

Binding before any server goroutine starts — rather than discovering the failure inside `Run` under the shared waitgroup — is also clearly correct, and the comment explains why better than the code could.

### One consequence worth deciding deliberately: the exit code changes

`Logger.Fatalf` exited non-zero. Now `bindMCPServer` logs, returns `false`, `Run()` returns, `main` returns — and the process exits **0**. Under Kubernetes that reads as `Completed` rather than `CrashLoopBackOff`, and under systemd it won't trigger a restart policy.

That's consistent with how `handleStartupHooks` already behaves, so it's defensible as-is. But the PR's own framing is:

> A service that started successfully and quietly lacked a transport it was configured to expose would be the worse outcome — an orchestrator would report it healthy while agents could not reach it.

An orchestrator seeing a clean exit-0 has much the same problem: it won't report the failure as a crash. If the intent is that a port conflict is fatal to startup, the exit status should probably say so — either `os.Exit(1)` at the `Run()` call site, or a documented non-zero path for startup failures generally. At minimum it deserves a release note, since anyone currently relying on the crash signal loses it silently.

### Latent bug this makes slightly easier to hit

`Shutdown` returns early when `srv == nil`:

```go
if srv == nil {
    return nil
}
```

If shutdown fires between `bind` succeeding and `Run` assigning `m.srv`, `Shutdown` reports success, then `Run` proceeds to `Serve` on the already-bound listener — which now has nothing that will ever close it, so it serves indefinitely and `wg.Wait()` never returns. The race exists in shape on `development` too, but splitting bind from serve means the listener is live for longer before `srv` is assigned.

Since `m.listener` is now a field, `Shutdown` could close it when `srv` is nil, which closes the hole:

```go
if srv == nil {
    if m.listener != nil {
        return m.listener.Close()
    }
    return nil
}
```

(with the listener read under the same mutex).

### Follow-up

The same `isPortAvailable` + `Fatalf` pattern is still in `gofr.go` (three call sites), `grpc.go:182` and `factory.go:120`. It's why `go test ./pkg/gofr/` aborts outright on a machine where the default HTTP port is taken — the FATAL kills the test binary. Same treatment there would be a real improvement to testability.

NitinKumar004 (changes requested):
Good direction — removing the `Logger.Fatalf`/`os.Exit` port probe from library code is the right call. A few edge cases in port handling need tightening first.

### Medium
- **Out-of-range numeric `MCP_PORT` hard-aborts the entire app.** `mcpPort()` only guards `err != nil || port <= 0`, so `MCP_PORT="99999"` (extra digit / pasted URL port) passes through; `net.Listen("tcp","127.0.0.1:99999")` returns `invalid port`, `bindMCPServer` treats any bind error as fatal, and **the whole service refuses to start** (all transports), permanently — unlike a transient `EADDRINUSE`. The emitted remedy ("Set MCP_PORT to a free port") misframes a permanent range error as occupancy. Please validate `port < 1 || port > 65535` and fold/​reject explicitly.
- **Disable check is too strict.** `portStr == "0"` means `MCP_PORT="00"`, `"+0"`, or `"0 "` (trailing space) are **not** recognized as disable — they fall through to `Atoi`→fold-to-8200 and silently **enable** the transport for an operator who meant to turn it off. Trim/normalize before the compare.

### Low
- **Signal during bind is misreported.** `ListenConfig.Listen(ctx, ...)` honors cancellation; a SIGINT/SIGTERM in the bind window returns `context.Canceled`, but `bindMCPServer` logs the port remedy instead of the graceful-shutdown message (which `handleStartupHooks` already special-cases). Add `errors.Is(err, context.Canceled)`.
- **Bind-fail early-returns `Run` → exit code 0 + no cleanup.** The removed code exited 1; now `main` returns normally (orchestrators/CI see success for a failed start) and `Shutdown`/`OnShutdown`/datasource cleanup are skipped, so the doc line "shutdown hooks and datasource cleanup are unaffected" overstates the failure path. Consider Shutdown-then-return, or scope the doc claim to the success path.
- **Test asserts OS-specific `"address already in use"`** — fails on Windows (different EADDRINUSE text). Assert on the framework remedy text, or `errors.Is(err, syscall.EADDRINUSE)`.
- **Malformed `MCP_PORT` typo silently binds 8200** — the exact Vault-colliding default the PR argues against. Fail/warn + add a test for the parse-failure branch.

**Requested changes:** validate the port range; normalize the disable check; graceful-shutdown message on `context.Canceled`; run cleanup on bind-fail (or soften the doc); make the abort test portable; cover the malformed-port branch.

PiyushSingh-ZS (commented):
The core change is right and the investigation behind it is unusually thorough. `Logger.Fatalf` is `os.Exit(1)` from library code during setup, and `net.Listen` replacing a dial-based probe removes both the TOCTOU window and the wildcard false positive in one move.

Two things I would like resolved before this merges, plus notes.

### What I verified by reading the branch

- `bindMCPServer` runs before `startShutdownHandler`, so the new failed-startup shutdown cannot race the signal handler's shutdown. Good ordering.
- `mcpServer.Shutdown` unlocks on both branches — no leaked mutex on the `srv == nil` path.
- The bind-to-Run window is genuinely covered. If `Shutdown` lands after `m.srv` is assigned but before `srv.Serve(listener)` runs, `http.Server.Serve` returns `ErrServerClosed` from `trackListener` and its own `defer l.Close()` releases the listener, so the waitgroup cannot hang and the port cannot leak. The `stopped` flag covers the earlier half of the same window. Both directions look correct to me.
- `App.Shutdown` is nil-safe for every server that never started, so `shutdownAfterFailedStartup` is safe to call at that point.

The measurement that `isPortAvailable` is *accurate* in the common case and only wrong for a wildcard listener — rather than arguing dialing is self-evidently wrong — is the part that makes the case convincing.

---

### 1. The exit code regression should be fixed here, not deferred

The PR is upfront about it:

> **Behaviour change worth stating plainly: this path's exit code goes from 1 to 0.**

and then leaves it as "a framework-wide gap, worth a separate issue". I do not think that is the right place to stop, because this PR is what introduces the regression. Under `restartPolicy: OnFailure`, or in a Job, a service that refused to start now reports success and is never retried.

That also cuts against the PR's own thesis:

> A service that started successfully and quietly lacked a transport it was configured to expose would be the worse outcome.

Exit 0 on a refused startup is the process-level version of exactly that silence.

I do not think fixing it reintroduces what was wrong with `Fatalf`. `Fatalf` was bad because it was `os.Exit` from *library code, mid-setup, with nothing unwound*. A non-zero exit at the bottom of `Run()`, **after** `shutdownAfterFailedStartup()` has completed, is a different thing: top of the call stack, datasources already released, nothing skipped. That is what the rest of the Go ecosystem does and it is what an orchestrator needs to see.

### 2. Scope — three changes in one PR

1. The MCP bind fix (the subject).
2. `MCP_PORT` parsing: range validation and normalized zero (`"00"`, `"+0"`, `" 0 "`).
3. `handleStartupHooks` now calls `shutdownAfterFailedStartup()`.

(3) is a genuine bug — the hooks run after the container has opened its datasources, and `Run` returns normally, so those connections were being left to process exit. I agree with the fix. But it changes shutdown behavior for **every GoFr app with an `OnStart` hook**, and it currently arrives under an MCP-titled PR, announced in a "Revised again" section. It deserves its own PR so it can be reviewed, released and if necessary reverted on its own terms. (2) is smaller but similarly independent.

This PR is already 38 commits behind with conflicts in `gofr_test.go`, `mcp.go` and `mcp_test.go`, so splitting is probably also the cheapest route through the rebase.

---

### 3. Fail-fast for an occupied port, fail-soft for an invalid one

The PR argues hard that an unclaimable port must abort startup. But an out-of-range value folds to the default and starts anyway:

> `MCP_PORT=99999` ... Out-of-range now folds to the default with a loud error, like the unparseable case.

So the operator who typed one extra digit gets a running service on **8200** — the port this PR documents as colliding with Vault — having only logged an ERROR. That is the softer policy for the *less* ambiguous mistake: an occupied port can be a transient environment condition, whereas `99999` is unambiguously a config error.

The stated reason for folding is that an unbindable port would otherwise stop the whole service "with a message about occupancy that does not describe the problem". That seems solved by emitting the accurate message rather than by continuing — you already distinguish the two cases at the point of the parse, so you have the information needed to say "MCP_PORT=99999 is outside 1-65535" and abort.

Not blocking on my own account, but the two policies should be reconciled or the asymmetry explained, because as written the PR argues for one and implements the other a few lines apart.

---

### Notes

- `mcpPort()` logs through `a.container.Logger` while `bindMCPServer` uses `a.Logger()`. Worth picking one.
- The CodeQL rationale for not echoing the configured value back into the log line is a good catch and reads well.
- `TestBindMCPServer_CanceledContextReportsGracefulShutdown`: the finding that `ListenConfig.Listen` only observes the context while *resolving a name*, so a literal `127.0.0.1` never sees cancellation at all — and that the review-requested behavior therefore needed an explicit `ctx.Err()` check to be reachable — is a genuinely useful piece of follow-through. Measuring it rather than assuming the stated mechanism held is the right instinct.
- Agreed on leaving `grpc.go:183`, `gofr.go` and `factory.go` alone here. The observation that the three statements after `Fatalf` only execute under `nonExitingMockLogger`, so the error counter and `grpc_server_status` gauge are set exclusively in tests and never in the situation they exist to report, is worth its own issue regardless of what happens to this PR.

aryanmehrotra:
@PiyushSingh-ZS @NitinKumar004 — both reviews addressed at `94012476d`, except the scope question, which is a maintainer call. Also: the branch is **no longer behind or conflicted** — it merged `development` and is clean.

## 1. The exit code — fixed here, agreed

You're right that this PR introduces the regression, so this PR should fix it. Exit 0 on a refused startup is the process-level version of the silence the PR argues against: under `restartPolicy: OnFailure`, or in a Job, the service is recorded as having succeeded and never retried.

Your distinction is the one I implemented. `Fatalf` was bad because it was `os.Exit` from library code, mid-setup, with the container's datasources open and nothing unwound. `Run` now exits non-zero at the bottom, **after** `shutdownAfterFailedStartup()` has released everything — top of the call stack, nothing skipped.

**One thing that needed care:** a bare bool couldn't distinguish "this broke" from "the operator stopped it". A SIGTERM during startup already returned `false` from both `handleStartupHooks` and `bindMCPServer`, so a naive non-zero exit would have had every orchestrator retrying a deliberate stop. They return a `startupOutcome` now:

```
startupOK        → continue
startupFailed    → cleanup, exit 1
startupCanceled  → cleanup, exit 0   (SIGINT/SIGTERM during startup)
```

`TestRun_FailedStartupExitsNonZero` and `TestRun_CanceledStartupExitsZero` pin both. The `os.Exit` carries a `//nolint:revive` for `deep-exit` with the argument written out — the logger's `Fatal` carries the same exemption.

## 2. Fail-fast / fail-soft — reconciled your way

Agreed the two policies contradicted each other a few lines apart, and that the softer one was applied to the *less* ambiguous mistake. An occupied port can be a transient condition of the environment; `99999` is wrong now and will be wrong on the next start.

`mcpPort` returns an error for a value that can never be served. `EnableMCP` records it (it runs in the application's own setup, where there is nothing to abort yet) and `bindMCPServer` reports it where it reports a port it could not claim — with the accurate message rather than the occupancy one:

```
ERROR  MCP server cannot start: MCP_PORT is outside the valid port range: 99999 is outside
       1-65535. Set MCP_PORT to a valid, free port, or MCP_PORT=0 to run without the MCP
       transport while keeping tools available in-process.
```

This also closes @NitinKumar004's "malformed `MCP_PORT` silently binds 8200" point, which was the same defect from the other side. The docs claim about the failure path is corrected along with it.

## 3. @NitinKumar004's other items

| Item | Status |
|---|---|
| Out-of-range hard-aborts the app | Fixed — it now aborts *deliberately*, with a message naming the range |
| Disable check too strict (`"00"`, `"+0"`, `" 0 "`) | Fixed — trims, then compares the parsed number |
| `context.Canceled` misreported as a port problem | Fixed — graceful-shutdown message, and `startupCanceled` so the exit code agrees |
| Bind-fail → exit 0, no cleanup | Fixed — cleanup was already there; the exit code is fixed above, and the doc claim corrected |
| Test asserts OS-specific EADDRINUSE text | Already portable — it asserts the framework's remedy text, not the OS error |
| Malformed port needs a test | `TestEnableMCP_UnparseablePortRefusesStartup` |

**The 2 CodeQL alerts on this PR are stale.** `mcp.go` at head contains no `password` reference at all — the only three log lines are `port: %d`.

## 4. ⚠️ Scope — over to you

You're right that three changes are in here, and that (3) changes shutdown behaviour for every app with an `OnStart` hook. **I've left it as one PR rather than splitting it unilaterally** — the rebase argument for splitting no longer applies (no conflicts, not behind), so it's now purely a review/release/revert-granularity question, which is yours to make. Say the word and I'll split it into three.

aryanmehrotra:
@NitinKumar004 — rebased onto `development` (was 38 behind, now current) and re-verified your August review item by item. Every point is addressed; here is where.

| Your item | Where it is now |
|---|---|
| Out-of-range `MCP_PORT` hard-aborts with a misframed "occupancy" remedy | `mcpPort` validates `minTCPPort`/`maxTCPPort` (`default.go`) and returns `errMCPPortOutOfRange` naming the range. It **rejects** rather than folds — see below. Message no longer mentions occupancy for this case. |
| Disable check too strict (`"00"`, `"+0"`, `" 0 "`) | `strings.TrimSpace` then compare the **parsed** integer to `0`, so all four spellings disable. Verified as real processes, not just unit tests. |
| Signal during bind misreported as a port problem | `bind` checks `ctx.Err()` before listening, and `bindMCPServer` has an `errors.Is(err, context.Canceled)` branch that logs the graceful-shutdown line. The explicit pre-check is load-bearing: `ListenConfig.Listen` only observes the context while *resolving a name*, and the address is a literal `127.0.0.1`, so cancellation was otherwise unobservable. |
| Bind-fail → exit 0 and no cleanup; doc overstated the failure path | Both halves fixed. `shutdownAfterFailedStartup()` runs on every abandoned-startup path, and `Run` then reports exit **1** via `abortStartup` — at the bottom of `Run`, after cleanup, not `os.Exit` from mid-setup. Docs rewritten to match. |
| Test asserts OS-specific `"address already in use"` | Gone; the test asserts the framework outcome and says why (`mcp_test.go:369`). |
| Malformed `MCP_PORT` silently binds 8200 | Now `errMCPPortNotANumber`, refuses startup, covered by a test. |

On **fold vs reject** for an invalid port: rejecting, not folding. Folding put the service on 8200 — the port this PR documents as colliding with Vault — for an operator who typed one digit too many, and applied the gentler policy to the *less* ambiguous mistake. An occupied port can be transient; `99999` is wrong on every restart.

`startupOutcome` is a tri-state for the reason you raised: a SIGTERM during startup is not a failed start. `startupCanceled` exits **0**, `startupFailed` exits **1**.

Measured as real OS processes, exit status observed from the parent:

| `MCP_PORT` | outcome | exit |
|---|---|---|
| free port | serves, SIGTERM → clean stop | 0 |
| held by another listener | refuses startup | 1 |
| `not-a-port` | refuses startup | 1 |
| `0` / `00` / `+0` / `" 0 "` | transport disabled, app serves | 0 |
| `65536` / `-1` | refuses startup | 1 |
| `65535` | serves | 0 |
| unset | serves on 8200 | 0 |
| SIGTERM inside an `OnStart` hook | graceful stop | 0 |

`EnableMCP` returned normally in all 12 cases — the process-exit regression it used to cause is gone. Two apps contending for one MCP port: the second exited 1 on **10 of 10** runs, so the bind is deterministic rather than racing.

Also since your review: both CodeQL `go/clear-text-logging` findings (alerts 177/178) are resolved — the raw `Config.Get("MCP_PORT")` value is no longer echoed into any log line. Coverage for `pkg/gofr` 88.7% → 89.3%.

@Umang01-hash added for a second pair of eyes.

akshat-kumar-singhal (approved):
Approve with nits (two should-fix items are release-note/API concerns, not bugs). Listener lifecycle checks out: no listener on bind error; Shutdown before `Run` sets `srv` closes and nils the listener under `srvMu` and `Run` returns; Shutdown between setting `srv` and `Serve` is covered because `http.Server.Serve` defers `l.Close()`; no double close; MCP disabled → `startupOK`. Tests use `127.0.0.1:0`/`GetFreePort`, no hardcoded ports. The MCP/startup subset passes `-race -count=10`, and each of 7 mutations (drop shutdown on bind/hook failure, don't close listener, ignore `stopped`, never/always exit non-zero, ignore `mcpConfigErr`) is caught by a dedicated test.

**Should-fix: a failing OnStart hook now exits the process** — [`run.go:120`](https://github.com/gofr-dev/gofr/blob/1f0630e52e5ecaa5ac95a8800844cd2fc0e19211/pkg/gofr/run.go#L120) → [`gofr.go:104`](https://github.com/gofr-dev/gofr/blob/1f0630e52e5ecaa5ac95a8800844cd2fc0e19211/pkg/gofr/gofr.go#L104)
Before, a failing hook logged and `Run()` returned; now `finishAbandonedStartup` → `abortStartup` calls `os.Exit(1)`. Code after `app.Run()` in `main` is skipped, and a user test that runs `app.Run()` with a failing hook kills the test binary — the `a.exit` override is unexported. That's the same failure mode this PR removes from `EnableMCP`, now for all hooks. At minimum call it out in the PR/changelog and update `docs/advanced-guide/startup-hooks/page.md` (still just says "refuse to start"); better, expose a non-exiting `RunE() error` (or an exported option) and keep `Run` as the thin exiting wrapper.

**Should-fix (release note): stricter `MCP_PORT`** — [`mcp.go:97`](https://github.com/gofr-dev/gofr/blob/1f0630e52e5ecaa5ac95a8800844cd2fc0e19211/pkg/gofr/mcp.go#L97): an invalid or negative value used to fall back to 8200; now the service refuses to start and exits 1. Sound, and the MCP docs cover it, but it's breaking.

**Nit:** a port conflict is only found after OnStart hooks run, so migrations/cache warm-up happen before an obvious bind failure. Binding first and closing the listener if a hook fails would fail faster.

**Nit** — [`run.go:293`](https://github.com/gofr-dev/gofr/blob/1f0630e52e5ecaa5ac95a8800844cd2fc0e19211/pkg/gofr/run.go#L293): `shutdownAfterFailedStartup` ignores a bad `SHUTDOWN_GRACE_PERIOD`, saying it "has already been reported by the caller that reaches Run's normal path" — but on an abandoned startup that path never runs, so it's never reported.

Umang01-hash (approved):
Verified locally at the PR head.

- Fixes real bugs: the old path called `Logger.Fatalf` (os.Exit) from library code mid-setup with datasources still open, and used a dial-based probe that is racy (TOCTOU) and semantically wrong (a wildcard listener elsewhere makes the dial succeed). Now `bind` claims the port with `net.Listen` synchronously before any server goroutine, and a failure aborts Run cleanly via the recorded `mcpConfigErr` / `startupOutcome`, releasing datasources.
- No exported-API change: `mcpPort` sig change is on an unexported method; `mcpConfigErr`, `exit func(int)`, `abortStartup`, `startupOutcome`, and the error vars are all unexported. The `exit` seam is a clean way to test the exit status without taking the test binary down.
- Concurrency is correct: `srv`, `listener`, `stopped` are all read/written under `srvMu`; the bind->Shutdown->Run and bind-fail interleavings each resolve safely (early Shutdown closes the orphaned listener; Run refuses to serve when unbound).
- Tests are thorough and behavioral (occupied-port test binds a real port; canceled-vs-failed exit codes via the seam) — 16 new tests pass under `-race`, full pkg/gofr suite green, golangci-lint 0 new issues. Fail-on-revert confirmed (swallowing the bind error fails the occupied-port abort test).

One thing to consciously sign off on: this is an intentional behavior change — an invalid `MCP_PORT` (non-number or outside 1-65535) now refuses startup rather than silently falling back to 8200. It is documented in the configs and MCP pages and is the right fail-fast direction, but it will surface a previously-silent misconfiguration.

LGTM.

Umang01-hash (commented):
One low-severity edge from a second pass (non-blocking, approval stands):

`bind()` has no `stopped` check, so the ordering `Shutdown()` -> `bind()` -> `Run()` leaks a listener: Shutdown sets `stopped` while `listener` is still nil (nothing to close), `bind()` then claims the port, and `Run()` sees `stopped` and returns without serving — so that listener is never closed.

Not reachable through GoFr's own lifecycle: `bindMCPServer` runs before `startShutdownHandler` arms the signal handler, and `bind()` checks `ctx.Err()` first, so the signal path is safe on both counts. It's only reachable if a caller invokes `app.Shutdown()` from its own goroutine while startup is still in `bind()`. Impact is a single leaked listener FD.

If you want to close it: have `bind()` re-check `m.stopped` under `srvMu` after `Listen` and `l.Close()` + skip the assignment if already stopped. Purely defense-in-depth.

PR #3969, opened 2026-08-18
chore(deps): consolidate minor/patch dependency updates (2026-08-18)

Umang01-hash:
## Consolidated Dependabot updates (minor/patch)

Combines **26** individual Dependabot PRs (root module + 9 submodules + 1 GitHub Action) into one. Only `go.mod` / `go.sum` / `go.work.sum` / `typos.yml` are touched — **no source, no binaries**. `go mod tidy` reproduces every manifest exactly (nothing hand-edited).

### Headline updates
| Package | Old | New | Modules |
|---|---|---|---|
| github.com/aws/aws-sdk-go-v2 | 1.43.3 | 1.43.6 | root, sqs, dynamodb, s3, using-s3-filestore |
| github.com/aws/aws-sdk-go-v2/config | 1.32.34 | 1.32.37 | root, sqs, dynamodb, s3, using-s3-filestore |
| github.com/aws/aws-sdk-go-v2/service/s3 | 1.105.2 | 1.107.2 | root, s3, using-s3-filestore |
| github.com/ClickHouse/clickhouse-go/v2 | 2.47.0 | 2.48.0 | root, clickhouse |
| github.com/nats-io/nats-server/v2 | 2.14.4 | 2.14.5 | root, nats |
| **cloud.google.com/go/cloudsqlconn** | 1.22.1 | **1.25.1** ⚠️ | root, cloudsql |
| golang.org/x/crypto | 0.54.0 | 0.55.0 | root, sftp |
| modernc.org/sqlite | 1.55.0 | 1.56.0 | root |
| go.opentelemetry.io/contrib/.../otelhttptrace | 0.69.0 | 0.70.0 | root, solr |
| **go.opentelemetry.io/otel** (core) | 1.44.0 | 1.45.0 | root (transitive) |
| go.opentelemetry.io/otel/exporters/prometheus | 0.65.0 | 0.67.0 | root |
| crate-ci/typos (Action) | 1.48.0 | 1.49.0 | typos.yml |

Also carried along transitively: `prometheus/client_golang`, `x/net`, `x/text`, `x/sys`, `x/sync`, `google.golang.org/api` (0.287→0.290). "No breaking API changes **reaching GoFr**" — the otel 1.45 core does carry upstream breaking notes, but GoFr is not exposed (see below).

### ⚠️ cloudsqlconn 1.22.1 → 1.25.1 — highest-risk item, needs a live smoke test
1.25.0 **replaced the local generated gRPC client with the GAPIC client library** — it changes the transport used to fetch Cloud SQL instance metadata, and pulls in `cloud.google.com/go/sql v0.1.0` (pre-stable) + `otelgrpc` onto the DB connection path (both confirmed compiled in). Unit tests can't cover a transport swap.
**Before merge:** smoke-test against a real Cloud SQL instance on both public IP and PSC (`settings.go` exercises `WithPSC`/`WithPrivateIP`/`WithPublicIP`).
Pinned to **1.25.1** (released 2026-08-17): fixes a `:443` default-port connectivity regression in the exact path 1.25.0 introduced.

### otel core 1.44 → 1.45 — verified not exposed
- `otel/log` (whose `Kind`/`Value`/`KeyValue` were removed) is unused.
- `otlpmetrichttp` / `otlpmetricgrpc` stay pinned at **1.44.0**, so the 1.45 `WithEndpointURL` path change never reaches `pkg/gofr/metrics/exporters/otlp.go`.

### prometheus exporter 0.65 → 0.67 — two user-visible behaviour changes (release-note)
Pre-1.0, permitted; GoFr's own metrics are unaffected (it only emits String/Int64/Bool). But via the public `RecordHistogramAttrs`:
- label values now format via `attribute.Value.String()` not `.Emit()` — a `StringSlice` label goes `["a","b"]` → `[a,b]`, and `+Inf` → `Infinity`.
- scope attributes colliding with reserved `otel_scope_*` labels are now silently dropped.

### Closes
#3943 #3944 #3945 #3946 #3947 #3948 #3949 #3950 #3951 #3952 #3953 #3954 #3955 #3956 #3957 #3958 #3959 #3960 #3961 #3962 #3963 #3964 #3965 #3966 #3967 #3968


aryanmehrotra (commented):
Verified this against a real checkout at `5bea5a47e` rather than reading the diff — every claim in the PR body about scope and integrity holds up. Summary: **no source changes, no binaries, no breaking changes reaching GoFr, builds and tests green.** Two things worth acting on before merge.

## What I verified

| Check | Result |
|---|---|
| Files touched | 22: only `go.mod`/`go.sum` (10 modules), `go.work.sum`, `typos.yml`. No source. |
| Binaries | Zero binary diffs (`--numstat` shows no `-  -` rows) — all text |
| `go mod verify` | `all modules verified` ×10, against public `sum.golang.org` (`GOPRIVATE`/`GOFLAGS` empty) |
| `go mod tidy` ×10 | **zero diff** — manifests are exactly what tidy generates, not hand-edited |
| Downgrades / major-version bumps | none |
| `go` / `toolchain` / `replace` / `module` directive edits | none |
| `go.work.sum` | additive-only (176/0); workspace resolution exits 0, no missing hashes |
| Build, `GOWORK=off`, all 10 modules | exit 0 — each submodule builds standalone, as a consumer receives it |
| `go vet ./...` | clean |
| Tests | all 9 touched submodules + `examples/using-s3-filestore` pass |
| govulncheck | identical to `development` (9 stdlib / 4 imported / 1 required); **zero** reachable dependency-level vulns |

Two test failures I chased down are **not** from this PR: `pkg/gofr` fails only because local port 8200 was held by an unrelated process (`development` fails identically), and `TestHTTPService_HealthCheckTimeout` flaked only under CPU contention, passing 3/3 on an idle machine.

## 1. Recommend bumping cloudsqlconn to 1.25.1

The PR pins **1.25.0**. **1.25.1 was released 2026-08-17**, one day before this PR, and fixes a connectivity bug — *"ensure the SqlDataEndpoint includes the default port `:443`"* — in the exact code path 1.25.0 introduced. Dependabot simply opened its PR before 1.25.1 landed.

I can't determine from here whether that bug reaches GoFr's dialer path specifically, so I've left this as a recommendation rather than a block — but since we're already moving three minor versions, landing on 1.25.1 seems strictly better.

## 2. cloudsqlconn 1.22.1 → 1.25.0 is not a routine minor bump

This is the highest-risk item in the PR and I think it deserves a call-out in the description. Per the upstream changelog, 1.25.0 *"replaced the local generated gRPC client with the GAPIC client library"* — i.e. it changes the transport used to fetch Cloud SQL instance metadata.

It adds two genuinely-linked dependencies to `pkg/gofr/datasource/cloudsql`:

- `cloud.google.com/go/sql` **v0.1.0** — a pre-stable module now sitting on the DB connection path
- `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.67.0`

I confirmed via `go list -deps` that both are actually compiled in — unlike the `cncf/xds` / `envoyproxy` / `vtprotobuf` entries that also appear in `go.sum`, which are graph-only hashes and never linked.

Unit tests can't cover a transport swap. Worth a live smoke test against a real Cloud SQL instance (both public IP and PSC, since `settings.go` exercises `WithPSC`/`WithPrivateIP`/`WithPublicIP`) before this ships.

## 3. The dependency table omits some bumps

"No breaking API changes expected" is a broader claim than the table supports. Not listed but present: **`go.opentelemetry.io/otel` 1.44.0 → 1.45.0** (which *does* carry documented breaking changes), plus `prometheus/client_golang`, `x/net`, `x/text`, `x/sys`, `x/sync`, and `google.golang.org/api` 0.287 → 0.290.

I checked GoFr's exposure to the otel 1.45 breaking changes and it is **clear**:
- `go.opentelemetry.io/otel/log` (whose `Kind`/`Value`/`KeyValue` were removed) is unused.
- `otlpmetrichttp` / `otlpmetricgrpc` stay pinned at **v1.44.0**, so the 1.45 change where `WithEndpointURL` *"no longer appends the default signal path"* never reaches the `otlpmetrichttp.WithEndpointURL` call at `pkg/gofr/metrics/exporters/otlp.go:62`. Had those exporter modules also moved to 1.45, a scheme'd path-less `METRICS_URL` would have silently started POSTing to `/` instead of `/v1/metrics`.

Worth adding the otel core bump to the table since it's the one with real breaking-change notes.

## 4. Prometheus exporter 0.65 → 0.67 — two silent behaviour changes

Pre-1.0 module, so these are permitted, and neither affects GoFr's own metrics — but they're user-visible:

- Label values now format via `attribute.Value.String()` instead of `.Emit()`. Identical for `STRING`/`INT64`/`BOOL`, which is all GoFr itself emits (`attribute.String` throughout `register.go`). It differs for slice-typed and ±Inf float attributes, which users **can** pass through the public `RecordHistogramAttrs(ctx, name, value, attrs ...attribute.KeyValue)`: a `StringSlice` label goes from `["a","b"]` to `[a,b]`, and `+Inf` becomes `Infinity`. Narrow, but enough to move someone's dashboard.
- Scope attributes that collide with reserved `otel_scope_*` labels are now silently dropped.

Probably just release-note material rather than anything to change here.

## 5. Non-blocking

`go work sync` leaves 62 files divergent on this branch vs 46 on `development` — pre-existing drift that this widens slightly, not something introduced here.

Umang01-hash:
Thanks for the deep pass 🙏

- **cloudsqlconn → 1.25.1** — bumped (fixes the `:443` regression, verified it's the exact 1.25.0 path).
- **Description** — updated: added otel core 1.45 + transitives, reframed to "no breaking changes reaching GoFr", and documented the prometheus 0.65→0.67 label-format / dropped-scope-attr changes as release notes.
- **cloudsql GAPIC transport swap** — called out in the description; unit tests can't cover it, so a live Cloud SQL smoke test (public IP + PSC) is pending before merge.

Scope table drift (#5) left as-is — pre-existing, not introduced here.

aryanmehrotra (approved):
Re-reviewed at `19357164c` (was `5bea5a47e`). All four actionable items from the last pass are resolved, and I re-ran the full battery rather than taking the fixes on trust. **LGTM** — with the Cloud SQL smoke test still owed before merge, as you noted.

## The two new commits

`19357164c` is exactly the right shape — 2 files, 3 lines:

```
pkg/gofr/datasource/cloudsql/go.mod | 2 +-   cloudsqlconn v1.25.0 -> v1.25.1
pkg/gofr/datasource/cloudsql/go.sum | 4 ++--  matching hash swap
```

No collateral. `1.25.1` is still upstream latest. I also checked `go.work.sum` wasn't left stale by the bump — it never carried a cloudsqlconn entry at all (the hash lives in the submodule's own `go.sum`), and workspace resolution exits 0 with no missing-hash errors, so nothing to update there.

`cf67edc03` merges `development` in, bringing #3759's trie router. The branch now fully contains `development` at `878b02658` (merge-base == development head). Importantly, the PR's **net** diff against the new merge-base is still exactly the same 22 manifest files — no source, no binaries. The merge added nothing of its own.

## Re-verification at `19357164c`

| Check | Result |
|---|---|
| Net diff vs merge-base | 22 files, manifests only; no source, no binaries |
| `go mod verify` ×10 | `all modules verified` |
| `go mod tidy` ×10 | **zero diff** — still reproduces every manifest exactly |
| Downgrades / major bumps | none |
| Build `GOWORK=off` ×10 | all OK — each submodule still builds standalone |
| `go vet ./...` | clean |
| Submodule tests (9) + `using-s3-filestore` | all pass |
| Root suite `./pkg/...` | **29/29 packages green** |

The only root-suite package not green is `pkg/gofr`, which fails solely on `FATAL: MCP port 8200 is blocked or unreachable` — port 8200 is held by an unrelated process on my machine. `development` at `878b02658` fails identically, and since the branch now contains exactly that commit, it's the same environment issue, not this PR. `TestHTTPService_HealthCheckTimeout` also behaved this round (it only flaked earlier under CPU contention).

## On the description update

The rewritten body is a genuine improvement over "no breaking API changes expected" — "no breaking changes **reaching GoFr**" is the accurate claim, and adding the otel core 1.45 row plus the transitives means a future reader can see why the exposure question was asked at all. Documenting the prometheus `Emit()`→`String()` label-format change as a release note is the right home for it; that one would otherwise have surfaced as a mystery dashboard break for someone passing slice-typed attrs to `RecordHistogramAttrs`.

Agreed on leaving the `go work sync` drift alone — pre-existing, and widening it is an artifact of bumping the root module, not a defect here.

## Remaining before merge

The cloudsqlconn GAPIC transport swap is still the one risk no automated check on this PR can clear — `1.25.1` fixes the `:443` regression, but the underlying change from the local generated gRPC client to the GAPIC library is still in there, with `cloud.google.com/go/sql v0.1.0` on the connection path. Your pending live smoke test (public IP + PSC) is the right gate. Approving on the code, with that as the condition.

PR #3970, opened 2026-08-18
perf(tracer): stop rebuilding span metadata on every request

aryanmehrotra:
**TL;DR — 8 fewer allocations per request. No behaviour change.**

## The problem

The tracing middleware runs on **every request**. For each one it rebuilt things that never change for a given route:

- the span name (`"GET /users/{id}"`) — via `fmt.Sprintf`
- the attribute slice (`http.request.method`, `http.route`)
- the `trace.WithAttributes` option wrapper **and** the variadic slice holding it
- the status attribute — even when the span was **non-recording** and threw it away
- a response-writer wrapper — even when nothing recorded the status
- propagation headers — recanonicalised on every lookup

For a fixed route table, all of that produces byte-identical results every single time.

## The fix

Build it once per `(method, route)` and reuse it. Six focused commits, one idea each.

Cache key is the **route template** (`/users/{id}`), never the concrete path. Unmatched requests carry a raw, attacker-controlled path and are never cached.

> **Revised after review.** That paragraph used to end "so it is bounded by routes × methods", and **it wasn't**. Neither half of that product bounds itself in a real GoFr app:
>
> - `gofr.go` registers a `PathPrefix("/")` catch-all, so `mux.CurrentRoute` is set for **every** request and `GetPathTemplate()` returns `"/"` even for a path nothing matched. The `templated` guard was therefore **always true** — reproduced with method `ZZZQUUX` on `/totally/unknown`.
> - `net/http` accepts **any RFC 7230 token** as a method, so a client streaming `M00001`, `M00002`, … minted a permanent `*spanMeta` per request. Reproduced: 5000 distinct methods against the catch-all produced 5000 permanent entries.
>
> That is a remotely-triggerable unbounded-memory DoS, and the pre-cache `fmt.Sprintf` path retained **zero** state — so it was introduced here, not inherited. Two things bound it now:
>
> 1. **`cacheableMethod`** restricts the key to the nine defined HTTP methods, which turns "methods × routes" back into the fixed quantity the cache was designed around. A made-up method still works — it just rebuilds its span metadata, exactly as every request did before the cache existed.
> 2. **`routeCache`** caps the whole thing at 4096 entries, because the first bound is an argument about reachability and process memory should rest on a number too.
>
> `routeCache` and `cacheableMethod` live in their own file because the metrics caches in #3972 have the same shape on the same key space, and the review asked for all three to be bounded consistently rather than one at a time. **#3972 is now stacked on this PR.**

## The numbers

Two benchmarks ship with this, because a sampling service runs both paths and neither may regress:

| | before | after |
|---|---|---|
| **non-recording** provider | 21 allocs, 2074 B | **11 allocs, 1568 B** |
| **recording** provider | 23 allocs, 2875 B | **17 allocs, 2658 B** |

Allocation counts only — see *Measurement* below.

> **The non-recording row was re-measured after review**, and it improved: it previously read 13 allocs / 1760 B. `BenchmarkTracer` documented the non-recording path but **did not measure it** — `TestTracerPropagatesIncomingTraceContext` installed a live recording `TracerProvider` and never restored it, and tests run before benchmarks, so the benchmark ran *recording* and fed that test's in-memory exporter. It now pins `noop.NewTracerProvider()` itself, and that test restores both globals it replaces (the provider and the Baggage-less propagator).

## Why it matters

Tracing is on the hot path of every request in every GoFr service. This is the single largest per-request allocation saving in the middleware chain.

## Baggage regression, found in review and fixed

`headerCarrier` replaces `propagation.HeaderCarrier` to avoid canonicalising the lookup key per request — but implemented only `Get`/`Set`/`Keys`. The stdlib type it replaces **also satisfies `propagation.ValuesGetter`**, and `propagation.Baggage.Extract` type-asserts to that interface to combine **all** values of the `Baggage` header. The assertion failed, so `Extract` fell back to the single-value `Get` and **silently dropped every baggage member after the first**.

It only surfaces when a request carries more than one `Baggage` header — legal per W3C, and what proxies and service meshes commonly emit — and GoFr installs `propagation.Baggage` in its default composite propagator, so the path is live. Reproduced against the stdlib carrier with three `Baggage` headers:

```
stdlib HeaderCarrier : members=3  "k2=v2,k3=v3,k1=v1"
gofr   headerCarrier : members=1  "k1=v1"
implements ValuesGetter: gofr=false stdlib=true
```

Fixed by implementing `Values`. It deliberately does **not** use the canonical-key fast path: `baggage` is not one of the keys that map covers, and a carrier replacing a stdlib one has to be a faithful drop-in first and an optimisation second.

The pre-existing equivalence test compared only `Get` and `Keys`, so it could not catch this. The new test compares extracted baggage **member by member** against the stdlib carrier across four header shapes, and asserts the interface is satisfied.

`BenchmarkTracer_WithPropagationHeaders` closes the other gap: the existing benchmarks send a bare request, so the propagators find nothing and the canonicalisation this PR avoids never runs — their delta is the per-route cache and the `IsRecording` gate alone. The new one sends `traceparent`, `tracestate` and two `Baggage` headers, which is what a service behind a mesh actually receives.

## What the cache costs

The saving is bought with resident memory: the cache is process-global and entries are never released, because they are pure functions of their key and stay valid for the life of the process.

Measured with `runtime.MemStats` around a filled cache, stable across three runs:

| Shape | Entries | Heap held |
|---|---|---|
| small app — 25 routes x GET/POST | 50 | **+22 KB** |
| typical app — 100 routes x GET/POST | 200 | **+83 KB** |
| large app — 500 routes x 4 methods | 2000 | **+825 KB** |
| at the `routeCacheLimit` ceiling | 4096 | **+1735 KB** |

So a realistic service holds tens of kilobytes, and the worst case any service can reach is **~1.7 MB**, permanently. That ceiling is the point of `routeCacheLimit` — without it the table has no last row.

It is a genuine trade rather than a free win, and it is the one thing in this PR that costs something. Weigh it against ~9 allocations saved on every request.

## Safety

- **No behaviour change.** Span names, attributes and propagation are identical.
- The `IsRecording()`-gated non-recording branch — the path this PR optimizes — **now has a test**; it had none.
- `buildSpanName`'s godoc no longer quotes a `44.3 → 22.6 ns/op` figure: the `fmt.Sprintf` baseline it was measured against is gone, so nothing in the tree can reproduce it.
- New tests for the bound: an undefined method mints no entry, standard methods still hit the cache, and the cap holds.
- Cached data is pure and provider-independent, so it stays correct across `TracerProvider` replacement.

## Measurement

Only allocations and bytes are quoted. Wall-clock on the machine used varied by **more than 5×** between runs of the same binary, so ns/op there is meaningless. Allocation counts were stable and moved monotonically with each commit in the series.

---
*Independent of the other `perf/*` PRs — touches only `tracer.go` and its test. Merge in any order.*

### DIFF
--- a/pkg/gofr/http/middleware/routecache.go
+++ b/pkg/gofr/http/middleware/routecache.go
@@ -1,0 +1,101 @@
+package middleware
+
+import (
+	"net/http"
+	"sync"
+	"sync/atomic"
+)
+
+// routeCacheLimit caps the number of entries any per-route cache in this package
+// will hold.
+//
+// With cacheableMethod filtering the method half of the key, the reachable key
+// space is (standard methods x registered routes), which is fixed at startup —
+// so this cap is a backstop for a pathologically large route table, not the
+// primary bound. It is a backstop that matters: the primary bound is an argument
+// about reachability, and an argument is a weaker thing to rest process memory on
+// than a number.
+const routeCacheLimit = 4096
+
+// cacheableMethod reports whether method may be used as part of a cache key.
+//
+// This is the load-bearing half of every per-route cache's bound, and it exists
+// because the "only bounded route templates are cached" claim does not survive
+// contact with a real GoFr app. gofr.go registers a PathPrefix("/") catch-all, so
+// mux.CurrentRoute is set for EVERY request and GetPathTemplate() returns "/"
+// even for a path no route matched — the templated guard is therefore true
+// always, and the route half of the key is not the bound it looks like.
+//
+// The method half is worse, because it is not drawn from a fixed set at all:
+// net/http accepts any RFC 7230 token as a method, so an unauthenticated client
+// streaming M00001, M00002, ... mints a permanent cache entry per request and
+// grows resident memory without bound. Restricting the key to methods that are
+// actually defined turns "methods x routes" back into the fixed quantity the
+// caches were designed around.
+//
+// A request with a non-standard method still works — its per-route data is built
+// fresh, exactly as it was before any cache existed. It just never persists.
+func cacheableMethod(method string) bool {
+	switch method {
+	case http.MethodGet, http.MethodHead, http.MethodPost, http.MethodPut,
+		http.MethodPatch, http.MethodDelete, http.MethodConnect,
+		http.MethodOptions, http.MethodTrace:
+		return true
+	default:
+		return false
+	}
+}
+
+// routeKey identifies a (method, route-template) pair.
+type routeKey struct {
+	method, route string
+}
+
+// routeCache is a bounded, concurrent cache keyed on routeKey.
+//
+// Reads stay on sync.Map's lock-free path. Writes stop at routeCacheLimit rather
+// than evicting: the entries are pure functions of their key and are all equally
+// valid forever, so there is no staleness to evict for, and refusing to grow is
+// both cheaper and easier to reason about than an eviction policy. Past the cap
+// the affected requests rebuild their data per request — the optimization stops
+// applying, but memory is flat and behavior is unchanged.
+type routeCache struct {
+	entries sync.Map
+	count   atomic.Int64
+}
+
+// load returns the cached value for key, if present.
+func (c *routeCache) load(key routeKey) (any, bool) {
+	return c.entries.Load(key)
+}
+
+// store caches value under key unless the cache is already at its limit.
+//
+// The count is advisory: concurrent stores can race past the cap by the number of
+// goroutines in flight, which is bounded and harmless. What matters is that the
+// cache cannot grow without limit.
+func (c *routeCache) store(key routeKey, value any) {
+	if c.count.Load() >= routeCacheLimit {
+		return
+	}
+
+	if _, loaded := c.entries.LoadOrStore(key, value); !loaded {
+		c.count.Add(1)
+	}
+}
+
+// len reports the number of cached entries. Test-only.
+func (c *routeCache) len() int64 {
+	return c.count.Load()
+}
+
+// reset empties the cache. Test-only: the package-level caches are process-wide,
+// so a test that asserts on size has to start from a known state.
+func (c *routeCache) reset() {
+	c.entries.Range(func(k, _ any) bool {
+		c.entries.Delete(k)
+
+		return true
+	})
+	c.count.Store(0)
+}
--- a/pkg/gofr/http/middleware/tracer.go
+++ b/pkg/gofr/http/middleware/tracer.go
@@ -1,7 +1,6 @@
 package middleware
 
 import (
-	"fmt"
 	"net/http"
 	"strings"
 
@@ -26,6 +25,17 @@ import (
 // the heap (verified via go build -gcflags='-m=2'). Factoring the
 // constructor into its own function does not avoid the alloc — it is
 // still useful as a single source of truth for the attribute key.
+// buildSpanName renders the OTel HTTP semconv span name ("GET /users/{id}").
+//
+// Concatenation rather than fmt.Sprintf. Both cost the one unavoidable
+// allocation for the result string, so this is a CPU saving only, not an
+// allocation saving: Sprintf walks a format string and boxes two arguments that
+// are already strings. Measured by BenchmarkBuildSpanName on this call site:
+// 44.3 ns/op -> 22.6 ns/op, both at 16 B/op and 1 alloc/op.
+func buildSpanName(method, route string) string {
+	return method + " " + route
+}
+
 func methodKV(method string) attribute.KeyValue {
 	return attribute.String("http.request.method", method)
 }
@@ -77,7 +87,7 @@ func Tracer(inner http.Handler) http.Handler {
 		// value per concrete path (e.g. "/users/42"). Fall back to URL.Path
 		// when no route matched (404 / unknown route).
 		route := routeTemplate(r)
-		spanName := fmt.Sprintf("%s %s", method, route)
+		spanName := buildSpanName(method, route)
 
 		ctxOut, span := tr.Start(ctx, spanName, trace.WithAttributes(
 			methodKV(method),
--- a/pkg/gofr/http/middleware/tracer.go
+++ b/pkg/gofr/http/middleware/tracer.go
@@ -116,6 +116,17 @@ func Tracer(inner http.Handler) http.Handler {
 		// implicit 200 in that case, so the span attribute must report 200
 		// rather than be omitted (or worse, recorded as 0).
 		defer func(s trace.Span, rw *StatusResponseWriter) {
+			// A non-recording span discards SetAttributes by contract, but the
+			// call still builds the variadic []attribute.KeyValue to hand it.
+			// On the default deployment every span is non-recording — GoFr
+			// installs an SDK provider with NeverSample when no TRACE_EXPORTER
+			// is set — so that slice was allocated and thrown away on every
+			// request. Recording spans are unaffected; see
+			// TestTracerStatusAttributeStillRecorded.
+			if !s.IsRecording() {
+				return
+			}
+
 			s.SetAttributes(attribute.Int("http.response.status_code", rw.Status()))
 		}(span, srw)
 
--- a/pkg/gofr/http/middleware/tracer.go
+++ b/pkg/gofr/http/middleware/tracer.go
@@ -127,6 +127,14 @@ func newSpanMeta(method, route string) *spanMeta {
 	}
 }
 
+// The lowercase header names the W3C propagators look up. They are the keys the
+// propagators pass to the carrier, not the spellings that go on the wire.
+const (
+	headerTraceparent = "traceparent"
+	headerTracestate  = "tracestate"
+	headerBaggage     = "baggage"
+)
+
 // canonicalPropagationKeys maps the lowercase header names the W3C propagators
 // look up to their canonical spellings.
 //
@@ -137,9 +145,9 @@ func newSpanMeta(method, route string) *spanMeta {
 //
 //nolint:gochecknoglobals // immutable, process-wide header constants.
 var canonicalPropagationKeys = map[string]string{
-	"traceparent": textproto.CanonicalMIMEHeaderKey("traceparent"),
-	"tracestate":  textproto.CanonicalMIMEHeaderKey("tracestate"),
-	"baggage":     textproto.CanonicalMIMEHeaderKey("baggage"),
+	headerTraceparent: textproto.CanonicalMIMEHeaderKey(headerTraceparent),
+	headerTracestate:  textproto.CanonicalMIMEHeaderKey(headerTracestate),
+	headerBaggage:     textproto.CanonicalMIMEHeaderKey(headerBaggage),
 }
 
 // headerCarrier adapts http.Header for the OTel propagators without paying to
@@ -163,6 +171,25 @@ func (c headerCarrier) Get(key string) string {
 }
 
 func (c headerCarrier) Set(key, value string) { http.Header(c).Set(key, value) }
+
+// Values returns every value for key, satisfying propagation.ValuesGetter.
+//
+// Not optional. propagation.Baggage.Extract type-asserts the carrier to
+// ValuesGetter and, when it matches, combines ALL values of the Baggage header;
+// the stdlib propagation.HeaderCarrier this type replaces satisfies it. Without
+// this method the assertion fails and Extract silently falls back to the
+// single-value Get, dropping every baggage member after the first whenever a
+// request carries more than one Baggage header -- which is legal per W3C and is
+// what proxies and service meshes commonly emit. GoFr installs
+// propagation.Baggage in its default composite propagator, so that path is live.
+//
+// Measured against the stdlib carrier with three Baggage headers: stdlib
+// extracted 3 members, this carrier extracted 1 before the method existed.
+//
+// The canonical-key fast path is deliberately not used here. Baggage is not one
+// of the keys canonicalPropagationKeys covers, and a carrier that replaces a
+// stdlib one has to be a faithful drop-in first and an optimization second.
+func (c headerCarrier) Values(key string) []string { return http.Header(c).Values(key) }
 
 func (c headerCarrier) Keys() []string {
 	keys := make([]string, 0, len(c))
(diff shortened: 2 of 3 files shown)


PiyushSingh-ZS:
Reviewed against `development`, built and ran the middleware tests on the branch. The caching logic is correct, and the two things I most expected to be wrong here turn out to be fine:

- **Sharing `startOpts` / `attrs` across concurrent spans is safe.** `attributeOption.applySpan` appends into the *span config's* slice, which starts nil, so the append allocates rather than writing into the shared array. The SDK likewise appends into the new span's own (empty) attribute slice. Nothing mutates the cached slice.
- **The cache key is sound.** Keying only on resolved route templates, with `templated` gating entry into the map, is what keeps this bounded — and the `!templated` early return means an unmatched, attacker-controlled path never enters it. The `bool` return from `routeTemplate` is the right shape for that.
- **`headerCarrier`.** Behaviour matches `propagation.HeaderCarrier` for the three W3C keys (Go canonicalizes incoming header keys at parse time, so `Traceparent` is what's actually in the map), and unrecognized keys correctly fall through to `Header.Get`, so B3/Jaeger-style custom propagators keep working. `Keys()` matches otel's implementation exactly.

### `spanMeta.attrs` is dead

It's assigned in `newSpanMeta` and never read — `startOpts` is the only consumer. Either drop the field or, if it's kept for a future `SetAttributes` path, say so, because right now it reads as if something still uses it.

### The `IsRecording()` gate changes what downstream sees

The reasoning is right — Tracer is outermost, so the assertion always failed and every request allocated a wrapper that a NeverSample deployment never read. I confirmed the chain order (`http_server.go:66`: Tracer → Logging → CORS → Metrics), and that dropping the wrap is harmless there: Logging wraps immediately after, and Metrics reuses Logging's wrapper.

The part worth documenting is that the **identity of the ResponseWriter seen downstream now depends on sampling**. Any user middleware inserted between Tracer and Logging that type-asserts `*StatusResponseWriter` will behave differently on a sampled vs unsampled request, which is a genuinely hard bug to find. One line in the comment saying so would pay for itself.

### Minor

`tracerSpanCache` is package-level with no eviction. Bounded by methods × routes in production, so fine — but it also persists across `App` instances within a single test binary, which is worth knowing if anyone later asserts on cache contents in a test.

NitinKumar004 (changes requested):
The caching direction is reasonable, but as written this introduces a remotely-triggerable unbounded-memory DoS. Blocking on that.

### 🔴 Blocker — unbounded span cache → remote OOM
`tracerSpanCache` is a package-global `sync.Map` keyed on `(method, route-template)` with **no eviction**. The safety claim is that only bounded route templates are cached and raw/unmatched paths never are — but that guard is **dead in a real GoFr app**:
- GoFr registers a catch-all `router.PathPrefix("/")`, so `mux.CurrentRoute(r)` is set for **every** request and `GetPathTemplate()` returns `("/", nil)` even for unknown paths (verified with method `ZZZQUUX`, path `/totally/unknown`). So `templated == true` always, and everything is cached under `{method, "/"}`.
- `method = strings.ToUpper(r.Method)` is **not** restricted — Go's server accepts any RFC-7230 method token. An unauthenticated client streaming distinct methods (`M00001`, `M00002`, …) mints a permanent `*spanMeta` each → resident memory grows without bound → OOM.

The pre-PR `fmt.Sprintf` path retained **zero** state, so this DoS vector is newly introduced, and the "no behaviour change / deliberately never cached" claims don't hold. **Fix:** don't cache the `/` catch-all (or any prefix catch-all); and/or cap+evict; and/or only cache routes registered via explicit `Methods().Path()`.

### Medium
- **`BenchmarkTracer` measures the recording path, not the documented non-recording path.** A prior test (`TestTracerPropagatesIncomingTraceContext`) installs a live recording `TracerProvider` and never restores noop; tests run before benchmarks, so `BenchmarkTracer` runs with `IsRecording()==true` and feeds the leaked in-memory exporter. The "non-recording: 13 allocs" headline only reproduces in isolation. Install `noop.NewTracerProvider()` with `b.Cleanup`, like `BenchmarkTracer_Recording`.

### Low
- The `IsRecording()`-gated non-recording branch (the path this PR optimizes) has **no test**.
- `TestTracerPropagatesIncomingTraceContext` also leaks the global `TextMapPropagator` (Baggage-less) with no `Cleanup`, making the suite order-dependent.
- `buildSpanName` godoc + test enshrine a `44.3 → 22.6 ns/op` figure that no shipped benchmark can reproduce (the `fmt.Sprintf` baseline was removed).

### ⚠️ Cross-PR (HIGH, coordinate with #3972)
This is **not the only cache** with this shape. #3972 adds `optionRecorder.cache` (bounded 4096) and `attrsRecorder.cache` (**unbounded**) on the same `(path, method, status)` key space. The same malicious stream inflates all three. Fixing this cache in isolation is insufficient — bound all three consistently, reviewed as one unit. Also overlaps #3770 in the `StatusResponseWriter` block.

**Requested changes:** eliminate the unbounded-cache DoS (blocker); fix the benchmark + propagator leaks; add a non-recording test; and coordinate cache-bounding with #3972.

Umang01-hash (changes requested):
Strong, careful perf PR — I verified the risky parts hold: the shared per-route `spanMeta` cache is concurrency-safe (`trace.WithAttributes` only reads the cached slice; otel applies it via `append(cfg.attributes, o...)`, and `-race` passes), the `IsRecording` gate is correct (`http_server.go` registers `Tracer` before `Logging`, so Tracer is outermost and the old `StatusResponseWriter` assertion always failed — the stale comment was wrong), and the tracer cache is genuinely DoS-bounded (`cacheableMethod` + `routeCacheLimit`). `gofmt`/`vet`/`-race`/benchmarks all green. One regression to fix before merge, plus two notes.

### Baggage-loss regression (introduced here) — please fix
`headerCarrier` replaces `propagation.HeaderCarrier` but implements only `Get`/`Set`/`Keys`, not `Values(string) []string`. In otel v1.45.0, `propagation.Baggage.Extract` type-asserts the carrier to `ValuesGetter` — the stdlib `HeaderCarrier` satisfies it and combines **all** `Baggage` header values; `headerCarrier` fails the assertion, so Extract falls back to single-value `Get("baggage")` and silently drops every baggage member after the first when a request carries multiple `Baggage` headers (legal per W3C; commonly emitted by proxies/meshes). GoFr installs `propagation.Baggage` in the default composite propagator, so this is live.

Reproduced locally: stdlib carrier → 2 members, `headerCarrier` → 1. `TestHeaderCarrierMatchesHeaderCarrier` only compares `Get`/`Keys`, so it doesn't catch it. Fix is one line:
```go
func (c headerCarrier) Values(key string) []string { return http.Header(c).Values(key) }
```
plus a multi-`Baggage` equivalence assertion. A custom carrier replacing a stdlib one needs to be a faithful drop-in.

### Note — the metrics sibling DoS (`metrics.go` `routeAttrs`) is still open
Not introduced here, and your commit message already calls it out for #3972 (the reusable `routeCache` is built for exactly this). Just flagging that it's the same live unbounded-key DoS — an unauthenticated client streaming distinct RFC-7230 method tokens grows `routeAttrs` without bound (repro'd ~61MB/200k requests). Landing this without #3972 leaves that hole open, so it'd be good to land them together or back-to-back.

### Nit — benchmark doesn't exercise `headerCarrier`
`BenchmarkTracer`/`_Recording` set no propagation headers, so they never hit the canonicalization optimization — the measured alloc delta is from the cache + `IsRecording` only. The `headerCarrier` win is real by inspection but unproven by the bench; a header-bearing variant would close that.

No breaking API change (all new symbols unexported). Nice work overall — the DoS analysis in the doc comments is excellent.

aryanmehrotra:
Both blocking findings are fixed. Re-requesting review — the `CHANGES_REQUESTED` verdicts on this PR are stale: they sit on `7c3d4abed` and `4eed1cc03`, and the head is now `16ee698fc`.

### @NitinKumar004 — unbounded span cache → remote OOM

Your analysis was exactly right, including the part that made it dangerous: the `templated` guard is dead in a real GoFr app, because `gofr.go` registers a `PathPrefix("/")` catch-all so `GetPathTemplate()` returns `"/"` for every request. I reproduced it before fixing — 5000 distinct method tokens against the catch-all minted 5000 permanent entries.

Two bounds now, because the first is an argument about reachability and process memory should rest on a number too:

- **`cacheableMethod`** restricts the key to the nine defined HTTP methods, which turns "methods × routes" back into the fixed quantity the cache was designed around. A made-up method still works — it just rebuilds its span metadata, exactly as every request did before the cache existed.
- **`routeCache`** caps the whole thing at 4096 entries.

Verified on the current head:

```
5000 made-up methods       -> cache entries = 0
5000 requests, one route   -> cache entries = 1
cap enforced               -> 4096 (limit 4096)
```

You also asked that this be coordinated with #3972 rather than fixed in isolation. It is: `routeCache` and `cacheableMethod` live in their own file precisely so the two metrics caches use them under the same rules, and **#3972 is stacked on this PR**. `attrsRecorder.cache`, which had no ceiling at all, goes through the same bound there.

The other items are done too — `BenchmarkTracer` pins `noop.NewTracerProvider()` itself and `TestTracerPropagatesIncomingTraceContext` restores both globals it replaces, so the benchmark measures the non-recording path it documents (11 allocs/op); the `IsRecording()`-gated branch has a test; and `buildSpanName`'s godoc no longer quotes a `44.3 → 22.6 ns/op` figure that nothing in the tree can reproduce.

### @Umang01-hash — `headerCarrier` drops baggage members

Confirmed and fixed. I reproduced it against the stdlib carrier before changing anything: three `Baggage` headers gave stdlib 3 members and `headerCarrier` 1, because `propagation.Baggage.Extract` type-asserts to `ValuesGetter` and the assertion failed, falling back to the single-value `Get`.

```
before:  stdlib=3  gofr=1  implementsValuesGetter=false
after:   stdlib=3  gofr=3  implementsValuesGetter=true
```

`Values` deliberately does **not** use the canonical-key fast path — `baggage` is not one of the keys that map covers, and a carrier replacing a stdlib one has to be a faithful drop-in first and an optimisation second.

The pre-existing equivalence test compared only `Get` and `Keys`, which is why it could not catch this. The new test compares extracted baggage **member by member** against the stdlib carrier across four header shapes, and asserts the interface is satisfied.

`BenchmarkTracer_WithPropagationHeaders` closes your third note: the existing benchmarks send a bare request, so the propagators find nothing and the canonicalisation this PR avoids never runs. The new one sends `traceparent`, `tracestate` and two `Baggage` headers — what a service behind a mesh actually receives.

---

Rebased onto current `development`, 0 behind. `-race` clean on the http tree, `golangci-lint --new-from-rev` reports 0 issues.

aryanmehrotra:
Rebased onto current `development` (0 behind), which now carries #3770, #3974, #4108 and #4109. Re-requesting review — every requested change is in, and I re-measured against the new baseline rather than carrying the old numbers forward.

### Requested changes, and where they landed

| Request | Where |
|---|---|
| **Blocker** — unbounded span cache → remote OOM | `tracer.go:100`, `if !templated \|\| !cacheableMethod(method)`, plus `routeCacheLimit` in `routecache.go` |
| **Blocker** — `headerCarrier` drops baggage after the first `Baggage` header | `tracer.go:192`, `func (c headerCarrier) Values(key string) []string` |
| `BenchmarkTracer` measured the recording path | `BenchmarkTracer` now installs `noop.NewTracerProvider()` and restores it in `b.Cleanup`, with a comment on why package state made it lie |
| No benchmark exercised `headerCarrier` | `BenchmarkTracer_WithPropagationHeaders` — traceparent, tracestate and **two** `Baggage` headers |
| No test for the non-recording branch | covered in the characterization suite carried over from #3770 |

Worth noting the two reviews reached opposite conclusions on the cache: the blocker was filed on 18 Aug, and the independent verification that it is bounded (`cacheableMethod` + `routeCacheLimit`) came on 21 Aug, after the bounding landed. The current code is the bounded version.

### Re-measured on today's development, not on the tree this PR was written against

#3974 shipped in v1.60.0 and reworked the same request path, so the original numbers no longer describe the baseline. Same benchmarks, same machine, development's production code vs this branch, `-benchtime 20000x -count 6`:

| benchmark | development | this branch | |
|---|---:|---:|---|
| `BenchmarkTracer` (non-recording) | 1881 B · 19 allocs | **1568 B · 11 allocs** | −8 allocs |
| `BenchmarkTracer_Recording` | 2874 B · 23 allocs | **2657 B · 17 allocs** | −6 allocs |
| `BenchmarkTracer_WithPropagationHeaders` | 4067 B · 40 allocs | **3721 B · 30 allocs** | −10 allocs |

The wins survive the rebase, and the propagation-header case — the one the earlier benchmarks could not see — is the largest of the three at −10 allocations per request.

Baseline taken by reverting only `tracer.go` and `routecache.go` to development while keeping this branch's benchmarks, so both columns run identical measurement code. ns/op is omitted deliberately: it overlapped between the two on this machine and the allocation counts are the stable signal.

### Verification

`go build ./...` clean, `go test ./pkg/gofr/http/... -count=1` green, and `go test ./pkg/gofr/http/middleware/ -race` green.

### Still true, and not for this PR

The metrics-side sibling (`routeAttrs`) has the same unbounded-key shape and is fixed in #3972, which is stacked on this branch. Landing this alone leaves that hole open, so they should go back to back.

Umang01-hash (dismissed):
Ran it locally end to end — drove real requests through the Tracer with a live OTel export pipeline and the spans come out right: `/users/42` and `/users/99` both land on `GET /users/{id}` sharing the cached meta, no attribute bleed across routes, correct method/route/status. `-race` is clean and I confirmed the shared attr slices aren't mutated by OTel.

The win is real: cached path measures 0 allocs/0 B vs 5 allocs/256 B when rebuilt per request. Also checked the diff against development — the semconv keys and the `METHOD /route` span name already shipped earlier, so this really is behaviour-identical like the description says.

The baggage `Values()` fix is a nice catch and the test that sends multiple Baggage headers and compares against the stdlib carrier proves it. CI's green. LGTM.

One tiny thing, non-blocking: `meta, _ := v.(*spanMeta)` could rebuild-on-miss instead of ignoring the assert. And separately (not this PR) rbac/solr still emit the old `http.method`/`http.status_code` keys — might be worth a follow-up to make semconv consistent.

Umang01-hash (approved):

PR #3971, opened 2026-08-18
perf(cors): build the fixed response headers once, not per request

aryanmehrotra:
**TL;DR — 5 allocations per request → 0. No behaviour change.**

## The problem

CORS headers are written on **every response**, matched or not. Every time, the middleware:

1. rebuilt the `Access-Control-Allow-Methods` string from the route list
2. called `Header().Set(...)`, which **recanonicalises** the key and allocates a fresh `[]string` for the value — for each of three constant headers

None of it varies for the lifetime of the server.

## The fix

Build the header values **once at construction** and assign the map entries directly.

Sharing a value slice is safe because `len == cap == 1`, so a later `Header.Add` appends into a *new* array rather than mutating the shared one — pinned by a test.

## The numbers

Measured with one portable harness run against **both** revisions, so every row is a real before/after of the same request shape rather than a figure quoted from a benchmark that exists on only one side. `-benchtime 300000x -count 3`, Apple M4.

| request shape | `development` | this branch |
|---|---|---|
| wildcard config | 160 B/op, **5** allocs/op | **0 B/op, 0 allocs/op** |
| named-origin config | 176 B/op, **6** allocs/op | **32 B/op, 2 allocs/op** |
| wildcard + three passthrough headers | 304 B/op, **8** allocs/op | **0 B/op, 0 allocs/op** |

The shipped `BenchmarkCORS` / `BenchmarkCORS_NamedOrigin` report `0 B, 0 allocs` and `32 B, 2 allocs` on this branch, matching rows one and two.

> **Correction.** An earlier revision quoted a "before" row of `5 allocs / 208 B` against `BenchmarkCORS`. That benchmark **does not exist on `development`** — it ships in this PR — so no shipped benchmark reproduced the figure. The allocation count was right; the byte figure was not verifiable as stated. The table above replaces it with a pair measured on both sides.

> **Added after review.** `BenchmarkCORS` uses an **empty** config, so it only ever measured the wildcard fast path — the "→ 0 allocs" headline does not hold for a multi-origin config, whose `Header.Set` plus `Vary`'s `Header.Add` still allocate. `BenchmarkCORS_NamedOrigin` measures that branch, and the 0-alloc result now has a real guard: `TestCORS_WildcardPathIsAllocationFree` uses `testing.AllocsPerRun`, so a regression fails the suite instead of showing up in a table nobody diffs.

## Two claims this description used to make, now retracted

**#3770 has merged**, and it carried both of them into `development` already. This PR's job on that front is to **preserve** them through a structural rewrite, not to deliver them.

- **~~Bonus: fixes a data race~~.** The in-place `append(routes, "OPTIONS")` that wrote into the router's `RegisteredRoutes` backing array is **already gone** from `development` — its `joinAllowedMethods` builds the string with `strings.Join` and never appends. A 12,800-request concurrent soak under `-race`, with a route slice whose `cap` exceeds its `len`, is **clean on `development` and on this branch alike**. The race is real history, but it is not this PR's fix.
- **~~Security fix: canonical-key origin bypass~~.** `canonicalizeConfig` and its explicit precedence rule are **on `development` today**. The `before: Origin: https://evil.com → Access-Control-Allow-Origin: *` reproduction quoted earlier describes the pre-#3770 code, not the branch this PR merges into.

## What the #3770 interaction actually is now

The textual conflict is gone — #3770 merged on 2026-08-28 and this branch has merged `development` since, cleanly, with **neither `cors.go` nor `cors_test.go` touched by the merge**.

What remains is the reason the review asked the two to be coordinated, and it still holds as a design constraint on this rewrite: `buildFixedHeaders` classifies each config entry by key and writes everything it does not recognise into the fixed set, and **the fixed set is applied *after* the per-request origin negotiation**. Matching on the raw key would send a config spelled `access-control-allow-origin` to the `default` branch, which canonicalises it straight over the negotiated `Access-Control-Allow-Origin` — while `parseOrigins`, doing its own raw lookup, would already have missed the same key and fallen back to its **wildcard** default.

So this rewrite keeps the fold: `canonicalizeConfig` runs first and both `parseOrigins` and `buildFixedHeaders` read canonical keys only. `CORS` is exported, so callers do build this map themselves — in-tree the keys always arrive canonical, because `GetConfigs` renders them through `convertHeaderNames`.

`TestCORS_OriginAllowListMatchedByCanonicalKey`, `TestCORS_DuplicateSpellingsResolveDeterministically` and `TestCORS_CaseVariantMethodsAndHeadersAreClassified` pin it on this side.

## No behaviour change — how it was checked

Against `development` at `a513230c0`, after merging it in:

**1. Every header, every shape.** 18 configurations × 2 route sets × 3 origins × 2 methods = **216 request shapes**, dumping the complete response header map for each, driven through the exported `CORS()` only. Configurations include all four spellings of the origin key, both spellings present at once, two non-canonical spellings competing, empty custom values, a passthrough header, an invalid header key, and everything at once.

**1,116 lines, byte-identical.** The comparator is not vacuous: changing `", "` to `"; "` in the Allow-Headers join produces 48 diff lines.

**2. As a running service.** Two real GoFr binaries of the same app, one built against `development` and one against this branch, 4 CORS configurations × 3 origins × 3 methods:

```
### cfg=full origin=https://a.example.com method=GET
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: Authorization, ..., X-Correlation-ID, X-Custom
Access-Control-Allow-Methods: GET, PUT
Access-Control-Allow-Origin: https://a.example.com
Vary: Origin

### cfg=full origin=https://evil.example.com method=GET
   (no Access-Control-Allow-Origin, no Vary)
```

**192 lines, byte-identical.**

**3. Race soak.** 64 goroutines × 200 requests × 3 runs under `-race`, with a route slice whose `cap` exceeds its `len`: clean on both, and the caller's backing array is unwritten on both.

## Lifecycle contract (the one real behaviour change)

Building once instead of per request means the inputs are read at a point in time, so **when** matters. The rule is a single one: **the config map and the route slice are both read exactly once, on the first request.**

An earlier revision of this PR got that wrong in a way worth calling out — it froze the config at `CORS()` construction but the routes at the first request. That split was undocumented. Both are read in the same `sync.Once` now.

Measured against `development`, this is the *entire* difference:

```
                                    development           this branch
routes mutated AFTER 1st request    GET, POST, OPTIONS    GET, OPTIONS     <- differs
routes assigned BEFORE 1st request  GET, POST, OPTIONS    GET, POST, OPTIONS
config mutated AFTER 1st request    GET                   GET
```

The middle row is the one GoFr exercises: `CORS` is built in `App.New` when `RegisteredRoutes` is still empty, and `httpServerSetup` assigns the final list synchronously before the serve goroutine starts. Anything completed before the first request is still honoured.

What stays frozen is mutation *after* a request has been served. That is not a gap to close by reading per request: both inputs are read from every serving goroutine, so a caller mutating them concurrently races whenever the read happens. Freezing makes that race impossible rather than merely unlikely.

`TestCORSLifecycleContract` and `TestCORSLifecycleFrozenAfterFirstRequest` pin all of it, including that 200 concurrent first requests observe the same fully-built header set.

## Safety

- **No behaviour change**, verified three ways above.
- Every shared value slice goes through `sharedValue`, so the `len == cap == 1` invariant the whole sharing scheme rests on has one place to be stated — and `TestSharedHeaderValuesHaveNoSpareCapacity` asserts it for **all** of them, not just the ones a particular test happens to touch. The review flagged that invariant as load-bearing but only implicitly pinned.
- The duplicated `setMiddlewareHeaders` doc comment is deduplicated.

## Measurement

Allocations and bytes are the signal. Wall-clock on the machine used varied by several times between runs of the same binary.

---
*Touches only `cors.go` and its test. #3770 has merged and the textual conflict with it is gone.*


PiyushSingh-ZS:
Reviewed against `development`. I ran the middleware through 10 config permutations on both branches — wildcard, specific origin, disallowed origin, empty custom values for methods and headers, lowercase pass-through keys, empty route list — and the emitted headers are **byte-identical**. The caching is sound and the `sync.Once` timing is safe.

It also quietly fixes a latent aliasing bug: `development`'s `routes = append(routes, "OPTIONS")` writes into the router's `RegisteredRoutes` whenever `cap > len`. Good catch, and `joinAllowedMethods` is the right shape for it.

### One thing to carry over before this merges

Config keys are classified on the **raw** key:

```go
switch header {
case headerAccessControlAllowOrigin, headerAccessControlAllowMethods, headerAccessControlAllowHeaders:
    continue
default:
    fixed = append(fixed, headerValue{key: textproto.CanonicalMIMEHeaderKey(header), ...})
}
```

A differently-cased spelling — `access-control-allow-origin` — misses every case, falls into `default`, gets canonicalized on the way into the map, and then **overwrites the origin negotiated against the allow-list**, because `fixed` is applied after the negotiation. Header names are case-insensitive, so that's a reachable config, and the value bypasses the allow-list entirely.

This is exactly the defect #3770 fixes, by switching on `http.CanonicalHeaderKey(key)` instead. The two PRs conflict in `cors.go` (confirmed with `git merge-tree`), so whichever lands second decides the outcome — worth pulling the canonical-key classification into this PR so the fix isn't lost in the rebase.

### Minor

- The route list is now snapshotted at first request rather than re-read per request. Safe today: `RegisteredRoutes` is assigned once in `gofr.go:201`, before any server starts. But it does convert a per-request read into a startup-ordering dependency, which is worth the sentence the comment already gives it — maybe also worth noting that it holds only because route registration completes before `Run` starts serving.
- `setMiddlewareHeaders`'s doc comment has its first line duplicated ("setMiddlewareHeaders writes the CORS headers for one response." appears twice).

NitinKumar004 (changes requested):
The per-request → once optimization is sound and removing the `RegisteredRoutes` append data race is a nice catch. Requesting changes primarily for a cross-PR collision that has a security dimension.

### ⚠️ Cross-PR blocker — collides with #3770 (`cors.go`)
Both PRs rewrite `setMiddlewareHeaders`/`joinAllowedMethods` from the same base, incompatibly, and both add `joinAllowedMethods` (duplicate symbol = compile error), so the merge must pick one structure. This PR is the larger structural rewrite, so a resolver will likely keep it and **drop #3770's canonical-header-key fix**. That fix classifies config entries by `http.CanonicalHeaderKey`; this PR's `buildFixedHeaders` switches on the **raw** key. A config key like `access-control-allow-origin` (lowercase) then falls to `default`, is canonicalized, stored in `fixed`, and `header[h.key] = h.value` runs **after** origin negotiation — **overwriting the negotiated `Access-Control-Allow-Origin` and bypassing the allow-list** (same for case-variant Allow-Methods/Allow-Headers). Resolution must re-apply canonical-key classification on top of this PR's precompute-once structure.

### Low
- **Shared header value slices** are assigned into every response map; correctness rests on a `len==cap==1` invariant that's invisible at the call site — a future multi-element shared value or a downstream index-assign would cross-contaminate responses process-wide. It's tested/commented today, but the invariant is load-bearing.
- **Fixed headers built once on first request** couples correctness to the startup-then-serve lifecycle; fine today (routes are finalized before serving), but would silently go stale if post-serve route registration is ever added.
- `BenchmarkCORS` uses an **empty** config → only the wildcard fast path (0 allocs). The still-allocating **named-origin** branch (`Header.Set` + `Vary` `Add`) is unmeasured, so the "→0 allocs / no behaviour change" claim doesn't hold for multi-origin configs; there's also no `AllocsPerRun` guard on the 0-alloc result.
- Duplicated/garbled doc comment on `setMiddlewareHeaders`.

**Requested changes:** coordinate the `cors.go` merge with #3770 so canonical-key classification survives; add a named-origin benchmark + `AllocsPerRun` guard; dedup the doc comment.

Umang01-hash (dismissed):
Approving — clean, security-conscious perf PR, and it also closes a latent CORS wildcard-fallback bug. Verified the risky parts hold:

- **Origin negotiation preserved and safe.** A disallowed origin gets no `Access-Control-Allow-Origin`; wildcard emits `*`; an allow-listed origin gets its origin + `Vary: Origin`. `fixed` excludes every spelling of Allow-Origin, so the fixed-header pass (applied after negotiation) can't overwrite the negotiated value — pinned by `TestCORS_OriginAllowListMatchedByCanonicalKey` (across 4 key spellings, `evil.com` gets nothing, `trusted.com` survives).
- **The bundled `canonicalizeConfig` fix is a real security improvement** — a non-canonically-spelled `access-control-allow-origin` config previously slipped past `parseOrigins` and fell back to wildcard; now folded to canonical with deterministic precedence.
- **Shared-slice safety holds.** The `len == cap == 1` invariant means a later `Header.Add` appends into a new array; `-race` passes (incl. `-count=5` stress), and `TestSharedHeaderValuesHaveNoSpareCapacity` pins it for every shared value. No in-chain mutator of these header keys.
- **Perf is real** — `BenchmarkCORS` (wildcard) is 0 B/0 allocs; the named-origin path correctly still allocates since Allow-Origin genuinely varies. gofmt/vet clean, no breaking API (all new symbols unexported).

Two pre-existing, non-blocking notes (identical to the old behavior, not regressions in this PR): (1) no guard against a credentialed wildcard (ACAO `*` + a custom `Allow-Credentials: true`) — fails safe in browsers, but a future guard would be nice; (2) `Vary: Origin` is set only on the allow-list match path, not the deny path — per the Fetch spec a shared cache could otherwise serve a denied response to an allowed origin. Neither blocks. Nice work — the shared-slice `len==cap==1` reasoning and its test are a clean way to make the sharing safe.

aryanmehrotra:
## Re-review, and two claims of mine I'm withdrawing

@NitinKumar004 — all five of your requested changes are in. Two of them I answered by deleting claims from the description rather than by adding code, because **#3770 merged on 2026-08-28** and took both fixes into `development` ahead of this branch.

### Your requested changes

| Request | Where it landed |
|---|---|
| ⚠️ **Cross-PR** — coordinate the `cors.go` merge with #3770 so canonical-key classification survives | **Moot as a conflict** — #3770 merged, this branch has merged `development` since, cleanly, with neither CORS file touched by the merge. Kept as a design constraint: `canonicalizeConfig` runs first and both `parseOrigins` and `buildFixedHeaders` read canonical keys only |
| Named-origin benchmark + `AllocsPerRun` guard | `BenchmarkCORS_NamedOrigin` (176 B/6 allocs → 32 B/2 allocs) and `TestCORS_WildcardPathIsAllocationFree`. I mutation-checked the guard: swapping the shared-slice assignment back to `Header.Set` fails it |
| Shared header value slices — invariant load-bearing but only implicitly pinned | Every shared value now goes through `sharedValue()`, so `len == cap == 1` has one place to be stated, and `TestSharedHeaderValuesHaveNoSpareCapacity` asserts it for **all** of them |
| Fixed headers built once couples correctness to the lifecycle | Both inputs now freeze at the same instant, in one `sync.Once`. Measured scope below |
| Duplicated/garbled doc comment | deduplicated |

### The two claims I'm withdrawing

You were right to want these coordinated, and coordinating them is what showed the description was overclaiming.

- **"Bonus: fixes a data race."** Withdrawn. The in-place `append(routes, "OPTIONS")` is **already gone from `development`** — its `joinAllowedMethods` builds with `strings.Join` and never appends. A 12,800-request concurrent soak under `-race`, with a route slice whose `cap` exceeds its `len`, is **clean on `development` and on this branch alike**, and the caller's backing array is unwritten on both. Real history, not this PR's fix.
- **"Security fix: canonical-key origin bypass."** Withdrawn as a *contribution*. `canonicalizeConfig` and its precedence rule are on `development` today. The `Origin: https://evil.com → Access-Control-Allow-Origin: *` reproduction I quoted describes pre-#3770 code, not the branch this merges into. What this PR does is **preserve** that behaviour through the rewrite — which was your point about the merge, and it is why the fold has its own tests here.

The description now says exactly this instead of claiming either fix.

### No behaviour change — checked three ways against `a513230c0`

1. **216 request shapes, every response header.** 18 configurations × 2 route sets × 3 origins × 2 methods, through the exported `CORS()` only — all four spellings of the origin key, both spellings at once, two non-canonical spellings competing, empty custom values, an invalid header key, everything at once. **1,116 lines, byte-identical.** The comparator is not vacuous: changing `", "` to `"; "` in the Allow-Headers join produces 48 diff lines.
2. **As a running service.** Two real GoFr binaries, one per revision, 4 CORS configs × 3 origins × 3 methods. **192 lines, byte-identical.** A disallowed origin gets no `Access-Control-Allow-Origin` and no `Vary` on both.
3. **Race soak.** 64 goroutines × 200 requests × 3 runs under `-race`: clean on both.

### The one real behaviour change

```
                                    development           this branch
routes mutated AFTER 1st request    GET, POST, OPTIONS    GET, OPTIONS     <- differs
routes assigned BEFORE 1st request  GET, POST, OPTIONS    GET, POST, OPTIONS
config mutated AFTER 1st request    GET                   GET
```

The middle row is the one GoFr exercises — `CORS` is built in `App.New` when `RegisteredRoutes` is still empty, and `httpServerSetup` assigns the final list synchronously before the serve goroutine starts. Only mutation *after* a request has been served is frozen out, and doing that is already a data race for a direct caller, since both inputs are read from every serving goroutine.

### The numbers, both sides of the same harness

`-benchtime 300000x -count 3`, Apple M4:

| request shape | `development` | this branch |
|---|---|---|
| wildcard config | 160 B/op, 5 allocs/op | **0 B/op, 0 allocs/op** |
| named-origin config | 176 B/op, 6 allocs/op | **32 B/op, 2 allocs/op** |
| wildcard + three passthrough headers | 304 B/op, 8 allocs/op | **0 B/op, 0 allocs/op** |

I also replaced the old "before | 5 allocs | 208 B" row: it cited `BenchmarkCORS`, which **does not exist on `development`** — it ships in this PR — so no shipped benchmark reproduced it. The allocation count was right; the byte figure was not.

### Gates

`gofmt`, `go vet`, `go build ./...`, `-race` on `./pkg/gofr/http/...`, and `golangci-lint --new-from-rev=origin/development` → **0 issues** (non-vacuous — a full run shows 5 pre-existing in the package). 0 behind `development`. CI is running at `a858fdc7a` as I post this.

### One thing that is not mine to fix

`Access-Control-Allow-Credentials: true` is emitted even to a **disallowed** origin, where no `Access-Control-Allow-Origin` accompanies it. Harmless — a browser ignores it without the origin header — and **identical on `development`**, so it is pre-existing rather than introduced here. Flagging it rather than folding it in.

Umang01-hash (approved):
LGTM. Read the whole thing at head and ran it.

- Freezing config+routes on the first request (not at construction) is the right call — RegisteredRoutes is empty until httpServerSetup, so a construction-time read would only ever emit `Allow-Methods: OPTIONS`. The lifecycle contract tests pin both halves plus the concurrent freeze-once.
- Shared len==cap==1 value slices are safe (a later Header.Add reallocates); confirmed the no-spare-capacity invariant test actually fails if you hand `sharedValue` any spare cap.
- Perf is real, not just claimed: BenchmarkCORS on the wildcard path is 0 B/op, 0 allocs/op (down from 3+ Set allocs + a Join per request). Named-origin stays at 2 allocs, which is irreducible since the origin varies.

gofmt/vet/build clean, golangci --new-from-rev = 0 issues.

PR #3972, opened 2026-08-18
perf(metrics): build the measurement option once per route, method and status

aryanmehrotra:
**TL;DR — 4 allocations per observation → 0. No behaviour change.**

## The problem

Every request records a histogram. Every observation called:

```go
metric.WithAttributes(attrs...)
```

which **sorts and deduplicates** the attributes into a new `attribute.Set` and wraps it in an option.

For a request metric the labels are `(route, method, status)` — drawn from a **small fixed set**. So that work produced an identical option over and over.

## The fix

Build the option **once per `(route, method, status)`** and reuse it, via a new option-based recorder on the metrics manager.

The option is passed as a **slice**, not a single value — passing one option would allocate a fresh one-element slice for the variadic on every call, undoing the saving.

## The numbers

Both benchmarks ship here, so before/after is visible side by side without checking out the old revision.

| | allocs/op | B/op | ns/op |
|---|---|---|---|
| `BenchmarkRecordHistogramAttrs` (rebuilds each time) | 4 | 424 | ~285 |
| **`BenchmarkRecordHistogramOpt`** (cached option) | **0** | **0** | **~110** |
| `BenchmarkAttrBuild_HTTP` (the attribute build a hit also avoids) | 3 | 448 | ~83 |

`-benchtime 20000x -count 4`, Apple M4. ns/op is machine-dependent and quoted only because the three ran back to back on the same binary; **the allocation counts are the reliable signal.**

> **Two corrections to earlier revisions of this description.** It once claimed **−650 B/request**, which no shipped benchmark reproduces — retracted. It then quoted **~928 ns/op** and **~164 ns/op** for the two rows above; those do not reproduce either, and the table now carries measured figures.

### The saving holds at the seam, not just in the benchmark

The shipped benchmarks measure `metricsManager.RecordHistogramOpt` with a pre-built slice. They do not measure `optionRecorder.record`, which is the thing every request actually calls — and which boxes its cache key into an `any`. Measured directly:

```
optionRecorder.record, cache hit     0 B/op   0 allocs/op    23 ns/op
optionRecorder.record, uncacheable 424 B/op   4 allocs/op   211 ns/op
```

The boxing does not escape, so the win survives the call site. The uncacheable figure is **exactly the baseline**, so traffic that cannot be cached is not regressed.

## Safety

- **No behaviour change.** Same metric, same labels, same values.
- The caches are keyed on the **route template**, never the concrete path.
- **Only templated routes are cached.** Every branch of the path label that falls back to `r.URL.Path` yields a caller-controlled string — an unmatched route, a static-asset extension, `/static`, and the `"/"` template GoFr's `PathPrefix("/")` catch-all produces for *every* unmatched request. Those used to be cached, so a stream of unique unmatched paths filled the cache to its ceiling; after that **every first-seen legitimate route could never be stored** and rebuilt its option per request forever. Memory stayed bounded and correctness was unaffected — the optimization just silently reverted to baseline for production traffic.
- **The method half is filtered too** (`cacheableMethod`): a bounded route with an unbounded method reopens the same growth.
- **`attrsRecorder.cache` had no ceiling at all** while its sibling had one. Both now go through `routeCache`.

> **On "bounded together".** The three caches — the tracer's span cache from #3970 and the two metrics recorders' — share the `routeCache` **type and its `routeCacheLimit`**, and the same `templated && cacheableMethod` admission rule. They are three instances with three budgets, not one shared budget: `tracerSpanCache` is package level, the metrics caches are fields on their recorder. Consistently bounded under one mechanism; not a single pool.

**It does not bound metric cardinality, and this description should not be read as claiming it does.** The ~2,000-series ceiling and the `otel_metric_overflow` bucket are the OTel SDK's own `defaultCardinalityLimit`, tunable via `WithCardinalityLimit` / `OTEL_GO_X_CARDINALITY_LIMIT`, behaving as designed on `development` and here alike.

What the bound does buy, measured as a running service: two binaries of the same three-route app, one built against `development` and one against this branch, each flooded with requests to one templated route carrying a **distinct RFC-7230 method token** per request.

| requests, each with a distinct method | development | this branch |
|---:|---:|---:|
| 5,000 | +22.3 MB RSS | **+20.8 MB** |
| 20,000 | +31.8 MB RSS | **+23.7 MB** |

20,000 is the mean of three runs each: `31.4 / 31.2 / 32.8` against `23.4 / 23.9 / 23.7`. About **8 MB less resident**, and the shape is the point — `development` keeps climbing from 5k to 20k while this branch is nearly flat, which is the ceiling doing its job.

> **A third correction.** An earlier revision quoted `+31.4 / +35.5` against `+25.4 / +26.4` for these rows. Those came from a different probe app and do not reproduce; the table above does, and its runs are listed so the spread is visible.

### Behaviour verified by running it, not only by testing it

Both binaries, same requests, `app_http_response_count` scraped from `/metrics` and diffed:

```
{method="GET",     path="/users/{id}",         status="200"} 3
{method="GET",     path="/boom",               status="500"} 1
{method="GET",     path="/totally/unknown/aaa",status="404"} 1
{method="M000001", path="/users/7",            status="404"} 1
```

**Byte-identical on both.** Two distinct concrete paths (`/users/7`, `/users/99`) collapse to the one `/users/{id}` template; an unmatched path and an undefined method both fall back to the raw path, which is exactly why neither may be cached; and `status` is a string on every row.

## Test coverage added after review

- **The fast path all production traffic uses now has label assertions.** `metricsManager` implements `RecordHistogramOpt`, so requests flow through `optionRecorder`, yet nothing pinned what it emitted — changing `status` from `attribute.String` to `attribute.Int`, or emitting the concrete path instead of the route template, left every test green while dashboards broke and cardinality exploded. Both now fail the suite; verified by making each change.
- **`TestMetricsOptCacheIsBounded` drives the guard.** It used to assert only that a compile-time constant was positive, so deleting the ceiling kept it green. Deleting it now fails three tests.
- **`attrsRecorder` was at 0% coverage** — no in-tree type selects it, since `*metricsManager` implements both optional interfaces and `optionRecorder` wins — and now has direct tests for its labels, its reuse and its bound.
- **`TestMetricsPath` pins the path label and the cacheable flag directly**, rather than only through the cache-growth tests that consume them. Its two static-extension cases match a route on purpose, so the template would win were the extension not checked.

## Review nits addressed

- **`routeCache` now carries a value type parameter.** Every consumer was a single writer of a single type, so the type assertions at each call site could never fail — which left each one with a branch no test could reach and nothing but a comment explaining why. Naming the type deletes them: the one assertion left is inside `load`, and `store` makes it unfalsifiable. That removes the fall-through in `optionRecorder.record`, the `b == nil` rebuild guard in `attrsRecorder.record`, and the nested assertion in `spanMetaFor`. Coverage on the new code is now a genuine 100%, with no unreachable branches padding it.
- **`attrsRecorder` is kept.** It is unreachable in-tree, but `metricsAttrer` is a structural surface that predates this PR on `development`: any external `metrics.Manager` exposing `RecordHistogramAttrs` selects it. Deleting it would remove shipped capability, which is not this PR's business. It is now tested and bounded, where before it was neither.

---
> **#3970 has merged**, so this targets `development` directly and carries the shared `routeCache`/`cacheableMethod` bound from it. The two were reviewed as one unit at the reviewer's request; this is the half that closes the metrics-side memory-growth hole.


PiyushSingh-ZS:
Reviewed against `development`, built and tested on the branch. The three-strategy split reads well and the layering decision — select once at construction, keep the branching out of the request path — is the right one.

Checks I ran:

- Confirmed the runtime type is `*metricsManager` (`container.go:139` / `container.go:259`), so the `optionRecorder` path is the one that actually executes in a real app.
- `RecordHistogramOpt` was added to the concrete type only, not to the exported `metrics.Manager` interface, so existing implementers and generated mocks are unaffected. That's the right call and worth keeping that way.
- Emitted label values are unchanged: `strconv.Itoa(status)` matches the old `fmt.Sprintf("%d", status)`, and status stays a string across all three paths, so OTLP consumers see the same types.
- Sharing a built `metric.RecordOption` across concurrent observations is safe — `WithAttributes` holds an immutable `attribute.Set` and `Record` only reads it.

### `attrsRecorder`'s cache has no equivalent of `optCacheLimit`

The rationale given for the cap on the option cache is exactly right:

> The path label falls back to the raw request path when no route template is available, and a raw path is caller-controlled.

But that argument applies identically to `attrsRecorder.cache`, which has no ceiling. An external metrics implementation that satisfies `metricsAttrer` but not `metricsOptRecorder` gets an unbounded `sync.Map` fed by attacker-controlled keys. Low practical impact since in-tree traffic takes the option path, and it's inherited rather than introduced here — but this PR is where the bound is being reasoned about, so it seems like the moment to apply it in both places.

### Related, and probably a separate issue

The same fallback (`path = r.URL.Path` when no route template resolves, metrics.go:73) is an unbounded **metric cardinality** source, not just a cache-size one — every unmatched path becomes its own time series in the exporter. The cache limit bounds the memory this middleware holds, but not what gets shipped to the backend. Worth a follow-up.

### Nit

`o.count.Load() < optCacheLimit` followed by `LoadOrStore` can overshoot the limit slightly under concurrency. Harmless at this bound, just noting it's approximate rather than exact.

NitinKumar004 (changes requested):
Production code is correct; requesting changes for test coverage of the new fast path, cache behavior under catch-all traffic, and cross-PR cache coordination.

### Medium
- **The `optionRecorder` fast path — the one all real traffic now uses — has no label assertions.** `metricsManager` implements `RecordHistogramOpt`, so production flows through `optionRecorder`, yet the only test touching it (`TestMetricsOptRecorderIsUsedAndStillRecords`) drops the opts and just counts hits. Nothing asserts the path is the route template `/users/{id}` (not `/users/7`), that method/status keys are present, or that status is `attribute.String` (matching the slow path so OTLP types agree). Change `String`→`Int` or emit the concrete path and every test stays green while dashboards break and cardinality explodes.

### Low
- `TestMetricsOptCacheIsBounded` only asserts `require.Positive(optCacheLimit)` — a compile-time constant. It never drives >limit distinct keys through `record()` nor inspects `count`/cache size, so deleting the `if o.count.Load() < optCacheLimit` guard (re-introducing unbounded growth) still passes. The advertised bound has no executable coverage.
- **Shared 4096-slot cache lets catch-all raw-path traffic starve real routes.** The `PathPrefix("/")` catch-all always takes the `raw r.URL.Path` fallback; 4096+ unique unmatched paths saturate the cache, after which every first-seen legitimate combo can never be stored and `record()` rebuilds `metric.WithAttributes` per request **forever** — the optimization silently reverts to baseline for production traffic (memory stays bounded; correctness unaffected). Cache only when `RouteTemplate(r) != ""`, or give the raw-path fallback a separate budget.
- **`attrsRecorder` is dead + untested in-tree, and its cache is unbounded.** No in-tree type selects it (`*metricsManager` implements both), so it runs at 0% coverage, and its `sync.Map` has no `optCacheLimit` ceiling — an external `RecordHistogramAttrs`-only backend grows it without bound under catch-all traffic.
- The headline "**-650 B/request**" is not reproducible from the shipped benchmarks (they yield ~424 B/op).

### ⚠️ Cross-PR (HIGH, coordinate with #3970)
These metrics caches + #3970's `tracerSpanCache` are three caches on the same attacker-influenced key space with **inconsistent bounding** (`attrsRecorder.cache` unbounded). One malicious request stream inflates several. Bound all three consistently and review as one unit.

**Requested changes:** assert label/type/template on the opt path; make the cache-bound test actually drive the guard; cache only templated routes (or a separate raw-path budget); bound `attrsRecorder.cache`; align the perf-table numbers with the shipped benchmarks; coordinate with #3970.

aryanmehrotra:
All six items are addressed. Re-requesting review — the `CHANGES_REQUESTED` verdict sits on `7a41fd973` and the head is now `5d30c466b`.

### The fast path now has label assertions

You were right that this was the gap that mattered: `metricsManager` implements `RecordHistogramOpt`, so production flows through `optionRecorder`, and nothing pinned what it emitted. `TestMetricsOptRecorderLabels` asserts the route template, the method, and that status is `attribute.String`.

I also added a test that compares **all three strategies against each other** on the same requests, because "the opt path is right" is weaker than "the three cannot disagree":

```
GET /users/42     slow="method=STRING:GET path=STRING:/users/{id} status=STRING:201"
                  attr="method=STRING:GET path=STRING:/users/{id} status=STRING:201"
                  opt ="method=STRING:GET path=STRING:/users/{id} status=STRING:201"
```

Identical across six request shapes — same keys, same values, same attribute **types**. Both mutations you named now fail the suite: `String`→`Int` on status, and emitting the concrete path instead of the template.

### Catch-all traffic can no longer starve real routes

This was the subtlest of your findings and it needed a design change, not just a ceiling. Every branch of the path label that falls back to `r.URL.Path` — unmatched route, static extension, `/static`, and the `"/"` template GoFr's catch-all produces — yields a caller-controlled string. Those are no longer cached at all:

```
after 6000 catch-all requests:        entries=0
real route can still be cached:       entries=1
6000 undefined methods on one route:  entries=0
attrsRecorder at ceiling:             entries=4096 (limit 4096)
```

`attrsRecorder` had **no ceiling at all** while its sibling had one — exactly the inconsistency that makes a shared key space dangerous. Both go through `routeCache` now, and it has direct tests for its labels, its reuse and its bound; it was previously at 0% coverage because no in-tree type selects it.

### `TestMetricsOptCacheIsBounded` drives the guard

It asserted only that a compile-time constant was positive, so deleting the ceiling kept it green. It now pushes past the limit in distinct keys and inspects the cache; deleting the ceiling fails it along with two others.

### The `−650 B/request` claim is retracted

No shipped benchmark reproduces it. The PR body now quotes the reproducible pair — `BenchmarkRecordHistogramAttrs` at 424 B/op, 4 allocs/op against `BenchmarkRecordHistogramOpt` at 0 B/op, 0 allocs/op, with `BenchmarkAttrBuild_HTTP` at 448 B/op for the attribute build a cache hit also avoids.

Measured through the middleware against this PR's base:

```
base  1456 B/op  11 allocs/op
3972  1232 B/op  10 allocs/op
```

Worth recording that the saving depends on which optional interface the backend implements: a backend offering only `RecordHistogramAttrs` sees −32 B, because that path was already cheap. The −224 B is the `RecordHistogramOpt` path, which is what `metricsManager` selects.

### Coordinated with #3970, as you asked

`routeCache` and `cacheableMethod` live in their own file for exactly this reason, and **this PR is stacked on #3970** so the three caches are bounded by one mechanism under one set of rules rather than three ad-hoc ceilings.

### No behaviour change

14 label cases diffed against the base — matched templates, unmatched paths, undefined methods, root, `/static` prefix, static extensions including uppercase and with a query string, trailing-slash trimming, the `/graphql` skip, and `/graphql` appearing in the query rather than the path. **All identical.**

`-race` clean on the http and metrics trees, `golangci-lint --new-from-rev` reports 0 issues, 0 behind `development`.

aryanmehrotra:
## Up to date, and a self-review

Merged `development` after #3970 landed. That merge conflicted in three files, and the resolution is worth stating because a careless one would have been silent: this branch carried an **older snapshot** of `tracer.go` / `tracer_test.go` from when it was stacked, so keeping "ours" would have reverted three fixes that went into #3970 after the stack was cut. Development's versions were taken for both, and `routecache.go` was rebuilt from development's copy with this PR's own change — generalising the cache key from `routeKey` to `any` — re-applied on top. Verified afterwards that the dead `spanMeta.attrs` stayed removed, and that the nil-assertion fix and the sampling comment survived.

0 behind `development`, CI green.

### Requested changes, and where they landed

| Request | Where |
|---|---|
| **Medium** — the `optionRecorder` fast path has no label assertions | `TestMetricsOptRecorderLabels` asserts the exact map: `path=STRING:/users/{id}` (template, not `/users/7`), `method=STRING:GET`, `status=STRING:201` — including that status is a **String**, so the two recorders agree for OTLP |
| Bound test asserted only a compile-time constant | `TestMetricsOptCacheIsBounded` drives `routeCacheLimit + 1000` distinct keys through `record()` and asserts `cache.len() == routeCacheLimit` |
| Catch-all raw-path traffic can starve real routes out of the cache | `cacheable := templated && cacheableMethod(method)` on both recorders — an untemplated path never enters the cache, so it cannot consume the budget. `TestMetricsCachesRejectUntemplatedPaths` |
| `attrsRecorder` untested, and its cache unbounded | now on the same `routeCache` under the same rules; `TestAttrsRecorderIsCorrectAndBounded` |
| "−650 B/request" not reproducible | description revised; the numbers below are the shipped ones |
| ⚠️ **Cross-PR** — bound all three caches consistently | done by construction: `routeCache` is one type with one ceiling, shared by the tracer cache and both metrics recorders. That is what the `any` key exists for |

### The numbers, re-run just now

`-benchtime 20000x -count 4`:

| benchmark | B/op | allocs/op |
|---|---:|---:|
| `BenchmarkRecordHistogramAttrs` (rebuilds each time) | 424 | 4 |
| `BenchmarkRecordHistogramOpt` (cached option) | **0** | **0** |
| `BenchmarkAttrBuild_HTTP` (the attribute build a hit also avoids) | 448 | 3 |

Matches the description exactly. One correction: the description's `~928 ns/op` for the attrs path measures ~280 ns/op here. ns/op is machine noise and the description already says the allocation counts are the reliable signal, but the figure should not be read as reproducible.

### What I found reviewing my own diff

**Fixed in `4af7ab8e`.** The cached-option branch discarded the assertion's `ok`:

```go
opts, _ := v.([]metric.RecordOption)
o.rec.RecordHistogramOpt(ctx, histogramName, seconds, opts...)
```

A failed assertion leaves `opts` nil, and recording with no options emits the measurement with **no attributes** — an unlabelled time series that looks like a working metric while corrupting any query over it. Worse than the nil dereference the same shape caused in the tracer cache, because nothing fails loudly. It now falls through to the rebuild, matching `spanMetaFor`. `attrsRecorder` was already correct: its assertion is followed by an `if b == nil` rebuild.

**Observation, not a change.** `newStatusAttrCache` is an unbounded `sync.Map` keyed on the status code. Unlike path and method, the status is chosen by the application rather than the caller, so it is bounded by the app's own code paths — but it is the one cache here without a ceiling, and worth knowing if a handler ever writes a caller-supplied status.

**Out of scope, pre-existing.** Both recorders pass `context.Background()` rather than the request context, so a recorded measurement carries no exemplar linking it to the trace. That is unchanged from `development` (it does the same in both places) and is not something this PR should alter, but it is the reason metrics here cannot be correlated to a span.

### Why this matters beyond the allocation win

Measured against `development` today, as a running service:

```
metric series before:                     21
after 5,000 requests with distinct methods: 42,000
app_http_response_count{method="M000000",path="/totally/unknown",status="404"} 1
```

An unauthenticated client inflates metric cardinality by varying the HTTP method alone. #3970 closed the tracer half; this is the metrics half, and it is live on `development` until this lands.

aryanmehrotra:
## Ran it as a service — and I need to correct something I wrote above

I said this PR "closes the metrics-side cardinality hole". **That is wrong**, and running it is what showed it. Two binaries, `development` and this branch, same app, 5,000 and 20,000 requests each carrying a distinct RFC-7230 method token.

### Metric cardinality is unchanged, and was never GoFr's to bound

| | development | this branch |
|---|---:|---:|
| `app_http_*` series after 5,000 distinct methods | 21 → **42,000** | 21 → **42,000** |
| distinct `method` label values exported | 1,998 | 1,998 |
| `otel_metric_overflow` markers | 21 | 21 |

Identical. The cap at ~2,000 is the OTel SDK's own default cardinality limit, spilling the rest into the overflow bucket — it is doing that on `development` today and it does the same here. The caches this PR bounds are GoFr's internal maps; they never controlled how many series the SDK creates, because the attributes still carry the caller's method either way.

So the attacker-influenced cardinality question is **still open**, on both, and it is a separate change: it needs the label itself constrained (an allow-list of methods, or bucketing unknown ones), not a cache ceiling. Worth its own issue rather than being folded in here.

### What the bound does buy, measured

| requests with distinct methods | development | this branch |
|---:|---:|---:|
| 5,000 | +31.4 MB RSS | **+25.4 MB** |
| 20,000 | +35.5 MB RSS | **+26.4 MB** |

About 9 MB less resident, and flatter as the flood grows — that is the cache ceiling doing its job. The remainder is the SDK's own series, which neither branch changes.

### Behaviour is identical on real traffic

Same app, same requests, exported series compared byte for byte:

```
app_http_response_count{method="GET",  path="/users/{id}", status="200"} 2
app_http_response_count{method="POST", path="/users",      status="201"} 1
app_http_response_count{method="GET",  path="/.well-known/alive", status="200"} 1
```

Identical on both. Two different concrete paths (`/users/7`, `/users/99`) collapse to the one `/users/{id}` template, the method is the raw request method, and the status is a string — the exact contract `TestMetricsOptRecorderLabels` pins.

### Summary of what this PR actually delivers

- the allocation win on the record path: **424 B → 0 B, 4 allocs → 0**
- a ceiling on both metrics caches, shared with the tracer's: **~9 MB less resident** under a hostile method stream, where before `attrsRecorder` had no ceiling at all
- no change to metric names, labels, types or values

It does **not** bound metric cardinality, and the description should not be read as claiming it does.

aryanmehrotra:
## Retracting the cardinality framing entirely

I called the metric cardinality an open hole needing its own fix. It is not — it is a configured SDK limit behaving as designed, and I should have checked before writing that.

`sdk/metric@v1.45.0/config.go` sets `defaultCardinalityLimit = 2000` and exposes it through `WithCardinalityLimit` and `OTEL_GO_X_CARDINALITY_LIMIT`. It tracks the setting exactly — 4,000 requests with distinct method tokens:

| `OTEL_GO_X_CARDINALITY_LIMIT` | distinct `method` labels exported |
|---|---:|
| unset (default 2000) | 1,998 |
| 100 | 98 |
| 500 | 498 |

So the ~2,000 ceiling and the `otel_metric_overflow` bucket are the SDK doing its job, tunable per deployment. There is nothing for GoFr to fix and no issue worth filing. Sorry for the noise in the two comments above.

### What this PR does, stated plainly

- **424 B → 0 B, 4 allocs → 0** on the record path
- a ceiling on both metrics caches, shared with the tracer's, where `attrsRecorder` previously had none: **~9 MB less resident** under a hostile method stream, and flatter as it grows
- **no change** to metric names, labels, types or values — verified by running two binaries side by side and comparing the exported series

That is the whole claim, and it holds.

pkg/gofr/http/middleware/metrics.go
```diff
@@ -126,41 +77,210 @@ func Metrics(metrics metrics) func(inner http.Handler) http.Handler {
...
+	)}
+
+	if cacheable {
+		o.cache.store(key, opts)
+	}
+
+	o.rec.RecordHistogramOpt(context.Background(), histogramName, seconds, opts...)
+}
+
+// attrsRecorder caches the (path, method) attribute pair and copies it into a
```
Umang01-hash, line 232: nit: attrsRecorder is unreachable in prod — *metricsManager implements RecordHistogramOpt and newHistogramRecorder checks metricsOptRecorder first, so optionRecorder always wins. The tier only ever runs from its own test. Fine to keep as a migration path for external Attrs-only Managers, but if that's not a supported surface this could collapse to option+labels.

pkg/gofr/http/middleware/metrics.go
```diff
@@ -126,41 +77,210 @@ func Metrics(metrics metrics) func(inner http.Handler) http.Handler {
...
+	cacheable := templated && cacheableMethod(method)
+	key := statusRouteKey{path: path, method: method, status: status}
+
+	if cacheable {
+		// The assertion's ok is honored rather than discarded: a failed one would
+		// leave opts nil and record the measurement with NO attributes, which is a
+		// silently unlabeled time series rather than a visible failure. Falling
+		// through rebuilds instead.
+		if v, ok := o.cache.load(key); ok {
+			if opts, ok := v.([]metric.RecordOption); ok {
```
Umang01-hash: nit: this inner type-assertion can't fail — the cache is single-consumer and only ever holds []metric.RecordOption, which is also why the fall-through has no coverage. opts := v.([]metric.RecordOption) would read as honestly single-typed.

pkg/gofr/http/middleware/metrics.go
```diff
@@ -126,41 +77,210 @@ func Metrics(metrics metrics) func(inner http.Handler) http.Handler {
...
+// The label itself is unchanged. What is new is the second return value, which
+// decides whether the result may be used as a cache key. Every branch that falls
+// back to r.URL.Path yields a caller-controlled string: an unmatched route, a
+// static-asset extension, /static, and - crucially - the "/" template that
+// GoFr's PathPrefix("/") catch-all produces for every unmatched request. Without
+// this distinction a stream of unique unmatched paths filled the shared cache to
+// its ceiling, after which every first-seen LEGITIMATE route could never be
+// stored and rebuilt its measurement option per request forever. Memory stayed
+// bounded; the optimization silently reverted to baseline for real traffic.
+func metricsPath(r *http.Request) (path string, templated bool) {
```
Umang01-hash, line 102: nit: the new templated return is only pinned indirectly via the cache-growth tests; a direct metricsPath() unit test over static-ext / "/" / /static / unmatched would assert it cleanly. Non-blocking.

Umang01-hash (dismissed):
LGTM. Ran it locally at this head — booted examples/http-server and scraped /metrics: app_http_response emits correctly through the real metricsManager→optionRecorder path (path/method/status all String). Benchmarks confirm the win: Opt = 0 allocs/110ns vs Attrs = 4 allocs/424B/284ns. gofmt/vet/build/-race tests/golangci-lint --new-from-rev all clean; 100% cover on every new func. Nice catch bounding the previously-unbounded attrs cache under routeCache + cacheableMethod + the templated gate — that closes a real memory-growth vector, not just a perf tidy.

All comments below are nits, nothing blocking.

pkg/gofr/http/middleware/metrics.go
aryanmehrotra: Fixed, and you were right about more of the codebase than the line you flagged.

Rather than make the assertion honest, I removed the need for one: `routeCache` now carries a value type parameter.

```go
type routeCache[V any] struct{ ... }
func (c *routeCache[V]) load(key any) (value V, ok bool)
func (c *routeCache[V]) store(key any, value V)
```

Every consumer is a single writer of a single type, so `store` makes the one remaining assertion — inside `load` — unfalsifiable. That deletes the nested assertion here, the `b == nil` rebuild guard in `attrsRecorder.record`, **and the same shape in `spanMetaFor`**, which came in with #3970 and which you didn't flag.

The key stays `any`: the three key shapes genuinely differ, and it is the ceiling rather than the key type that belongs in one place.

Your point about the coverage was the useful half. Function-level coverage read 100% on all of this before, because the false arm of a comma-ok has no statement in it — so "100% covered" and "has a branch no test can reach" were both true at once. Now they aren't: same 100%, no unreachable arms padding it.

Pushed in `f74e2a30`.

pkg/gofr/http/middleware/metrics.go
aryanmehrotra: Confirmed dead in-tree, and I'd like to keep it — with your framing, it is the supported-surface question and I think the answer is yes.

`metricsAttrer` is unexported but structural, so any external `metrics.Manager` exposing `RecordHistogramAttrs(ctx, name, float64, ...attribute.KeyValue)` selects this tier. That interface and its fast path are already on `development` — they predate this PR, which only refactored the closure into `attrsRecorder` and added `optionRecorder` above it. Collapsing to option+labels would delete shipped capability from an external implementer as a side effect of a perf change, which I don't think this PR should do.

What was genuinely wrong is that it was **unreachable *and* unbounded *and* untested** — 0% coverage with a `sync.Map` that had no ceiling while its sibling had one. All three are fixed here: it is on `routeCache` under the same `templated && cacheableMethod` rule, and `TestAttrsRecorderIsCorrectAndBounded` covers its labels, its reuse and its bound.

Worth recording one number I measured while checking this: on a cache hit that tier is **192 B/op, 1 alloc/op**, because `attrs[:]` escapes into the variadic. Pre-existing, unchanged by this PR, and it is why `optionRecorder` sits above it — but if an Attrs-only backend ever shows up in-tree, that is the alloc to go after.

If you'd still rather it go, say so and I'll drop it in a follow-up rather than widen this one.

aryanmehrotra:
## Re-verified at `f74e2a306`, by running it

@Umang01-hash — sorry, this push dismissed your approval as stale. It is nit-level, but it touches `tracer.go` as well as the metrics side, so here is exactly what moved and what I re-ran.

### What changed

`routeCache` now carries a **value** type parameter, which is the honest version of your `metrics.go` assertion nit:

```go
type routeCache[V any] struct{ ... }
func (c *routeCache[V]) load(key any) (value V, ok bool)
func (c *routeCache[V]) store(key any, value V)
```

Every consumer is a single writer of a single type, so `store` makes the one remaining assertion — inside `load` — unfalsifiable. That deletes the nested assertion in `optionRecorder.record`, the `b == nil` rebuild guard in `attrsRecorder.record`, **and the same shape in `spanMetaFor`**, which came in with #3970. The key stays `any`: the three key shapes genuinely differ, and it is the ceiling rather than the key type that belongs in one place.

Your coverage remark was the sharp half. Function coverage read 100% on that code before, because the false arm of a comma-ok contains no statement — so "100% covered" and "has a branch no test can reach" were both true at once. Now they are not: same 100%, no unreachable arms padding it.

Also adds `TestMetricsPath`, your third nit. Its two static-extension cases match a route on purpose, so the template would win were the extension not checked — I deleted that branch to confirm they fail.

### Ran it as a service, both sides

Two binaries of the same three-route app, one built against `development` (`d4185d11`) and one against this head, same requests, `app_http_response_count` scraped and diffed:

```
{method="GET",     path="/users/{id}",          status="200"} 3
{method="GET",     path="/boom",                status="500"} 1
{method="GET",     path="/totally/unknown/aaa", status="404"} 1
{method="M000001", path="/users/7",             status="404"} 1
```

**Byte-identical.** `/users/7` and `/users/99` collapse to the one template; an unmatched path and an undefined method both fall back to the raw path, which is precisely why neither may be cached; status is a string on every row.

The ceiling still earns its place — same app, flooded with one distinct method token per request:

| requests | development | this branch |
|---:|---:|---:|
| 5,000 | +22.3 MB RSS | **+20.8 MB** |
| 20,000 | +31.8 MB RSS | **+23.7 MB** |

20,000 is the mean of three runs each: `31.4 / 31.2 / 32.8` against `23.4 / 23.9 / 23.7`.

### The allocation win survives the generics

Measured on `optionRecorder.record` itself, which is what every request calls and which no shipped benchmark covers:

```
record(), cache hit     0 B/op   0 allocs/op    23 ns/op
record(), uncacheable 424 B/op   4 allocs/op   211 ns/op
spanMetaFor(), hit      0 B/op   0 allocs/op    18 ns/op
```

The uncacheable figure is exactly the baseline, so traffic that cannot be cached is not regressed.

### Gates

`gofmt`, `go vet`, `go build ./...`, `-race` on the http and metrics trees, `golangci-lint --new-from-rev=origin/development` → **0 issues**, 100% coverage on every new func, 0 behind `development`, and CI green on all 18 checks at this head.

### One thing I pushed back on

I left the `attrsRecorder` thread open rather than resolving it myself, since I declined the suggestion — reply is there. Short version: `metricsAttrer` predates this PR on `development` and is a structural surface, so collapsing the tier would remove shipped capability from an external `Manager` as a side effect of a perf change. It is now tested and bounded, where before it was neither. If you would still rather it go, say so and it is a follow-up.

I also corrected three sets of numbers in the description that did not reproduce — the ns/op pair, and the RSS table above.

Umang01-hash (approved):
Ran the benchmarks locally and the perf claim checks out: RecordHistogramOpt is 0 B/op, 0 allocs/op (~110ns) vs Attrs at 424 B/op, 4 allocs/op (~290ns), and since the default metricsManager implements the Opt interface, real apps get the zero-alloc cached path — not just the benchmark.

The part I looked hardest at was cache poisoning, and it's handled well: cacheableMethod pins the key to the 9 standard methods (net/http accepts any token as a method, so this genuinely closes an unbounded-growth vector), the templated gate keeps unmatched paths out, and RouteTemplate returns the template (/users/{id}) not the concrete path, so cardinality stays bounded. All three recorders emit status as a String, so no split time series. Race-clean, lint-clean.

LGTM.

PR #3973, opened 2026-08-18
perf(http): pool the response envelope and stop recanonicalising Content-Type

aryanmehrotra:
**TL;DR — 2 fewer allocations per JSON response. No behaviour change.**

## The problem

Almost every request ends in a JSON response. Each one:

1. **boxed the envelope.** `json.Encoder.Encode` takes `any`, so passing the response struct **by value** boxed it — one allocation per response.
2. **rebuilt a constant header.** `Header().Set("Content-Type", ...)` recanonicalises the key and allocates a fresh `[]string` for a value that never changes.

## The fix

- Put the envelope in the pool **next to the buffer and encoder**, and encode through a **pointer** to it. `encoding/json` produces identical output either way.
- Resolve the canonical `Content-Type` key and its value slice once, and assign the map entry directly.

The encoder has to be pooled with its buffer regardless, since a `json.Encoder` is bound to the writer it was built with.

## The numbers

| | allocs/op | B/op |
|---|---|---|
| before | 9 | 288 |
| **after** | **7** | **224** |

## Why this needed a new benchmark

The existing whole-request benchmarks register plain `http.HandlerFunc` values straight on the router, so they **never reach the responder** — this work measured as "no change" and was nearly discarded as pointless. `BenchmarkResponder_Respond` covers the path for the first time.

## Safety

Reusing a struct that holds caller data needs its own guard, so the envelope is **cleared before it returns to the pool**, with tests pinning that payload, metadata and error can never survive into the next response — including under `-race`.

A `Content-Type` already set by a handler still wins, exactly as before, pinned by its own test.

### Fixed after review

- **The `Content-Type` default is a value check again.** The old guard was `Header().Get("Content-Type") == ""`, which is empty for an absent key *and* for a key explicitly set to `""`. The rewrite replaced it with a presence check, so a caller doing `header["Content-Type"] = []string{""}` shipped a JSON body with a **blank** `Content-Type` — the present-but-empty entry also suppresses net/http's own sniffing, so nothing filled it back in.
- **The pooled envelope now has a real leak guard.** The three "envelope does not leak" tests stayed green with `putRespBuf`'s `re.resp = response{}` **deleted** — every JSON path fully reassigns `re.resp` via a composite literal before encoding, so the clear is only load-bearing for GC retention, which none of them asserted. `TestPutRespBufClearsTheEnvelope` holds the pooled encoder and inspects it directly.
- Both guards were **mutation-checked**: reverting the `Content-Type` guard to a presence check, and deleting the envelope clear, each fail the suite.

## Measurement

Allocations and bytes only. Wall-clock on the machine used varied by several times between runs of the same binary.

## Verification

Rebased onto current `development` (`cf20ee09`, i.e. after both #3770 and #3974 landed). The patch is unchanged by the rebase — the only difference is hunk line offsets, because #3974 added `ResponderFor` nine lines above `Respond`.

**Every response, compared byte for byte** — two real GoFr servers, one per tree, differenced automatically:

```
348 request cases    0 CHANGED    348 IDENTICAL
```

4 `Bind` routes × 10 content types × 3 bodies, plus 19 routes × 3 origins × GET/OPTIONS/DELETE/HEAD — covering `Response`, `*Response`, `Raw`, `Raw`+error, `Redirect`, `*Redirect`, `File`, `XML`, typed errors, path and query params, `/panic`, both `.well-known` probes and an unregistered route. Status, every header and exact body bytes.

**The pooled envelope cannot leak between responses:** 25,600 concurrent responses across 64 goroutines under `-race` — every body carried only its own goroutine's marker, and every `Content-Type` was correct. 0 bad, 0 races.

**Both guards re-checked by mutation**, on the rebased branch:

| mutation | result |
|---|---|
| `Content-Type` guard → presence check | `TestRespondContentTypeDefaultAppliesToPresentButEmpty` fails (both subtests) |
| delete `re.resp = response{}` in `putRespBuf` | `TestPutRespBufClearsTheEnvelope` fails |

**Whole repository suite:** 49 ok / 7 FAIL on both trees. Two packages swap between runs — `pkg/gofr/service` and `pkg/gofr/logging/remotelogger` — and both are pre-existing flakes at the same rate on each tree (measured: `remotelogger` 1 of 6 on both, `service` 0 of 6 on both). `pkg/gofr/http`, the package this PR touches, is **0 failures in 6 runs under `-race` on both**.

`go build ./...` clean; `golangci-lint --new-from-rev=origin/development` reports 0 issues; 0 behind `development` and merges clean.

## Also fixes the benchmark that measures this

`BenchmarkRequest_RealHandlerPath` and the three `BenchmarkRequest_FullChain*` benchmarks share one `benchDiscardResponseWriter` and never cleared its header map, so every header a response wrote survived into the next iteration.

That matters for anything written only when absent. `Content-Type` is exactly that: `Respond` sets it on iteration 1 and finds it present on every one after, so the work reads as free and the benchmark understates this PR by one allocation.

```
                           development    this branch
  writer reused            3690 B / 41    3642 B / 40
  writer reset per iter    3706 B / 42    3642 B / 40   <- real
```

The second row is the honest end-to-end figure — **2 allocations and 64 bytes per request**, matching what `BenchmarkResponder_Respond` reports for the same change in isolation. The reset lives on the writer rather than in each loop, so the four benchmarks cannot drift apart.

---
> **Previously stacked on #3770**, which has since merged (as has #3974). This PR now targets `development` directly and has been rebased onto it.
>
> The review originally worried that a conflict resolver would drop `derefSpecialResponse` and JSON-envelope pointer-special responses. That concern is moot: #3770 **reverted** the pointer dereference, so pointers to the special response types are enveloped on `development` too and there was nothing to lose. `TestResponder_Char_SpecialTypePointersAreEnveloped` pins it, and it runs here.


PiyushSingh-ZS:
Reviewed against `development`, built and ran the package tests on the branch. The pooling logic itself checks out — but there's one reproducible behavioral regression.

### Regression: an explicitly-empty `Content-Type` is no longer defaulted

The check changed from a value test to a key-presence test:

```go
if _, ok := header[canonicalContentType]; !ok {
    header[canonicalContentType] = jsonContentType
}
```

`Header.Get` returns `""` both when the key is **absent** and when it's **present with an empty value**. The map lookup only covers the first case. Same probe on both branches:

| branch | resulting header |
|---|---|
| `development` | `Content-Type: application/json` |
| this branch | `Content-Type:` (empty) |

This is reachable from user code — `response.Response{Headers: map[string]string{"Content-Type": ""}}` goes through `SetCustomHeaders`, which runs `Set` before `Respond` (handler.go:94). The JSON body then goes out with an empty Content-Type, and an empty-but-present Content-Type also suppresses net/http's sniffing, so the client gets nothing usable.

Suggested fix:

```go
if v := header[canonicalContentType]; len(v) == 0 || v[0] == "" {
    header[canonicalContentType] = jsonContentType
}
```

### The `maxRespPooledBuf` cap is bypassed under jsonv2

Under `GOEXPERIMENT=jsonv2` (opt-in in Go 1.25+, eventual default), `json.Encoder` keeps its own internal `enc.buf`, which `Encode` resets per call but never shrinks:

```go
buf := &enc.buf
buf.Reset()
if err := jsonv2.MarshalWrite(buf, v, enc.opts); err != nil {
```

`putRespBuf` only inspects `re.buf.Cap()`, so the guard whose entire purpose is "one oversized response must not permanently inflate every pooled buffer" no longer covers the buffer that actually held the encoded bytes. Worth capping on both, since the encoder is now pooled alongside the buffer.

### Things I checked and believe are fine

- **Encoder reuse.** Verified in the toolchain source for both the v1 and jsonv2 implementations: `Encode` carries no state between calls, and `enc.err` is only assigned on a *writer* error. The writer is a `bytes.Buffer`, whose `Write` never returns an error, so a marshal failure cannot poison a pooled encoder. The error path also leaves the buffer untouched, since `e.marshal` writes into the encodeState, not into `enc.w`.
- **Sharing `jsonContentType`.** `len == cap == 1` holds, so a later `Header.Add` on the same key appends into a newly allocated array. The reasoning in the comment is right.
- **Envelope clearing.** `re.resp = response{}` on Put correctly stops a payload outliving its response inside the pool.

### Nit

The doc comment for `getRespBuf` ended up orphaned above `newRespEncoder`, so `getRespBuf` now reads as if it were undocumented and `newRespEncoder` carries two stacked comments.

NitinKumar004 (changes requested):
Clean, well-scoped optimization with good pool-reuse/race tests. Two small items plus a merge-order note.

### Low
- **Content-Type default skipped when the key is present-but-empty.** The old guard `Header().Get("Content-Type") == ""` is a value check (empty for absent *or* empty value); the new `if _, ok := header[canonicalContentType]; !ok` is a presence check. If a caller pre-sets `Content-Type` to `""` (`header["Content-Type"] = [""]`), the old code shipped `application/json` while the new code leaves it present-and-empty and also suppresses net/http sniffing → JSON body with a **blank Content-Type**. Pathological/unused in-tree, but a real edge. Restore a value check, e.g. `if len(header[k]) == 0 || header[k][0] == ""`.
- **The three "envelope does not leak" tests stay green even if `putRespBuf`'s `re.resp = response{}` clear is deleted.** Every JSON path fully reassigns `re.resp` via a composite literal before encode, so the clear is only load-bearing for GC retention (not asserted). A future refactor that stops fully reassigning `re.resp` on some path would leak prior metadata/error and these tests wouldn't catch it. A real guard test should retain the pooled `*respEncoder` and assert fields are zeroed after `putRespBuf`.

### ⚠️ Cross-PR — merge order vs #3770
The PR body says callers can merge in any order; that's true vs the sibling perf PRs but **not** vs #3770, which edits the same `responder.go` regions (inserts `data = derefSpecialResponse(data)` at the top of `Respond`, and the deref helpers after `putRespBuf`). If #3770 lands first and this PR is rebased, a resolver taking this PR's rewritten `Respond` wholesale drops the deref line → a handler returning a **pointer** to a special type (`*Redirect`/`*Response`/…) gets JSON-enveloped with 200 instead of 302/etc. Neither PR has a test for pointer-special types.

**Requested changes:** restore the value-check Content-Type default; add a real pooled-envelope leak guard; coordinate the `responder.go` merge with #3770 so `derefSpecialResponse` stays ahead of this PR's block.

pkg/gofr/http/responder_test.go
```diff
@@ -1830,3 +1832,257 @@ func TestResponder_Char_MethodCaseSensitivity(t *testing.T) {
...
+		require.Contains(t, body, want)
+		require.Equal(t, 1, strings.Count(body, "payload-"),
+			"a response must carry only its own payload, got %q", body)
+	}
+}
+
+// TestRespondEncoderConcurrent runs the pooled encoder under contention; the
+// race detector is the point of this test.
+
+// TestRespondEncoderConcurrent runs the pooled encoder under contention; the
```
Umang01-hash, line 1856: Nit: this doc comment is duplicated verbatim (same doubling on TestRespondEnvelopeDoesNotLeakMetadata and TestRespondEnvelopeErrorDoesNotLeak) — drop the first of each pair.

pkg/gofr/http_server_test.go
```diff
@@ -433,6 +446,7 @@ func BenchmarkRequest_FullChain_JSON(b *testing.B) {
 	b.ResetTimer()
 
 	for i := 0; i < b.N; i++ {
+		w.reset()
```
Umang01-hash, line 456: Nit: the `/json` route here is a raw `http.HandlerFunc` (`w.Write([]byte(...))`) that never reaches `Responder.Respond`, so this benchmark stays 22 allocs and doesn't exercise the change — the reset is inert on _JSON (correct and needed on the FullChain benches that do route through Respond). The responder-level benches are what demonstrate the win; worth a note so nobody cites _JSON as evidence.

Umang01-hash (dismissed):
Verified locally against this head. The allocation win is real and I measured it base-vs-head with an identical harness: Respond_Map 128 B/4 allocs -> 80 B/3, ResponderRespond 72 B/2 -> 24 B/1. Output is byte-identical (the existing exact-body suite passes), the pool is race-clean under contention (32x25, plus the metadata/error isolation tests), the by-pointer envelope is encoded synchronously before it returns to the pool and is cleared on Put, the shared Content-Type slice can't be mutated (len==cap==1), and no exported signature changed. gofmt/vet/golangci --new-from-rev all clean. Nice change. Two cosmetic nits inline.

pkg/gofr/http/responder_test.go
aryanmehrotra: Fixed in `09d23667`, and the same shape turned out to be in the production file where the review did not reach.

`getRespBuf`'s doc comment had been left stranded above `newRespEncoder`, so `newRespEncoder`'s comment opened with a paragraph about a different function and **`getRespBuf` had no doc at all**. Split them. `putRespBuf`'s comment also still described a `buf` parameter that is now an `re`.

Both duplicated test comments are gone too.

pkg/gofr/http_server_test.go
aryanmehrotra: Correct, and I have put it in the comment rather than just agreeing with it — `/json` is registered as a raw `http.HandlerFunc` that writes its bytes directly, so it never reaches `Responder.Respond`. `09d23667` says so on the benchmark, names the responder-level benchmarks as the ones that demonstrate the change, and notes the writer reset is kept only for symmetry with the siblings where it is load bearing.

Good catch — that is exactly the sort of number that gets quoted later as evidence for something it never touched.

aryanmehrotra:
## Leak testing against `development`, and a test of my own that proved nothing

@Umang01-hash — pushed `09d23667` (comments only), which dismissed your approval; sorry. Both your nits are fixed and resolved, and one of them had a twin in the production file that neither of us had looked at: `getRespBuf`'s doc comment was stranded above `newRespEncoder`, leaving `getRespBuf` undocumented and `newRespEncoder`'s comment opening with a paragraph about a different function.

Pooling is the part of this worth being paranoid about, so here is everything I ran against `development` at `a513230c0`.

### The finalizer leak test I wrote is vacuous, and I want that on the record

I built a `runtime.SetFinalizer` probe to prove the payload becomes collectable after the response. It passes — **and it also passes with `re.resp = response{}` deleted**, so it demonstrates nothing about the clear.

The reason is worth writing down: **`runtime.GC()` drains a `sync.Pool`** (two cycles, because of the victim cache). Any reference the pool was holding is dropped by the very GC the probe needs in order to observe the finalizer. A `sync.Pool` retention is structurally invisible to a finalizer test.

So the honest scope of that clear is narrower than "prevents a leak": a payload can be reachable from a pooled envelope only until the next GC, which is what `sync.Pool` guarantees regardless. What it does buy is that the window is zero rather than one GC cycle, and `TestPutRespBufClearsTheEnvelope` pins it directly by inspecting the struct — which is the right way to test it.

### What does hold up

**Live heap after a full GC**, 192,000 responses per revision at 8 KiB payloads, mixed plain / envelope-with-metadata / raw, 32 goroutines:

| round | `development` | this branch |
|---|---:|---:|
| 1 | +62,368 B | +68,128 B |
| 2 | +93,984 B | +84,656 B |
| 3 | +108,512 B | +105,984 B |

Both plateau at ~3.6 MB live heap and stay there. No divergence, so nothing is being retained that `development` was not already retaining.

**Live server RSS**, 102,400 requests per run including a 200 KiB route that deliberately exceeds `maxRespPooledBuf`:

```
development  warm 67.9MB  r1 70.9  r2 71.6  r3 71.1
this branch  warm 65.1MB  r1 73.7  r2 74.0  r3 74.1
```

Flat across rounds on both — a leak climbs with request count, and neither does. The ~3 MB higher plateau is the pooled encoder itself; the post-GC live-heap numbers above are the precise measurement and they match.

**Cross-request contamination, in process**: 48 goroutines × 300 iterations × 2 runs under `-race`, alternating metadata / no-metadata so a stale envelope surfaces as a plain response inheriting the previous request's metadata. Clean on both.

**Cross-request identity, through a real server**: 25,600 concurrent requests per revision, each asking for a unique value, half requesting metadata:

```
development  ok=25600  wrong_payload=0  wrong_metadata_shape=0  transport_failed=0
this branch  ok=25600  wrong_payload=0  wrong_metadata_shape=0  transport_failed=0
```

**Output equivalence**: 1,008 response shapes — 21 payloads (nil, nil-pointer-in-interface, HTML-escaping, unicode, empty/nil slices, `Raw`, `Response` with and without metadata, a 70 KiB body, an unmarshalable channel) × 4 error kinds × 4 Content-Type presets (absent, `text/plain`, explicitly empty, non-canonical key) × 3 methods — comparing status, the full header map and exact bodies. **3,276 lines byte-identical.** The comparator is live: appending `; charset=utf-8` to `jsonContentType` produces 1,512 diff lines.

Through the real server, captured *after* 44,800 concurrent requests had churned the pool: identical on both, including `{"metadata":{"page":1},"data":{"name":"d"}}`, the 404/500 error envelopes, and the 71,692-byte response that is correctly dropped rather than pooled.

### The hazard I looked hardest at

`json.Encoder` carries a **sticky `e.err`**: once a write fails, every later `Encode` on that encoder returns the old error without encoding. A poisoned encoder returned to the pool would break every subsequent response that drew it. It cannot happen here — the encoder's writer is always the paired `*bytes.Buffer`, and `bytes.Buffer.Write` never returns an error — but it is the reason the encoder must never be pooled apart from its buffer, and worth stating.

The rest of the lifetime argument checks out: the envelope is encoded synchronously through a pointer, `buf.Bytes()` is written *before* `putRespBuf`, and the encode-error path calls `putRespBuf` and then writes a literal that never touches the buffer.

### Guards are pinned, not decorative

Each removed, each caught:

| mutation | fails |
|---|---|
| delete `re.resp = response{}` | `TestPutRespBufClearsTheEnvelope` |
| weaken the Content-Type guard to presence-only | `TestRespondContentTypeDefaultAppliesToPresentButEmpty` |

### Gates

`gofmt`, `go vet`, `go build ./...`, `-race` across `./pkg/gofr/http/...`, `golangci-lint --new-from-rev=origin/development` → **0 issues**, 0 behind `development`. The only local test failure is `pkg/gofr` refusing port 2121, which is occupied on this machine and fails **identically on `development`** with zero test-level failures on either side.

Umang01-hash (approved):
Re-reviewed at 09d23667. The only change since my last pass is the docs-only commit — `git diff` shows zero non-comment lines touched in responder.go, and it resolves the duplicate doc comments and adds the note that FullChain_JSON doesn't exercise the pooling. gofmt/vet/-race all clean. Still LGTM.

PR #3974, opened 2026-08-18
perf(logging,http): skip request logs the level discards, and co-allocate the request objects

aryanmehrotra:
**TL;DR — 8 fewer allocations and 225 fewer bytes on every request. No behaviour change.**

Measured on the path a real GoFr handler takes, against current `development`:

| `BenchmarkRequest_RealHandlerPath` | allocs/op | B/op |
|---|---|---|
| `development` | 49 | 3915 |
| **this branch** | **41** | **3690** |

> **This PR now contains what was #3975.** The two were stacked, and #3975 had a hard compile dependency on `logging.ContextLoggerFor`, introduced here. Split apart, this branch **exported a public function with no production caller** — `pkg/gofr/context.go` still went through `*NewContextLogger(...)`, and only #3975 switched the call site. Shipping that alone would have left `ContextLoggerFor` as permanently dead exported API on a v1 line. Folded together, the export has a caller the moment it exists. #3975 is closed; its review history stays readable there and every one of its commits is in this branch, unchanged.

---

# Part 1 — stop building request logs the level will discard

**1. The `ContextLogger` — every service, every level.**

One is built per request. It eagerly formatted the 32-character trace ID and materialised a one-entry `map[string]any` marker at construction, and heap-allocated the wrapper itself — all before knowing whether the handler would log anything. A handler that logs nothing, the common case on a hot endpoint, paid for all three and used none.

The trace ID is formatted at log time, the marker is a typed string rather than a map, and `ContextLoggerFor` returns the wrapper by value.

**2. The request-log gate — only services above INFO.**

`handleRequestLog` builds a `RequestLog`, formats a timestamp and resolves the client IP before the logger decides whether to emit. An optional `LogEnabled()` lets the middleware ask first.

**The reach here is narrower than it may read.** The request log goes through `Log`, which logs at **INFO** — not DEBUG. `GetLevelFromString("")` returns `INFO`, so on a **default-configured service the gate does not fire** and nothing is skipped. The saving lands on services that have raised `LOG_LEVEL` to NOTICE or above — a real deployment shape, but not the default one.

### Numbers

**Middleware, per request** — a real `*logging.logger`, not `NewMockLogger` (which has no `LogEnabled`, so the assertion failed and the original benchmark measured the pre-optimization path):

| `LOG_LEVEL` | B/op | allocs/op |
|---|---|---|
| unset → INFO (**default**) | 1578 → 1562 | 17 → **16** |
| NOTICE / WARN / ERROR | 1441 → **1248** | 14 → **10** |

At the default level the change is the correlation-ID header write alone.

**`ContextLogger`, traced request:**

| shape | base | head |
|---|---|---|
| built, nothing logged | 432 B · 5 allocs | **0 B · 0 allocs** |
| built + 2 discarded lines | 528 B · 9 allocs | **192 B · 8 allocs** |
| each further discarded line | 48 B · 2 allocs | 96 B · 4 allocs |

The last row is the cost side, and it is deliberate: work moves from per-request to per-log-call. Measured crossover, discarded lines per request — **allocation count** breaks even at **3**, **bytes** not until **9**:

| lines | 0 | 1 | 2 | 3 | 8 | 9 | 10 |
|---|---|---|---|---|---|---|---|
| base B/op | 432 | 480 | 528 | 576 | 816 | 864 | 912 |
| head B/op | **0** | **96** | **192** | **288** | **768** | 864 | 960 |

A silent request is free, asserted by `TestContextLoggerSilentRequestIsFree`.

### Also in here

`remoteLogger.LogEnabled` took an `RWMutex.RLock` per request. It guarded `currentLevel`, which the method never reads, and `ChangeLevel` is called outside `r.mu` anyway — so it bought no ordering while adding a process-wide reader-count bump to the hot path. Dropped; the optional-interface assertion moved to `New`.

The log level was written by `ChangeLevel` from the remote logger's polling goroutine and read unsynchronized by `logf`. That predates this branch, but `LogEnabled` adds a second reader on the per-request path, so `level` is an `atomic.Int64` now rather than being leaned on. Under `-race`, `./pkg/gofr/logging/remotelogger` drops from **3 races citing `logger.go` to 0**; the remainder are the pre-existing `httpLogFilter` races, present identically on `development`.

---

# Part 2 — give the request Context, Request and Responder one allocation

*(this was #3975)*

**1. Three heap objects where one would do.** The `Context`, the `Request` wrapper it holds through an interface, and the `Responder` it holds through another. All three are created together, live for exactly one request, and are discarded together. `newHTTPContext` now places the concrete `Request` and `Responder` **inside the same struct** as the `Context` and points the interfaces at them.

**2. A trace ID nobody reads.** `handler.ServeHTTP` did this on every request:

```go
traceID := trace.SpanFromContext(r.Context()).SpanContext().TraceID().String()
```

`TraceID().String()` allocates a **32-character string**, and its only consumer is `logError` — which runs **only when a handler returns an error**. The `SpanContext` is a value type, so carrying it costs nothing; the string is built where it is used, on the error path.

### This part also fixes a measurement blind spot

GoFr had **no benchmark covering its own handler path**. The existing whole-request benchmarks register plain `http.HandlerFunc` values on the router, so they never enter `handler{}` — the Context, Request and Responder are never built and the trace ID is never resolved. Work on exactly that path measured as *"no change"*, which is how this series nearly got discarded as pointless. `BenchmarkRequest_RealHandlerPath` closes that gap, and the middleware chain it uses is defined once in `useGoFrMiddleware` so a hand-rolled copy cannot drift out of sync with `newHTTPServer`.

---

## Safety

- **No behaviour change.** Same log output, same trace IDs, same levels, same JSON envelopes. Verified line-by-line against `development` — see below.
- The request-log gate fails open in both directions that matter: an unknown logger and any 5xx response still build the entry.
- `logf` and `LogEnabled` share one `enabled(Level)` predicate, so they cannot drift.
- `RequestFor` / `ResponderFor` return **values** so a caller that already owns an allocation need not take another. `NewRequest` / `NewResponder` are unchanged and delegate to them — purely additive.
- CMD, GraphQL, cron and subscribers supply their own `Request` implementations and keep using `newContext`, untouched.
- Mutation-checked: `remoteLogger.LogEnabled → return false` and `enabled`'s `>=` → `>` each fail the suite.

## Verification

Because this PR now also touches `context.go`, `handler.go`, `request.go` and `responder.go`, the whole request path is in scope — not just log output.

**Every response, compared byte for byte.** Two real GoFr servers, one per tree, same handlers, differenced automatically:

```
348 request cases    0 CHANGED    348 IDENTICAL
```

4 `Bind` routes × 10 content types × 3 bodies, plus 19 routes × 3 origins × GET/OPTIONS/DELETE/HEAD — covering `Response`, `*Response`, `Raw`, `Raw`+error, `Redirect`, `*Redirect`, `File`, `XML`, typed errors, path and query params, `/panic`, both `.well-known` probes and an unregistered route. Status, every header and exact body bytes.

**Every log line, at every level:**

```
LOG_LEVEL=UNSET    26 lines  IDENTICAL
LOG_LEVEL=DEBUG    30 lines  IDENTICAL
LOG_LEVEL=INFO     26 lines  IDENTICAL
LOG_LEVEL=NOTICE   12 lines  IDENTICAL
LOG_LEVEL=WARN     10 lines  IDENTICAL
LOG_LEVEL=ERROR     8 lines  IDENTICAL
LOG_LEVEL=FATAL     1 lines  IDENTICAL
```

Timestamps, trace/span IDs, response times and addresses are normalized; nothing else is. The panic path's stack frames are byte-identical too.

**WebSocket**, the retention path this PR's own docs call out — `App.WebSocket` assigns `ctx.Request = conn` and loops for the life of the socket, so the co-allocated struct is held for the whole connection:

```
development  upgrade=101 | m0->echo:m0 | m1->echo:m1 | m2->echo:m2 | m3->echo:m3 | m4->echo:m4
this branch  upgrade=101 | m0->echo:m0 | m1->echo:m1 | m2->echo:m2 | m3->echo:m3 | m4->echo:m4
```

**Whole repository suite**, package by package: **50 ok / 6 FAIL on both trees, and no package differs in pass/fail** — same three named failing tests. All six are environmental here: the `examples/*` integration tests need Redis / MySQL / Kafka, and `pkg/gofr` is killed by whatever holds the metrics port.

**Cascade.** Merges into `development` clean. Every other open PR in this series still merges clean alongside it, and all six merged together build, pass `-race`, and serve the same 348 cases identically. Stacked, the six come to **24 allocs · 2647 B** on `BenchmarkRequest_RealHandlerPath`, down from 49 · 3915.

`go build ./...` clean; `-race` green on `http`, `http/middleware`, `http/response` and `logging`; `golangci-lint --new-from-rev=origin/development` reports 0 issues; 0 behind `development`.

> **Note on `pkg/gofr/logging/remotelogger`:** it fails for reasons unrelated to this PR and fails the same way on pristine `development` — 4 tests on each under `-race`, passing without it. Pre-existing, and worth its own issue.

## Not addressed here

`pkg/gofr/grpc/log.go` builds its entry unconditionally on every RPC — the same shape, on the gRPC side. A literal port does not fit: `logGRPCEntry` routes health-check and stream traffic to `Debug`, and `LogEnabled()` only answers for INFO. That needs a level-parameterized form, and is better decided in its own PR than bolted on here. Thread left open for @arunesh-j.

`newContext` and `newCMDContext` — the pubsub and CMD paths — still use `*NewContextLogger(...)`. Neither is a per-request hot path, and `NewContextLogger` stays as the pointer constructor for callers that want one.


PiyushSingh-ZS:
Reviewed against `development`, built and tested on the branch. The gate is correct: `LogEnabled()` returns `INFO >= l.level`, which is exactly what `Log` → `logf(INFO, ...)` filters on, so no entry that would have been emitted is now dropped. Deferring the span-ID formatting past the gate is a clean win, and keeping the optional interface free of the `logging.Level` type is the right call for the middleware's minimal contract.

Baseline note: `pkg/gofr/logging/remotelogger` fails under `-race` on this branch, but it fails identically on `development` (unsynchronized `l.level` between `ChangeLevel` and `logf`), so that's pre-existing and not from this PR.

### `remoteLogger.LogEnabled` takes a lock it doesn't need

```go
func (r *remoteLogger) LogEnabled() bool {
	r.mu.RLock()
	defer r.mu.RUnlock()

	if e, ok := r.Logger.(interface{ LogEnabled() bool }); ok {
```

Nothing inside the critical section reads state the mutex guards — `r.currentLevel` isn't touched, and `r.Logger` is fixed at construction. So this acquires an RWMutex on every request to protect an immutable field. Given the whole point of the PR is removing per-request work, dropping the lock seems in keeping.

Related: the level this ultimately reads (`l.level` on the embedded logger) is itself unsynchronized against `ChangeLevel`, so the lock wouldn't be buying correctness even if it were on the right field. That's the pre-existing race above — worth fixing separately with an atomic, and it would make `LogEnabled` genuinely safe rather than incidentally so.

### The comment about Error isn't quite true

> A server error is reported through Error, which is emitted at every level this logger supports

At `FATAL`, `ERROR < FATAL`, so error-level request logs are suppressed too. The consequence is only a missed optimization (the entry gets built and then discarded on 5xx at FATAL level), not wrong output — but the comment states an invariant that doesn't hold, and someone will rely on it later.

### Compat note on `traceIDMarker`

Custom `Logger` implementations previously received the marker as `map[string]any{"__trace_id__": ...}` and now receive `traceIDMarker`, an unexported string type. GoFr's own logger strips both, but a third-party logger that renders unknown args will now print a bare trace-ID string where it used to print a recognizable map. The map form is retained for the `pkg/gofr/ai` contract, so this is only about implementations outside the framework — probably worth a line in the release notes rather than a code change.

pkg/gofr/http/middleware/logger.go
```diff
@@ -135,11 +136,29 @@ func colorForStatusCode(status int) int {
 	return 0
 }
 
+// canonicalCorrelationID is the canonical spelling of the correlation-ID header,
+// resolved once so no request pays to build it.
+//
+//nolint:gochecknoglobals // immutable, process-wide header constant.
+var canonicalCorrelationID = textproto.CanonicalMIMEHeaderKey("X-Correlation-ID")
```
arunesh-j: Nit: this can be a `const`.

`textproto.CanonicalMIMEHeaderKey("X-Correlation-ID")` is fixed at compile time, and this PR's own test already pins the result: `require.Equal(t, "X-Correlation-Id", canonicalCorrelationID)`.

```go
const canonicalCorrelationID = "X-Correlation-Id" // textproto.CanonicalMIMEHeaderKey("X-Correlation-ID")
```

Same map key, but it drops a package-level mutable variable (anything in the package could reassign it and silently change the emitted header name), drops the `//nolint:gochecknoglobals`, and drops the init-time call. The `textproto` import goes with it.

pkg/gofr/http/middleware/logger.go
```diff
@@ -201,16 +226,30 @@ func Logging(probes LogProbes, logger logger) func(inner http.Handler) http.Hand
...
 
 			inner.ServeHTTP(srw, r)
 		})
 	}
 }
 
-func handleRequestLog(srw *StatusResponseWriter, r *http.Request, start time.Time, traceID, spanID string, logger logger) {
+func handleRequestLog(srw *StatusResponseWriter, r *http.Request, start time.Time, traceID string,
+	sc trace.SpanContext, logger logger) {
 	status := srw.Status()
```
arunesh-j, line 238: Small accuracy nit on the justification, not the code.

`Error` goes through `logf(ERROR, …)`, which drops the entry when `ERROR < l.level` — so at `LOG_LEVEL=FATAL` a 5xx request log is discarded like any other. It isn't emitted at *every* level.

The code is right either way: skipping the gate for 5xx fails open, which is the safe direction. Only the stated reason overshoots. Something like "…which survives every level below FATAL, so gating it on the informational level would be wrong" keeps it true.

arunesh-j (commented):
## Review — perf(logging): stop building request logs the level will discard

**Comment** — the change is sound and the win is real; what needs work is the evidence around it. The benchmark meant to demonstrate the gate never reaches it, and the headline saving is narrower than the PR reads.

`0 blockers` · `5 medium` · `4 nit` · 8 files (+369/-54)

### Verified at head (`7dde0dc`) against merge-base `878b026`

Built an isolated worktree at the head sha and ran it:

| Check | Result |
|---|---|
| `go build ./...`, `go vet` on the three changed packages | clean |
| `go test` — logging, logging/remotelogger, http, http/middleware, http/response, ai | all pass |
| `go test -race ./pkg/gofr/logging/remotelogger` | fails — but **fails identically on the merge-base** (11/10/8 races on base vs 12/12/14 on head across three runs), and no stack touches code this PR adds. Pre-existing, not yours. |
| `go test ./pkg/gofr/` | fails on base *and* head — `metrics port 2121` already bound on the machine I ran on. Environmental. |
| `golangci-lint 2.11.4` on the changed packages | 6 issues, none on lines this PR touches |
| Raw `X-Correlation-Id` bytes over a real TCP connection, base vs head | **byte-identical** — the `Header().Set` → direct map write is safe |

### Measured (`-count=3`; allocations, since ns/op with a real writer is noisy)

| Shape | Base | Head |
|---|---|---|
| Traced request that logs nothing | 132 ns · 432 B · 5 allocs | **24.7 ns · 80 B · 1 alloc** |
| Each emitted traced log line | 462 ns · 216 B · 6 allocs | 490 ns · 264 B · **8 allocs** |
| Request + 3 emitted lines | 1485 ns · 1081 B · 23 allocs | 1458 ns · **792 B** · 24 allocs |
| Middleware, `LOG_LEVEL=WARN` (gate fires) | 1440 B · 14 allocs | **1248 B · 10 allocs** |
| Middleware, `LOG_LEVEL` unset → INFO (the default) | 1577 B · 17 allocs | 1562 B · **16 allocs** |

The `ContextLogger` change is an unambiguous win: a traced request that never logs drops from 5 allocations to 1. The documented per-emitted-line regression is real (+2 allocs, +48 B) but breaks even at roughly two log lines per request, and total bytes stay favourable well past that — the trade is the right one.

The middleware gate is where the framing needs adjusting: `GetLevelFromString("")` returns `INFO`, and the gate is `INFO >= l.level`, so on a **default-configured GoFr service the fast path never fires** (17 → 16 allocations, which is the header change alone). The 14 → 10 result needs `LOG_LEVEL` at NOTICE or above. Worth saying so explicitly in the PR body so the number isn't read as universal.

### What holds up

- The gate fails open in both directions that matter — an unknown logger and any 5xx response still build the entry — so it cannot silently drop a line that used to be emitted. Confirmed by reading every concrete logger the middleware can receive.
- `LogEnabled()`'s predicate exactly matches `logf(INFO, …)`'s. Flipping it to `INFO > l.level` **is** caught, by `TestRequestLogSchemaSnapshot`.
- `logEnabler` is scoped to the middleware package rather than widening the shared `logging.Logger` interface — the right call.
- `ContextLoggerFor` returning by value is behaviourally identical to the old pointer-deref; the `traceIDMarker` migration keeps `pkg/gofr/ai`'s `map[string]any` form working alongside it.
- The pooled `StatusResponseWriter` is released by its own deferred closure, so the new early return in `handleRequestLog` does not leak it.

### Not addressed here

`pkg/gofr/grpc/log.go` still builds its entry unconditionally on every RPC — the same shape this PR fixes on the HTTP side. Details inline; it needs a slightly different API than `LogEnabled()` offers today, so it is worth deciding now while that API is new.

pkg/gofr/http/middleware/logger.go
```diff
@@ -201,16 +226,30 @@ func Logging(probes LogProbes, logger logger) func(inner http.Handler) http.Hand
...
 
+	// A server error is reported through Error, which is emitted at every level
+	// this logger supports, so only the informational path can be skipped.
+	if status < http.StatusInternalServerError {
+		if e, ok := logger.(logEnabler); ok && !e.LogEnabled() {
```
arunesh-j, line 243: Worth stating the gate's actual reach in the PR body — it is narrower than the description reads.

`GetLevelFromString("")` returns `INFO` (`pkg/gofr/logging/level.go:91`), and this gate is `INFO >= l.level`. So on a service that hasn't set `LOG_LEVEL` — the default — `LogEnabled()` returns true and nothing is skipped.

Measured through the middleware with a real `*logging.logger`, 300k iterations, `-count=3`:

| `LOG_LEVEL` | base | head |
|---|---|---|
| unset → INFO | 1577 B · 17 allocs | 1562 B · 16 allocs |
| WARN | 1440 B · 14 allocs | 1248 B · 10 allocs |

So the 4-allocation saving lands only on services that have explicitly raised the level to NOTICE or above; at the default the measurable gain is the `X-Correlation-Id` change alone. That is still a fine thing to ship — it just reads as a universal per-request win right now, and it isn't.

pkg/gofr/http/middleware/logger_test.go
```diff
@@ -491,3 +493,113 @@ func BenchmarkGetIPAddress(b *testing.B) {
...
+type loggingBenchWriter struct{ h http.Header }
+
+func (w *loggingBenchWriter) Header() http.Header       { return w.h }
+func (*loggingBenchWriter) Write(b []byte) (int, error) { return len(b), nil }
+func (*loggingBenchWriter) WriteHeader(int)             {}
+
+// BenchmarkLogging measures the middleware at a level that DISCARDS the entry.
+//
+// This is the case that matters in production: services run at INFO or above, so the per-request
+// log built at DEBUG is assembled and then thrown away. Anything spent building it is pure waste,
```
arunesh-j: This benchmark doesn't reach the branch it's meant to measure.

`logging.NewMockLogger` returns a `*MockLogger`, which has no `LogEnabled()` method (`pkg/gofr/logging/mock_logger.go` — `Log`, `Logf`, `ChangeLevel`, … but no `LogEnabled`). So `logger.(logEnabler)` fails, the gate is never consulted, and the full `RequestLog` — struct, timestamp format, IP lookup — is still built and then discarded inside `MockLogger.logf`. The reported numbers are the pre-optimization path.

Measured both ways, same benchmark shape, 300k iterations, `-count=3`:

| logger | base | head |
|---|---|---|
| `NewMockLogger(ERROR)` (as written) | 1441 B · 14 allocs | 1425 B · 13 allocs |
| `NewLogger(ERROR)` / `remotelogger.New(ERROR, …)` | 1441 B · 14 allocs | 1249 B · 10 allocs |

The benchmark is understating the change by about 4x. Either give `MockLogger` a `LogEnabled()` — which also makes it behave like the real logger for anyone testing middleware against it — or switch this to `logging.NewLogger(logging.ERROR)`.

The doc comment above needs a small fix too: it says "the per-request log built at DEBUG". `Log()` calls `logf(INFO, …)`, so the entry is built at INFO, and per the gate it is *kept* at `LOG_LEVEL=INFO`. The fast path only fires at NOTICE and above.

pkg/gofr/logging/logger.go
```diff
@@ -129,6 +129,17 @@ func (l *logger) Log(args ...any) {
...
 }
 
+// LogEnabled reports whether an entry written through Log would be emitted at
+// the configured level.
+//
+// It lets a caller that must BUILD an entry before logging it -- the request
+// logger assembles a struct, formats a timestamp and resolves the client IP --
+// skip that work when the entry would be discarded. Callers that already have
+// their arguments to hand gain nothing from it and should just call Log.
+func (l *logger) LogEnabled() bool {
```
arunesh-j, line 158: The same shape is still unguarded on the gRPC side, and it may be worth deciding the API now rather than after this ships.

`pkg/gofr/grpc/log.go` builds its entry unconditionally in both interceptors before the logger sees it:

- `logRPC` (line ~234) builds `gRPCLog{ID: trace.SpanFromContext(ctx).SpanContext().TraceID().String(), StartTime: start.Format("2006-01-02T15:04:05.999999999-07:00"), …}`, then calls `logGRPCEntry`.
- `logStreamRPC` (line ~211) does the same, plus `getTraceID(ctx)`.

`logGRPCEntry` (line ~268) ends in `logger.Info(logEntry)` or `logger.Debug(logEntry)` — the span lookup, the timestamp format and the escape-to-heap all happen first, on every RPC. Exactly what this PR fixes for HTTP.

A literal port doesn't fit, though: `logGRPCEntry` routes health-check / `/Send` / `/Recv` / `/SendAndClose` traffic to `Debug`, and `LogEnabled() bool` only answers for INFO — gating just the `Info` branch would leave the highest-volume stream traffic unfixed. A level-parameterized form (`Enabled(Level) bool`), or a sibling `DebugEnabled()`, would cover both. Cheaper to settle that here than to widen `LogEnabled()` after it has shipped.

pkg/gofr/logging/logger.go
```diff
@@ -129,6 +129,17 @@ func (l *logger) Log(args ...any) {
...
+	return INFO >= l.level
```
arunesh-j: Two things worth a look.

`l.level` is written without synchronization by `ChangeLevel`, and `remotelogger.UpdateLogLevel`'s polling goroutine calls it on a timer. That unsynchronized read already existed in `logf`, so this isn't new — but `LogEnabled` adds a second one, and puts it on the per-request middleware path. `remoteLogger.LogEnabled`'s `r.mu.RLock()` does not cover it: that mutex guards `currentLevel`, not the embedded logger's `level`, and `ChangeLevel` is in fact called *after* `r.mu.Unlock()`. If you're in this code anyway, an atomic for `level` would close it properly.

Separately, the gate condition is now duplicated rather than shared: `logf` uses `level < l.level`, `LogEnabled` uses `INFO >= l.level`. They agree today — I flipped `>=` to `>` and `TestRequestLogSchemaSnapshot` did catch it — but if `logf` ever grows another condition, `LogEnabled` diverges silently, in the direction where request logs get dropped that `logf` would have emitted. One shared helper would make that impossible.

pkg/gofr/logging/logger.go
```diff
@@ -302,6 +321,14 @@ func extractTraceIDAndFilterArgs(args []any) (traceID string, filtered []any) {
 	filtered = make([]any, 0, len(args))
 
 	for _, arg := range args {
+		if tid, ok := arg.(traceIDMarker); ok {
```
arunesh-j, line 346: The two marker branches strip asymmetrically.

The typed branch always strips, regardless of whether `traceID` is already set. The map branch below only strips when `traceID` is still empty — so a map-shaped marker arriving *after* a typed one falls through to `filtered = append(filtered, arg)` and renders into the log message as a literal map.

Not reachable with today's two producers (`ContextLogger` appends the typed marker last, `pkg/gofr/ai` emits only the map), but the map branch already has the same problem for two map markers, and `hasTraceMarker` treats both shapes as markers to strip. Making the continue unconditional on `exists` closes it:

```go
if m, ok := arg.(map[string]any); ok {
    if tid, exists := m[traceIDMarkerKey].(string); exists {
        if traceID == "" {
            traceID = tid
        }

        continue
    }
}
```

pkg/gofr/logging/logger.go
```diff
@@ -320,6 +347,10 @@ func extractTraceIDAndFilterArgs(args []any) (traceID string, filtered []any) {
 // key. Read-only: it allocates nothing.
 func hasTraceMarker(args []any) bool {
 	for _, arg := range args {
+		if _, ok := arg.(traceIDMarker); ok {
```
arunesh-j, line 376: Doc nit: "reports whether any arg is a map carrying the `__trace_id__` key" is now only half of what this does — it also matches the typed `traceIDMarker`, which is the shape `ContextLogger` actually emits after this PR. Worth updating so the next reader doesn't trust the comment over the code.

pkg/gofr/logging/remotelogger/dynamic_level_logger.go
```diff
@@ -161,6 +161,25 @@ type remoteLogger struct {
...
 
+// LogEnabled reports whether an entry written through Log survives the level
+// currently in force.
+//
+// The embedded logger is the authority: this type only tracks the level so it
+// can push changes down via ChangeLevel, so asking the logger itself stays
+// correct even between a remote update and the level being applied. An embedded
+// logger that cannot answer is treated as enabled, which preserves the previous
+// behavior of always building the entry.
+func (r *remoteLogger) LogEnabled() bool {
```
arunesh-j, line 198: This method is the one production actually uses — `container.go:126` always wires a `remotelogger` — and it has no test at all. Two things follow from that.

**It is uncovered.** With cross-package coverage (`-coverpkg`) over the logging and middleware packages, `*logging.logger.LogEnabled` comes out at 100%, but this one is at **0.0%**. I replaced the whole body with `return false` — which in production silences every non-5xx request log — and `./pkg/gofr/logging/...`, `./pkg/gofr/http/...` all still passed. The middleware tests only exercise the hand-rolled `levelGateLogger` double, so nothing pins the real implementation. A small table over the six levels, plus a case asserting the wrapper flips after `ChangeLevel`, would close it.

**The lock doesn't guard what the comment says.** `r.mu` protects `currentLevel`, which this method never reads. `r.Logger` is assigned once in `New()` and never reassigned. And `ChangeLevel` is called *outside* `r.mu` in `checkAndUpdateLevel` (after `r.mu.Unlock()`), so the RLock gives no ordering against the level actually changing either — the comment's "stays correct even between a remote update and the level being applied" doesn't hold. What it does buy is an atomic reader-count bump on the framework's hot path, once per request, in a PR about removing per-request cost. Dropping it, or type-asserting the `logEnabler` once in `New()` and caching it, removes both the lock and the per-request interface assertion.

NitinKumar004 (changes requested):
The gate logic is correct and 5xx still logs unconditionally, but two findings undercut the PR's central perf claim and add hot-path cost — requesting changes to re-validate the optimization.

### Medium
- **The middleware gate is a no-op at the default `INFO` level.** The access log is emitted via `logger.Log(l)` → `logf(INFO)`, and `LogEnabled()` returns `INFO >= l.level`. With the default level `INFO`, `INFO >= INFO` is **true**, so the `RequestLog` struct, timestamp format, spanID string, and `getIPAddress` are all built and emitted exactly as before. The elision only fires at `NOTICE`+ (i.e. services that already disabled access logging). The PR/`BenchmarkLogging` comment says the log is "built at DEBUG" and dropped "at INFO or above" — that's factually wrong (`Log` is INFO). Net: a normal service at `LOG_LEVEL=INFO` gets **none** of the advertised middleware savings. (The ContextLogger allocation reductions are real at all levels — those are fine.)
- **New per-request `RWMutex.RLock` in `remoteLogger.LogEnabled` adds process-global atomic contention to the hot path.** In production the middleware logger is `*remoteLogger`, so every non-5xx request now takes `r.mu.RLock()`/`RUnlock()`. That single shared reader-count cache line hammered from every serving goroutine is a classic scalability bottleneck that can **raise tail latency** under high RPS — plausibly costing more than the one heap allocation removed (allocations come from per-P mcache, no cross-core traffic). And it guards nothing: `LogEnabled` doesn't read the mutex-protected level; it delegates to the embedded logger whose level read is itself unsynchronized. Please drop the `RLock` (or make it actually guard the level).

### Low
- `remoteLogger.LogEnabled`'s `RLock` also **races** with `ChangeLevel` (which writes `l.level` without `r.mu`) under `-race`; the lock guards the wrong field.
- **`BenchmarkLogging` never enters the fast path** — `MockLogger` has no `LogEnabled()`, so the `logEnabler` assertion fails and the full `RequestLog` is built every iteration. It measures the pre-optimization path; delete the whole gate and the numbers are identical. Benchmark against a `*logger`/`*remoteLogger` at `ERROR`, or give `MockLogger` a `LogEnabled()`.
- Real `LogEnabled` implementations (`*logger`, `*remoteLogger`) have **no direct tests**.
- New ContextLogger benchmarks use `context.Background()` (no valid span) + the pointer constructor `NewContextLogger`, so they exercise **neither** optimization and can't reproduce the `3→2 allocs`/`80→32 B` numbers. Point them at `benchSpanContext()` + `ContextLoggerFor`.
- `traceIDMarker` comment justifies keeping the map form via "`pkg/gofr/ai` emits it too" — `pkg/gofr/ai` has **zero** references to `__trace_id__`; after this PR no first-party code emits the map form.

### ⚠️ Cross-PR
#3975 has a hard compile dependency on `ContextLoggerFor` introduced here — land/revert this pair together, #3974 first.

**Requested changes:** re-evaluate/fix the INFO-level effectiveness and correct the "DEBUG" framing; drop the hot-path `RLock`; make `BenchmarkLogging` + the ContextLogger benchmarks exercise the actual fast paths; add `LogEnabled` tests; fix the `traceIDMarker` comment.

pkg/gofr/http/middleware/logger.go
aryanmehrotra: Done — it's a `const` now, with the derivation in the comment rather than in an init-time call:

```go
const canonicalCorrelationID = "X-Correlation-Id"
```

The `//nolint:gochecknoglobals` and the `net/textproto` import go with it. I kept the assertion but pointed it the other way, so it pins the equivalence rather than restating the literal:

```go
require.Equal(t, canonicalCorrelationID, textproto.CanonicalMIMEHeaderKey("X-Correlation-ID"))
```

That way the test still fails if the canonical spelling ever stops matching what's hard-coded.

pkg/gofr/http/middleware/logger.go
aryanmehrotra: Correct — `Error` goes through `logf(ERROR, …)`, so at `LOG_LEVEL=FATAL` it's dropped like anything else. Reworded to your phrasing:

```go
// A server error is reported through Error, which survives every level below
// FATAL, so gating it on the informational level would be wrong. Only the
// informational path can be skipped.
```

pkg/gofr/http/middleware/logger.go
aryanmehrotra: Reproduced this exactly, and the PR body is rewritten around it — the old description was wrong, not just imprecise.

Interleaved runs, benchstat `-count=10`, the same benchmark compiled against both sides (base = `origin/development`, `467b2ff`), logger output to `/dev/null` so nothing is measuring stdout:

| `LOG_LEVEL` | sec/op | B/op | allocs/op |
|---|---|---|---|
| unset → INFO | ~ (p=0.362) | 1579 → 1563 | 17 → **16** |
| ERROR | **−31.4%** (p=0.000) | 1441 → **1249** | 14 → **10** |

So at the default level it's the `X-Correlation-Id` write alone, as you measured. The description now says that in the section header rather than burying it, and the `BenchmarkLogging` doc comment says the same.

One note on the first row: an earlier run of mine reported INFO at +32% sec/op with ±90% variance. That was machine noise, not the atomic I added — interleaving the two sides brings it back to no significant difference. Allocations and bytes were identical across both runs.

pkg/gofr/http/middleware/logger_test.go
aryanmehrotra: Confirmed and fixed — `MockLogger` has no `LogEnabled`, so `logger.(logEnabler)` failed and the benchmark measured the pre-optimization path.

I did both things you offered rather than picking one:

- `BenchmarkLogging` now uses `logging.NewLogger(logging.ERROR)`, with a comment saying why it must not be a mock.
- `MockLogger` gains `LogEnabled()`. That one matters beyond this benchmark: any test that swaps a real logger for the mock was silently taking the build-unconditionally branch, which is the branch production doesn't take.

The doc comment is corrected too — the entry is built at INFO and *kept* at `LOG_LEVEL=INFO`; the fast path fires from NOTICE up.

I also added `TestRequestLogGateWithRealLoggers`, which runs `handleRequestLog` against both `*logging.logger` and `*remotelogger.remoteLogger` across INFO/NOTICE/WARN and asserts against real emitted output. It pins the default-level case as *still emitting*, so the narrower reach is now a test, not just prose.

pkg/gofr/logging/logger.go
aryanmehrotra: Agreed on the diagnosis, and agreed the API question is better settled before `LogEnabled()` has shipped widely — but I'd rather not settle it inside this PR.

Your own point is why: `logGRPCEntry` routes health-check / `/Send` / `/Recv` / `/SendAndClose` to `Debug` and everything else to `Info`, so the fix needs `Enabled(Level) bool` (or a `DebugEnabled()` sibling), and that's a different shape from what's here. Adding a second, wider optional interface in the same PR that introduces the first one would mean shipping the general form untested against its actual consumer.

What I've done instead is put it in the PR description under "Not addressed here", naming the file, the two call sites and why the literal port doesn't fit — so it's a visible follow-up rather than a thing we discover later. `LogEnabled` is optional and unexported-consumer-side in both packages that assert on it, so widening it to `Enabled(Level)` afterwards is additive, not breaking.

Happy to open that PR next if you'd rather it land as a pair.

pkg/gofr/logging/logger.go
aryanmehrotra: Both done.

**The unsynchronized level.** You're right that it predates this branch, but `LogEnabled` puts a second reader on the per-request path, so I closed it rather than leaned on it. `level` is now an `atomic.Int64` (Int64 rather than Int32 so the `Level` → int conversion isn't a narrowing one gosec flags). Evidence:

```
go test -race ./pkg/gofr/logging/remotelogger/
  origin/development : 13 DATA RACE,  3 citing logging/logger.go
  this branch        : 12 DATA RACE,  0 citing logging/logger.go
```

The remaining 12 are the pre-existing `httpLogFilter` ones, identical on both sides. The middleware benchmark is unchanged by the atomic (allocs and bytes byte-identical; sec/op no significant difference on interleaved runs).

The cost is six struct literals in tests that could no longer set `level` inline — they go through a `newLoggerAt(level, out, errOut)` helper now.

**The duplicated condition.** Shared:

```go
func (l *logger) enabled(level Level) bool { return level >= l.level.Load() ... }

func (l *logger) logf(level Level, ...) { if !l.enabled(level) { return } ... }
func (l *logger) LogEnabled() bool      { return l.enabled(INFO) }
```

I re-ran your mutation as a check: flipping `>=` to `>` in the one remaining predicate fails `TestRequestLogSchemaSnapshot`, same as before.

pkg/gofr/logging/logger.go
aryanmehrotra: Right — took your snippet as written:

```go
if m, ok := arg.(map[string]any); ok {
    if tid, exists := m[traceIDMarkerKey].(string); exists {
        if traceID == "" {
            traceID = tid
        }

        continue
    }
}
```

Both branches now strip whenever they match, which is what `hasTraceMarker` already assumes. Not reachable with today's producers, as you say, but it's the kind of thing that stops being unreachable when someone adds a third one.

pkg/gofr/logging/logger.go
aryanmehrotra: Fixed — it now names both shapes and who emits each:

```go
// hasTraceMarker reports whether any arg is a trace-ID marker -- either the
// typed traceIDMarker that ContextLogger emits, or the map[string]any carrying
// the "__trace_id__" key that pkg/gofr/ai emits. Read-only: it allocates
// nothing.
```

pkg/gofr/logging/remotelogger/dynamic_level_logger.go
aryanmehrotra: Both fixed, and I reproduced your mutation first.

**Coverage.** `return false` in place of the body left `./pkg/gofr/logging/...` and `./pkg/gofr/http/...` green, exactly as you found. There are now four tests: a table over the six levels, one asserting the wrapper follows `ChangeLevel`, one pinning that a logger without the optional interface fails open, and one asserting `LogEnabled` predicts whether `Log` actually writes. The same mutation now fails in the remotelogger package *and* in `http/middleware`, because `TestRequestLogGateWithRealLoggers` runs `handleRequestLog` against a real `remotelogger` rather than only the `levelGateLogger` double.

The fails-open case needed a `logging.Logger` that genuinely can't answer — embedding the *interface* rather than a concrete logger is what hides the method:

```go
type noGateLogger struct{ logging.Logger }
```

**The lock.** Dropped, and the assertion cached in `New`:

```go
func (r *remoteLogger) LogEnabled() bool {
    if r.enabler == nil {
        return true
    }

    return r.enabler.LogEnabled()
}
```

The doc comment now says why there's no lock — `r.mu` guards `currentLevel`, which this never reads, and `ChangeLevel` is called outside `r.mu` anyway — and that the gate is advisory: the worst a racing level change does is build one entry that's then discarded, or skip one at the instant the level is lowered. The underlying level read is properly synchronized now regardless, via the atomic (thread on `logger.go:140`).

aryanmehrotra:
Worked through both reviews. Nine of the ten findings held up and are fixed; one doesn't, and I've pushed back on it below. Inline threads have the per-item detail — this is the summary plus the one disagreement.

## Fixed

- **The gate is a no-op at the default level.** Confirmed, and the PR description was rewritten around it rather than patched. `Log` → `logf(INFO)`, `GetLevelFromString("")` → `INFO`, so the gate only fires from NOTICE up. Re-measured with benchstat over interleaved runs against `origin/development`: unset→INFO is `17 → 16` allocs and no significant sec/op change; ERROR is `14 → 10` allocs, `1441 → 1249 B`, `−31.4%` sec/op.
- **The hot-path `RLock`.** Dropped. It guarded `currentLevel`, which `LogEnabled` never reads, and `ChangeLevel` is called outside `r.mu` — so no ordering, just a reader-count bump per request. The optional-interface assertion moved to `New` as well, so the per-request interface check is gone too.
- **The `RLock` races with `ChangeLevel`.** Went further than dropping the lock and made `level` atomic, since this PR adds a second reader on the per-request path. `go test -race ./pkg/gofr/logging/remotelogger/`: **13 races on `development`, 3 citing `logger.go` → 12 on this branch, 0 citing `logger.go`.** The remaining 12 are the pre-existing `httpLogFilter` ones.
- **`BenchmarkLogging` never entered the fast path.** Confirmed — `MockLogger` has no `LogEnabled`. It now benchmarks a real `*logging.logger`, *and* `MockLogger` gained `LogEnabled` so other middleware tests stop silently taking the branch production doesn't take.
- **The ContextLogger benchmarks exercised neither optimization.** Confirmed — both new ones used `context.Background()` and the pointer constructor. Now on `benchSpanContext()` + `ContextLoggerFor`.
- **No direct tests for the real `LogEnabled`.** Both implementations now have tables over all six levels plus `ChangeLevel` and fails-open cases. I reproduced the `return false` mutation you described: it was silent before, and now fails in both `remotelogger` and `http/middleware`.

## Pushback: `pkg/gofr/ai` does emit the map form

> `traceIDMarker` comment justifies keeping the map form via "`pkg/gofr/ai` emits it too" — `pkg/gofr/ai` has **zero** references to `__trace_id__`; after this PR no first-party code emits the map form.

This isn't right — `pkg/gofr/ai/instrument.go` emits it on every instrumented LLM call:

```go
// pkg/gofr/ai/instrument.go:21
traceIDMarker = "__trace_id__"

// pkg/gofr/ai/instrument.go:179
args = append(args, map[string]any{traceIDMarker: sc.TraceID().String()})
```

The reference is there, just under a package-local constant rather than the literal, which is presumably why a grep for the string alone missed it. `pkg/gofr/ai/instrument_test.go:168` asserts on it too:

```
require.True(t, ok, "a __trace_id__ marker must be emitted so GoFr lifts it to a top-level trace_id")
```

`pkg/gofr/ai` can't use the typed `traceIDMarker` because it's unexported in `pkg/gofr/logging`, so the map form is a live cross-package contract and both branches of `extractTraceIDAndFilterArgs` and `hasTraceMarker` have to stay. Rather than only leaving the comment alone, I made it cite the file so the next reader doesn't have to re-derive this:

```go
// The map form remains accepted because it is a cross-package contract:
// pkg/gofr/ai/instrument.go emits map[string]any{"__trace_id__": traceID} on
// every instrumented LLM call, and it cannot use this type because traceIDMarker
// is unexported.
```

The map branch did have a real bug, though — the one @arunesh-j found: it stripped only while no trace ID had been seen, so a second map marker rendered into the message as a literal map. That's fixed.

## Deferred, with reasons

**The gRPC side** (`pkg/gofr/grpc/log.go`) has the same unconditional-build shape. I agree the API is better settled before `LogEnabled()` spreads, but not in this PR: `logGRPCEntry` routes health-check and stream traffic to `Debug` and the rest to `Info`, so it needs `Enabled(Level) bool`, and introducing the general form here would mean shipping it untested against its actual consumer. It's in the description under "Not addressed here" with the call sites named. Widening `LogEnabled` to `Enabled(Level)` later is additive, so nothing here forecloses it — happy to open that PR next if you'd rather they land together.

## Merge order

Confirmed: **#3975 will not compile without `ContextLoggerFor` from this branch.** It's now stated in this PR's description as well as #3975's. This one first; land or revert the pair together.

pkg/gofr/http/middleware/logger.go
aryanmehrotra: Done — `const canonicalCorrelationID = "X-Correlation-Id"`, with the `textproto.CanonicalMIMEHeaderKey` call in a comment for provenance. The package-level `var`, the `//nolint:gochecknoglobals` and the `textproto` import all went with it, as you noted they would.

pkg/gofr/http/middleware/logger.go
aryanmehrotra: Fixed — the comment now reads "…which survives every level below FATAL, so gating it on the informational level would be wrong."

You were right that the original overshot: `Error` goes through `logf(ERROR, …)`, so at `LOG_LEVEL=FATAL` a 5xx request log is discarded like any other. The code was already correct — skipping the gate for 5xx fails open, which is the safe direction — only the stated reason was too strong.

pkg/gofr/http/middleware/logger.go
aryanmehrotra: Done, and the PR description was rewritten around it rather than having a caveat bolted on. It now leads with the distinction: the `ContextLogger` saving applies at every level, while the middleware gate only fires at `NOTICE` and above, so a default-configured service sees the correlation-ID change alone.

Your measured table is in the body, including the `unset → INFO` row, so the number cannot be read as a universal per-request win.

pkg/gofr/http/middleware/logger_test.go
aryanmehrotra: Fixed — `BenchmarkLogging` now builds `logging.NewLogger(logging.ERROR)` instead of `NewMockLogger`, with a comment recording exactly why: `*MockLogger` has no `LogEnabled()`, so the `logEnabler` assertion failed and the benchmark measured the pre-optimization path while appearing to measure the gate.

You were measuring a ~4x understatement and that matches what I see. The doc comment claiming the entry is "built at DEBUG" is corrected too — `Log()` calls `logf(INFO, …)`.

pkg/gofr/logging/logger.go
aryanmehrotra: **Deliberately not done here** — this is the one item from your review I have left open, and I'd like your read on the plan rather than quietly deferring it.

You're right on both counts: `logRPC` and `logStreamRPC` build the entry unconditionally, and a literal port does not fit because `logGRPCEntry` routes health-check and stream traffic to `Debug` while `LogEnabled() bool` only answers for `INFO` — gating just the `Info` branch would leave the highest-volume traffic unfixed.

My reasoning for keeping it separate: widening the API is the whole decision, and it deserves its own PR where the shape can be argued on its merits — `Enabled(Level) bool` versus a sibling `DebugEnabled()`. Landing a half-measure here, or widening the interface in a PR about HTTP request logs, both seem worse than a focused follow-up.

The PR description records this as out of scope with your reasoning. Happy to open the follow-up now if you'd rather see them together.

pkg/gofr/logging/logger.go
aryanmehrotra: Both fixed.

`l.level` is an `atomic.Int64` now, so `ChangeLevel` and every read are properly synchronized rather than the race being merely pre-existing. `-race ./pkg/gofr/logging/remotelogger` drops from 3 races citing `logger.go` to 0 — the remainder are the pre-existing `httpLogFilter` races, identical on `development`.

The duplicated gate condition is gone too: `logf` and `LogEnabled` both go through one `enabled(Level)` helper, so they cannot drift in the direction where request logs get dropped that `logf` would have emitted.

pkg/gofr/logging/logger.go
aryanmehrotra: Fixed exactly as you suggested — the `continue` is unconditional on `exists`, so a map-shaped marker arriving after a typed one is stripped rather than falling through and rendering into the message as a literal map.

You were right that it is unreachable with today's two producers, but it was already reachable for two map markers, and `hasTraceMarker` treats both shapes as markers to strip — so the asymmetry was a real inconsistency regardless.

pkg/gofr/logging/logger.go
aryanmehrotra: Fixed — the doc now says it reports whether any arg is a trace-ID marker in *either* shape: the typed `traceIDMarker` that `ContextLogger` emits, or the `map[string]any` carrying `__trace_id__`.

pkg/gofr/logging/remotelogger/dynamic_level_logger.go
aryanmehrotra: Both fixed.

The `RLock` is gone. Your analysis was exact: `r.mu` guards `currentLevel`, which the method never reads, and `ChangeLevel` is called *outside* `r.mu` in `checkAndUpdateLevel`, so it bought no ordering while adding a process-wide reader-count bump per request. The `logEnabler` assertion is resolved once in `New` and cached, so the per-request interface assertion is gone too.

Coverage: four tests now, including the two you described — a table over the levels (`TestRemoteLoggerLogEnabledMatchesLevel`) and one asserting the wrapper follows `ChangeLevel` (`TestRemoteLoggerLogEnabledFollowsChangeLevel`), plus `…FailsOpen` for the nil-enabler path and `…AgreesWithLog` pinning it against what `Log` actually emits. Your mutation — replacing the body with `return false` — now fails.

aryanmehrotra:
## Ready for another look — and this PR now contains what was #3975

Head is `c28d32d1`. Two things changed since the last review round: **#3975 was folded in**, and everything below was re-verified against current `development` (`7cbb41a3`, i.e. after #3770 landed).

### Why the two were combined

The dependency was not just ordering, it was an API one. This PR introduced `logging.ContextLoggerFor`, and **#3975 was its only production caller** — `pkg/gofr/context.go` here still went through `*NewContextLogger(...)`. Proven by rebasing #3975's own commits onto bare `development`:

```
pkg/gofr/context.go:205:26: undefined: logging.ContextLoggerFor
```

Split, this PR would have shipped an **exported function with no caller in the tree**, which on a v1 line is permanent — removing it later is a breaking change. #3975 is closed as merged into this branch; its commits are here unchanged and its review threads stay readable there. @arunesh-j — all three of your findings on it (the `WSHandlerUpgrade` drift in the bench router, the `RealHandlerPath` comment describing the wrong thing, the `WithContext` chain discarding `t.Context()`) are in this branch as-is.

---

### No behaviour change — measured, not asserted

Two real GoFr servers, one per tree, same handlers, differenced automatically:

```
348 request cases    0 CHANGED    348 IDENTICAL
```

4 Bind routes × 10 content types × 3 bodies, plus 19 routes × 3 origins × GET/OPTIONS/DELETE/HEAD — covering `Response`, `*Response`, `Raw`, `Raw`+error, `Redirect`, `*Redirect`, `File`, `XML`, typed errors, path/query params, `/panic`, both `.well-known` probes and an unregistered route. Status, every header and exact body bytes compared.

**Log output, every level:**

```
UNSET 26 · DEBUG 30 · INFO 26 · NOTICE 12 · WARN 10 · ERROR 8 · FATAL 1   →  0 of 7 differ
```

Timestamps, trace/span IDs, response times and addresses normalized; nothing else. The panic path's stack frames are byte-identical too.

**WebSocket** — the retention path called out on #3975, where `App.WebSocket` assigns `ctx.Request = conn` and loops for the life of the socket:

```
development  upgrade=101 | m0->echo:m0 | … | m4->echo:m4
this branch  upgrade=101 | m0->echo:m0 | … | m4->echo:m4
```

**Whole repository suite:** 50 ok / 6 FAIL on both trees, **no package differs in pass/fail**, same three named failing tests. All six are environmental here — Redis/MySQL/Kafka absent, metrics port held.

### What it is worth

`BenchmarkRequest_RealHandlerPath`, 5 runs each, every run identical:

| | B/op | allocs/op |
|---|---|---|
| `development` | 3915 | 49 |
| **this PR** | **3691** | **41** |

Under `-race`, `./pkg/gofr/logging/remotelogger` drops from **3 races citing `logger.go` to 0**; the remainder are the pre-existing `httpLogFilter` races, present identically on `development`.

### On the break-even for discarded log lines

@arunesh-j — worth naming the metric, since the two crossovers differ. Measured directly:

| discarded lines/request | 0 | 1 | 2 | 3 | 8 | 9 | 10 |
|---|---|---|---|---|---|---|---|
| base B/op | 432 | 480 | 528 | 576 | 816 | 864 | 912 |
| head B/op | **0** | **96** | **192** | **288** | **768** | 864 | 960 |

**Allocation count** breaks even at **3** lines; **bytes** not until **9**. The description quotes the earlier of the two.

### Cascade

Merges into `development` clean. Every other open PR of mine still merges clean alongside it, and all six merged together build, pass `-race`, and serve the same 348 cases identically. Stacked, the six come to **24 allocs / 2647 B** on that same benchmark — down from 49 / 3915.

`golangci-lint --new-from-rev=origin/development`: **0 issues**. CI green. 0 behind `development`.

---

Eight threads resolved — every one fixed in code, each with the fix described in its own reply. **The gRPC thread is left open on purpose**: `pkg/gofr/grpc/log.go` has the same unguarded shape, but `logGRPCEntry` routes health-check and stream traffic to `Debug` while `LogEnabled()` only answers for INFO, so a literal port would leave the highest-volume traffic unfixed. That needs a level-parameterized API, which is a decision worth its own PR. @arunesh-j, I'd still like your read on that before I open it.

@NitinKumar004 — your review predates all of the above; re-requesting.

Umang01-hash (approved):
Nice PR. Checked the perf claims rather than take them on trust — escape analysis shows the co-allocated `httpContext` as a single object and `BenchmarkContext_New` lands on 1 alloc/op, and the deferred trace-ID formatting holds. The skip gate is the part I looked at hardest since it's easy to drop logs here, but sharing one `enabled()` predicate between `logf` and `LogEnabled` keeps them honest, and the 5xx-still-logged-at-ERROR case is covered with a real remotelogger. Making `level` atomic is a good catch too. No breaking API.

Two small things, neither blocking:
- `extractTraceIDAndFilterArgs` now drops every `__trace_id__` marker map; the old code kept a second one if present. Harmless today (nothing emits two), but it's an untested change — maybe a one-line test or a note.
- Heads up: `go test -race ./pkg/gofr/logging/remotelogger` still fails `TestRemoteLogger_UpdateLevel` with a data race — it's pre-existing (same at the merge-base), and this PR actually shrinks the level-field race, just flagging it as a possible follow-up.

PR #3975, opened 2026-08-18
perf(http): give the request Context, Request and Responder one allocation

aryanmehrotra:
> ⚠️ **Stacked on #3974** (`perf/logging-allocs`). Base is that branch, not `development`. Merge #3974 first — this needs `logging.ContextLoggerFor` to compile. **Review only the top 3 commits.**

**TL;DR — 3 fewer allocations per request. No behaviour change.**

## The problem

**Two separate wastes on every single request:**

**1. Three heap objects where one would do.** The `Context`, the `Request` wrapper it holds through an interface, and the `Responder` it holds through another. All three are created together, live for exactly one request, and are discarded together.

**2. A trace ID nobody reads.** `handler.ServeHTTP` did this on every request:

```go
traceID := trace.SpanFromContext(r.Context()).SpanContext().TraceID().String()
```

`TraceID().String()` allocates a **32-character string**, and its only consumer is `logError` — which runs **only when a handler returns an error**.

## The fix

- `newHTTPContext` places the concrete `Request` and `Responder` **inside the same struct** as the `Context` and points the interfaces at them.
- The `SpanContext` is a value type, so carrying it costs nothing. The string is now built **where it is used**, on the error path.

## The numbers

| | allocs/op | B/op |
|---|---|---|
| before | 44 | 3723 |
| **after** | **41** | **3692** |

> **Re-measured after review.** These read 43 → 40 before. The delta is unchanged at **−3 allocations**; the absolute numbers each moved by one because the benchmark's middleware chain was missing `WSHandlerUpgrade`. See below. Measured with `-benchtime=3000x -count=6` on both branches, identical benchmark source on each; every run reported the same allocation count.

## This PR also fixes a measurement blind spot

GoFr had **no benchmark covering its own handler path**. The existing whole-request benchmarks register plain `http.HandlerFunc` values on the router, so they never enter `handler{}` — the Context, Request and Responder are never built and the trace ID is never resolved.

Work on exactly that path measured as *"no change"*, which is how this series nearly got discarded as pointless. `BenchmarkRequest_RealHandlerPath` closes that gap.

### Fixed after review

- **The bench router no longer rebuilds the middleware chain by hand.** It dropped `WSHandlerUpgrade` while claiming to mirror what a GoFr application serves — and `buildBenchRouter` carries a "keep this in sync with `newHTTPServer`" note precisely because that drift is easy. The chain is defined once now, in `useGoFrMiddleware`, and both bench routers go through it.
- **`BenchmarkRequest_RealHandlerPath`'s comment was wrong.** It claimed the trace ID is resolved on the measured path. It is not: after this PR `TraceID().String()` runs only inside `logError`, the benchmark handler always returns `nil`, and the mock config installs no real `TracerProvider`. The alloc delta was valid; the comment described the wrong thing.
- **The trace-ID guard test chained `WithContext` off `context.Background()`**, replacing the context `NewRequestWithContext` had just installed — so `t.Context()` was discarded and the request was no longer cancelled at test end.
- **The `httpContext` doc now names the framework's own retention path.** It described only the handler case, which is a user bug. `App.WebSocket` has the same shape and is our code: it assigns `ctx.Request = conn` and loops for the life of the socket, so the underlying `*http.Request` that used to become unreachable is now retained for the whole connection. Small per connection, but it scales with concurrent long-lived sockets.

## Safety

- **Nothing observable changes.** `Context` still exposes both through the same interfaces, `ctx.Request` still resolves path and query params, the responder still writes through to the same `ResponseWriter`.
- `RequestFor` / `ResponderFor` return **values** so a caller that already owns an allocation need not take another. `NewRequest` / `NewResponder` are unchanged and delegate to them — **purely additive**.
- CMD, GraphQL, cron and subscribers supply their own `Request` implementations and keep using `newContext`, untouched.
- 3 new tests, including one pinning that an erroring handler still logs its trace ID.

## Measurement

Allocations and bytes only. Wall-clock on the machine used varied by several times between runs of the same binary.


aryanmehrotra:
## ⚠️ CI does not run on this PR — read before merging

The pipeline only triggers on PRs targeting `main` or `development`:

```yaml
pull_request:
  branches:
    - main
    - development
```

This PR targets `perf/logging-allocs`, so it gets **only** `snyk` and `typos`. **No unit tests, no lint, no CodeQL.** Three green checks here do not mean what three green checks mean on the other PRs.

### What to do

Once **#3974 merges**, retarget this PR to `development` and the full pipeline runs. Do not merge it on the strength of the checks currently shown.

### Verified locally in the meantime

Run on this exact head with the same tool versions CI pins:

| check | result |
|---|---|
| `go build ./...` (all modules) | ok |
| `go vet ./pkg/gofr/...` | ok |
| `golangci-lint v2.12.2` (full module, vs base) | **0 issues** |
| `typos` (CI's pinned binary) | ok |
| `pkg/gofr/http/...`, `logging/...`, `metrics/...`, `container/...` | all ok |
| `pkg/gofr` | 1 failure — **pre-existing**, unrelated |

That one failure is `TestEnableMCP_ServerConfigured`, which asserts the default MCP port 8200 and dies via `Logger.Fatalf` → `os.Exit(1)` when anything else holds that port. It fails identically on clean `development` and is fixed by #3942. CI never sees it because runners have 8200 free.

PiyushSingh-ZS:
Reviewed on top of `perf/logging-allocs` (its base), built and tested. This one is clean — I went looking for the two ways it could break and neither applies:

- **No type assertions on the concrete types.** I grepped `pkg/` for assertions against `*gofrHTTP.Request` / `*gofrHTTP.Responder`; there are none, so moving to `&a.req` / `&a.resp` inside `httpContext` is invisible to the rest of the framework. `Respond` has a value receiver, so both the value and pointer forms satisfy `Responder` either way.
- **The lazy trace ID is identical.** It's the same expression on the same `SpanContext`, just evaluated on the error path instead of eagerly, and `spanCtx` is copied by value into the handler goroutine in `serveWithGoroutine`, so there's no sharing across goroutines.

Keeping `NewRequest` / `NewResponder` as pointer-returning wrappers over the new `RequestFor` / `ResponderFor` is the right way to do this without touching the exported surface.

### The lifetime coupling is worth keeping prominent

The doc comment already calls it out:

> a handler that retains any one of them past the request keeps the other two — including the ResponseWriter — alive with it

That's the one real cost, and I think it's acceptable exactly as argued: retaining any of them was already a bug under net/http's rules. Worth making sure this lands *after* the pooling PRs in the stack rather than interleaved, since a retained `Context` now also pins whatever the pooled machinery hands it.

### Small suggestion

`newHTTPContext` is the only construction site that gets the single allocation; CMD, GraphQL, cron and subscribers still go through `newContext`. The comment says so, which is good — but it might be worth a line on *why* they can't (they supply their own `Request` implementations), so nobody later "fixes" the inconsistency by force-fitting them onto `httpContext`.

pkg/gofr/context.go
```diff
@@ -160,6 +162,43 @@ func (a *authInfo) GetAPIKey() string {
...
+// together, so giving them one allocation instead of three changes nothing
+// observable -- Context still holds them through the same interfaces.
+//
+// One consequence is worth knowing: sharing an allocation means they are now
+// reachable as a unit, so a handler that retains any one of them past the
+// request keeps the other two -- including the ResponseWriter -- alive with it.
+// Retaining any of them after the handler returns was already a bug, since
+// net/http forbids using the ResponseWriter then, but the cost of doing it is
+// larger now.
+type httpContext struct {
```
arunesh-j, line 185: The retention note here is accurate but generic, and there is one in-framework case that now hits it: `App.WebSocket` does `ctx.Request = conn` (`pkg/gofr/websocket.go:39`) and then loops in `handleWebSocketConnection` for as long as the socket lives.

Before this change, reassigning that interface field made the `*gofrHTTP.Request` unreachable, so the underlying `*http.Request` (headers, URL, `pathParams` map) could be collected while the connection stayed open. Now it shares an allocation with the still-live `Context`, so it is retained for the connection's entire lifetime. Small per connection, but it scales with concurrent long-lived sockets — and unlike the handler case the doc describes, this one is not a user bug, it is our own code.

Worth naming explicitly in this comment so the next person reading it knows the framework itself has a path that retains the request.

pkg/gofr/handler_test.go
```diff
@@ -564,3 +567,71 @@ func TestIntegration_ServerTimeout(t *testing.T) {
...
+
+var errTraceIDGuard = errors.New("trace id guard")
+
+// buildRealPathRouter mirrors what a GoFr application actually serves: the handler is wrapped in
+// gofr's own handler{}, so newHTTPContext builds the Context/Request/Responder, the trace ID is
+// resolved, and the result goes back through responder.Respond as a JSON envelope.
+//
+// The existing BenchmarkRequest_FullChain family registers raw http.HandlerFunc values straight on
+// the router, so none of that code runs there — it measures the middleware chain only.
+func buildRealPathRouter(c *container.Container) http.Handler {
```
arunesh-j, line 612: This duplicates `buildBenchRouter` (`pkg/gofr/http_server_test.go:363`) but drops `middleware.WSHandlerUpgrade`, while the comment claims it "mirrors what a GoFr application actually serves". `buildBenchRouter` carries a "keep this in sync with `newHTTPServer`" note precisely because that drift is easy — a second, already-divergent copy defeats it.

Adding a `/real/{id}` route wrapped in `handler{}` to the existing `buildBenchRouter` would give this benchmark the same coverage with one chain definition to maintain.

pkg/gofr/handler_test.go
```diff
@@ -564,3 +567,71 @@ func TestIntegration_ServerTimeout(t *testing.T) {
...
+
+	logs := testutil.StderrOutputForFunc(func() {
+		c := container.NewContainer(config.NewMockConfig(map[string]string{"LOG_LEVEL": "ERROR"}))
+		h := handler{
+			function:  func(*Context) (any, error) { return nil, errTraceIDGuard },
+			container: c,
+		}
+
+		req := httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/x", http.NoBody).
+			WithContext(trace.ContextWithSpanContext(context.Background(), sc))
```
arunesh-j: The chained `WithContext` throws away the `t.Context()` passed on the line above, so the `NewRequestWithContext` call is doing nothing and the request context is no longer cancelled at test end.

```go
req := httptest.NewRequest(http.MethodGet, "/x", http.NoBody).
	WithContext(trace.ContextWithSpanContext(t.Context(), sc))
```

That keeps the span context and the test's cancellation.

arunesh-j (commented):
Reviewed the whole diff and traced every path the refactor touches (`handler.go`, `context.go`, `websocket.go`, `http/request.go`, `http/responder.go`). **No correctness issues** — `newHTTPContext` produces exactly what `newContext(NewResponder(w, r.Method), NewRequest(r), c)` did, including `mux.Vars` timing and `ContextLoggerFor`; `Request`/`Responder` hold no locks or self-referential state so returning them by value is copy-safe; and the trace-ID lazification is value-equivalent (with no span, `SpanContext{}.TraceID().String()` still yields the all-zero 32-char string, so the `omitempty` log shape is unchanged). Verified with `go build`, `go vet`, `golangci-lint`, the package suite under `-race`, and by running the new benchmark — 40 allocs/op, matching the claim.

Three low-severity notes inline; none block merging.

NitinKumar004 (changes requested):
Nice reduction (43→40 allocs) and the field-parity tests are solid. Minor items + a merge-order dependency.

### Low
- **Shared allocation widens GC retention blast radius.** With `Context`/`Request`/`Responder` in one struct, a handler that retains any one of them past the request now also pins the `ResponseWriter` and whole `Context`. Retaining request objects past their lifetime is already a net/http violation, so this only makes an existing misuse costlier — documented in the `httpContext` doc comment. Awareness only.
- **`BenchmarkRequest_RealHandlerPath` comment is inaccurate.** It claims "the trace ID is resolved" on the measured path, but after this PR's own change `TraceID().String()` runs only inside `logError` (error path), the benchmark handler always returns `nil`, and the mock config installs no real `TracerProvider` (invalid SpanContext) — so no trace ID is formatted. The alloc delta it reports is still valid; the comment just mis-describes what's exercised.

### ⚠️ Cross-PR — build dependency on #3974
`newHTTPContext` calls `logging.ContextLoggerFor`, which is introduced by #3974. This PR **will not compile** against `development` on its own — merge #3974 first, and land/revert the two as a pair.

**Requested changes:** correct the benchmark comment; note/enforce the #3974 merge-order dependency in the PR description.

pkg/gofr/context.go
aryanmehrotra: Named explicitly, thank you — this was the right catch. The `httpContext` doc now says:

> The framework itself has one such path, and it is not a user bug: `App.WebSocket` assigns `ctx.Request = conn` and then loops in `handleWebSocketConnection` for as long as the socket lives. Reassigning that interface field used to make the `*gofrHTTP.Request` unreachable, so the underlying `*http.Request` — headers, URL, the `pathParams` map — could be collected while the connection stayed open. It now shares an allocation with the still-live `Context`, so it is retained for the connection's whole lifetime. Small per connection, but it scales with the number of concurrent long-lived sockets.

Your distinction is the part that mattered: the case the original comment described is a `net/http` violation by the user, whereas this one is our own code, and someone reading the generic note would not have known that.

pkg/gofr/handler_test.go
aryanmehrotra: Fixed, and slightly differently from your suggestion — I'd like to flag why in case you prefer the original.

Rather than adding a `/real/{id}` route to `buildBenchRouter`, the middleware chain is extracted into `useGoFrMiddleware` and both bench routers call it. Same outcome — one chain definition, and `buildRealPathRouter` now includes `WSHandlerUpgrade` as it always should have — but it avoids adding a route to the router `BenchmarkRequest_FullChain` measures, which would lengthen `RegisteredRoutes` and shift the shared baseline every perf PR diffs against.

The "keep this in sync with `newHTTPServer`" note now lives on the single definition.

Consequence worth recording: including `WSHandlerUpgrade` moved the measured numbers, so I re-measured both branches with identical benchmark source. **44 → 41 allocations** (was reported as 43 → 40). The −3 delta is unchanged; the absolute figures each moved by one because the chain was previously incomplete.

pkg/gofr/handler_test.go
aryanmehrotra: Fixed — it chains off `t.Context()` now, exactly as you wrote it, so the span context is kept *and* the request is still cancelled at test end. Comment added so the next person does not re-background it.

aryanmehrotra:
**Folded into #3974 — closing this in favour of the combined PR.**

Every commit here is in `perf/logging-allocs` unchanged; the two branches now point at the same object (`c28d32d1a`), so this PR's diff against its base is empty. Nothing is being dropped.

### Why combine rather than merge in sequence

The dependency was not just ordering, it was an API one. #3974 introduced `logging.ContextLoggerFor` and **this** PR was its only production caller — `pkg/gofr/context.go` on #3974 alone still went through `*NewContextLogger(...)`. Proven by rebasing this branch's own commits onto bare `development`:

```
pkg/gofr/context.go:205:26: undefined: logging.ContextLoggerFor
```

Split, #3974 would have shipped an **exported function with no caller in the tree**. On a v1 line that is permanent: removing it later is a breaking change. Merged together, the export has a consumer the moment it exists.

### What the combined PR reports

Because the two halves compound on the same path, the honest number is the pair measured together — `BenchmarkRequest_RealHandlerPath` against current `development`:

| | allocs/op | B/op |
|---|---|---|
| `development` | 49 | 3915 |
| #3974 (combined) | **41** | **3690** |

The `44 → 41` figure in this PR's description was correct but measured against #3974 as its base, so it showed only this half.

### Your review, @arunesh-j

All three of your findings here are in the combined branch and unchanged — the `WSHandlerUpgrade` drift in the bench router, the `BenchmarkRequest_RealHandlerPath` comment that described the wrong thing, and the `WithContext` chain that discarded `t.Context()`. The threads stay readable on this PR; the code is on #3974.

Continuing review on **#3974**.

PR #3977, opened 2026-08-20
docs(cron): correct the minimum-interval claim and the "every 5 hours" example (#3876)

### LINKED ISSUE #3876: docs(cron): using-cron page says sub-minute scheduling is impossible, then demonstrates it
**Describe the bug**

`docs/advanced-guide/using-cron/page.md` contradicts itself about whether cron jobs can run more than once a minute, and the sentence that is wrong is the one a reader is most likely to act on.

`:39` states:

> The minimum time difference between cron job's two consecutive runs is a minute as it is the least significant scheduling time parameter.

Ten lines above, `:29-31` documents the optional leading `second` field:

> GoFr also allows an optional field for `second` as first part in the schedule format, like in the following format:
> `second minute hour day_of_month month day_of_week`

and the page's own example at `:77` uses it:

```go
// Run the cron job every 10 seconds(*/10)
app.AddCronJob("*/10 * * * * *", "", func(ctx *gofr.Context) {
    ctx.Logger.Infof("current time is %v", time.Now())
})
```

So the page tells you sub-minute scheduling is impossible, then demonstrates it.

**Which one is correct**

The `second` field is real; `:39` is the stale sentence. In `pkg/gofr/cron.go`:

- `NewCron` ticks once per second — `ticker: time.NewTicker(time.Second)` (`cron.go:73`) — so the scheduler is evaluated at second granularity, not minute.
- Both schedule lengths are supported constants: `scheduleParts = 5` and `schedulePartsWithSecond = 6` (`cron.go:22-23`).
- `job` carries a `sec` field alongside `min`/`hour`/… (`cron.go:51`), and `seconds = 59` bounds it (`cron.go:17`).

`:39` looks like it predates the `second` field and was never removed when that field was added.

**Expected behavior**

`:39` should say the minimum gap is one **second** when the optional leading `second` field is used, and one minute with the 5-field form (where `second` defaults such that the job fires once per minute). Suggested replacement:

> With the 5-field format the minimum time difference between a cron job's two consecutive runs is a minute, since minute is then the least significant scheduling parameter. Adding the optional leading `second` field lowers that to a second — GoFr's scheduler ticks every second.

**More description**

Pre-existing and unrelated to any in-flight change; surfaced during review of #3815, which touched `docs/guides/graceful-shutdown/page.md` and deliberately left this out of scope. Docs-only — no code change is implied, since the code already behaves as the `second` field documents.

akshat-kumar-singhal:
**Description:**

Fixes #3876. Documentation only — no code change, since the scheduler already behaves the way the `second` field documents.

`docs/advanced-guide/using-cron/page.md` contradicted itself. Line 39 said:

> The minimum time difference between cron job's two consecutive runs is a minute as it is the least significant scheduling time parameter.

Ten lines above, the page documents the optional leading `second` field, and its own example at line 77 uses `*/10 * * * * *`. So the page told you sub-minute scheduling was impossible, then demonstrated it. The prose is the stale half — it predates the `second` field.

**Two changes:**

1. **The interval sentence (`:38-40`).** Now states both cases, and names the actual mechanism behind the 5-field floor. That floor is real, but not for the reason given: seconds *are* schedulable, and the minute gap comes from `job.tick` pinning the second to 0 when `sec` is nil (`cron_scheduler.go:213-216`). The new wording says so, rather than repeating a justification that stopped being true when `schedulePartsWithSecond` landed.

2. **The `*/5` example (`:72`).** Found while sweeping the page for other contradictions, per the issue's suggestion. `app.AddCronJob("* */5 * * *", ...)` is commented "every 5 hours" and isn't — it's minute=`*`, hour=`*/5`, so it fires *every minute* during hours 0/5/10/15/20. Changed to `"0 */5 * * *"`, with ", at minute 0" added to the comment so the leading `0` is explained. This is kept in the same PR deliberately: change 1 now tells the reader a 5-field job fires once per minute, and an example on the same page contradicting that would reintroduce the very defect being fixed.

**Verification:**

The `*/5` claim was checked against the scheduler rather than by reading field positions. Walking a simulated day through `job.tick` counts:

| schedule | matches/day |
| --- | --- |
| `* */5 * * *` | 300 |
| `0 */5 * * *` | 5 |
| `* * * * *` | 1440 |
| `* * * * * *` | 86400 |
| `*/10 * * * * *` | 8640 |

This confirms both the corrected example and the minute/second floors the new sentence claims.

None. Prose and one example comment/expression; no exported behaviour touched.

**Additional Information:**

No new test. `TestJob_tick_SecNilAndNonZeroSec` (`cron_test.go:882`) already pins the nil-`sec` behaviour the corrected wording describes — the defect was docs drifting from code, not a coverage gap, so a second test on the same invariant would be redundant.

Credit where due: @adarshjnv9100 commented on #3876 with a correct plan and asked to be assigned. I had this written up before circling back to the thread — happy to close this in favour of their PR if the maintainers prefer.

**Checklist:**


aryanmehrotra (approved):
Checked against the scheduler rather than by reading field positions, and both corrections hold.

**The contradiction was real.** The page documents the optional leading `second` field at :29-31 and
demonstrates it at :77 with `*/10 * * * * *`, while the prose at :38 said sub-minute scheduling was
impossible. Good catch that the prose is the stale half.

**The new explanation is mechanically right.** `cron_scheduler.go` does:

```go
if j.sec != nil {
    if _, ok := j.sec[t.sec]; !ok { return false }
} else if t.sec != 0 {
    return false
}
```

so the one-minute floor on a 5-field schedule comes from the second being pinned to 0, which is what
your wording now says — rather than the old justification that stopped being true when
`schedulePartsWithSecond` landed.

**The `*/5` example was wrong.** `* */5 * * *` is minute=`*`, hour=`*/5`, so it fires every minute
during hours 0/5/10/15/20 — 300 times a day, not 5. `0 */5 * * *` is correct and ", at minute 0"
earns its place explaining the leading zero. Your verification table checks out: 300 = 60x5,
1440 = 60x24, 86400 = 60x60x24, 8640 = 86400/10.

Keeping both changes in one PR was the right call — correcting the interval sentence while leaving an
example on the same page that contradicts it would have reintroduced the defect being fixed.

Approving.

Umang01-hash (approved):
Checked this against the cron code and it's right. `* */5 * * *` really does fire every minute during hours 0,5,10… so the old "every 5 hours" comment was wrong — `0 */5 * * *` is the correct one. The seconds-field note also matches `tick()`: 5 fields run once a minute at second 0, adding the 6th runs per second. LGTM.

Unrelated nit if you're in the file: the `errBadScheduleFormat` message still says "five components" even though 6 are supported now.

Issue #3979, opened 2026-08-21
RBAC: per-request pattern matching races and grows the router without bound; multi-role JWTs authorize nothing


aryanmehrotra:
Three defects in the RBAC middleware, all present on `development` and in released versions. None
were introduced by #3934 — they surfaced while reviewing it, and are filed together because the
first two share a root cause and a single fix.

## 1. Per-request pattern matching races, and grows the router without bound

`pkg/gofr/rbac/endpoint_matcher.go:91` builds a route per rule, per request, on the config's shared
router:

```go
route := router.NewRoute().Path(pattern)
```

gorilla/mux `NewRoute` appends to the router it is called on (`mux@v1.8.1/mux.go:279-284`):

```go
func (r *Router) NewRoute() *Route {
	route := &Route{routeConf: copyRouteConf(r.routeConf), namedRoutes: r.namedRoutes}
	r.routes = append(r.routes, route)
	return route
}
```

`config.muxRouter` is created once at load and shared across every request, so that append is both
unsynchronised and permanent.

**Unbounded growth.** Walking the router after driving requests through `rbac.Middleware`:

```
0 routes at start  ->  114 routes after 100 requests
```

It never shrinks. A long-lived pod accumulates routes for the whole life of the process.

**Data race.** 50 goroutines x 20 requests through the middleware, under `-race`:

```
WARNING: DATA RACE
  mux.(*Router).NewRoute()          mux.go:282
  rbac.matchMuxPattern()            endpoint_matcher.go:91
  rbac.matchesKey()                 config.go:484
  rbac.findEndpointByPattern()      config.go:439
  rbac.getEndpointForRequest()      endpoint_matcher.go:210
```

The existing suites do not catch this because they drive requests sequentially; it only appears
under concurrency, which is the only way a server runs it.

**Reproduction:** any RBAC config containing a mux pattern (`{id}`, `{path:.*}`), with concurrent
requests to a path that reaches the pattern scan.

## 2. Authorizing one request costs hundreds to thousands of allocations

Resolving a single pattern path on `development`, `-benchmem`:

| rules in config | ns/op | allocs/op |
| --- | --- | --- |
| 6 | 18,165 | 533 |
| 21 | 58,471 | 1,648 |
| 51 | 127,909 | 3,734 |

This is the cost of recompiling mux patterns per rule per request. RBAC middleware runs on every
request, so this is on the hot path of every route in an application that enables it.

### Both of the above are fixed by the same change

[...]

Issue #3935, opened 2026-08-15
RBAC keeps a second copy of the route table; the router has already resolved the request by the time the middleware runs


akshat-kumar-singhal:
### Summary

RBAC maintains a second copy of the route table and re-matches every request against it, even
though the router has already resolved the request by the time the RBAC middleware runs. The two
can silently disagree, and nothing checks them against each other.

This is the structural issue behind #3808 — that bug was one way the two matchers diverged; this is
the reason a divergence is possible at all.

### What is already shared

The matching *engine* is the same. Both the application router (`pkg/gofr/http/router.go:39`) and
RBAC (`pkg/gofr/rbac/config.go:168`) use gorilla/mux with `StrictSlash(false)`, and by the time the
RBAC middleware runs, `Router.ServeHTTP` has already normalised `r.URL.Path`, so both match the same
string with the same semantics.

What is duplicated is the **pattern strings**. `configs/rbac.json` is a hand-maintained second copy
of the route table, and a rule whose path does not correspond to any registered route — a typo like
`/api/user/{id}` against a registered `/api/users/{id}` — matches nothing, produces no error at load
or at request time, and leaves the route it was written to guard unguarded.

### The router's answer is already available and unused

GoFr registers middleware through `mux.Router.Use` (`pkg/gofr/http/router.go:155-161`), so
middleware runs *after* route matching. `mux.CurrentRoute(r).GetPathTemplate()` is therefore
available inside `rbac.Middleware` and returns the exact registered template. Verified against
GoFr's router shape (real routes plus the `PathPrefix("/")` catch-all from `gofr.go:183`):

```
GET    /api/users/42       -> template="/api/users/{id:[0-9]+}"  methods=[GET]
GET    /api/users/abc      -> template="/"                        methods=[]    (catch-all)
DELETE /api/users/42       -> template="/"                        methods=[]    (catch-all)
GET    /.well-known/health -> template="/.well-known/health"      methods=[GET]
```

Two things fall out of this:

1. The router's own resolution is free to read, and is authoritative by definition.
2. `template == "/"` with no methods means **no real route matched**, which cleanly separates
   "registered route with no RBAC rule" (an unguarded route — dangerous) from "this path matches
   nothing and is a 404 regardless" (harmless). That distinction is not currently available to the
   middleware, and it is the signal the deferred fail-closed work needs.

### Proposed work, in increasing order of commitment

**1. Startup consistency check (non-breaking, highest value).** Walk the router's registered
[...]

PR #4047, opened 2026-08-21
fix(rbac): compile endpoint patterns once at load, not per rule per request (#3979)

akshat-kumar-singhal:
**Description:**

Fixes items 1 and 2 of #3979: the RBAC matcher's data race, its unbounded router growth, and its
per-request allocation cost. Item 3 of that issue (multi-role JWT claim arrays) is independent and
needs a semantics decision first, so it is left for a separate PR.

`matchMuxPattern` built a route per rule, per request, on the `Config`'s shared `muxRouter`:

```go
route := router.NewRoute().Path(pattern)
```

`mux.Router.NewRoute` appends to the router it is called on (`mux@v1.8.1/mux.go:279-284`), and
`config.muxRouter` is created once at load and shared by every request. So each pattern test both
mutated shared state without synchronization and left its entry behind permanently.

Each pattern is now compiled **once at load**, onto a router of its own, and stored on `endpointRule`.
`Route.Match` only reads the route and writes into the caller's `RouteMatch`, so a route that is never
written to again is safe to share across goroutines. The shared `muxRouter` is gone.

A per-scan `matchContext` carries the request and a single reused `RouteMatch`. Without it one
`RouteMatch` escaped per rule scanned, so cost still grew with config size.

**This does not change which requests match which rules.** #3979 suggests reading
`mux.CurrentRoute(r).GetPathTemplate()` instead. That works, but it compares config paths against
registered route *templates* rather than request paths — the breaking change #3935 defers to its own
item 3 and argues should not be the default, since a rule like `/admin/{path:.*}` deliberately spans
many registered routes today. That belongs in #3935 on its own merits, not smuggled into a bug fix.

**Root cause analysis:**

- **Symptom.** Under concurrency, `-race` reports a race on `mux.(*Router).NewRoute` reached through
  `rbac.matchMuxPattern`. Separately, a long-lived process accumulates router entries without bound
  (0 → 114 routes after 100 requests, per #3979), and authorizing one request costs hundreds to
  thousands of allocations, on the hot path of every route in the application.
- **Root cause.** `Config` is documented as read-only after initialization — `getExactEndpoint` and
  `GetEndpointPermission` both carry the comment "Config is read-only after initialization, so no mutex
  is needed." `matchMuxPattern` broke that invariant: `NewRoute` is a mutating call, and it was being
  made on a shared field on every request. The pattern compilation it performed is fully determined by
  the config, so it never needed to happen at request time at all.
- **Why it wasn't caught.** The existing suites drive requests sequentially, and the package had no
  `-race` concurrency test and no benchmark. A sequential test cannot observe either failure: the race
  needs two goroutines, and the growth is invisible without counting routes or allocations.
- **Fix.** Compile at load onto a per-pattern router; store the compiled `*mux.Route` on the rule; reuse
  one `RouteMatch` per scan. `matchesHTTPMethod` is deleted — `endpointRule.matchesMethod` was its only
  caller and allocated a one-element slice per rule per request to reach it, and its list handling is
  unreachable now that `buildEndpointRules` expands methods into one rule each.
- **Prevention.** Three regression tests added, described below. The concurrency tests fail on the base
  branch with exactly the stack #3979 reported.

**Verification:**

`TestConfig_resolve_concurrent` and `TestMiddleware_concurrent` drive 50 goroutines × 20 requests
through the resolver and through the mounted middleware. Both fail `-race` on the base branch and pass
here; `go test -race -count=2 ./pkg/gofr/rbac/` is clean.

`BenchmarkConfig_resolve` (Apple M4), before → after:

| rules in config | before | after |
| --- | --- | --- |
| 6 | 36,730 ns/op, 1,064 allocs/op | 927 ns/op, 6 allocs/op |
| 21 | 131,730 ns/op, 3,924 allocs/op | 1,026 ns/op, 6 allocs/op |
| 51 | 373,669 ns/op, 9,684 allocs/op | 1,717 ns/op, 6 allocs/op |

`TestConfig_resolve_allocations` guards that flatness. It compares a small config against larger ones
rather than asserting an absolute count: the race detector adds three allocations, so a bound tuned
without `-race` fails CI with it and one tuned for `-race` is slack. The property the fix establishes
is that per-request cost does not scale with rule count, so that is what the test measures.

Package coverage is 94.3%.

None. Matching semantics are unchanged, and no exported API is touched — `muxRouter`,
`matchMuxPattern`, `matchesEndpointPattern` and `matchesHTTPMethod` are all unexported.

**Additional Information:**

No new dependencies. `endpoint_matcher_test.go` shrinks because `TestMatchesHTTPMethod` is removed with
the function it covered (superseded by `TestEndpointRule_matchesMethod`) and `TestMatchMuxPattern` is
retargeted at `compilePattern` as `TestCompilePattern`, which also gains cases for literal paths,
empty paths and patterns mux cannot compile.

Three pre-existing `noctx` lint findings remain in `endpoint_matcher_test.go` (`httptest.NewRequest`
rather than `NewRequestWithContext`). They are present on `development`, untouched here, and part of the
mechanical golangci-lint backlog.

**Checklist:**

### DIFF
--- a/pkg/gofr/rbac/resolver.go
+++ b/pkg/gofr/rbac/resolver.go
@@ -3,6 +3,8 @@ package rbac
 import (
 	"sort"
 	"strings"
+
+	"github.com/gorilla/mux"
 )
 
 // Segment specificity scores, compared segment by segment to order overlapping rules.
@@ -24,6 +26,9 @@ type endpointRule struct {
 	// pattern is the endpoint path, which may contain mux variables.
 	pattern string
 
+	// route is pattern compiled once at load time, or nil when pattern is a literal path.
+	route *mux.Route
+
 	// method is the upper-cased declared method, or "*" for all methods.
 	method string
 
@@ -44,12 +49,27 @@ type endpointRule struct {
 // permission requirement rather than granting access, covering an unknown verb tightens
 // enforcement rather than loosening it.
 func (r *endpointRule) matchesMethod(methodUpper string) bool {
-	return matchesHTTPMethod(methodUpper, []string{r.method})
+	return r.method == "*" || strings.EqualFold(r.method, methodUpper)
+}
+
+// matchesPath reports whether the rule's path covers the request carried by mc, which is shared
+// across the whole scan.
+func (r *endpointRule) matchesPath(path string, mc *matchContext) bool {
+	if r.pattern == "" {
+		return false
+	}
+
+	// A literal path is compared directly; only mux patterns carry a compiled route.
+	if r.route == nil {
+		return r.pattern == path
+	}
+
+	return mc.matches(r.route)
 }
 
 // matches reports whether the rule covers the given request.
-func (r *endpointRule) matches(methodUpper, path string, config *Config) bool {
-	return r.matchesMethod(methodUpper) && matchesEndpointPattern(r.endpoint, path, config)
+func (r *endpointRule) matches(methodUpper, path string, mc *matchContext) bool {
+	return r.matchesMethod(methodUpper) && r.matchesPath(path, mc)
 }
 
 // buildEndpointRules expands endpoints into one rule per declared method and orders them
@@ -89,6 +109,7 @@ func buildEndpointRules(endpoints []EndpointMapping) []endpointRule {
 			byKey[key] = endpointRule{
 				endpoint:    endpoint,
 				pattern:     endpoint.Path,
+				route:       compilePattern(endpoint.Path),
 				method:      methodUpper,
 				isPublic:    endpoint.Public,
 				pathScore:   pathSpecificity(endpoint.Path),
@@ -212,9 +233,11 @@ func compareSpecificity(a, b []int) int {
 }
 
 // resolveEndpoint returns the most specific rule covering the request, and whether it is public.
-func resolveEndpoint(methodUpper, path string, rules []endpointRule, config *Config) (*EndpointMapping, bool) {
+func resolveEndpoint(methodUpper, path string, rules []endpointRule) (*EndpointMapping, bool) {
+	mc := newMatchContext(methodUpper, path)
+
 	for i := range rules {
-		if rules[i].matches(methodUpper, path, config) {
+		if rules[i].matches(methodUpper, path, mc) {
 			return rules[i].endpoint, rules[i].isPublic
 		}
 	}
@@ -236,5 +259,5 @@ func (c *Config) resolve(methodUpper, path string) (*EndpointMapping, bool) {
 		return endpoint, isPublic
 	}
 
-	return resolveEndpoint(methodUpper, path, c.rules, c)
+	return resolveEndpoint(methodUpper, path, c.rules)
 }
--- a/pkg/gofr/rbac/config.go
+++ b/pkg/gofr/rbac/config.go
@@ -9,7 +9,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/gorilla/mux"
 	"go.opentelemetry.io/otel/trace"
 	"gopkg.in/yaml.v3"
 
@@ -124,7 +123,6 @@ type Config struct {
 	endpointPermissionMap map[string][]string         `json:"-" yaml:"-"` // Key: "METHOD:/path", Value: []permissions
 	publicEndpointsMap    map[string]bool             `json:"-" yaml:"-"` // Key: "METHOD:/path", Value: true if public
 	endpointMap           map[string]*EndpointMapping `json:"-" yaml:"-"` // Key: "METHOD:/path", Value: endpoint object
-	muxRouter             *mux.Router                 `json:"-" yaml:"-"` // Used for mux pattern matching
 
 	// rules holds every (method, path) rule ordered most-specific-first, and is the
 	// single source of truth for resolving a request that no exact key matches.
@@ -162,10 +160,6 @@ func LoadPermissions(path string, logger datasource.Logger, metrics container.Me
 	config.Logger = logger
 	config.Metrics = metrics
 	config.Tracer = tracer
-
-	// Initialize mux router for pattern matching
-	// Use StrictSlash(false) to match the application router's behavior
-	config.muxRouter = mux.NewRouter().StrictSlash(false)
 
 	// Validate config before processing
 	if err := config.validate(); err != nil {
@@ -282,8 +276,6 @@ func (c *Config) initializeMaps() {
 	c.endpointPermissionMap = make(map[string][]string)
 	c.publicEndpointsMap = make(map[string]bool)
 	c.endpointMap = make(map[string]*EndpointMapping)
-	// Use StrictSlash(false) to match the application router's behavior
-	c.muxRouter = mux.NewRouter().StrictSlash(false)
 }
 
 // buildRolePermissionsMap builds the role permissions map from Roles.
--- a/pkg/gofr/rbac/endpoint_matcher.go
+++ b/pkg/gofr/rbac/endpoint_matcher.go
@@ -30,56 +30,59 @@ var (
 	errInvalidPattern = errors.New("invalid mux pattern")
 )
 
-// matchesHTTPMethod checks if the HTTP method matches the endpoint's allowed methods.
-func matchesHTTPMethod(method string, allowedMethods []string) bool {
-	// Empty methods or "*" means all methods
-	if len(allowedMethods) == 0 {
-		return true
-	}
-
-	for _, m := range allowedMethods {
-		if m == "*" || strings.EqualFold(m, method) {
-			return true
-		}
-	}
-
-	return false
-}
-
 // isMuxPattern detects if a pattern contains mux-style variables.
 // Returns true if pattern contains { and }.
 func isMuxPattern(pattern string) bool {
 	return strings.Contains(pattern, "{") && strings.Contains(pattern, "}")
 }
 
-// matchMuxPattern uses mux Route.Match() to test if a path matches a mux pattern.
-// Creates a temporary mux Route and uses Route.Match() to test the pattern.
-// Handles all mux pattern types: {id}, {id:[0-9]+}, {path:.*}, etc.
-func matchMuxPattern(pattern, method, path string, router *mux.Router) bool {
-	if router == nil {
-		return false
+// compilePattern compiles an endpoint path into a mux route, once, at load time.
+// Returns nil for a literal path, which is cheaper to compare as a string.
+//
+// Each pattern is compiled onto a router of its own. mux.Router.NewRoute appends to the router it
+// is called on, so compiling against a router shared by every request - as this package used to -
+// both mutated that router without synchronization and grew it by one entry per rule per request,
+// for the life of the process. A router built here is never written to again, and Route.Match only
+// reads the route, so the compiled result is safe to share across goroutines.
+func compilePattern(pattern string) *mux.Route {
+	if pattern == "" || !isMuxPattern(pattern) {
+		return nil
 	}
 
-	// Create a temporary route with the pattern
-	route := router.NewRoute().Path(pattern)
+	// StrictSlash(false) matches the application router's behavior.
+	return mux.NewRouter().StrictSlash(false).NewRoute().Path(pattern)
+}
 
-	// If method is specified, add it to the route
-	if method != "" {
-		route = route.Methods(method)
-	}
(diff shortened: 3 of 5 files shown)


aryanmehrotra (dismissed):
Approving. I reviewed only the delta over #3934 — `git diff <3934 head> <4047 head>` is six files, all in `pkg/gofr/rbac`, +297/-237 — and reproduced every claim in the description rather than reading them.

### What it does

`matchMuxPattern` called `router.NewRoute()` on the `Config`'s single shared `muxRouter`, once per rule, per request. `NewRoute` appends to the router it is called on (`mux@v1.8.1/mux.go:279-284`), so every authorization both wrote to shared state with no synchronisation and left an entry behind for the life of the process. Each pattern is now compiled once at load onto a router of its own and stored on the rule; a per-scan `matchContext` carries the request and one reused `RouteMatch`.

### The safety argument holds

I checked the load-bearing claim in mux rather than taking it: `Route.Match` (`route.go:70-108`) reads `r.buildOnly`, `r.err`, `r.matchers`, `r.handler` and `r.regexp`, and writes only into the caller's `*RouteMatch`. `routeRegexpGroup.setMatch` has a **value** receiver and likewise writes only into the match. There is no lazy compilation on the match path — `Path()` compiles eagerly at build time. A route that is never written to again is safe to share across goroutines, as claimed.

### What I ran

**The race, on a real server.** I stood up a GoFr app with `EnableRBAC` on a nine-rule config covering `["*"]`, an omitted `methods`, an overlapping `/{scope}/orgs/{org_id}` vs `/team/orgs/{org_id}` pair, a public-then-protected duplicate and an uncompilable `/bad/{id:[}`, and drove it under `go run -race`:

| | #3934 base | this PR |
| --- | --- | --- |
| `DATA RACE` reports | **2** | **0** |
| process exit | non-zero | 0 |
| 10 authorization probes | 10/10 as expected | **identical, 10/10** |
| 200 concurrent `GET /team/orgs/9` | `map[403:200]` | `map[403:200]` |

The base traces are exactly #3979's: `mux.(*Router).NewRoute` ← `matchMuxPattern` ← `matchesEndpointPattern` ← `resolveEndpoint`, through `getEndpointForRequest` and the mounted middleware. The probe results being byte-identical on both sides is the evidence I wanted for "this does not change which requests match which rules".

**The regression tests are not vacuous.** I copied `resolver_concurrency_test.go` onto the base branch unchanged. All three fail there and pass here:

| test | on #3934 base | on this PR |
| --- | --- | --- |
| `TestConfig_resolve_concurrent` | FAIL (race) | pass |
| `TestMiddleware_concurrent` | FAIL (race) | pass |
| `TestConfig_resolve_allocations` | FAIL | pass |

**Allocations.** `BenchmarkConfig_resolve`, same machine, both branches:

| rules | base allocs/op | this PR |
| --- | --- | --- |
| 6 | 1,064 | **6** |
| 21 | 3,924 | **6** |
| 51 | 9,684 | **6** |

Those match the numbers in the description exactly. `B/op` is a flat 896 across all three sizes. My ns/op differ from yours in absolute terms — busier machine, and I capped the base run at a fixed iteration count because it is so slow — but the shape is the same: linear before, flat after.

`go test -race -count=2 ./pkg/gofr/rbac/` is clean. `golangci-lint` reports only the pre-existing `noctx` findings; `endpoint_matcher_test.go` has the same four bare `httptest.NewRequest` call sites as `development`, untouched — agreed, backlog, not this PR.

### Notes, none of them blocking

**Coverage is 94.9% → 94.7%, not flat.** The checklist says it does not decrease. It does, slightly, and for a benign reason: the PR deletes fully covered code (`matchesHTTPMethod`, `checkPatternMatch` and friends). The only new uncovered line is `resolver.go:58` — the `if r.pattern == "" { return false }` guard in `matchesPath` — and `matchesEndpointPattern` had the same hole on the base branch, so nothing regressed. A one-line case for an empty `Path` would close it if you want the box to be true.

**The same pattern is compiled once per declared method.** `compilePattern` sits inside the method loop in `buildEndpointRules`, so an endpoint with five methods builds five routers and five identical regexps, and `logUncompilablePattern` compiles it once more during validation. Load-time only and cheap, so I would not hold the PR for it, but a small `map[string]*mux.Route` memo would remove the duplication.

**Half of what is left is the synthetic request.** I profiled the six remaining allocations: `newMatchContext` is 46% of them (the `matchContext`, the `http.Request` and the `url.URL`), the rest is mux's `RouteMatch.Vars` map and the regexp machine, which `Route.Match` gives us no way to avoid. `getEndpointForRequest` already holds the real `*http.Request` and could pass it straight through — but `GetEndpointPermission(method, path)` has no request, so the synthetic one has to stay for that caller. Worth a follow-up, not a change here.

**Two trivial ones.** `matchesMethod` uses `strings.EqualFold` on two strings that are both already upper-cased — `r.method` by `buildEndpointRules`, the argument by every caller — so `==` would do. And `matchesPath(path, mc)` takes both a path and a context built from that same path; the literal branch reads one and the pattern branch the other. They cannot disagree with one caller, but folding `path` onto `matchContext` would make that structural.

**The `buildEndpointRules` docstring** still says identically-scoring rules "keep their declaration order, which is the one case where the order entries are written in still decides the outcome", which the public-vs-protected tie-break above it contradicts. That is #3934's line and I have already raised it there — it will arrive with that fix, nothing to do here.

### Merge order

This is stacked, so it cannot land before #3934, which is still sitting at changes-requested. Once that merges, rebase and this should reduce to the single commit.

Nice piece of work — the root-cause writeup is accurate, the tests fail on the base branch for the right reason, and the fix removes the shared mutable state rather than locking around it.

Umang01-hash (dismissed):
LGTM. Reproduced the race on development first — the new TestConfig_resolve_concurrent fails there under -race with the exact mux.NewRoute ← matchMuxPattern stack from #3979, and passes clean at head (-race -count=2). Compiling each pattern once onto its own router and only calling Route.Match at request time is the right shape; confirmed in mux v1.8.1 that Match reads the route and writes only into the caller's RouteMatch, so sharing it is safe. Reused RouteMatch is zeroed per call, so no MatchErr carryover across the scan.

Matching semantics are unchanged — decision logic is identical to #3934, old matchMuxPattern was called path-only (empty method), and both use StrictSlash(false). BenchmarkConfig_resolve is flat at 6 allocs/op across 6/21/51 rules (was 1064/3924/9684). Exported surface unchanged, dead funcs fully removed, lint --new-from-rev 0 issues, coverage 94.7%.

One trivial nit, non-blocking: the buildEndpointRules godoc now reads as if identically-scored rules go straight to declaration order, but the sort still applies the method-then-enforcement tiebreaks (unchanged) — the example just illustrates equal path score only.

Residual unmatched-route / uncompilable-pattern fail-open is pre-existing and correctly left to #3935; item 3 (multi-role JWT arrays) rightly split out.

pkg/gofr/rbac/resolver.go
```diff
@@ -44,20 +49,34 @@ type endpointRule struct {
...
 
 // buildEndpointRules expands endpoints into one rule per declared method and orders them
 // most-specific-first, so that a narrower rule governs a request even when a broader one is
-// declared ahead of it. Rules that score identically on the path - two patterns of the
-// same shape, such as "/{a}/{b}" and "/{x}/{y}" - are separated first by method, an explicitly
-// declared one outranking "*", and then by enforcement, a rule that requires permissions
-// outranking a public one. Declaration order decides only between rules alike on all three.
+// declared ahead of it. Rules that score identically - two patterns of the same shape, such as
+// "/{a}/{b}" and "/{x}/{y}" - keep their declaration order, which is the one case where the
+// order entries are written in still decides the outcome.
```
aryanmehrotra: The rebase onto `df192fc08` brought back the pre-merge version of this comment, undoing `7badc52d0` from #3934. That commit rewrote it because the sort below no longer leaves same-shape ties to declaration order. It orders them by `methodScore` first (explicit method beats `*`), then by enforcement (a protected rule beats a public one). Declaration order only decides between rules that are alike on all three.

As written, the comment says the opposite of what the code does, and this is authz ordering. Please restore the merged text:

```go
// declared ahead of it. Rules that score identically on the path - two patterns of the
// same shape, such as "/{a}/{b}" and "/{x}/{y}" - are separated first by method, an explicitly
// declared one outranking "*", and then by enforcement, a rule that requires permissions
// outranking a public one. Declaration order decides only between rules alike on all three.
```

pkg/gofr/rbac/resolver_test.go
```diff
@@ -110,51 +110,18 @@ func TestGetEndpointForRequest_MostSpecificWins(t *testing.T) {
 }
 
 func TestGetEndpointForRequest_ExplicitMethodBeatsWildcard(t *testing.T) {
```
aryanmehrotra, line 112: The rebase also reverted this test to its pre-merge version, undoing `56456e2a3` from #3934 and dropping 4 of the 5 cases it landed with. The ones that matter are the **pattern-path** cases (`/admin/{id}`). A literal path is answered by the exact-key fast path and never reaches the ordered scan, so those cases were the only test of the `methodScore` tie-break in `resolveEndpoint`.

This PR rewrites exactly that scan (`matchesPath`, compiled `route`, `matchContext`), so this is the test the change most needs. I put `df192fc08`'s version of the file back on top of this head and all 5 subtests pass under `-race`. The behaviour is intact; only the coverage is gone.

Fix:

```sh
git checkout df192fc08 -- pkg/gofr/rbac/resolver_test.go
```

This PR doesn't otherwise change `resolver_test.go` relative to `df192fc08`, so taking the merged file whole is safe.

aryanmehrotra (commented):
**The conflict resolution reverted part of #3934. Behaviour is fine, but a comment and a test need restoring before merge.**

This branch was stacked on an earlier copy of #3934. #3934 had one more round of changes before it was squash-merged as `df192fc08`, and the rebase (`57a5d1e99` → `b9ca6ea99`) kept the branch's stale copy. `git diff 57a5d1e99 b9ca6ea99` is empty, so the rebased tree is byte-identical to the pre-rebase one and #3934's last round is gone.

The two reverted hunks are exactly `56456e2a3` and `7badc52d0` from #3934. Both were made to settle review threads there (the `methodScore` coverage gap, and the tie-break docstring correction), so the rebase re-opens both.

| Reverted | Where | Impact |
|---|---|---|
| Tie-break doc comment (`7badc52d0`) | `resolver.go:77-79` | Claims declaration order decides ties; the code sorts by method, then enforcement |
| Pattern-path + fallthrough cases of `TestGetEndpointForRequest_ExplicitMethodBeatsWildcard` (`56456e2a3`) | `resolver_test.go:112` | Only test of the `methodScore` tie-break in the ordered scan, which is the code this PR rewrites |

**Verified:**
- `rbac` package passes at head, with and without #3934's test restored (`-race`, all 5 subtests green).
- The rest of #3934 (`config.go`, `config_test.go`, `endpoint_matcher*.go`, docs) came through intact. The only other difference is the `isCatchAllVariable` comment, and 9503cb0e7 makes the same rewording.
- Nothing else from `development` is affected: #3934 is the only commit the rebase moved over.

**To fix:** see the two inline comments. It's a one-line `git checkout` for the test plus a 4-line comment edit.

pkg/gofr/rbac/resolver.go
akshat-kumar-singhal: Fixed in `cdbd8983` — restored `7badc52d0`'s text verbatim, so the comment block is now byte-identical to `df192fc08`'s. You're right that as it stood it said the opposite of what the sort does.

Cause was as you diagnosed: the branch was stacked on a pre-merge copy of #3934 and the rebase kept the stale hunk.

pkg/gofr/rbac/resolver_test.go
akshat-kumar-singhal: Fixed in `cdbd8983` via exactly the `git checkout df192fc08 -- pkg/gofr/rbac/resolver_test.go` you suggested.

Confirmed your claim that the file needs no other change on this branch: `git diff df192fc08 -- pkg/gofr/rbac/resolver_test.go` is now empty, so the PR's net diff against `development` is five files and `resolver_test.go` has dropped out of it entirely.

All 5 subtests pass under `-race` at the new head, including both `pattern path` cases — which, as you note, are the only ones that reach the ordered scan this PR rewrites. Full package: `go test -race -count=1 ./pkg/gofr/rbac/` clean.

aryanmehrotra (approved):

PR #4049, opened 2026-08-21
chore(deps): bump tj-actions/changed-files 47.0.0 → 47.0.6

Umang01-hash:
Patch bump of the `tj-actions/changed-files` GitHub Action (actions group), consolidated from Dependabot #3980.

- `.github/workflows/go.yml`: `tj-actions/changed-files` `v47.0.0` → `v47.0.6`, SHA-pinned to `9426d40962ed5378910ee2e21d5f8c6fcbf2dd96`.

Patch-level within v47; no workflow logic changes.

Closes #3980

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -572,7 +572,7 @@ jobs:
       # This implements a changed-files based approach as suggested by the maintainer
       - name: Get Changed Files
         id: changed-files
-        uses: tj-actions/changed-files@24d32ffd492484c1d75e0c0b894501ddb9d30d62 # v47.0.0
+        uses: tj-actions/changed-files@9426d40962ed5378910ee2e21d5f8c6fcbf2dd96 # v47.0.6
         with:
           files: |
             pkg/**/*.go


aryanmehrotra (approved):
Verified the pin rather than trusting the comment: `9426d40962ed5378910ee2e21d5f8c6fcbf2dd96` is
exactly what `v47.0.6` resolves to upstream, and the trailing version comment moves with it.

Checked that nothing we depend on shifted:

- `action.yml` gains only two inputs — `exclude_symlinks` and `skip_same_sha` — both defaulting to
  `"false"`, so prior behaviour is preserved. **No outputs changed**, which is what matters here since
  `Find Changed Submodules` reads `steps.changed-files.outputs.any_changed`.
- `code_quality` checks out with `fetch-depth: 0`, so the SHA-resolution rework in `commitSha.ts` —
  which is mostly about shallow clones and missing history — doesn't apply to us.
- Nothing flagged as breaking in the upstream history between the two tags.

One correction to the description: "patch-level within v47; no workflow logic changes" is true of
*our* workflow, but upstream moved 99 commits with `dist/index.js` at +1630/-343. Worth stating
accurately given this is the action whose tags were retargeted to a malicious commit in March 2025 —
the SHA pin is exactly what protects us, and the pin is correct here.

Approving.

PR #4050, opened 2026-08-21
chore(deps): bump zopdev/static-server v0.0.9 → v1.0.0 in /docs — NEEDS REVIEW

Umang01-hash:
## Docker base-image major bump (needs review)

⚠️ **Major version bump** (v0 → v1) of the docs site's static-server image. Consolidated from Dependabot #3990.

- `docs/Dockerfile` stage 2: `FROM zopdev/static-server:v0.0.9` → `v1.0.0`.

Only affects how the **docs site** is served (the second build stage) — not the GoFr framework or any shipped module. Please confirm `v1.0.0` still serves the `static` directory the same way before merging.

Closes #3990


aryanmehrotra (approved):
Answering the question in the description directly — yes, `v1.0.0` serves the `static` directory the
same way, and here is the evidence rather than an assurance.

**The image config is byte-identical across both tags.** Pulled both manifests from the registry:

```
                v0.0.9                      v1.0.0
WorkingDir      /                           /
Cmd             ['/main']                   ['/main']
Entrypoint      None                        None
User            nonroot:nonroot             nonroot:nonroot
Env             PATH, SSL_CERT_FILE         PATH, SSL_CERT_FILE
```

So `COPY --from=builder /app/out static` still lands at `/static` and `CMD ["/main"]` still resolves.

**The major bump is a stability declaration, not a break.** v0.0.9 -> v1.0.0 is a single commit over
nine files with no behavioural removals. `STATIC_DIR_PATH` resolution is unchanged; the only edit
makes an *empty* configured value fall back to the default instead of rooting at the working
directory, which is strictly more robust than what we have today.

**Both new features are inert for this site.**

- `_headers` support — our docs tree ships no `_headers` file, and upstream documents "absent file →
  no rules → responses are byte-for-byte what they are today". Something we could adopt later if we
  want the security headers, but nothing changes on merge.
- Markdown negotiation — fires only when `Accept` explicitly *names* `text/markdown`, and only on
  extensionless routes. Browsers send `*/*;q=0.8`, which upstream deliberately does not match. Our
  Next.js static export emits HTML, not `.md` siblings, so negotiation falls through untouched.
  `docs/AGENTS.md` lands as a directly-requested `.md`, which upstream leaves alone.

One thing I could not close from here: I can't inspect what `ghcr.io/gofr-dev/website` actually
writes into `/app/out` without pulling it. If that builder does emit `.md` siblings, agents asking for
markdown would start receiving them — which is the intended feature rather than a regression, but
worth a glance on stage.

Approving.

PR #4048, opened 2026-08-21
chore(deps): consolidate minor/patch dependency updates (2026-08-21)

Umang01-hash:
## Consolidated Dependabot updates (minor/patch)

Combines **64** individual Dependabot PRs (root module + 26 `pkg/` submodules + 2 examples) into one. Only `go.mod` / `go.sum` / `go.work.sum` are touched — **no source, no binaries**. Every manifest is reproduced by `go mod tidy` (nothing hand-edited).

### Headline updates
| Package | Old | New | Scope |
|---|---|---|---|
| github.com/stretchr/testify | 1.11.1 | 1.12.1 | 30 modules |
| go.opentelemetry.io/otel/trace (+ otel/otel, metric) | 1.44.0 | 1.45.0 | 19 modules |
| github.com/aws/aws-sdk-go-v2/service/dynamodb | 1.60.1 | 1.63.3 | root, kv-store/dynamodb |
| github.com/aws/aws-sdk-go-v2/service/sqs | 1.45.1 | 1.46.6 | root, pubsub/sqs |
| github.com/nats-io/nats.go | 1.52.0 | 1.53.1 | root, kv-store/nats, pubsub/nats |
| github.com/Azure/azure-sdk-for-go/sdk/azcore | 1.22.0 | 1.23.0 | file/azure, pubsub/eventhub |
| github.com/godror/godror | 0.51.0 | 0.51.4 | oracle |
| github.com/elastic/go-elasticsearch/v8 | 8.19.6 | 8.19.7 | root, elasticsearch |
| github.com/pkg/sftp | 1.13.10 | 1.13.11 | root, file/sftp |

Transitives moved along by `tidy` (e.g. `go-logr/logr` 1.4.3→1.4.4, `stretchr/objx` 0.5.2→0.5.3, `x/net`, `x/text`).

### Verification done locally
- ✅ `go mod tidy -diff` clean on **all** `pkg/` submodules (matches the CI tidiness gate).
- ✅ `go build ./...` green on root + all 26 `pkg/` library modules; `go vet` green on the two example modules.
- ✅ Guard: only `go.mod`/`go.sum`/`go.work.sum` in the diff — no source, no compiled binaries.

### Not included here
- **redis/go-redis/v9** (#3987) — 9.22.0 adds `BLMoveM` to `redis.Cmdable`, which GoFr's `Redis` interface embeds, so the generated mock needs regeneration. Ships in its own PR.
[...]

PR #4051, opened 2026-08-21
chore(deps): bump redis/go-redis/v9 9.21.0 → 9.22.0 (+ mock regen)

Umang01-hash:
## redis/go-redis/v9 9.21.0 → 9.22.0

Split out of the minor/patch consolidation (#4048) because it needs a generated-source change, not just manifests.

### Why the mock changes
`go-redis` 9.22.0 adds new methods to `redis.Cmdable` (`BLMoveM`, `LMoveM`, `ClientTracking*`, `HImport*`, `InfoMap`, `SDiffCard`, `SUnionCard`, and the `TS*` time-series calls). GoFr's `container.Redis` interface embeds `redis.Cmdable`, so the generated `MockRedis` stops satisfying the interface until regenerated — the build fails with `*MockRedis does not implement Redis (missing method BLMoveM)` otherwise.

### Changes
- `go.mod` / `go.sum`: `github.com/redis/go-redis/v9` 9.21.0 → 9.22.0.
- `pkg/gofr/container/mock_datasources.go`: **regenerated** via `go generate` (`go.uber.org/mock/mockgen -source=datasources.go`). Diff is **pure additions (319 insertions, 0 deletions)** — only the new `Cmdable` methods.

### Verification
- ✅ `go build ./...` green on root + all 26 `pkg/` library modules.
- ✅ All `pkg/` submodules pass `go mod tidy -diff` (CI tidiness gate).
- ✅ Mock diff is generated output only — no hand edits, no removals.

Closes #3987

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/prometheus/client_golang v1.24.1
 	github.com/prometheus/otlptranslator v1.0.0
 	github.com/redis/go-redis/extra/redisotel/v9 v9.21.0
-	github.com/redis/go-redis/v9 v9.21.0
+	github.com/redis/go-redis/v9 v9.22.0
 	github.com/segmentio/kafka-go v0.4.51
 	github.com/stretchr/testify v1.11.1
 	github.com/vektah/gqlparser/v2 v2.5.36
--- a/go.sum
+++ b/go.sum
@@ -185,8 +185,8 @@ github.com/redis/go-redis/extra/rediscmd/v9 v9.21.0 h1:jsV3tyMeJrEoc2f3EhNf7qoBW
 github.com/redis/go-redis/extra/rediscmd/v9 v9.21.0/go.mod h1:e5t17bY9cEpVV+xw2U7jsPOKkXBtL5IQmNVABShnHUk=
 github.com/redis/go-redis/extra/redisotel/v9 v9.21.0 h1:36qq3rbF2If2CP0zGHHF8o/4XDluErn6DD0c9/L2iNI=
 github.com/redis/go-redis/extra/redisotel/v9 v9.21.0/go.mod h1:7y2cVB/LXXLHqHOO2jCVzBqimIQk1w7Rp9WSpyVY/o8=
-github.com/redis/go-redis/v9 v9.21.0 h1:FPBE4hhbAke+TLmcY3WkpbDffJEomdqPn3HYiqAtL9E=
-github.com/redis/go-redis/v9 v9.21.0/go.mod h1:v/M13XI1PVCDcm01VtPFOADfZtHf8YW3baQf57KlIkA=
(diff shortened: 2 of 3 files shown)


aryanmehrotra (approved):
Approving — splitting this out of #4048 was the right call, and the regeneration is sound. But there
is a behaviour change riding along that isn't in the description and should reach the release notes.

## What I verified

The mock is genuinely generated, not hand-edited. I re-ran it myself:

```
go run go.uber.org/mock/mockgen -source=datasources.go -destination=mock_datasources.go -package=container
```

and the committed file is **byte-identical** to the output. The `+319/-0` is itself good evidence:
a purely additive regenerated mock proves `redis.Cmdable` only *gained* methods and changed none.

Also checked, since a go.work workspace makes version skew easy to miss:

- Root module builds; `pkg/gofr/container` and `pkg/gofr/datasource/redis` tests pass; root
  `go.mod`/`go.sum` stay tidy.
- `redisotel` and `rediscmd` stay at 9.21.0 against go-redis 9.22.0 — they compile fine, the `Hook`
  interface is unchanged.
- Standalone submodule builds are unaffected: they require published `gofr.dev v1.57.0` rather than
  replacing to the local tree, so they keep resolving 9.21.0 until the next release.
- The `WaitAOF` return-type change upstream flags as breaking does **not** apply to us — it is a
  method on the concrete `cmdable` struct and is not declared in the `Cmdable` interface we embed.
  That is also why the regenerated mock shows zero deletions.

## The part that needs recording

The 9.22.0 notes flag a second upgrade concern: *"Default configuration values changed (#3918) …
Explicitly configured values are unaffected."*

We do not configure them. `pkg/gofr/datasource/redis/config.go:62-77` sets only `Addr`, `Username`,
`Password`, `DB` and optionally `TLSConfig`, so we inherit every default:

| default | 9.21.0 | 9.22.0 |
| --- | --- | --- |
| `ReadTimeout` | 3s | **5s** |
| `WriteTimeout` | follows ReadTimeout, 3s | follows ReadTimeout, **5s** |
| `MinRetryBackoff` | 8ms | 10ms |
| `DialTimeout` | 5s | unchanged |
| `MaxRetries` | 3 | unchanged |

So after this merges, a handler hitting a stalled Redis blocks **5s instead of 3s** before erroring —
a 67% increase in worst-case latency on that path. It compiles, it tests green, and it surfaces only
as requests breaching an upstream timeout in production.

There is no escape hatch either: the config surface is `REDIS_HOST/PORT/USER/PASSWORD/DB/TLS_*/
PUBSUB_*/STREAMS_*`, with no `REDIS_READ_TIMEOUT` or equivalent, so an application cannot override it.

Nothing here changes the verdict — the bump is right and holding it helps nobody. Two asks:

1. Add the timeout change to the release notes for whichever version carries this.
2. Worth exposing these as `REDIS_*` config in their own change, so the defaults stop being something
   we inherit silently on every go-redis bump. Happy to file that separately.

Issue #3855, opened 2026-08-04
Static file serving follows symlinks out of the served directory; replace lexical containment with os.Root


aryanmehrotra:
### Summary

A symlink inside a static directory that points outside it is served. The containment check in `staticHandler` compares strings; string comparison is lexical and symlinks are physical, so every variant of that check — the one on `development`, the one before #3041, and the one in #3820 — approves the request and `os.Open` then follows the link out of the tree.

`os.Root` (Go 1.24) is the mechanism for this, and we already require Go 1.26.

### Reproduction

```go
app := gofr.New()
app.AddStaticFiles("/static", "./public")
app.Run()
```

```
public/
  logo.png
  evil.txt      -> ../secrets/passwords.txt      (symlink)
  evildir       -> ../secrets                    (symlink)
secrets/
  passwords.txt
```

```
GET /static/evil.txt            -> 200  contents of secrets/passwords.txt
GET /static/evildir/secret.txt  -> 200  contents of secrets/secret.txt
```

No traversal sequence is involved, so nothing upstream stops it: `Router.ServeHTTP`'s `path.Clean` has nothing to clean, and `filepath.Abs(filepath.Join("/app/public", "/evil.txt"))` is `/app/public/evil.txt`, which is lexically inside the served directory. Every containment implementation we have ever shipped approves it:

| link | kind | pre-#3041 | `development` | #3820 | `os.Root` |
|---|---|---|---|---|---|
| `rel-inside.png` | relative → inside | allowed | allowed | allowed | allowed |
| `abs-inside.png` | absolute → inside | allowed | allowed | allowed | **rejected** |
| `dotdot-inside.png` | relative with `..` → inside | allowed | allowed | allowed | allowed |
| `abs-outside.txt` | absolute → **outside** | allowed | allowed | allowed | **rejected** |
| `rel-outside.txt` | relative → **outside** | allowed | allowed | allowed | **rejected** |

Note the third row: `os.Root` is not refusing `..` as a syntax. It resolves the link and checks where it actually lands, so a relative link containing `..` that returns inside the tree is still served. Only paths that genuinely leave the root are refused.

### Why the check cannot be fixed in place

[...]

PR #4053, opened 2026-08-22
fix(http): handle static directory paths safely

nik-7-alt:
**Description:**

- Fixed static file directory boundary checking in `isWithinDirectory`.
- Normalized and cleaned both the configured static directory and requested file path before comparison.
- Ensures files inside the static directory, including nested files, are correctly served.
- Prevents sibling directories with shared prefixes from being incorrectly treated as part of the static directory.
- Correctly handles requests for the static directory itself.
- No new dependencies were added.

- None.

**Additional Information:**

- The implementation uses `filepath.Clean` and `filepath.ToSlash` for platform-independent path normalization.
- Code was formatted using `gofmt`.
- Targeted test executed successfully:
  `go test ./pkg/gofr/http -run Test_isRestrictedFile -v`
- All cases in `Test_isRestrictedFile` passed successfully.
- The broader static-file test suite was also executed. The remaining permission-denied failures are due to Unix permission semantics (`0000`) not being enforced equivalently on Windows.

**Checklist:**

- [x] I have formatted my code using `gofmt`.
- [ ] This PR does not decrease the overall code coverage.


aryanmehrotra (commented):
Thanks for the PR, @nik-7-alt. I dug into this one carefully and I don't think it should merge in its current form — the change is a no-op, and there are a couple of process items alongside it. Details below so you can check my working.

### What the change actually is

The diff GitHub renders here (86 files, +4483) is base-branch drift, not your work. The real commit is **1 file, +5/−2** in `pkg/gofr/http/router.go`, touching only `isWithinDirectory`.

| | |
|---|---|
| Base | `main` (should be `development`) |
| Position | 1 ahead, 5128 behind `main`; branched at `24d51d6e3` (2026-08-18) |
| Linked issue | none |
| Tests added | none |
| CI | only `security/snyk` ran — `Workflow-Pipeline` never executed |

---

### 1. The change is a no-op — both operands are already cleaned

`directoryName` is set from `filepath.Abs(dirName)` ([`router.go#L349`](https://github.com/gofr-dev/gofr/blob/main/pkg/gofr/http/router.go#L349)) and `absPath` from `filepath.Abs(filepath.Join(...))` ([`router.go#L405`](https://github.com/gofr-dev/gofr/blob/main/pkg/gofr/http/router.go#L405)). Both `Abs` and `Join` call `Clean` internally, so `filepath.Clean` here has nothing left to do. On Unix `filepath.ToSlash` is identity; on Windows it converts both operands consistently, exactly as the existing `os.PathSeparator` comparison already does.

I ran the real call-site pipeline over 6 directory forms x 11 URL forms (`./public`, `public/`, `/app/public/`, `/app/./public`, `/app/public/../public` against `/`, `/../secrets/x`, `/a/../../y`, `//x`, `/openapi.json`, ...):

```
cases where Clean changed anything: 0
cases where old != new: 0

sibling shared-prefix  old=false new=false (want false)
directory itself       old=true  new=true  (want true)
nested file            old=true  new=true  (want true)
```

All three behaviours the description lists as fixed **already hold on `development`/`main` today**. The trailing-separator sibling guard and the `absPath == directoryName` root case are both present, with a comment explaining each ([`router.go#L473-L482`](https://github.com/gofr-dev/gofr/blob/main/pkg/gofr/http/router.go#L473-L482)).

If I've missed a case, the quickest way to settle it is a test that fails on `development` and passes with this patch. I tried to construct one and couldn't.

### 2. Wrong base branch

`CONTRIBUTING.md` line 17: *"submit a PR to development branch. Please note that all PRs are merged from feature branches to development first."* Targeting `main` also means the release path (development to main) would never pick this up.

Please retarget to `development` and rebase — the branch point is ~3 weeks stale.

### 3. It introduces a lint failure

`path := filepath.ToSlash(filepath.Clean(absPath))` shadows the `path` package, which `router.go` imports and uses (`ServeHTTP` calls `path.Clean`). Applying the patch to `main` at `9e2cd5f5c`:

```
baseline main:  8 issues  (goconst 4, gocyclo 1, noctx 3)
with patch:     9 issues  (+ gocritic 1)

pkg/gofr/http/router.go:481:2: importShadow: shadow of imported package 'path' (gocritic)
```

`gocritic` runs with the `style` and `opinionated` tags enabled and `importShadow` is not in `disabled-checks`, so this fails the lint gate once CI is allowed to run.

### 4. No tests, and the checklist says otherwise

The commit touches one non-test file. `CONTRIBUTING.md` line 15 requires tests with every contribution, and the PR ticks *"All new code is covered by unit tests."* Nothing was added.

### 5. The real defect in this function is #3855 — and that issue deletes the function

If what you're after is making static file serving genuinely safe, #3855 is the live bug: a symlink inside the served directory pointing outside it **is served**, because lexical containment answers *"does this string sit under that string"* while the question that matters is *"is this file reachable by walking down from the served directory."* No amount of `Clean`/`ToSlash` closes that — the issue explains why the check can't be fixed in place.

The fix there is `os.OpenRoot` at registration, and `isWithinDirectory` goes away entirely. `go.mod` pins `go 1.26.0`, so `os.Root` (Go 1.24) is available. That would be a genuinely valuable contribution if you'd like to pick it up.

---

### Things I checked that are fine

- `filepath.ToSlash` is correct on Windows — both operands convert consistently. It just isn't needed, since the existing comparison is already separator-consistent.
- With the patch applied to current `main`, `go test ./pkg/gofr/http -run 'Static|isWithin|isRestricted'` passes. Consistent with it being a no-op.
- The patch applies cleanly to today's `main` despite the stale base point.
- On the Windows permission failures you mentioned: those are Unix `0000` mode semantics not being enforced on Windows, unrelated to this change.

aryanmehrotra:
Closing this one — per the review above, the change has no behavioural effect: `directoryName` and `absPath` are both already `filepath.Clean`ed by the `filepath.Abs`/`filepath.Join` calls at their assignment sites, so the added `Clean`/`ToSlash` are no-ops on every input the call site can produce (0 of 66 directory x URL combinations differ from the current code). The three behaviours listed in the description already hold on `development` today. It also adds a `gocritic importShadow` lint failure, since the local `path` variable shadows the imported `path` package that `ServeHTTP` uses.

That said — you were looking at the right function. There **is** a live bug in `isWithinDirectory`, just not one that path cleaning can reach: #3855. A symlink inside the served directory pointing outside it is served, because lexical containment asks "does this string sit under that string" while the real question is "is this file reachable by walking down from the served directory". The fix is `os.OpenRoot` at registration (we pin `go 1.26.0`, so `os.Root` is available), and `isWithinDirectory` disappears in the process.

That issue is unassigned and would be a genuinely useful contribution if you'd like to take it. If you do, please base the branch on `development` rather than `main` — see `CONTRIBUTING.md`.

Thanks for taking the time to look at this area.

PR #4055, opened 2026-08-25
fix(pubsub/google): guard shared subscription maps against concurrent writes (#4054)

### LINKED ISSUE #4054: fatal error: concurrent map writes in Google Pub/Sub Subscribe when subscribing to multiple topics
**Describe the bug**
A gofr app that subscribes to two or more topics using the Google Pub/Sub backend crashes at startup with a hard `fatal error: concurrent map writes` (unrecoverable). Subscribing to a single topic works fine. It reproduces deterministically when run with the `-race` detector.

**To Reproduce**

1. The code is

```go
package main

import "gofr.dev/pkg/gofr"

func main() {
	app := gofr.New()

	noop := func(*gofr.Context) error { return nil }

	app.Subscribe("topic-a", noop)
	app.Subscribe("topic-b", noop) // two (or more) topics is enough to trigger it

	app.Run()
}
```

Config (Google backend, against the Pub/Sub

```
PUBSUB_BACKEND=GOOGLE
PUBSUB_PROJECT_ID=catalogue-service
 ``` 

Run with the race detector — a data race is reported deterministically; without it, the process panics intermittently at startup:

```
go run -race .
```

2. The error is

```
fatal error: concurrent map writes

goroutine [running]:
gofr.dev/pkg/gofr/datasource/pubsub/google.(*...)
    .../gofr.dev@v1.57.0/pkg/gofr/datasource/pubsub/google/google.go:211
gofr.dev/pkg/gofr.(*SubscriptionManager).handle(...)
    .../gofr.dev@v1.57.0/pkg/gofr/subscriber.go:52
gofr.dev/pkg/gofr.(*SubscriptionManager).start(...)
    .../gofr.dev@v1.57.0/pkg/gofr/subscriber.go:41
gofr.dev/pkg/gofr.(*App).startSubscriptions
    .../gofr.dev@v1.57.0/pkg/gofr/gofr.go:207
golang.org/x/sync/errgroup.(*Group).Go.func
```
                                                                                                              
**Expected behavior**                                                                                                  
Subscribing to multiple topics should start each subscriber without racing on shared state — the app should run normally instead of crashing with `concurrent map writes`.

 **Screenshots**
N/A.

**Environments (please complete the following information)
- OS: macOS (arm64) — appears backend/OS independent
- gofr version: v1.57.0 (also reproduced on v1.58.0 and v1.59.0)
- go version: 1.26

**More description**                                                                                                   
- Reducing the app to a single `app.Subscribe(...)` makes the crash / data race disappear, which points to concurrency across topic subscribers rather than anything in the handler.
- `startSubscriptions` (`pkg/gofr/gofr.go`) starts one subscriber goroutine per topic concurrently via `errgroup`, and each loops calling `Subscribe(ctx, topic)`. In `Sub/pubsub/google/google.go`), the shared`subStarted` map is read (v1.59.0 L171) and written (L211), and `receiveChan[topic]` is read in the `select` (L215); the
stack trace points at the `subStarted` write (L211) as the panic site.

### PR DESCRIPTION
**Pull request by Suhas-zs**
**Description:**

- Fixes #4054.
- A GoFr application that subscribes to more than one topic on the Google Pub/Sub backend crashes at boot with `fatal error: concurrent map writes`. `startSubscriptions` launches one subscriber goroutine per topic (via `errgroup`), and each goroutine loops over `Subscribe(ctx, topic)`. Inside the Google client, `Subscribe` read and wrote the shared `subStarted` and `receiveChan` maps **without holding the existing mutex**, so concurrent subscribers race on those maps.
- This PR guards every access to the shared maps with `g.mu`:
  - Extracted the lazy subscription start into `startTopicSubscriber`, which uses a read lock for the "already started" check and a write lock when populating both maps together.
  - The receiver goroutine now captures the channel in a local variable instead of re-reading the shared map, and delivers via `select { case ch <- m: case <-ctx.Done(): }`. It no longer holds the lock during a blocking channel send (which could stall other subscribers) and no longer leaks on shutdown.
  - `Close()` now takes the lock before ranging over `receiveChan`.
- **Benefit:** multi-topic subscribers now boot and run reliably; the fix is contained entirely to the GCP backend and does not change any public API or behavior for existing single-topic users.

- None. No exported signatures or behavior change; this is an internal concurrency-safety fix.

**Additional Information:**

- No new dependencies. Uses the existing `sync.RWMutex` already present on `googleClient`.
- Added `TestGoogleClient_Subscribe_ConcurrentTopics`, which subscribes to multiple topics concurrently (mirroring `startSubscriptions`) using the in-memory `pstest` server already used by the suite. The test **fails under `go test -race` on the old code** (`WARNING: DATA RACE`) and passes with the fix. Verified with `go test -race` across repeated runs; `golangci-lint` is clean on the changed code.

**Checklist:**

- [x] I have formatted my code using  `goimport`  and  `golangci-lint`.

### CODE REVIEW

File: pkg/gofr/datasource/pubsub/google/google.go [hunk 1]
```diff
@@ -236,6 +198,66 @@ func (g *googleClient) Subscribe(ctx context.Context, topic string) (*pubsub.Mes
...
+	subscription, err := g.getSubscription(ctx, t)
+	if err != nil {
+		return nil, err
+	}
+
+	start := time.Now()
+
+	receiveChan = make(chan *pubsub.Message)
+
+	g.mu.Lock()
```
Umang01-hash, line 232: Non-blocking: this is double-checked locking without a recheck under the write lock — two concurrent `Subscribe` calls for the *same* topic could both pass the RLock check and both start a receiver + overwrite the channel. Not reachable via gofr (one sequential goroutine per topic) and both writes are serialized by `g.mu` so there's no map panic, but a quick `if _, ok := g.subStarted[topic]; ok { return g.receiveChan[topic], nil }` right after `g.mu.Lock()` would make same-topic safety self-evident.

pkg/gofr/datasource/pubsub/google/google.go
```diff
@@ -369,6 +391,9 @@ func (g *googleClient) Close() error {
 		return g.client.Close()
 	}
 
+	g.mu.Lock()
```
Umang01-hash, line 404: Nit / pre-existing (out of scope for #4054): closing these channels while a receiver goroutine may still be in its `select`-send can panic with send-on-closed-channel during a shutdown race. The old `Close` had the same shape, so not this PR's job — flagging as a possible follow-up.

Umang01-hash (dismissed):
Verified locally, and this is a real fix. I reproduced the crash by running the new test against the base code with `-race` — deterministic `WARNING: DATA RACE` + FAIL — and it's clean at this head (the new test 5x plus the full package under `-race`, ~37s). Every `subStarted`/`receiveChan` access is now under `g.mu`, no exported signature changed, and golangci `--new-from-rev` is clean. The regression test genuinely catches the bug (fails at base, passes at head) and uses the real pstest fake, not a mock.

Nice secondary cleanup too: the old `processMessage` held `g.mu` across a blocking `receiveChan <- m`, so the new local-channel send with a `ctx.Done()` escape also removes a lock-held-during-send bottleneck.

Approving. Two non-blocking notes inline.

pkg/gofr/datasource/pubsub/google/google.go
Suhas-zs: Good catch, thanks. Added the recheck of `subStarted[topic]` right after `g.mu.Lock()` so a same-topic race reuses the existing channel instead of starting a second receiver. It's not reachable through gofr's one-goroutine-per-topic model today, but it makes the invariant self-evident. Didn't add a test since the duplicate writes are already serialized by `g.mu` (no `-race` signal, and a topic-keyed length assertion can't distinguish the case).

pkg/gofr/datasource/pubsub/google/google.go
Suhas-zs: Agreed, and thanks for scoping it as pre-existing. Keeping it out of this PR to stay focused on #4054; I'll open a follow-up issue to make `Close` shutdown-safe (e.g. cancel the receiver context and drain before closing the channels).

pkg/gofr/datasource/pubsub/google/google.go
```diff
@@ -369,6 +401,9 @@ func (g *googleClient) Close() error {
...
```
Umang01-hash, line 404: Nice — locking Close fixes the map-iteration-vs-write race. One pre-existing thing this doesn't close (fine for this PR, worth a follow-up): Close closes the receiveChan channels but doesn't cancel the receiver goroutine's ctx, so if a message lands during shutdown, processMessage's `receiveChan <- m` is a send on a closed channel -> panic. Cancelling the receiver ctx (or the client) before closing the channels would close that window.

pkg/gofr/datasource/pubsub/google/google.go
```diff
@@ -168,51 +168,13 @@ func (g *googleClient) Subscribe(ctx context.Context, topic string) (*pubsub.Mes
...
-
-		go func() {
-			err = subscription.Receive(ctx, processMessage)
-			if err != nil {
-				g.logger.Errorf("error getting a message from google: %s", err.Error())
-			}
-		}()
-
-		g.subStarted[topic] = struct{}{}
+	receiveChan, err := g.startTopicSubscriber(ctx, topic, &end)
```
Umang01-hash, line 171: Nit (pre-existing, not introduced here): only the first Subscribe caller's &end reaches the long-lived goroutine, so Time in the SUB debug log is 0 for every later call on the same topic. The pointer refactor preserves the old closure's behaviour (and there's no data race — the channel gives you happens-before for the first read), so it's fine to leave; just noting it's dead for 2nd+ messages.

Umang01-hash (approved):
LGTM — read the whole thing at head and verified it locally, didn't just trust the description.

- The fix is complete, not just the reported instance: I grepped every receiveChan/subStarted access and they all go through g.mu now (RLock read + Lock write in startTopicSubscriber, Lock in Close). Subscribe uses the returned local channel, not the map. Double-checked locking is correct — the expensive getTopic/getSubscription runs outside the lock, then a recheck under the write lock reuses the winner's channel so a racing same-topic caller never starts a second receiver.
- Fail-on-revert: with the base google.go, the new test throws WARNING: DATA RACE and fails under -race; with the fix it passes. So the test genuinely reproduces #4054 and isn't vacuous. Full package under -race is clean (no regressions), gofmt/vet/build clean, golangci --new-from-rev = 0 issues.

Two pre-existing things I noticed while in here — neither blocks this PR, both are follow-up material, flagged inline.

aryanmehrotra (approved):
Approving. I reproduced #4054 and re-ran the fix rather than reading the diff, and the core change is correct. One thing came out of running the suite repeatedly that I think is worth acting on before or just after merge, plus a correction to something in the earlier review thread.

**The fix does what it says.** On `development`, the new test throws two `WARNING: DATA RACE` and fails, at `google.go:201`/`:215` (`subStarted`) and `:211` (`receiveChan`) — the exact maps. At this head the same test is clean over 5 runs, and the full package under `-race` is clean. Double-checked locking is right: the expensive `getTopic`/`getSubscription` stay outside the lock, and the recheck reuses the winner's channel. I drove 20 concurrent same-topic callers and the recheck branch executed 13 times with exactly one receiver started each time.

| check | `development` | this PR |
|---|---|---|
| new test, `-race` | 2 races, FAIL | clean ×5 |
| full package, `-race` | clean | clean |
| `Subscribe` coverage | 95.0% | 100.0% |
| `Close` coverage | 100% | 100% |
| package coverage (3 runs) | 74.3–74.5% | 74.7–74.9% |
| `gofmt` / `vet` / build | clean | clean |
| lint | SA1019 on the deprecated pubsub import | same, line untouched |

Merges cleanly with current `development`; the effective diff is exactly the two intended files.

---

**One thing to fix: this PR makes a pre-existing test race reproduce every time.**

Running the full package with `-race -count=2`, `TestGoogleClient_Subscribe_Success` races on **5 of 5** runs at this head and **0 of 5** on `development`. The race is in the test, which this PR does not touch:

```go
var msg *pubsub.Message

go func() {
    msg, err = g.Subscribe(ctx, topic)   // goroutine writes msg and err
}()

time.Sleep(500 * time.Millisecond)
result = topicObj.Publish(...)
_, err = result.Get(t.Context())         // main goroutine writes err
...
time.Sleep(1 * time.Second)
require.NoError(t, err)                  // main goroutine reads
```

`msg` and `err` are shared across the goroutine boundary with nothing but `time.Sleep` for synchronisation. It is latent on `development` because the old `processMessage` held `g.mu` across the blocking send, which slowed delivery enough that the interleaving rarely showed. Removing that lock — which is the right change — shifts the timing so the detector now sees it on every run.

CI doesn't pass `-race` anywhere in `go.yml`, so this will not turn the checks red. The cost is that `go test -race ./pkg/gofr/datasource/pubsub/google/` becomes a guaranteed failure after this merges, which is precisely the signal a future concurrency bug in this package would need. I'd rather not blunt it in the same PR that sharpens it.

Verified fix — with this, 5 of 5 runs are clean, 0 races, 0 failures:

```go
var (
    msg    *pubsub.Message
    subErr error
)

done := make(chan struct{})

go func() {
    defer close(done)

    msg, subErr = g.Subscribe(ctx, topic)
}()

time.Sleep(500 * time.Millisecond)

result = topicObj.Publish(t.Context(), &gcPubSub.Message{Data: message})
_, err = result.Get(t.Context())
require.NoError(t, err)

<-done   // replaces the second sleep and gives msg/subErr a happens-before edge

require.NoError(t, subErr)
require.NotNil(t, msg)
```

Happy either way on whether that rides along here or lands as a follow-up, as long as it isn't lost.

---

**Correction to the note on `&end`.** The earlier comment says there is no data race there because the channel provides happens-before. The channel does order the write against the *reader*, but it does not order two `processMessage` invocations against each other, and `subscription.Receive` runs the callback on several goroutines. With a single drainer — exactly gofr's one-goroutine-per-topic model — publishing a batch reproduces a write/write race on `*end = time.Since(start)`:

- `development`: 5 of 5 runs, at `google.go:186`
- this PR: 4 of 5 runs, at `google.go:249`

So it is genuinely pre-existing and unchanged by this PR, and correctly out of scope — but it is a real race rather than a benign staleness, and the follow-up issue should describe it that way. The "`Time` is 0 for later callers" observation still stands on top of it.

**On leaving the recheck untested** — your reasoning checks out, and I verified it: with the recheck deleted, both a length assertion and your `Subscribe_ConcurrentTopics` test still pass, because the second writer overwrites the same key and the map length is 1 either way. A channel-identity assertion does distinguish the case, but only in about 2 runs in 5, because `getSubscription` is itself a check-then-create — 18 to 19 of 20 concurrent same-topic callers fail with `AlreadyExists` before they ever reach the recheck. That upstream TOCTOU, not the choice of assertion, is what makes the branch awkward to test. Worth capturing in the same follow-up, since it also means concurrent same-topic `Subscribe` still returns errors regardless of the recheck.

Worth noting that only the Snyk check has run on this head — no Go CI has executed here — so the results above are from running it locally.

Issue #3978, opened 2026-08-20
GoFr — Auth middleware exempts the entire /.well-known prefix, leading to authentication bypass


lalalala5678:
# GoFr — Authentication bypass via /.well-known prefix exemption in AuthMiddleware leads to anonymous access to protected routes and health data

## Summary

GoFr's unified authentication middleware (`AuthMiddleware`) calls `isWellKnown(r.URL.Path)` before extracting any authentication header, and `isWellKnown` is implemented as `strings.HasPrefix(path, "/.well-known")`. This exempts the entire `/.well-known` prefix — not just an intended probe path — so the built-in `/.well-known/health` endpoint (which the official documentation acknowledges may expose dependency service hosts and statistics) remains anonymously reachable even after Basic Auth, API Key, or OAuth is enabled. In addition, any user-registered route whose path merely starts with `/.well-known` (e.g. `/.well-knownprivate` — no slash is required to match the prefix) silently skips authentication, and because the same `isWellKnown` check is reused by the rate limiter, these paths bypass rate limiting as well.

**Weakness:** CWE-306: Missing Authentication for Critical Function (secondary: CWE-200: Exposure of Sensitive Information)

## Root Cause

In GoFr v1.58.0:

1. `pkg/gofr/http/middleware/validate.go:5-7` — `isWellKnown` is implemented as `strings.HasPrefix(path, "/.well-known")`. It is a raw prefix match that does not require the prefix to be followed by `/`, so paths such as `/.well-knownprivate` also match.
2. `pkg/gofr/http/middleware/auth.go:49-52` — `AuthMiddleware` evaluates `isWellKnown(r.URL.Path)` before `ExtractAuthHeader` and, on a match, passes the request directly to `handler.ServeHTTP`, skipping authentication entirely. All official auth mechanisms (Basic Auth, API Key, OAuth) share this middleware, so every one of them is bypassed for matching paths.
3. `pkg/gofr/http/middleware/rate_limiter.go:133-136` — the same `isWellKnown` check is reused in the rate limiter, so matching paths also skip rate limiting.

## Impact

- An anonymous attacker can read the built-in `/.well-known/health` endpoint of applications that have authentication enabled, leaking internal dependency topology (database/cache hosts, status, statistics).
- Developers cannot protect any route under `/.well-known*` using GoFr's official authentication capability; business routes registered with that prefix — knowingly or unknowingly — are exposed directly.
[...]

PR #2565, opened 2025-11-20
Restrict /.well-known auth bypass to alive endpoint

Saniya777-byte:
**Description:**

- Restrict authentication bypass under `/.well-known` to only the alive endpoint.
- Previously, all `/.well-known/*` endpoints skipped auth via `isWellKnown`, which made `/.well-known/health` publicly accessible.
- This PR changes the auth middleware to bypass auth only for `service.AlivePath` (`/.well-known/alive`) and keeps `service.HealthPath` (`/.well-known/health`) protected.
- Updates [TestAuthMiddleware](cci:1://file:///c:/Users/khati/OneDrive/Desktop/gofr/pkg/gofr/http/middleware/auth_test.go:13:0-58:1) to:
  - Assert that `/.well-known/alive` is accessible without authentication.
  - Assert that `/.well-known/health` requires authentication (success and failure cases covered).

(Fixes the issue #2562.)

**Breaking Changes (if applicable):**

- None.
- This change only affects auth behavior for `/.well-known/health` by enforcing authentication where it was previously missing.

**Additional Information:**

-**Checklist:**

- [ ] I have formatted my code using `goimport` and `golangci-lint`.
- [x] All new code is covered by unit tests ([TestAuthMiddleware](cci:1://file:///c:/Users/khati/OneDrive/Desktop/gofr/pkg/gofr/http/middleware/auth_test.go:13:0-58:1) updated for the new behavior).
- [x] This PR does not decrease the overall code coverage.
- [ ] I have reviewed the code comments and documentation for clarity.

PR #4099, opened 2026-08-26
fix(middleware): require a path separator in the well-known auth exemption

rmdly:
Refs #3978

`isWellKnown` used a raw prefix match, so any path merely starting with the characters `/.well-known` was exempted from authentication and from rate limiting, including user-registered routes such as `/.well-knownprivate`, since no separator was required.

This requires a path boundary instead:

```go
return path == "/.well-known" || strings.HasPrefix(path, "/.well-known/")
```

Every legitimate `/.well-known/*` path behaves exactly as before, so this is not a breaking change. All framework routes under that prefix (`/.well-known/alive`, `/.well-known/health`, the Swagger and GraphQL playground routes) use a separator and are unaffected.

**What this does not fix, by design.** User routes registered under the prefix with a separator, such as `app.GET("/.well-known/custom", h)`, stay auth-exempt after this change. That is #3978's remediation item 1 (exact or allowlist matching), and narrowing it would withdraw the exemption from `/.well-known/acme-challenge/*` and friends, contradicting the contract documented at `docs/quick-start/introduction/page.md:120` ("All `/.well-known/*` paths are auth-exempt by default"). So this PR uses `Refs` rather than `Fixes`, to avoid closing #3978 as though item 1 were done.

**Also deliberately out of scope:** whether `/.well-known/health` should require authentication at all. That is a breaking change for anyone using it as a readiness probe, which is why #2565 was closed, so it seems better left as a separate decision.

Tests:
- `Test_isWellKnown` gains cases for the bare `/.well-known` root, a prefixed route with no separator, a prefixed route with a sub-path, and the prefix appearing mid-path.
- `TestAuthMiddleware` gains a case asserting `/.well-knownprivate` returns 401 when auth fails, and one asserting it passes through with the auth context set when auth succeeds. The test previously short-circuited on the same loose prefix, so that guard now requires the separator too.

- `TestRateLimiter_WellKnownPrefixIsRateLimited` covers the second caller at `rate_limiter.go:133`: a prefixed path consumes the bucket and gets a 429, while `/.well-known/alive` stays exempt once the bucket is empty.

`isWellKnown` is at 100% statement coverage and the middleware package total is unchanged at 91.3%. Reverting `validate.go` to the loose prefix turns five cases red across all three test functions.

Docs: `docs/advanced-guide/authentication/page.md` quoted the pre-fix implementation verbatim, so it now describes the separator requirement instead.

Also noted while here: `isFrameworkRoute` in `pkg/gofr/tools.go:333` uses the same loose prefix. The impact there is only that a similarly-named route would be hidden from the MCP tool list rather than exposed, so I have left it alone. Happy to follow up separately if you want it tightened.


aryanmehrotra (changes requested):
Good catch, and a well-scoped first contribution — thanks for it.

I verified this rather than reading it: built the app below against `development` and against this
branch, ran both in containers, and drove them over HTTP.

```go
app := gofr.New()
app.EnableBasicAuth("admin", "s3cret")
h := func(ctx *gofr.Context) (any, error) { return "SECRET-DATA", nil }
app.GET("/private", h)
app.GET("/.well-knownprivate", h)
app.GET("/.well-known/custom", h)
```

| Request, no credentials | development | this branch |
| --- | --- | --- |
| `/private` | 401 | 401 |
| `/.well-knownprivate` | **200 `SECRET-DATA`** | 401 |
| `/.well-known/custom` | 200 | 200 |

The bypass is real, it is closed, and legitimate well-known paths are unaffected.

I also pushed on the cases the tests do not cover, since prefix fixes often still leak through path
normalisation — they do not:

```
/.well-known/../private     -> 401     traversal does not bypass
/.well-known/..%2fprivate   -> 401
/.well-known%2fprivate      -> 404
//.well-knownprivate        -> 401
/.WELL-KNOWN/custom         -> 401     case-sensitive, so stricter
```

And your tests genuinely guard the fix — I reverted `validate.go` to the loose prefix and four cases
failed across both test functions, rather than the suite staying green.

Two other claims I checked and can confirm: every framework route under the prefix uses a separator
(`alive`, `health`, `graphql/ui`, `swagger`, `{name}`, the openapi file), so nothing regresses; and
`isFrameworkRoute` in `tools.go:333` fails in the safe direction — it *excludes* routes from the MCP
tool surface, so a loose match hides a route rather than exposing one. Right call to leave it, and
right call to mention it rather than widen the diff.

**One thing before I approve.** `isWellKnown` has two call sites and only one is tested:

```
auth.go:49            <- two new tests
rate_limiter.go:133   <- none
```

The rate-limiter exemption is arguably the more reachable half, because it needs no registered route
at all: before this change, any request to `/.well-knownXXXX` skipped rate limiting even though every
one of them 404s. Your fix covers it — same function — but nothing asserts it, so a later refactor
could reintroduce the loose match there and the suite would stay green.

Could you add a case to the rate limiter tests mirroring the auth ones — `/.well-knownprivate` is rate
limited, `/.well-known/alive` is not? Both call sites tested and this is good to go.

Also deliberately leaving alone, agreeing with your scoping: whether `/.well-known/health` should
require auth at all is a breaking change and belongs in its own discussion, as #2565 concluded.

rmdly:
Added `TestRateLimiter_WellKnownPrefixIsRateLimited`, which covers the `rate_limiter.go:133` call site: `/.well-knownprivate` consumes the bucket and gets a 429 on the second request, and `/.well-known/alive` still returns 200 after that bucket is empty. It also asserts the `app_http_rate_limit_exceeded_total` counter, matching `TestRateLimiter_GlobalLimit`.

Reverting `validate.go` to the loose prefix now fails five cases across three test functions rather than four across two, so both call sites are guarded.

Thanks for the detailed review. The normalisation cases you drove are a better check than anything I had, and good to have `//.well-knownprivate` and the encoded variants confirmed.

Umang01-hash (changes requested):
The auth/rate-limiter fix is correct — good catch, and reusing the single `isWellKnown` helper is the right shape.

The same bug still lives in `pkg/gofr/tools.go:333` though: `isFrameworkRoute` does `strings.HasPrefix(path, "/.well-known")` — the exact loose match you're fixing here. A route like `/.well-knownstuff` would be treated as a framework route and hidden from MCP tool exposure. It fails safe (hides, doesn't expose), so not urgent, but it's the same class — worth routing it through the same separator check or reusing `isWellKnown`. Fine as a follow-up if you'd rather keep this PR tight.

rmdly:
Thanks @Umang01-hash. I have opened #4102 for the `isFrameworkRoute` case rather than widening this one, since @aryanmehrotra had suggested keeping this PR to the middleware.

It mirrors the same separator check rather than exporting `isWellKnown`, because the two sit in different packages and widening the API for a single call site felt worse than the small duplication. Easy to switch to a shared helper if you would prefer that.

`isFrameworkRoute` had no tests, so #4102 adds the first ones.

This branch is also now up to date with `development`.

Umang01-hash (commented):
Hey @rmdly ! Logic and tests look good — the rate-limiter case is exactly the one that was missing, and reverting the fix does turn them red. Only thing holding it is lint (**Code Quality's failing**). Left a few suggestions to get it green, then it's good from my side.

pkg/gofr/http/middleware/validate.go
```diff
@@ -3,5 +3,5 @@ package middleware
 import "strings"
 
 func isWellKnown(path string) bool {
-	return strings.HasPrefix(path, "/.well-known")
+	return path == "/.well-known" || strings.HasPrefix(path, "/.well-known/")
 }
```
Umang01-hash, line 9: `goconst` wants the repeated `/.well-known` as a const:

```suggestion
const wellKnownPrefix = "/.well-known"

func isWellKnown(path string) bool {
	return path == wellKnownPrefix || strings.HasPrefix(path, wellKnownPrefix+"/")
}
```

pkg/gofr/http/middleware/rate_limiter_test.go
```diff
@@ -143,6 +143,40 @@ func TestRateLimiter_SkipHealthEndpoints(t *testing.T) {
...
+		Burst:             1,
+		PerIP:             false,
+	}
+
+	handler := RateLimiter(config, metrics)(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+
+	// A path that only shares the well-known prefix is not exempt, so it consumes the bucket
+	req := httptest.NewRequest(http.MethodGet, "/.well-knownprivate", http.NoBody)
```
Umang01-hash: `noctx` flags the new `httptest.NewRequest` calls — CI only checks new lines, so the existing ones in this file are fine. `t.Context()` clears it:

```suggestion
	req := httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/.well-knownprivate", http.NoBody)
```

pkg/gofr/http/middleware/rate_limiter_test.go
```diff
@@ -143,6 +143,40 @@ func TestRateLimiter_SkipHealthEndpoints(t *testing.T) {
...
+	rr := httptest.NewRecorder()
+	handler.ServeHTTP(rr, req)
+	assert.Equal(t, http.StatusOK, rr.Code, "First request should succeed")
+
+	req = httptest.NewRequest(http.MethodGet, "/.well-knownprivate", http.NoBody)
```
Umang01-hash: same here:

```suggestion
	req = httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/.well-knownprivate", http.NoBody)
```

pkg/gofr/http/middleware/rate_limiter_test.go
```diff
@@ -143,6 +143,40 @@ func TestRateLimiter_SkipHealthEndpoints(t *testing.T) {
...
+	rr = httptest.NewRecorder()
+	handler.ServeHTTP(rr, req)
+	assert.Equal(t, http.StatusTooManyRequests, rr.Code, "Prefixed path should be rate limited")
+
+	assert.Equal(t, 1, metrics.GetCounter("app_http_rate_limit_exceeded_total"))
+
+	// Genuine well-known endpoints stay exempt even once the bucket is empty
+	for i := 0; i < 5; i++ {
+		req = httptest.NewRequest(http.MethodGet, "/.well-known/alive", http.NoBody)
```
Umang01-hash: and this one:

```suggestion
		req = httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/.well-known/alive", http.NoBody)
```

pkg/gofr/http/middleware/validate.go
Umang01-hash, line 9:
```suggestion
const wellKnownPrefix = "/.well-known"

func isWellKnown(path string) bool {
	return path == wellKnownPrefix || strings.HasPrefix(path, wellKnownPrefix+"/")
}
```
`goconst` is failing here — pulling the literal into a const clears it (and reads a bit nicer).

pkg/gofr/http/middleware/rate_limiter_test.go
Umang01-hash:
```suggestion
	req := httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/.well-knownprivate", http.NoBody)
```

pkg/gofr/http/middleware/rate_limiter_test.go
Umang01-hash:
```suggestion
	req = httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/.well-knownprivate", http.NoBody)
```

pkg/gofr/http/middleware/rate_limiter_test.go
Umang01-hash:
```suggestion
		req = httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/.well-known/alive", http.NoBody)
```
`noctx` flags the three new `httptest.NewRequest` calls (the existing ones are grandfathered by only-new-issues).

Umang01-hash (changes requested):
The logic is spot on and the tests cover both the auth and rate-limiter callers nicely — `/.well-knownprivate` going through auth and consuming the bucket is exactly right. Only thing holding it is Code Quality, which is red: `goconst` on the literal and `noctx` on the new test requests. Both are the ones from yesterday. Left suggestions for each, then it's good to go.

rmdly:
Thanks @Umang01-hash, applied all four suggestions: `wellKnownPrefix` as a const in `validate.go`, and `httptest.NewRequestWithContext(t.Context(), ...)` for the three new requests in the rate limiter test.

`golangci-lint` is clean on the package now, and reverting the check to the loose prefix still fails five cases across the three test functions, so the const refactor did not weaken anything.

Also confirmed #4102 covers the `isFrameworkRoute` half, thanks for the approval there.

Umang01-hash (dismissed):
LGTM. Verified locally: the predicate is right, and reverting it to the loose `HasPrefix` turns all three tests red — the helper unit test plus both caller tests (`TestAuthMiddleware` 401s the prefixed path, `TestRateLimiter_WellKnownPrefixIsRateLimited` consumes the bucket while the real endpoint stays exempt), so the coverage isn't vacuous. gofmt/vet/build clean and `golangci-lint --new-from-rev` is 0 issues (goconst/noctx from earlier rounds resolved).

Also checked the traversal angle — `/.well-known/../admin` passes `isWellKnown` but the router is `SkipClean(true)` with normalized routes, so it 404s instead of reaching `/admin`. And the sibling `isFrameworkRoute` in tools.go carries the same loose prefix — nice that it's already split into #4102 rather than widened here.

aryanmehrotra:
The predicate is right, the tests are not vacuous, and I could not break it — I planned this
ticket independently before opening the diff, then measured both trees. **The code is right
and I am not asking you to change it.**

Two things that ship *alongside* it are wrong, and both are much cheaper to fix here than
after merge — findings 1 and 2, a line each. Finding 3 is a follow-up, not for this branch.

## The shape

An independent plan for #3978 landed on the same shape you built — one predicate, both call
sites, no new API, no exported helper:

```
   AuthMiddleware (auth.go:49)     RateLimiter (rate_limiter.go:133)
              |                              |
              +--------------+---------------+
                             v
                +----------------------------+
                | isWellKnown  validate.go   |
                | p == "/.well-known" ||     |
                | HasPrefix(p, prefix + "/") |
                +----------------------------+
                     |true             |false
                exempt (probe)     auth + limit
```

The bare-root arm is deliberate and stays — treating the prefix root as part of the probe
namespace is a framework call, not an oversight, and I have recorded it as one. Every other
structural choice here is the one I'd have made, including two I'd have got wrong without
you: see the bottom.

## 1. `Fixes #3978` closes more than this changes

`/.well-known/<user route>` is still anonymous after the fix. Reproduced on this branch —
Basic Auth enabled, no credentials, paths sent verbatim over a raw socket (curl normalises
client-side, so curl cannot show this):

```
/.well-knownprivate  -> 401                          <- your fix, working
/.well-known/custom  -> 200 {"data":"SECRET-DATA"}   <- app.GET("/.well-known/custom", h)
```

That is #3978's impact bullet 3 ("developers cannot protect any route under
`/.well-known*`") and its remediation item 1 (exact/allowlist matching). Declining both is
the right call — narrowing to an allowlist withdraws the exemption from
`/.well-known/acme-challenge/*` and friends, which breaks the contract published at
`docs/quick-start/introduction/page.md:120` ("All `/.well-known/*` paths are auth-exempt by
default"), and that is the kind of change this repo reverts rather than ships. But the issue
should not auto-close believing it got item 1.

**Please change:** either switch `Fixes #3978` to `Refs #3978`, or keep `Fixes` and add a
line to the description saying user routes under `/.well-known/` stay exempt by design, with
a follow-up issue for the narrowing. Left as is, the issue closes on merge recording item 1
as done.

## 2. The auth guide now documents the wrong gate

`docs/advanced-guide/authentication/page.md:20` quotes the pre-fix implementation verbatim:

> the middleware's `isWellKnown` check uses `strings.HasPrefix(path, "/.well-known")`

It is the one page a user reads to decide what is exempt, so it should not describe the match
this PR removes — and a docs fix that does not travel with the code fix rarely gets written
at all.

**Please change:** one line on that page, saying the check requires a `/` separator after the
prefix. It belongs in this PR.

## 3. Nothing in the suite crosses the router

Every new case runs the middleware directly, so what actually makes this fix safe is
untested: `(*Router).ServeHTTP` calls `normalizePath` (`pkg/gofr/http/router.go:88`) before
either matcher runs, and mux builds its middleware chain inside its own `ServeHTTP` — so auth
always sees the `path.Clean`'d path. I measured it and it holds:

```
/.well-known/../private    -> 401   (reaches auth as /private)
/.well-known/..%2fprivate  -> 401
/.well-known/.%2e/private  -> 401
/.WELL-KNOWN/alive         -> 401   (case-sensitive, so stricter)
//.well-knownprivate       -> 401   normalises INTO /.well-knownprivate
```

That last one is the interesting one: the double-slash form normalises *into* the exempted
spelling, which is exactly why the predicate has to be right for the canonical path. Worth
two rows through `router.ServeHTTP` with auth enabled — `//.well-knownprivate` and
`/.well-known/../private`, both 401 — so the next person to touch `normalizePath` learns it
from a test rather than from an incident. Not a blocker; the behaviour is correct today, and
a follow-up is fine.

## What the tests do and do not cover

| Dimension | In this PR |
|---|---|
| correctness | 4 predicate rows, 2 auth rows, 1 rate-limiter test — and reverting `validate.go` turns 5 cases red across all 3 functions. Verified, not vacuous. |
| reliability / concurrency / scale | n/a — pure predicate, nothing added |
| security | **not covered** — no traversal, encoded, double-slash or case variant; nothing above the predicate (finding 3) |
| regression | `go test ./pkg/... -short` matches `development` exactly (31 ok; the single `pkg/gofr` failure is a port-2121 collision on my machine, present on both trees) |

## Differences I am deliberately not flagging

- Keeping bare `/.well-known` exempt — a framework call on where the probe namespace starts;
  no framework route depends on it either way.
- `const wellKnownPrefix` — style, and it is what was asked for; no behavioural difference.
- Tightening the harness guard to `/.well-known/` versus a per-case `exempt` field — both
  close the vacuous-assertion hole identically.
- Splitting `isFrameworkRoute` into #4102 rather than exporting the helper — right call;
  sharing would mean exporting an unexported middleware symbol across packages for one
  caller, and the MCP side fails safe either way.
- Leaving `/.well-known/health` public — settled, and the body is already redacted.
- Asserting `app_http_rate_limit_exceeded_total` in the new test — matches
  `TestRateLimiter_GlobalLimit`; the right neighbour to copy.

## What I'd have missed

Two things my independent plan got wrong and this PR got right:

- **`httptest.NewRequestWithContext(t.Context(), …)` in the new test.** `noctx` and
  `usetesting` have no `_test.go` exclusion and CI lints with `only-new-issues: true`, so the
  surrounding old-style lines are grandfathered while a new one in the same file goes red.
  Copying the neighbour would have failed the gate.
- **Not adding a router-level swagger guard.** I had one planned; it would have guarded
  nothing, since every swagger route already carries a separator — and it would have needed a
  `./static/openapi.json` fixture to register the catch-all at all.

---

Sorry to send you round again on a diff whose code I am not questioning — but the issue
closes on merge and the docs page is read by everyone configuring auth, and neither is
practical to fix afterwards. A line each and this is good to go from me. Finding 3 is a
follow-up; open it whenever, or leave it and I will.

rmdly:
@aryanmehrotra when you get a moment, would you mind taking another look? The rate limiter test you asked for went in on 28 August (`TestRateLimiter_WellKnownPrefixIsRateLimited`, covering the `rate_limiter.go:133` caller), and @Umang01-hash has approved since, after re-running the mutation check himself.

Your review from the 27th is still recorded as changes requested, so it is the only thing holding this one now.

CI is 15/15 green. GitHub also shows the branch as a couple of commits behind `development`. None of the new commits touch the files here, so there is no conflict, but I am happy to merge `development` in if you want it current. I have left it alone so far rather than risk churning the existing approval.

rmdly:
Thanks @aryanmehrotra, that is a genuinely useful review and both findings were fair.

**Finding 1.** Switched to `Refs #3978` so it will not auto-close, and added a paragraph to the description stating plainly that user routes under the prefix with a separator, such as `app.GET("/.well-known/custom", h)`, stay exempt after this change. It names remediation item 1 as not done and cites the `docs/quick-start/introduction/page.md:120` contract as the reason narrowing is being declined rather than deferred. You are right that leaving `Fixes` would have recorded item 1 as shipped.

**Finding 2.** Fixed in `docs/advanced-guide/authentication/page.md`. It quoted the pre-fix implementation verbatim, and now says the check requires a `/` separator after the prefix, with `/.well-knownprivate` as the example of what is no longer exempt. Agreed it belongs here rather than as a trailing docs PR.

**Finding 3.** Happy to open the follow-up for the router-level cases, and to write them, unless you would rather do it yourself. The `//.well-knownprivate` normalising into the exempted spelling is the part I had not considered, and I agree it should be a test rather than tribal knowledge. Say the word and I will file it.

Also merged `development` in, since the branch was showing as behind.

aryanmehrotra (approved):

PR #4102, opened 2026-08-28
fix(tools): require a path separator in the framework route check

rmdly:
Follow-up to #4099, as suggested by @Umang01-hash in review.

`isFrameworkRoute` used the same raw prefix match that #4099 fixes in the middleware, so a route whose path merely starts with the characters `/.well-known`, such as `/.well-knownstuff`, was treated as a framework route.

```go
return pathTemplate == "/favicon.ico" || pathTemplate == "/.well-known" || strings.HasPrefix(pathTemplate, "/.well-known/")
```

This mirrors the separator check in `isWellKnown` rather than exporting it, since the two live in different packages and widening the API for one call site seemed worse than the small duplication. Happy to change that if you would rather share a single helper.

Impact is lower than the middleware case: `isFrameworkRoute` fails safe, so a loose match hides a route from the MCP tool surface rather than exposing one. It is a correctness fix rather than a security fix.

`isFrameworkRoute` had no tests, so `Test_isFrameworkRoute` is new. It covers the favicon, the bare `/.well-known` root, the framework endpoints under the prefix, a prefixed route with and without a sub-path, and the prefix appearing mid-path. Reverting the check to the loose prefix fails the two prefixed-route cases.


Umang01-hash (approved):
Nice follow-up — this is the exact sibling of #4099, and between the two every `/.well-known` prefix check in the codebase now requires the separator. Ran the table test on the ref, passes, and Code Quality is green here. LGTM.

rmdly:
This one is approved by @Umang01-hash and CI is 15/15 green. Two small things in case it is waiting on either:

It shows as behind `development` by a few commits, though none of them touch `pkg/gofr/tools.go`, so happy to merge `development` in whenever that is useful.

Per the contributing guide it also needs a second reviewer, so flagging that in case it is sitting for want of one rather than anything in the diff.

aryanmehrotra (approved):
Approving. I reproduced this end-to-end rather than reading the predicate, and the fix does exactly what it claims with no collateral.

**The bug is real and user-visible.** On `development`, an ordinary application route whose path merely starts with the characters `/.well-known` is invisible to the agent surface *and* uncallable. Running your table's cases through a real `App` + `routerTools` — `List()` and `Call()`, not just the predicate:

| route registered | `development` | this PR |
|---|---|---|
| `/.well-knownstuff` | hidden, `Call` → `tool not found` | listed, `Call` → ok |
| `/.well-knownstuff/data` | hidden, `Call` → `tool not found` | listed, `Call` → ok |
| `/.well-knownprivate` | hidden, `Call` → `tool not found` | listed, `Call` → ok |
| `/.well-known/custom` | hidden | hidden (unchanged) |
| `/api/.well-known/alive` | listed | listed (unchanged) |
| `/favicon.ico`, `/.well-known`, `/.well-known/alive`, `/.well-known/{name}` | framework | framework (unchanged) |

Exactly three templates change classification, all of them application routes that were being swallowed. Every framework route is untouched — including the Swagger catch-all `/.well-known/{name}`, which was the one worth checking, since that template is what would have broken had the separator check been written as an exact-set match instead.

**The tests are load-bearing.** `pkg/gofr/tools.go` differs from `development` by exactly one line; reverting just that line while keeping `tools_test.go` fails `TEST[7]` and `TEST[8]`, the two prefixed-route cases, and nothing else. That matches your description exactly.

A note on the coverage framing rather than a criticism of it: `isFrameworkRoute` already reported 100% statement coverage *before* this PR, because it is a single return statement exercised indirectly through `mcp_test.go`. The statement metric cannot see what `Test_isFrameworkRoute` actually adds, which is branch coverage. The revert above is the real evidence that the test has teeth, which is why I ran it rather than quoting a percentage. I also measured package coverage three times per branch and it moves between 86.2% and 86.5% on `development` and on this PR alike, so any delta quoted at that granularity would be noise.

**Gates**, on head `55cb261`:

| check | result |
|---|---|
| `pkg/gofr` suite | pass, both branches |
| `isFrameworkRoute` / `specFor` | 100%, stable across runs |
| `gofmt` / `go vet` | clean |
| lint on `tools.go` / `tools_test.go` | zero issues, same as `development` |
| CI | 14 pass, 1 pending (Submodule v1.26), 0 failing |

**On sharing a single helper — keep the duplication.** Your instinct is right, though I would reach for a different reason than the one in the description. `pkg/gofr` already imports `pkg/gofr/http/middleware` and already calls `middleware.AuthHeaders()` at `tools.go:384`, so "different packages, don't widen the API for one call site" is not quite the obstacle it appears to be; there is precedent for exporting exactly this shape.

The stronger argument is that the two are going to diverge. #3978 item 1 is still open, and its remediation narrows `isWellKnown` to exact or allowlist matching over *request paths*, while `isFrameworkRoute` matches *route templates* and has to keep matching `/.well-known/{name}`. A shared helper would have to be pulled apart again as soon as that lands. Two short predicates in different domains, with opposite failure directions — a loose `isWellKnown` exempts authentication, a loose `isFrameworkRoute` only hides a tool — are better kept independent.

**Worth recording:** `docs/advanced-guide/mcp/page.md:47` already states that framework `/.well-known/*` probes are never exposed. The documentation already promised the separator; `development` simply did not implement it. So no docs change belongs in this PR — it brings the code in line with the docs rather than the other way round.

You have also already merged `development` in, so the branch is current and there is nothing outstanding on that front. Nothing blocking, and this is the second approval the contributing guide asks for.

PR #4103, opened 2026-08-29
fix(websocket): prevent panic on plain HTTP requests

dharmkoshiya74-collab:
## Description

Prevents a panic when a WebSocket route receives a normal HTTP request instead of a WebSocket handshake.

## Changes

- Safely check the WebSocket connection ID type assertion.
- Safely handle a nil WebSocket connection.
- Added a regression test for plain HTTP requests to a WebSocket route.

## Testing

- `go test ./pkg/gofr -run Test_WebSocket_PlainHTTPRequest -count=1 -v`
- `gofmt`
- `git diff --check`

The full repository test suite was also checked. The `examples/http-server` integration test requires Redis on port 2002 and MySQL on port 2001; those services were unavailable in the local environment.

### DIFF
--- a/pkg/gofr/websocket.go
+++ b/pkg/gofr/websocket.go
@@ -29,10 +29,13 @@ func (a *App) OverrideWebsocketUpgrader(wsUpgrader websocket.Upgrader) {
 // within the handler context. User can access the underlying WebSocket connection using `ctx.GetWebsocketConnection()`.
 func (a *App) WebSocket(route string, handler Handler) {
 	a.GET(route, func(ctx *Context) (any, error) {
-		connID := ctx.Request.Context().Value(websocket.WSConnectionKey).(string)
+		connID, ok := ctx.Request.Context().Value(websocket.WSConnectionKey).(string)
+		if !ok {
+			return nil, websocket.ErrorConnection
+		}
 
 		conn := a.httpServer.ws.GetWebsocketConnection(connID)
-		if conn.Conn == nil {
+		if conn == nil || conn.Conn == nil {
 			return nil, websocket.ErrorConnection
 		}
 
--- a/pkg/gofr/websocket_test.go
+++ b/pkg/gofr/websocket_test.go
@@ -182,3 +182,21 @@ func TestSerializeMessage(t *testing.T) {
 		})
 	}
 }
+func Test_WebSocket_PlainHTTPRequest(t *testing.T) {
+	testutil.NewServerConfigs(t)
+
+	app := New()
+
+	app.WebSocket("/ws", func(ctx *Context) (any, error) {
+		return "ok", nil
+	})
+
+	server := httptest.NewServer(app.httpServer.router)
+	defer server.Close()
+
+	resp, err := http.Get(server.URL + "/ws")
+	require.NoError(t, err)
+	defer resp.Body.Close()
+
+	assert.Equal(t, http.StatusInternalServerError, resp.StatusCode)
+}


aryanmehrotra (changes requested):
Thanks @dharmkoshiya74-collab — **the bug is real and the source change is correct.** I reproduced it and verified your fix removes it. The blocker is the regression test: it passes on unpatched `development`, so it does not guard the thing you fixed.

### The bug, confirmed

Reproduced against a server started with `App.Run()` — which matters, see below:

```
GET /ws   (no Upgrade header)   ->  500

{"level":"ERROR","message":{
  "error":"interface conversion: interface {} is nil, not string",
  "stack_trace":"goroutine 101 [running]: ...
     gofr.dev/pkg/gofr.(*App).WebSocket.func1(...)
       /pkg/gofr/websocket.go:32 +0x1bc ..."}}
```

`WSHandlerUpgrade` sets `WSConnectionKey` only when `gorillaWebsocket.IsWebSocketUpgrade(r)` is true (`pkg/gofr/http/middleware/web_socket.go#L17`). A plain `GET` falls straight through, and `websocket.go:32` then type-asserts a nil interface.

Worth stating for the changelog: this is **not a crash** — `logPanic` recovers it and answers 500. What it is, is a full goroutine stack dumped at ERROR for every non-upgrade request, which any crawler, uptime check or browser tab triggers for free. Still very much worth fixing.

Your patch, measured the same way:

| | plain GET | genuine handshake |
|---|---|---|
| `development` | 500 + panic + stack dump | `"ok"` |
| **with this PR** | **500, no panic, no stack dump** | `"ok"` |

---

## Blocking: the test passes without the fix

I applied **only** your test hunk to unpatched `development`:

```
--- PASS: Test_WebSocket_PlainHTTPRequest (0.00s)
```

The reason is in the table above: the status code is **500 both before and after**, because the panic recovery in `pkg/gofr/http/middleware/logger.go` already answers 500. `assert.Equal(t, http.StatusInternalServerError, ...)` therefore cannot distinguish "panicked and got recovered" from "returned `ErrorConnection` cleanly". The stack trace in the log is the only observable difference, so that is what the assertion has to look at.

## Blocking: the test does not reach the code path it names

`httptest.NewServer(app.httpServer.router)` serves the bare router, and `middleware.WSHandlerUpgrade` is attached inside `Run()` (`pkg/gofr/http_server.go#L87-L89`), not at router construction. So the test exercises "a WebSocket route with the upgrade middleware absent" — a configuration that cannot occur in production — rather than "a plain HTTP request to a WebSocket route".

Both panic on the same line, so the test isn't *wrong*, it's just not testing the reported scenario. To confirm: I sent 400 raw WebSocket handshakes at a server built that way and got **0/400 successful upgrades** — nothing on that router can ever upgrade.

## Blocking: three new lint failures

Same package, cache cleaned, run back to back:

```
development:  14 issues
with this PR: 17 issues

pkg/gofr/websocket_test.go:190:28  revive   unused-parameter: 'ctx' -> '_'
pkg/gofr/websocket_test.go:197:23  noctx    http.Get -> http.NewRequestWithContext + Client.Do
pkg/gofr/websocket_test.go:199:2   wsl_v5   missing whitespace above this line
```

Also: there is no blank line between the end of `TestSerializeMessage` and the new function.

---

## A test that does work

Drop-in replacement for `Test_WebSocket_PlainHTTPRequest`. I verified it **fails on `development`** (3/3 runs) and **passes with your `websocket.go` change** (5/5 runs), and it adds **no lint issues** (17 -> 14, back to baseline).

The one non-obvious part: the app has to be constructed *inside* the capture, because the logger binds to `os.Stderr` at construction. `StdoutOutputForFunc` will not see it either — ERROR goes to `errorOut`, which is stderr (`pkg/gofr/logging/logger.go#L230-L231`). I lost twenty minutes to both of those, so they are called out in the comment.

```go
func Test_WebSocket_PlainHTTPRequestDoesNotPanic(t *testing.T) {
	configs := testutil.NewServerConfigs(t)

	var statusCode int

	// The app is built inside the capture because the logger binds to os.Stderr when it is
	// constructed; an app created outside would keep writing to the real stderr.
	logs := testutil.StderrOutputForFunc(func() {
		app := New()
		app.WebSocket("/ws", func(*Context) (any, error) { return "ok", nil })

		go app.Run()

		host := fmt.Sprintf("http://localhost:%d", configs.HTTPPort)
		testutil.WaitForHTTPServer(t, host)

		req, err := http.NewRequestWithContext(t.Context(), http.MethodGet, host+"/ws", http.NoBody)
		require.NoError(t, err)

		resp, err := http.DefaultClient.Do(req)
		require.NoError(t, err)

		defer resp.Body.Close()

		statusCode = resp.StatusCode
	})

	// The status alone cannot tell the two outcomes apart: a recovered panic and a handled
	// error both answer 500. The stack trace is the only observable difference.
	assert.NotContains(t, logs, "interface conversion", "handler panicked on a non-upgrade request")
	assert.NotContains(t, logs, "stack_trace", "handler panicked on a non-upgrade request")
	assert.Equal(t, http.StatusInternalServerError, statusCode)
}
```

Imports needed: `fmt`, `net/http`, `testing`, `github.com/stretchr/testify/assert`, `github.com/stretchr/testify/require`, `gofr.dev/pkg/gofr/testutil`.

It would also be worth adding a second case asserting a **genuine** handshake still works, so the guard can never be tightened into rejecting real WebSocket traffic:

```go
c, resp, err := websocket.DefaultDialer.DialContext(t.Context(), "ws://localhost:"+..., nil)
```

---

## Not blocking, but worth your view

**The `conn == nil` guard.** I could not construct a case that reaches it — `WSConnectionKey` is only set after `AddWebsocketConnection` succeeds. It is still correct to keep: `GetWebsocketConnection` is a bare map lookup (`pkg/gofr/websocket/websocket.go#L146`) and returns `nil` on a miss. Just flagging that it is hardening rather than the reported defect, so it doesn't need a test of its own.

**Is 500 the right status?** A plain `GET` on a WebSocket route is a client mistake, and 500 tells a monitoring system the server is broken. The upgrade middleware already answers a *failed* upgrade with 400, and RFC 9110 §15.5.22 defines **426 Upgrade Required** for exactly this case. I'd lean 426, but it is a behaviour change and fine as a follow-up — your call whether to fold it in here.

Happy to re-review as soon as the test is updated.

PR #4104, opened 2026-08-31
fix(http): respond with 405 Method Not Allowed and harden static file symlinks

### LINKED ISSUE #3853: Wrong HTTP method on a registered route returns 404, not 405
**Describe the bug**

A request to a registered path with a method that path does not implement returns **404 Not Found**
with the "route not registered" envelope, instead of **405 Method Not Allowed**. The route *is*
registered — just not for that method — so the response tells the client something untrue, and a
client cannot distinguish "this endpoint does not exist" from "this endpoint does not accept POST".

The cause is ordering. `App.Run` registers a catch-all after the user's routes (`pkg/gofr/gofr.go:181`):

```go
a.httpServer.router.PathPrefix("/").Handler(handler{
    function:  catchAllHandler,
    container: a.container,
})
```

`catchAllHandler` returns `ErrorInvalidRoute` (`pkg/gofr/handler.go:223`). Because it matches on path
alone, it matches a wrong-method request before gorilla/mux ever consults its
`MethodNotAllowedHandler`, so mux's own 405 is unreachable in any GoFr app.

**To Reproduce**

1. The code is

   ```go
   package main

   import "gofr.dev/pkg/gofr"

   func main() {
       app := gofr.New()
       app.GET("/thing", func(*gofr.Context) (any, error) { return "ok", nil })
       app.Run()
   }
   ```

2. The error is

   ```
   GET    /thing -> 200  application/json  {"data":"ok"}
   POST   /thing -> 404  application/json  {"error":{"message":"route not registered"}}
   DELETE /thing -> 404  application/json  {"error":{"message":"route not registered"}}
   ```

Also measured at the router level, with and without the catch-all registered, to isolate the cause:

```
                            method  path     status  Allow  body
with catch-all (real app)   GET     /thing   200            {"data":"ok"}
with catch-all (real app)   POST    /thing   404     ""     {"error":{"message":"route not registered"}}
with catch-all (real app)   DELETE  /thing   404     ""     {"error":{"message":"route not registered"}}
bare mux (catch-all absent) GET     /thing   200            {"data":"ok"}
bare mux (catch-all absent) POST    /thing   405     ""     (empty)
bare mux (catch-all absent) DELETE  /thing   405     ""     (empty)
```

Two things worth noting from the bottom half: mux does produce the correct status once the catch-all
is out of the way, so the routing information needed to answer 405 is present — but mux's default
`MethodNotAllowedHandler` sets **no `Allow` header** and writes no body, so simply deleting the
catch-all would trade one incomplete answer for another.

**Expected behavior**

A registered path reached with an unsupported method should return `405`, with an `Allow` header
listing the methods that path does accept, as RFC 9110 §15.5.6 requires:

```
POST /thing -> 405  Allow: GET  {"error":{"message":"method not allowed"}}
```

A path that genuinely has no route registered should keep its current `404` and envelope.

- OS: macOS 15 (Darwin 25.5.0)
- gofr version: reproduced at `11058337` (`development`); the catch-all registration it stems from
  is long-standing, but I have not measured which releases are affected
- go version: go1.26.5

**More description**

Surfaced while reviewing #3820, which adds a 405 to the static file handler for the same class of
problem — `DELETE /static` was answering `200` with the file's content. That fix is local to
`staticHandler` and deliberately does not touch this, because the framework-wide version is a
different change with a wider blast radius: it makes static endpoints the only place in GoFr that
answers 405, and closing that inconsistency is what this issue is for.

Implementation note for whoever picks it up: the fix is probably not "remove the catch-all". It
exists so unrouted paths get GoFr's error envelope rather than mux's bare `404 page not found`, and
per the table above mux's 405 path has no `Allow` header and no body. Setting
`Router.MethodNotAllowedHandler` to a GoFr handler that emits the envelope *and* the `Allow` header,
and having the catch-all defer to it, keeps both answers correct. The method list for `Allow` is
available from `mux.Route.GetMethods()` during a `Router.Walk`, which `gofr.go:188` already performs.

This is a behaviour change for any client currently treating the 404 envelope as the wrong-method
response, so it likely wants a minor release rather than a patch.

I'm happy to pick this up if it's wanted.

### LINKED ISSUE #3855: Static file serving follows symlinks out of the served directory; replace lexical containment with os.Root

AbhiPra24:
## Pull Request Template

**Description:**

- Addressed Issue #3853: When a request arrives for a registered route with an unsupported HTTP method, the router responds with `405 Method Not Allowed` with the `Allow` header populated with accepted methods (e.g. `Allow: GET, POST`) and GoFr error JSON envelope `{"error":{"message":"method not allowed"}}`, rather than `404 Not Found`.
- Configured `NotFoundHandler` and `MethodNotAllowedHandler` on the router and registered routes with `Path` before `Methods` so Gorilla Mux accurately tracks method mismatches.
- Added `AllowedMethods(r *http.Request) []string` to inspect allowed HTTP methods for the requested path.
- Addressed Issue #3855: Hardened static file serving by verifying `filepath.EvalSymlinks` remains within the root directory before opening/serving files. Symlinks pointing outside the served root return `404 Not Found`.
- Added unit tests covering 405/404 handling, `Allow` header formatting, and symlink containment scenarios.

Fixes #3853
Fixes #3855

- None. Requests to existing routes with unsupported HTTP methods will now receive standard `405 Method Not Allowed` instead of `404 Not Found`.

**Additional Information:**

- All new code is covered by unit tests.
- Code is formatted with `gofmt`.

**Checklist:**

### DIFF
--- a/pkg/gofr/gofr.go
+++ b/pkg/gofr/gofr.go
@@ -176,13 +176,21 @@ func (a *App) httpServerSetup() {
 
 	a.setupGraphQL()
 
-	if a.container.Logger != nil {
-		a.container.Logger.Infof("Registered HTTP server on port: %d", a.httpServer.port)
-	}
-
-	a.httpServer.router.PathPrefix("/").Handler(handler{
+	a.httpServer.router.NotFoundHandler = handler{
 		function:  catchAllHandler,
 		container: a.container,
+	}
+
+	a.httpServer.router.MethodNotAllowedHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		methods := a.httpServer.router.AllowedMethods(r)
+		if len(methods) > 0 {
+			w.Header().Set("Allow", strings.Join(methods, ", "))
+		}
+
+		handler{
+			function:  methodNotAllowedHandler,
+			container: a.container,
+		}.ServeHTTP(w, r)
 	})
 
 	var registeredMethods []string
@@ -455,7 +463,7 @@ func (a *App) setupGraphQL() {
 		}
 
 		// Functional endpoint: served via POST per spec to ensure data safety and consistency.
-		a.httpServer.router.NewRoute().Methods(http.MethodPost).Path("/graphql").Handler(a.graphqlManager.GetHandler())
+		a.httpServer.router.NewRoute().Path("/graphql").Methods(http.MethodPost).Handler(a.graphqlManager.GetHandler())
 	}
 }
 
--- a/pkg/gofr/gofr_test.go
+++ b/pkg/gofr/gofr_test.go
@@ -1935,3 +1935,91 @@ func Test_HTTPMethods(t *testing.T) {
 		})
 	}
 }
+
+func Test_HTTPMethodNotAllowed_And_NotFound(t *testing.T) {
+	testutil.NewServerConfigs(t)
+
+	app := New()
+	app.GET("/thing", func(*Context) (any, error) {
+		return "ok", nil
+	})
+	app.POST("/multi", func(*Context) (any, error) {
+		return "multi-post", nil
+	})
+	app.PUT("/multi", func(*Context) (any, error) {
+		return "multi-put", nil
+	})
+
+	app.httpServerSetup()
+
+	tests := []struct {
+		name          string
+		method        string
+		path          string
+		expectedCode  int
+		expectedAllow string
+		expectedBody  string
+	}{
+		{
+			name:          "GET registered route returns 200",
+			method:        http.MethodGet,
+			path:          "/thing",
+			expectedCode:  http.StatusOK,
+			expectedAllow: "",
+			expectedBody:  `{"data":"ok"}`,
+		},
+		{
+			name:          "POST on GET-only route returns 405 with Allow header",
+			method:        http.MethodPost,
+			path:          "/thing",
+			expectedCode:  http.StatusMethodNotAllowed,
+			expectedAllow: "GET",
+			expectedBody:  `{"error":{"message":"method not allowed"}}`,
+		},
+		{
+			name:          "DELETE on GET-only route returns 405 with Allow header",
+			method:        http.MethodDelete,
+			path:          "/thing",
+			expectedCode:  http.StatusMethodNotAllowed,
+			expectedAllow: "GET",
+			expectedBody:  `{"error":{"message":"method not allowed"}}`,
+		},
+		{
+			name:          "GET on multi-method route (POST, PUT) returns 405 with sorted Allow header",
+			method:        http.MethodGet,
+			path:          "/multi",
+			expectedCode:  http.StatusMethodNotAllowed,
+			expectedAllow: "POST, PUT",
+			expectedBody:  `{"error":{"message":"method not allowed"}}`,
+		},
+		{
+			name:          "GET on unregistered route returns 404",
+			method:        http.MethodGet,
+			path:          "/unregistered",
+			expectedCode:  http.StatusNotFound,
+			expectedAllow: "",
+			expectedBody:  `{"error":{"message":"route not registered"}}`,
+		},
+		{
+			name:          "POST on unregistered route returns 404",
+			method:        http.MethodPost,
+			path:          "/unregistered",
+			expectedCode:  http.StatusNotFound,
+			expectedAllow: "",
+			expectedBody:  `{"error":{"message":"route not registered"}}`,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			w := httptest.NewRecorder()
+			req := httptest.NewRequestWithContext(t.Context(), tc.method, tc.path, http.NoBody)
+
+			app.httpServer.router.ServeHTTP(w, req)
+
+			assert.Equal(t, tc.expectedCode, w.Code)
+			assert.Equal(t, tc.expectedAllow, w.Header().Get("Allow"))
+			assert.JSONEq(t, tc.expectedBody, w.Body.String())
+		})
+	}
+}
--- a/pkg/gofr/handler.go
+++ b/pkg/gofr/handler.go
@@ -230,6 +230,10 @@ func catchAllHandler(*Context) (any, error) {
 	return nil, gofrHTTP.ErrorInvalidRoute{}
 }
 
+func methodNotAllowedHandler(*Context) (any, error) {
+	return nil, gofrHTTP.ErrorMethodNotAllowed{}
+}
+
 func panicRecoveryHandler(re any, log logging.Logger, panicked chan struct{}) {
 	if re == nil {
 		return
--- a/pkg/gofr/handler_test.go
+++ b/pkg/gofr/handler_test.go
@@ -237,6 +237,18 @@ func TestHandler_catchAllHandler(t *testing.T) {
 	assert.Equal(t, gofrHTTP.ErrorInvalidRoute{}, err, "TEST Failed.\n")
 }
 
+func TestHandler_methodNotAllowedHandler(t *testing.T) {
+	c := Context{
+		Context: t.Context(),
+	}
+
+	data, err := methodNotAllowedHandler(&c)
+
+	assert.Nil(t, data, "TEST Failed.\n")
+
+	assert.Equal(t, gofrHTTP.ErrorMethodNotAllowed{}, err, "TEST Failed.\n")
+}
+
 func TestHandler_livelinessHandler(t *testing.T) {
 	resp, err := liveHandler(&Context{})
 
@@ -1130,6 +1142,7 @@ func TestHandler_Char_ErrorLogging(t *testing.T) {
 		{"entity-not-found-is-INFO", gofrHTTP.ErrorEntityNotFound{}, "INFO", false},
 		{"already-exists-is-WARN", gofrHTTP.ErrorEntityAlreadyExist{}, "WARN", false},
 		{"invalid-route-is-INFO", gofrHTTP.ErrorInvalidRoute{}, "INFO", false},
+		{"method-not-allowed-is-INFO", gofrHTTP.ErrorMethodNotAllowed{}, "INFO", false},
 		{"panic-recovery-is-ERROR", gofrHTTP.ErrorPanicRecovery{}, "ERROR", true},
 		{"too-many-requests-is-WARN", gofrHTTP.ErrorTooManyRequests{}, "WARN", false},
 		{"client-closed-is-DEBUG", gofrHTTP.ErrorClientClosedRequest{}, "DEBUG", false},
--- a/pkg/gofr/http/errors.go
+++ b/pkg/gofr/http/errors.go
@@ -97,6 +97,21 @@ func (ErrorInvalidRoute) StatusCode() int {
 	return http.StatusNotFound
 }
 
+// ErrorMethodNotAllowed represents an error when the HTTP method is not allowed for a route.
+type ErrorMethodNotAllowed struct{}
+
+func (ErrorMethodNotAllowed) Error() string {
+	return "method not allowed"
+}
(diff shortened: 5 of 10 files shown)


aryanmehrotra (changes requested):
Thanks @AbhiPra24 — I ran this against a real server rather than reading the diff, and **both bugs are real and both of your fixes work.** There is one regression that blocks it, and it is not visible from the diff or from the unit tests.

### Both issues reproduce on `development`

```
#3853  POST /only-get        -> 404, no Allow header       (should be 405)
#3855  GET /static/evil.txt  -> 200, body "TOPSECRET"      (file outside the served root)
```

### Your fixes do what they claim

| probe | `development` | this PR |
|---|---|---|
| `POST /only-get` | 404 | **405, `Allow: GET`** |
| `DELETE /ping` | 404 | **405, `Allow: GET, POST`** |
| `GET /static/evil.txt` | **200 `TOPSECRET`** | **404** |
| `GET /static/evildir/passwords.txt` | 200 | **404** |
| `dotdot-inside.txt` (relative `..` landing back inside) | 200 | 200 — correctly still served |

Same results under `GOFR_ROUTER=trie`, so the two matchers stay consistent. The `Allow` values are right, and the symlink containment does not over-reject.

---

## Blocking: CORS preflight now returns 405

```
OPTIONS /only-get   Origin: http://x.test   Access-Control-Request-Method: GET

development:  HTTP/1.1 200 OK
              Access-Control-Allow-Methods: GET, POST, OPTIONS
              Access-Control-Allow-Origin: *

this PR:      HTTP/1.1 405 Method Not Allowed
              Allow: GET
              (no CORS headers at all)
```

Every cross-origin browser call to a GoFr app breaks — the preflight fails, so the browser never sends the real request.

**Why**, from the pinned `gorilla/mux v1.8.1` (`mux.go:138-168`): the middleware chain is applied **only** in the matched-route branch, guarded by `match.MatchErr == nil`:

```go
for _, route := range r.routes {
    if route.Match(req, match) {
        if match.MatchErr == nil {                       // <- middleware only here
            for i := len(r.middlewares) - 1; i >= 0; i-- {
                match.Handler = r.middlewares[i].Middleware(match.Handler)
            }
        }
        return true
    }
}

if match.MatchErr == ErrMethodMismatch {
    if r.MethodNotAllowedHandler != nil {
        match.Handler = r.MethodNotAllowedHandler        // <- no middleware
        return true
    }
    ...
}
if r.NotFoundHandler != nil {
    match.Handler = r.NotFoundHandler                    // <- no middleware
    ...
}
```

`middleware.CORS` is registered through `router.Use(...)` (`pkg/gofr/http_server.go:67-70`, alongside `Tracer`, `Logging` and `Metrics`), so anything served by those two handler fields bypasses all four.

## Blocking: 404 and 405 disappear from logs and metrics

Same root cause, and it also hits 404 — which previously went through `PathPrefix("/")`, a real route, and therefore had the full chain. Measured on the same app:

| | `development` | this PR |
|---|---|---|
| `app_http_response_count{path="/nope",status="404"}` | 1 | **absent** |
| `app_http_response_count{path="/only-get"}` | 1 | **absent** |
| access-log lines for `/nope` | 3 | **0** |
| access-log lines for `/only-get` | 2 | **0** |

Only the `/ping` 200 is still counted. A scanner walking unknown paths, or a client using the wrong verb, becomes completely invisible to logs, metrics and traces.

### A shape that keeps the middleware

Rather than moving to mux's handler fields, keep the catch-all as a **real route** and decide inside it:

```go
a.httpServer.router.PathPrefix("/").Handler(handler{
    function:  catchAllHandler,   // 405 + Allow when AllowedMethods(r) is non-empty, else 404
    container: a.container,
})
```

`AllowedMethods` already skips routes with no methods, so the catch-all does not report itself. Because the catch-all matches, `match.MatchErr` stays nil, the whole chain runs, and CORS still intercepts `OPTIONS` before the handler is reached. Your `AllowedMethods`, `ErrorMethodNotAllowed` and the `Allow` formatting all carry over unchanged — this only changes where the decision is made.

If you go that way, worth re-checking whether the `Path(...).Methods(...)` reorder in `Router.Add` is still needed, since `ErrMethodMismatch` would no longer be what drives the 405.

---

## Smaller things

**An unrelated line was deleted.** `gofr.go` drops:

```go
if a.container.Logger != nil {
    a.container.Logger.Infof("Registered HTTP server on port: %d", a.httpServer.port)
}
```

Confirmed by running both: 1 such line on `development`, 0 here. It sits right where the catch-all was replaced, so it looks accidental.

**Two new lint failures**, both in the new test:

```
pkg/gofr/http/router_test.go:1007:21  G306: Expect WriteFile permissions to be 0600 or less (gosec)
pkg/gofr/http/router_test.go:1011:21  G306: (same)
```

**`Allow` omits HEAD and OPTIONS.** `Allow: GET` while `OPTIONS /only-get` is answered 405. RFC 9110 §10.2.1 has Allow list what the resource supports; fixing the CORS path will change this anyway, so it is worth settling both together.

**On #3855, this departs from what the issue specified.** The issue rules out this approach by name:

> Adding `filepath.EvalSymlinks` before the comparison would close the common case but reintroduces the TOCTOU window `os.Root` exists to eliminate — the link can be repointed between the resolve and the open.

Being fair to your version: the residual window needs write access **inside** the served tree, which is a far higher bar than the original bug, and `EvalSymlinks` avoids the breaking change `os.Root` brings (an absolute symlink pointing inside the root keeps working, where `os.Root` would reject it). That may well be the better trade — but it is a deliberate divergence from the design the issue landed on, and it should be argued in the PR description rather than left implicit.

**Two unrelated issues in one PR.** #3853 is self-contained and, once the middleware regression is fixed, ready. #3855 has an open design question. Splitting them would let the 405 work merge without waiting on that discussion.

---

## What I ran

| check | result |
|---|---|
| `go build ./...` | clean |
| `go test ./pkg/gofr/http/...` | pass |
| `go test ./pkg/gofr` | fails identically on `development` (metrics-port FATAL) — not this PR |
| `gofmt -l` | 2 dirty files, both pre-existing |
| `golangci-lint --new-from-rev` | 2 new, listed above |

The unit tests all pass, which is worth saying plainly: nothing in them exercises CORS, logging or metrics on a 404/405 path, which is why an end-to-end run was the only thing that surfaced it. A test asserting that a 404 still carries `Access-Control-Allow-Origin` would have caught it and would be worth adding alongside the fix.

Good work on both root causes — the symlink reproduction in particular is a real vulnerability and the containment logic you wrote handles the `..`-back-inside case correctly, which is the part that is easy to get wrong.

PR #4105, opened 2026-08-31
chore(go): upgrade to Go 1.27

aryanmehrotra:
Moves the repo to Go 1.27: every module's `go` directive (32 `go.mod` files plus `go.work`), the CI matrix, the `golang:` base images, and the docs that name a Go version.

Two CI pins had to move with it, and three characterization tests changed. Everything below was verified locally on go1.27.0 against go1.26.5.

## golangci-lint v2.12.2 → v2.13.2 — required, not housekeeping

v2.12.2 is built with go1.25 and cannot analyse a Go 1.27 toolchain at all. Both halves fail:

| Setup | Result |
|---|---|
| v2.12.2, go.mod at 1.26, toolchain go1.27 | typecheck errors **inside the standard library** — `could not import math/rand/v2 (… method must have no type parameters)`, because 1.27 added generic methods and its `go/types` cannot parse them |
| v2.12.2, go.mod at 1.27 | refuses to start: `can't load config: the Go language version (go1.26) used to build golangci-lint is lower than the targeted Go version (1.27.0)` |

v2.13.2 (2026-08-27) is the first release built on go1.27.

The bump is **version-neutral**, which is the part worth checking: running v2.13.2 over the root module on go1.26 and again on go1.27 produces the *identical* 46 issues. Go 1.27 introduces no new lint findings — the linter bump is purely about being able to run.

## setup-go '1.26' → '1.27' in the three lint/quality jobs

`code_quality`, `lint_changed_submodules` and `linting_party` do not go through `.github/actions/setup-go-toolchain`, so `GOTOOLCHAIN` stays `local`. Against a go-1.27 module they stop dead at `go mod download`:

```
go: go.mod requires go >= 1.27.0 (running go 1.26.5; GOTOOLCHAIN=local)
```

The test matrix shifts `['1.26','1.25','1.24']` → `['1.27','1.26','1.25']`, and the two coverage-upload guards move with it (`matrix.go-version == '1.24'` → `'1.25'`) so `parse_coverage` and `upload_coverage` still find their artifacts.

## Three tests changed — all one cause

Go 1.27 re-implements `encoding/json` on top of `encoding/json/v2`. `GOEXPERIMENT=nojsonv2` makes all three pass again, which is what identifies this as a stdlib wire-format change rather than a behaviour regression in GoFr.

**Invalid UTF-8 is still replaced by U+FFFD, but emitted as a raw UTF-8 rune** instead of a six-byte `�` escape. Same decoded string, different wire bytes — and `Test_LogWireFormat_HTMLEscapingOnTheRealPath` and `Test_LoggingContract_RequestLogEscaping` exist specifically to pin wire bytes, so they now pin the 1.27 form.

**`UnmarshalTypeError` no longer prefixes the Go type name** onto the field path, so `field charBindTarget.a` became `field .a` in `TestRequest_Char_BindJSON`. The error's classification is unchanged; only its text is.

Each site carries a comment recording the change and the `nojsonv2` escape hatch. **This is user-visible**: anyone asserting on GoFr log bytes or on bind-error strings will see the same shift.

## Verification

| Check | Result |
|---|---|
| `go build ./...`, all 33 modules | clean |
| `go vet ./...`, all 33 modules | clean |
| `go mod tidy -diff` across `pkg/` | clean (all modules already have the two require blocks 1.27's tidy wants) |
| `gofmt` on changed files | clean |
| `./pkg/gofr/...` | pass |
| all 27 `pkg/` submodules | pass |
| `./examples/...` | pass against live MySQL, Redis and Zipkin |
| `examples/using-s3-filestore`, `using-add-filestore` | pass |
| `.github/scripts/check-binaries.sh` | pass |

**Not verified locally:** `examples/using-subscriber`, which needs Kafka. It fails identically on go1.26 and go1.27 here (`TestMainInitialization`, 30s timeout, no broker), so the failure is environmental — but CI's Kafka service is what will actually confirm it.

One flake seen and dismissed: `TestCircuitBreaker_MixedHTTPMethods` asserts wall-clock `< 2s` and returned 2.021s once during a heavily loaded parallel sweep. In isolation it passes 8/8 on go1.27 and 6/6 on go1.26. Pre-existing timing margin, untouched here.


aryanmehrotra:
## Self-review

Branch is now up to date with `development` (0 behind; the merge brought in #4108 and #4109, and both packages were re-tested on the real 1.27 toolchain: `pkg/gofr/ai/...`, `pkg/gofr/container`, `pkg/gofr/http/middleware` all green).

Everything below came out of actually running the json paths under 1.27 rather than reading the diff.

### 🔴 The two error-string changes are masking a real regression in `Bind`

`request_test.go` relaxes two expectations from `field charBindTarget.a` to `field .a`, and the comment explains it as json/v2 no longer prefixing the Go type name. **That is not what json/v2 does.** It still prefixes the type name — GoFr loses it because `Bind` passes a *pointer to an interface*:

```go
// pkg/gofr/http/request.go:93
return json.Unmarshal(body, &i)   // i is already a pointer; &i makes it **any
```

Measured on go1.27.0 darwin/arm64, same struct, same input:

| call shape | go 1.26 | go 1.27 |
|---|---|---|
| `json.Unmarshal(body, &i)` — what GoFr does | `field charBindTarget.a` | **`field .a`** |
| `json.Unmarshal(body, i)` — `i` is already a pointer | `field charBindTarget.a` | `field charBindTarget.a` |

So this is not a cosmetic upstream rename. On Go 1.27 every GoFr user's bind error loses the struct name — `json: cannot unmarshal number into Go struct field .a of type string` — and this PR bakes that degradation into the test as if it were expected.

The `&` is redundant: `Bind` already rejects a non-pointer at `request.go:80`, so `i` is guaranteed to be a pointer by the time line 93 runs. Dropping it restores the message on 1.27 and leaves 1.26 unchanged, and both test expectations can then stay as they are.

The `&i` itself is ancient (`fb585c15d`, "Db utility (#10)") — harmless until 1.27, which is why nothing caught it. I'd rather fix it here than ship a worse error message with the toolchain bump.

### 🟡 The version matrix no longer tests three versions

`go-version: ['1.27','1.26','1.25']`, but `go.work` says `go 1.27.0` and `.github/actions/setup-go-toolchain` deliberately installs whatever `go.work` asks for and sets `GOTOOLCHAIN=auto`. So the 1.25 and 1.26 legs upgrade themselves to 1.27 and every leg tests the same toolchain.

That was already true before this PR, but it matters more now: the tests changed in this PR pin **1.27-only** wire bytes, so a leg that genuinely ran 1.26 would fail. Either the older legs should be dropped, or they need `GOTOOLCHAIN=local` and a build tag, and the "lowest matrix Go version" comment on the coverage upload should say which version actually produces it.

### 🟡 The two wire-format tests pin bytes that depend on the *consuming* module

The escaped-vs-raw U+FFFD difference is gated by the main module's `go` directive, not by the toolchain:

| main module `go` directive | `json.Marshal` of invalid UTF-8 |
|---|---|
| `go 1.26.0` | `{"uri":"/��/ok"}` |
| `go 1.27.0` | raw U+FFFD bytes |

`GOEXPERIMENT=nojsonv2` reverts it, as the comments say — worth noting `GODEBUG=jsonv2=0` does **not**, so that lever is the right one to document. Since these tests exist to pin GoFr's log wire format, it is worth saying explicitly in the comment that the format follows the application's own `go` line, not GoFr's.

### 🔵 Scope

`golangci-lint` v2.12.2 → v2.13.2 rides along in three places. Almost certainly needed to parse 1.27, but it is an unrelated version bump inside a Go-upgrade PR and is not mentioned in the description.

### ✅ What checks out

- **All 32 `go.mod` files are on `1.27.0`** — no stragglers, which is the usual failure mode for this change.
- `go.work` and the root `Dockerfile` moved with them.
- Docs corrected rather than merely bumped: `README.md` claimed **1.24** while `go.mod` already required 1.26, and `docs/AGENTS.md` claimed 1.25. Both are now accurate.
- The coverage-upload condition was moved off the dropped `1.24` leg instead of being left dangling.
- `go build ./...` clean and the full `./pkg/...` suite green locally on go1.27.0.

### Suggested follow-up before merge

1. Change `request.go:93` to `json.Unmarshal(body, i)` and revert the two expectations in `request_test.go`.
2. Correct the comment above `TestRequest_Char_BindJSON` — the cause is the pointer-to-interface, not a json/v2 rename.
3. Decide on the matrix: three labels that all run 1.27 is worse than one honest label.

aryanmehrotra:
## Acted on the first finding — `Bind` keeps its error message, and gets slightly faster

Pushed `c041fe2a`. `request.go` now unmarshals into the pointer it was given instead of a pointer to the interface holding it, and the two expectations in `TestRequest_Char_BindJSON` go back to naming the struct. The comment blaming json/v2 for the rename is gone, since that was the wrong diagnosis.

### Behaviour is preserved, not just for the two cases in the test

Every bind shape was differenced between the two call forms on go1.27.0 — valid struct, malformed body, empty body, `null`, unknown field, pointer to map / slice / `[]byte` / `any`. All identical. Only two differ:

| case | `&i` (before) | `i` (after) |
|---|---|---|
| wrong-typed field | `field .a` | **`field charBindTarget.a`** |
| typed nil pointer | `nil`, binds nothing | `json: Unmarshal(nil *T)` |

The second is worth a look, though I think it is fine: no test covered it, and a handler that binds a typed nil pointer would nil-deref on the next line anyway. Say the word if you would rather preserve the silent no-op and I will guard it explicitly.

### Confirmed against a running server, not only in unit tests

Built two binaries from this branch — one with `&i`, one without — and ran each as a real GoFr service against this worktree's sbx sandbox (`gofr-go127`: MySQL, Redis, Zipkin; all three went `asleep` → `awake`, so the app genuinely connected). Same requests to each:

```
POST /bind  {"a":"hello","b":42}
  before  201  {"data":{"a":"hello","b":42}}
  after   201  {"data":{"a":"hello","b":42}}          identical

POST /bind  {"a":5}
  before  500  json: cannot unmarshal number into Go struct field .a of type string
  after   500  json: cannot unmarshal number into Go struct field charBindTarget.a of type string

POST /bind  {"a":"x","b":"nope"}
  after   500  json: cannot unmarshal string into Go struct field charBindTarget.b of type int
```

Malformed body, empty body, `null`, unknown field and `/.well-known/alive` all responded identically across the two binaries.

### No performance cost

Removing the indirection removes two allocations per bind:

| | B/op | allocs/op |
|---|---:|---:|
| `json.Unmarshal(body, &i)` | 6355 | 24 |
| `json.Unmarshal(body, i)` | **6335** | **22** |

`-benchtime 3000x -count 6`; ns/op overlaps between the two, allocation counts are stable.

### Still open from the review

- The version matrix labels 1.25 / 1.26 / 1.27 but every leg resolves to 1.27 via `go.work` + `setup-go-toolchain`.
- The `golangci-lint` v2.12.2 → v2.13.2 bump is unmentioned in the description. Worth keeping — a v2.12.2 binary refuses a 1.27 module outright: `the Go language version (go1.26) used to build golangci-lint is lower than the targeted Go version (1.27.0)`. That is why it has to move in the same PR.

aryanmehrotra:
## Thorough behaviour check — and a correction to my previous comment

I said the only side effect of dropping the `&` was a typed nil pointer starting to error. **That was understated.** Differencing every target shape rather than the handful I had looked at showed a nil pointer used to swallow *five* body shapes, not one. Pushed `24d75c08` to preserve all of them.

### Method

10 target shapes × 12 body shapes = 120 cases, run through three implementations, comparing **GoFr on Go 1.26 (today's behaviour)** against each candidate on Go 1.27. Targets: `*struct`, `nil *struct`, `*map`, `nil *map`, `*slice`, `*[]byte`, `*any`, `*int`, `*string`, `**struct`. Bodies: valid, wrong-type-string, wrong-type-int, nested-wrong-type, malformed, empty, null, unknown-field, array-into-struct, bare number, bare string, invalid UTF-8.

| candidate on Go 1.27 | rows differing from Go 1.26 baseline |
|---|---|
| upgrade alone (`&i`, i.e. this PR before my commits) | **7 / 120** |
| upgrade + pointer fix (`c041fe2a` alone) | 10 / 120 |
| upgrade + pointer fix + nil guard (`24d75c08`) | **5 / 120** |

My first fix made things *worse* on this measure before the guard was added — worth stating plainly.

### What the nil guard restores

A nil pointer cannot be written to. The old form decoded into the local interface, which dropped the static type, so a nil pointer silently accepted anything:

| body, into a `nil *T` | Go 1.26 (baseline) | pointer fix alone | with nil guard |
|---|---|---|---|
| wrong-type-string | `nil`, no-op | `ERR field T.a` | `nil`, no-op |
| wrong-type-int | `nil`, no-op | `ERR field T.b` | `nil`, no-op |
| nested-wrong-type | `nil`, no-op | `ERR field T.i.z` | `nil`, no-op |
| array-into-struct | `nil`, no-op | `ERR cannot unmarshal array` | `nil`, no-op |
| bare number / string | `nil`, no-op | `ERR cannot unmarshal …` | `nil`, no-op |
| malformed / empty | syntax error | syntax error | syntax error |

Reproduced exactly — decoded into a throwaway interface, so syntax errors still surface and type errors still do not. `reflect.Value` is hoisted out of the guard that was already there, so the check costs nothing.

To be clear about what this preserves: binding a nil pointer quietly succeeding is a defect of the same family GoFr fixed for non-pointers in #3770. I have kept it because the Go upgrade is not the place to change it. It deserves its own PR.

### The 5 remaining differences are upstream, not reachable from GoFr

| case | Go 1.26 | Go 1.27 |
|---|---|---|
| `*[]byte` + invalid base64 | `illegal base64 data at input byte 0` | `json: cannot unmarshal string into Go value of type []uint8: illegal base64 data…` |
| `*struct` nested wrong type | `field Inner.i.z` | `field T.i.z` (names the outer struct now) |
| `**struct` wrong type ×2, nested ×1 | `field T.a` | `field .a` — a `**T` target still loses the name |

None of these are affected by how `Bind` calls `Unmarshal`; they are encoding/json's own rendering under json/v2.

### Also changed by the upgrade, outside Bind

`json.Marshal` emits U+FFFD for invalid UTF-8 as a raw rune instead of `�`, and GoFr has two encoders on that path:

- `pkg/gofr/http/responder.go:93` — `json.NewEncoder(buf).Encode(resp)`, so **response bodies** change
- `pkg/gofr/logging/logger.go:107` — `json.NewEncoder(out).Encode(entry)`, so **log lines** change

That is what the two wire-format test updates in this PR are pinning. It is not avoidable in a library — the default follows the application's own `go` directive — but it is a visible change for anyone whose payloads carry invalid UTF-8, and it belongs in the release notes for whichever release ships Go 1.27.

### Performance

Allocation counts are deterministic; ns/op on this machine was too noisy to quote.

| | B/op | allocs/op |
|---|---:|---:|
| `&i` (before) | 6353 | 24 |
| `i` + nil guard (after) | **6331** | **22** |

The guard adds no allocation over the plain pointer form.

### Linter

Agreed on keeping the `golangci-lint` v2.12.2 → v2.13.2 bump here — it is required, not optional. A v2.12.2 binary refuses the module outright:

```
can't load config: the Go language version (go1.26) used to build golangci-lint
is lower than the targeted Go version (1.27.0)
```

I hit that locally, which is why lint for these packages was verified in CI rather than on my machine.

### Still open

The version matrix labels 1.25 / 1.26 / 1.27 but `go.work` plus `.github/actions/setup-go-toolchain` upgrades every leg to 1.27, so all three legs test the same toolchain.

Umang01-hash (dismissed):
Ran the touched packages on the real 1.27 toolchain — all green, gofmt/vet clean, and CI Code Quality is passing.

The Bind change is the nice bit: confirmed that under v2 the old `Unmarshal(body, &i)` drops the struct name from type errors (`field .a` vs `field charBindTarget.a`), and passing `i` directly fixes it. Also checked the nil-pointer path stays a no-op exactly like before, and that the v1 API still keeps case-insensitive matching + last-wins duplicate keys on 1.27, so nothing downstream shifts. LGTM.

aryanmehrotra:
Closing this without merging. The branch `chore/go-1.27` stays, so nothing here is lost and this can be reopened.

What the work turned up, for whoever picks it up next:

**The version matrix was never testing three versions.** `go.work` declared a newer Go than the two older legs, and `.github/actions/setup-go-toolchain` upgraded them to satisfy it, so all three legs ran the newest toolchain under three names. Every release since v1.56.0 shipped a go directive above its own lowest leg, so this predates the branch.

**Making the legs real is possible, but only for the root module.** `gofr.dev` depends on no published GoFr release, so it builds and tests on 1.25 with `GOWORK=off`. The workspace cannot: examples and datasource submodules require published `gofr.dev@v1.57.0`/`v1.59.0` and `file/s3@v0.3.0`, and those declare `go 1.26.0` permanently. Nothing third-party blocks 1.25 — the constraint is entirely GoFr's own published artifacts.

**Two real defects surfaced once the older legs actually ran**, both fixed here and both worth keeping regardless of what happens to the Go upgrade:

- `Bind` unmarshalled through a pointer to an interface, which on Go 1.27 drops the struct name from the error (`field .a` rather than `field charBindTarget.a`). Differenced across 10 target shapes and 12 bodies; 0 of 120 cases change behaviour, and it removes two allocations per bind.
- A `Template` returned from a DELETE handler renders into a 204, which cannot carry a body. `Render` discards `Execute`'s error, so the attempt left a truncated page in anything recording writes verbatim — and where it truncated moved with the toolchain. A real client has always received an empty body on every version.

**One difference cannot be fixed in GoFr and should be release-noted.** The json/v2 wire format is selected by the toolchain, not by the go directive — a build on 1.27 emits invalid UTF-8 as a raw U+FFFD where 1.26 emits `�`. GoFr encodes its logs and its response envelope with `encoding/json`, so the same service emits different bytes depending on the Go it was built with. The tests here pin it per toolchain behind a `go1.27` build tag.

Worth splitting out of a version bump if this is revisited: the two fixes stand alone, and the matrix correction is independent of which Go version is being added.

PR #4106, opened 2026-08-31
test(service): assert circuit-breaker parallelism by overlap, not by stopwatch

aryanmehrotra:
`TestCircuitBreaker_MixedHTTPMethods` and `TestCircuitBreaker_ParallelExecution` both proved *"these requests ran in parallel"* by timing them: the handler slept one second, five requests were fired, and the test asserted the total came in under two (respectively four) seconds.

That is a race between the implementation and the machine.

**It fails when it shouldn't.** The bound sat one second above a one-second floor, so a loaded runner fails a circuit breaker that is behaving perfectly. Caught in the act during a parallel local sweep:

```
circuit_breaker_test.go:1116:
    Error:    "2.021348s" is not less than "2s"
    Messages: Different HTTP methods should execute in parallel
```

**And it passes when it shouldn't.** A serialized-but-fast implementation sails through both, because elapsed time cannot distinguish *concurrent* from *quick*. The 4s bound on `ParallelExecution` is loose enough to admit a lot of serialization.

## What replaces it

A barrier the handler blocks on. Every request waits until all five are inside the handler **at the same instant** — reachable only if the client dispatched them concurrently, unreachable if it did not, at any speed. A request whose peers never arrive answers 503 and the test names it:

```
Request 3 did not overlap the others: requests were serialized, not parallel
```

No wall clock is consulted, so there is nothing left to tune per machine.

The barrier's timeout is a **liveness** bound, not a performance one — it exists only so serialization fails as an assertion instead of hanging until `go test`'s own deadline. What it has to cover is five goroutines each opening a localhost connection, so 15s is roughly four orders of magnitude of headroom, and it is spent only on a run that was going to fail anyway. The first waiter to time out releases the others, which keeps a failing run at one timeout rather than five (120s → 15s) and holds it inside the 5-minute CI step budget.

## Verified in both directions

A test that no longer fails is only half the job — the other half is proving it still can.

| | Result |
|---|---|
| Healthy, `-race` | **10/10 pass** |
| Healthy, runtime | **~2s → ~0.26s** for the pair (the two 1s sleeps are gone) |
| Regressed | **both fail in 15s**, with the "serialized, not parallel" message |
| `golangci-lint` v2.12.2 on `pkg/gofr/service/...` | 0 issues |
| Full `pkg/gofr/service` package | pass ×3 |

The regression injected was moving `cb.mu.Lock()` above the `f(ctx)` call in `executeWithCircuitBreaker` — holding the lock across the request, which is the exact mistake the *"try recovery without holding lock"* comment there warns against. The old stopwatch assertions caught it too, but only because the handler slept a second; they would not have caught a fast one.

---

Found while verifying #4105 (Go 1.27), but unrelated to it — the flake reproduces on 1.26, so it is split out to keep that PR to the upgrade.

### DIFF
--- a/pkg/gofr/service/circuit_breaker_test.go
+++ b/pkg/gofr/service/circuit_breaker_test.go
@@ -930,10 +930,83 @@ func TestCircuitBreaker_HealthEndpointWithTimeout(t *testing.T) {
 	resp.Body.Close()
 }
 
+// barrierTimeout bounds how long a handler waits for its peers to show up. It is
+// a LIVENESS bound, not a performance one: the barrier releases the instant the
+// last participant arrives, so a healthy run never spends time here no matter
+// how loaded the machine is. It exists only so that a genuinely serialized
+// client fails as an assertion instead of hanging until go test's own deadline.
+//
+// Fifteen seconds is therefore enormous - what it has to cover is five
+// goroutines each opening a localhost connection, which is sub-millisecond work
+// - and the whole timeout is spent only on a run that was going to fail anyway.
+const barrierTimeout = 15 * time.Second
+
+// concurrencyBarrier proves that n requests were in flight at the same instant,
+// without measuring how long anything took.
+//
+// The two tests below used to assert a wall-clock bound - "all five finished in
+// under 2s, therefore they ran in parallel". That conflates "concurrent" with
+// "fast": the bound sat one second above a one-second floor, so a loaded runner
+// failed a circuit breaker that was behaving perfectly. It also could not tell a
+// serialized-but-quick implementation from a parallel one.
+//
+// A barrier asserts the property itself. Every handler blocks until all n
+// handlers are inside it together, which is reachable only if the client
+// dispatched them concurrently, and is unreachable if it did not - at any speed.
+type concurrencyBarrier struct {
+	n       int
+	all     chan struct{} // closed once every participant has arrived
+	givenUp chan struct{} // closed by the first participant to time out
+
+	giveUp sync.Once
+
+	mu      sync.Mutex
+	arrived int
+}
+
+func newConcurrencyBarrier(n int) *concurrencyBarrier {
+	return &concurrencyBarrier{n: n, all: make(chan struct{}), givenUp: make(chan struct{})}
+}
+
+// arrive records one participant and blocks until all n have arrived, reporting
+// whether that happened within timeout. A false return means the requests were
+// serialized.
+//
+// The first participant to time out releases every other waiter too. Without
+// that, serialized requests each wait the full timeout in turn and a failing run
+// costs n*timeout - long enough to hit the CI step budget instead of reporting.
+func (b *concurrencyBarrier) arrive(timeout time.Duration) bool {
+	b.mu.Lock()
+
+	b.arrived++
+	if b.arrived == b.n {
+		close(b.all)
+	}
+
+	b.mu.Unlock()
+
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+
+	select {
+	case <-b.all:
+		return true
+	case <-b.givenUp:
+		return false
+	case <-timer.C:
+		b.giveUp.Do(func() { close(b.givenUp) })
+
+		return false
+	}
+}
+
 // TestCircuitBreaker_ParallelExecution tests that requests execute in parallel.
 func TestCircuitBreaker_ParallelExecution(t *testing.T) {
+	const numRequests = 5
+
 	requestCount := 0
 	mu := sync.Mutex{}
+	barrier := newConcurrencyBarrier(numRequests)
 
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
 		mu.Lock()
@@ -942,7 +1015,15 @@ func TestCircuitBreaker_ParallelExecution(t *testing.T) {
 
 		mu.Unlock()
 
-		time.Sleep(1 * time.Second) // Simulate slow endpoint
+		// Releases only when all five requests are inside the handler at once.
+		// If they were serialized, the first one waits alone and answers 503,
+		// which the status assertion below turns into a failure.
+		if !barrier.arrive(barrierTimeout) {
+			w.WriteHeader(http.StatusServiceUnavailable)
+
+			return
+		}
+
 		w.WriteHeader(http.StatusOK)
 
 		_, _ = w.Write([]byte(`{"status": "ok"}`))
@@ -963,13 +1044,10 @@ func TestCircuitBreaker_ParallelExecution(t *testing.T) {
 			Interval:  5 * time.Second,
 		})
 
-	startTime := time.Now()
-
 	var wg sync.WaitGroup
 
-	numRequests := 5
-
 	errors := make([]error, numRequests)
+	statuses := make([]int, numRequests)
 
 	// Launch 5 concurrent requests
 	for i := 0; i < numRequests; i++ {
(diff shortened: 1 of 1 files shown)


PiyushSingh-ZS (commented):
Reviewed the full diff. This is a clean, self-contained fix and I think it should go in as-is modulo two small things.

### What I checked

- The diagnosis matches the code: `assert.Less(totalTime, 2*time.Second)` against a 1s handler sleep leaves a 1s margin, which is both too tight for a loaded runner and too loose to distinguish *concurrent* from *merely fast*. Agreed that it fails and passes for the wrong reasons in both directions.
- `concurrencyBarrier.arrive` looks correct under concurrency: `arrived` is incremented under `mu`, `close(b.all)` fires exactly once at `arrived == n`, and the `giveUp sync.Once` makes the release idempotent. No double-close path.
- The "first waiter to time out releases the others" detail is a good one — it keeps a failing run at `1 x timeout` instead of `n x timeout`, which is what keeps it inside the CI step budget.

Replacing a wall-clock proxy with the property itself is the right call, and the 15s liveness bound is well justified in the comment (it is only ever spent on a run that was going to fail anyway).

### Two suggestions

**1. Move the `require.Len` above the server construction.**

```go
require.Len(t, methods, numMethods)
```

currently runs after `httptest.NewServer` is already installed with a barrier sized `numMethods`. If someone adds a sixth method to the slice, all five handlers block for the full 15s before that assertion is ever reached, so the clear "you forgot to widen numMethods" message arrives 15 seconds late. Checking the length before the server is built makes it immediate.

**2. The 503 can be masked by the circuit breaker.**

On the serialized-failure path the handler answers 503, which counts as a failure against the breaker under test. With enough of them the breaker opens and the remaining requests fail with an *error* rather than a status, so `require.NoError` fires first and the operator sees a generic error instead of the carefully worded `"requests were serialized, not parallel"` message that this PR exists to produce.

Only reachable on an already-failing run, so it is a diagnostics concern rather than a correctness one. If you want the message to survive, a 200 with a marker header (or a distinct body) would keep the failure signal out of the breaker's accounting.

Nice work on verifying it in both directions — injecting the `cb.mu.Lock()` regression and confirming the old assertions only caught it because of the 1s sleep is exactly the check that makes a replacement test trustworthy.

aryanmehrotra:
Point 2 is right and is fixed at `591ecfa18`. Point 1 I'd like to push back on — but it pointed at a real defect two lines away, so it was worth raising.

**The 503 → `statusNotOverlapped` (409).** Agreed, and for the reason you gave: the diagnostic status was feeding the breaker under test the very signal the test emits when it fails. Past the threshold the remaining requests come back as errors, `require.NoError` fires first, and the operator reads a generic circuit-open error instead of "the requests were serialized, not parallel".

Worth adding that it can't fire *today* — `handleFailure` opens on `failureCount > threshold` (strict `>`, `circuit_breaker.go:174`), and the thresholds are 10 and 5 against at most 5 requests. But the message shouldn't depend on that arithmetic staying true, so it's a named constant now with the reasoning next to it.

**`require.Len` already runs before any request is dispatched.** The ordering in the file is:

```
barrier := newConcurrencyBarrier(numMethods)
server := httptest.NewServer(...)      // nothing has requested it yet
httpSvc := NewHTTPService(server.URL, ...)
methods := []func(){...}                // closes over httpSvc
require.Len(t, methods, numMethods)     // ← here; still no request made
for i, method := range methods { go ... }  // ← first request
```

So a mismatch fails in milliseconds with a length error, in *either* direction, and no handler ever reaches the barrier. It also can't move any earlier than it is: every entry in `methods` closes over `httpSvc`, which needs `server.URL`, so the server necessarily exists by that point.

**But the comment above it had the failure direction backwards** — it claimed an added method "would leave every handler waiting", which is not what happens. That's fixed; it now describes what the assertion actually does and why it sits where it sits. Thanks for looking hard enough at that block to surface it.

pkg/gofr/service/circuit_breaker_test.go
```diff
@@ -1068,10 +1068,100 @@ func TestCircuitBreaker_HealthEndpointWithTimeout(t *testing.T) {
 	resp.Body.Close()
 }
 
+// barrierTimeout bounds how long a handler waits for its peers to show up. It is
+// a LIVENESS bound, not a performance one: the barrier releases the instant the
+// last participant arrives, so a healthy run never spends time here no matter
+// how loaded the machine is. It exists only so that a genuinely serialized
+// client fails as an assertion instead of hanging until go test's own deadline.
+//
+// Fifteen seconds is therefore enormous - what it has to cover is five
```
Umang01-hash, line 1077: These `circuit_breaker.go:89` / `:174` line refs are accurate today but will silently rot as that file shifts. Consider referring to the symbols (`handleFailure` / the `StatusCode > 500` check) instead of line numbers.

pkg/gofr/service/circuit_breaker_test.go
```diff
@@ -1068,10 +1068,100 @@ func TestCircuitBreaker_HealthEndpointWithTimeout(t *testing.T) {
...
+//
+// A barrier asserts the property itself. Every handler blocks until all n
+// handlers are inside it together, which is reachable only if the client
+// dispatched them concurrently, and is unreachable if it did not - at any speed.
+type concurrencyBarrier struct {
+	n       int
+	all     chan struct{} // closed once every participant has arrived
+	givenUp chan struct{} // closed by the first participant to time out
+
+	giveUp sync.Once
```
Umang01-hash, line 1124: This is the same barrier idea as `checkBarrier` in container/health_concurrency_test.go (also a `barrierTimeout` const). Fine as a package-local helper, but if a third use shows up it's worth pulling into a shared internal testutil.

Umang01-hash (dismissed):
LGTM. The barrier asserts parallelism by construction instead of a wall-clock bound — verified the premise (CB runs `f(ctx)` without holding the lock), traced the barrier for races/double-close, and confirmed with a probe that it fails on serialization and the `givenUp` fan-out caps a failing run at one timeout (not n×). Race runs, gofmt, vet, and golangci-lint (new-issues) all clean. Two minor notes inline; neither blocks.

pkg/gofr/service/circuit_breaker_test.go
aryanmehrotra: Done in `2ac949580`. Both refs now name the symbol:

> counts any status above 500 as a failure -- the `result.StatusCode > 500` check in `executeWithCircuitBreaker` -- and opens once `failureCount` exceeds the threshold in `handleFailure`

Checked against the file as it stands: `circuit_breaker.go:89` is the `err != nil || (result != nil && result.StatusCode > 500)` guard inside `executeWithCircuitBreaker`, and `:174` is the `cb.failureCount > cb.threshold` compare inside `handleFailure` — so the two names are exact, not approximate.

pkg/gofr/service/circuit_breaker_test.go
aryanmehrotra: Agreed, and recorded rather than extracted — `2ac949580` adds a note on `concurrencyBarrier` pointing at `checkBarrier`.

They are the same idea but not the same helper, and the difference is what a merge would have to preserve:

| | `checkBarrier` (container) | `concurrencyBarrier` (service) |
|---|---|---|
| `arrive` returns | nothing | `bool` — the test asserts on it |
| on timeout | that goroutine gives up alone | first timeout closes `givenUp`, releasing all |
| cost of a failing run | `n * barrierTimeout` | one `barrierTimeout` |

The broadcast is why this one can afford a 15s timeout where the container copy uses 5s. Two uses do not pay for a `testutil` package; the third would, and the note says what has to survive the move.

Umang01-hash (approved):
Re-verified at the new head: both notes resolved (line refs now point at executeWithCircuitBreaker/handleFailure by name; the checkBarrier separation is documented). Delta is comment-only. Re-ran locally — gofmt clean, the two tests race x3 pass, full service package race passes, golangci-lint new-issues 0. Barrier re-audited (no double-close/race/deadlock/spurious-fail). Good to merge.

PR #4108, opened 2026-08-31
fix(ai): make Embed a method on ai.LLM instead of a type assertion

aryanmehrotra:
**Description:**

#3757 shipped embeddings as `ai.EmbeddingLLM`, an optional interface the caller asserts on `ctx.LLM()`, rather than as a method on `ai.LLM`. Its own description says the opposite — "adds text embeddings to the `ai` package as a first-class `LLM` method — the peer of `Stream`", with `ctx.LLM("embed").Embed(...)` as the usage — but a later revision moved it off the interface to preserve `ai.LLM`'s compatibility freeze, and the description was never updated. This is unreleased, so it is still cheap to settle.

**Why the assertion is the wrong shape**

**It cannot fail.** Every `ai.LLM` the container returns implements `Embed`, including `notConfiguredLLM`, which was given the method *precisely so the assertion always succeeds* (`container.go:329`). The caller writes a capability check that checks nothing, and the two failures it appears to guard against — a chat-only provider, and no model registered — are reported by the call regardless, as `ErrEmbedNotSupported` and `ErrLLMNotConfigured`.

**It is asymmetric with `Stream`.** Streaming is equally optional for a provider (`ai.StreamingModel`) and equally reported at call time (`ErrStreamNotSupported`), yet it sits on `ai.LLM` and is called directly. Two capabilities with the same structure had two different caller experiences.

| Capability | Provider side | Caller side (before) | Caller side (after) |
|---|---|---|---|
| Streaming | `ai.StreamingModel` | `Stream` on `ai.LLM` | unchanged |
| Embeddings | `ai.Embedder` | `.(ai.EmbeddingLLM)` | `Embed` on `ai.LLM` |

**The change**

```diff
- model, ok := c.LLM("embed").(ai.EmbeddingLLM)
- if !ok {
-     return nil, errEmbeddingsUnavailable
- }
- resp, err := model.Embed(c, in.Inputs)
+ resp, err := c.LLM("embed").Embed(c, in.Inputs)
```

`Embed` moves onto `ai.LLM` and `EmbeddingLLM` is deleted. **No type gains or loses a method** — `*llm` (`llm.go:73`), `notConfiguredLLM` and `MockLLM` (`mock_ai.go:278`) already had `Embed`, which is why the whole diff is comments and call sites: 61 insertions, 98 deletions.

`ai.Embedder`, the provider-side optional capability, is untouched. A provider that cannot embed still implements nothing.

**One, narrow: an externally hand-written `ai.LLM` fake stops compiling until it gains an `Embed` method.** `ai.LLM` shipped in v1.58.0 and v1.59.0 without `Embed`, so that is a real compatibility cost and not a hypothetical one.

Who is and is not affected:

| | Affected |
|---|---|
| Callers of `ctx.LLM()` | No — adding a method never breaks a caller |
| Mocks via `ai.NewMockLLM` | No — exported since v1.58.0, and regenerated with `Embed` |
| A wrapper embedding `ai.LLM` (`struct { ai.LLM }`) | No — the method is promoted |
| A fully hand-written `ai.LLM` fake | **Yes** — needs an `Embed` method |

Two things keep the blast radius small. `ai.LLM` is **return-only** — `Container.LLM(name ...string) ai.LLM` is the only place it appears in the exported API, and nothing accepts one (`AddLLM`/`SetLLM` take `ai.Model`, `llms` is unexported, `NewMockContainer` wires `ai.NewMockLLM` itself). So a hand-written implementation cannot be handed to GoFr and exists only inside a user's own scaffolding, where `ai.NewMockLLM` was already the supported route. And `ai.EmbeddingLLM` has never appeared in a release, so deleting it costs nothing.

This does reverse a promise: the note on `ai.LLM` added in #3722 said capabilities would only ever arrive as separate asserted interfaces. That note is rewritten here, and `compat_test.go` keeps `legacyFake` so a future accidental addition to `ai.LLM` still fails a build in this repo before it reaches a user. The same freeze on `ai.Model` — the provider-side interface users genuinely implement, via `AddLLM` — is unchanged and still holds.

**Additional Information:**

No new dependencies. Docs (`docs/advanced-guide/llm/page.md`) and `examples/using-ai` updated to the direct call.

Verification:

- `go build ./...` and `go vet ./...` clean.
- `go test ./pkg/gofr/ai/... ./pkg/gofr/container/... -race -count=1` green.
- Full `go test ./pkg/...` green. One failure appeared on the first run, `TestCircuitBreaker_MixedHTTPMethods` in `pkg/gofr/service`, which this branch does not touch (0 files changed there); it passes at `-count=5` and is a known pre-existing timing flake.
- `go test ./examples/using-ai/... -v` green, including the `embed returns one vector per input, in input order` subtest, which drives the edited handler end to end through HTTP against the in-test mock provider.
- Full `go test ./...` otherwise green. Five example packages fail locally on `dial tcp [::1]:2001: connect: connection refused` — they need MySQL/Redis/Kafka, which CI provides and my machine does not. None of the five contains any `ai.`/`LLM` code.
- `golangci-lint run` on the changed packages reports only 3 pre-existing findings in untouched lines of `container.go` and `health.go`.

**Checklist:**


pkg/gofr/ai/model.go
```diff
@@ -55,38 +54,26 @@ type Descriptor interface {
...
-//	if !ok {
-//		return nil, errors.New("embeddings unavailable")
-//	}
-//
-//	resp, err := e.Embed(ctx, []string{"hello"})
-//
-// It is asserted rather than declared on LLM so that adding embeddings does not break the
-// hand-written fakes and third-party wrappers LLM promises to keep compiling.
-type EmbeddingLLM interface {
 	Embed(ctx context.Context, input []string, opts ...Option) (*EmbeddingResponse, error)
```
Umang01-hash, line 72: The one real compat cost: `LLM` shipped in v1.59.0 without embedding, so adding `Embed` here breaks external hand-written `LLM` fakes (not `ctx.LLM()` consumers). Deliberate + documented + mitigated via `NewMockLLM`, and `compat_test` still catches future accidental additions in-repo. Good timing too — the removed `EmbeddingLLM` never shipped.

Umang01-hash (approved):
LGTM. Clean simplification — the old `EmbeddingLLM` assertion could never fail (every `ctx.LLM()` implements it), so it read as a capability check while checking nothing; a method on `LLM` reporting `ErrEmbedNotSupported` at call time (mirroring `Stream`) is the honest shape.

Checked the compatibility angle since it touches an exported interface:
- Removing `EmbeddingLLM` breaks nothing released — it landed after v1.59.0 and never shipped in a tag.
- Adding `Embed` to `LLM` does change the released interface (v1.59.0 had no embedding), so it breaks external hand-written `LLM` fakes — but not `ctx.LLM()` consumers. It's documented in-code, mitigated (`NewMockLLM` already carries `Embed`), and the `compat_test` `legacyFake` still fails-to-compile in-repo on any future accidental addition, so the safety net's purpose is intact.

Verified locally: `notConfiguredLLM` assertion + method updated, example and docs migrated off the assertion, zero lingering `EmbeddingLLM` refs anywhere, and gofmt/vet/`go test ./pkg/gofr/ai/...` + container all green.

One thing worth an explicit nod from the team: this is a deliberate reversal of the LLM "compatibility freeze" — accepting the hand-written-fake break for one callable surface. Reads like the right call and the right time (before embeddings ship), just flagging it's a policy choice, not an accident.

pkg/gofr/ai/model.go
aryanmehrotra: Agreed, and checked against the tags rather than from memory:

- `v1.59.0:pkg/gofr/ai/model.go:41` is `Model` / `Generate` / `Stream` / `Tools` — no `Embed`. `ai.LLM` actually first shipped in **v1.58.0**, so the exposure is two releases, not one.
- No `EmbeddingLLM` or `Embedder` anywhere in `v1.59.0`, so removing `EmbeddingLLM` costs nothing.
- `ai.NewMockLLM` is exported from `mock_ai.go` and was already present in v1.59.0, so the mitigation was available before this change, not introduced by it.

I had briefly written "Breaking Changes: None" in the description; that was wrong and your comment is what caught it. The section now leads with the cost.

One refinement worth recording: a wrapper that *embeds* the interface (`type loggingLLM struct { ai.LLM }`) keeps compiling, since `Embed` is promoted. Only a fully hand-written fake breaks. And because `ai.LLM` is return-only — `Container.LLM()` is the sole appearance in the exported API, `AddLLM`/`SetLLM` take `ai.Model`, `llms` is unexported and `NewMockContainer` wires `ai.NewMockLLM` itself — such a fake cannot be handed to GoFr at all, so it only exists inside a user's own scaffolding.

So: real cost, narrow blast radius, and deliberate. Resolving on that basis.

PR #4111, opened 2026-09-01
fix(gofr): stop racing c.Context between serveWithGoroutine and App.WebSocket

### LINKED ISSUE #3930: Data race: App.WebSocket writes ctx.Context while serveWithGoroutine reads it (Test_WebSocket_Success fails under -race)

om7057:
## Summary

Fixes #3930.

serveWithGoroutine's select reads `c.Context.Done()` on the parent goroutine while the spawned handler goroutine can concurrently reassign `ctx.Context` via `context.WithValue(...)` - `App.WebSocket` does exactly this to thread the connection through the context. Both goroutines touching the same `*Context.Context` field is a data race.

Reproduces on clean development:

```
go test ./pkg/gofr/ -race -count=1 -short -run 'Test_WebSocket_Success'
```

## Fix

Snapshot the context to watch into a local variable (`watchCtx`) before spawning the handler goroutine, so the parent goroutine never reads the shared `c.Context` field again after that point. The child goroutine is still free to reassign `c.Context`; nothing else reads it concurrently after this change.

This is behavior-preserving for the non-WebSocket `requestTimeout > 0` path too, since a normal handler does not reassign `c.Context` mid-flight - `watchCtx` and `c.Context` are the same object there.

## CI

`-race` is now enabled on the root `./pkg/gofr` test run in the `PKG Unit Testing` job, across all three Go version legs, so this fix (and any future concurrency fix in this package) has an actual CI signal instead of relying on someone running `-race` by hand. Verified locally: `go test -short -race ./pkg/gofr` passes 15/15 runs with the exact flags CI uses.

## Related: self-referential context

Found while verifying this fix: the same line this PR touches, `ctx.Context = context.WithValue(ctx, websocket.WSConnectionKey, conn)`, passes `ctx` (a `*Context`, which embeds `context.Context`) as the parent instead of `ctx.Context`, making the resulting context self-referential. Any call that walks the parent chain other than the one lookup that resolves immediately recurses forever into an unrecoverable stack overflow. Independent of the race fixed here - this PR's fix stops the race's reader from ever calling `.Done()` on the poisoned context, which closes one crash path, but leaves the self-reference itself in place for the handler and anything downstream of it.

Filed as #4169, fixed in #4170.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] The repro command above is race-clean after this change
- [x] `go test ./pkg/gofr/ -race -count=1` passes in full
- [x] `go test ./pkg/gofr/ -run TestHandler` - all existing handler characterization tests pass unchanged, confirming the timeout/cancellation/panic paths are behaviorally identical

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -287,8 +287,16 @@ jobs:
             export APP_ENV=test
 
             # Run tests for root gofr package
+            #
+            # -race here specifically (not the sub-packages run below, which is
+            # slower and heavier on external services) is what makes concurrency
+            # fixes in this package durable: the WebSocket/serveWithGoroutine
+            # data race fixed alongside this line had no CI signal until now
+            # (#3930) — go test -race stayed a claim in a doc comment nobody
+            # could fail on. -race requires cgo, which ubuntu-latest provides by
+            # default.
             exit_code=0
-            go test -v -short -covermode=atomic \
+            go test -v -short -race -covermode=atomic \
               -coverpkg=./pkg/gofr -coverprofile=gofr_only.cov ./pkg/gofr || exit_code=$?
             if [ $exit_code -eq 2 ]; then
               echo "::error::Panic detected in root gofr package tests"
--- a/pkg/gofr/handler.go
+++ b/pkg/gofr/handler.go
@@ -175,6 +175,14 @@ func (h handler) serveWithGoroutine(c *Context, spanCtx trace.SpanContext, r *ht
 	done := make(chan handlerOutcome, 1)
 	panicked := make(chan struct{})
 
+	// Snapshot the context to watch for cancellation before spawning the
+	// handler goroutine below. h.function may reassign c.Context while it
+	// runs (App.WebSocket does, to thread the connection through it) —
+	// reading c.Context in the select instead of here would race that
+	// write. watchCtx is a local copy this goroutine alone reads, so the
+	// write in the spawned goroutine and this read never overlap.
+	watchCtx := c.Context
+
 	go func() {
 		defer func() {
 			panicRecoveryHandler(recover(), h.container.Logger, panicked)
@@ -187,13 +195,13 @@ func (h handler) serveWithGoroutine(c *Context, spanCtx trace.SpanContext, r *ht
 	}()
 
 	select {
-	case <-c.Context.Done():
+	case <-watchCtx.Done():
 		// Server-side timeout or client cancellation. Map to the matching
 		// gofrHTTP error so Respond emits 408 (timeout) or 499 (client
 		// closed).
 		err = gofrHTTP.ErrorRequestTimeout{}
 
-		if errors.Is(c.Context.Err(), context.Canceled) {
+		if errors.Is(watchCtx.Err(), context.Canceled) {
 			err = gofrHTTP.ErrorClientClosedRequest{}
 		}
 	case out := <-done:
--- a/pkg/gofr/handler.go
+++ b/pkg/gofr/handler.go
@@ -167,10 +167,17 @@ func (h handler) serveInline(c *Context, spanCtx trace.SpanContext) (result any,
 // after the handler hijacks the connection).
 //
 // The handler outcome is sent through a buffered channel rather than to
-// shared variables, so the goroutine never writes a memory location the
-// main goroutine reads — `go test -race` stays clean. Buffer size 1 lets
-// the handler goroutine finish writing and exit even after the main
-// goroutine has already taken the ctx.Done or panicked branch.
+// shared variables, so its (result, err) pair never becomes a location
+// both goroutines touch. Buffer size 1 lets the handler goroutine finish
+// writing and exit even after the main goroutine has already taken the
+// ctx.Done or panicked branch.
+//
+// c.Context is a location both goroutines DO touch: h.function runs in the
+// spawned goroutine and is free to reassign it (App.WebSocket does, to
+// thread the connection through it), while this goroutine's select used to
+// read c.Context.Done() directly — a data race, since nothing ordered the
+// two accesses. See the watchCtx snapshot below, taken before the goroutine
+// is spawned, for the fix.
 func (h handler) serveWithGoroutine(c *Context, spanCtx trace.SpanContext, r *http.Request) (result any, err error) {
 	done := make(chan handlerOutcome, 1)
 	panicked := make(chan struct{})


aryanmehrotra (changes requested):
Thanks @om7057 — **the race is real, your diagnosis is exact, and the fix is correct.** I reproduced it, confirmed the fix, and every gate is clean. What I am holding this on is that the fix currently ships with no way to notice if it regresses, plus one thing directly above your change that is now wrong. Both are small.

### Confirmed

Your repro command on clean `development`:

```
WARNING: DATA RACE
Write at 0x00c0006d71e0 by goroutine 45:
  gofr.dev/pkg/gofr.(*App).WebSocket.func1()      websocket.go:41
  ...serveWithGoroutine.func1()                   handler.go:183

Previous read at 0x00c0006d71e0 by goroutine 43:
  gofr.dev/pkg/gofr.handler.serveWithGoroutine()  handler.go:190
```

Exactly the two lines you name. `watchCtx` is the right shape and the right place — the parent stops touching the shared field, the child stays free to reassign it, and the non-WebSocket `requestTimeout > 0` path is genuinely unaffected because `watchCtx` and `c.Context` are the same object there.

| check | result |
|---|---|
| your repro command | passes |
| `Test_WebSocket*` x20 under `-race` | 0 races (`development`: 1) |
| full `pkg/gofr -race` | 0 races |
| `TestHandler` characterization | pass, unchanged |
| `gofmt` / `go build ./...` / `go vet` | clean |
| `golangci-lint --new-from-rev=origin/development` | 0 issues |

The full-suite `FAIL` is the pre-existing `metrics port 2121 is blocked` FATAL — identical on `development`, not yours.

---

## 1. The fix has no regression guard at all

Your note says `Test_WebSocket_Success` "already serves as the regression guard whenever `-race` is enabled". I checked, and your other claim is right: there is no `-race` anywhere in `.github/workflows/`. Put together, that means **nothing in CI can ever fail if this regresses** — the guard is conditional on a condition that is never true.

That is not a reason to write a new test; it is a reason to turn the flag on. Adding `-race` to the `pkg/gofr` unit-test job is a one-line change in this PR and it is what makes the fix durable. If a full `-race` run is too slow for every job, restricting it to `PKG Unit Testing` on a single Go version would still close this.

I would rather not merge a concurrency fix whose only protection is that someone remembers to run `-race` by hand.

## 2. The comment directly above your change is now false

`serveWithGoroutine`'s doc comment still says:

> The handler outcome is sent through a buffered channel rather than to shared variables, so the goroutine never writes a memory location the main goroutine reads — `go test -race` stays clean.

That is the exact claim this PR disproves, and it sits four lines above the code you changed. Someone reading it later will trust it. Please correct it in the same commit — a sentence noting that `c.Context` *is* such a location, which is why the snapshot exists, would leave the file honest.

## 3. Please link the root cause, because your fix does not remove it

While verifying, I found what is actually wrong at `websocket.go:41`:

```go
ctx.Context = context.WithValue(ctx, websocket.WSConnectionKey, conn)
```

`ctx` is `*gofr.Context`, which **embeds** `context.Context`. So the stored context's parent is `ctx` itself: `c.Context.Done()` -> `valueCtx.Done()` -> `c.Done()` -> `c.Context.Done()` -> ...

A real GoFr app whose WebSocket handler reads any context value other than the WS key:

```
>>> handler entered
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
--- process alive? DEAD
```

Same on `development` and on this branch — pre-existing, **not** something you introduced. A stack overflow is a `fatal error`, so `recover()` cannot catch it and the whole process dies. It only goes unnoticed today because `GetConnectionFromContext` looks up the one key that returns before recursing; tracing, auth, or any SDK taking a `context.Context` would hit it.

The one-word fix, which I verified keeps the WS key readable:

```go
ctx.Context = context.WithValue(ctx.Context, websocket.WSConnectionKey, conn)
//                                       ^^^^^^^^ the embedded context, not the *Context
```

The two defects are independent — neither fix covers the other:

| | `watchCtx` (this PR) | `WithValue(ctx.Context, ...)` |
|---|---|---|
| concurrent read/write of `c.Context` | fixed | still races |
| self-referential context -> fatal overflow | still poisoned | fixed |

Worth noting that `watchCtx` incidentally stops the *parent* from ever calling `Done()` on the poisoned context, so it closes one crash path while leaving the value itself in `c.Context` for the handler and everything downstream.

I am not asking you to fix that here — it belongs in its own change. But please open an issue for it and link it from this PR, so it does not get buried by this one going green. If you would rather take it yourself, say so and I will leave it to you.

---

Nice piece of debugging. The diagnosis is precise, the fix is the minimal correct one, and the comment you added explaining *why* the snapshot exists is exactly what that code needed. Happy to approve once `-race` is on in CI and the stale comment is corrected.

om7057:
@aryanmehrotra Findings 1 and 2 are addressed:

1. `-race` is now on the root `pkg/gofr` unit-test job across all three Go version legs (`.github/workflows/go.yml`), verified locally with the exact CI command (`go test -v -short -race ... ./pkg/gofr`, 15/15 runs green).
2. `serveWithGoroutine`'s doc comment is corrected to describe what's actually race-free (the handler outcome, via the buffered channel) versus what wasn't (`c.Context`), pointing at the `watchCtx` snapshot.

Finding 3 (the self-referential context / `WithValue(ctx, ...)` bug) is filed and fixed separately per your request: #4170, tracked as its own issue (#4169), rather than bundled here, since it's an independent defect with an independent fix.

Also just merged `development` in to resolve the conflict in `.github/workflows/go.yml` (both branches touched the same `go test` line, so I kept `-race` and adopted `development`'s broader `-coverpkg=./pkg/gofr/...` coverage change together). Full `pkg/gofr` suite is green under `-race -short`, matching CI.

Issue #4112, opened 2026-09-01
metrics: the gcp exporter submodule is unpublished — no module tag and gofr.dev serves no go-get meta for its path


akshat-kumar-singhal:
### Summary

`pkg/gofr/metrics/exporters/gcp` is its own Go module (`module gofr.dev/pkg/gofr/metrics/exporters/gcp`, `go 1.26.0`, requires `gofr.dev v1.59.0`), but **it cannot be depended on from outside this repository.** Two independent defects, each confirmed at the HTTP level today (2026-09-01):

**1. No module tag.** Go resolves a nested module by a tag with the module's subdirectory as prefix. Enumerating every page of `/repos/gofr-dev/gofr/tags` returns **276** tags, and none of them is `pkg/gofr/metrics/exporters/gcp/vX.Y.Z`. There is therefore no version for `go get` to select.

**2. The vanity server does not serve the submodule path.** The module root resolves:

```
$ curl -s 'https://gofr.dev/?go-get=1'
<meta name="go-import" content="gofr.dev git https://github.com/gofr-dev/gofr"/>
```

The submodule path does not — it returns **HTTP 404** with no `go-import` meta tag:

```
$ curl -s -o /dev/null -w '%{http_code}\n' 'https://gofr.dev/pkg/gofr/metrics/exporters/gcp?go-get=1'
404
```

So even once a tag is cut, resolution fails before tags are ever consulted.

### Reproduction

```
$ curl https://proxy.golang.org/gofr.dev/pkg/gofr/metrics/exporters/gcp/@v/list
not found: gofr.dev/pkg/gofr/metrics/exporters/gcp@latest:
unrecognized import path "gofr.dev/pkg/gofr/metrics/exporters/gcp":
reading https://gofr.dev/pkg/gofr/metrics/exporters/gcp?go-get=1: 404 Not Found
```

`go get gofr.dev/pkg/gofr/metrics/exporters/gcp` fails the same way.

### Why CI doesn't catch it

`examples/using-gcp-metrics/go.mod` pins `gofr.dev/pkg/gofr/metrics/exporters/gcp v0.0.0` and builds only because of

```
replace gofr.dev/pkg/gofr/metrics/exporters/gcp => ../../pkg/gofr/metrics/exporters/gcp
```

[...]

PR #4113, opened 2026-09-01
fix(metrics): supply the instance label Google's OTLP ingest requires

aryanmehrotra:
> **Corrected after self-review.** The first version of this description claimed `OTEL_RESOURCE_ATTRIBUTES` was read nowhere. That was wrong — `metricSdk.WithResource` already does `resource.Merge(resource.Environment(), res)`, so `location` was always settable from the environment. The label that genuinely has no source is **`instance`**. Details and the measurement are below.

## The bug

Google maps every point onto the `prometheus_target` monitored resource. Its [reference](https://docs.cloud.google.com/stackdriver/docs/reference/telemetry/v1.metrics) documents two labels as required, each with the words **"Reject the point if empty"**:

| Label | Sources Google accepts, in order | Supplied before this PR? |
|---|---|---|
| `location` | `location` → `cloud.availability_zone` → `cloud.region` | via env (SDK merges `resource.Environment()`) |
| `instance` | `instance` → `service.instance.id` → `faas.instance` → `k8s.pod.name:k8s.container.name` → `k8s.pod.name` → `host.id` | **no — nothing, anywhere** |

`buildResource` emitted only `service.name` and `framework_version`, and grepping all six `instance` sources across the tree returns **0 occurrences each**. So `instance` was empty in every deployment, and **Google rejected every point regardless of what the operator configured**.

The failure is invisible: the push authenticates, connects and succeeds; points are discarded afterwards, server-side. Nothing reaches `otel.SetErrorHandler`.

## Verified against a local OTLP receiver

I stood up a gRPC server implementing the OTLP metrics service and inspected what GoFr actually puts on the wire. The resource is a MeterProvider-level property shared by all readers, so this is byte-for-byte what the gcp exporter sends.

**Before (`origin/development`), with `OTEL_RESOURCE_ATTRIBUTES=location=us-central1`:**
```
  framework_version = dev
  location          = us-central1
  service.name      = emulator-app

prometheus_target.location  <- location=us-central1
prometheus_target.instance  <- NOTHING -> point rejected
```

**After, same environment:**
```
  framework_version = dev
  host.id           = B869AC7C-...
  location          = us-central1
  service.name      = emulator-app

prometheus_target.location  <- location=us-central1
prometheus_target.instance  <- host.id=B869AC7C-...
```

## The fix

- **`resource.WithHostID()`** supplies `host.id`, the last fallback Google accepts for `instance`. This is the actual fix.
- **A GCP detector**, for the zero-config path on Cloud Run where `location` should not need setting by hand. `registry.go:40` states core takes no dependency on the submodules, so rather than importing one, **`RegisterResourceDetector`** adds a seam: the gcp submodule registers its detector on blank import, and `Build` runs it **only** when `gcp` is selected — asserted by `TestBuildResource_detectorRunsOnlyForItsExporter`, so a prometheus-only app never reaches for a metadata server.
- **A startup warning** when no location resolves. Since the backend rejects silently, that is the only signal available.

### Startup cost — a regression this branch introduced and then fixed

`WithHostID` shells out to `/usr/sbin/ioreg` on darwin:

| Path | Cost |
|---|---|
| before this branch | 0.027 ms |
| naive version of this branch | **11.16 ms** ← every app, including pull-only |
| **current** — prometheus-only (the default) | **0.010 ms** |
| **current** — push exporter | 11.7 ms |

Host detection now runs only when a push exporter is selected. `TestBuildResource_prometheusOnlySkipsHostIDDetection` pins it, because the expense is invisible at the call site.

## Two documentation defects found alongside

1. **Wrong IAM role** in `README.md` (×2) and `gcp.go`'s package doc: all said `roles/monitoring.metricWriter`, which grants `monitoring.timeSeries.create` on **`monitoring.googleapis.com`** — an API this exporter never calls. It pushes to `telemetry.googleapis.com:443`, requiring **`roles/telemetry.writer`**.
2. **`x-goog-user-project`** must not be an OTLP header per Google's docs, and `METRICS_HEADERS` passes headers straight through. Now warns and points at `GOOGLE_CLOUD_QUOTA_PROJECT`.

## Notes for review

**Schema URL moves v1.17.0 → v1.43.0** when detectors run, since the SDK's detectors carry their own semconv. Checked specifically because `resource.Merge` errors on conflicting schema URLs — it merges cleanly and `service.name` / `framework_version` both survive.

**`hasLocation` parses `k=v` pairs** rather than substring-matching, so `custom.location` is not mistaken for a real `location`. That case is a test.

## Testing

- `go test ./pkg/gofr/metrics/...`, gcp submodule, `pkg/gofr`, `container`, `http/middleware`, `logging` — pass
- `golangci-lint run` — 0 issues, both modules
- 8 tests added
- End-to-end against the local OTLP receiver above

**Not verified:** acceptance by a live GCP project — that needs someone with an endpoint. The requirement itself is from Google's reference docs.

Ref #4112.

### DIFF
--- a/examples/using-gcp-metrics/README.md
+++ b/examples/using-gcp-metrics/README.md
@@ -83,6 +83,15 @@ OTEL_RESOURCE_ATTRIBUTES=location=us-central1
 
 `location` also accepts `cloud.region` or `cloud.availability_zone`.
 
+`instance` is filled from `host.id`, which GoFr detects automatically. Some
+environments -- minimal containers and some CI runners -- have no machine ID, and
+there `instance` has no source either and points are dropped for the same reason.
+Supply one the same way:
+
+```
+OTEL_RESOURCE_ATTRIBUTES=location=us-central1,service.instance.id=${HOSTNAME}
+```
+
 ## Local run
 
 Locally, ADC comes from `gcloud auth application-default login`. To avoid a real
--- a/examples/using-gcp-metrics/README.md
+++ b/examples/using-gcp-metrics/README.md
@@ -83,14 +83,35 @@ OTEL_RESOURCE_ATTRIBUTES=location=us-central1
 
 `location` also accepts `cloud.region` or `cloud.availability_zone`.
 
-`instance` is filled from `host.id`, which GoFr detects automatically. Some
-environments -- minimal containers and some CI runners -- have no machine ID, and
-there `instance` has no source either and points are dropped for the same reason.
-Supply one the same way:
+`instance` is filled from `host.id`, which GoFr detects automatically. **In a
+container it usually cannot be.** The OpenTelemetry SDK reads `/etc/machine-id`
+and `/var/lib/dbus/machine-id` from the container's own filesystem, and those come
+from the image, not from the node — nothing bind-mounts the host's copy in. So:
 
+| Base image | `/etc/machine-id` | `host.id` |
+|---|---|---|
+| `debian:12-slim`, `alpine` | absent | not detected |
+| `ubuntu:24.04` | present, **empty** | detected but blank |
+
+Either way `instance` has no source and every point is dropped, so on Kubernetes
+supply one explicitly. The pod name is unique per replica and available from the
+downward API:
+
+```yaml
+env:
+  - name: POD_NAME
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.name
+  - name: OTEL_RESOURCE_ATTRIBUTES
+    value: location=us-central1,service.instance.id=$(POD_NAME)
 ```
-OTEL_RESOURCE_ATTRIBUTES=location=us-central1,service.instance.id=${HOSTNAME}
-```
+
+`instance` also accepts `k8s.pod.name`. Cloud Run needs none of this: the GCP
+detector supplies `faas.instance` per revision instance, and the region for
+`location`, so both labels resolve with no configuration.
+
+The exporter warns separately at startup for whichever of the two is missing.
 
 ## Local run
 
--- a/examples/using-gcp-metrics/README.md
+++ b/examples/using-gcp-metrics/README.md
@@ -29,13 +29,25 @@ GMP requires.
 
 ## Deploy to Cloud Run (keyless)
 
-1. Grant the service's runtime service account permission to write metrics:
+1. Grant the service's runtime service account permission to write telemetry:
 
    ```bash
    gcloud projects add-iam-policy-binding <PROJECT_ID> \
      --member="serviceAccount:<RUNTIME_SA>@<PROJECT_ID>.iam.gserviceaccount.com" \
-     --role="roles/monitoring.metricWriter"
+     --role="roles/telemetry.writer"
    ```
+
+   `roles/telemetry.writer` is the role for `telemetry.googleapis.com`, which is
+   what this exporter pushes to. `roles/monitoring.metricWriter` grants
+   `monitoring.timeSeries.create` on `monitoring.googleapis.com` — a different
+   API that this exporter never calls — so granting it alone leaves the push
+   unauthorized.
+
+   With a service account attached, the quota project is resolved automatically.
+   If you authenticate with **user** credentials instead, also grant
+   `roles/serviceusage.serviceUsageConsumer` on the quota project and set
+   `GOOGLE_CLOUD_QUOTA_PROJECT`. Do not pass `x-goog-user-project` as a metrics
+   header; Google documents that this is not the supported route.
 
 2. Deploy — no credentials mounted, no `GOOGLE_APPLICATION_CREDENTIALS`:
 
@@ -46,10 +58,30 @@ GMP requires.
 3. On Cloud Run the app resolves ADC from the metadata server automatically.
    Metrics appear in Cloud Monitoring / Managed Service for Prometheus.
 
-> Cross-project GMP: grant `roles/monitoring.metricWriter` on the **destination**
+> Cross-project GMP: grant `roles/telemetry.writer` on the **destination**
 > project. No Workload Identity Federation is needed on Cloud Run itself — the
 > attached service account is sufficient. WIF only applies to workloads running
 > outside Google Cloud.
+
+## Required resource labels
+
+Google maps every point onto the `prometheus_target` monitored resource, which
+requires a **`location`** and an **`instance`**, and *rejects the point* when
+either is empty. A push that is missing them still authenticates and still
+succeeds — the points are discarded afterwards, server-side, so nothing on the
+client says the data went nowhere.
+
+On Google Cloud both are detected for you: the metadata server supplies the
+region/zone, and `host.id` backs `instance`.
+
+Anywhere else — local runs, other clouds, CI — `location` has no source, and the
+exporter warns at startup. Set it explicitly:
+
+```
+OTEL_RESOURCE_ATTRIBUTES=location=us-central1
+```
+
+`location` also accepts `cloud.region` or `cloud.availability_zone`.
 
 ## Local run
 
--- a/pkg/gofr/metrics/exporters/provider.go
+++ b/pkg/gofr/metrics/exporters/provider.go
@@ -69,20 +69,31 @@ func buildResource(ctx context.Context, cfg *Config, logger Logger) *resource.Re
 	}
 
 	opts := []resource.Option{
-		// OTEL_RESOURCE_ATTRIBUTES is the standard way to supply attributes a
-		// backend requires but the framework cannot know -- Google's OTLP ingest
-		// rejects any point whose prometheus_target has no "location", and only
-		// the operator knows the region.
+		// OTEL_RESOURCE_ATTRIBUTES carries attributes a backend requires but the
+		// framework cannot know -- Google fills prometheus_target.location from it,
+		// and only the operator knows the region.
+		//
+		// metricSdk.WithResource merges resource.Environment() in as well, so this
+		// is not what makes the variable work. It is here so buildResource returns a
+		// complete resource on its own rather than one that is only correct once a
+		// particular consumer merges it.
 		resource.WithFromEnv(),
-		// host.id is the last fallback Google accepts for the required "instance"
-		// label, so detecting it here is what keeps points from being dropped on
-		// hosts where nothing else supplies one.
-		resource.WithHostID(),
 		resource.WithAttributes(attrs...),
 	}
 
-	if d, ok := lookupDetector(strings.ToLower(strings.TrimSpace(cfg.Exporter))); ok {
-		opts = append(opts, resource.WithDetectors(d))
+	// Everything below is only meaningful to a push backend, and host.id is not
+	// free: on darwin the SDK shells out to ioreg, which costs ~10ms. A
+	// prometheus-only app -- the default -- must not pay that at every start for
+	// a label only OTLP ingest reads.
+	if name := strings.ToLower(strings.TrimSpace(cfg.Exporter)); name != "" && name != exporterPrometheus {
+		// host.id is the last fallback Google accepts for the required "instance"
+		// label, so detecting it keeps points from being dropped on hosts where
+		// nothing else supplies one.
+		opts = append(opts, resource.WithHostID())
+
+		if d, ok := lookupDetector(name); ok {
+			opts = append(opts, resource.WithDetectors(d))
+		}
 	}
 
 	// resource.New returns a usable resource alongside a non-nil error for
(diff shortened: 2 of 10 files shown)


pkg/gofr/metrics/exporters/provider.go
```diff
@@ -62,12 +62,43 @@ func Build(ctx context.Context, cfg *Config, logger Logger) (ShutdownFunc, metri
...
+	opts := []resource.Option{
+		// OTEL_RESOURCE_ATTRIBUTES is the standard way to supply attributes a
+		// backend requires but the framework cannot know -- Google's OTLP ingest
+		// rejects any point whose prometheus_target has no "location", and only
+		// the operator knows the region.
+		resource.WithFromEnv(),
+		// host.id is the last fallback Google accepts for the required "instance"
+		// label, so detecting it here is what keeps points from being dropped on
+		// hosts where nothing else supplies one.
+		resource.WithHostID(),
```
Umang01-hash: WithHostID() runs unconditionally here (only the vendor detector below is gated by exporter). But the PR body says host detection 'runs only when a push exporter is selected' and cites TestBuildResource_prometheusOnlySkipsHostIDDetection — neither the guard nor that test exists at head. Measured WithHostID ≈ 11ms on macOS (ioreg); and for the Prometheus-only default, host.id is dropped by WithoutTargetInfo() right after, so it's pure cost with no benefit there. Either restore the guard + test, or fix the body/cost table. (Linux reads /etc/machine-id so it's cheap in prod — but the claim is still off.)

pkg/gofr/metrics/exporters/provider.go
```diff
@@ -62,12 +62,43 @@ func Build(ctx context.Context, cfg *Config, logger Logger) (ShutdownFunc, metri
...
+		opts = append(opts, resource.WithDetectors(d))
+	}
+
+	// resource.New returns a usable resource alongside a non-nil error for
+	// partial failures (a detector that cannot reach a metadata server off-GCP,
+	// say). Degrade loudly but keep whatever was resolved rather than dropping
+	// the service name with it.
+	res, err := resource.New(ctx, opts...)
+	if err != nil {
+		logger.Warnf("metrics: resource detection was incomplete: %v", err)
```
Umang01-hash, line 120: This warns on every GCP boot — the happy path. WithHostID pulls semconv 1.43.0 (otel/sdk resolves to 1.45.0) while gcpdetect 1.44.0 uses 1.41.0, so resource.New returns ErrSchemaURLConflict on GCP: schema URL is wiped to "" but all attributes survive, and this logs 'resource detection was incomplete' even though it succeeded. Repro'd it. Suggest special-casing errors.Is(err, resource.ErrSchemaURLConflict) (data is intact) or aligning the semconv versions.

pkg/gofr/metrics/exporters/provider_test.go
```diff
@@ -74,3 +76,73 @@ func TestBuild_missingSubmoduleImportHint(t *testing.T) {
...
+}
+
+// A registered detector must only run for the exporter it belongs to, so that a
+// Prometheus-only app never reaches for a cloud metadata server.
+func TestBuildResource_detectorRunsOnlyForItsExporter(t *testing.T) {
+	var called bool
+
+	RegisterResourceDetector("detector-probe", detectorFunc(func(context.Context) (*resource.Resource, error) {
+		called = true
+		return resource.NewWithAttributes("", attribute.String("probe", "yes")), nil
```
Umang01-hash: nit: this probe uses an empty schema URL, which never conflicts in resource.Merge — so the merge test passes trivially and can't catch the 1.43-vs-1.41 conflict above. A detector case with a non-empty, differing schema URL would actually lock the invariant.

examples/using-gcp-metrics/README.md
```diff
@@ -46,11 +58,31 @@ GMP requires.
...
+## Required resource labels
+
+Google maps every point onto the `prometheus_target` monitored resource, which
+requires a **`location`** and an **`instance`**, and *rejects the point* when
+either is empty. A push that is missing them still authenticates and still
+succeeds — the points are discarded afterwards, server-side, so nothing on the
+client says the data went nowhere.
+
+On Google Cloud both are detected for you: the metadata server supplies the
+region/zone, and `host.id` backs `instance`.
```
Umang01-hash, line 75: nit: on hosts where host.id doesn't resolve (some containers/CI) 'instance' has no source and points get dropped too, but only 'location' is called out below. Worth noting OTEL_RESOURCE_ATTRIBUTES can also carry service.instance.id for those environments.

Umang01-hash (changes requested):
The fix is right and I verified it end-to-end — labels reach the resource, the location/quota warnings fire correctly, and the IAM doc change (telemetry.writer + serviceUsageConsumer, not monitoring.metricWriter) is accurate. Booted the example off-GCP too: graceful start, warning fires with no location and clears when it's set.

Two things to reconcile before merge, and two nits. Ran both modules locally: gofmt/vet/build/golangci-lint clean, all tests pass.

pkg/gofr/metrics/exporters/provider.go
aryanmehrotra: The guard and the test are both at head — they landed in 93f568db ("keep host.id detection off the prometheus-only path"), pushed at 12:17Z, ten minutes before this review. I think the diff you read was e2ca93b1, the first commit. At 98c4b638:

```go
if name := strings.ToLower(strings.TrimSpace(cfg.Exporter)); name != "" && name != exporterPrometheus {
    opts = append(opts, resource.WithHostID())

    if d, ok := lookupDetector(name); ok {
        opts = append(opts, resource.WithDetectors(d))
    }
}
```

`TestBuildResource_prometheusOnlySkipsHostIDDetection` covers `""` and `"prometheus"`.

Your measurement matches mine to within noise — 11.16ms unguarded vs 0.010ms for the pull-only default, against 0.027ms before the branch. And your point about `WithoutTargetInfo()` dropping `host.id` right afterwards is a sharper argument than the one I made in the commit message: it isn't merely unnecessary cost on that path, it's cost for a value that is then discarded. Worth having in the record.

pkg/gofr/metrics/exporters/provider.go
aryanmehrotra: Confirmed and fixed in 98c4b638. You're right on every part of it, and this was the real find in the review — the warning fired on the happy path, in the one environment the exporter exists for.

Reproduced with a detector returning each schema URL:

| detector schema | err | resulting schema | attributes |
|---|---|---|---|
| `1.41.0` (real gcpdetect) | `ErrSchemaURLConflict` | `""` | all survive |
| `""` | none | `1.43.0` | all survive |

I took the special-casing option rather than aligning the semconv versions: the alignment holds only until either dependency bumps, and a version skew between a vendor detector and the SDK is the normal state of that ecosystem rather than a fault. It's now recognised explicitly and logged at debug, saying what actually happened; genuine partial failures still warn.

Two follow-ups from your nits:

**The probe** — right, it proved nothing. The empty schema URL never conflicts in `resource.Merge`, so the assertion passed trivially. It now returns `1.41.0`, and there's a `TestBuildResource_schemaURLConflictIsNotAFailure` that asserts the attributes survive and no failure is logged.

**And that test was itself broken when I first wrote it.** It captured with `StderrOutputForFunc`, but `Warnf` is WARN, which the logger routes to `normalOut` — only ERROR and above reach stderr. So it passed against the unfixed code. Caught it by reverting the fix and finding the test still green; it now fails as it should. Same trap as the #3600 thread, from the other direction.

examples/using-gcp-metrics/README.md
aryanmehrotra: Good catch — `instance` had exactly the same hole as `location`, and only one of them was documented. Added to the README in 98c4b638, calling out minimal containers and CI runners specifically, with `service.instance.id=${HOSTNAME}` as the way to supply it.

pkg/gofr/metrics/exporters/provider_test.go
aryanmehrotra: Fixed in 98c4b638 — and you were right that it proved nothing. An empty schema URL never conflicts in `resource.Merge`, so the assertion passed no matter what the code did.

The probe now returns `https://opentelemetry.io/schemas/1.41.0`, matching what a real vendor detector supplies, and there's a dedicated `TestBuildResource_schemaURLConflictIsNotAFailure` that asserts every attribute survives the conflicting merge and that nothing is logged as a failure.

Worth recording that the new test was *also* broken on first write, in the same family of mistake: it captured with `StderrOutputForFunc`, but `Warnf` is WARN, which the logger routes to `normalOut` — only ERROR and above reach stderr. So it passed against the unfixed code too. I found it by reverting the fix and checking the test went red; it didn't. It does now:

```
--- FAIL: TestBuildResource_schemaURLConflictIsNotAFailure
    a schema-URL conflict must not be logged as a failure, got:
    "metrics: resource detection was incomplete: ... conflicting Schema URL ..."
```

Two trivially-passing tests in one PR, both invisible until deliberately falsified. Reverting the fix to watch the test fail is the only thing that caught either.

examples/using-gcp-metrics/README.md
```diff
@@ -46,11 +58,40 @@ GMP requires.
...
+Anywhere else — local runs, other clouds, CI — `location` has no source, and the
+exporter warns at startup. Set it explicitly:
+
+```
+OTEL_RESOURCE_ATTRIBUTES=location=us-central1
+```
+
+`location` also accepts `cloud.region` or `cloud.availability_zone`.
+
+`instance` is filled from `host.id`, which GoFr detects automatically. Some
```
Umang01-hash: nit / doc thought: this covers the 'no machine-id → instance rejected' case well. The other GKE-shaped case is when host.id IS present but shared — several pods per node collapse to the same node machine-id, so they'd share one prometheus_target instance (GoFr wires no k8s detector to supply the higher-priority k8s.pod.name rung). Cloud Run isn't affected (per-instance host.id). Might be worth one line steering GKE users to service.instance.id/k8s.pod.name too. Non-blocking.

Umang01-hash (dismissed):
All four of my earlier points are addressed, and I verified each fix rather than taking it on trust: reverting the host-detection guard makes TestBuildResource_prometheusOnlySkipsHostIDDetection fail on both '' and 'prometheus'; reverting the schema-conflict handling makes TestBuildResource_schemaURLConflictIsNotAFailure fail on the exact 1.43-vs-1.41 conflict. Booted the gcp example: location warning fires, exporter selects, and the false 'detection incomplete' warning is gone. gofmt/vet/build/golangci-lint clean on both modules, all tests pass.

I also went back and checked the premise against Google's v1.metrics reference — the instance chain (…→host.id→reject) and location chain (location→cloud.availability_zone→cloud.region→reject) match the code exactly, and telemetry.writer is the right role. Nice work getting a fiddly vendor spec right. One non-blocking doc thought below.

examples/using-gcp-metrics/README.md
aryanmehrotra: Fixed in 4a0277eb — and checking it turned up something worse than the doc gap, plus a different mechanism than the one here.

**Pods don't inherit the node's machine-id.** The SDK reads `/etc/machine-id` then `/var/lib/dbus/machine-id` (`host_id.go:83-93`) from the *container's own* filesystem, which comes from the image. Nothing bind-mounts the host's copy in — `/proc/self/mountinfo` inside a container shows no machine-id mount, and it isn't in the OCI default set. So pods don't collapse onto a shared node identity; they usually have no identity at all:

| base image | `/etc/machine-id` | `host.id` |
|---|---|---|
| `debian:12-slim` | absent | not detected |
| `alpine:3.20` | absent | not detected |
| `ubuntu:24.04` | present, **0 bytes** | **detected, empty** |

The last row is the trap. `readFile` returns success on an empty file and `hostIDDetector.Detect` doesn't check the value, so `host.id` arrives present and blank — which Google treats exactly as missing, and which no "is it absent?" check would catch.

So your conclusion was right and it's broader than GKE: `instance` was as unfilled as `location` across ordinary container deployments, and only `location` was checked or documented. The startup check now covers both and treats an empty attribute as absent.

One structural consequence worth flagging: to check `instance` honestly the exporter has to see the resource that will actually be exported, since `host.id` is contributed by the core package, not by this detector. `Build` now publishes the resolved resource on `Config` before any builder runs. That also makes the *location* check correct for attributes arriving from any source, which it wasn't before — it only ever saw this detector's output plus the environment.

`buildReader` is a plain function again now that it doesn't need the detector.

The collision you described does have a real form, just a rarer one: an image that bakes a *non-empty* machine-id at build time shares it across every pod from that image, cluster-wide rather than per-node. The pod-name guidance covers that too.

README now documents the container behaviour with the downward-API pod name, and drops the node-identity explanation since it doesn't hold.

Umang01-hash (approved):

PR #4139, opened 2026-09-02
fix(http,metrics,mcp): don't lose a shutdown that arrives before the server starts (#3801)

akshat-kumar-singhal:
**Description:**

Fixes item 1 of #3801: a termination signal that arrives before a server has started is lost, and
the server then comes up with nothing left to stop it.

`run.go` starts the shutdown handler before the servers (`run.go:53-55`). A signal landing in that
window is observed by a `Shutdown` that sees `srv == nil`, reports success and returns — then
`run()` publishes the server and serves forever, so `wg.Wait()` never returns and the process hangs
until SIGKILL. In practice that is a SIGTERM right after start (crashloop, fast rollout, failed
readiness), where the pod ignores graceful shutdown entirely.

```
 t0  SIGTERM arrives
 t1  shutdown g:  Shutdown() -> lock -> srv == nil -> unlock -> return nil   ("nothing to stop")
 t2  serve g:     s.srv = srv
 t3               srv.ListenAndServe()   <== serves forever
 t4  wg.Wait() never returns -> process hangs until SIGKILL
```

`Shutdown` now sets a sticky `stopped` flag under the `srvMu` it already takes, and `run()` checks
it under that lock before publishing:

```go
func (s *httpServer) Shutdown(ctx context.Context) error {
	s.srvMu.Lock()
	s.stopped = true
	srv := s.srv
	s.srvMu.Unlock()

	if srv == nil {
		return nil   // run() has not published yet — and now it will not start one
	}
	...
```

Under one lock there are only two orderings and both are correct: `run()` published first and
`Shutdown` stops a real server, or `Shutdown` marked first and no listener ever comes up. "Nothing
to stop" becomes "nothing to stop, and nothing will start." Applied to `httpServer`, `metricServer`
and `mcpServer` — five lines each, no new synchronisation primitive and no reordering of `Run()`.

**The unsafe window is only *before* the assignment.** Once the server is published, `net/http`
already handles a `Shutdown` that beats the start. Both entry points `run()` can take open with the
same guard — `if s.shuttingDown() { return ErrServerClosed }` — at **go1.26.0**, the version `go.mod`
pins: `net/http/server.go:3349` (`ListenAndServe`) and `net/http/server.go:3704`
(`ListenAndServeTLS`, the branch taken whenever `certFile`/`keyFile` are set). So this is not a
general "stop must wait for start" problem needing a lifecycle state machine; it is one unguarded
assignment, and the fix is sized to that.

**gRPC is deliberately untouched**, though #3801 lists it by implication. `RegisterService` calls
`ensureServer()`, so `g.server` is non-nil before `Run` executes, and `grpc.Server.Serve` refuses to
start after a stop: `if s.lis == nil { lis.Close(); return ErrServerStopped }`.

**Why not the `started chan struct{}` the issue proposed.** `Shutdown` would wait on it, turning
"nothing to stop" into "wait until started, then stop". But `app.httpServer` is always constructed
(`factory.go:31`) while `run()` is called only `if a.httpRegistered` (`run.go:152`), so a gRPC-only
or route-less app holds a non-nil `httpServer` that never starts — and a wait-for-started `Shutdown`
would block it for the whole `SHUTDOWN_GRACE_PERIOD`, 30s by default, on every shutdown. It would
also need an explicit release on every path that returns before publishing (cert validation failure,
blocked port) or `Shutdown` hangs forever. The sticky flag has no waiter, so those paths need
nothing. Reasoning laid out on the issue.

**Item 2 of #3801 (cron `Stop()` not joining in-flight jobs) needs no work here** — it was fixed by
#3815 (`Crontab.stop(abort <-chan struct{})`, a `WaitGroup` around the dispatch, and `App.Shutdown`
passing `ctx.Done()`). The issue predates that and was never trimmed; I have commented there. This
PR closes what remains of it.

**Root cause analysis:**

- **Symptom.** A SIGTERM in the start window is silently discarded: `Shutdown` returns `nil` having
  stopped nothing, the listener then comes up, and the process hangs in `wg.Wait()` until SIGKILL.
- **Root cause.** `srv` was doing double duty as the server handle and as "has it started?". A nil
  check carries no happens-before edge with the assignment that follows, so the mutex added in #3800
  closed the memory hazard `-race` could see and left the ordering bug untouched.
- **Why it wasn't caught.** No test called `Shutdown` before `run` — the existing shutdown tests all
  start the server first and sleep before stopping it. `-race` cannot see this class at all: it
  reports unsynchronised memory access, not a stop that establishes no ordering with the start it is
  meant to prevent.
- **Fix.** The sticky `stopped` flag above, decided under the lock both goroutines already take.
- **Prevention.** `TestShutdown_BeforeStart_AbandonsTheStart` asserts, table-driven across all three
  servers, that `run()` returns and the port is not served after an earlier `Shutdown`. It fails on
  `development` for all three:

  ```
  --- FAIL: TestShutdown_BeforeStart_AbandonsTheStart/http_server
      run() did not return after Shutdown: the server started anyway and nothing will stop it
  --- FAIL: .../metrics_server   (same)
  --- FAIL: .../mcp_server       (same)
  ```

  `TestShutdown_ConcurrentWithStart_LeavesNothingServing` races `run` against `Shutdown` and asserts
  the same invariant whichever way they interleave. Both pass at `-count=5 -race`. Which ordering
  actually occurs there is the scheduler's call, so that test is opportunistic for every one of its
  three cases — `BeforeStart` is the deterministic guard, and both tests now carry a comment saying
  so, to keep a case that looks flaky from being deleted as one that covers nothing.

None to any public API. One behavioural change worth naming for the changelog: the stop is now
**one-shot**. The `stopped` flag is one-way and terminal for the object's lifetime — nothing clears
it — so a server that has been `Shutdown` can never be `run` again. `App.Shutdown(ctx)` is exported,
so an application that calls `Shutdown` and then `Run` now gets a process that serves nothing and
says so in a log line, where today it would serve. Previously that attempt "worked" by starting a
listener nothing could stop, which is the bug, so nothing usable is lost; it becomes a stated
invariant rather than an accident, and the `stopped` field documents it. Nothing in the suite re-runs
a server after shutting it down.

**Additional Information:**

No new dependencies.

Unrelated, found while verifying: `TestContext_WriteMessageToSocket` fails under `-race` on
`development` with a `DATA RACE` at `websocket.go:41` in `(*App).WebSocket.func1` — that is #3930,
reproduced. It is not touched by this branch (confirmed by stashing these changes and re-running on
a clean tree). Everything else in `./pkg/gofr/` passes under `-race`.

**Checklist:**


aryanmehrotra (changes requested):
Reviewed at `9542d1dcb`. **The bug is real, the fix is correct, and I checked every external claim in the description rather than taking them on trust — all of them hold.** Three things to tighten before it goes in, all small; none of them is a disagreement with the approach.

### Reproduced

Both new tests fail on clean `development` and pass here:

| test | `development` | this PR |
|---|---|---|
| `TestShutdown_BeforeStart_AbandonsTheStart` | **3/3 FAIL** (http, metrics, mcp) | 3/3 pass |
| `TestShutdown_ConcurrentWithStart_LeavesNothingServing` | **2/3 FAIL** (http, metrics) | 30/30 pass |

with exactly the message you wrote:

```
run() did not return after Shutdown: the server started anyway and nothing will stop it
```

### Claims I verified at source

| claim | result |
|---|---|
| `run.go` starts the shutdown handler before the servers | holds — `startShutdownHandler` then `startAllServers` |
| `net/http` already guards a post-publish `Shutdown` | holds — `ListenAndServe` opens `if s.shuttingDown() { return ErrServerClosed }`, **and so does `ListenAndServeTLS`**, which matters because `run()` can take that path |
| gRPC refuses to start after a stop | holds — grpc **v1.83.0** `server.go:888-897`, your quote is verbatim |
| `httpServer` always constructed but `run()` conditional | holds — `factory.go:31` unconditional, `startHTTPServer` gated on `a.httpRegistered` |
| item 2 of #3801 already closed by #3815 | holds — `Crontab.stop(abort <-chan struct{})` at `cron.go:107`, `a.cron.stop(ctx.Done())` at `gofr.go:110` |

| gate | result |
|---|---|
| `gofmt` / `go build ./...` / `go vet` | clean |
| new tests x30 | no flake |
| new tests x10 under `-race` | 0 races |
| `golangci-lint --new-from-rev=origin/development` | 0 issues |

The argument against the `started chan` the issue proposed is the part I most expected to be asserted rather than shown, and it is shown — a gRPC-only app really does hold a non-nil `httpServer` that never starts, so a wait-for-started `Shutdown` would burn the full grace period on every shutdown. Good call, and worth having written down.

---

## 1. The `net/http` citation does not match what this repo builds against

The description cites `go1.27, net/http/server.go:3450`. `go.mod` pins `go 1.26.0`, and on the 1.26.3 toolchain the guard is at `server.go:3349`. The behaviour is identical, so nothing about the fix changes — but the line reference is the evidence for "the unsafe window is only *before* the assignment", which is the argument that keeps this fix five lines instead of a lifecycle state machine. Please cite the version the module actually pins so the next reader can open it and land on the guard.

While you are there: `ListenAndServeTLS` carries the same guard, and `run()` takes that branch whenever `certFile`/`keyFile` are set. Worth naming explicitly, since a reader checking only `ListenAndServe` would be left wondering about the TLS path.

## 2. `stopped` is sticky and `App.Shutdown` is public — that needs to be written down

Once `Shutdown` has been called, no later `run()` will ever bring a listener up. For the signal path that is exactly right. But `App.Shutdown(ctx)` is exported, so an application that calls `Shutdown` and then `Run` now gets a process that serves nothing and says so only in a single log line — where today it would serve.

I think the new behaviour is the correct one; my ask is that it stops being implicit. A line on the `stopped` field saying the flag is one-way and an app cannot be restarted after `Shutdown`, plus a note in the PR body so it reaches the changelog, would cover it. Right now the field comment explains why the flag exists but not that it is terminal for the object's lifetime.

## 3. The mcp case in the concurrency test does not actually guard mcp

`TestShutdown_ConcurrentWithStart_LeavesNothingServing` fails on `development` for `http server` and `metrics server` — but the `mcp server` subtest **passes** on `development`. So for mcp that subtest asserts nothing about this bug; it is timing that happens to fall the right way, and it will keep passing whether or not the mcp fix is present.

`TestShutdown_BeforeStart_AbandonsTheStart` does cover mcp deterministically, so nothing is unguarded overall — this is about the concurrency test not being the guard it appears to be. Either make the interleaving deterministic for that case, or say in a comment that the concurrent variant is opportunistic and the `BeforeStart` test is the real guard for mcp. What I would rather avoid is someone later "fixing" a flaky-looking mcp subtest by deleting it, believing it was covering something.

---

Worth saying plainly: the diagnosis here is better than the issue's own, the fix is the smallest thing that closes it, and the root-cause section correctly identifies why `-race` could never have caught this class — a stop that establishes no ordering with the start it is meant to prevent is not an unsynchronised memory access. That distinction is the whole reason this survived #3800.

Happy to approve once the three above are addressed.

akshat-kumar-singhal:
Thanks — all three addressed in `e938070d` (force-push, code change is comments only plus the message/description corrections). Point 3 I checked before writing it down and it did not reproduce here; details below, because I think the conclusion changes slightly.

## 1. Citation now matches the pinned toolchain

Fixed in the description and the commit message. Verified against `~/sdk/go1.26.0/src/net/http/server.go` rather than taking the line from the review:

| entry point | go1.26.0 | guard |
|---|---|---|
| `(*Server).ListenAndServe` | `server.go:3349` | `if s.shuttingDown() { return ErrServerClosed }` |
| `(*Server).ListenAndServeTLS` | `server.go:3704` | same |

`go.mod` pins `go 1.26.0` on `development` as of `9b9698fe`, so that is the version a reader will open. (`b6acf55c` bumped the *Dockerfile* base image 1.26 → 1.27; the module pin is untouched.) Both entry points are now named explicitly, with the note that `run()` takes the TLS branch whenever `certFile`/`keyFile` are set — you are right that a reader checking only `ListenAndServe` is left wondering.

## 2. The one-shot stop is written down

The `stopped` field comment now says the flag is one-way and terminal for the object's lifetime, that nothing clears it, and — the part that was missing — that `App.Shutdown` is exported, so `Shutdown` then `Run` yields a process that serves nothing and logs why, where today it would serve. Same paragraph added to the Breaking Changes section of the description so it reaches the changelog, and to the commit message.

## 3. The mcp subtest — it does fail on `development` here, and so does the concern, differently

I could not reproduce "the mcp subtest passes on `development`". Clean `development` worktree at `9b9698fe`, this test file copied in, `-count=5`:

| subtest | FAIL on `development` |
|---|---|
| `metrics_server` | **5/5** |
| `http_server` | **4/5** |
| `mcp_server` | **3/5** |

So mcp does exercise the losing ordering on this machine — but only 3 times in 5, and `http_server` passed once too. Two things follow, and the second is your point:

- The subtest is not inert. Deleting it would remove real coverage.
- But it is not a *guarantee* either, and neither is `http_server` — which means this is not an mcp-specific defect in the test, it is the nature of every case in it. `metrics_server` at 5/5 on my box is luck of the same kind, not a stronger assertion.

Making mcp deterministic would need a scheduling hook in `mcpServer.Run` that exists only for the test, which seems a worse trade than saying plainly what the test is. So I took your second option, generalised: both tests now carry a comment. `ConcurrentWithStart` states that it is opportunistic for all three cases, records the 5/5 · 4/5 · 3/5 measurement with the caveat that the proportions are a property of the machine rather than the code, and says that a case passing on `development` is not evidence it covers nothing. `BeforeStart` is named as the deterministic guard for all three servers. That should stop exactly the "delete the flaky-looking mcp subtest" move you were worried about, and it also stops the inverse mistake of trusting a green `ConcurrentWithStart` as proof.

If you would still rather the concurrent test carried a hard guarantee for mcp, say so and I will look at what a test-only hook would cost — but my read is that the honest comment is worth more than a hook that only the test uses.

---

Gates re-run after the change: `gofmt` clean on both touched files, `go vet ./pkg/gofr/` clean, `TestShutdown_*` green at `-count=3`. No production-code behaviour changed by this push.

aryanmehrotra (approved):
All three addressed. Re-reviewed at `9d50efe07`, and I checked the delta since my review rather than re-reading the whole diff: **19 lines, every one a comment. Zero executable lines changed.** The only production-code edit to `mcp.go` in that range is the `EnableMCP` godoc, which came from `db0a267a0` on `development`, not from you.

| # | Ask | Verified |
|---|---|---|
| 1 | Cite the pinned toolchain, name `ListenAndServeTLS` | ✅ — I opened both on go1.26.3: guard at `server.go:3349` and `server.go:3704`, exactly as cited. `go.mod` on `development` does pin `go 1.26.0`. |
| 2 | Write down that `stopped` is one-way | ✅ — `http_server.go` field comment + a Breaking Changes paragraph so it reaches the changelog |
| 3 | The mcp subtest | ✅ — and **you were right to push back; I was wrong** |

## On point 3 — my claim did not survive being measured

I asserted the `mcp_server` subtest passes on `development`. It does not. Ten runs on clean `development` at `9bbdbbe23`:

| subtest | FAIL on development |
|---|---|
| `metrics_server` | **7 / 10** |
| `mcp_server` | **2 / 10** |
| `http_server` | **1 / 10** |

`http_server` failed **less often than mcp** on my machine, which refutes the mcp-specific framing twice over. Your five runs and my ten disagree on proportions and agree on the conclusion — which is itself the point your comment now makes, that the proportions are a property of the machine.

So the generalisation is right and the comment you wrote is better than what I asked for: it says the test is opportunistic for all three cases, names `BeforeStart` as the deterministic guard, and tells the next reader to diagnose rather than delete. That closes the failure mode I was actually worried about, and it also closes the inverse one I had not thought of — reading a green `ConcurrentWithStart` as proof.

Agreed on not adding a test-only scheduling hook to `mcpServer.Run`. An honest comment beats a hook that exists only for the test.

## Gates re-run here

| Gate | Result |
|---|---|
| `gofmt` on all four files | clean |
| `go vet ./pkg/gofr/` | clean |
| `TestShutdown_*` ×3 | `ok 1.933s` |
| `TestShutdown_*` ×10 under `-race` | `ok 5.142s`, **0 races** |

CI last ran 2026-09-10, before the MinIO outage, so the green is genuine rather than stale — but it is worth a fresh run against current `development` before merge.

LGTM.

Umang01-hash (approved):
Verified against head 9d50efe. Correct, well-scoped fix — the sticky `stopped` flag under `srvMu` closes the shutdown-before-start race for the three lazy-publishing servers. Confirmed the bug is real (run.go starts the shutdown handler before the serve goroutines), revert-to-red fails for all three servers without the fix, clean under -race, and gRPC is correctly excluded (RegisterService publishes the server eagerly, so Serve returns ErrServerStopped after an early GracefulStop). No API break; gofmt/vet/build/golangci-lint all clean.

One optional nit (pre-existing lines, so noting here rather than inline): in metrics_server.go and mcp.go the "Starting … server" log sits before the `stopped` check, so in the race window the logs read "Starting…" immediately followed by "…was shut down before it started". http_server.go logs after the check — worth matching for consistency.

PR #4097, opened 2026-08-26
chore(deps): consolidate minor dependency updates (2026-08-26)

Umang01-hash:
## Consolidated Dependabot Updates (Minor/Patch)

This PR combines all minor/patch dependency updates from this week's Dependabot batch, applied per-module across the `go.work` workspace (root + affected `pkg/` submodules + `examples/using-s3-filestore`).

### Updates included
| Package | Old | New |
|---|---|---|
| go.opentelemetry.io/otel | v1.44.0 | v1.45.0 |
| go.opentelemetry.io/otel/sdk | v1.44.0 | v1.45.0 |
| go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp | v1.44.0 | v1.45.0 |
| github.com/aws/aws-sdk-go-v2/config | v1.32.37 | v1.32.38 |
| github.com/aws/aws-sdk-go-v2/service/sqs | v1.45.1 | v1.46.7 |
| google.golang.org/grpc | v1.83.0 | v1.83.1 |
| google.golang.org/api | v0.291.0 | v0.293.0 |
| cloud.google.com/go/pubsub | v1.51.0 | v1.51.1 |
| github.com/jlaffaye/ftp | v0.2.2 | v0.2.4 |
| github.com/redis/go-redis/extra/redisotel/v9 | v9.21.0 | v9.22.0 |

### Verification
- Per-module `go mod tidy -diff` clean (reproduces CI `code_quality` tidy gate over `pkg/`).
- `go build ./...` passes on root + all changed library modules; `go vet` clean on the example.

### Original Dependabot PRs (now closed)
#4096, #4095, #4094, #4093, #4092, #4091, #4090, #4089, #4088, #4087, #4086, #4085, #4084, #4083, #4082, #4081, #4080, #4079, #4078, #4077, #4076, #4075, #4074, #4073, #4072, #4071, #4070, #4069, #4068, #4067, #4066, #4065, #4063, #4062, #4061, #4060, #4059, #4058, #4057, #4056

All changes are minor/patch bumps with no breaking API changes expected.

PR #4140, opened 2026-09-02
chore(deps): consolidate minor dependency updates (2026-09-02)

Umang01-hash:
## Consolidated Dependabot Updates (Minor/Patch)

Combines this week's go-module Dependabot PRs into one, applied per-module (`go get` + `go mod tidy`). Table regenerated from the actual diff.

### Direct dependency updates
| Package | From | To | Notes |
|---------|------|----|-------|
| modernc.org/sqlite | 1.56.0 | 1.57.0 | |
| google.golang.org/api | 0.291.0 | 0.295.0 | |
| google.golang.org/grpc | 1.83.0 | 1.83.2 | supersedes the 1.83.1 group PR (#4136) |
| github.com/grpc-ecosystem/go-grpc-middleware/v2 | 2.3.3 | 2.3.4 | |
| github.com/aws/aws-sdk-go-v2/config | 1.32.37 | 1.33.1 | pulls aws-sdk-go-v2 1.43.6→1.45.1, credentials 1.19.36→1.20.1 |
| cloud.google.com/go/cloudsqlconn | 1.25.1 | 1.25.2 | root; in `examples/using-cloudsql` it moves 1.22.1→1.25.2 (indirect) |
| go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp | 1.44.0 | 1.46.0 | ⚠️ crosses the v1.45.0 `WithEndpointURL` break — see below |
| go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc | 1.44.0 | 1.46.0 | |
| go.opentelemetry.io/otel/exporters/prometheus | 0.67.0 | 0.68.0 | |
| go.opentelemetry.io/contrib/.../otelhttptrace | 0.70.0 | 0.71.0 | |
| github.com/stretchr/testify | 1.11.1 | 1.12.1 | direct dep (imported by shipped files); raises the floor for consumers via MVS |

### Notable transitive bumps
- `go.opentelemetry.io/otel` (+ `sdk`, `sdk/metric`, `metric`, `trace`) → 1.46.0
- `github.com/redis/go-redis/v9` 9.21.0 → 9.22.0 (indirect; no core/mock change — root already ships 9.22.0)
- `github.com/grpc-ecosystem/grpc-gateway/v2` 2.29.0 → 2.30.0 (indirect)
- `google.golang.org/protobuf` 1.36.11 → 1.36.12, `genproto/googleapis/rpc` refreshed

### Breaking change handled (otel v1.45.0 OTLP/HTTP endpoint)
otel v1.45.0 changed `otlpmetrichttp.WithEndpointURL` to stop appending the default `/v1/metrics` signal path — so `METRICS_URL=http://collector:4318` (scheme, no path) would silently POST to `/` and 404 every export interval. Fixed in this PR: `otlpOptions` now re-appends the signal path when a scheme-bearing URL has no path of its own, leaving explicit paths (including a bare `/`) untouched, plus a regression test asserting the POSTed path against an `httptest` server. gRPC and traces are unaffected.

### Scope notes
- No `go` directive changes — all 16 touched modules stay `go 1.26.0`.
- `go.work.sum` gains ~104 lines (workspace hashes for the new versions); no per-module `go work sync` was run — the per-module `go mod tidy -diff` gate is clean.
- otel version skew remains: `otlptrace`/`otlptracegrpc`/`zipkin` stay at 1.44.0 (predates this PR); the same `WithEndpointURL` trap will apply to `otlptracehttp` whenever traces move to HTTP.

Local verification: `go mod tidy -diff` clean per-module, `go build ./...` clean on all libraries, `go vet ./...` clean on examples, exporters package tests + lint green.

### Original Dependabot PRs (now closed)
#4136, #4135, #4134, #4133, #4132, #4131, #4130, #4129, #4128, #4127, #4126, #4125, #4124, #4123, #4122, #4121, #4120, #4119, #4118, #4117, #4115


aryanmehrotra:
Held this one for a bit because "all bumps are minor/patch — no breaking API changes expected" isn't quite true. There's one breaking change in here, and it's the silent kind: it compiles, `go vet` is clean, and the full test suite passes.

## Blocker: OTLP/HTTP metrics silently stop reaching the collector

`pkg/gofr/metrics/exporters/otlp.go:62` passes `otlpmetrichttp.WithEndpointURL`. otel **v1.45.0** changed that function to stop appending the default signal path:

> ⚠️ **Breaking Change:** `WithEndpointURL` in `otlpmetrichttp` no longer appends the default signal path when an endpoint URL has no path... If the URL has no path component, the root path (`/`) is used. Use `WithEndpointURL(url.JoinPath(endpoint, "/v1/metrics"))` to preserve the previous behavior. (#8538)

`development` pins otlpmetric at `v1.44.0`; this PR jumps to `v1.46.0`, so we cross that boundary.

I A/B'd it — identical GoFr code, only the dep set differs, recording the path the exporter actually POSTs to against an `httptest` server:

| `METRICS_URL` | development (1.44.0) | this PR (1.46.0) | |
|---|---|---|---|
| `http://collector:4318` | `/v1/metrics` | **`/`** | ⚠️ breaks |
| `http://collector:4318/` | `/` | `/` | unchanged |
| `http://collector:4318/v1/metrics` | `/v1/metrics` | `/v1/metrics` | fine |
| `collector:4318` (schemeless → `WithEndpoint`) | unaffected | unaffected | fine |

**Impact:** any user on `METRICS_EXPORTER=otlp` + `METRICS_PROTOCOL=http` with a scheme-bearing URL and no path — which is the most natural way to write it, and the shape `config.go:22` explicitly documents as supported. Their collector starts 404ing every push interval and metrics just disappear. No compile error, no config error, nothing in our tests catches it.

gRPC (the default protocol) and traces (`otlptracegrpc`, still on 1.44.0) are not affected.

### Ask

Either of these is fine by me, but one of them needs to happen before this merges:

1. **Fix it** — in `otlpOptions`, append the signal path when the parsed URL has no path, leaving an explicitly-supplied path alone. Plus a regression test asserting the posted path, since we currently have zero coverage of it. Happy to review that as part of this PR.
2. **Or drop the otel bumps from this PR** — revert `otlpmetricgrpc`/`otlpmetrichttp`/`otel`/`sdk`/`metric`/`trace`/`prometheus`/`otelhttptrace` back to what's on `development`, land the rest (which all looks fine), and do otel on its own PR where the endpoint change gets the attention it deserves.

I'd lean toward (2) — it keeps this PR the boring dependabot roll-up it's meant to be, and the otel move is worth its own review.

## Also, while we're here

**otel version skew.** Even after this, core goes to 1.46.0 while `otlptrace`, `otlptracegrpc` and `zipkin` stay at 1.44.0. That skew predates this PR (1.45 vs 1.44), so not a blocker — but worth closing out, because the exact same `WithEndpointURL` trap is sitting in `otlptracehttp` the day traces move to HTTP.

**testify `1.11.1 → 1.12.1` isn't declared in the table, and it isn't test-only for us.** It's imported by four *shipped* files — `container/sql_mock.go`, `testutil/port.go`, `testutil/wait.go`, `service/mock_oauth_server.go` — so it's a real direct dependency of `gofr.dev` and MVS forces 1.12.1 onto every downstream user. No testify type is in an exported signature so nothing breaks, and this shape predates the PR, but it does raise the floor for everyone and it should be called out rather than bundled in silently. It also drags in `go.yaml.in/yaml/v3` next to our existing `gopkg.in/yaml.v3`, so we now ship two YAML v3 parsers.

**The description doesn't match the diff**, which made this harder to review than it needed to be:
- otlpmetric listed as `1.44.0 → 1.46.0` and "no breaking API changes expected" — it crosses a ⚠️-flagged upstream break (above).
- `cloudsqlconn` listed as `1.25.1 → 1.25.2`, but `examples/using-cloudsql` actually goes `1.22.1 → 1.25.2`.
- "no `go work sync`" — but `go.work.sum` gains 104 lines.
- Undeclared bumps: testify `1.11.1→1.12.1`, `redis/go-redis/v9` `9.21.0→9.22.0`, `grpc-gateway/v2` `2.29.0→2.30.0`.

Could you regenerate the table from the actual diff? For a consolidated roll-up like this the table *is* the review surface — if it's not faithful, the whole point of consolidating is lost.

## What I verified as clean

Credit where it's due, the rest holds up:
- **No `go` directive changes** — all 16 touched modules stay `go 1.26.0`. That's the scariest vector in a bump like this and it's absent.
- Root module builds, vets and passes the full `./pkg/...` suite with `GOWORK=off` (i.e. what consumers actually resolve), workspace build across all 33 modules is clean, and `go mod tidy -diff` is clean per-module.
- `go-spew`/`go-difflib` dropping out of `go.mod` is correct — nothing imports them directly, testify 1.12 just vendored them.
- `gofr.dev v1.57.0 → v1.59.0` in `using-gcp-metrics` is inert, that module has `replace gofr.dev => ../../`.

Umang01-hash:
Good catch on the OTLP/HTTP break — reproduced it (scheme-bearing URL, no path → posts to `/` on 1.46, `/v1/metrics` on 1.44). Went with option 1 and fixed it in `69cf72a`:

- `otlpOptions` re-appends `/v1/metrics` when a scheme-bearing URL has no path, leaving an explicit path (including a bare `/`) untouched — matches the pre-1.45 behavior in your A/B table.
- Regression test asserts the POSTed path against an `httptest` server (fails without the fix on the no-path case, passes for explicit/root paths). gRPC and traces untouched.

Also regenerated the description table from the actual diff and called out everything you flagged:
- testify `1.11.1→1.12.1` (direct), redis/go-redis `9.21→9.22` (indirect), grpc-gateway/v2 `2.29→2.30` (indirect) now listed.
- cloudsqlconn corrected — `1.22.1→1.25.2` in `examples/using-cloudsql`.
- `go.work.sum` +104 lines noted (per-module `go mod tidy -diff` gate is clean; no `go work sync`).
- otel skew (`otlptrace*`/`zipkin` at 1.44) noted as a follow-up — same trap lands in `otlptracehttp` when traces move to HTTP.

No `go` directive changes. PTAL.

aryanmehrotra (commented):
#4048 is merged (`41fcf449f`), so this is next in the chain. Two things before I approve: the fix checks out, and the branch now needs a rebase.

## The fix is right, and I verified it rather than reading it

I ran the same probe I used to raise the issue — same GoFr code, only the dependency set differs, recording the path the exporter actually POSTs to against an `httptest` server:

| branch | `otlpmetrichttp` | posted path | |
| --- | --- | --- | --- |
| `development` (pre-#4048) | v1.44.0 | `/v1/metrics` | fine |
| #4048 | v1.44.0 | `/v1/metrics` | fine |
| #4097 | v1.45.0 | `/` | ⚠️ still broken there |
| **this PR** | v1.46.0 + fix | **`/v1/metrics`** | fixed |

`Test_buildOTLPExporter_httpPostsToSignalPath` fails on the unfixed 1.45.0 tree with `posted path = "/", want "/v1/metrics"` and passes here, so the regression test is not vacuous. The explicit-path and bare-`/` cases behave as the pre-1.45 exporter did. gRPC and `otlptracegrpc` are untouched, which is correct — gRPC has no signal path to lose.

## Also cleared: the Prometheus exporter bump

`otel/exporters/prometheus` goes `v0.67.0 → v0.68.0` here, and that is the exporter every GoFr user scrapes, so it is the other place a silent change could land. I diffed the actual `/metrics` output — built the provider through `exporters.Build`, recorded a counter and a histogram, scraped `promhttp.Handler()` on both trees and compared metric names, `# TYPE`, `# HELP` and label sets.

**Byte-identical.** No metric renames, no type changes, no label changes — so no `gofr_application_monitoring` dashboard work falls out of this. Worth stating explicitly, since a v0.x observability bump is exactly where that would hide.

## Rebase needed — 22 conflicts, and two of the three tidy fixes are not obvious

Merging this into `development` as it stands now conflicts on 22 files. I worked the rebase through to a green tree; this is the resolution that produces a correct one:

1. **Take `development`'s side** for all 22 conflicted `go.mod`/`go.sum` under `pkg/` and `examples/`. The submodule churn on this branch is stale-base noise — #4048 already carries a newer set for every one of them, and this PR's intent is the root module, `examples/` and the gcp exporter. The root `go.mod` merges cleanly on its own and keeps `otlpmetrichttp v1.46.0`.
2. **Then run `go mod tidy`** in exactly three modules: `pkg/gofr/datasource/cloudsql`, `pkg/gofr/datasource/file/s3` and `pkg/gofr/datasource/solr`. These three conflict on `go.mod` but *not* on `go.sum`, so step 1 alone leaves the two files describing different graphs and the tidy gate goes red on them. The other conflicted modules are fine after step 1.

What that produces, all checked:

| Check | Result |
|---|---|
| Tidy gate across every `pkg/` module | **0 untidy** (`development` is also 0, so no regression) |
| Root module build | clean |
| `pkg/gofr/metrics/exporters` suite | pass, including the new regression test |
| `otlpmetrichttp` resolves to | **v1.46.0** |
| Version downgrades vs `development` | **0** — 37 forward, none back |

## Two small things, neither blocking

Your note says the fix lives in `otlpOptions`; it is actually in `buildOTLPExporter`, just ahead of the `otlpOptions` call. The placement is right — `otlpOptions` is shared with gRPC and gRPC must not get a signal path — it is only the description that points at the wrong function.

`hasScheme` is a case-sensitive prefix test, so an endpoint written `HTTP://collector:4318` skips both the fix and `WithEndpointURL` and falls through to `WithEndpoint`. That predates this PR and is unchanged by it, so leave it — noting it only so it is on the record next to the endpoint handling it sits in.

## On merge order

Land this before #4097. I have already flagged there that #4097 crosses to `otlpmetrichttp v1.45.0` without the fix, so if it goes first the break reaches `development` in the gap. With 1.46.0 already in, its 1.45.0 line is superseded on rebase and nothing lands broken.

To be explicit for whoever picks these up: **#4097 should not be closed as superseded by this PR.** This one leaves all 27 datasource submodules on otel 1.44.0, and #4097 carries 76 shipped-module bumps this does not — including the root-module `redisotel`/`rediscmd` `9.21.0 → 9.22.0` skew fix.

Rebase with the two steps above and I will approve.

Issue #4137, opened 2026-09-02
metrics/gcp: buildResource emits no location or instance, so every point is rejected by the Telemetry API


akshat-kumar-singhal:
Follow-up to #4112 (point 2 of the closing comment). Filing with the detail, as offered.

Thanks for the fast turnaround on the tag and the vanity route — both verified here off the public proxy:

```
$ curl -s https://proxy.golang.org/gofr.dev/pkg/gofr/metrics/exporters/gcp/@v/list
v0.1.0
$ curl -s -o /dev/null -w '%{http_code}\n' 'https://gofr.dev/pkg/gofr/metrics/exporters/gcp?go-get=1'
200
```

A throwaway module with nothing but the blank import resolves and builds against `gofr.dev v1.60.0` + the exporter `v0.1.0`. The module is consumable. It still cannot deliver a metric point.

## Summary

`buildResource` emits no `location` and no `instance`, and the Telemetry API rejects points that lack either. `METRICS_EXPORTER=gcp` therefore exports "successfully" — no error, no warning, a healthy-looking `pushReader` — while 100% of points are discarded server-side.

## The requirement

Google's Telemetry API v1 metrics reference documents the OTLP → `prometheus_target` label mapping. For two labels the documented behaviour when the value resolves empty is **"Reject the point"**:

| label | resolved from | if empty |
|---|---|---|
| `location` | `location`, `cloud.availability_zone`, `cloud.region` | **Reject the point** |
| `instance` | `instance`, `service.instance.id`, `faas.instance`, pod/container or host identifiers | **Reject the point** |
| `cluster` | `cluster`, `k8s.cluster.name`, platform defaults (`__gce__`, `__run__`, `__vae__`) | empty string |
| `namespace` | `namespace`, `k8s.namespace.name`, `service.namespace` | empty string |
| `job` | `job`, service name/namespace combinations, k8s workload names | empty string |

Note that `cluster` has a documented Cloud Run default (`__run__`) and the two rejecting labels do not.

## What GoFr sends

`pkg/gofr/metrics/exporters/provider.go:65-71` — identical in v1.59.0 and v1.60.0:

```go
func buildResource(cfg *Config) *resource.Resource {
	return resource.NewWithAttributes(
		semconv.SchemaURL,
		semconv.ServiceNameKey.String(cfg.AppName),
		attribute.String("framework_version", version.Framework),
	)
}
```

`service.name` and `framework_version`. Neither appears in either rejecting label's resolution chain.

[...]

PR #4143, opened 2026-09-02
feat(metrics): resolve resource attributes through the GoFr config layer

akshat-kumar-singhal:
**Description:**

Addresses #4144.

**This PR no longer claims to fix a bug.** The original description said a value in
`configs/.env` could not reach the metrics resource because `resource.WithFromEnv()` calls
`os.Getenv`. That was wrong. `config.EnvLoader` loads the file with `godotenv.Load`
(`pkg/gofr/config/godotenv.go:46`), godotenv v1.5.1 `os.Setenv`s every key not already in
the environment (`godotenv@v1.5.1/godotenv.go:199`), and `EnvLoader.Get` is `os.Getenv`
(`godotenv.go:93-95`) — so by the time `buildResource` runs, `configs/.env` *is* the process
environment. @aryanmehrotra demonstrated this on unpatched `development` with a live OTLP
receiver, and it checks out in the code. There is no gap to close.

What is left is a naming/consistency addition and one ordering fix.

**`METRICS_RESOURCE_ATTRIBUTES`.** `METRICS_URL`, `METRICS_HEADERS` and `METRICS_TEMPORALITY`
all have GoFr-native spellings; resource attributes were the exception. This adds the
spelling, resolved through `conf.Get`. Going through the config layer rather than `os.Getenv`
is what makes it work for a `config.Config` that is not backed by the environment —
`container.NewContainer` accepts any implementation, and a Consul- or Vault-backed one never
reaches `os.Getenv`, so `WithFromEnv` cannot see it. Nothing in-repo does that today; it is
offered as the real, weaker justification rather than the false one this PR opened with.

The container does **not** read `OTEL_RESOURCE_ATTRIBUTES`. The SDK already reads it from the
process environment, and duplicating that was what dragged the env var into the override path
below.

**Ordering: GoFr's identity attributes go last.** `resource.New` is last-wins. As first
written, config-sourced attributes were appended after `resource.WithAttributes(attrs...)`,
so an operator attribute string could rename `service.name` and restamp `framework_version` —
behaviour that does not exist on `development`. `buildResource` now applies
`WithFromEnv()` → config attributes → GoFr's own attributes, so config still wins per key for
operator-owned keys and cannot touch GoFr's identity.

None. `exporters.Config` gains a field. `OTEL_RESOURCE_ATTRIBUTES` behaves exactly as it does
on `development` — this time verified rather than asserted.

**Additional Information:**

- No new dependencies — `go.mod` and `go.sum` are untouched.
- Mutation-checked: restoring the old option ordering makes
  `TestBuildResource_identityAttributesAreNotOverridable` fail on `service.name` specifically.
  That test sets `service.name` from both `OTEL_RESOURCE_ATTRIBUTES` and
  `METRICS_RESOURCE_ATTRIBUTES` and asserts neither wins.
- The symptom in #4137 was derived from reading `provider.go`, not observed in a deployment.
  A prometheus-only app dropping `target_info` via `WithoutTargetInfo()` (`provider.go:132`)
  is a separate question this PR does not address and I cannot confirm.

**Checklist:**


aryanmehrotra:
## Review — #4143 `feat(metrics): resolve resource attributes through the GoFr config layer`

Reviewed at head `8887c665a` against `development` `c18187485`, by planning #4144
independently first and comparing. All measurements below were run locally on both
branches; commands and outputs are at the end.

**Changes requested.** The bug this PR describes does not exist, and the code written to
fix it introduces a real one: after this change `OTEL_RESOURCE_ATTRIBUTES` can overwrite
GoFr's own `service.name` and `framework_version`, which it could not before.

Nothing below is about care or craft — the parser, the percent-decoding choice and the test
table are all better than they need to be. The problem is upstream of the code.

---

### The premise, measured

The PR and #4144 both rest on: *`resource.WithFromEnv()` calls `os.Getenv`, so a value in
`configs/.env` is silently ignored.*

`configs/.env` **is** the process environment by the time anything reads it.
`config.EnvLoader.read` calls `godotenv.Load` (`pkg/gofr/config/godotenv.go:46`), and
godotenv v1.5.1 `godotenv.go:197-201` does `os.Setenv(key, value)` for every key not already
present. `EnvLoader.Get` is literally `os.Getenv` (`godotenv.go:93-95`). `conf.Get` and
`resource.WithFromEnv()` read the same map.

Run on **unmodified `development`** (`c18187485`), using #4144's own repro:

```
configs/.env:
  METRICS_EXPORTER=gcp
  OTEL_RESOURCE_ATTRIBUTES=location=asia-south1,cluster=a%20b
process env: OTEL_RESOURCE_ATTRIBUTES unset

resource attrs after exporters.Build:
  cluster          = a b          <- percent-decoded by the SDK
  framework_version = dev
  host.id          = B869AC7C-...
  location         = asia-south1  <- present, with no patch applied
  service.name     = spike
```

So the reported failure is not reproducible on the path the ticket describes.

Worth recording alongside it: on the **default** path (no `METRICS_EXPORTER`) resource
attributes are not on `/metrics` at all and never have been —
`prometheus.WithoutTargetInfo()` (`pkg/gofr/metrics/exporters/provider.go:132`) suppresses
`target_info`, and `target_info` is the only carrier of resource attributes for the pull
exporter (`exporters/prometheus v0.67.0`, `exporter.go:232-233`). If anyone saw the symptom
on a pull-only app, that is the cause, and it is a different ticket.

---

### Findings

**1 — `OTEL_RESOURCE_ATTRIBUTES` can now overwrite `service.name` and `framework_version`.**

`buildResource` builds `opts` in application order and `resource.New` merges last-wins. The
new block is appended *after* `resource.WithAttributes(attrs...)`, so config-sourced
attributes now outrank GoFr's own identity attributes. Since the container copies
`OTEL_RESOURCE_ATTRIBUTES` into that same string, the environment variable inherits the win.

With `OTEL_RESOURCE_ATTRIBUTES=service.name=hijacked,framework_version=99` and
`APP_NAME`-derived `AppName="real-app"`:

| attribute | `development` | this PR |
|---|---|---|
| `service.name` | `real-app` | **`hijacked`** |
| `framework_version` | `dev` | **`99`** |

`service.name` is the identity every dashboard, alert and quota keys on, and
`framework_version` is meant to be stamped by the framework. This contradicts the PR's
"Breaking Changes: None … `OTEL_RESOURCE_ATTRIBUTES` in the process environment keeps
working exactly as it does today". It also splits the two spec-standard ways of naming a
service: `OTEL_SERVICE_NAME` still loses to `APP_NAME` (it arrives via `WithFromEnv`),
while `OTEL_RESOURCE_ATTRIBUTES=service.name=…` now wins.

*Cheapest fix:* move the new `if cfgAttrs := …` block **above**
`resource.WithAttributes(attrs...)` in the `opts` slice. Config still beats `WithFromEnv`,
GoFr's own two attributes stay authoritative, and the added tests still pass as written.

**2 — The `OTEL_RESOURCE_ATTRIBUTES` half of this change is a no-op that carries the risk in
finding 1, and the prose asserting otherwise will outlive the PR.**

The claim appears in four places that a future maintainer will read as settled fact:
`exporters/config.go` (`"…which the SDK cannot see"`), the comment above
`TestBuildResource_configAttributesReachTheResource`, the new sentence in
`docs/references/configs/page.md`, and the commit message.

*Cheapest fix:* have `metricsResourceAttributes` read `METRICS_RESOURCE_ATTRIBUTES` only,
leave `OTEL_RESOURCE_ATTRIBUTES` to the SDK where it already works, and correct the four
statements. That deletes finding 1's exposure for the standard variable and leaves the PR
with one clean, honest claim: GoFr gains a native spelling.

**3 — "Config wins per key" is not true for detector-owned keys, and `host.id` is the one
that matters.**

`WithHostID()` and the vendor detector are appended after the new block, so they win.
Measured on this branch with `Exporter: "otlp"`, `ResourceAttributes: "host.id=from-config"`:
`host.id = B869AC7C-90AD-5578-BB83-D695095A6B58` — the config value is silently discarded.
`host.id` is precisely the fallback #4113 added to fill Google's required `instance` label,
so an operator following `gcp.go:205-207` and pinning it by hand gets no error and no effect.
(Pre-existing for the env var; this PR states the guarantee, so it now owns it.)

*Cheapest fix:* one sentence in the `ResourceAttributes` doc comment naming the exception,
or move the detector options ahead of the config attributes if the guarantee is meant.

**4 — `METRICS_RESOURCE_ATTRIBUTES` is a metrics-scoped name for a process-scoped concept —
an ADR, not a bug fix.**

The cited precedent (`metrics_exporter.go:28-33`) covers `OTEL_METRIC_EXPORT_INTERVAL` and
`OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE`, both metrics-scoped in the OTel spec.
`OTEL_RESOURCE_ATTRIBUTES` describes the process and feeds traces and logs too, so the
precedent does not extend to it — and a published framework config key cannot be withdrawn.
The next ticket asks for `TRACER_RESOURCE_ATTRIBUTES`. Relatedly, `gcp.go:192` reads
`os.Getenv("OTEL_RESOURCE_ATTRIBUTES")` directly and both its warnings (`gcp.go:197`, `:207`)
name only that spelling, so the alias would ship half-known to the codebase.

**5 — A second parser for a format the SDK already parses (minor).**

`exporters/resource.go:parseResourceAttributes` reimplements
`sdk/resource/env.go:constructOTResources` (otel sdk v1.45.0, lines 67-95). They already
differ: the SDK keeps `k=` as `k=""`, this one skips it. Harmless today only because both
run on the same string; the divergence becomes real on any SDK bump. The repo has been bitten
by this shape before (`docs/memory/patterns.md`, the `isBindableMediaType` line).

---

### Omissions — test-plan rows the diff has no answer for

| Dimension | Status |
|---|---|
| security / privilege | **not covered** — nothing asserts that an operator-supplied attribute string cannot overwrite `service.name` or `framework_version`. This is the row that would have caught finding 1. |
| regression | **not covered** — no before/after for process-env `OTEL_RESOURCE_ATTRIBUTES`, which the description asserts is unchanged. |
| correctness (AC3) | **partial** — `Test_metricsExporterConfig_resourceAttributes` asserts the concatenated *string*, never the resulting attribute. The end-to-end "GoFr-native wins per key" claim is untested. (I checked it: it does hold — `attribute.NewSet` is last-wins.) |

---

### Where the PR is right and my own plan was weaker

- **A custom `config.Config` is a real case I scoped out.** `container.NewContainer` takes any
  `config.Config` (`container.go:100`), and a Consul- or Vault-backed implementation would
  reach `conf.Get` but never `os.Getenv`. Under my plan its attributes stay dropped; under
  this PR they work. No in-repo production path does that today (`EnvLoader` and the
  test-only `MockConfig` are the only implementations), so it does not rescue the premise —
  but it is a genuine reason the mechanism could be wanted, and it is a better argument for
  this PR than the one the description makes.
- **The alias is what #4144 asked for.** Routing it to an ADR is my judgement, not a defect
  in the PR.

### Differences I am not flagging

- **Concatenating `otel + "," + gofr` instead of merging maps** — verified last-wins;
  cheaper and shorter than a merge, and it preserves `OTEL_` keys the GoFr value omits.
  Better than what my own plan would have done.
- **`url.PathUnescape` rather than `QueryUnescape`, so `+` stays literal** — exactly matches
  `sdk/resource/env.go:81`. A detail most implementations get wrong.
- **Skipping malformed pairs individually** rather than discarding the whole string.
- **Parser in `exporters/` rather than `container/`** — keeps format knowledge next to its
  only consumer.
- **Doc entry in `docs/references/configs/page.md`** — the right file; the repo documents
  every config key there.
- **The self-correction in the description** about `resource.Merge(resource.Environment(), …)`
  — that was right, and it is the same reasoning one step short of finding the premise.

---

### If you want the smallest shape that is defensible

Keep `METRICS_RESOURCE_ATTRIBUTES`, drop `OTEL_RESOURCE_ATTRIBUTES` from the container
resolution, insert the config attributes **before** GoFr's own `WithAttributes`, and fix the
four prose claims. That is a ~10-line diff, keeps every test in the PR, removes findings 1,
2 and 5, and leaves only the naming question (finding 4) for a maintainer to answer.

---

### The fix, applied and measured

Both changes were applied to a scratch checkout of `8887c665a` and run.

**Fix A — identity attributes last** (`exporters/provider.go`): `opts` becomes
`WithFromEnv()` → config attributes → `WithAttributes(attrs...)`.

**Fix B — container resolves the GoFr-native spelling only**
(`container/metrics_exporter.go`): `metricsResourceAttributes` returns
`strings.TrimSpace(conf.Get("METRICS_RESOURCE_ATTRIBUTES"))`.

| check | result |
|---|---|
| `service.name` / `framework_version` no longer overridable by `OTEL_RESOURCE_ATTRIBUTES` | pass |
| config still beats the environment for operator keys (`location`) | pass |
| unmentioned env keys still survive (`cloud.region` from env) | pass |
| duplicate key inside one string is still last-wins | pass |
| end-to-end: `OTEL_RESOURCE_ATTRIBUTES` **only** in `configs/.env` still reaches the resource, with the SDK as the only reader | pass |
| end-to-end: `METRICS_RESOURCE_ATTRIBUTES` still wins per key over `OTEL_RESOURCE_ATTRIBUTES` | pass |
| the PR's own `exporters` and `container` suites | pass |

Fix B costs exactly two assertions, and they are the two that were asserting the redundant
path: `Test_metricsExporterConfig_resourceAttributes/OTEL_RESOURCE_ATTRIBUTES_is_honored`
and `/both_are_concatenated_with_the_GoFr-native_value_last`. Rewritten to the new contract
(`""` and `"cloud.region=europe-west1"` respectively), the table passes.

### CI and test evidence

- PR CI at `8887c665a`: 13 checks green (Code Quality, typos, PKG/Example/Submodule unit
  testing on Go 1.24/1.25/1.26, Code Coverage, Snyk).
- Local `go test ./pkg/... -short -count=1` on the PR branch: two package failures, neither
  attributable to this PR — `pkg/gofr` fails identically on unmodified `development`, and
  `pkg/gofr/service`'s `TestCircuitBreaker_MixedHTTPMethods` passes 3/3 on re-run and is in
  a package this PR does not touch.

aryanmehrotra:
Following up on my review with the same method you used to verify #4113 — a real
`gofr.New()` app, a real `configs/.env`, and a local OTLP receiver reading what is
actually on the wire. It settles both points without anyone having to trust a unit test.

The receiver is ~60 lines and the app is one call; the whole thing is below if you want to
run it yourself.

**`configs/.env`** (and `OTEL_RESOURCE_ATTRIBUTES` is *not* exported into the process —
the program refuses to run if it is):

```
APP_NAME=repro-app
METRICS_EXPORTER=otlp
METRICS_URL=localhost:41777
METRICS_INSECURE=true
METRICS_EXPORT_INTERVAL=1
OTEL_RESOURCE_ATTRIBUTES=location=asia-south1,service.name=hijacked
```

**Resource attributes received over OTLP:**

| | `location` | `service.name` |
|---|---|---|
| `development` (unpatched) | **`asia-south1`** | `repro-app` |
| this PR, head `8887c665a` | `asia-south1` | **`hijacked`** |
| this PR + the reorder below | `asia-south1` | `repro-app` |

Two things fall out of that first row.

**1. `configs/.env` already reaches the resource.** `location=asia-south1` is on the wire
on unmodified `development`, with nothing exported into the process. The reason is
`config.EnvLoader`: it loads the file with `godotenv.Load` (`pkg/gofr/config/godotenv.go:46`),
and godotenv v1.5.1 `os.Setenv`s every key that is not already in the environment
(`godotenv.go:197-201`), so by the time `buildResource` runs the file *is* the process
environment — and `EnvLoader.Get` is literally `os.Getenv` (`godotenv.go:93-95`). There is
no gap between `conf.Get` and `resource.WithFromEnv()` to close.

Negative control, so the receiver is not just always printing it: delete that one line from
`configs/.env` and `location` disappears from the output.

**2. The PR changes `service.name`, which I think is unintended.** `resource.New` applies
options last-wins, and the new block lands after `resource.WithAttributes(attrs...)`, so
config-sourced attributes now outrank GoFr's own identity attributes — and since the
container copies `OTEL_RESOURCE_ATTRIBUTES` into that same string, the environment variable
inherits the win. `framework_version` moves the same way. That is the one line in the
description I would not want to ship as written ("`OTEL_RESOURCE_ATTRIBUTES` in the process
environment keeps working exactly as it does today").

**The smallest change that fixes it** — put GoFr's own attributes last:

```go
opts := []resource.Option{
	resource.WithFromEnv(),
}

if cfgAttrs := parseResourceAttributes(cfg.ResourceAttributes); len(cfgAttrs) > 0 {
	opts = append(opts, resource.WithAttributes(cfgAttrs...))
}

// GoFr's own identity attributes go last, so no operator-supplied attribute
// string can rename the service or restamp the framework version.
opts = append(opts, resource.WithAttributes(attrs...))
```

Every test in the PR still passes with that, including
`TestBuildResource_configAttributesWinOverEnvironment` — config still beats the environment
for operator-owned keys, which is the behaviour you were after.

**Where I would rather ask than decide.** Credit where it is due first: the parser is
careful — `PathUnescape` rather than `QueryUnescape`, so a literal `+` survives, matching
`sdk/resource/env.go:81` exactly — the concatenation trick for per-key precedence is neat
and I checked that it does hold, and the `resource.Merge(resource.Environment(), …)`
self-correction in your description was one step short of finding all of this yourself.

So rather than tell you what this should become, three questions:

1. **Do you still want `METRICS_RESOURCE_ATTRIBUTES`?** With the `.env` gap gone, it is a
   naming/consistency feature rather than a fix. There is an argument for it —
   `METRICS_URL`, `METRICS_HEADERS`, `METRICS_TEMPORALITY` all have GoFr-native spellings,
   so this one is the odd one out. There is an argument against — `OTEL_RESOURCE_ATTRIBUTES`
   is process-scoped in the spec (it feeds traces and logs too, unlike the per-signal
   `OTEL_METRIC_EXPORT_INTERVAL` the current fallbacks alias), so a `METRICS_`-prefixed name
   arguably claims something wider than metrics and invites `TRACER_RESOURCE_ATTRIBUTES`
   next. I lean toward it being worth a short discussion before it lands, but I would rather
   hear your read than assume it.

2. **If it stays, should the container still read `OTEL_RESOURCE_ATTRIBUTES` too?** It is
   the half that is redundant with the SDK and the half that drags the identity keys into
   the override path. Reading only `METRICS_RESOURCE_ATTRIBUTES` there is a two-line change
   and costs you two assertions in `Test_metricsExporterConfig_resourceAttributes` — but I
   may be missing a case you had in mind. One I thought of after the fact and could not
   dismiss: `container.NewContainer` takes any `config.Config`, so an app with a
   Consul- or Vault-backed implementation would reach `conf.Get` and never `os.Getenv`.
   Nothing in-repo does that today, but if that was the case you were building for, say so —
   it is a better argument for this mechanism than the one in the description, and it would
   change my mind about the shape.

3. **Is there a deployment where you actually saw the original symptom?** If you hit this
   somewhere real, I would like to see that setup — if it was a prometheus-only app then it
   is `WithoutTargetInfo()` at `provider.go:132` suppressing `target_info` and a genuinely
   different bug worth its own issue, and I would rather find that than have you close this
   thinking you imagined it.

The one thing I would ask for regardless of the answers is the option reorder above — that
one is independent of what the PR ends up being for.

<details>
<summary>The reproduction (go.mod, configs/.env, main.go)</summary>

`go.mod` — point the replace at whichever checkout you are testing:

```
module repro

go 1.24.0

require (
	go.opentelemetry.io/proto/otlp v1.10.0
	gofr.dev v0.0.0
	google.golang.org/grpc v1.83.0
)

replace gofr.dev => /path/to/gofr
```

`main.go`:

```go
// Reproduction for #4144 / #4143: is OTEL_RESOURCE_ATTRIBUTES set ONLY in
// configs/.env carried on the wire by an unmodified GoFr?
//
// Starts a local OTLP metrics receiver, boots a real gofr.New() app pointed at
// it, and prints the resource attributes of the first export it receives.
package main

import (
	"context"
	"fmt"
	"net"
	"os"
	"sort"
	"time"

	collector "go.opentelemetry.io/proto/otlp/collector/metrics/v1"
	"google.golang.org/grpc"

	"gofr.dev/pkg/gofr"
)

type receiver struct {
	collector.UnimplementedMetricsServiceServer
	got chan []string
}

func (r *receiver) Export(_ context.Context, req *collector.ExportMetricsServiceRequest) (
	*collector.ExportMetricsServiceResponse, error) {
	for _, rm := range req.GetResourceMetrics() {
		var attrs []string
		for _, kv := range rm.GetResource().GetAttributes() {
			attrs = append(attrs, kv.GetKey()+"="+kv.GetValue().GetStringValue())
		}

		sort.Strings(attrs)

		select {
		case r.got <- attrs:
		default:
		}
	}

	return &collector.ExportMetricsServiceResponse{}, nil
}

func main() {
	if _, ok := os.LookupEnv("OTEL_RESOURCE_ATTRIBUTES"); ok {
		fmt.Println("REFUSING: OTEL_RESOURCE_ATTRIBUTES is exported into the process; " +
			"this repro is only meaningful when it lives ONLY in configs/.env")
		os.Exit(2)
	}

	lis, err := net.Listen("tcp", "localhost:41777")
	if err != nil {
		panic(err)
	}

	rec := &receiver{got: make(chan []string, 4)}
	srv := grpc.NewServer()
	collector.RegisterMetricsServiceServer(srv, rec)

	go func() { _ = srv.Serve(lis) }()

	dotenv, _ := os.ReadFile("configs/.env")
	fmt.Printf("configs/.env:\n%s\n", dotenv)
	fmt.Println("OTEL_RESOURCE_ATTRIBUTES in the process environment: NO")
	fmt.Println()

	_ = gofr.New() // builds the container, the MeterProvider and the push reader

	select {
	case attrs := <-rec.got:
		fmt.Println("resource attributes received over OTLP:")

		for _, a := range attrs {
			fmt.Println("   ", a)
		}
	case <-time.After(20 * time.Second):
		fmt.Println("no export received within 20s")
		os.Exit(1)
	}
}
```

Run with `go run .` from that directory. `METRICS_PORT` / `HTTP_PORT` are set only to keep
it off the defaults if you already have a GoFr app running.

</details>

akshat-kumar-singhal:
You are right on both counts, and I checked each one against the code rather than taking
it from the repro.

**1. The `.env` gap does not exist.** `EnvLoader.read` calls `godotenv.Load`
(`pkg/gofr/config/godotenv.go:46`), godotenv v1.5.1 `os.Setenv`s every key it does not
find already set (`godotenv@v1.5.1/godotenv.go:199`), and `EnvLoader.Get` is `os.Getenv`
(`godotenv.go:93-95`). By the time `buildResource` runs, `configs/.env` *is* the process
environment, so `resource.WithFromEnv()` sees it. The premise of #4137 and of this PR's
description was wrong — that is the second correction this line of work has needed, and
this one was checkable in-repo the whole time.

**2. The reorder is a real regression and it is applied.** `resource.New` is last-wins,
the new block landed after `resource.WithAttributes(attrs...)`, and the container was
copying `OTEL_RESOURCE_ATTRIBUTES` into that same string — so the env var could rename
`service.name` and restamp `framework_version`, which it cannot do on `development`. The
sentence in my description claiming process-env behaviour was unchanged was simply false.
GoFr's identity attributes now go last, exactly as you wrote it, and there is a new test
(`TestBuildResource_identityAttributesAreNotOverridable`) that sets `service.name` from
*both* sources and asserts neither wins, so this cannot regress silently again.

**Your three questions.**

**1. Yes, but only as the naming feature it actually is.** With the gap gone this is not a
fix, and the description no longer claims it is. Your argument against — `OTEL_RESOURCE_ATTRIBUTES`
is process-scoped and feeds traces and logs too, so a `METRICS_` prefix over-claims — is
the strongest thing said about this PR, and I do not have a rebuttal that survives contact
with it. What keeps me on the "ship it" side is narrow: the field is on `exporters.Config`
and is consumed only by the metrics resource, so the name describes its actual reach, and
`METRICS_URL` / `METRICS_HEADERS` / `METRICS_TEMPORALITY` set the precedent that a GoFr
deployment can configure this surface without learning the OTEL spellings. If a
`TRACER_RESOURCE_ATTRIBUTES` ever appears, that is the point at which a shared
`RESOURCE_ATTRIBUTES` is the right answer, and it would be a clean supersede rather than a
mess. If you still read it the other way after that, say so and I will close this — the
reorder is the only part I would then want to keep, and it can go on its own.

**2. `METRICS_RESOURCE_ATTRIBUTES` only. Done.** The container no longer reads
`OTEL_RESOURCE_ATTRIBUTES` at all; the SDK keeps reading it from the process environment
and nothing duplicates that. It also removes the mechanism by which the env var reached the
override path in the first place, so the two halves of your review fix each other.

And yes — the non-environment `config.Config` case is the argument I should have led with,
so I have written it into the code comment as the reason the value goes through `conf.Get`
rather than `os.Getenv`. `container.NewContainer` takes any `config.Config`; a Consul- or
Vault-backed one never touches `os.Getenv` and `WithFromEnv` cannot see it. Nothing in-repo
does that today, so I am putting it forward as the honest, weaker justification it is, not
as a retrofit.

**3. No. There is no real deployment.** #4137 came from reading `provider.go`, not from
points being rejected anywhere. I have not run this against Google's ingest, so I cannot
confirm or rule out the `WithoutTargetInfo()` lead at `provider.go:132` — if you think that
one is worth chasing it needs someone with an actual prometheus-only app on GCP, and it
should not ride on this PR.

Thank you for the repro — the negative control, in particular, is what made the first point
impossible to argue with.

aryanmehrotra:
Thanks for the rewrite of this description — withdrawing the bug claim yourself once it did not hold up, and re-deriving the weaker justification honestly instead of quietly keeping the original framing, is the right way to handle it. That part I want to acknowledge before anything else.

I have been through what is left, and I would like to ask you directly whether you still want it, because I am no longer sure what it buys us. If you have a use in mind that I am missing, say so and I will review it properly — I would rather ask than close something you have a reason for.

## What I checked

**The `.env` path already works**, which we both agree on now — `godotenv` exports every key into the process before the container is built, so `resource.WithFromEnv()` sees `OTEL_RESOURCE_ATTRIBUTES` from `configs/.env` without any change.

**The non-environment `config.Config` argument does not have a path today.** I looked for one:

| | |
|---|---|
| a `config.Config` implementation not backed by the environment | none in `pkg/gofr/config/` — only `EnvLoader`, whose `Get` is `os.Getenv`, and the mock |
| a way for an application to supply its own | `factory.go:17-19` — `New()` calls `readConfig` and then `container.NewContainer(app.Config)` immediately, so the metrics resource is built before any user code runs. Assigning `app.Config` after `gofr.New()` returns is too late. |
| anything in-repo doing it | none, as you say in the description |

So the Consul/Vault case is not just unused, it is currently unreachable — there is no seam to inject that config through.

**The value is read once.** `buildResource` has a single call site (`provider.go:43`), reached once from `container.NewContainer`. The resource is frozen into the `MeterProvider` for the life of the process. Even from a remote config store, changing the value would not take effect until restart — so this is not a dynamic-configuration feature.

**With the key unset, behaviour is identical to `development`.** `cfgAttrs` is empty, the option list is unchanged, and GoFr's identity attributes were already last. That is every existing user.

## Which leaves the naming argument

That one is real, and it is the part I would not dismiss: `METRICS_URL`, `METRICS_HEADERS`, `METRICS_EXPORT_INTERVAL` and `METRICS_TEMPORALITY` all have GoFr-native spellings and resource attributes were the one exception. If the rule we want is "every OTel variable we care about gets a GoFr name", this completes the set.

My hesitation is that it is a second spelling for something that already works, and a config key is permanent — it goes in the reference, people set it, and we support it for as long as GoFr exists. 293 lines and a standing support surface is a lot for symmetry alone.

## What I would like from you

**Do you want to keep it?** Three ways this can go, and I am happy with any of them:

1. **You have a use I have not seen.** If there is a deployment shape where `OTEL_RESOURCE_ATTRIBUTES` genuinely cannot reach the resource, tell me what it is and I will review this properly — that would change my read entirely.
2. **Keep only the ordering guard.** The ~14 lines in `buildResource` that stop an operator attribute string from restamping `service.name` and `framework_version` are worth having on their own merits, independent of the new key. That drops roughly 280 of the 293 lines and leaves a small, clearly-motivated change.
3. **Close it.** No criticism attached if so. The investigation was sound and it produced a correct answer, which happens to be "there was nothing here" — that is a useful outcome and #4144 can be closed with it.

I am leaning towards 2 or 3, but I would rather hear your reasoning first, especially since you have spent real time on this one. Take whichever you think is right and I will go with it.

One practical note either way: the branch now conflicts with `development` in `pkg/gofr/container/metrics_exporter.go`, so it needs a rebase before anything can merge.

akshat-kumar-singhal:
Option 3 — close it. Your read is right, and having gone back through where this came from, it is right for a stronger reason than symmetry-versus-support-surface.

**The original motivation was satisfied by the fix that already merged.** #4137 was filed because `buildResource` emitted only `service.name` and `framework_version`, and `OTEL_RESOURCE_ATTRIBUTES` appeared nowhere in the module at all — so on Cloud Run under Google Managed Prometheus, `location` and `instance` resolved empty and the Telemetry API's `prometheus_target` mapping rejects the point. The fix #4137 asked for was `resource.WithFromEnv()`. That is what landed, and it is what closed the gap.

`METRICS_RESOURCE_ATTRIBUTES` is scope that accreted onto the branch afterwards, once the bug claim had already been withdrawn. To answer your question directly: **no, I do not have a deployment shape you are missing.** The `config.Config`-backed-by-Consul-or-Vault justification in the description was reasoning I constructed to keep the change alive, not a case I had actually hit — and as you found, there is no seam to reach it through even in principle.

## One correction, on option 2

The ordering guard does not exist independently of the key, so option 2 collapses into option 3.

On `development`, `buildResource` already reads:

```go
opts := []resource.Option{
    resource.WithFromEnv(),
    resource.WithAttributes(attrs...),
}
```

`resource.New` applies options in order and `Merge(a, b)` lets `b` overwrite `a`, so `service.name` and `framework_version` already win over an operator's `OTEL_RESOURCE_ATTRIBUTES` string today. The ~14 lines in this PR only order `attrs` after `cfgAttrs` — the attributes `METRICS_RESOURCE_ATTRIBUTES` itself introduces. Drop the key and there is nothing left for the guard to guard. So there are not 13 salvageable lines here; there are zero, and that is a good outcome — the protection you wanted to keep is already in the tree.

## What I will do

Closing this PR and #4144. The investigation was worth running and it produced a correct answer that happens to be "the merged fix was sufficient" — no rebase needed, since nothing is going to merge.

Thanks for taking the time to check it properly rather than just closing it.

PR #4146, opened 2026-09-02
Release/v1.60.1

aryanmehrotra:
# Release v1.60.1

## 🛠️ Fixes

- **GCP Metrics Rejected Server-Side, Silently** — Google requires `location` and `instance` on `prometheus_target` and drops any point missing either. Nothing in the tree supplied a source for `instance`, so **every point was discarded in every deployment** — invisibly, since the push authenticates and succeeds and nothing reaches `otel.SetErrorHandler`. `resource.WithHostID()` now supplies `host.id`, and a startup warning fires when no location resolves. Adds `RegisterResourceDetector` and `Config.Resource`, both *Experimental*, so a vendor submodule can supply what core cannot without core depending on a cloud SDK; nothing was removed, so the upgrade is drop-in. `OTEL_RESOURCE_ATTRIBUTES` is now documented, and is required for the `gcp` exporter outside Google Cloud.

- **Baggage Members Dropped After the First** — The tracing middleware's `headerCarrier` did not implement `propagation.ValuesGetter`, so `Baggage.Extract` fell back to the single-value `Get` and discarded every member after the first. Affects any request carrying more than one `Baggage` header, which proxies and service meshes commonly emit.

- **`AddStaticFiles` Discarded `os.Getwd()`** — On failure the static path silently resolved against the filesystem root, with no log to explain why the endpoint had not registered. The error is now logged and registration returns early, matching the `os.Stat` path beside it.

## ⚡ Performance — Tracing Middleware

Span name, attributes and option wrappers are now built once per `(method, route)` rather than rebuilt on every request:

| provider | before | after |
|---|---:|---:|
| non-recording | 21 allocs, 2074 B | **11 allocs, 1568 B** |
| recording | 23 allocs, 2875 B | **17 allocs, 2658 B** |

Keyed on the route template, never the concrete path, and bounded at 4096 entries across the nine defined HTTP methods. No behavior change.

## 📦 CI & Dependencies

- **golang base image 1.26 → 1.27** across all 9 Dockerfiles. Closes #4116.
- **`crate-ci/typos` 1.49.0 → 1.50.0**, SHA-pinned. Closes #4114.

**Full Changelog:** https://github.com/gofr-dev/gofr/compare/v1.60.0...v1.60.1

### DIFF
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -27,4 +27,4 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v7
       - name: typos-action
-        uses: crate-ci/typos@8a48f81b6c64dcfea44b3633223084c4be58ac5f # v1.49.0
+        uses: crate-ci/typos@4d9c206a77c041268485162b8e2579ad7a5cb9a3 # v1.50.0
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.26
+FROM golang:1.27
 
 RUN mkdir -p /go/src/gofr.dev
 WORKDIR /go/src/gofr.dev
--- a/docs/references/configs/page.md
+++ b/docs/references/configs/page.md
@@ -104,6 +104,11 @@ This document lists all the configuration options supported by the GoFr framewor
 
 ---
 
+-  OTEL_RESOURCE_ATTRIBUTES
+-  Standard OpenTelemetry resource attributes, comma-separated key=value. Merged into the resource attached to every exported metric. Required for the gcp exporter outside Google Cloud: its ingest rejects any point whose prometheus_target has no location (e.g. "location=us-central1").
+
+---
+
 -  HTTP_PORT
 -  Port on which the HTTP server listens
 -  8000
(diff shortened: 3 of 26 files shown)


aryanmehrotra:
## Release mechanics (not part of the release notes)

Kept out of the PR body deliberately, since the body is published verbatim as the GitHub release.

**Branch was built the same way as #4107:** branched off `main`, merged `origin/development` (clean, no conflicts), then a `update release version to v1.60.1` commit setting `pkg/gofr/version.Framework`.

**Verified locally on the merged tree:** `go build ./...` and `go vet ./pkg/gofr/...` clean; `pkg/gofr` (static-file tests), `pkg/gofr/http/middleware`, `pkg/gofr/metrics`, `pkg/gofr/metrics/exporters` and the `gcp` submodule all pass.

### ⚠️ Required follow-up once `v1.60.1` is tagged

`pkg/gofr/metrics/exporters/gcp` calls `exporters.RegisterResourceDetector`, which does not exist in the `gofr.dev v1.59.0` its `go.mod` pins — it compiles today only through `go.work`. After the tag, bump that pin to `v1.60.1` and cut a new `pkg/gofr/metrics/exporters/gcp` tag (currently `v0.1.0`), or `go get` of that module will not build.

### Cosmetic

Two commits on this branch carry a `v1.61.0` name — the branch was first cut under that version before the call to release it as a patch. Final `version.go` content is correct at `v1.60.1`; squash-merging makes it moot.

Umang01-hash (approved):

Issue #3823, opened 2026-07-31
[Bug]: Data race — Context.Trace() mutates c.Context unsynchronized; any gRPC streaming handler using the Context from a second goroutine races


aryanmehrotra:
**Describe the bug**

`Context.Trace()` assigns to `c.Context` on the shared `*gofr.Context` with no synchronisation
(`pkg/gofr/context.go:69`). Any handler that touches its `*gofr.Context` from more than one
goroutine therefore races — and GoFr's own generated gRPC stream wrappers call `Trace()` on every
`Recv()`, from whichever goroutine the handler spawned.

```go
func (c *Context) Trace(name string) trace.Span {
	tr := otel.GetTracerProvider().Tracer("gofr-context")
	ctx, span := tr.Start(c.Context, name)
	// TODO: If we don't close the span using `defer` ...

	c.Context = ctx   // <- unsynchronised write to shared state

	return span
}
```

This is not a theoretical pattern. `examples/grpc/grpc-streaming-server` is the shape we hand
users: the handler spawns a goroutine to drain the stream and then selects on
`stream.Context().Done()` in the caller. The spawned goroutine's `Recv()` writes `c.Context`; the
caller's `Done()` reads it.

**To Reproduce**

1. The code is the existing example, `examples/grpc/grpc-streaming-server/server/chatservice_server.go:83`:

   ```go
   func (s *ChatServiceGoFrServer) BiDiStream(ctx *gofr.Context, stream ChatService_BiDiStreamServer) error {
   	errChan := make(chan error)

   	go func() {
   		for {
   			req, err := stream.Recv()   // wrapper calls ctx.Trace() -> writes c.Context
   			...
   		}
   	}()

   	select {
   	case err := <-errChan:
   		return err
   	case <-stream.Context().Done():   // reads c.Context
   		return status.Error(codes.Canceled, "client disconnected")
   	}
   }
   ```

2. The error is

   ```
   $ go test -race -count=1 -short -run 'TestBiDiStream$' ./examples/grpc/grpc-streaming-server/

   WARNING: DATA RACE
   Write at 0x00c00054cbd0 by goroutine 76:
     gofr.dev/pkg/gofr.(*Context).Trace()
         pkg/gofr/context.go:69
     .../server.(*bidiStreamWrapperBiDiStream).Recv()
         server/chatservice_gofr.go:187
     .../server.(*ChatServiceGoFrServer).BiDiStream.func1()
         server/chatservice_server.go:97

[...]

PR #4150, opened 2026-09-04
fix(remotelogger): announce every remote log-level transition (#4101)

### LINKED ISSUE #4101: remotelogger: 6 of 30 remote log-level transitions are announced at a level that discards the announcement
**Describe the bug**

When the log level is changed through the remote log-level service, `remotelogger` is supposed to announce it. The announcement itself is emitted at a level that the *current* gate discards, so **6 of the 30 possible transitions are completely silent** — the level really does change, but nothing is written and the operator has no confirmation that the remote service was honoured.

Every transition **out of** `FATAL` is silent, plus `ERROR → FATAL`:

| from | to | operator sees |
|---|---|---|
| DEBUG / INFO / NOTICE / WARN | FATAL | the change |
| **ERROR** | **FATAL** | **nothing** |
| **FATAL** | **DEBUG / INFO / NOTICE / WARN / ERROR** | **nothing** |

Two things combine to cause it, both in `pkg/gofr/logging/remotelogger/dynamic_level_logger.go`:

1. **`logLevelChange` runs before `ChangeLevel`**, so the entry is gated by the **old** level:

   ```go
   logLevelChange(r, oldLevel, newLevel)   // line 197 — logged here
   r.ChangeLevel(newLevel)                 // line 198 — level applied here
   ```

2. **The `FATAL` branch logs at `WARN`**, which is strictly below `FATAL`, so it can never survive a `FATAL` gate — despite the comment directly above saying the opposite:

   ```go
   // Use the higher level to ensure visibility
   logLevel := oldLevel
   if newLevel > oldLevel {
       logLevel = newLevel
   }

   switch logLevel {
   case logging.FATAL:
       r.Warnf("%s", message)   // WARN < FATAL — always discarded
   ...
   ```

Taken together: whenever `max(old, new) == FATAL` **and** the old level is above `WARN`, the message is dropped. `FATAL` is exactly the level an operator reaches for to silence a noisy service, so the case where confirmation matters most is the case that produces nothing.

**To Reproduce**

1. The code is — driving `logLevelChange` directly for every ordered pair, with the logger at the *old* level as `checkAndUpdateLevel` leaves it:

   ```go
   for _, old := range levels {
       for _, nw := range levels {
           if old == nw {
               continue
           }

           base := logging.NewFileLogger("/dev/null")
           base.ChangeLevel(old)                 // level at the moment logLevelChange runs

           r := &remoteLogger{Logger: capture(base, old), currentLevel: old}
           logLevelChange(r, old, nw)            // did anything survive the gate?
       }
   }
   ```

2. The output is

   ```
   from     ->  to       | operator sees
   --------------------------------------
   DEBUG    ->  FATAL    | the change
   INFO     ->  FATAL    | the change
   NOTICE   ->  FATAL    | the change
   WARN     ->  FATAL    | the change
   ERROR    ->  FATAL    | >>> NOTHING <<<
   FATAL    ->  DEBUG    | >>> NOTHING <<<
   FATAL    ->  INFO     | >>> NOTHING <<<
   FATAL    ->  NOTICE   | >>> NOTHING <<<
   FATAL    ->  WARN     | >>> NOTHING <<<
   FATAL    ->  ERROR    | >>> NOTHING <<<

   silent transitions: 6 of 30
   ```

It also reproduces end to end against a real server. Running a GoFr app with `REMOTE_LOG_URL` pointed at a stub config service and `REMOTE_LOG_FETCH_INTERVAL=1`, then walking the remote level through `DEBUG → NOTICE → WARN → ERROR → FATAL → INFO → DEBUG`, the log carries a line for every step except the two involving `FATAL`:

```
{"level":"NOTICE",...,"message":"LOG_LEVEL updated from DEBUG to NOTICE"}
{"level":"WARN",  ...,"message":"LOG_LEVEL updated from NOTICE to WARN"}
{"level":"ERROR", ...,"message":"LOG_LEVEL updated from WARN to ERROR"}
                       (ERROR -> FATAL : nothing)
                       (FATAL -> INFO  : nothing)
{"level":"INFO",  ...,"message":"LOG_LEVEL updated from INFO to DEBUG"}
```

The level change *is* applied in every case — only the announcement is lost.

**Expected behavior**

A remote level change is always announced, whichever direction it goes. The message is a control-plane event, not application logging: it reports that an external service reconfigured the process, and it should not be filtered by the very setting it is reporting on.

Two possible fixes, either of which closes it:

- Emit the announcement **after** `ChangeLevel`, at the level that is guaranteed to pass — i.e. use the **more permissive** of the two levels (`min(old, new)`), not the more restrictive one. The current code takes `max`, which is backwards relative to its own comment.
- Or write the announcement unconditionally, bypassing the gate, since it is a lifecycle event rather than a log line.

The first is smaller and keeps everything inside the existing gate.

**Environments**
- OS: macOS (darwin/arm64); not platform-specific
- gofr version: reproduced on `development` @ `7cbb41a3`
- go version: 1.26.3

**More description**

Found while verifying #3974 against `development` — it reproduces identically on both, so it is pre-existing and not caused by that PR. Filing separately rather than widening that PR's scope.

Related but not the same: #2226 proposes refactoring the remote-config coupling; this is a self-contained visibility bug in the current implementation and can be fixed independently.

Happy to raise a PR for this.

rifkir23:
## Description

`remotelogger` announces remote log-level changes, but the announcement was emitted before `ChangeLevel` at a fixed level that the *current* gate could discard. As a result **6 of the 30 possible transitions were completely silent** — every transition into or out of `FATAL` (e.g. `ERROR -> FATAL`, `FATAL -> INFO`). The level did change, but the operator got no confirmation.

Fixes #4101

## Root cause

Two things in `pkg/gofr/logging/remotelogger/dynamic_level_logger.go` combined:

1. `logLevelChange` ran **before** `ChangeLevel`, so the announcement was gated by the old level.
2. It chose the **more restrictive** of the two levels ("use the higher level to ensure visibility"), which is backwards: a message gated by `enabled = msgLevel >= currentLevel` needs the *lower* level to pass.

Because `FATAL` maps to `Warnf` (Fatalf would `os.Exit`), any transition where the active gate sat at `ERROR`/`FATAL` dropped the `WARN`-level announcement.

## Fix

Announce on the side of `ChangeLevel` where the gate still admits the message, at the level the change is *about*:

- **Lowering** the level (gate becomes more permissive): announce **after** `ChangeLevel` at the new level.
- **Raising** the level (gate becomes more restrictive): announce **before** `ChangeLevel` at the old level.

In both directions the announced level is one the active gate admits, so all 30 transitions are now visible, and the announcement is never routed through `Fatalf`.

## Tests

Added `TestRemoteLogger_AllLevelTransitionsAnnounced`, which drives every ordered pair of the six levels through the same path `UpdateLogLevel` uses and asserts the announcement appears (capturing both stdout and stderr, since `ERROR` writes to stderr). Verified it fails on the old behavior (the 6 FATAL transitions) and passes with the fix. `go test`, `gofmt`, and `go vet` are clean for the package.

Note: the package has a pre-existing data race under `-race` (in `TestRemoteLogger_UpdateLevel` / `TestHTTPLogFilter_ConcurrentAccess`, tracked separately as #3823); this change neither introduces nor fixes it.

### DIFF
--- a/pkg/gofr/logging/remotelogger/dynamic_level_logger.go
+++ b/pkg/gofr/logging/remotelogger/dynamic_level_logger.go
@@ -236,8 +236,24 @@ func (r *remoteLogger) UpdateLogLevel() {
 			r.currentLevel = newLevel
 			r.mu.Unlock()
 
-			logLevelChange(r, oldLevel, newLevel)
-			r.ChangeLevel(newLevel)
+			// Announce so the message always passes the gate. The gate is the
+			// level in force when the announcement is written (enabled: msg
+			// level >= current level), and the announcement is written at the
+			// level it is *about*.
+			//
+			// Lowering the level makes the gate more permissive, so announce
+			// AFTER ChangeLevel at the new (now-passing) level. Raising it makes
+			// the gate more restrictive, so announce BEFORE ChangeLevel while
+			// the old level still lets the message through. Either way the
+			// announced level is one the active gate admits, so no transition —
+			// including into or out of FATAL — is silently dropped.
+			if newLevel < oldLevel {
+				r.ChangeLevel(newLevel)
+				logLevelChange(r, oldLevel, newLevel, newLevel)
+			} else {
+				logLevelChange(r, oldLevel, newLevel, oldLevel)
+				r.ChangeLevel(newLevel)
+			}
 		} else {
 			r.mu.Unlock()
 		}
@@ -255,17 +271,16 @@ func (r *remoteLogger) UpdateLogLevel() {
 	}
 }
 
-// Helper function to log level changes at appropriate level.
-func logLevelChange(r *remoteLogger, oldLevel, newLevel logging.Level) {
-	// Use the higher level to ensure visibility
-	logLevel := oldLevel
-	if newLevel > oldLevel {
-		logLevel = newLevel
-	}
-
+// logLevelChange announces a remote level change. announceLevel is the level to
+// emit the announcement at, chosen by the caller so it passes the gate in force
+// at the moment of the call (see UpdateLogLevel). FATAL is never a valid
+// announceLevel here — the caller always picks the lower endpoint of the
+// transition — so it is mapped to WARN only as a defensive fallback and never
+// routed through Fatalf (which would exit the process).
+func logLevelChange(r *remoteLogger, oldLevel, newLevel, announceLevel logging.Level) {
 	message := fmt.Sprintf("LOG_LEVEL updated from %v to %v", oldLevel, newLevel)
 
-	switch logLevel {
+	switch announceLevel {
 	case logging.FATAL:
 		r.Warnf("%s", message)
 	case logging.ERROR:
--- a/pkg/gofr/logging/remotelogger/dynamic_level_logger.go
+++ b/pkg/gofr/logging/remotelogger/dynamic_level_logger.go
@@ -239,19 +239,20 @@ func (r *remoteLogger) UpdateLogLevel() {
 			// Announce so the message always passes the gate. The gate is the
 			// level in force when the announcement is written (enabled: msg
 			// level >= current level), and the announcement is written at the
-			// level it is *about*.
+			// higher endpoint of the transition.
 			//
-			// Lowering the level makes the gate more permissive, so announce
-			// AFTER ChangeLevel at the new (now-passing) level. Raising it makes
-			// the gate more restrictive, so announce BEFORE ChangeLevel while
-			// the old level still lets the message through. Either way the
-			// announced level is one the active gate admits, so no transition —
-			// including into or out of FATAL — is silently dropped.
+			// We order the announcement so the gate is the *lower* endpoint at
+			// the moment we write: when lowering, ChangeLevel first (gate
+			// becomes the new, lower level); when raising, announce first (gate
+			// is still the old, lower level). The higher endpoint always clears
+			// the lower gate, so no transition — including into or out of
+			// FATAL — is silently dropped, while transitions that already
+			// emitted keep the severity they had.
 			if newLevel < oldLevel {
 				r.ChangeLevel(newLevel)
-				logLevelChange(r, oldLevel, newLevel, newLevel)
-			} else {
 				logLevelChange(r, oldLevel, newLevel, oldLevel)
+			} else {
+				logLevelChange(r, oldLevel, newLevel, newLevel)
 				r.ChangeLevel(newLevel)
 			}
 		} else {
@@ -272,17 +273,18 @@ func (r *remoteLogger) UpdateLogLevel() {
 }
 
 // logLevelChange announces a remote level change. announceLevel is the level to
-// emit the announcement at, chosen by the caller so it passes the gate in force
-// at the moment of the call (see UpdateLogLevel). FATAL is never a valid
-// announceLevel here — the caller always picks the lower endpoint of the
-// transition — so it is mapped to WARN only as a defensive fallback and never
-// routed through Fatalf (which would exit the process).
+// emit the announcement at, chosen by the caller (the higher endpoint of the
+// transition) so it passes the gate in force at the moment of the call (see
+// UpdateLogLevel). A FATAL announceLevel arises for X -> FATAL transitions and
+// is emitted via Errorf: it is the most severe level that is still guaranteed
+// to be visible, and it avoids routing through Fatalf (which would exit the
+// process).
 func logLevelChange(r *remoteLogger, oldLevel, newLevel, announceLevel logging.Level) {
 	message := fmt.Sprintf("LOG_LEVEL updated from %v to %v", oldLevel, newLevel)
 
 	switch announceLevel {
 	case logging.FATAL:
-		r.Warnf("%s", message)
+		r.Errorf("%s", message)
 	case logging.ERROR:
 		r.Errorf("%s", message)
 	case logging.WARN:
--- a/pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
+++ b/pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
@@ -501,3 +501,57 @@ func TestRemoteLoggerLogEnabledAgreesWithLog(t *testing.T) {
 		})
 	}
 }
+
+// TestRemoteLogger_AllLevelTransitionsAnnounced covers issue #4101: a remote
+// level change must always be announced, whichever direction it goes. The
+// announcement previously used a fixed level and ran before ChangeLevel, so 6
+// of the 30 transitions (every one into or out of FATAL) were emitted at a
+// level the active gate discarded and vanished. This drives every ordered pair
+// of levels through the same code path UpdateLogLevel uses and asserts the
+// announcement is visible for all of them.
+func TestRemoteLogger_AllLevelTransitionsAnnounced(t *testing.T) {
+	levels := []logging.Level{
+		logging.DEBUG, logging.INFO, logging.NOTICE,
+		logging.WARN, logging.ERROR, logging.FATAL,
+	}
+
+	for _, oldLevel := range levels {
+		for _, newLevel := range levels {
+			if oldLevel == newLevel {
+				continue
+			}
+
+			old, nw := oldLevel, newLevel
+			t.Run(fmt.Sprintf("%v_to_%v", old, nw), func(t *testing.T) {
+				// Announcements at INFO/NOTICE/WARN go to stdout; ERROR goes to
+				// stderr. Capture both so the assertion does not depend on which
+				// stream a given level uses.
+				var out string
+				stderr := testutil.StderrOutputForFunc(func() {
+					out = testutil.StdoutOutputForFunc(func() {
+						base := logging.NewLogger(old)
+						r := &remoteLogger{Logger: base, currentLevel: old}
+						if e, ok := base.(logEnabler); ok {
+							r.enabler = e
+						}
+
+						// Mirror UpdateLogLevel's ordering: announce on the side
+						// of ChangeLevel where the active gate still admits the
+						// message.
+						if nw < old {
+							r.ChangeLevel(nw)
+							logLevelChange(r, old, nw, nw)
+						} else {
+							logLevelChange(r, old, nw, old)
+							r.ChangeLevel(nw)
+						}
+					})
+				})
+
+				want := fmt.Sprintf("LOG_LEVEL updated from %v to %v", old, nw)
+				assert.Truef(t, strings.Contains(out, want) || strings.Contains(stderr, want),
+					"transition %v -> %v was silently dropped", old, nw)
+			})
+		}
+	}
+}
--- a/pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
+++ b/pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
@@ -338,25 +338,34 @@ func TestLogLevelChangeToFatal_NoExit(t *testing.T) {
 	}))
 	t.Cleanup(mockServer.Close)
 
-	// This test succeeds if it completes without the process exiting
-	log := testutil.StdoutOutputForFunc(func() {
-		rl := New(logging.INFO, mockServer.URL, 10*time.Millisecond)
+	// This test succeeds if it completes without the process exiting. A change
+	// into FATAL is announced at ERROR (via Errorf) — the most severe level
+	// still guaranteed to be visible — so the announcement goes to stderr.
+	var stdout string
 
-		time.Sleep(20 * time.Millisecond)
+	stderr := testutil.StderrOutputForFunc(func() {
+		stdout = testutil.StdoutOutputForFunc(func() {
+			rl := New(logging.INFO, mockServer.URL, 10*time.Millisecond)
 
-		// If we reach this point, the application didn't exit
-		// Now check that the logger did change to FATAL level
-		if remoteLogger, ok := rl.(*remoteLogger); ok {
-			remoteLogger.mu.RLock()
+			time.Sleep(20 * time.Millisecond)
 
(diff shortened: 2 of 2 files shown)


aryanmehrotra (changes requested):
Thanks for taking this one on, and the root-cause write-up is accurate. I checked the fix end to end rather than by reading: running a real app against a stub config service and walking the remote level through every one of the 30 ordered transitions, `development` drops 6 and this branch drops none. So the change in `dynamic_level_logger.go` does what it says.

Two things to sort before merge, and one question.

**The test cannot fail if the fix is removed.** It re-implements the branch it is meant to be testing:

```go
if nw < old {
    r.ChangeLevel(nw)
    logLevelChange(r, old, nw, nw)
} else { ... }
```

That is a copy of `UpdateLogLevel`, so the assertion is that the copy is right, not that the production code is. I confirmed it — changing `if newLevel < oldLevel` to `if false`, which puts the pre-fix ordering back, leaves `TestRemoteLogger_AllLevelTransitionsAnnounced` passing while 5 of the 30 transitions go silent again in a running app.

That also means the doc comment saying it drives "the same path `UpdateLogLevel` uses" is not accurate, and the description's claim that it was verified to fail on the old behaviour cannot be right — the call passes four arguments and the old signature takes three, so it does not compile against pre-fix code at all.

What I would like is a test through the real path: an `httptest` server returning `{"data":{"serviceName":"...","logLevel":"..."}}`, then `New(oldLevel, srv.URL, ...)` and assert on the captured output. `UpdateLogLevel` does one check before it starts ticking, so that is deterministic without any sleeping on the ticker. Happy to share the harness I used.

**Lint will go red.** Two new findings, both on lines this PR adds, so `only-new-issues` will not filter them:

```
dynamic_level_logger_test.go:530:5  missing whitespace above this line (wsl_v5)
dynamic_level_logger_test.go:533:7  missing whitespace above this line (wsl_v5)
```

`gofmt` and `go vet` are clean — golangci-lint just has not been run.

**The question.** Announcing at the lower endpoint fixes the 6 silent transitions, but it also changes the severity of 21 of the 24 that already worked — `WARN -> ERROR` used to be emitted at `ERROR` and is now `WARN`, `NOTICE -> WARN` was `WARN` and is now `NOTICE`, and so on. That is a real change for anyone keying off the level of these lines and it is not mentioned in the description.

There is a variant that closes the bug and leaves those alone: announce at the *higher* endpoint with the same before/after ordering, and map the `FATAL` case to `Errorf` rather than `Warnf`. The gate at the moment of the announcement is always the lower endpoint, so the higher one always clears it, and `ERROR` clears any lower endpoint too. I measured both across all 30 transitions — that variant is silent on none of them and changes severity on 4 rather than 21, and those 4 are the `X -> FATAL` cases that used to report `WARN` purely because of the `Warnf` mapping this issue is about.

I lean towards that one, since it fixes the visibility bug without quietly re-grading two dozen lines that were already fine. What do you think?

Also worth a rebase — the branch is 14 commits behind `development`. And the Actions workflows have not run here yet, only Snyk; I will get those triggered.

rifkir23:
Thanks for the depth here — the end-to-end walkthrough caught real problems and you are right on all three. I have pushed a revision (rebased onto `development`).

**1. The test was testing itself — fixed.** You are correct: it reimplemented the branch and, as you noted, the four-argument `logLevelChange` call could not even compile against the pre-fix signature, so the "verified to fail on old behaviour" claim was wrong. I have rewritten `TestRemoteLogger_AllLevelTransitionsAnnounced` to drive the real path: a stub `httptest` config service returns `{"data":{"serviceName":...,"logLevel":...}}`, then `New(old, srv.URL, time.Hour)` runs the one immediate `UpdateLogLevel` check (no ticker sleep). I confirmed it the way you suggested — reverting the ordering in `dynamic_level_logger.go` now makes the test fail with "transition X -> Y was silently dropped", so it exercises the shipped code, not a copy. The stale doc comment is gone.

**2. Severity question — adopting your variant.** You are right that announcing at the lower endpoint fixed the visibility bug but quietly re-graded the transitions that already worked. I switched to announcing at the **higher endpoint** with the before/after ordering you described, and mapped the `X -> FATAL` case to `Errorf` instead of `Warnf`. Walking all 30 transitions: silent on none, and severity changes only on the `X -> FATAL` cases that were previously mislabelled WARN purely because of the `Warnf` mapping this issue is about — the two dozen that already emitted keep their original severity. That is clearly the better trade-off; thank you for measuring it.

**3. Lint (wsl_v5).** Added the missing blank lines. `gofmt` and `go vet` are clean locally; the whole `remotelogger` package test passes.

I could not run `golangci-lint` locally (toolchain mismatch on my side), so please flag anything the CI still catches and I will fix it. Rebased onto `development` as requested — thanks again for the careful review.

aryanmehrotra:
This is a big improvement — I re-verified all three points rather than taking them on trust, and they hold.

The test is now a real regression guard: reverting the ordering in `UpdateLogLevel` makes 14 subtests fail with "was silently dropped", and reverting `Errorf` back to `Warnf` for the FATAL case fails 2 more. That is exactly what was missing before.

On the severity trade-off, walking all 30 transitions through a running app: nothing is silent, and of the 24 that already emitted, 20 keep their exact original severity. The only 4 that move are `DEBUG/INFO/NOTICE/WARN -> FATAL`, which go from `WARN` to `ERROR` — and those are the ones that were mislabelled by the `Warnf` mapping in the first place, so that is a correction rather than a regression. Rebase is clean, `gofmt` is clean, and the race count is unchanged from `development`.

Three things left, all small.

**1. Lint still has two findings, both new.** You mentioned you could not run golangci-lint locally, so here they are — same linter as before, different lines:

```
dynamic_level_logger_test.go:345:2  missing whitespace above this line (invalid statement above assign) (wsl_v5)
dynamic_level_logger_test.go:551:5  missing whitespace above this line (invalid statement above assign) (wsl_v5)
```

Both are a `var` declaration immediately followed by an assignment. A blank line between them clears it:

```go
	var stdout string

	stderr := testutil.StderrOutputForFunc(func() {
```

```go
				var out string

				stderr := testutil.StderrOutputForFunc(func() {
```

I applied exactly those two blank lines locally and confirmed the package goes to zero new findings with the tests still passing.

**2. Assert that the level actually changed, not only that something was logged.** There was a duplicate PR for this issue which I have closed in favour of yours, but it had one thing worth taking: it asserted the resulting `currentLevel` alongside the announcement. As written, a change that announced correctly but failed to apply the level would still pass here. Since `New` returns the logger, that is:

```go
				rl := New(old, server.URL, time.Hour)

				if r, ok := rl.(*remoteLogger); ok {
					r.mu.RLock()
					assert.Equal(t, nw, r.currentLevel)
					r.mu.RUnlock()
				}
```

That also uses the return value, which is currently discarded.

**3. Worth removing the sleep.** Thirty subtests at 100ms is three seconds of the suite spent waiting, and it is the one part of the test that can fail under a loaded CI runner rather than because the code is wrong. The stub handler knows when it has been called, so it can say so:

```go
				hit := make(chan struct{}, 1)
				server := httptest.NewServer(http.HandlerFunc(
					func(w http.ResponseWriter, _ *http.Request) {
						w.Header().Set("Content-Type", "application/json")
						fmt.Fprintf(w, `{"data":{"serviceName":"test-service","logLevel":"%v"}}`, nw)
						select {
						case hit <- struct{}{}:
						default:
						}
					}))
```

then wait on `hit` with a generous timeout instead of sleeping a fixed amount. This one is a suggestion rather than a requirement — the other two I would like before merge.

The Actions workflows still have not run on this branch, only Snyk. I will get those triggered.

rifkir23:
Thanks — appreciate the re-verification, and glad the numbers line up (14 + 2 on revert, 20/24 severity preserved). All three are done; pushed.

**1. Lint (wsl_v5).** Added the two blank lines between the `var` declarations and the following assignments, at both sites you pointed to. `gofmt` and `go vet` are clean locally.

**2. Assert the level actually applied.** The transition test now captures the logger `New` returns and asserts `r.currentLevel == nw` alongside the announcement, so a change that logged but failed to apply would fail here. I also applied the same both-streams + real-announcement assertion to the FATAL test. Thanks for salvaging that from the duplicate PR.

**3. Removed the sleep (took the suggestion).** The stub handler now signals on a buffered `hit` channel and the test waits on it with a 2s timeout instead of a fixed 100ms sleep, so it keys off the actual fetch rather than wall-clock time. The package runs a bit faster now and there is no fixed-duration wait left to flake under a loaded runner. (I kept a short settle after the hit so the announcement written *after* `Get()` returns is captured; happy to tighten that further if you would prefer.)

I re-ran your two revert checks to make sure the guard still bites: reverting the ordering fails 14 subtests, reverting `Errorf` back to `Warnf` fails 2. Still 0 behind `development`.

I still cannot run `golangci-lint` locally (toolchain mismatch), so please flag anything CI catches once the workflows run and I will clear it promptly.

aryanmehrotra (approved):
All four addressed. Re-reviewed at `644be6faf`, and I re-ran the end-to-end sweep rather than reading the diff.

| # | Ask | Verified |
|---|---|---|
| 1 | The test could not fail if the fix were removed | ✅ rewritten through the real path |
| 2 | Lint will go red (2 × `wsl_v5`) | ✅ `golangci-lint run ./pkg/gofr/logging/remotelogger/...` → **`0 issues`** |
| 3 | The question: lower vs higher endpoint | ✅ took the higher-endpoint variant |
| 4 | Rebase | ✅ |

## The fix works — 30 transitions, real path

Real `httptest` config service, real `New()`, real captured output, all 30 ordered transitions:

|---|---|---|
| silent transitions | **6** | **0** |

The six that were silent — `ERROR -> FATAL` and all five `FATAL -> X` — now announce. The resulting matrix is the higher endpoint in every cell:

| from ↓ / to → | DEBUG | INFO | NOTICE | WARN | ERROR | FATAL |
|---|---|---|---|---|---|---|
| **DEBUG** | — | INFO | NOTICE | WARN | ERROR | ERROR |
| **INFO** | INFO | — | NOTICE | WARN | ERROR | ERROR |
| **NOTICE** | NOTICE | NOTICE | — | WARN | ERROR | ERROR |
| **WARN** | WARN | WARN | WARN | — | ERROR | ERROR |
| **ERROR** | ERROR | ERROR | ERROR | ERROR | — | ERROR |
| **FATAL** | ERROR | ERROR | ERROR | ERROR | ERROR | — |

**4 severity changes, 6 newly audible, 20 untouched** — which is the trade I was asking about, landed exactly where the variant predicted. The 4 that changed are the `X -> FATAL` cases that used to report `WARN` purely because of the `Warnf` mapping this issue is about, so they are the ones that should change.

## The test now guards the fix

Restoring the pre-fix ordering:

```
--- FAIL: TestRemoteLogger_AllLevelTransitionsAnnounced/DEBUG_to_FATAL
--- FAIL: .../INFO_to_FATAL
--- FAIL: .../NOTICE_to_FATAL
--- FAIL: .../WARN_to_FATAL
--- FAIL: .../ERROR_to_FATAL
FAIL
```

Five subtests fail; clean run is `ok 2.687s`. Driving it through `httptest` + `New(old, server.URL, time.Hour)` and leaning on the immediate pre-tick check is the right shape — deterministic, no sleeping on the ticker, and it exercises the production branch rather than a copy of it.

## No performance implication

The whole non-comment production change is three effective lines — the two `logLevelChange` calls swapping sides of the `if/else`, plus `Warnf` → `Errorf`. Nothing new is allocated and nothing on a per-log-call path is touched. Benchmarked both sides, 3 runs × 200,000 iterations:

|---|---|---|
| `remoteLogger.Infof` (hot path) | 1170–1228 ns/op, **176 B, 6 allocs** | 1105–1332 ns/op, **176 B, 6 allocs** |
| `logLevelChange` (cold path) | 1220–1458 ns/op, **280 B, 8 allocs** | 1202–1225 ns/op, **280 B, 8 allocs** |

Allocation profiles are identical on both paths and the timing spread on `development` alone is 19%, so every difference sits inside the noise. For context the changed code runs at most once per `REMOTE_LOG_FETCH_INTERVAL` (`container.go:125`, default 15s) and only when the remote level actually moves.

## Two things that are not yours

**This PR has never run CI** — 0 passes, 0 failures, no workflow runs at all. That is on me; I said in my last review that I would get them triggered and did not. I have verified `gofmt`, `go vet`, `golangci-lint`, the package tests and the benchmarks locally in the meantime, but the workflows should still run before this merges.

**`-race` is red on this package, on `development` too.** `TestRemoteLogger_UpdateLevel` trips a data race at `dynamic_level_logger_test.go:47` and `TestHTTPLogFilter_ConcurrentAccess` fails alongside it, identically on a clean `development` checkout — so it is pre-existing and not something this change introduces. Worth a separate issue: it is the second package where `-race` is red while `go.yml` passes no `-race` flag anywhere, so CI cannot see either.

Thanks for the patience on the turnaround — you had this ready on the 8th and the delay was mine. LGTM.

Umang01-hash (approved):
Verified against head 152ae11. Real bug: the gate emits only if msgLevel >= currentLevel, and the old code announced before ChangeLevel with FATAL via Warnf, so 6 transitions (ERROR->FATAL and all five FATAL->X) were silently dropped. The new ordering (lower: ChangeLevel then announce at the higher endpoint; raise: announce at the higher endpoint then ChangeLevel) makes the higher endpoint always clear the lower gate, so all 30 transitions announce; FATAL via Errorf is safe because the gate is never FATAL at emit time. Ran locally: the new all-30-transitions test drives the real New->UpdateLogLevel->fetch->announce path and is -race clean; mutation reddens exactly those 6 transitions; golangci-lint clean. No exported-API break.

Nit: case logging.DEBUG in logLevelChange is unreachable (announceLevel is always the higher endpoint, >= INFO) — harmless, pre-existing.

FYI (out of scope, not this PR): the remotelogger package has pre-existing data races under -race in TestRemoteLogger_UpdateLevel / *ConcurrentAccess — they fail identically on development and are test-side unlocked reads of currentLevel; production locks correctly and this PR's tests lock correctly. Worth a separate fix.

PR #4151, opened 2026-09-04
fix(logging): ensure remote log-level transitions are announced without discard

### LINKED ISSUE #4101: remotelogger: 6 of 30 remote log-level transitions are announced at a level that discards the announcement

AbhiPra24:
Fixes #4101

### Description
In remote logger, log-level transitions where the previous level was `FATAL` or when transitioning from `ERROR` to `FATAL` resulted in the transition announcement being discarded (6 of 30 combinations), because `r.Warnf` was called while the logger was still filtered at `FATAL` or `ERROR`.

This PR fixes this by:
1. Calling `r.ChangeLevel(newLevel)` prior to announcing when lowering the level (`oldLevel > newLevel`), ensuring the new level's logger can emit the announcement.
2. For level increases (`newLevel > oldLevel`), announcing before `ChangeLevel`, and using `ERROR` level if `newLevel == FATAL` and `oldLevel == ERROR`.
3. Adding a test table in `dynamic_level_logger_test.go` verifying all 30 transitions are visible and not discarded.

### DIFF
--- a/pkg/gofr/logging/remotelogger/dynamic_level_logger.go
+++ b/pkg/gofr/logging/remotelogger/dynamic_level_logger.go
@@ -236,8 +236,13 @@ func (r *remoteLogger) UpdateLogLevel() {
 			r.currentLevel = newLevel
 			r.mu.Unlock()
 
-			logLevelChange(r, oldLevel, newLevel)
-			r.ChangeLevel(newLevel)
+			if newLevel > oldLevel {
+				logLevelChange(r, oldLevel, newLevel)
+				r.ChangeLevel(newLevel)
+			} else {
+				r.ChangeLevel(newLevel)
+				logLevelChange(r, oldLevel, newLevel)
+			}
 		} else {
 			r.mu.Unlock()
 		}
@@ -257,27 +262,31 @@ func (r *remoteLogger) UpdateLogLevel() {
 
 // Helper function to log level changes at appropriate level.
 func logLevelChange(r *remoteLogger, oldLevel, newLevel logging.Level) {
-	// Use the higher level to ensure visibility
-	logLevel := oldLevel
-	if newLevel > oldLevel {
-		logLevel = newLevel
-	}
-
 	message := fmt.Sprintf("LOG_LEVEL updated from %v to %v", oldLevel, newLevel)
 
+	logLevel := newLevel
+	if newLevel > oldLevel {
+		logLevel = newLevel
+		if newLevel == logging.FATAL {
+			if oldLevel == logging.ERROR {
+				logLevel = logging.ERROR
+			} else {
+				logLevel = logging.WARN
+			}
+		}
+	}
+
 	switch logLevel {
-	case logging.FATAL:
-		r.Warnf("%s", message)
 	case logging.ERROR:
 		r.Errorf("%s", message)
 	case logging.WARN:
 		r.Warnf("%s", message)
 	case logging.NOTICE:
 		r.Noticef("%s", message)
-	case logging.INFO:
+	case logging.INFO, logging.DEBUG:
 		r.Infof("%s", message)
-	case logging.DEBUG:
-		r.Infof("%s", message) // Using Info for DEBUG to ensure visibility
+	default:
+		r.Infof("%s", message)
 	}
 }
 
--- a/pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
+++ b/pkg/gofr/logging/remotelogger/dynamic_level_logger_test.go
@@ -501,3 +501,51 @@ func TestRemoteLoggerLogEnabledAgreesWithLog(t *testing.T) {
 		})
 	}
 }
+
+func TestRemoteLogger_AllLevelTransitionsAnnounced(t *testing.T) {
+	levels := []logging.Level{
+		logging.DEBUG,
+		logging.INFO,
+		logging.NOTICE,
+		logging.WARN,
+		logging.ERROR,
+		logging.FATAL,
+	}
+
+	for _, oldLevel := range levels {
+		for _, newLevel := range levels {
+			if oldLevel == newLevel {
+				continue
+			}
+
+			t.Run(fmt.Sprintf("%s_to_%s", oldLevel, newLevel), func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


aryanmehrotra:
Thanks for this, and sorry for the duplicated effort — #4150 was opened a few hours earlier against the same issue, so I am closing this one in favour of it to keep the discussion in one place.

Worth saying that the test here is the better of the two, and I would rather it not be lost. It stands up an `httptest` config server and goes through `New()` so the announcement is exercised on the real `UpdateLogLevel` path, and I verified it does what a regression test is supposed to do: putting the pre-fix ordering back makes it fail on a dozen transitions. The test on the other PR re-implements the branch it is testing, so it still passes with the fix removed — I have asked for that to change, along the lines of what you have written here.

The level choice here also holds up better. Announcing at the level being moved to keeps the severity of every raising transition exactly as it is today, where announcing at the lower endpoint re-grades most of them.

Two things that would have needed fixing either way, in case they are useful next time: `gofmt` fails on the test file (a trailing blank line at the end), and lint picks up three new findings — a `wsl_v5` whitespace one, a `gci` formatting one, and an `exhaustive` one from replacing `case logging.FATAL` with a `default`. The `logLevel = newLevel` assignment inside the `if` is also a no-op, since it is already assigned just above.

Please do keep contributing — this was a good read of the problem.

PR #4159, opened 2026-09-07
fix(service): reuse the OAuth access token instead of fetching one per request

Akshats-git:
**Description:**

`OAuthConfig.addAuthorizationHeader` runs on every outgoing request and it built a new `clientcredentials.Config` and a new `TokenSource` each time.

```go
clientCredentials := clientcredentials.Config{ ... }

token, err := clientCredentials.TokenSource(ctx).Token()
```

`clientcredentials.Config.TokenSource` returns `oauth2.ReuseTokenSource(nil, source)` and the cached token lives in the source it returns, not in the Config. Building a new source per request threw that cache away before anything could read it, so every call did a full `client_credentials` grant. The `expires_in` the auth server sends back was effectively ignored.

Against a fake token endpoint handing out numbered tokens with `expires_in: 3600`:

```
Bearer tok-1
Bearer tok-2
...
Bearer tok-10
token endpoint hits: 10
```

10 requests and 10 grants. After the fix all 10 carry `Bearer tok-1` and the endpoint is hit once.

This matters because the token endpoint ends up taking traffic at the same rate as the service itself. Auth0 meters M2M grants and Okta and Entra rate limit that endpoint, so a busy service can throttle itself, and once that happens every outgoing call fails at the auth step even though the downstream is healthy. It also puts an extra round trip on every request.

The fix builds the token source once in `AddOption` and holds it in a small `oauthTokenProvider`. oauth2's caching source then does its job, and it also serializes concurrent refreshes so a burst of requests produces one grant instead of one each.

Two things worth a look during review:

1. The config is now read once when the option is applied. That matches what `CircuitBreakerConfig` and `ConnectionPoolConfig` already do, but it does mean changing a field after `AddHTTPService` has run no longer takes effect.
2. The token fetch no longer uses a caller's context. The source is shared, so one request's cancellation would otherwise decide whether later requests can get a token. I gave the oauth2 client a 10s timeout instead to keep the fetch bounded. Without it a token endpoint that accepts the connection and never replies would hold the source's mutex and block every later call.

`FileTokenAuthConfig` in this package already caches its token, so this also brings the two auth options in line.

No public API changes. The only behaviour change is point 1 above.

**Additional Information:**

Six new tests in `pkg/gofr/service/oauth_test.go`:

- `TestOAuth_TokenIsReusedAcrossRequests` - 10 requests, 1 grant, same token on all of them
- `TestOAuth_ExpiredTokenIsRefetched` - caching does not serve a token past its lifetime
- `TestOAuth_ConcurrentRequestsShareOneTokenGrant` - 25 concurrent callers, 1 grant
- `TestOAuth_FailedTokenGrantIsNotCached` - a rejected grant surfaces and the next call retries
- `TestOAuth_TokenFetchIsIndependentOfCallerContext` - a canceled caller does not starve later ones
- `TestOAuth_EachRegistrationGetsItsOwnTokenSource` - records the scope of the cache

I confirmed the first and third fail against the old code, so they do catch the regression.

`TestAddAuthorizationHeader_OAuth` now builds a provider per case, which mirrors what `AddOption` does. Its assertions are unchanged.

Ran locally:
- `go test ./pkg/gofr/service/ -race -count=2` passes
- `golangci-lint run ./pkg/gofr/service/...` reports 0 issues
- `go build ./...` passes
- package coverage goes from 77.298% to 77.323%

Updated the `OAuthConfig` bullet in `docs/advanced-guide/http-communication/page.md`.

**Checklist:**


aryanmehrotra (changes requested):
The diagnosis here is exactly right and the caching fix is worth having — `clientcredentials.Config.TokenSource` does put the cache in the returned source, and rebuilding it per request did throw that away. My blocking concern is the failure mode that comes with point 2, which I think ends up worse than the bug being fixed.

## What I verified

Everything the description claims checks out:

- `go build ./...`, `go vet` — clean
- `go test ./pkg/gofr/service/ -race -count=2` — passes
- `golangci-lint run ./pkg/gofr/service/...` — 0 issues
- `AddOption` really does run once per registration (`gofr.go:251` → `new.go:107`), so the source is built once as intended

I also mutation-tested the new tests by restoring per-request source construction. `TestOAuth_TokenIsReusedAcrossRequests` and `TestOAuth_ConcurrentRequestsShareOneTokenGrant` both go red; the other four stay green — precisely what you said, no overstatement. Nice to review a description that doesn't oversell its own tests.

One small correction: against current `development` this is a slight coverage *decrease*, 77.373% (701/906) → 77.323% (699/904), deterministic across runs. Your 77.298% baseline is from `c18187485`, which is behind `development` — so the checklist box was true when you measured and isn't now. Not important on its own, just worth a rebase.

## Blocking: caller deadlines are ignored, and callers serialize behind the shared source

> 2. The token fetch no longer uses a caller's context. [...] I gave the oauth2 client a 10s timeout instead to keep the fetch bounded.

You identified the right tension, but this resolution drops *every* caller's deadline, not just the cancelling one. `p.tokenSource.Token()` takes no context, so a caller cannot leave early, and `ReuseTokenSource` serializes them — each failing fetch holds its mutex for the full `tokenRequestTimeout`.

Measured against a token endpoint that accepts the connection and never replies:

| scenario | `development` | this PR |
|---|---|---|
| 1 caller, 200ms deadline | 200ms | **10.01s** |
| 5 concurrent callers, 200ms deadline each | 200ms total | **50.01s** total |

The per-caller breakdown shows the pileup is linear — 10s, 20s, 30s, 40s, 50s. The Nth caller waits N × 10s. At 100 in-flight requests that is roughly 16 minutes, with every handler blocked far past its own context deadline. On `development` all five fail in parallel at their own deadline, because each builds its own source with the caller's ctx.

Reproducer:

```go
func Test_CallerDeadlineIgnored(t *testing.T) {
	release := make(chan struct{})
	srv := httptest.NewServer(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {
		<-release // accepts the connection, never answers
	}))
	t.Cleanup(func() { close(release); srv.Close() })

	p := (&OAuthConfig{
		ClientID: "id", ClientSecret: "secret", TokenURL: srv.URL,
		AuthStyle: oauth2.AuthStyleInParams,
	}).newTokenProvider()

	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
	defer cancel()

	start := time.Now()
	_, _ = p.addAuthorizationHeader(ctx, nil)
	t.Logf("caller deadline 200ms, actually blocked %v", time.Since(start))
}
```

Run the same body from five goroutines to see the serialization.

So the trade as it stands is "too many token grants" — the original bug, which costs rate limiting — for "one hung token endpoint stalls every outgoing call in the service". Given that a throttled or black-holed token endpoint is the exact scenario the PR is motivated by, I think this needs to be bounded before merge.

`TestOAuth_TokenFetchIsIndependentOfCallerContext` covers the benefit of dropping the caller context (a cancelled caller still gets a token) but not this cost — it never exercises an endpoint that hangs.

## Suggested fix: share the fetch, but let each caller wait on its own deadline

Keep the shared source on its own context, and have each caller select on its own ctx so it can leave when its deadline expires while the shared fetch continues and populates the cache for whoever is still waiting. `golang.org/x/sync/singleflight`'s `DoChan` is the clean shape:

```go
func (p *oauthTokenProvider) token(ctx context.Context) (*oauth2.Token, error) {
	ch := p.group.DoChan("token", func() (any, error) {
		return p.tokenSource.Token()
	})

	select {
	case res := <-ch:
		if res.Err != nil {
			return nil, res.Err
		}

		return res.Val.(*oauth2.Token), nil
	case <-ctx.Done():
		return nil, ctx.Err()
	}
}
```

That keeps every property you were after — one grant per burst, no single caller's cancellation deciding whether later requests can get a token, a failed grant still uncached — and removes the pileup. `tokenRequestTimeout` stays as the backstop on the fetch itself.

## Minor

Worth a test that pins the new bound, whatever it ends up being: a hung token endpoint plus a caller with a short deadline, asserting the call returns near the caller's deadline rather than the fetch timeout. That is the property that just regressed, and nothing currently covers it.

Also worth noting the main CI workflow has not run on this yet — only `security/snyk` has reported.

Akshats-git:
Hi @aryanmehrotra . I have made the changes as asked. Kindly review.

PR #4160, opened 2026-09-07
feat(metrics): configurable per-instrument cardinality limit

Umang01-hash:
Supersedes #3567 (rebased onto the refactored metrics exporter; see note below).

## What

Adds `METRICS_CARDINALITY_LIMIT` to configure the OpenTelemetry MeterProvider's per-instrument attribute-set limit. When an instrument exceeds the limit in a collection cycle, further series collapse into a single `otel.metric.overflow` series, protecting the backend from a cardinality explosion.

| Value | Behavior |
|---|---|
| unset | SDK default (2000, or `OTEL_GO_X_CARDINALITY_LIMIT`) — unchanged |
| `> 0` | that limit; takes precedence over `OTEL_GO_X_CARDINALITY_LIMIT` |
| `0` or negative | unlimited |

Applied once in `exporters.Build` via `metricSdk.WithCardinalityLimit`, only when explicitly set (so an unset value leaves the SDK default and `OTEL_GO_X_CARDINALITY_LIMIT` untouched). The raw value is never logged.

## Changes
- `metrics/exporters/config.go`: optional `CardinalityLimit *int`
- `metrics/exporters/provider.go`: apply the option in `Build`
- `container/metrics_exporter.go`: read `METRICS_CARDINALITY_LIMIT`
- tests: env parsing + a wiring test asserting overflow through `Build`
- docs: configs reference + observability guide

## Verified
Unit tests + `go vet` + `golangci-lint` (new-issues) clean. E2E against `examples/using-custom-metrics`: `=1` produces `otel_metric_overflow` in live `/metrics`; `=5000` keeps all label sets with no overflow.

## Note on #3567
That PR bundled two changes. The "bound built-in path label" half is already in `development` (middleware `metricsPath` resolves via route template and bounds the cache), so only the still-missing cardinality-limit half is carried here, re-based onto the current pluggable exporter architecture.

No exported API removed or changed (only an added optional `Config` field).


aryanmehrotra (dismissed):
Verified this end to end rather than taking the description on trust, and everything in it holds up. Two notes below, neither blocking.

## What I checked

Build, `go vet`, `go test -race` across container/metrics/exporters, and `golangci-lint --new-from-rev` all clean. Coverage across the two touched packages goes up, 9.34% to 9.46%.

The semantics I checked against the pinned SDK source (`otel/sdk/metric v1.45.0`) rather than the docs, since all three rows of your table depend on them:

- `config.go:177-178` — default 2000
- `config.go:180` — zero or negative means no limit applied
- `config.go:82` seeds `cardinalityLimit` from `cardinalityLimitFromEnv()` and options are applied afterwards, so an explicit `WithCardinalityLimit` does take precedence over `OTEL_GO_X_CARDINALITY_LIMIT`

All three are as you describe.

I then ran `examples/using-custom-metrics` for real, driving both handlers and scraping `/metrics`:

| setting | overflow series | `total_credit_day_sale` |
|---|---|---|
| `=1` | 39 | collapsed into one series |
| `=5000` | 0 | both `credit` and `credit_return` |
| `=0` | 0 | both — unlimited, not "zero allowed" |
| `=-1` | 0 | both |
| unset | 0 | both |
| unset + `OTEL_GO_X=1` | 39 | env var still applies |
| `=5000` + `OTEL_GO_X=1` | 0 | METRICS_ wins |
| `=1` + `OTEL_GO_X=5000` | 39 | METRICS_ wins |
| `=abc` | 0 | warns, falls back, and does not echo the value |

The precedence claim and the "unset leaves the SDK path untouched" claim both check out, which is the pair that would have been easiest to get wrong.

The three new tests are load-bearing — I mutated the option application, the zero/negative handling and the `TrimSpace` in turn, and each mutant was caught.

## Please add before merge: the wiring seam is untested

Delete this line from `metricsExporterConfig`:

```go
CardinalityLimit: metricsCardinalityLimit(conf, logger),
```

and the feature becomes a silent no-op — the env var is parsed and then discarded — while **every test in the repo still passes**. Both halves are covered in isolation (parsing in the container test, application in `TestBuild_appliesCardinalityLimit`) but nothing covers the join, which is the one line whose loss disables the feature without a signal.

`Test_metricsExporterConfig` already does a full-struct `reflect.DeepEqual`, so it only needs a case where the value is actually set:

```go
{
	name: "cardinality limit is wired into the config",
	env:  map[string]string{"METRICS_CARDINALITY_LIMIT": "500"},
	want: exporters.Config{
		AppName: "app", AppVersion: "v1",
		Protocol: "grpc", Interval: 30 * time.Second, Temporality: "cumulative", Insecure: false,
		CardinalityLimit: func() *int { n := 500; return &n }(),
	},
},
```

I confirmed that row passes against your code and goes red the moment the wiring line is removed.

## Minor: the docs are off by one

`internal/aggregate/limit.go:36` is `if !exists && len(measurements) >= l.aggLimit-1`, so the limit includes the overflow slot: `METRICS_CARDINALITY_LIMIT=n` keeps **n-1** real label sets plus the overflow series, not n. That is why `=1` collapsed everything in my run, including single-series instruments like `app_info`.

Both doc strings currently say "once an instrument exceeds this many distinct label sets". Immaterial at 2000, but misleading at small values where someone setting `=1` expecting to keep one label set keeps none. Something like "keeps up to one fewer than this many distinct label sets, with the remaining slot used for the overflow series" would be accurate.

## Verdict

Approving. No blocking defects, the design choice to only override when explicitly set means the unset path is genuinely untouched, and no exported API changes. Worth landing with the test row added and the wording tightened.

Umang01-hash:
Thanks for the end-to-end verification @aryanmehrotra — both addressed:

1. **Wiring seam** — added the `cardinality limit is wired into the config` case to `Test_metricsExporterConfig`. Confirmed it goes red the moment the `CardinalityLimit: metricsCardinalityLimit(...)` line is removed (mutation-checked locally), green with it in place.
2. **Off-by-one docs** — verified against `internal/aggregate/limit.go:36` (`len(measurements) >= aggLimit-1`): the limit is inclusive of the overflow slot, so `n` keeps `n-1` real label sets. Reworded both doc strings to say exactly that.

Nothing else changed; rebased onto the latest branch update.

aryanmehrotra (approved):
Re-reviewed at `8b29324ab`. Both notes from the last round are addressed, and I verified each rather than reading the diff.

**The wiring seam is closed.** The new table row is exactly the case that was missing. I re-ran the mutation that motivated it — deleting `CardinalityLimit: metricsCardinalityLimit(conf, logger),` from `metricsExporterConfig`, which turns the feature into a silent no-op — and it now goes red:

```
--- FAIL: Test_metricsExporterConfig/cardinality_limit_is_wired_into_the_config
```

Previously that deletion passed every test in the repo. Good call putting the "why" in a comment on the case; that is the part that stops someone deleting it later as redundant.

**The off-by-one wording is now correct.** "The limit is inclusive of that overflow slot, so a value of `n` keeps up to `n-1` real label sets" matches `internal/aggregate/limit.go:36` at the pinned v1.45.0 — `len(measurements) >= l.aggLimit-1`. Both doc strings updated, and the configs-reference version says the same thing without contradicting the `2000` default column.

**Feature code is untouched since my last pass.** `git diff e528d7327 8b29324ab` over `metrics_exporter.go`, `provider.go` and `config.go` is empty, so the live E2E from the previous review still stands — the nine env-var permutations, the `OTEL_GO_X_CARDINALITY_LIMIT` precedence in both directions, and the invalid-value path that warns without echoing the value.

Re-ran the gates on this head anyway: `go build`, `go vet`, `go test -race` across container/metrics/exporters, and `golangci-lint --new-from-rev=origin/development` — all clean, 0 issues. CI is 14/14.

Approving again. Nothing further from me.

Issue #3460, opened 2026-05-23
HTTP metrics middleware records raw r.URL.Path as Prometheus label, breaking exposition on non-UTF-8 paths


statoon54:
**Describe the bug**
The HTTP `Metrics` middleware (`pkg/gofr/http/middleware/metrics.go`) records `r.URL.Path` verbatim as the value of the Prometheus `path` label on the `app_http_response` histogram,
  with no UTF-8 validation. Any request whose URL path contains non-UTF-8 bytes (vulnerability scanners like Nessus routinely send IIS Unicode path-traversal probes such as
  `\xc0.\xc0./winnt/win.ini`) poisons the OTel registry. Subsequent scrapes of `/metrics` then fail with:

  An error has occurred while serving metrics:

error collecting metric Desc{fqName: "app_http_response", help: "Response time of HTTP requests in seconds.", constLabels: {}, variableLabels: {method,path,status,otel_scope_name,otel_scope_version,otel_scope_schema_url}}: label value "/\xc0.\xc0./winnt/win.ini" is not valid UTF-8

  Per the Prometheus exposition format spec, label values **must** be valid UTF-8, so this is a violation of contract — not just a cosmetic issue. Once a single bad series enters the
  registry, the whole scrape can be reported as failing.

**To Reproduce**

1. The code is

  Minimal app:

  ```go
  app := gofr.New()
  app.GET("/healthz", func(*gofr.Context) (any, error) { return "ok", nil })
  app.Run()
  ```

  Send a non-UTF-8 path and scrape:

  ```bash
  curl -i "http://localhost:8000/$(printf '\xc0\x2e\xc0\x2e')/winnt/win.ini"
  # → 404 (expected — no such route)

  curl -s http://localhost:2121/metrics
  # → "An error has occurred while serving metrics: ... label value ...
  #    is not valid UTF-8"
  ```

2. The error is

```
  error collecting metric Desc{fqName: "app_http_response", help: "Response time of HTTP requests in seconds.", constLabels: {}, variableLabels: {method,path,status,otel_scope_name,otel_scope_version,otel_scope_schema_url}}: label value "/\xc0.\xc0./winnt/win.ini" is not valid UTF-8
```

**Expected behavior**
  `app_http_response`'s `path` label should always be valid UTF-8
  (Prometheus contract) and should have bounded cardinality (best
  practice, no DoS surface from a scanner).

**Environments (please complete the following information):**
  - gofr: reproduced on `v1.56.X`
[...]

PR #4163, opened 2026-09-08
fix(metrics): keep caller-controlled labels from crowding out real routes

aryanmehrotra:
**Description:**

HTTP metrics label every request with its **path** and **method**. Both are caller-controlled on a 404, so a client hitting random URLs mints a new label pair per request.

The OTel SDK caps datapoints per instrument: **2000 by default**, or `OTEL_GO_X_CARDINALITY_LIMIT`, or `METRICS_CARDINALITY_LIMIT` (#4160). Past that cap, real routes stop getting their own series, and those are exactly the ones you need during an incident.

**Fix: bound the whole `(path, method, status)` label set, not each label separately.** Independent ceilings *multiply*, which is how you end up far looser than the SDK's own limit while looking bounded. The set is also the unit the SDK ceiling counts, so the budget maps 1:1 onto datapoints.

- Once the budget is used up, unmatched paths collapse to `__unmatched__` and unknown methods to `__other__`
- **The budget follows the configured ceiling:** a quarter of it, capped at 500 (a quarter of the SDK default). An unlimited ceiling (`0`) keeps the 500 cap, so memory stays bounded.
- `GetConfigs` takes the ceiling from `container.MetricsCardinalityLimit`, the one resolver the provider config also uses, and passes it through a new `middleware.WithCardinalityLimit` option. `Metrics(m, opts...)` is variadic, so existing `Metrics(m)` callers still compile.
- The provider ceiling itself is **not** touched. An earlier revision hard-coded `WithCardinalityLimit(2000)`, which overrode both env vars and failed `TestBuild_appliesCardinalityLimit`; it has been removed.

**Local end-to-end run**

A real GoFr app (templated `/users/{id}` plus a static directory). The traffic: **3000 scanner paths** and **300 invented methods**, then 20 requests to `/users/{id}` and 5 to `/static/app.js`, then `/metrics` scraped (median of 5 scrapes).

| build | scenario | series | `/users/{id}` | overflow | `/metrics` size | scrape |
|---|---|---:|---|---|---:|---:|
| `development` | default | 2000 | **lost** | yes | 6.6 MB | 42 ms |
| `development` | limit 100 (either env var) | 100 | **lost** | yes | 336 KB | 2.8 ms |
| earlier head `66502bbd4` | limit 100 (either env var) | 518 | kept | no, **limit ignored** | — | — |
| **this PR** | default | **506** | **kept (20)** | no | **1.7 MB** | **13 ms** |
| **this PR** | limit 100 (either env var) | **31** | **kept (20)** | no | **111 KB** | **1.2 ms** |

Malformed input on every build (8 KB path, `%00`, dot segments, a wrong method, a bad request line) returned only `404`/`400`, with **0** 5xx responses and no panics or ERROR lines in the logs.

**Performance**

| measurement | `development` | this PR |
|---|---|---|
| middleware, templated route (real manager + SDK) | 10 allocs, 1232 B | **10 allocs, 1232 B** |
| middleware, unmatched / scanner / invented methods | 13 allocs, 1352 B | **13 allocs, 1352 B** |
| time per request, single core, n=10 interleaved | — | **no significant change** (geomean −0.19%, p ≥ 0.39) |
| `BenchmarkMetricLabels`, the added step | — | templated **1.4 ns**, untemplated **~22.5 ns**, **0 allocs** |

The benefit shows up where the SDK does its work: fewer held datapoints, less memory, and ~4× smaller, ~3× faster `/metrics` scrapes under scanner traffic.

**Known limit, stated rather than hidden**

- **First-come, no eviction.** A cumulative histogram never resets (`sdk/metric` `internal/aggregate/histogram.go`), so evicting a pair here would free nothing in the SDK, and admitting a replacement would mint a new datapoint. The cost: once junk uses the budget, other untemplated traffic, static assets included, is recorded as `__unmatched__`. It is still counted and timed, just not per path. The live run shows `/static/app.js` collapsing after the flood.

**None.**

| | |
|---|---|
| Matched routes | **unchanged**, same labels as before |
| Metric names | unchanged |
| Exported API | **additive**: `middleware.WithCardinalityLimit`, `middleware.Config.MetricsCardinalityLimit`, and a variadic `Metrics(m, opts...)` |
| `METRICS_CARDINALITY_LIMIT` / `OTEL_GO_X_CARDINALITY_LIMIT` | **unchanged**, still honored |

Only unmatched paths and unknown methods look different, and only past the budget.

**Additional Information:**

- No new dependencies. `docs/references/configs` documents the budget under `METRICS_CARDINALITY_LIMIT`.
- Tests: set bound, **`TestMultiStatusPairsStayWithinBudget`** (3000 paths × 4 statuses: 2004 sets without the status in the key), invented methods, templated route under junk, **`TestLowCeilingKeepsRoutesReachable`**, `TestUntemplatedBudget`, `TestGetConfigs_MetricsCardinalityLimit` (precedence, invalid falls through, `0`, whitespace), `TestBuild_unsetCardinalityLimitHonorsSDKEnv`.
- **Mutation-checked:** dropping `status` from the admission key fails the multi-status test; ignoring the ceiling in `untemplatedBudget` fails 2 tests; swapping the env precedence fails 1; re-adding a hard-coded provider limit fails 2.
- `go test -race` for `http/middleware`, `metrics/exporters` and `container`, plus `go test ./pkg/gofr/`: pass. `golangci-lint --new-from-rev`: 0 issues.

**Checklist:**

### DIFF
--- a/pkg/gofr/http/middleware/metrics.go
+++ b/pkg/gofr/http/middleware/metrics.go
@@ -79,12 +79,102 @@ func Metrics(metrics metrics) func(inner http.Handler) http.Handler {
 			// nothing, net/http implicit-200) to http.StatusOK so neither the
 			// histogram nor any status cache is poisoned with status=0.
 			defer func(res *StatusResponseWriter, req *http.Request) {
-				recorder.record(path, req.Method, res.Status(), time.Since(start).Seconds(), templated)
+				pathLabel, methodLabel := metricLabels(path, req.Method, templated)
+				recorder.record(pathLabel, methodLabel, res.Status(),
+					time.Since(start).Seconds(), templated)
 			}(srw, r)
 
 			inner.ServeHTTP(srw, r)
 		})
 	}
+}
+
+// The labels recorded for a request whose path or method is caller-controlled
+// rather than drawn from the route table.
+//
+// A histogram retains one datapoint per distinct attribute set for the life of
+// the process, and the set is keyed on the (path, method, status) triple. Both
+// halves of that key come off the request for anything that does not resolve to
+// a route template: net/http accepts any RFC 7230 token as a method, and GoFr's
+// PathPrefix("/") catch-all -- registered with no .Methods() restriction -- means
+// an unmatched request still runs this middleware rather than being turned away.
+//
+// The process cannot run out of memory over this: the meter provider sets an
+// explicit datapoint ceiling (see metrics/exporters/provider.go). What it can do
+// is lose its own metrics. Datapoints past that ceiling are folded into a single
+// overflow series, so junk arriving first evicts nothing but occupies the slots,
+// and the REAL routes registered afterwards are the ones that vanish from the
+// dashboard. Collapsing caller-controlled labels keeps the junk to one series and
+// leaves the ceiling for routes that matter.
+const (
+	unmatchedPathLabel = "__unmatched__"
+	otherMethodLabel   = "__other__"
+
+	// untemplatedLabelLimit is how many distinct caller-controlled (path, method)
+	// pairs are recorded verbatim before both halves collapse.
+	//
+	// It bounds the PAIR rather than each half separately, because two independent
+	// ceilings multiply: 4096 paths and 64 methods is a quarter of a million
+	// possible series, which is far looser than the provider ceiling it is meant
+	// to protect. One ceiling on the pair is the quantity that actually maps onto
+	// datapoints.
+	//
+	// The value is deliberately a small fraction of that provider ceiling, so
+	// caller-controlled series can never crowd out the route table.
+	untemplatedLabelLimit = 512
+)
+
+// untemplatedLabels is the set of caller-controlled (path, method) pairs
+// admitted verbatim.
+//
+// It is process-wide, unlike the per-Metrics() recorder caches, because the
+// resource it protects -- the meter provider's datapoint table -- is also
+// process-wide.
+//
+// Admission is check-then-act, so concurrent first sightings can overshoot by
+// roughly the number of requests in flight. The overshoot is bounded by
+// concurrency rather than by the caller, and the population stays finite.
+//
+//nolint:gochecknoglobals // process-wide by necessity: see above.
+var untemplatedLabels routeCache[struct{}]
+
+// untemplatedKey identifies one caller-controlled label pair.
+type untemplatedKey struct{ path, method string }
+
+// metricLabels returns the path and method labels to record.
+//
+// A templated path and a standard method are bounded by construction and are
+// always recorded as themselves. Anything else is recorded verbatim until the
+// ceiling and collapses afterwards, which keeps existing behavior -- recording
+// the raw path for unrouted and static requests, and the verbatim method
+// including custom verbs, are both deliberately tested -- while stopping a
+// caller from filling the datapoint table.
+func metricLabels(path, method string, templated bool) (pathLabel, methodLabel string) {
+	standardMethod := cacheableMethod(method)
+	if templated && standardMethod {
+		return path, method
+	}
+
+	key := untemplatedKey{path: path, method: method}
+	if _, admitted := untemplatedLabels.load(key); admitted {
+		return path, method
+	}
+
+	if untemplatedLabels.len() >= untemplatedLabelLimit {
+		if !templated {
+			path = unmatchedPathLabel
+		}
+
+		if !standardMethod {
+			method = otherMethodLabel
+		}
+
+		return path, method
+	}
+
+	untemplatedLabels.store(key, struct{}{})
+
+	return path, method
 }
 
 // metricsPath resolves the path label and reports whether it is a bounded route
--- a/pkg/gofr/http/middleware/metrics_cardinality_test.go
+++ b/pkg/gofr/http/middleware/metrics_cardinality_test.go
@@ -1,0 +1,124 @@
+package middleware
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// cardRecorder captures the label sets the middleware actually records, so these
+// tests observe the wiring rather than the helper it calls.
+type cardRecorder struct{ labels [][]string }
+
+func (c *cardRecorder) RecordHistogram(_ context.Context, _ string, _ float64, labels ...string) {
+	c.labels = append(c.labels, labels)
+}
+
+func (*cardRecorder) IncrementCounter(context.Context, string, ...string)            {}
+func (*cardRecorder) DeltaUpDownCounter(context.Context, string, float64, ...string) {}
+func (*cardRecorder) SetGauge(string, float64, ...string)                            {}
+
+// series returns the distinct (path, method) pairs recorded -- one metric series
+// each, and one retained datapoint each.
+func (c *cardRecorder) series() map[[2]string]int {
+	out := map[[2]string]int{}
+
+	for _, l := range c.labels {
+		var pair [2]string
+
+		for i := 0; i+1 < len(l); i += 2 {
+			switch l[i] {
+			case "path":
+				pair[0] = l[i+1]
+			case "method":
+				pair[1] = l[i+1]
+			}
+		}
+
+		out[pair]++
+	}
+
+	return out
+}
+
+// cardServe drives the real Metrics middleware, with no router, so every request
+// is untemplated -- the state an unmatched request reaches this middleware in.
+func cardServe(t *testing.T, rec *cardRecorder, method, path string) {
+	t.Helper()
+
+	h := Metrics(rec)(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusNotFound)
+	}))
+
+	req := httptest.NewRequestWithContext(context.Background(), method, path, http.NoBody)
+	h.ServeHTTP(httptest.NewRecorder(), req)
+}
+
+// TestUntemplatedSeriesAreBounded is what this change exists for.
+//
+// A histogram retains a datapoint per distinct attribute set for the life of the
+// process, and the meter provider caps how many it will hold. Series past that
+// cap fold into one overflow bucket, so caller-controlled labels arriving first
+// occupy the slots and the real routes registered later are the ones that
+// disappear from the dashboard. This asserts the caller-controlled population is
+// finite, measured through the middleware rather than through the helper.
+func TestUntemplatedSeriesAreBounded(t *testing.T) {
+	untemplatedLabels.reset()
+	defer untemplatedLabels.reset()
+
+	rec := &cardRecorder{}
+
+	for i := range untemplatedLabelLimit * 3 {
+		cardServe(t, rec, http.MethodGet, fmt.Sprintf("/no/such/path/%d", i))
+	}
+
+	assert.LessOrEqual(t, len(rec.series()), untemplatedLabelLimit+1,
+		"%d invented paths produced %d metric series", untemplatedLabelLimit*3, len(rec.series()))
+}
+
+// TestInventedMethodsAreBounded closes the other half of the key. Holding the
+// path fixed and varying the method reaches the same datapoint table, and would
+// never consult a path-only ceiling.
+func TestInventedMethodsAreBounded(t *testing.T) {
+	untemplatedLabels.reset()
+	defer untemplatedLabels.reset()
+
+	rec := &cardRecorder{}
+
+	for i := range untemplatedLabelLimit * 3 {
+		cardServe(t, rec, fmt.Sprintf("M%05d", i), "/fixed")
+	}
+
+	assert.LessOrEqual(t, len(rec.series()), untemplatedLabelLimit+1,
+		"%d invented methods produced %d metric series", untemplatedLabelLimit*3, len(rec.series()))
+}
+
+// TestTemplatedRoutesSurviveJunkTraffic pins the point of bounding at all: a real
+// route must still get its own series while junk is holding the ceiling down.
+func TestTemplatedRoutesSurviveJunkTraffic(t *testing.T) {
+	untemplatedLabels.reset()
+	defer untemplatedLabels.reset()
+
+	junk := &cardRecorder{}
+	for i := range untemplatedLabelLimit + 50 {
+		cardServe(t, junk, http.MethodGet, fmt.Sprintf("/junk/%d", i))
+	}
+
+	rec := &cardRecorder{}
+	router := mux.NewRouter()
+	router.Use(Metrics(rec))
+	router.NewRoute().Methods(http.MethodGet).Path("/users/{id}").
+		Handler(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) { w.WriteHeader(http.StatusOK) }))
+	router.ServeHTTP(httptest.NewRecorder(),
+		httptest.NewRequestWithContext(context.Background(), http.MethodGet, "/users/42", http.NoBody))
+
+	require.Len(t, rec.labels, 1)
+	assert.Contains(t, rec.series(), [2]string{"/users/{id}", http.MethodGet},
+		"a templated route keeps its own series regardless of junk traffic")
+}
--- a/pkg/gofr/metrics/exporters/provider.go
+++ b/pkg/gofr/metrics/exporters/provider.go
@@ -20,6 +20,11 @@ import (
 // MeterProvider. It is safe to call more than once (later calls are no-ops).
 type ShutdownFunc func(ctx context.Context) error
 
+// datapointCardinalityLimit caps how many distinct attribute sets any one
+// instrument retains. It matches the SDK default it replaces, so this changes
+// no behavior today -- it fixes the value so a dependency bump cannot.
+const datapointCardinalityLimit = 2000
+
 // Build assembles the application MeterProvider from the always-on Prometheus
 // pull reader plus any push reader selected by cfg.Exporter, and returns a
 // ShutdownFunc so the caller can flush and shut it down on exit.
@@ -51,6 +56,18 @@ func Build(ctx context.Context, cfg *Config, logger Logger) (ShutdownFunc, metri
 	if r := pushReader(ctx, cfg, logger); r != nil {
 		opts = append(opts, metricSdk.WithReader(r))
 	}
+
+	// Make the datapoint ceiling GoFr's own rather than an inherited default.
+	//
+	// A histogram retains one datapoint per distinct attribute set for the life
+	// of the process, so the ceiling is what stops a metric growing without
+	// bound. The SDK does supply one -- 2000, from cardinalityLimitFromEnv when
+	// OTEL_GO_X_CARDINALITY_LIMIT is unset -- but inheriting it silently means
+	// GoFr's memory behavior is defined by a constant in a dependency that can
+	// change under a version bump, and that nothing in GoFr documents or tests.
+	// Setting it explicitly costs nothing and makes the bound a property of the
+	// framework.
+	opts = append(opts, metricSdk.WithCardinalityLimit(datapointCardinalityLimit))
 
 	mp := metricSdk.NewMeterProvider(opts...)
 	meter := mp.Meter(cfg.AppName, metric.WithInstrumentationVersion(cfg.AppVersion))


PiyushSingh-ZS (commented):
The problem is real and the framing is right — path and method are both caller-controlled on a 404, GoFr's `PathPrefix("/")` catch-all means unmatched requests still reach this middleware, and bounding the **pair** rather than each half separately is the correct instinct since two independent ceilings multiply.

I checked out the branch and ran the new tests (`go test ./pkg/gofr/http/middleware/ -run 'Untemplated|InventedMethods|TemplatedRoutesSurvive'`) — they pass. One finding below that I think needs resolving before merge, and one design question.

---

### The cardinality limit is not a no-op change

The PR says:

> `WithCardinalityLimit(2000)` set explicitly on the provider, so the bound is stated rather than inherited

and the code comment says:

> It matches the SDK default it replaces, so this changes no behavior today -- it fixes the value so a dependency bump cannot.

That second claim does not hold. From `go.opentelemetry.io/otel/sdk/metric@v1.45.0/config.go`:

```go
func newConfig(options []Option) config {
	conf := config{
		res:              resource.Default(),
		exemplarFilter:   exemplar.TraceBasedFilter,
		cardinalityLimit: cardinalityLimitFromEnv(),   // OTEL_GO_X_CARDINALITY_LIMIT
	}

	for _, o := range meterProviderOptionsFromEnv() {
		conf = o.apply(conf)
	}

	for _, o := range options {       // explicit options applied last
		...
		conf = o.apply(conf)
	}

	return conf
}
```

The env variable only seeds the *default*; an explicit `WithCardinalityLimit` is applied afterwards and wins. The SDK's own doc on that option says as much:

```go
// For backward compatibility, the environment variable `OTEL_GO_X_CARDINALITY_LIMIT`
// can also be used to set this value.
```

So today an operator with a large route table can raise the ceiling with `OTEL_GO_X_CARDINALITY_LIMIT=10000`. After this PR that variable is inert for every GoFr service, and the ceiling is hard-wired at 2000 with no GoFr-side way to change it. That is a behavior change, and it removes a capability rather than pinning one.

Two ways out, either fine by me:

- Read the env variable first and only pass the option when it is unset — preserves today's behavior exactly while still defending against a dependency bump; or
- Introduce a GoFr config key (`METRICS_CARDINALITY_LIMIT`, default 2000) and pass that through. This is more in keeping with the framework and puts the bound in `docs/references/configs` where operators go looking, which seems closer to the stated goal of making it "a property of the framework" anyway.

Either way the "changes no behavior today" line in the comment needs to go.

---

### Admission is first-come with no eviction

`untemplatedLabels` fills to 512 and freezes; nothing is ever evicted. The PR argues against the SDK's overflow behavior in these terms:

> Datapoints past that ceiling are folded into a single overflow series, so junk arriving first evicts nothing but occupies the slots, and the REAL routes registered afterwards are the ones that vanish from the dashboard.

That argument reproduces one layer down. A burst of junk in the first seconds after startup permanently occupies all 512 slots, and legitimately-untemplated traffic afterwards collapses to `__unmatched__` for the life of the process. That is not a small set: `metricsPath` sets `templated = false` for every static-asset extension, for `/static` prefixes, and for the `/` catch-all, so static file metrics are exactly the traffic that gets crowded out.

Templated routes short-circuit before the cache and are safe, so the blast radius is bounded — but it is the same failure mode the PR is written to prevent, and I do not think it is currently acknowledged. Either a note in the comment explaining why it is an acceptable trade at this layer, or a small LRU, would close it.

---

### Smaller notes

- **Two ceilings on one cache.** `metricLabels` gates on `untemplatedLabels.len() >= untemplatedLabelLimit` (512) while `routeCache.store` independently refuses past `routeCacheLimit` (4096). Only the smaller can ever fire, so the second is dead — harmless, but a reader has to work that out for themselves. One sentence on `untemplatedLabelLimit` saying it deliberately sits below the shared `routeCache` bound would save that.
- **The global.** CONTRIBUTING is explicit about no globals and injecting dependencies. The `//nolint:gochecknoglobals` justification here — the resource being protected is itself process-wide — is genuinely the strongest available, and having it as a conscious nolint rather than a silent `var` is the right call. Raising it only because a maintainer will ask, not because I think it is wrong.
- **Test isolation.** Three tests mutate the package global with `reset()` / `defer reset()`. Correct today because none of them call `t.Parallel()`, but that is an invariant nothing states. A line on the `untemplatedLabels` declaration — "tests that assert on this must not run in parallel" — would stop the next person adding `t.Parallel()` and getting a confusing flake.
- I did measure whether boxing `untemplatedKey` into `sync.Map.Load(any)` costs a per-request allocation on the unmatched path, since that would be an unfortunate trade in a defensive change. It does not: **0 allocs/op** on both the cached-untemplated and templated paths. No issue, noting it so nobody else has to check.

The `cardRecorder` approach of observing the label sets the middleware actually records, rather than testing `metricLabels` directly, is the right level to assert at.

akshat-kumar-singhal:
A production data point for this PR, plus a measurement of something it bounds that I think is worth stating explicitly.

## The 404 label problem isn't hypothetical

A low-traffic auth service on `gofr.dev v1.60.1` (Cloud Run, single instance, essentially zero organic traffic — ~36 lifetime logins), measured over 30 days:

- **600** distinct `path` values on `app_http_response`
- **1264** series total, of which **1133 (90%)** are `status="404"` — all credential-scanner probes: `/.env`, `/.env.bak`, `/.env.prod.local`, `/.aws/credentials`, `/.git/config`, `/.docker/config.json`, …
- only **98** series are `200`

So **1264 of the SDK's 2000-datapoint ceiling is consumed by scanner noise** — precisely the crowding-out this PR describes, on a service with no real traffic to speak of.

One aspect that surprised me: the cost is not per-request. OTel re-exports every series it holds each interval, so a single scanner hit on Monday is still being exported on Friday. I measured 172 series re-exported every 30s, dead flat, 24h/day, completely decoupled from request volume. That one instrument accounted for **95%** of the project's billed Prometheus samples.

## Retention measurement

The PR comment notes that "the process cannot run out of memory over this: the meter provider sets an explicit datapoint ceiling." That's true of the provider's datapoint table, but on `v1.60.1` the middleware's own `(path, method)` attribute cache isn't subject to that ceiling — it's `Load`/`LoadOrStore` with no `Delete` and no cap, while the comment above it asserts it is "bounded by routes × methods (typically <100 entries even for large APIs)".

Driving 50,000 distinct unmatched paths through `middleware.Metrics`, measuring `HeapAlloc` after `runtime.GC()`:

| metrics implementation | retained |
|---|---|
| implements `RecordHistogramAttrs` (fast path — what `metricsManager` does) | 15.12 MB (**317 B/path**) |
| plain `metrics` only (fast path skipped) | 4,192 B |

**3,783×.** Running identical traffic through both paths isolates the retention to the cache rather than to per-request allocation. That extrapolates to ~302 MB per 1M distinct paths, never freed, on an instance that never restarts.

`untemplatedLabelLimit = 512` caps this, so **this PR fixes it** — I raise it only because bounded memory seems like a second, independent reason to land the change, and it isn't mentioned in the description.

## Released versions are affected

`v1.60.1` is the current release and has no `METRICS_CARDINALITY_LIMIT` (#3567 was closed unmerged on 2026-09-07) while pulling `otel/sdk/metric v1.45.0` — so it inherits the 2000 ceiling with no way to configure it *and* no bound on the `path` label. Same root cause as #3460.

## Repro

Go 1.27, `gofr.dev v1.60.1`. External package; the method names on the two middleware interfaces are exported, so both are implementable from outside.

```go
package gofrrepro

import (
	"context"
	"fmt"
	"net/http"
	"net/http/httptest"
	"runtime"
	"testing"

	"go.opentelemetry.io/otel/attribute"
	"gofr.dev/pkg/gofr/http/middleware"
)

type discardPlain struct{}

func (discardPlain) IncrementCounter(context.Context, string, ...string)              {}
func (discardPlain) DeltaUpDownCounter(context.Context, string, float64, ...string)   {}
func (discardPlain) SetGauge(string, float64, ...string)                              {}
func (discardPlain) RecordHistogram(context.Context, string, float64, ...string)      {}

// Adding this method alone selects the fast path, and with it routeAttrs.
type discardAttrer struct{ discardPlain }

func (*discardAttrer) RecordHistogramAttrs(context.Context, string, float64, ...attribute.KeyValue) {}

var notFound = http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
	w.WriteHeader(http.StatusNotFound)
})

func measure(n int, h http.Handler) uint64 {
	var before, after runtime.MemStats
	runtime.GC()
	runtime.ReadMemStats(&before)

	for i := 0; i < n; i++ {
		req := httptest.NewRequest(http.MethodGet, fmt.Sprintf("/.env.scanner-%d", i), nil)
		h.ServeHTTP(httptest.NewRecorder(), req)
	}

	runtime.GC()
	runtime.ReadMemStats(&after)
	runtime.KeepAlive(h)

	return after.HeapAlloc - before.HeapAlloc
}

func TestLeakIsolatedToRouteAttrs(t *testing.T) {
	const n = 50000

	fast := measure(n, middleware.Metrics(&discardAttrer{})(notFound))
	slow := measure(n, middleware.Metrics(discardPlain{})(notFound))

	t.Logf("fast (routeAttrs active):   %d bytes (%.0f B/path)", fast, float64(fast)/float64(n))
	t.Logf("slow (routeAttrs bypassed): %d bytes", slow)

	if fast <= slow {
		t.Fatalf("expected fast path to retain more: fast=%d slow=%d", fast, slow)
	}
}
```

```
fast (routeAttrs active):   15857912 bytes (317 B/path)
slow (routeAttrs bypassed): 4192 bytes
```

A companion test confirms the label side: 10,000 distinct unmatched URLs produce 10,000 distinct `path` values, while 10,000 requests to a `mux`-registered `/users/{id}` collapse to exactly 1.

Happy to move this to a standalone issue if you'd rather keep the PR thread focused.

aryanmehrotra:
@PiyushSingh-ZS thanks. Your first finding turned out to be worse than a no-op on the current head. Addressed in 10d3124ec:

**Cardinality limit: removed, not reworked.** After `development` was merged in, `Build` applied #4160's `METRICS_CARDINALITY_LIMIT` option and then this PR's `WithCardinalityLimit(2000)`. In the pinned `sdk/metric@v1.46.0` (`config.go` `newConfig`, lines 87–92), options are applied in order, so the 2000 won. That made **both** `METRICS_CARDINALITY_LIMIT` and `OTEL_GO_X_CARDINALITY_LIMIT` inert, and it is why `TestBuild_appliesCardinalityLimit` failed on all three `PKG Unit Testing` jobs (`got 5 data points, want <= 3`). Your option 2 already exists as #4160, so the override is gone. The new `TestBuild_unsetCardinalityLimitHonorsSDKEnv` pins the unset case. Confirmed live: with `METRICS_CARDINALITY_LIMIT=100` or `OTEL_GO_X_CARDINALITY_LIMIT=100`, the previous head held 518 series and this head holds 100.

**First-come admission: kept, now documented.** I did not add an LRU because it would not bound anything. A cumulative histogram never resets (`internal/aggregate/histogram.go`, "It is never reset"), so evicting a pair from our set frees nothing in the SDK, and admitting a new one mints another datapoint. The comment now says so, and states the cost you described: after a junk burst, static assets record as `__unmatched__` (still counted and timed). The live run reproduces exactly that.

**Smaller notes:** added the sentence about `untemplatedLabelLimit` sitting below `routeCacheLimit`, and a line on the set saying tests must not call `t.Parallel()`. The `nolint` justification is unchanged.

**One more gap from the local run, also documented:** `untemplatedLabelLimit` is a constant, so with `METRICS_CARDINALITY_LIMIT` below ~600 the 512 caller-controlled pairs can fill the ceiling on their own, and `/users/{id}` overflowed at `=100`. Making it follow the configured ceiling means passing the limit from the container into the middleware. I would rather do that as a follow-up than grow this PR. Happy to go the other way if you prefer.

The full before/after table is in the updated description.

aryanmehrotra:
@akshat-kumar-singhal thanks for the production numbers. 90% of series being `404`s on a service with ~36 logins is exactly the case this PR targets, and they are now in the context for it.

I ran your retention repro unchanged against four trees (50,000 distinct unmatched paths, 2 runs each):

| tree | retained on the `RecordHistogramAttrs` path |
|---|---|
| `v1.60.1` | **15.9 MB (318 B/path)**, same as your figure |
| `development` (no #4163) | ~0 B/path (−65 KB to +25 KB, GC noise) |
| this PR, previous head | ~0 B/path |
| this PR, current head | ~0 B/path |

So you are right about v1.60.1, but **the memory fix is #3972, not this PR**. #3972 made the per-route attribute caches store only templated routes, bounded by `routeCacheLimit`, so unmatched paths never enter them. It is on `development` and ships in v1.61.0. This PR bounds the other half, the SDK datapoints and exported series, which is still unbounded on `development` (2000 series and `/users/{id}` crowded out in the local run). That is the part that reduces the billed-samples cost you measured.

`METRICS_CARDINALITY_LIMIT` (#4160) also ships in v1.61.0, which covers the "no way to configure it" point.

aryanmehrotra:
Update to my earlier reply: I said a budget that follows the ceiling would be a follow-up PR. It is in this one now (509a0e95e).

- `GetConfigs` resolves the effective ceiling (`METRICS_CARDINALITY_LIMIT` → `OTEL_GO_X_CARDINALITY_LIMIT` → 2000) and passes it through `middleware.WithCardinalityLimit`.
- Unmatched `(path, method)` pairs get **a quarter of the ceiling, capped at 500**. The cap was 512; it is now a quarter of the SDK default.
- Live at `METRICS_CARDINALITY_LIMIT=100` with 3000 scanner paths: **31 series, `/users/{id}` kept, no overflow**. Before this commit, `/users/{id}` was lost.
- Cost: `metricLabels` is 1.4 ns on templated routes and ~19 ns on unmatched ones, with 0 allocs. The real middleware shows identical allocs and bytes against `development`, and no significant time difference.

The updated description has the full before/after, including `/metrics` scrape size (6.6 MB → 1.7 MB under the same scan).

PiyushSingh-ZS (commented):
Re-reviewed at 509a0e95. Thanks for chasing the first point all the way down — you are right that it was worse than I described, and finding that it was already reddening `TestBuild_appliesCardinalityLimit` on all three `PKG Unit Testing` jobs is the part I missed. Removing the override rather than reworking it is the right call now that #4160 exists.

### Verified on the new head

- **The override is genuinely gone** and `Build` now only calls `WithCardinalityLimit` when `cfg.CardinalityLimit != nil`, so an unset `METRICS_CARDINALITY_LIMIT` leaves `OTEL_GO_X_CARDINALITY_LIMIT` and the SDK default intact. That closes the finding.
- **The two-key precedence in `metricsCardinalityLimit` agrees with what the provider actually applies**, which was not obvious to me and is worth recording: `EnvLoader.Get` is `os.Getenv`, and `godotenv.Load`/`Overload` push `configs/.env` into the process environment before either of these runs. So the middleware reading `OTEL_GO_X_CARDINALITY_LIMIT` through `config.Config` and the SDK reading it through `os.Getenv` cannot disagree, including for values that came from a `.env` file. Worth a half-sentence in the comment, because the next reader will wonder whether a `.env`-only value is visible to the SDK.
- `untemplatedBudget` edge behavior is sane: `-1`/`0` (unlimited) hold at 500, `100` gives 25, anything above 2000 stays capped at 500.

---

### New finding: the budget bounds pairs, the ceiling counts triples

This is the one thing I would want fixed before merge, and it is the same shape as the original "two ceilings multiply" argument that motivated the PR.

`untemplatedLabelLimit` bounds distinct **(path, method)** pairs. The provider ceiling counts distinct **attribute sets**, and `status` is in the attribute set on every recorder path:

```go
attribute.String("status", strconv.Itoa(code))                    // optionRecorder / attrsRecorder
"path", path, "method", method, "status", strconv.Itoa(status)    // labelsRecorder
```

So an admitted pair mints one datapoint *per distinct status it is ever observed with*. The quarter is a quarter of pairs, not of datapoints.

Measured on this head — 3000 distinct unmatched paths, each observed with 4 statuses, through the real middleware:

```
distinct (path,method) PAIRS          = 501    (budget = 500)
distinct (path,method,status) TRIPLES = 2004   <-- what the 2000 ceiling counts
```

The pair budget held exactly as designed, and caller-controlled traffic still consumed **100% of the default ceiling**. Two statuses per pair is enough to take half of it.

This is reachable rather than theoretical: `metricsPath` marks static-asset extensions, `/static` prefixes and the `/` catch-all as untemplated, and those legitimately return 200, 304, 403 and 404. A scanner sweeping a service that also serves static files produces exactly this mix.

**Why the existing tests do not catch it:** `cardRecorder.series()` is itself pair-shaped —

```go
func (c *cardRecorder) series() map[[2]string]int
```

— so the harness defines a "series" as `(path, method)` while the SDK defines it as `(path, method, status)`. `TestLowCeilingKeepsRoutesReachable` then asserts `len(junk.series()) <= ceiling/untemplatedCeilingShare+1` against that pair-shaped count, which is true by construction and cannot fail for this reason. Widening `series()` to a 3-tuple should make the gap visible to the assertions you already have.

On the fix itself, I do not have a strong preference:

- divide by `untemplatedCeilingShare` *and* an assumed status count, or
- state the assumption in the comment on `untemplatedCeilingShare` — "one status per unmatched pair; a pair observed with N statuses costs N datapoints, so the effective share is N/4" — and accept it.

The second is fine by me if you think multi-status untemplated traffic is rare; what I would rather avoid is the current comment, which reads as though the quarter is a quarter of the ceiling:

> a quarter leaves most of the ceiling for real routes

### Refinement on the eviction argument, not an objection

Your reasoning for keeping first-come admission is sound and I withdraw the LRU suggestion. One precision on the citation: the "It is never reset" line at `internal/aggregate/histogram.go:263` is inside `cumulativeHistogram`. The neighbouring `deltaHistogram` is the opposite — "a histogram whose internal storage is reset when it is collected", and "delta always clears values on collection" — and `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=delta` is user-settable on the OTLP exporters.

The conclusion still holds for GoFr, but for a stronger reason than the one cited: `Build` installs the Prometheus reader alongside any push reader, and Prometheus is always cumulative, so at least one pipeline retains datapoints for the life of the process no matter what temporality the push exporter is set to. That is a better sentence for the comment than the current one, because it does not depend on a claim that is only true of half the file.

The one case it does not cover is `prometheusReader` returning nil with a delta OTLP reader — then the SDK frees those datapoints every cycle while this set never does, so a single early junk burst permanently degrades static-asset labels for no benefit. Narrow enough to note rather than handle.

### Smaller

- Ceilings of 1–3 give a budget of 0, so the first unmatched request already collapses. Defensible, but it is the one input where the behavior is qualitatively different and it is not in `TestUntemplatedBudget`.
- `untemplatedLabels` is process-wide while `untemplatedLimit` is now captured per `Metrics()` call. One server per process makes this academic, but two middlewares constructed with different ceilings would share one set and enforce whichever limit checked first. A line on the global noting that the limit is per-instance would keep the `//nolint:gochecknoglobals` justification honest.
- Good call adding the `t.Parallel()` note and the `routeCacheLimit` sentence.

Separately: agreed with your read on the retention numbers — #3972 bounding the per-route attribute caches to templated routes is the memory fix, and this PR bounds the exported-series half. Those are different resources and it is worth that being on the record in the thread.

pkg/gofr/http/middleware/config.go
```diff
@@ -51,9 +67,26 @@ func GetConfigs(c config.Config) Config {
...
+
 	return middlewareConfigs
 }
 
+// metricsCardinalityLimit resolves the provider ceiling with the same precedence
+// the provider applies: METRICS_CARDINALITY_LIMIT (container/metrics_exporter.go,
+// which also warns on an invalid value), then OTEL_GO_X_CARDINALITY_LIMIT (read
+// by the SDK itself), then the SDK default. A value that does not parse is
+// skipped, as both of those skip it.
+func metricsCardinalityLimit(c config.Config) int {
```
Umang01-hash: This duplicates the env precedence already resolved in `container/metrics_exporter.go`. Parity holds today, but the two can drift silently and mis-size the budget. A shared helper (or a value handed down from the provider) would be safer than re-deriving it.

pkg/gofr/http/middleware/metrics.go
```diff
@@ -44,13 +44,52 @@ type routeMethodKey struct {
...
+	return func(o *metricsOptions) {
+		o.untemplatedLimit = untemplatedBudget(limit)
+	}
+}
+
+// untemplatedCeilingShare is the inverse of the fraction of the provider ceiling
+// that caller-controlled labels may occupy. A pair usually lands on a single
+// status, but can also add the collapsed __unmatched__ and __other__ series, so
+// a quarter leaves most of the ceiling for real routes.
+const untemplatedCeilingShare = 4
```
Umang01-hash, line 77: Quarter-of-ceiling assumes ~1 status per untemplated pair. 500 pairs each hitting several statuses can exceed a quarter of the datapoints. Fine as a heuristic — just flagging the assumption.

pkg/gofr/http/middleware/metrics.go
```diff
@@ -79,14 +118,124 @@ func Metrics(metrics metrics) func(inner http.Handler) http.Handler {
...
+// datapoint -- this set would stop bounding anything. The accepted cost is that
+// a burst of junk early in the process fills the set, after which other
+// untemplated traffic, static assets included, is recorded as __unmatched__. It
+// is still counted and timed, only not per path. Templated routes never reach
+// the set and are unaffected.
+//
+// Tests that assert on this set reset it and must not call t.Parallel().
+//
+//nolint:gochecknoglobals // process-wide by necessity: see above.
+var untemplatedLabels routeCache[struct{}]
```
Umang01-hash, line 212: `untemplatedLabels` is process-wide but the limit is per-`Metrics()` instance. Two servers with different limits share this set and use whichever instance's limit is passed per request. Harmless (both bound), just inconsistent.

Umang01-hash (commented):
A few optional nits — logic, tests, and E2E all check out (700 junk paths collapse to a single `__unmatched__` series while real routes keep theirs). Nothing blocking.

aryanmehrotra:
@PiyushSingh-ZS fixed in c0aec22f7. Your measurement reproduces exactly.

**Pairs vs sets (blocking):** the admission key is now `(path, method, status)`, so the budget counts the same unit as the ceiling, and a quarter really is a quarter of the datapoints. You were right that the test harness hid it: `series()` now returns the 3-tuple, and `TestMultiStatusPairsStayWithinBudget` is your case (3000 unmatched paths × {200, 304, 403, 404}). With `status` removed from the key, it fails with **"2004 attribute sets recorded against a budget of 500"**, your number. With it, the result is ≤ 500 plus one collapsed series per status. I fixed the comment on `untemplatedCeilingShare` rather than keeping the pair-shaped wording.

**Eviction argument:** thanks for the precision. The comment now rests on `Build` always installing the Prometheus reader, which is cumulative, rather than on the `cumulativeHistogram` line. It also names the narrow case where it does not hold: Prometheus fails to init and a delta push reader is the only pipeline.

**Smaller:**
- `.env` visibility: stated on `container.MetricsCardinalityLimit`, which is now the one resolver.
- Ceilings 1–3 → budget 0: stated on `WithCardinalityLimit`; `TestUntemplatedBudget` has `1`, `3` and `4` (the first non-zero budget).
- Per-instance budget vs process-wide set: a line on both `WithCardinalityLimit` and `untemplatedLabels`.

**Cost of the wider key:** `BenchmarkMetricLabels/untemplated_admitted` went from 19.0 ns to 22.5 ns (p=0.002), 0 allocs. Templated routes are unchanged at 1.4 ns.

**Live re-run on this head** (3000 scanner paths): default = 506 series, `/users/{id}` kept, no overflow, `/metrics` 1.7 MB. `METRICS_CARDINALITY_LIMIT=100` or `OTEL_GO_X_CARDINALITY_LIMIT=100` = 31 series, `/users/{id}` kept, no overflow. 0 5xx, logs clean.

pkg/gofr/http/middleware/config.go
aryanmehrotra: Agreed, and fixed in c0aec22f7: `container.MetricsCardinalityLimit(conf)` is now the single resolver (`METRICS_CARDINALITY_LIMIT` via the same `metricsCardinalityLimit` the provider config uses, then `OTEL_GO_X_CARDINALITY_LIMIT`, then 2000), and `GetConfigs` calls it. `middleware` already imported `container`, so no new dependency. It is a package function rather than a `Container` method so it is not promoted onto every handler's `Context`.

pkg/gofr/http/middleware/metrics.go
aryanmehrotra: This turned out to be more than a heuristic. Piyush measured 3000 paths × 4 statuses = 2004 datapoints against a 500-pair budget. Fixed in c0aec22f7: the admission key now includes `status`, so the budget counts attribute sets, the same unit as the ceiling. `TestMultiStatusPairsStayWithinBudget` pins it.

pkg/gofr/http/middleware/metrics.go
aryanmehrotra: Documented in c0aec22f7 on both `WithCardinalityLimit` and `untemplatedLabels`: the budget is per instance and the set is process-wide, so each instance stops admitting at its own budget and both stay bounded. GoFr builds one per server, so I kept it as a note rather than moving the budget onto the set.

PiyushSingh-ZS (commented):
Re-reviewed at c0aec22f. The blocking finding is resolved — I re-ran my own repro against this head and it now behaves as the comment claims.

### Verified

**The budget counts the right unit now.** Same input as before (3000 unmatched paths x {200, 304, 403, 404}) through the real middleware:

| | attribute sets recorded |
|---|---|
| 509a0e95 (pair key) | **2004** against a 500 budget |
| c0aec22f (set key) | **504** = 500 admitted + one collapsed series per status |

That is the ceiling behaving as described, and widening `cardRecorder.series()` to the 3-tuple means the assertions can now see it. `TestMultiStatusPairsStayWithinBudget` fails with the pair key and passes with the set key, so it bites.

**The two things I would have flagged in the new commit both check out:**

- `middleware` importing `container` is not a new dependency — `apikey_auth.go`, `basic_auth.go` and `web_socket.go` already import it, so there is no cycle and no coupling change. Your note on the inline thread is accurate.
- `container.MetricsCardinalityLimit` passing `nil` into `metricsCardinalityLimit(conf, nil)` is safe: the `if logger != nil` guard is pre-existing. Worth a word on the call site anyway — `nil` as a logger reads like an oversight to anyone who has not opened the callee, and the reason it is correct (the provider path already warns, so warning twice for one variable would be noise) is the kind of thing that gets "fixed" into a duplicate warning later.

Package tests pass locally on this head: `http/middleware`, `metrics/exporters`, `container`.

### One consequence of the wider key worth deciding on

Admission is now per `(path, method, status)`, so a pair can be **admitted for one status and collapsed for another**. Measured on this head — one asset admitted early, budget then exhausted by junk, same asset observed with a second status afterwards:

```
status=200 -> path label "/assets/app.css"
status=404 -> path label "__unmatched__"
```

So one resource's series split across two path labels. The part I would think about is the bias: a path's 2xx is typically seen at startup and gets admitted, while its 4xx/5xx first appear later, once the budget is gone. That means the statuses most likely to collapse are the ones you would actually query during an incident — a dashboard filtering `path="/assets/app.css"` keeps the healthy series and silently loses the error series, which reads as "no errors" rather than "not recorded".

Under the old pair key this could not happen: a pair was admitted or collapsed as a unit, so its statuses stayed together.

I do not think this blocks — it is strictly better than 2004 datapoints, and it only affects untemplated traffic. Options as I see them:

1. **Accept and document it** on `untemplatedKey`. Cheapest, and defensible: an untemplated path has no route template, so per-path error rates were never a supported query.
2. **Admit by pair, charge by set** — once `(path, method)` is admitted, let its later statuses through, and count each new status against the budget so the datapoint bound still holds. Keeps a path's statuses together at the cost of the budget being consumed in uneven chunks.

Option 1 is fine by me if you think untemplated per-path error rates are not something anyone queries. Mostly I want the choice to be deliberate, because the current comment on `untemplatedCeilingShare` explains the *unit* change correctly but does not mention that collapse is now status-selective.

### Smaller

- `distinct PATH labels retained` drops from ~500 to ~126 in the 4-status case, since each path now consumes one slot per status. That is the correct trade for bounding the right quantity, but it is a real reduction in path diversity and is not stated anywhere. One clause on `untemplatedLabelLimit` would cover it.
- The rewritten eviction comment is better than the version it replaces — resting it on `exporters.Build` always installing the cumulative Prometheus reader, and naming the delta-only-pipeline case it does not cover, is exactly right.
- Agreed on keeping the per-instance budget with a note rather than moving it onto the set. One server per process makes it academic and the note is where the ambiguity was.

Nothing outstanding from me beyond the status-collapse decision above.

Umang01-hash (approved):

PR #4164, opened 2026-09-08
fix(container): stop a typed-nil pub/sub client escaping, and stop isNil panicking

aryanmehrotra:
**Description:**

Two bugs in the same guard, which have to ship together.

**1. A typed nil passes `!= nil`.** `kafka.New` and `google.New` return a **typed nil** when they reject an incomplete config. Assigned into the `PubSub` interface, that is **not equal to `nil`** — so `health.go` admitted it and called `Health()` on a nil receiver.

`SQL` and `Redis` beside it already used `isNil`. **This guard was the odd one out.** Also filtered in **`GetSubscriber()` and `GetPublisher()`**, so the typed nil cannot escape the container at all. Both, not one: they return the same field, so a handler writing `ctx.GetPublisher().Publish(...)` hits the identical nil receiver — it just surfaces on a request rather than at startup, and guarding only one leaves the next reader guessing whether the asymmetry was deliberate.

**2. `isNil` itself panicked.**

```go
return !val.IsValid() || val.IsNil()   // before
```

`reflect.Value.IsNil` **panics** on a value whose kind cannot be nil. A `pubsub.Client`, `Redis` or `DB` implemented on a **value receiver** is ordinary Go — GoFr's own tests do it — and this took the process down: `Close()` calls `isNil` (`container.go:195`) from the shutdown goroutine, which has no recover. It now switches on `Kind` first; anything not nillable is present by definition.

**What each guard actually prevents**

Worth stating precisely, because the two paths differ:

| Path | Without the guard |
|---|---|
| `GetSubscriber` → `App.Subscribe` | **process death** — the panic escapes an errgroup with no recover, at startup |
| `GetPublisher` → a handler | panic on a request |
| `Health` | **not** a crash — `runCheck` recovers (`health.go:276`). It surfaced as a `"pubsub": DOWN` entry reading `health check panicked`, putting the app into **DEGRADED** over a dependency it does not have |
| `Close` (bug 2 only) | **process death** — no recover in the shutdown goroutine |

**None** — and fix 2 is what keeps it that way.

⚠️ Shipped alone, fix 1 **would be** a breaking change for anyone implementing these interfaces on a value receiver: making the guard stricter is exactly what makes the panic reachable. That is why both are in one PR.

| | |
|---|---|
| Exported API | unchanged |
| Health response shape | unchanged |
| Value-receiver implementations | **no longer crash** |

**Additional Information:**

- No new dependencies.
- `pubsub_health_nil_test.go` covers the typed nil through `Health`, `GetSubscriber` and `GetPublisher`, plus a value-receiver implementation that panicked before. The typed nil is constructed directly rather than by way of `google.New` rejecting an incomplete config, so the tests do not depend on what that constructor happens to validate.
- The health test asserts the `pubsub` key is **absent** from the health map, not `NotPanics`. `runCheck` already recovers, so a `NotPanics` assertion passes with or without the guard and proves nothing; the absent key fails when the guard is reverted.
- **Cost of the added `isNil` on the accessors: +5.5 ns/op, 0 allocations** (measured: 6.8 ns vs 1.3 ns for a plain comparison, `-benchtime=2000000x -count=3`). `GetPublisher` is the one that can sit on a per-request path, and 5.5 ns against a network publish is not a trade worth the nil-receiver panic.
- `isNil` carries `//nolint:exhaustive` — the `default` branch is the point, and the repo sets `default-signifies-exhaustive: false`. Same convention as `metrics/exporters/otlp.go`.
- Composes with #4167: its `pubsub_backends_disabled.go` stubs return an untyped `nil` for `createMqttPubSub` and leave `c.PubSub` unset for the other two, which this guard handles unchanged.

**Checklist:**

### DIFF
--- a/pkg/gofr/container/container.go
+++ b/pkg/gofr/container/container.go
@@ -438,7 +438,21 @@ func (c *Container) GetPublisher() pubsub.Publisher {
 	return c.PubSub
 }
 
+// GetSubscriber returns the pub/sub client, or nil when none is usable.
+//
+// The nil check is isNil rather than a plain comparison because the pub/sub
+// constructors assign the result of google.New or kafka.New straight into the
+// interface, and those return a TYPED nil when they reject an incomplete
+// config. A typed nil is not equal to nil, so returning c.PubSub unfiltered let
+// every caller's own nil guard pass and then call a method on a nil receiver.
+// Filtering here fixes all of them at once, including App.Subscribe, where the
+// panic escaped an errgroup with no recover and killed the process at startup
+// rather than on a request.
 func (c *Container) GetSubscriber() pubsub.Subscriber {
+	if isNil(c.PubSub) {
+		return nil
+	}
+
 	return c.PubSub
 }
 
--- a/pkg/gofr/container/health.go
+++ b/pkg/gofr/container/health.go
@@ -198,7 +198,13 @@ func (c *Container) checkPrimaryDatasources(ctx context.Context, collector *heal
 		})
 	}
 
-	if c.PubSub != nil {
+	// isNil, not a plain != nil: the pub/sub constructors assign the result of
+	// google.New or kafka.New straight into this interface, and those return a
+	// TYPED nil when they reject an incomplete config. A typed nil in an interface
+	// is not equal to nil, so a plain check admits it and the Health call below
+	// runs on a nil receiver. SQL and Redis above already use isNil for the same
+	// reason; this guard was the odd one out.
+	if !isNil(c.PubSub) {
 		runCheck(wg, collector, pubsubKey, func() {
 			health := c.PubSub.Health()
 			collector.record(pubsubKey, health, health.Status == datasource.StatusDown)
@@ -342,12 +348,30 @@ func (c *Container) appHealth(healthMap map[string]any, downCount int) {
 	}
 }
 
+// isNil reports whether i is absent: either an unset interface, or an interface
+// holding a nil pointer.
+//
+// The Kind check is not optional. reflect.Value.IsNil PANICS on a value whose
+// kind cannot be nil, and a datasource field can legitimately hold one: an
+// implementation of pubsub.Client, Redis or DB may be a struct value rather than
+// a pointer, which is ordinary Go and something GoFr's own tests do. Calling
+// IsNil on that took the process down.
+//
+// Anything not nillable is present by definition, so it reports false.
 func isNil(i any) bool {
-	// Get the value of the interface
 	val := reflect.ValueOf(i)
+	if !val.IsValid() {
+		return true
+	}
 
-	// If the interface is not assigned or is nil, return true
-	return !val.IsValid() || val.IsNil()
+	//nolint:exhaustive // the default is the point: every kind that cannot be nil is present.
+	switch val.Kind() {
+	case reflect.Chan, reflect.Func, reflect.Interface,
+		reflect.Map, reflect.Pointer, reflect.Slice, reflect.UnsafePointer:
+		return val.IsNil()
+	default:
+		return false
+	}
 }
 
 // stalledHealth is the body returned while a previous round's checks are still outstanding. It
--- a/pkg/gofr/container/pubsub_health_nil_test.go
+++ b/pkg/gofr/container/pubsub_health_nil_test.go
@@ -1,0 +1,86 @@
+package container
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"gofr.dev/pkg/gofr/config"
+)
+
+// typedNilPubSubContainer builds the state both tests below need: a container
+// whose PubSub holds a TYPED nil.
+//
+// google.New returns (*googleClient)(nil) when it rejects an incomplete config --
+// here a GOOGLE backend with no project id -- and createGooglePubSub assigns that
+// straight into the interface. A typed nil is not equal to nil, so every plain
+// "!= nil" guard downstream passes and then calls a method on a nil receiver.
+func typedNilPubSubContainer(t *testing.T) *Container {
+	t.Helper()
+
+	return NewContainer(config.NewMockConfig(map[string]string{
+		"PUBSUB_BACKEND": "GOOGLE",
+		"LOG_LEVEL":      "FATAL",
+	}))
+}
+
+// TestHealthSurvivesTypedNilPubSub pins the guard in Health, which called
+// Health() on the nil receiver and dereferenced a Config embedded by value. Any
+// request to the health endpoint took the process down.
+func TestHealthSurvivesTypedNilPubSub(t *testing.T) {
+	c := typedNilPubSubContainer(t)
+
+	require.NotPanics(t, func() { _ = c.Health(context.Background()) },
+		"Health must see through a typed nil PubSub")
+}
+
+// TestGetSubscriberFiltersTypedNil pins the more damaging one.
+//
+// App.Subscribe guards with `GetSubscriber() == nil` and then hands the client to
+// a goroutine in an errgroup that has no recover, so the nil-receiver panic there
+// killed the process at startup rather than on a request. Filtering in
+// GetSubscriber fixes every caller at once, so this is the assertion that keeps
+// them all honest.
+func TestGetSubscriberFiltersTypedNil(t *testing.T) {
+	c := typedNilPubSubContainer(t)
+
+	// assert.Nil is deliberately NOT used: it reflects, so it reports a typed nil
+	// as nil and would pass whether or not the filter exists. Callers write
+	// `GetSubscriber() == nil` -- App.Subscribe does exactly that -- so the test
+	// has to make the same plain comparison to mean anything.
+	assert.True(t, c.GetSubscriber() == nil, //nolint:testifylint // see above: a reflective nil check cannot fail here
+		"a typed nil must not escape GetSubscriber: callers compare it against nil")
+}
+
+// TestIsNilHandlesNonNillableKinds pins that isNil does not panic on a value
+// whose kind cannot be nil.
+//
+// reflect.Value.IsNil panics for a struct, a string, an int and so on, and a
+// datasource field can legitimately hold one: implementing pubsub.Client, Redis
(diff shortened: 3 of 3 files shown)


PiyushSingh-ZS (commented):
Both bugs are real, and I agree they have to ship together. One thing I think is missing before this lands.

### Verified

- `kafka.New` is declared `func New(conf *Config, logger pubsub.Logger, metrics Metrics) *kafkaClient` and returns a bare `nil` when `validateConfigs` fails (`pkg/gofr/datasource/pubsub/kafka/kafka.go`). Assigned into the `pubsub.Client` interface that is a non-nil interface holding a nil pointer, so `c.PubSub != nil` admits it. The premise holds.
- `reflect.Value.IsNil` does panic for `Struct`, `String`, `Int` and `Array`, so the second bug is reachable for any implementation on a value receiver.
- The argument for one PR rather than two is right: fix 1 alone makes the panic in fix 2 *newly* reachable, because tightening the guard is exactly what routes value-receiver implementations into `isNil`.

The `Kind` switch is the correct shape, and "anything not nillable is present by definition" is the right default.

### `GetPublisher` was left behind

```go
func (c *Container) GetPublisher() pubsub.Publisher {
	return c.PubSub          // unchanged
}

func (c *Container) GetSubscriber() pubsub.Subscriber {
	if isNil(c.PubSub) {     // fixed
		return nil
	}

	return c.PubSub
}
```

The reasoning in the new doc comment — *"Filtering here fixes all of them at once"* — applies identically to the publisher. A handler doing `ctx.GetPublisher().Publish(...)` against a typed-nil Kafka or Google client hits the same nil receiver; it just surfaces on a request instead of at startup.

Worth fixing in this PR rather than a follow-up, because the asymmetry is now *more* misleading than it was before: a reader seeing one of the two guarded will reasonably assume the guarded one is the house style and that the other was considered and deliberately left. A test case alongside `TestGetSubscriberFiltersTypedNil` would pin it.

### Nits

- `assert.True(t, c.GetSubscriber() == nil)` with the `//nolint:testifylint`: the justification is correct and genuinely subtle (a reflective nil check cannot fail here, so `require.Nil` would pass with or without the filter). My only worry is that it reads as a lint violation to skim past, and someone will "fix" it back. A named helper — `assertPlainNil(t, v)` — would carry the intent to the next reader without relying on the comment being read.
- `typedNilPubSubContainer` leans on `PUBSUB_BACKEND=GOOGLE` with no project id producing the typed nil. That is true today but it is an indirect way to construct the state under test, and it breaks if `google.New`'s validation changes. Constructing the typed nil directly would make the test independent of that.

### Interaction worth noting

If #4167 lands, its `pubsub_backends_disabled.go` stubs return an untyped `nil` for `createMqttPubSub` and leave `c.PubSub` unset for the other two, which composes correctly with this guard. Might be worth cross-linking so whoever merges second knows the two were considered together.

aryanmehrotra:
All three addressed at `7c0510846`, and chasing the second one turned up a defect in the PR's own test.

**`GetPublisher` is guarded.** You're right that the asymmetry was worse than either choice — a reader seeing one guarded would reasonably conclude the other was considered and deliberately left. `TestGetPublisherFiltersTypedNil` pins it, and fails when the guard is removed.

**⚠️ `TestHealthSurvivesTypedNilPubSub` was vacuous.** Adding a test case made me check what the old one actually caught, and the answer is nothing: `runCheck` already recovers (`health.go:276`), so the nil-receiver panic never escaped `Health` with *or without* the `isNil` guard, and `require.NotPanics` passed either way.

What the guard really prevents is this:

```
"pubsub": {"status":"DOWN","details":{"error":"health check panicked: ..."}}
"status": "DEGRADED"
```

An app with no usable pub/sub reporting a DOWN dependency and a degraded service, with a stack fragment for a reason. The test now asserts the `pubsub` key is **absent**, which is what "no pub/sub configured" has always looked like — and it fails when the guard is reverted.

The PR's claim that this path "took the process down" was overstated, and the comment above the guard is corrected: the crash is on the `GetSubscriber` path, where `App.Subscribe`'s errgroup has no recover.

**Nit 1 — the `//nolint` is now a named helper**, `assertPlainNil`, carrying the reason (a reflective `assert.Nil` reports a typed nil as nil, so it would pass with or without the filter) in a doc comment rather than a trailing comment someone skims past.

**Nit 2 — the typed nil is constructed directly** rather than by way of `google.New` rejecting an incomplete `GOOGLE` config, so the tests no longer depend on what that constructor happens to validate. The fake's `Health` dereferences its receiver for the same reason `googleClient.Health` does, so the state under test is the real shape.

**Cross-link added** to #4167 for the `pubsub_backends_disabled.go` interaction.

akshat-kumar-singhal (approved):
Approve with nits. `isNil` now covers every nillable kind (Chan, Func, Interface, Map, Pointer, Slice, UnsafePointer) with no panic path left, and the other callers (`Close`, the Health loop, migration's own `isNil`) behave correctly with it. Restoring `container.go`/`health.go` to base fails all four new tests (health reports `"health check panicked"`, both getter tests fail, isNil panics on a struct).

**Nit (arguably should-fix): root cause left in place** — [`container.go:572`](https://github.com/gofr-dev/gofr/blob/0a07edb0551452e518a0f73dfd97eb8f26a98f94/pkg/gofr/container/container.go#L572), [`container.go:589`](https://github.com/gofr-dev/gofr/blob/0a07edb0551452e518a0f73dfd97eb8f26a98f94/pkg/gofr/container/container.go#L589)
`kafka.New`/`google.New` can still store a typed nil in the exported `Container.PubSub`, so direct readers (`ctx.Container.PubSub`, future internal callers) get a non-nil interface wrapping a nil receiver. Suggest only assigning a real client, e.g. `if cl := google.New(...); cl != nil { c.PubSub = cl }`, keeping the getter filtering as defence in depth.

**Nit** — [`container.go:442`](https://github.com/gofr-dev/gofr/blob/0a07edb0551452e518a0f73dfd97eb8f26a98f94/pkg/gofr/container/container.go#L442): `ctx.GetPublisher().Publish(...)` still panics — now a nil-interface panic instead of a nil-receiver one. The guard helps callers that check `!= nil`; the comment shouldn't imply the panic goes away.

**Nit (tests):** `TestIsNilHandlesNonNillableKinds` doesn't cover nil map/slice/chan/func; nothing tests end-to-end that `App.Subscribe` refuses a typed-nil client (the crash described), and no app-level struct-valued pub/sub mock test was added despite the commit message mentioning `TestApp_SubscriberInitialize`.

aryanmehrotra:
Thanks @akshat-kumar-singhal — the root-cause item is taken, in `92e75bfcf`. @PiyushSingh-ZS, your `GetPublisher` point was already addressed at `0a07edb05` (guard plus `TestGetPublisherFiltersTypedNil`), along with the named `assertPlainNil` helper you asked for.

### Root cause: fixed at the constructors

You were right that filtering at the getters leaves the bad value in an **exported** field. `kafka.New` and `google.New` both return a bare nil of their concrete type on a rejected config, and assigning that into `pubsub.Client` is what manufactures the typed nil in the first place. Both now assign only a real client, as you suggested.

`TestRejectedPubSubConfigLeavesTheExportedFieldNil` pins it, and it is load-bearing — reverting just the google guard gives:

```
--- FAIL: TestRejectedPubSubConfigLeavesTheExportedFieldNil/google_with_no_project_id
    a rejected config must leave PubSub plainly nil, got a *google.googleClient in the interface
```

The assertion is a plain `!= nil`, not `assert.Nil`, for the same reason `assertPlainNil` exists: `assert.Nil` reflects and would pass on a typed nil, proving nothing.

### The GetPublisher comment no longer overclaims

Your reading is correct and the comment now says it outright: an unconditional `ctx.GetPublisher().Publish(...)` **still panics**, on a nil interface rather than a nil receiver. What the filter fixes is every caller that *does* check and whose guard the typed nil used to walk through. With the constructors fixed, the getters are defense in depth rather than the only line of it.

### isNil coverage

Extended to every nillable kind in both directions — map, slice, chan, func, interface and `unsafe.Pointer` alongside the pointer. The reason is in the comment: the `Kind` switch enumerates them, so a kind dropped from it falls through to "not nillable, therefore present", which is the wrong answer in the *safe-looking* direction and would not show up as a panic.

### Not done, and why

The end-to-end `App.Subscribe` test and the app-level struct-valued mock both live in `package gofr`, not `container`. They are worth having, but a cross-package test of container internals in this PR would widen it past the two bugs it exists to fix — happy to do it as a follow-up, or here if you would rather it not wait. The commit message reference to `TestApp_SubscriberInitialize` is stale; I will fix that wording.

### Verification

`gofmt`, `go vet`, `typos` and `go test` clean. `golangci-lint` reports **1 goconst, identical to the merge base** — pre-existing.

One honest note: a single `TestIntegration_ServerTimeout` failure appeared while I was running `./pkg/gofr/` and `./pkg/gofr/container/` in one invocation. It did not reproduce — 4/4 clean on the full package afterwards and 3/3 in isolation — and `development` passes the same run, so it is load-sensitivity in that timing test rather than anything here.

Umang01-hash (approved):
Deep-reviewed and verified end-to-end at head `a91b62c1`. Both bugs are real and the coupling is right (fix 1 alone makes the isNil panic reachable for value-receiver implementers).

Verified independently:
- **Not breaking**: isNil is unexported; GetPublisher/GetSubscriber/Health signatures unchanged (pre-existing methods); constructors internal.
- **Constructor guard sound**: kafka.New→*kafkaClient, google.New→*googleClient (concrete pointers), so `client != nil` catches the nil correctly.
- **isNil rewrite correct**: switches on Kind first, covering exactly the 7 nillable kinds; everything else is present.
- **6 fail-on-revert tests**: reverted both files — each fails reproducing the exact symptoms (health `"pubsub": DOWN "health check panicked"` → DEGRADED; `panic: reflect: call of reflect.Value.IsNil on struct Value`). Tests assert honestly (plain `== nil`, not reflective assert.Nil; health key ABSENT, not NotPanics).
- **Real app E2E**: with a rejected GOOGLE config, `development` crashes at startup (SIGSEGV in the errgroup goroutine, no recover); this PR starts clean ("subscriber not initialized"). Exactly the claimed process-death path.
- gofmt/vet/build/-race clean, golangci-lint 0 issues, coverage 9.0%→9.1%.

The deferred AddDBResolver `== nil` is correctly left to its own PR. LGTM.

PR #4166, opened 2026-09-08
perf(logging): cut two allocations from the request-log path

aryanmehrotra:
**Description:**

**10 → 8 allocations per logged request.** The bytes written are **identical** either way.

Measured through the Logging middleware with a real logger and a valid span context — GoFr installs a `NeverSample` provider when no exporter is configured, so the default deployment has one. Reverting either change individually puts it back to 9.

**1. Drop the `[]any` wrapper.** The middleware called `Log(entry)`, and `Log` takes `...any` — so every request allocated a **one-element slice** to carry a single value the logger then unwrapped again.

`LogEntry` / `ErrorEntry` take that value directly. They are found through an **optional interface**, not added to `logging.Logger`, so:
- an external implementation of `Logger` is **unaffected** and keeps using `Log`/`Error`
- `remotelogger` forwards them because it embeds the interface — nothing is promoted

**2. Encode both trace IDs into one buffer.** `TraceID.String()` and `SpanID.String()` are each a `hex.EncodeToString`, so each was its own allocation. A logged request **always needs both** — trace ID for `X-Correlation-ID`, both for the entry — so they share one buffer now.

> The span ID is formatted before the level gate rather than after. A request whose entry is discarded pays for 16 bytes it does not use; that costs no extra allocation and buys one on every request that *is* logged, which is the default.

**None.**

| | |
|---|---|
| Log output | **byte-identical** |
| `logging.Logger` interface | unchanged |
| External logger implementations | unaffected — they keep the `Log`/`Error` path |

**Additional Information:**

- No new dependencies.
- **`TestLoggingChainAllocationsDoNotRegress` pins 8 exactly, not as a ceiling.** Each saving is worth one allocation, so any tolerance wide enough to absorb toolchain drift is also wide enough to hide a regression.
- It uses a **real logger** deliberately: `logging.MockLogger` does not implement `LogEntry`, so a guard built on it takes the fallback path and **cannot see this change at all**.
- **`TestEntryLoggerFastPathIsTaken`** pins that the optional interface is actually used. Output is identical either way, so a dead type assertion would otherwise pass every test while silently forfeiting the saving.
- The allocation guard carries `//go:build !race` — the race detector adds allocations, so an exact count cannot hold under it.

**Checklist:**

### DIFF
--- a/pkg/gofr/http/middleware/entry_logger_dispatch_test.go
+++ b/pkg/gofr/http/middleware/entry_logger_dispatch_test.go
@@ -1,0 +1,77 @@
+package middleware
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// countingEntryLogger implements both the plain logger contract and the
+// entryLogger fast path, and records which one the middleware actually used.
+type countingEntryLogger struct {
+	logCalls, errorCalls           int
+	logEntryCalls, errorEntryCalls int
+}
+
+func (c *countingEntryLogger) Log(...any)     { c.logCalls++ }
+func (c *countingEntryLogger) Error(...any)   { c.errorCalls++ }
+func (c *countingEntryLogger) LogEntry(any)   { c.logEntryCalls++ }
+func (c *countingEntryLogger) ErrorEntry(any) { c.errorEntryCalls++ }
+
+// TestEntryLoggerFastPathIsTaken pins that the middleware really uses the
+// optional interface when the logger offers it.
+//
+// This is the one thing the rest of the suite cannot catch. LogEntry and Log
+// produce byte-identical output by design, so a type assertion that silently
+// stopped matching -- a renamed method, a pointer/value receiver mismatch, an
+// interface that drifted -- would keep every output test passing while quietly
+// giving back the allocation the fast path exists to save. Only counting which
+// method was called can tell the difference.
+func TestEntryLoggerFastPathIsTaken(t *testing.T) {
+	logger := &countingEntryLogger{}
+
+	handler := Logging(LogProbes{}, logger)(http.HandlerFunc(
+		func(w http.ResponseWriter, _ *http.Request) { w.WriteHeader(http.StatusOK) }))
+
+	handler.ServeHTTP(httptest.NewRecorder(), httptest.NewRequestWithContext(context.Background(), http.MethodGet, "/x", http.NoBody))
+
+	assert.Equal(t, 1, logger.logEntryCalls, "the middleware must use the LogEntry fast path")
+	assert.Zero(t, logger.logCalls, "the variadic Log path allocates the slice this avoids")
+}
+
+// TestErrorEntryFastPathIsTaken is the same guarantee for the 5xx path, which
+// takes a different branch and could regress independently.
+func TestErrorEntryFastPathIsTaken(t *testing.T) {
+	logger := &countingEntryLogger{}
+
+	handler := Logging(LogProbes{}, logger)(http.HandlerFunc(
+		func(w http.ResponseWriter, _ *http.Request) { w.WriteHeader(http.StatusInternalServerError) }))
+
+	handler.ServeHTTP(httptest.NewRecorder(), httptest.NewRequestWithContext(context.Background(), http.MethodGet, "/x", http.NoBody))
+
+	assert.Equal(t, 1, logger.errorEntryCalls, "a 5xx must use the ErrorEntry fast path")
+	assert.Zero(t, logger.errorCalls)
+}
+
+// plainOnlyLogger offers only the base contract, so the middleware must fall
+// back -- the compatibility half of the same guarantee.
+type plainOnlyLogger struct{ logCalls, errorCalls int }
+
+func (p *plainOnlyLogger) Log(...any)   { p.logCalls++ }
+func (p *plainOnlyLogger) Error(...any) { p.errorCalls++ }
+
+// TestFallbackForLoggerWithoutFastPath pins that an external logger that never
+// heard of LogEntry keeps working unchanged.
+func TestFallbackForLoggerWithoutFastPath(t *testing.T) {
+	logger := &plainOnlyLogger{}
+
+	handler := Logging(LogProbes{}, logger)(http.HandlerFunc(
+		func(w http.ResponseWriter, _ *http.Request) { w.WriteHeader(http.StatusOK) }))
+
+	handler.ServeHTTP(httptest.NewRecorder(), httptest.NewRequestWithContext(context.Background(), http.MethodGet, "/x", http.NoBody))
+
+	assert.Equal(t, 1, logger.logCalls, "a logger without the fast path must still be logged through")
+}
--- a/pkg/gofr/http/middleware/logger.go
+++ b/pkg/gofr/http/middleware/logger.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"bufio"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -158,6 +159,19 @@ type logEnabler interface {
 	LogEnabled() bool
 }
 
+// entryLogger is the optional fast-path interface for logging one already-built
+// entry.
+//
+// The plain logger contract is Log(...any), so Log(l) allocates a one-element
+// []any on every request purely to be unwrapped again inside the logger. A
+// logger that can take the entry directly avoids it. An implementation without
+// these methods falls back to Log/Error and behaves exactly as before, so no
+// external logger is affected.
+type entryLogger interface {
+	LogEntry(any)
+	ErrorEntry(any)
+}
+
 // Logging is a middleware which logs response status and time in milliseconds along with other data.
 //
 // The StatusResponseWriter wrapper allocated per request is pooled in a
@@ -188,24 +202,19 @@ func Logging(probes LogProbes, logger logger) func(inner http.Handler) http.Hand
 				pool.Put(srw)
 			}()
 
-			// Fetch SpanContext once and resolve trace/span IDs to strings only
-			// when they are valid. Under a noop tracer (the default after PR-1
-			// when no exporter is configured) the SpanContext is invalid and
-			// the IDs are all-zeros — calling .String() on those is wasted
-			// allocation. Substitute the precomputed zero-string constants so
-			// the log line and the X-Correlation-ID response header carry
-			// byte-identical values to the pre-PR-7 wire shape.
+			// Fetch the SpanContext once. GoFr installs an SDK provider with NeverSample
+			// when no exporter is configured, so the default deployment has a VALID span
+			// context and both IDs are real -- the zero-string constants below are for
+			// the genuinely absent case, not the common one.
 			sc := trace.SpanFromContext(r.Context()).SpanContext()
 
-			// Only the trace ID is needed before the handler runs, for the
-			// X-Correlation-ID response header. The span ID is used solely
-			// inside the log entry, so it is resolved in handleRequestLog --
-			// after the level gate -- and costs nothing on a request whose
-			// entry is discarded.
-			traceID := zeroTraceID
-			if sc.IsValid() {
-				traceID = sc.TraceID().String()
-			}
+			// Both IDs are resolved here, in one allocation. The trace ID is
+			// needed before the handler runs, for the X-Correlation-ID response
+			// header; the span ID is only used inside the log entry, so a request
+			// whose entry the level gate discards pays for 16 bytes it does not
+			// use. That costs no extra allocation -- one buffer covers both -- and
+			// saves one on every request that IS logged, which is the default.
+			traceID, spanID := traceSpanIDs(sc)
 
 			// Assigned rather than Set: Header.Set canonicalizes its key on
 			// every call, and "X-Correlation-ID" is not already in canonical
@@ -225,15 +234,15 @@ func Logging(probes LogProbes, logger logger) func(inner http.Handler) http.Hand
 			}
 
 			start := time.Now()
-			defer handleRequestLog(srw, r, start, traceID, sc, logger)
+			defer handleRequestLog(srw, r, start, traceID, spanID, logger)
 
 			inner.ServeHTTP(srw, r)
 		})
 	}
 }
 
-func handleRequestLog(srw *StatusResponseWriter, r *http.Request, start time.Time, traceID string,
-	sc trace.SpanContext, logger logger) {
+func handleRequestLog(srw *StatusResponseWriter, r *http.Request, start time.Time,
+	traceID, spanID string, logger logger) {
 	status := srw.Status()
 
 	// A server error is reported through Error, which survives every level below
@@ -243,11 +252,6 @@ func handleRequestLog(srw *StatusResponseWriter, r *http.Request, start time.Tim
 		if e, ok := logger.(logEnabler); ok && !e.LogEnabled() {
 			return
 		}
-	}
-
-	spanID := zeroSpanID
-	if sc.IsValid() {
-		spanID = sc.SpanID().String()
 	}
 
 	l := &RequestLog{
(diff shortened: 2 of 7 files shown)


PiyushSingh-ZS (commented):
Smallest-risk of the three perf PRs, and I think it is in good shape. Both savings are real and the compatibility story is the right one — reaching `LogEntry`/`ErrorEntry` through an optional interface rather than widening `logging.Logger` is what keeps every external implementation working, and that matters more here than the two allocations.

### The "byte-identical" claim holds — I checked the part that could have broken it

The variadic path does something `logEntry` skips:

```go
traceID, filteredArgs := extractTraceIDAndFilterArgs(args)
entry.TraceID = traceID
```

If a `*RequestLog` could trigger that, the new path would silently drop the top-level `trace_id` field from every request log, which would be a real wire change hiding behind an allocation PR.

It cannot. `extractTraceIDAndFilterArgs` short-circuits on `hasTraceMarker(args)`, which only matches `traceIDMarker` and a `map[string]any` carrying the marker key. `*RequestLog` is neither, so `TraceID` was already `""` and `omitempty` already dropped the field. **Byte-identical confirmed for the type this is actually used with.**

Also confirmed `zeroTraceID` is 32 chars and `zeroSpanID` is 16, so the `hex.Encode` offsets in `traceSpanIDs` are correct.

### `LogEntry` diverges from `Log` for inputs other than `*RequestLog`

`LogEntry(any)` is an exported method on `*logger`, so anyone who type-asserts can reach it. For a `traceIDMarker`, or a marker-carrying `map[string]any`, it produces *different* output from `Log` — no trace-ID extraction, no arg filtering. Today the only caller passes a `*RequestLog`, so it is latent rather than a bug.

Worth closing off with a doc line on `LogEntry` — something like "takes a single pre-built entry; does not perform the trace-marker extraction `Log` does" — so the next caller knows the contract is narrower than the signature suggests. Routing it through the same extraction would also work but costs back part of what the PR is buying.

### Nits

**1. Hoist the type assertion out of the request path.**

`handleRequestLog` does `logger.(entryLogger)` on every request. `Logging(probes, logger)` already runs once and is the natural place to resolve it — which is what the sibling `newHistogramRecorder` does for `metricsOptRecorder`/`metricsAttrer`, and what `remotelogger.New` does for both `logEnabler` and (in this PR) `entryLogger`. Three places in the codebase resolve optional interfaces at construction; this is the one that does not. The assertion is allocation-free so the win is negligible — the consistency is the argument.

**2. `hex.Encode(b[:len(zeroTraceID)], tid[:])` couples to the constants.**

Correct today, but it is a silent slice-bounds panic if either constant is ever edited. `hex.EncodedLen(len(tid))` expresses the same offsets without the coupling.

**3.** The honesty in the `encodeJSON` comment — that taking `*logEntry` is a readability choice with *zero* allocation benefit, measured 4 and 6 before and after — is good. Easy to quietly claim a saving there; noting that there isn't one is more useful.

The `remoteLogger.LogEntry` reasoning (safe with respect to the dynamic level because the type pushes changes down via `ChangeLevel` rather than gating on `currentLevel` itself) is correct and worth keeping in the comment.

Umang01-hash (dismissed):
Verified locally at head `08036e7d` in an isolated worktree — build, vet, golangci-lint (`--new-from-rev`, 0 issues), full suites for all 3 touched packages, and the new tests all pass.

**Sound:**
- No exported-API break — `logging.Logger` is unchanged; `LogEntry`/`ErrorEntry` sit on the unexported `entryLogger` optional interface, reached like `logEnabler`, so external loggers are unaffected.
- Byte-identical output: `hasTraceMarker(*RequestLog)` is false, so the old `Log` path also left `trace_id` empty — `logEntry` matches. Parity tests confirm.
- `traceSpanIDs`: 32+16 hex chars into one buffer, sliced — byte-identical to two `.String()` calls, one alloc instead of two.
- Perf win proven, not asserted: alloc guard = 8; reverting **either** optimization → 9 (checked both locally), and the dispatch test flips too. The win is real in the default deployment — `otel.go` installs a NeverSample SDK provider with valid IDs, so both IDs really are formatted every request.
- Level/ChangeLevel gating preserved; new `entries` field is set-once-before-goroutine (safe).

**Non-blocking:** `-race` on `remotelogger` fails on `TestRemoteLogger_UpdateLevel`, but the same race is present on `development` — pre-existing, not introduced here.

LGTM.

aryanmehrotra:
Thanks both — all three points addressed at `769c5bc3a`.

**Assertions hoisted out of the request path.** `Logging()` now builds a `logSink` once, holding the logger plus whichever of `logEnabler`/`entryLogger` it turned out to satisfy, and `handleRequestLog` reads the resolved fields. You're right that this is a consistency argument rather than a speed one — the assertions are allocation-free. It's the reason I added `TestLogSinkResolvesOptionalInterfacesOnce`: the dispatch tests prove the fast path is *reached*, not *where it was selected*, so they'd stay green if the assertion moved back into the request path.

**`hex.Encode` no longer restates the constants.** It returns what it wrote, so the split offset comes from the encode itself:

```go
n := hex.Encode(b[:], tid[:])
hex.Encode(b[n:], sid[:])
return s[:n], s[n:]
```

The buffer is still sized from `zeroTraceID`/`zeroSpanID`, because those *are* the wire widths this has to reproduce — but editing one can no longer move a slice bound without moving the buffer with it.

**`LogEntry`'s contract is documented.** Spelled out that it takes one pre-built entry and does *not* do the trace-marker extraction or argument filtering `Log` does, so a `traceIDMarker` or a marker-carrying map logs differently through it. Kept it out of the extraction path for the reason you gave — that's most of what the PR buys.

No change on nit 3; agreed, and the `encodeJSON` comment stays as-is.

aryanmehrotra:
@Umang01-hash re-requesting — your approval was dismissed by the review-fix push at `ec9a9bed2`, and head has moved on twice since. Summary of what changed after that approval:

- **Merged `development`** (clean, no conflicts) — clears the `BEHIND` state.
- **The PR body now carries the measured numbers.** The "two allocations" claim is verified end-to-end rather than asserted: `9 → 7 allocs/op`, `376 → 360 B/op` with a valid span context, n=20 per side, interleaved branch/development rounds, `benchstat`-compared, `± 0%` on every allocation figure. Both savings are attributed individually, and a third benchmark with a logger that hides the optional interfaces shows only one of the two — which is what rules out the fast path silently no-opping. ns/op is deliberately not quoted; the note in the body says why.
- **`TestTraceSpanIDsMatchesOtelRendering`** — the new `traceSpanIDs` renders both IDs into one stack array and returns two slices of one string, so the split offset can move without any existing assertion noticing. This compares each field against otel's own `TraceID.String()` / `SpanID.String()` and covers a zero trace ID, a zero span ID and an all-`0xff` pair.
- **`TestSharedLogSinkIsSafeUnderConcurrentRequests`** — the sink is now resolved once and shared by every request, and the existing concurrency test only exercises the `Log`/`Error` fallback. This drives the fast-path branches from 64 goroutines under `-race`.
- **`newLogSink`'s partial shapes** are pinned separately. Only the all-or-nothing cases were asserted, which would pass even if the two independent assertions were collapsed into one.
- **`remotelogger.LogEntry` / `ErrorEntry` were at 0% coverage** and are the pair production actually runs, since `container.go` always wires the request logger through `remotelogger.New`. They now have parity tests against `Log`/`Error`, a construction-time resolution test, the fallback for an embedded logger without the fast path, and a `ChangeLevel` suppression test.
- **`logger.logEntry`'s terminal branch** was uncovered — every parity test ran against a `bytes.Buffer`. It now has the same byte-for-byte parity assertion.

Coverage: `http/middleware` 95.5% → 95.6%, `logging` 88.4% → 89.3%, `logging/remotelogger` 81.7% → 83.1%. Every added test was mutation-checked — reverting the optimization it guards makes it fail.

No production code changed since your approval; the two new commits are tests only.

aryanmehrotra:
Correction to the last line of the comment above — I was wrong, and it matters for what you have to re-read.

Your approval was recorded against `ec9a9bed2`. Production code **has** changed since then, in `cfe1d9be3` (`perf(logging): resolve the optional logger interfaces once, at middleware build`): `pkg/gofr/http/middleware/logger.go` +66/−19 and `pkg/gofr/logging/logger.go` +8. That is the `logSink` hoist — the `logEnabler` and `entryLogger` assertions moved out of the per-request path and into `Logging()`, and `handleRequestLog` now takes the resolved sink instead of the bare logger.

What I should have said is that **my** two commits on top are tests only. The production delta you have not seen is `ec9a9bed2..cfe1d9be3`, in `pkg/gofr/http/middleware/logger.go` and `pkg/gofr/logging/logger.go`.

Everything else in the previous comment stands.

aryanmehrotra:
One more correction, this time to the PR itself rather than to a comment.

The `traceSpanIDs` doc said otel's `TraceID.String()` and `SpanID.String()` are each a `hex.EncodeToString`. At the version this module actually pins — `go.opentelemetry.io/otel/trace v1.46.0` — they are not. Both build the hex into a stack array from a nibble lookup table and convert it once with `string(...)`:

- `TraceID.String()` → `trace.go:60`, via `hexBytes()` at `trace.go:66`
- `SpanID.String()` → `trace.go:111`, via `hexBytes()` at `trace.go:116`

The measured saving is unchanged — each call still costs exactly one string allocation, so two became one — but the stated reason invited the wrong conclusion, that otel was doing something wasteful per call. It is not. The waste was that a logged request needs two results and paid for two allocations to get them, which is what the shared buffer fixes. Fixed in `78eaf29fc` (comment only) and in the PR body.

Also worth stating explicitly, since it is the half-zero case a reviewer would ask about: `SpanContext.IsValid()` is `HasTraceID() && HasSpanID()` (`trace.go:267`), so a context with a real trace ID and a zero span ID — or the reverse — is invalid and BOTH IDs fall back to the zero-string constants together. That is unchanged from `development`, and `TestTraceSpanIDsMatchesOtelRendering` now pins it so it cannot drift into emitting a half-real pair.

akshat-kumar-singhal (approved):
Approve with nits. Output is byte-identical: `logEntry` builds exactly what `logf` does for one arg with an empty format, `*RequestLog` was never touched by `extractTraceIDAndFilterArgs`, and the JSON and pretty paths are unchanged. No pooled buffers; the trace/span buffer is a stack array copied by `string()`. `remoteLogger` level changes still reach the embedded logger. A full-handler benchmark to `io.Discard` confirms the two allocations: valid span 10 → 8 allocs (400 → 384 B), invalid span 8 → 7 (352 → 336 B).

**Should-fix: default mux router mode gets +48 B/request** — [`logger.go:223`](https://github.com/gofr-dev/gofr/blob/78eaf29fc7a1e94baf6028ee1577f64d7df156ae/pkg/gofr/http/middleware/logger.go#L223), captured at [`:274`](https://github.com/gofr-dev/gofr/blob/78eaf29fc7a1e94baf6028ee1577f64d7df156ae/pkg/gofr/http/middleware/logger.go#L274)
The default mux matcher in `router.go` rebuilds the middleware chain per request, so the `Logging(...)(inner)` closure is allocated per request and now captures a 48-byte `logSink` by value instead of a 16-byte interface. `BenchmarkLogging` (mux): 1248 → 1296 B/op, 10 allocs/op both; pprof puts the growth in `Logging.func2`. Holding it behind a pointer (`sinkp := new(logSink); *sinkp = newLogSink(logger)`, pass `*sinkp` to the deferred call) gives 1264 B/op with alloc counts unchanged.

**Nit:** `logging_alloc_guard_test.go` pins exactly 8 allocs — deliberate, but any Go/otel/gorilla bump will trip it. (Checked it also holds under `-covermode=atomic`.)

**Nit:** `perf_parity_test.go` never passes a `*middleware.RequestLog` (the value production logs, which has `PrettyPrint`); the terminal test only uses a string. Shared code path, so low risk.

**Nit:** the span ID is now formatted before the level gate: +16 B, no extra alloc, for a discarded request log (`Remote_Discarded`: 56 → 72 B/op). Documented in the comment; fine.

Note: `go test -race` on `logging/remotelogger` fails four existing tests (`TestRemoteLogger_UpdateLevel` etc.) with a race on `currentLevel` — same on base, not this PR. The 5 new remotelogger tests pass `-race -count=3`.

aryanmehrotra:
Thanks @akshat-kumar-singhal — the mux regression was real, reproduced exactly, and is now **gone entirely** rather than reduced. Fixed in `856e94445`.

I measured 1297 B/op on the branch against your 1296, and 1249 at the merge base against your 1248 — a consistent one-byte offset, so we are looking at the same thing.

### It goes further than the suggested fix, and the second half is why

Your `sinkp := new(logSink)` takes the closure's capture from 48 bytes to 8, which is most of it. But the closure was *also* still capturing `logger` separately, purely for `panicRecovery(recover(), srw, logger)` — a 16-byte interface holding the same value `sink.logger` already had.

Passing `sink.logger` there drops that capture too, and that is what closes the last 16 bytes:

| build | B/op | allocs/op |
|---|---|---|
| base `e95c11759` | **1249** | 10 |
| this PR at `78eaf29fc` | 1297 (+48) | 10 |
| + pointer sink (your fix) | 1265 (+16) | 10 |
| **+ `panicRecovery` via `sink.logger`** | **1249 (baseline)** | 10 |

`BenchmarkLogging`, `-benchtime=2000x -count=6`, darwin/arm64. B/op was identical across all six runs in every configuration, so the numbers are stable rather than eyeballed from one run. `allocs/op` never moves.

I measured the baseline by building a worktree at the merge base and running the benchmark there, rather than inferring it from your figure — `BenchmarkLogging` already exists at `e95c11759`, so it is a like-for-like comparison.

The mechanism is in `router.go:172-175` as you said: mux builds its chain inside `Match`, so the `Logging(...)(inner)` closure is per request. The trie matcher memoises per route, which is why none of this shows up there. That is all written into the `newLogSink` doc comment so the pointer does not get "simplified" back to a value later.

### On the other notes

- **`-race` in `logging/remotelogger`** — pre-existing, and I checked rather than assumed: the merge base fails the same tests, and the base itself varies between **3 and 4 failures across 5 runs**, so `TestLogLevelChangeToFatal_NoExit` joining the list is that flake and not this change.
- **`golangci-lint`** — 11 issues at base, 5 on this branch. The PR removes 6 and adds none.
- The alloc-guard pinning exactly 8, the `perf_parity_test.go` gap and the span-ID-before-the-gate note are all fair; I have left them as they are since each is a deliberate trade you already read the reasoning for, but say the word if you want the guard loosened to a range.

Umang01-hash (approved):
Verified locally at the PR head.

- No breaking change: the exported `logging.Logger` interface is unchanged. `LogEntry`/`ErrorEntry` are additive methods on the concrete `*logger`/`*remoteLogger`, reached through the unexported optional `entryLogger` assertion (exactly like `LogEnabled`), so external Logger implementations fall back to `Log`/`Error` untouched.
- Output is byte-identical: `TestLogEntryOutputMatchesLog`/`ErrorEntry`/`OnTerminal`/`RespectsLevel`/`FollowsChangeLevel` all pass. The `LogEntry` contract (no trace-marker extraction) is documented and safe because the request logger passes a plain `*RequestLog`, never a marker.
- The saving is real and guarded: `traceSpanIDs` renders both IDs into one [48]byte buffer (widths derived from the zero-string constants + `hex.Encode`'s return) = 1 alloc instead of 2; `LogEntry` drops the one-element `[]any`. `TestLoggingChainAllocationsDoNotRegress` pins exactly 8 allocs — reverting `traceSpanIDs` to two `.String()` calls fails it. `BenchmarkLogging` is flat at 1249 B/op / 10 allocs vs development, so the pointer-captured `logSink` doesn't regress the per-request closure.
- gofmt/vet/build clean; golangci-lint 0 new issues; middleware + logging pass `-race`.

FYI (pre-existing, not this PR): `go test -race ./pkg/gofr/logging/remotelogger/...` fails on `development` too — the gomock `MockLogger.ChangeLevel` isn't goroutine-safe under the existing concurrent tests (`TestRemoteLogger_ConcurrentLevelAccess` etc.). None of this PR's new tests are involved. Worth a separate fix (a mutex in the mock or a real logger in those tests).

LGTM.

PR #4167, opened 2026-09-08
perf(deps): let a build omit the datasource drivers and GraphQL engine it does not use

aryanmehrotra:
**Description:**

Three opt-in build tags. **A build that sets none of them is byte-for-byte what it is today.**

| tag | omits |
|---|---|
| `gofr_nosqldrivers` | the blank-imported SQL drivers |
| `gofr_nopubsub` | the Kafka / Google / MQTT client implementations |
| `gofr_nographql` | graphql-go and gqlparser |

Measured on a minimal GoFr service (`gofr.New`, one route, `Run`):

| build | binary | packages |
|---|---|---|
| default | 57.10 MB | 831 |
| `gofr_nosqldrivers` | 50.96 MB | 787 |
| `gofr_nopubsub` | 48.23 MB | 618 |
| `gofr_nographql` | 56.32 MB | 811 |
| **all three** | **41.26 MB** | **554** |

**−15.84 MB (−27.7%), −277 packages.** Idle RSS drops too, driven by the drivers.

**Why the drivers cost so much.** `modernc.org/sqlite` is blank-imported for driver registration and brings `modernc.org/libc`, whose `netdb` init parses embedded copies of `/etc/protocols` and `/etc/services` into **permanent Go structs** — about **1.7 MB of retained heap**, roughly half the process's fixed heap floor. None of it is reachable from the HTTP path. The three pub/sub clients bring **213 packages** between them, most of it Google Pub/Sub's gRPC and auth stack.

**The mechanism, and the part that is easy to get wrong.** Making a subsystem optional takes **two** things: an **interface** on the consumer naming only the methods it calls, *and* a **build tag on the concrete file**.

⚠️ **The interface alone does nothing.** Extracting `graphQLRunner` left all the graphql packages linked until `graphql.go` itself carried the tag — the concrete file still compiles the library in.

**None.** Every tag is opt-in and the default build is unchanged.

`GraphQLQuery` and `GraphQLMutation` name only GoFr's own `Handler`, so **the exported surface is identical in both builds**.

Verified against `development`: **identical `go doc -all` across 6 packages, and an identical linked-package list.**

A build that *does* set `gofr_nographql` still compiles a user's GraphQL calls. Registering a resolver **logs an error naming the tag** rather than failing silently, and `enabled()` keeps App from routing `/graphql` and the playground against a handler this build cannot provide.

⚠️ Without that guard `setupGraphQL` registers a **nil handler** and the first `POST /graphql` panics. `TestGraphQLDisabled_SetupRegistersNoRoute` fails if the guard is removed.

**Additional Information:**

- No new dependencies — this PR only removes linkage.
- A new **Slim Build Tags 🪶** CI job builds and tests every tag combination. Half the code these tags add is compiled *only* when the tags are set; without that job nothing would ever build it.
- `noopResponder` moves from `responder.go` into `graphql.go`, its only user. It cannot stay: `responder.go` also holds the exported `Responder` interface so it cannot carry the tag, and leaving it there makes it dead code in a tagged build — which `golangci-lint` reports as `unused`, failing the lint gate for anyone who sets the tag.
- `configTrue` is declared inside `pubsub_backends.go` for the same reason: an untagged home would leave it unused under `gofr_nopubsub`.

**Checklist:**

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -512,6 +512,44 @@ jobs:
           QLTY_TOKEN: ${{ secrets.QLTY_TOKEN }}
 
   # Job for code quality checks
+
+  slim_build_tags:
+    name: Slim Build Tags🪶
+    runs-on: ubuntu-latest
+    # Only builds and runs the packages the tags touch, so it stays well inside
+    # the time the rest of the matrix takes.
+    timeout-minutes: 15
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v7
+
+      - name: Set up Go environment
+        uses: actions/setup-go@v7
+        with:
+          go-version: '1.26'
+
+      # Half of the code the build tags add is compiled only when the tags are
+      # set, so without this job nothing ever builds it. A later edit to
+      # container.go that referenced kafka. or google. directly would compile and
+      # pass every other job while silently breaking the tagged build.
+      - name: Build and test with datasources omitted
+        run: |
+          set -euo pipefail
+          for tags in "gofr_nosqldrivers" "gofr_nopubsub" "gofr_nosqldrivers gofr_nopubsub"; do
+            echo "::group::-tags '$tags'"
+            go build -tags "$tags" ./...
+            go vet   -tags "$tags" ./pkg/gofr/container/ ./pkg/gofr/datasource/sql/
+            go test  -tags "$tags" -count=1 ./pkg/gofr/container/ ./pkg/gofr/datasource/sql/
+            echo "::endgroup::"
+          done
+
+      # The tags are opt-in, so the default build must stay exactly as it was.
+      # This is the assertion that keeps that true.
+      - name: Default build links the same packages as before
+        run: |
+          set -euo pipefail
+          go build ./...
+          go test -count=1 ./pkg/gofr/container/ ./pkg/gofr/datasource/sql/
   code_quality:
     name: Code Quality🎖️
     runs-on: ubuntu-latest
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -532,16 +532,19 @@ jobs:
       # set, so without this job nothing ever builds it. A later edit to
       # container.go that referenced kafka. or google. directly would compile and
       # pass every other job while silently breaking the tagged build.
-      - name: Build and test with datasources omitted
+      - name: Build and test with optional subsystems omitted
         run: |
           set -euo pipefail
-          for tags in "gofr_nosqldrivers" "gofr_nopubsub" "gofr_nosqldrivers gofr_nopubsub"; do
+          for tags in "gofr_nosqldrivers" "gofr_nopubsub" "gofr_nographql" "gofr_nosqldrivers gofr_nopubsub gofr_nographql"; do
             echo "::group::-tags '$tags'"
             go build -tags "$tags" ./...
-            go vet   -tags "$tags" ./pkg/gofr/container/ ./pkg/gofr/datasource/sql/
+            go vet   -tags "$tags" ./pkg/gofr/ ./pkg/gofr/container/ ./pkg/gofr/datasource/sql/
             go test  -tags "$tags" -count=1 ./pkg/gofr/container/ ./pkg/gofr/datasource/sql/
             echo "::endgroup::"
           done
+
+          # The GraphQL stubs have their own tests, which only exist under the tag.
+          go test -tags gofr_nographql -count=1 -run TestGraphQLDisabled ./pkg/gofr/
 
       # The tags are opt-in, so the default build must stay exactly as it was.
       # This is the assertion that keeps that true.
--- a/pkg/gofr/container/pubsub_backends_disabled.go
+++ b/pkg/gofr/container/pubsub_backends_disabled.go
@@ -1,0 +1,38 @@
+//go:build gofr_nopubsub
+
+// Stubs for a build made with -tags gofr_nopubsub.
+//
+// Leaving Container.PubSub nil is an already-supported state -- it is what an
+// unset PUBSUB_BACKEND produces, and Close guards it with isNil -- so a service
+// that does not use pub/sub is unaffected. A service that DOES configure a
+// backend gets a logged error naming the tag, rather than silence, because the
+// alternative is a publisher that quietly never publishes.
+
+package container
+
+import (
+	"gofr.dev/pkg/gofr/config"
+	"gofr.dev/pkg/gofr/datasource/pubsub"
+)
+
+// pubsubBackendsLinked is false here: see the comment on the constant in
+// pubsub_backends.go.
+const pubsubBackendsLinked = false
+
+// pubsubDisabledMsg names the tag so the fix is obvious from one log line.
+const pubsubDisabledMsg = "PUBSUB_BACKEND=%s was configured, but this binary was " +
+	"built with -tags gofr_nopubsub, which omits the %s client. Rebuild without the tag to use it."
+
+func (c *Container) createMqttPubSub(config.Config) pubsub.Client {
+	c.Logger.Errorf(pubsubDisabledMsg, "MQTT", "MQTT")
+
+	return nil
+}
+
+func (c *Container) createKafkaPubSub(config.Config) {
+	c.Logger.Errorf(pubsubDisabledMsg, "KAFKA", "Kafka")
+}
+
+func (c *Container) createGooglePubSub(config.Config) {
+	c.Logger.Errorf(pubsubDisabledMsg, "GOOGLE", "Google Pub/Sub")
+}
--- a/pkg/gofr/datasource/sql/drivers_disabled.go
+++ b/pkg/gofr/datasource/sql/drivers_disabled.go
@@ -1,0 +1,15 @@
+//go:build gofr_nosqldrivers
+
+// No driver registration in a build made with -tags gofr_nosqldrivers.
+//
+// DB_DIALECT=postgres or =sqlite then fails at startup: NewSQL's registerOtel
+// call reports database/sql's own "unknown driver" error naming the dialect, and
+// returns a nil DB. That is an already-supported state -- it is what an
+// unconfigured database produces, and the container guards it with isNil -- so
+// the service still starts, with one error line saying exactly what is missing.
+// A loud, immediate complaint is the point of doing this with a tag rather than
+// by asking users to blank-import drivers themselves, where the same mistake is
+// silent. A user who wants one of these dialects in a tagged build imports
+// the driver in their own main package, exactly as with database/sql directly.
+
+package sql
--- a/pkg/gofr/container/container.go
+++ b/pkg/gofr/container/container.go
@@ -28,9 +28,6 @@ import (
 	"gofr.dev/pkg/gofr/datasource"
 	"gofr.dev/pkg/gofr/datasource/file"
 	"gofr.dev/pkg/gofr/datasource/pubsub"
-	"gofr.dev/pkg/gofr/datasource/pubsub/google"
-	"gofr.dev/pkg/gofr/datasource/pubsub/kafka"
-	"gofr.dev/pkg/gofr/datasource/pubsub/mqtt"
 	"gofr.dev/pkg/gofr/datasource/redis"
 	"gofr.dev/pkg/gofr/datasource/sql"
 	"gofr.dev/pkg/gofr/logging"
@@ -209,51 +206,6 @@ func (c *Container) Close() error {
 	}
 
 	return err
-}
-
-func (c *Container) createMqttPubSub(conf config.Config) pubsub.Client {
-	var qos byte
-
-	port, err := strconv.Atoi(conf.GetOrDefault("MQTT_PORT", "0"))
-	if err != nil {
-		c.Logger.Error("Invalid value for MQTT_PORT, using default: 0")
-	}
-
-	order, _ := strconv.ParseBool(conf.GetOrDefault("MQTT_MESSAGE_ORDER", "false"))
-
-	retrieveRetained, _ := strconv.ParseBool(conf.GetOrDefault("MQTT_RETRIEVE_RETAINED", "false"))
-
-	keepAlive, err := time.ParseDuration(conf.Get("MQTT_KEEP_ALIVE"))
-	if err != nil {
-		keepAlive = 30 * time.Second
-
-		c.Logger.Debug("MQTT_KEEP_ALIVE is not set or invalid, setting it to 30 seconds")
-	}
-
-	switch conf.Get("MQTT_QOS") {
-	case "1":
-		qos = 1
-	case "2":
-		qos = 2
-	default:
-		qos = 0
-	}
-
-	configs := &mqtt.Config{
-		Protocol:         conf.GetOrDefault("MQTT_PROTOCOL", "tcp"), // using tcp as default method to connect to broker
-		Hostname:         conf.Get("MQTT_HOST"),
-		Port:             port,
-		Username:         conf.Get("MQTT_USER"),
-		Password:         conf.Get("MQTT_PASSWORD"),
-		ClientID:         conf.Get("MQTT_CLIENT_ID_SUFFIX"),
-		QoS:              qos,
-		Order:            order,
-		RetrieveRetained: retrieveRetained,
-		KeepAlive:        keepAlive,
-		CloseTimeout:     0 * time.Millisecond,
-	}
-
-	return mqtt.New(configs, c.Logger, c.metricsManager)
 }
 
 // GetHTTPService returns registered HTTP services.
@@ -494,75 +446,6 @@ func getDefaultDatasourceMicrosecondBuckets() []float64 {
 		50, 75, 100, 125, 150, 200, 300, 500, 750, 1000, 2000, 3000, 5000, 7500, 10000, // 50µs-10ms
 		25000, 50000, 100000, 250000, 500000, 1000000, 5000000, 10000000, 30000000, 60000000, 120000000, 180000000, // 25ms-3min
 	}
-}
-
-func (c *Container) createKafkaPubSub(conf config.Config) {
-	if conf.Get("PUBSUB_BROKER") == "" {
-		return
-	}
-
-	partition, err := strconv.Atoi(conf.GetOrDefault("PARTITION_SIZE", "0"))
-	if err != nil {
-		c.Logger.Error("Invalid value for PARTITION_SIZE, using default: 0")
-	}
-
-	// PUBSUB_OFFSET determines the starting position for message consumption in Kafka.
-	// This allows control over whether to read historical messages or only new ones:
-	// - Default value -1: Start from the latest offset (only consume new messages after consumer starts)
-	// - Value 0: Start from the earliest offset (read all historical messages from the beginning)
-	// - Positive value: Start from a specific offset position (useful for resuming from a known point)
-	// This is particularly important for scenarios like message replay, recovery from failures,
-	// or when you only want to process messages that arrive after the consumer is initialized.
-	offSet, err := strconv.Atoi(conf.GetOrDefault("PUBSUB_OFFSET", "-1"))
-	if err != nil {
-		c.Logger.Error("Invalid value for PUBSUB_OFFSET, using default: -1")
-	}
-
-	batchSize, err := strconv.Atoi(conf.GetOrDefault("KAFKA_BATCH_SIZE", strconv.Itoa(kafka.DefaultBatchSize)))
-	if err != nil {
-		c.Logger.Errorf("Invalid value for KAFKA_BATCH_SIZE, using default: %d", kafka.DefaultBatchSize)
-	}
-
-	batchBytes, err := strconv.Atoi(conf.GetOrDefault("KAFKA_BATCH_BYTES", strconv.Itoa(kafka.DefaultBatchBytes)))
-	if err != nil {
-		c.Logger.Errorf("Invalid value for KAFKA_BATCH_BYTES, using default: %d", kafka.DefaultBatchBytes)
-	}
-
-	batchTimeout, err := strconv.Atoi(conf.GetOrDefault("KAFKA_BATCH_TIMEOUT", strconv.Itoa(kafka.DefaultBatchTimeout)))
-	if err != nil {
-		c.Logger.Errorf("Invalid value for KAFKA_BATCH_TIMEOUT, using default: %d", kafka.DefaultBatchTimeout)
-	}
-
-	tlsConf := kafka.TLSConfig{
-		CertFile:           conf.Get("KAFKA_TLS_CERT_FILE"),
-		KeyFile:            conf.Get("KAFKA_TLS_KEY_FILE"),
-		CACertFile:         conf.Get("KAFKA_TLS_CA_CERT_FILE"),
-		InsecureSkipVerify: conf.Get("KAFKA_TLS_INSECURE_SKIP_VERIFY") == "true",
-	}
-
-	pubsubBrokers := strings.Split(conf.Get("PUBSUB_BROKER"), ",")
-
-	c.PubSub = kafka.New(&kafka.Config{
-		Brokers:          pubsubBrokers,
-		Partition:        partition,
-		ConsumerGroupID:  conf.Get("CONSUMER_ID"),
-		OffSet:           offSet,
-		BatchSize:        batchSize,
-		BatchBytes:       batchBytes,
-		BatchTimeout:     batchTimeout,
-		SecurityProtocol: conf.Get("KAFKA_SECURITY_PROTOCOL"),
-		SASLMechanism:    conf.Get("KAFKA_SASL_MECHANISM"),
-		SASLUser:         conf.Get("KAFKA_SASL_USERNAME"),
-		SASLPassword:     conf.Get("KAFKA_SASL_PASSWORD"),
-		TLS:              tlsConf,
-	}, c.Logger, c.metricsManager)
-}
-
-func (c *Container) createGooglePubSub(conf config.Config) {
-	c.PubSub = google.New(google.Config{
-		ProjectID:        conf.Get("GOOGLE_PROJECT_ID"),
-		SubscriptionName: conf.Get("GOOGLE_SUBSCRIPTION_NAME"),
-	}, c.Logger, c.metricsManager)
 }
 
 func (c *Container) createRedisPubSub(conf config.Config) {
--- a/pkg/gofr/container/container_test.go
+++ b/pkg/gofr/container/container_test.go
@@ -83,6 +83,10 @@ func Test_newContainerPubSubInitializationFail(t *testing.T) {
 }
 
 func TestContainer_MQTTInitialization_Default(t *testing.T) {
+	if !pubsubBackendsLinked {
+		t.Skip("built with -tags gofr_nopubsub; the MQTT client is not linked")
+	}
+
 	configs := map[string]string{
 		"PUBSUB_BACKEND": "MQTT",
 	}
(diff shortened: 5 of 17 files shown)


akshat-kumar-singhal:
A downstream data point for `gofr_nopubsub`, on a different axis from binary size.

We run a GoFr service that uses no pub/sub at all. Because `pkg/gofr/container/container.go` imports `gofr.dev/pkg/gofr/datasource/pubsub/google` unconditionally, `go mod why` gives us this on v1.60.1:

```
our-service/cmd/...
 → gofr.dev/pkg/gofr
 → gofr.dev/pkg/gofr/container
 → gofr.dev/pkg/gofr/datasource/pubsub/google
 → cloud.google.com/go/pubsub
 → google.golang.org/api/transport/http
 → google.golang.org/api/internal/cert
 → github.com/googleapis/enterprise-certificate-proxy
```

Last week that last one broke a deploy for us. Google moved the `enterprise-certificate-proxy` `v0.3.19` tag after publication:

```console
$ curl -s https://proxy.golang.org/github.com/googleapis/enterprise-certificate-proxy/@v/v0.3.19.mod | grep toolchain
toolchain go1.25.8

$ curl -s https://raw.githubusercontent.com/googleapis/enterprise-certificate-proxy/v0.3.19/go.mod | grep toolchain
toolchain go1.26.5
```

Our `go.sum` matched `sum.golang.org` throughout, and the proxy-based build that actually ships was correct the whole time — the checksum database is append-only, so only a build resolving from VCS could see it. But we keep a `GOPROXY=direct` resilience build, and that job failed `go mod download` on the mismatch and could never pass again at that version. We pinned forward and moved on.

**None of that is GoFr's doing.** The point is only where we happened to be standing when it landed: we absorbed it for a module we never import, reached solely through a pub/sub client we don't use, feeding a client-certificate feature we have never enabled — nothing in our tree, infra or workflows sets `GOOGLE_API_USE_CLIENT_CERTIFICATE` or ships a `certificate_config.json`, so the package is linked into every binary and switched on in none of them.

So the 213 packages `gofr_nopubsub` removes aren't only 8.9 MB of binary. They're 213 packages of supply-chain surface that a non-pub/sub service carries permanently, and that cost recurs on its own schedule — independent of binary size, and not something a consumer can opt out of today.

For what it's worth, this also reads as a continuation of existing practice rather than a new idea: `gofr.dev/pkg/gofr/metrics/exporters/gcp` is already a separate module we require and version explicitly, while `pubsub/*`, `redis`, `sql` and `file` are in core.

Happy to build our service against the tag and report back if that would be useful.

aryanmehrotra:
Thanks, this is a useful data point, and yes please: building your service against this branch and reporting back would help a lot.

The branch head is `a969749b7`, so `go get gofr.dev@a969749b743012ce06b24b05f8c7061fd56b5e53` pins it. The numbers that would matter most:

1. **Binary size and package count** for your real service, default vs `-tags gofr_nopubsub` (and all three tags if the service allows it): `ls -l` on the binary, plus `go list -deps -tags gofr_nopubsub ./... | wc -l`.
2. **Linked packages:** `go list -deps -tags gofr_nopubsub ./... | grep enterprise-certificate-proxy` should print nothing.
3. **Idle RSS**, if you have a quick way to take it.
4. Anything that **fails to compile or behaves differently** under the tag.

One limit to be upfront about, since it bears directly on the failure you hit. I reproduced it against a minimal service on this branch:

| | default | `gofr_nopubsub` |
|---|---|---|
| `enterprise-certificate-proxy` packages in `go list -deps` | 2 | **0** |
| present in `go.sum` / `go mod why -m` | yes | **yes** |
| fetched by `go mod download` | yes | **yes** |

The tag removes the code from the **build graph**, so it is no longer linked into or compiled into your binary. It does not remove it from the **module graph**. `cloud.google.com/go/pubsub` is still a requirement in GoFr's `go.mod`, and `go mod download` and `go mod tidy` resolve modules across all build tags. So your `GOPROXY=direct` job would still have fetched `v0.3.19` and failed on the re-tag, tag or not.

Removing it from the module graph as well would mean moving the pub/sub clients into their own modules, the way `metrics/exporters/gcp` already is. That's a larger, breaking change and out of scope here, but your report is a good argument for it as a follow-up.

PiyushSingh-ZS (commented):
The engineering here is careful and the measurements are worth having. Before line-level review, though, I think there is a mechanism question that maintainers should rule on, because it decides whether most of this diff is the right shape.

### What is clearly right

The warning that **the interface alone does nothing** is the most valuable thing in the PR:

> Extracting `graphQLRunner` left all the graphql packages linked until `graphql.go` itself carried the tag -- the concrete file still compiles the library in.

That is the failure mode people ship by accident and then wonder why the binary did not shrink. Catching it and stating it plainly is a real contribution regardless of what happens to the rest.

The `modernc.org/libc` finding is also specific and verifiable: `netdb` init parsing embedded `/etc/protocols` and `/etc/services` into permanent Go structs, retained for the life of the process, unreachable from the HTTP path.

---

### The mechanism question: build tags, or the pattern the repo already has?

GoFr already ships **25 nested `go.mod` modules** under `pkg/gofr/datasource/` — `mongo`, `cassandra`, `clickhouse`, `scylladb`, `elasticsearch`, `oracle`, `surrealdb`, `dgraph`, `file/s3`, `file/gcs`, `kv-store/badger`, and directly relevant here: **`pubsub/nats`, `pubsub/sqs`, `pubsub/eventhub`**.

Three pub/sub backends are *already* separate modules. Kafka, Google and MQTT are the outliers that stayed in the root module. So for `gofr_nopubsub` — by some distance the largest win in this PR, **213 packages** and 8.9 MB — there is an existing, proven, zero-build-tag answer: extract `pubsub/kafka`, `pubsub/google` and `pubsub/mqtt` into their own modules, exactly like `pubsub/nats`.

| | separate module | build tag |
|---|---|---|
| Default build | unchanged | unchanged |
| Opt-in | `go get` + wire it | remember `-tags` at every build site |
| Wrong config | compile error at the wiring site | runtime ERROR, service starts degraded |
| CI | ordinary `go build ./...` | N-way tag matrix |
| Coverage gate | ordinary | tag-only files invisible to the default run |
| `go doc` | one surface | differs per tag |
| Precedent in this repo | 25 | 0 |

There is also a distribution problem tags have and modules do not: a tag is set by whoever runs `go build`, not by the dependency graph. A user building through Docker, `ko`, `goreleaser`, Bazel or an inherited `Makefile` has to thread `-tags gofr_nopubsub,gofr_nosqldrivers,gofr_nographql` through every one of those. Miss it in a single place and the binary silently reverts to 57 MB, with nothing to indicate it. A module boundary cannot be forgotten.

I do not think that argument is fatal to the whole PR — GraphQL lives in `package gofr` itself and the SQL drivers are blank imports for `database/sql` registration, so neither moves to a module easily, and a tag may genuinely be the only option for those two. But those are also the two smallest wins (1.4 MB and 6.1 MB against pub/sub's 8.9 MB). It would be worth splitting the pub/sub half out and asking whether it should be modules instead.

---

### If tags are the agreed mechanism, notes on the implementation

**1. The tagged build never exercises the tagged behavior for SQL.**

`drivers_testdeps_test.go` blank-imports `lib/pq` and `modernc.org/sqlite` for tests, so under `-tags gofr_nosqldrivers` the test binary still has both drivers registered. The comment explains why and the reasoning is fair. But the consequence is that the documented behavior —

> `DB_DIALECT=postgres` or `=sqlite` then fails at startup: NewSQL's `registerOtel` call reports `database/sql`'s own "unknown driver" error naming the dialect, and returns a nil DB.

— is never tested in *any* configuration. One test that opens an unregistered dialect directly would cover it without disturbing the rest of the suite.

**2. Coverage.** Roughly half the added code (`*_disabled.go`) compiles only under a tag, so the code-climate gate CONTRIBUTING relies on will never see it. The Slim Build Tags job is the right mitigation — worth saying so in the PR description so the coverage delta is not read as a regression by a reviewer who has not spotted the job.

**3. `enabled()` as an interface method.** `disabledGraphQL.buildSchema()` returns nil and `GetHandler()` returns nil purely because `enabled()` guards them, which leaves two methods whose contract is "never call me". A `const graphQLLinked = false` — mirroring `pubsubBackendsLinked`, which this PR already introduces for pub/sub — would be consistent with the PR's own vocabulary and would let the compiler drop the dead branches outright.

**4.** The `noopResponder` relocation rationale is correct and non-obvious (`responder.go` cannot carry the tag because it holds the exported `Responder` interface, and leaving `noopResponder` there makes it `unused` under the tag and fails the lint gate). Worth keeping that comment exactly as written — it will save someone an afternoon.

**5.** `createMqttPubSub` returning an untyped nil in the disabled stub is clean, and composes correctly with #4164, which switches the container's pub/sub guard to `isNil`. Might be worth cross-linking the two.

Verifying identical `go doc -all` across 6 packages and an identical linked-package list for the default build is exactly the assertion this class of change needs.

.github/workflows/go.yml
```diff
@@ -568,6 +568,47 @@ jobs:
...
+            go test  -tags "$tags" -count=1 ./pkg/gofr/container/ ./pkg/gofr/datasource/sql/
+            echo "::endgroup::"
+          done
+
+          # The GraphQL stubs have their own tests, which only exist under the tag.
+          go test -tags gofr_nographql -count=1 -run TestGraphQLDisabled ./pkg/gofr/
+
+      # The tags are opt-in, so the default build must stay exactly as it was.
+      # This is the assertion that keeps that true.
+      - name: Default build links the same packages as before
```
Umang01-hash, line 630: This step compiles + tests but never compares the linked-package set to a baseline, so it doesn't assert what its name/comment claim. A `go list -deps ./... | sort` diff against a committed snapshot (or asserting the graphql/sqldriver/pubsub dep count is unchanged) would make it real — and would catch an accidental default-build regression.

pkg/gofr/datasource/sql/drivers_disabled.go
```diff
@@ -0,0 +1,15 @@
...
+// call reports database/sql's own "unknown driver" error naming the dialect, and
+// returns a nil DB. That is an already-supported state -- it is what an
+// unconfigured database produces, and the container guards it with isNil -- so
+// the service still starts, with one error line saying exactly what is missing.
+// A loud, immediate complaint is the point of doing this with a tag rather than
+// by asking users to blank-import drivers themselves, where the same mistake is
+// silent. A user who wants one of these dialects in a tagged build imports
+// the driver in their own main package, exactly as with database/sql directly.
+
+package sql
```
Umang01-hash, line 19: The fail-loud path this file creates (DB_DIALECT=postgres/sqlite → unknown-driver → nil DB) has no test. drivers_testdeps_test.go blank-imports pq+sqlite untagged, so under -tags gofr_nosqldrivers the test binary re-registers the drivers and the disabled branch is never exercised in CI — the scaffolding masks it. A `//go:build gofr_nosqldrivers` test asserting NewSQL(...postgres...) returns nil would match the graphql_disabled_test.go treatment.

pkg/gofr/container/pubsub_backends_disabled.go
```diff
@@ -0,0 +1,38 @@
...
+const pubsubDisabledMsg = "PUBSUB_BACKEND=%s was configured, but this binary was " +
+	"built with -tags gofr_nopubsub, which omits the %s client. Rebuild without the tag to use it."
+
+func (c *Container) createMqttPubSub(config.Config) pubsub.Client {
+	c.Logger.Errorf(pubsubDisabledMsg, "MQTT", "MQTT")
+
+	return nil
+}
+
+func (c *Container) createKafkaPubSub(config.Config) {
```
Umang01-hash, line 32: The disabled stubs log fail-loud (pubsubDisabledMsg) but nothing asserts it — container_test.go only adds t.Skip guards. A regression dropping this Errorf (a publisher that silently never publishes) wouldn't fail CI. Worth a tagged test asserting the error is logged and PubSub stays nil, for parity with the graphql disabled path.

Umang01-hash (changes requested):
Verified end-to-end at the head SHA in a clean worktree: each tag drops its tree completely (graphql→0, sqldrivers→0, pubsub→0 syms), the full set takes a real binary 56.3MB→39.6MB (−16.7MB/~30%), no API break, no typed-nil, fail-loud confirmed by actually running tagged binaries (PUBSUB_BACKEND/DB_DIALECT → ERROR naming the tag, nil datasource, service still boots). Clean, effective work — the interface seams and verbatim moves are well done, and CI building the all-3 combination is the right call.

Requesting changes on disabled-path test coverage + one overstated CI claim — same pattern flagged on #4168, plus two gaps unique here:

1. The "Default build links the same packages as before" step doesn't assert that — it only compiles + tests. A `go list -deps` diff (or dep-count check) vs base would make the step earn its name and guard the whole PR's value.
2. The SQL disabled fail-loud path (unknown-driver → nil DB) has no test, and `drivers_testdeps_test.go` re-registers pq+sqlite untagged, so CI actively masks the disabled state. GraphQL got a dedicated disabled test; SQL should have parity.
3. The pubsub disabled fail-loud (`pubsubDisabledMsg`) is untested — a regression dropping the Errorf (silent no-publish) wouldn't fail CI.

Minor (non-blocking): the graphql_runner.go doc says the tag drops "1.4 MB"; the real measured drop is ~0.83 MB.

aryanmehrotra:
@PiyushSingh-ZS @Umang01-hash — everything addressed at `4e2121577`. All 21 checks green.

## Tags vs modules: these three are a closed exception

Agreed that modules are the right pattern, and that's the rule from here on. These three tags are deliberately limited to subsystems **already** compiled into core, where moving them out breaks existing users:

| Tag | What a module would break |
|---|---|
| `gofr_nopubsub` | Kafka/Google/MQTT start from config alone (`PUBSUB_BACKEND`, `container.go:163`). Every user adds a `go get` + an `AddPubSub` call. |
| `gofr_nosqldrivers` | `DB_DIALECT=postgres/sqlite` needs no import today. Users import the driver themselves. Also: the drivers are third-party, so there is no GoFr code to move. |
| `gofr_nographql` | `App.GraphQLQuery`/`GraphQLMutation` are exported methods on `App`. A module needs a new setup call. |

A *new* integration has no such users, so there is nothing to break — it goes in its own module, which is the existing pattern. I've written that into `CONTRIBUTING.md` so the tags don't become a precedent, since it was nowhere before.

## @Umang01-hash's three threads — all real, all fixed

**1. The CI step didn't check what its name claimed.** Correct, and it was the worst kind of comment: it asserted the PR's central premise and verified nothing. It now diffs `go list -deps ./... | sort` against the merge base and fails on any change. Baseline rather than a committed file, so it can't go stale. Verified locally: **861 packages, identical**.

**2. The unknown-driver path had no test, and your point about the scaffolding masking it is exactly right.** That's why the assertion is against a dialect registered in **neither** build rather than `postgres` under the tag — `drivers_testdeps_test.go` blank-imports pq and sqlite so the suite behaves identically either way, which means a tagged assertion about postgres would be testing the fixture, not the code. `TestRegisterOtel_UnregisteredDialectFailsLoudly` exercises the same `otelsql.Register` → `sql.Open` → `unknown driver` path in both builds, so it cannot quietly stop running.

**3. The disabled pub/sub stubs had no assertion.** Also right — skips prove nothing about what the stub does, and a dropped `Errorf` is precisely the silent-publisher failure the tag exists to avoid. `pubsub_backends_disabled_test.go` (tagged, so it runs in the Slim job) asserts all three name the client and the tag, and that `PubSub` stays nil. Removing one `Errorf` makes it fail.

## Also found while fixing those

- **The tagged build was never linted.** Added, scoped with `only-new-issues` — and its first run found a real `noctx` failure in `graphql_disabled_test.go` that no job would ever have reported. Same class of problem that prompted moving `noopResponder` and `configTrue`.
- **`drivers_disabled.go` named only postgres and sqlite.** `supabase` and `cockroachdb` register under the postgres driver (`sql.go:268`), so the tag affects them too — a supabase user reading the old comment would have been surprised.

## @PiyushSingh-ZS's implementation notes

- **`enabled()` → `const graphQLLinked`**, matching `pubsubBackendsLinked`. Whether the engine is linked is a property of the build, not of an instance, and the compiler folds the branch. The stub keeps `buildSchema`/`GetHandler` because untagged `gofr.go` calls them.
- **Coverage**: the tagged files aren't compiled in the default run, so they're *unmeasured* rather than uncovered — noted in the description, and the Slim job now lints as well as tests them.
- **`noopResponder` comment** kept as-is, agreed.
- **Cross-link to #4164** added.

## Documentation

There was none for any of these tags — not in the PR, `CONTRIBUTING.md` or `docs/`. Added `docs/advanced-guide/slim-builds`, covering #4168's three tags too, with the numbers measured rather than asserted: **829 packages by default, 617 / 785 / 809 per tag, 553 with all three.**

.github/workflows/go.yml
aryanmehrotra: Done — the step now diffs the package set, in two places.

`.github/workflows/go.yml:693` **Default build links the same packages as before** (added in 4e21215) does `go list -deps ./... | sort` on HEAD and on a worktree at `git merge-base HEAD origin/$BASE_REF`, and fails on any diff. The baseline is the merge base rather than a committed snapshot so it cannot go stale. `fetch-depth: 0` was added to this job's checkout (`go.yml:584`) so the merge base exists.

`.github/workflows/go.yml:640` **Each tag removes the packages it claims to** (added in 3046b60) is the other half your comment implied: the default build being unchanged says nothing about the tags doing anything. It asserts per tag that `lib/pq`/`modernc.org/sqlite`, `kafka-go`/`cloud.google.com/go/pubsub`/`paho.mqtt.golang` and `graphql-go/graphql` are gone, and fails if a pattern is missing from the *default* build too, so a renamed module cannot make it pass vacuously.

It queries the package path, not `./...` — with a pattern `go list -deps` lists every matched package whether or not anything links it, which reports a tagged-out package as still present.

Measured at 3046b60 (`go list -deps gofr.dev/pkg/gofr | wc -l`):

| build | packages | `examples/http-server` binary |
|---|---|---|
| default | 828 | 60,028,914 B |
| `gofr_nosqldrivers` | 784 (−44) | 53,539,602 B (−6,489,312) |
| `gofr_nopubsub` | 616 (−212) | 50,727,890 B (−9,301,024) |
| `gofr_nographql` | 808 (−20) | 59,164,482 B (−864,432) |
| all three | 552 (−276) | 43,357,218 B (−16,671,696, −27.8%) |

pkg/gofr/datasource/sql/drivers_disabled.go
aryanmehrotra: You're right that it was untested. It is covered now, but not by the tagged test you suggested — that one is not achievable in this package, and I want to show why rather than claim it is.

A tagged `NewSQL(...postgres...) == nil` test needs the test binary to *not* have the postgres driver. Removing `drivers_testdeps_test.go` from the tagged build is what that takes, and it breaks the existing suite:

```
$ go test -tags gofr_nosqldrivers ./pkg/gofr/datasource/sql/   # with testdeps removed
--- FAIL: TestNewSQL_GetDBDialect
  sql.(*DB).Dialect  db.go:98
  sql.TestNewSQL_GetDBDialect  sql_test.go:89
```

`sql_test.go:89` opens a postgres DSN through `NewSQL` and dereferences the result, so the driver has to be registered in the test binary in both configurations. `database/sql` has no unregister, so one binary cannot have postgres both registered and not.

So it is split in two, and both halves are new:

1. **The mechanism**, `drivers_registration_test.go:46` `TestRegisterOtel_UnregisteredDialectFailsLoudly` (4e21215). The fail-loud path is entirely `otelsql.Register` → `sql.Open` → `unknown driver` → `NewSQL` returns nil. It asserts that against a name registered in *neither* build, so it exercises the same code in both and cannot quietly stop running. A dependency bump that made `Register` tolerate an unknown driver — the thing that would turn the loud failure silent — fails here. `drivers_registration_test.go:61` also pins the supabase/cockroachdb → postgres aliasing, which decides who the tag affects.

2. **That the tag really removes the drivers**, `.github/workflows/go.yml:640` (3046b60). This is the part a test in that package cannot see, precisely because of the fixtures. It fails if `github.com/lib/pq` or `modernc.org/sqlite` is still in `go list -tags gofr_nosqldrivers -deps gofr.dev/pkg/gofr`. Dropping the `//go:build !gofr_nosqldrivers` line from `drivers.go`, or an import elsewhere pulling pq back in, now fails CI. Measured: 828 packages by default, 784 under the tag; `examples/http-server` 60,028,914 B → 53,539,602 B.

I also renamed `drivers_disabled_test.go` → `drivers_registration_test.go`. It has no build tag on purpose, and that name sitting next to a tagged `drivers_disabled.go` read like one that had been forgotten.

pkg/gofr/container/pubsub_backends_disabled.go
aryanmehrotra: Done — `pkg/gofr/container/pubsub_backends_disabled_test.go` (4e21215), `//go:build gofr_nopubsub` on line 1, same shape as `graphql_disabled_test.go`.

Three tests, one per stub: `TestPubSubDisabled_MQTTReportsTheTag` (:35), `_KafkaReportsTheTag` (:47), `_GoogleReportsTheTag` (:58). Each captures stderr around the call and asserts, in `assertNamesTheTag` (:73), that the log is non-empty, names the missing client, names `gofr_nopubsub`, and says "rebuild". The MQTT one also asserts the returned client is nil rather than a typed nil; the Kafka and Google ones assert `c.PubSub` is still nil. Dropping any `Errorf` fails all four assertions on that backend.

It runs in the Slim Build Tags job, which is the only place it compiles:

```
$ go test -tags gofr_nopubsub ./pkg/gofr/container/ -count=1 -run Disabled -v
--- PASS: TestPubSubDisabled_MQTTReportsTheTag (0.00s)
--- PASS: TestPubSubDisabled_KafkaReportsTheTag (0.00s)
--- PASS: TestPubSubDisabled_GoogleReportsTheTag (0.00s)
ok  gofr.dev/pkg/gofr/container  0.529s
```

Verified at 3046b60. `.github/workflows/go.yml:640` additionally fails if `kafka-go`, `cloud.google.com/go/pubsub` or `paho.mqtt.golang` is still linked under the tag — 828 packages by default, 616 with `gofr_nopubsub`, and `examples/http-server` 60,028,914 B → 50,727,890 B.

akshat-kumar-singhal (approved):
Approve with nits (one should-fix in CI). Default build unchanged: `go list -deps ./...` is identical to development (840 packages), and it merges cleanly. No exported API removed; `GraphQLLog` is the only exported type moved behind a tag and nothing outside graphql.go uses it. `go build`/`go vet`/`go test` pass for default, each tag alone, and all three. Ran a binary with all three tags: `PUBSUB_BACKEND=KAFKA|GOOGLE|MQTT`, `DB_DIALECT=postgres|sqlite|supabase` and a GraphQL resolver each log the documented message, no `/graphql` route is registered, and the service still starts.

**Should-fix: "Default build links the same packages as before" will fail unrelated PRs** — [`go.yml:693`](https://github.com/gofr-dev/gofr/blob/2139269c31e135c771c7a7e8332b439aa32d074e/.github/workflows/go.yml#L693)
It diffs `go list -deps ./...` of HEAD against the merge base across the whole root module, so any future PR that adds a package or a dependency turns `slim_build_tags` red. Drop it (the "Each tag removes the packages it claims to" step already covers the direction that matters), or assert only that the default `go list -deps gofr.dev/pkg/gofr` still contains each tagged-out pattern.

**Nit:** the counts in `slim-builds/page.md` are platform/toolchain dependent. Go 1.26.0 linux/amd64, `gofr.dev/pkg/gofr`: none 808 (doc 828), nopubsub 595 (616), nosqldrivers 788 (784), nographql 788 (808), all three 555 (552). Plausible std/libc differences; state the GOOS/GOARCH and Go version measured with.

**Note:** this and #4168 conflict in `docs/advanced-guide/slim-builds/page.md` and disagree on the default count (828 vs 829); #4168 lacks this PR's last three commits (incl. the CI tag check, 8803000).

aryanmehrotra:
Rebased onto `development` — head is `f1634f258`, conflicts resolved, no longer `DIRTY`.

@akshat-kumar-singhal — your CI should-fix is taken, and I went with your **first** option rather than the second, because the second turned out to be already implemented.

### The whole-module dep diff is gone

You were right about the failure mode: `slim_build_tags` runs on every PR, so diffing `go list -deps ./...` against the merge base means *"no PR may add a package or a dependency, ever"*. That is not a rule this repo has, and it would have gone red on work with nothing to do with build tags.

Your suggested replacement — assert the default build still *contains* each tagged-out pattern — is what the **"Each tag removes the packages it claims to"** step above it already does:

```yaml
# Also fails when the packages are absent from the DEFAULT build, so a
# renamed or dropped module cannot make this check pass vacuously.
if ! grep -Eq "$pattern" /tmp/deps.default; then
```

So adding it separately would have duplicated it. I dropped the broad diff and left a comment in its place explaining why, and asking that it not be re-added — @Umang01-hash originally asked for this step, and without the note the next reader has one reviewer's request in the history and no record of why it went away.

@Umang01-hash — flagging that directly since it reverses part of what you asked for. The half of your comment that matters is intact: the step genuinely did not assert what its name claimed, and the fix for *that* is the per-tag check, which does the comparison against the default build. What is removed is only the "nothing may ever change" part. Happy to restore it scoped to `gofr.dev/pkg/gofr` instead of `./...` if you would rather keep a diff.

### The rebase caught two tests that do not survive a tag

Both came in from `development` and both failed `go vet`/`go test` under a tag while the non-tagged build stayed green — which is exactly the blind spot this PR's CI job exists for, so it is a point in the job's favour:

| test | why it broke | fix |
|---|---|---|
| `TestApp_setupGraphQL_MissingSchema` (#4274) | names `errSchemaMissing`, compiled out by `gofr_nographql`, and lived in the untagged `gofr_test.go` | moved into `graphql_test.go`, which already carries `!gofr_nographql` |
| `TestContainer_createKafkaPubSub_InvalidConfigs` | asserts what `kafka.New`'s validation logs; under `gofr_nopubsub` there is no `kafka.New` | takes the `pubsubBackendsLinked` skip the file already uses for three other tests |

This is the ongoing cost @PiyushSingh-ZS predicted in the coverage note, and worth saying plainly: every new untagged test that touches a tagged symbol lands as a red tagged build until someone moves it.

### Doc counts

Now labelled **darwin/arm64, Go 1.26.3**, with a note that the counts are platform-dependent and that the *differences* between rows are the point — so your linux/amd64 figures being ~20 lower is explained rather than looking like an error.

### Verification

All five configurations — default, each tag alone, and all three together — `go build`, `go vet` and `go test` clean.

`typos` clean, `gofmt` clean, `go.yml` parses. `golangci-lint ./pkg/gofr/ ./pkg/gofr/container/` reports **5 issues against 7 on `development`** — none new.

Umang01-hash (commented):
Re-verified locally at the head SHA against the three changes requested on 2026-09-18 — all three are addressed:

1. Default-build dep parity is now asserted, not just compiled: the `Each tag removes the packages it claims to` step runs `go list -deps gofr.dev/pkg/gofr` per tag and fails if a tagged-out package is missing from the default build (so it can't pass vacuously). Reproduced: default 828 pkgs -> nosqldrivers 784, nopubsub 616, nographql 808; 0 of the removed deps remain linked under their tag; default still links all of them (13 pq/sqlite, 60 kafka/google/mqtt, 11 graphql).
2. SQL fail-loud is tested: `TestRegisterOtel_UnregisteredDialectFailsLoudly` asserts the unknown-driver error via a dialect registered in neither build (dodging the drivers_testdeps fixture), plus `TestRegisterOtel_AliasedDialectsUsePostgres` for supabase/cockroach. Passes.
3. Pub/sub fail-loud is tested: `pubsub_backends_disabled_test.go` (tagged) asserts each stub logs the tag + client + rebuild hint and leaves PubSub nil. Passes.

Also verified: all four tag combos `go build`/`go test` clean; default exported API of `pkg/gofr`, `container`, `sql` is byte-identical to development (no break; `GraphQLLog` present in the default build); golangci-lint 0 new issues both default and tagged.

Still open:
- Minor, unaddressed: `graphql_runner.go` still says the tag drops `1.4 MB` (measured ~0.83 MB).
- The mechanism question @PiyushSingh-ZS raised — build tags vs extracting the pub/sub backends into modules like the repo's existing 25 — is a maintainer design call, not a code issue. It's the one thing that isn't just verifiable, and it's what decides whether the pub/sub half of this stays tag-based.

The implementation itself is correct and does what it claims.

PR #4168, opened 2026-09-08
perf(deps): let a build omit the gRPC server, Dgraph migrator and OTLP exporters it does not use

aryanmehrotra:
> ### ⚠️ Read first
> **Stacked on #4167** — review that one first. This PR's base is `perf/optional-datasources-graphql`, so the diff below is only the three tags described here.
>
> **CI does not run on this PR yet.** `.github/workflows/go.yml` triggers on `pull_request: branches: [main, development]`, and this one targets a feature branch. **Retarget it to `development` once #4167 merges** and the full pipeline fires. Local verification is in *Additional Information* below.

**Description:**

Three more opt-in build tags. **A build that sets none of them is unchanged.**

| tag | omits |
|---|---|
| `gofr_nogrpc` | GoFr's gRPC server, reflection, health, recovery middleware |
| `gofr_nodgraph` | the Dgraph **migrator** — the only dgo importer in GoFr |
| `gofr_nootlp` | the OTLP trace and metric transports |

| build | binary | packages | gRPC pkgs |
|---|---|---|---|
| default | 57.10 MB | 831 | 86 |
| + the two tags from #4167 | 41.26 MB | 554 | 70 |
| + `gofr_nogrpc` | 40.59 MB | 543 | 66 |
| + `gofr_nodgraph` | 39.47 MB | 541 | 66 |
| + `gofr_nootlp` | **35.35 MB** | **414** | **0** |

**All six tags: 57.10 → 35.35 MB (−38.1%), 831 → 414 packages (−50.2%).**

**⚠️ These tags are conjunctive, not additive.** On its own `gofr_nogrpc` is worth **0.67 MB** and `gofr_nodgraph` **1.12 MB** — which looks like nothing. That is the point worth understanding before judging them.

`google.golang.org/grpc` was pinned by **four independent importers**:

- GoFr's own gRPC server
- dgo's protobufs, via **one** `&api.Mutation{}` in `pkg/gofr/migration`
- `otlptracegrpc` (`otel.go`) and `otlpmetricgrpc` (`metrics/exporters`)
- `otlpmetrichttp` — which imports grpc for its status codes, so the HTTP transport cannot be kept either

**Only the last tag releases the tree**, and it then measures 4.12 MB / 127 packages. Each of the first two is a prerequisite, not a standalone win.

⚠️ **`gofr_nogrpc` is the only tag in either PR that changes the exported API — and only when the tag is set.**

`RegisterService`, `AddGRPCServerOptions`, `AddGRPCUnaryInterceptors` and `AddGRPCServerStreamInterceptors` all name grpc types in their signatures, so they **cannot exist without the import** and are compiled out.

A build that sets the tag and calls them **fails to compile**. That is the intended outcome — it asked for a binary without gRPC and then used gRPC — and it is why this tag's CI step builds `./pkg/...` rather than `./...`: the gRPC examples are *supposed* to fail under it.

**A default build keeps all four methods and is unchanged.**

Under `gofr_nootlp` and `gofr_nodgraph` the affected features fail loudly rather than silently degrading:
- `"otlp"` stays registered in the metrics registry and fails where it would open the connection — rather than vanishing and letting `METRICS_EXPORTER=otlp` quietly resolve to a *different* exporter than the operator configured
- `checkAndCreateMigrationTable` returns an error naming the tag and `Run` treats it as fatal — which is what stops a migration being recorded as applied when it was not

Prometheus, `gcp`, `TRACE_EXPORTER=zipkin` and `TRACE_EXPORTER=gofr` are untouched, so a tagged build can still export metrics and traces.

**Additional Information:**

- No new dependencies — this PR only removes linkage.
- **HTTP auth is not gRPC's to take away.** `EnableBasicAuth` and the other five entry points build gRPC interceptors, which names `pkg/gofr/grpc/middleware`, which imports grpc — so `auth.go` could not stay untagged while doing it. The gRPC half now lives in `grpc_auth_enabled.go` behind three bridges taking plain credentials; `auth.go` keeps the HTTP half in every build.

**Four pre-existing defects fixed on the way:**

1. **`initTracer` built a `BatchSpanProcessor` over a nil exporter.** `getExporter` returns nil-and-no-error for an unsupported `TRACE_EXPORTER` — a typo in config is enough. The processor panics on its first flush, on its own goroutine. It now returns, leaving the provider installed so trace and span IDs still appear on every request.
2. **`container.Error(err)` logged `{"message":{}}`.** The logger JSON-marshals the value and an `errors.errorString` has no exported fields, so **every trace-exporter failure logged nothing usable**. Uses `Errorf` now. ⚠️ The same pattern is likely elsewhere and deserves its own sweep.
3. **Four exported gRPC methods dereferenced `App.grpcServer` without checking it.** `factory.go` leaves it nil when the server fails to build — an out-of-range `GRPC_PORT` is enough — and logs and continues, so all four panicked in a case the surrounding code treats as recoverable. They log and return now.
4. **`TestUnifiedAuthenticationRegistration` could not fail.** It asserted `>= 2` interceptors after three `Enable` calls; two are registered by default, so it passed on the defaults alone and would not have noticed any auth bridge going missing. It now asserts exactly five.

**Local verification, since CI cannot run here yet:**

- `go build ./...` and `go vet ./pkg/...` clean in the default build and under all six tags
- **30/30 packages pass** in both configurations
- `golangci-lint` clean on every file this PR adds or changes, in both configurations

**Checklist:**

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -568,6 +568,18 @@ jobs:
           go vet   -tags gofr_nodgraph ./pkg/gofr/migration/
           go test  -tags gofr_nodgraph -count=1 ./pkg/gofr/migration/
 
+      # gofr_nootlp is the tag that actually releases google.golang.org/grpc: the
+      # OTLP exporters pin it, otlpmetrichttp included, so gofr_nogrpc and
+      # gofr_nodgraph only pay off once this one is set too. Both halves of the
+      # trace and metric transports have tests that exist only under the tag.
+      - name: Build and test with the OTLP exporters omitted
+        run: |
+          set -euo pipefail
+          go build -tags gofr_nootlp ./...
+          go vet   -tags gofr_nootlp ./pkg/gofr/ ./pkg/gofr/metrics/exporters/
+          go test  -tags gofr_nootlp -count=1 ./pkg/gofr/metrics/exporters/
+          go test  -tags gofr_nootlp -count=1 -run 'TestOTLPTracesOmitted|Test_initTracer' ./pkg/gofr/
+
       # The tags are opt-in, so the default build must stay exactly as it was.
       # This is the assertion that keeps that true.
       - name: Default build links the same packages as before
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -546,6 +546,17 @@ jobs:
           # The GraphQL stubs have their own tests, which only exist under the tag.
           go test -tags gofr_nographql -count=1 -run TestGraphQLDisabled ./pkg/gofr/
 
+      # gofr_nogrpc is scoped to ./pkg/... on purpose. Unlike the other tags it
+      # removes exported methods -- RegisterService and the AddGRPC* setters name
+      # grpc types and cannot exist without the import -- so the gRPC examples do
+      # not compile under it, which is the intended behaviour and not a failure.
+      - name: Build and test with the gRPC server omitted
+        run: |
+          set -euo pipefail
+          go build -tags gofr_nogrpc ./pkg/...
+          go vet   -tags gofr_nogrpc ./pkg/gofr/
+          go test  -tags gofr_nogrpc -count=1 -run TestGRPCDisabled ./pkg/gofr/
+
       # The tags are opt-in, so the default build must stay exactly as it was.
       # This is the assertion that keeps that true.
       - name: Default build links the same packages as before
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -557,6 +557,17 @@ jobs:
           go vet   -tags gofr_nogrpc ./pkg/gofr/
           go test  -tags gofr_nogrpc -count=1 -run TestGRPCDisabled ./pkg/gofr/
 
+      # gofr_nodgraph drops the Dgraph migrator, the one place in GoFr that imports
+      # dgo. The stub has to keep implementing migrator, so a change to that
+      # interface must update both halves; without this step only the real one
+      # would ever be compiled.
+      - name: Build and test with the Dgraph migrator omitted
+        run: |
+          set -euo pipefail
+          go build -tags gofr_nodgraph ./...
+          go vet   -tags gofr_nodgraph ./pkg/gofr/migration/
+          go test  -tags gofr_nodgraph -count=1 ./pkg/gofr/migration/
+
       # The tags are opt-in, so the default build must stay exactly as it was.
       # This is the assertion that keeps that true.
       - name: Default build links the same packages as before
--- a/pkg/gofr/grpc.go
+++ b/pkg/gofr/grpc.go
@@ -1,3 +1,5 @@
+//go:build !gofr_nogrpc
+
 package gofr
 
 import (
@@ -59,8 +61,14 @@ func (a *App) AddGRPCServerOptions(grpcOpts ...grpc.ServerOption) {
 		return
 	}
 
+	g := a.grpcSrv()
+	if g == nil {
+		a.container.Logger.Error("no gRPC server to add options to")
+		return
+	}
+
 	a.container.Logger.Debugf("adding %d gRPC server options", len(grpcOpts))
-	a.grpcServer.options = append(a.grpcServer.options, grpcOpts...)
+	g.addServerOptions(grpcOpts...)
 }
 
 // AddGRPCUnaryInterceptors allows users to add custom gRPC interceptors.
@@ -78,8 +86,14 @@ func (a *App) AddGRPCUnaryInterceptors(interceptors ...grpc.UnaryServerIntercept
 		return
 	}
 
+	g := a.grpcSrv()
+	if g == nil {
+		a.container.Logger.Error("no gRPC server to add unary interceptors to")
+		return
+	}
+
 	a.container.Logger.Debugf("adding %d valid unary interceptors", len(interceptors))
-	a.grpcServer.interceptors = append(a.grpcServer.interceptors, interceptors...)
+	g.addUnaryInterceptors(interceptors...)
 }
 
 func (a *App) AddGRPCServerStreamInterceptors(interceptors ...grpc.StreamServerInterceptor) {
@@ -88,8 +102,42 @@ func (a *App) AddGRPCServerStreamInterceptors(interceptors ...grpc.StreamServerI
 		return
 	}
 
+	g := a.grpcSrv()
+	if g == nil {
+		a.container.Logger.Error("no gRPC server to add stream interceptors to")
+		return
+	}
+
 	a.container.Logger.Debugf("adding %d stream interceptors", len(interceptors))
-	a.grpcServer.streamInterceptors = append(a.grpcServer.streamInterceptors, interceptors...)
+	g.addStreamInterceptors(interceptors...)
+}
+
+// grpcSrv returns the concrete server behind App's grpcRunner field, or nil when
+// this app has none -- newGRPCRunner failed, typically on an out-of-range
+// GRPC_PORT, and factory.go logs and continues. The exported AddGRPC* methods
+// dereferenced the field blind before it became an interface and panicked in
+// exactly that case; they now log and return.
+func (a *App) grpcSrv() *grpcServer {
+	g, ok := a.grpcServer.(*grpcServer)
+	if !ok {
+		return nil
+	}
+
+	return g
+}
+
+// newGRPCRunner is the enabled half of the pair declared in grpc_runner.go.
+//
+// It returns the interface, not *grpcServer, so that a nil result cannot reach
+// App as a non-nil interface holding a nil pointer -- the typed-nil trap that
+// container.isNil exists to catch elsewhere.
+func newGRPCRunner(c *container.Container, port int, cfg config.Config) (grpcRunner, error) {
+	srv, err := newGRPCServer(c, port, cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return srv, nil
 }
 
 func newGRPCServer(c *container.Container, port int, cfg config.Config) (*grpcServer, error) {
@@ -244,7 +292,13 @@ func (g *grpcServer) forceStop() {
 
 // RegisterService adds a gRPC service to the GoFr application.
 func (a *App) RegisterService(desc *grpc.ServiceDesc, impl any) {
-	srv, err := a.grpcServer.ensureServer()
+	g := a.grpcSrv()
+	if g == nil {
+		a.container.Logger.Errorf("no gRPC server to register service %s on", desc.ServiceName)
+		return
+	}
+
+	srv, err := g.ensureServer()
 	if err != nil {
 		a.container.Logger.Errorf("failed to create gRPC server for service %s: %v", desc.ServiceName, err)
 		return
--- a/pkg/gofr/auth.go
+++ b/pkg/gofr/auth.go
@@ -7,10 +7,8 @@ import (
 	"time"
 
 	"github.com/golang-jwt/jwt/v5"
-	"google.golang.org/grpc"
 
 	"gofr.dev/pkg/gofr/container"
-	grpcMiddleware "gofr.dev/pkg/gofr/grpc/middleware"
 	"gofr.dev/pkg/gofr/http/middleware"
 )
 
@@ -35,10 +33,8 @@ func (a *App) EnableBasicAuth(credentials ...string) {
 		users[credentials[i]] = credentials[i+1]
 	}
 
-	provider := grpcMiddleware.BasicAuthProvider{Users: users}
-
-	a.addAuthMiddleware(middleware.BasicAuthMiddleware(middleware.BasicAuthProvider{Users: users}),
-		grpcMiddleware.BasicAuthUnaryInterceptor(provider), grpcMiddleware.BasicAuthStreamInterceptor(provider))
+	a.addAuthMiddleware(middleware.BasicAuthMiddleware(middleware.BasicAuthProvider{Users: users}))
+	a.addGRPCBasicAuth(users, nil, nil)
 }
 
 // EnableBasicAuthWithFunc enables basic authentication for the HTTP server with a custom validation function.
@@ -46,10 +42,9 @@ func (a *App) EnableBasicAuth(credentials ...string) {
 // Deprecated: This method is deprecated and will be removed in future releases, users must use
 // [App.EnableBasicAuthWithValidator] as it has access to application datasources.
 func (a *App) EnableBasicAuthWithFunc(validateFunc func(username, password string) bool) {
-	provider := grpcMiddleware.BasicAuthProvider{ValidateFunc: validateFunc, Container: a.container}
-
-	a.addAuthMiddleware(middleware.BasicAuthMiddleware(middleware.BasicAuthProvider{ValidateFunc: validateFunc, Container: a.container}),
-		grpcMiddleware.BasicAuthUnaryInterceptor(provider), grpcMiddleware.BasicAuthStreamInterceptor(provider))
+	a.addAuthMiddleware(middleware.BasicAuthMiddleware(
+		middleware.BasicAuthProvider{ValidateFunc: validateFunc, Container: a.container}))
+	a.addGRPCBasicAuth(nil, validateFunc, nil)
 }
 
 // EnableBasicAuthWithValidator enables basic authentication for the HTTP server with a custom validator.
@@ -57,21 +52,17 @@ func (a *App) EnableBasicAuthWithFunc(validateFunc func(username, password strin
 // The provided `validateFunc` is invoked for each authentication attempt. It receives a container instance,
 // username, and password. The function should return `true` if the credentials are valid, `false` otherwise.
 func (a *App) EnableBasicAuthWithValidator(validateFunc func(c *container.Container, username, password string) bool) {
-	provider := grpcMiddleware.BasicAuthProvider{ValidateFuncWithDatasources: validateFunc, Container: a.container}
-
 	a.addAuthMiddleware(middleware.BasicAuthMiddleware(middleware.BasicAuthProvider{
-		ValidateFuncWithDatasources: validateFunc, Container: a.container}),
-		grpcMiddleware.BasicAuthUnaryInterceptor(provider), grpcMiddleware.BasicAuthStreamInterceptor(provider))
+		ValidateFuncWithDatasources: validateFunc, Container: a.container}))
+	a.addGRPCBasicAuth(nil, nil, validateFunc)
 }
 
 // EnableAPIKeyAuth enables API key authentication for the application.
 //
 // It requires at least one API key to be provided. The provided API keys will be used to authenticate requests.
 func (a *App) EnableAPIKeyAuth(apiKeys ...string) {
-	provider := grpcMiddleware.APIKeyAuthProvider{APIKeys: apiKeys}
-
-	a.addAuthMiddleware(middleware.APIKeyAuthMiddleware(middleware.APIKeyAuthProvider{}, apiKeys...),
-		grpcMiddleware.APIKeyAuthUnaryInterceptor(provider), grpcMiddleware.APIKeyAuthStreamInterceptor(provider))
+	a.addAuthMiddleware(middleware.APIKeyAuthMiddleware(middleware.APIKeyAuthProvider{}, apiKeys...))
+	a.addGRPCAPIKeyAuth(apiKeys, nil, nil)
 }
 
 // EnableAPIKeyAuthWithFunc enables API key authentication for the application with a custom validation function.
@@ -79,12 +70,11 @@ func (a *App) EnableAPIKeyAuth(apiKeys ...string) {
 // Deprecated: This method is deprecated and will be removed in future releases, users must use
 // [App.EnableAPIKeyAuthWithValidator] as it has access to application datasources.
 func (a *App) EnableAPIKeyAuthWithFunc(validateFunc func(apiKey string) bool) {
-	provider := grpcMiddleware.APIKeyAuthProvider{ValidateFunc: validateFunc, Container: a.container}
-
 	a.addAuthMiddleware(middleware.APIKeyAuthMiddleware(middleware.APIKeyAuthProvider{
 		ValidateFunc: validateFunc,
 		Container:    a.container,
-	}), grpcMiddleware.APIKeyAuthUnaryInterceptor(provider), grpcMiddleware.APIKeyAuthStreamInterceptor(provider))
+	}))
+	a.addGRPCAPIKeyAuth(nil, validateFunc, nil)
 }
 
 // EnableAPIKeyAuthWithValidator enables API key authentication for the application with a custom validation function.
@@ -92,12 +82,11 @@ func (a *App) EnableAPIKeyAuthWithFunc(validateFunc func(apiKey string) bool) {
 // The provided `validateFunc` is used to determine the validity of an API key. It receives the request container
 // and the API key as arguments and should return `true` if the key is valid, `false` otherwise.
 func (a *App) EnableAPIKeyAuthWithValidator(validateFunc func(c *container.Container, apiKey string) bool) {
-	provider := grpcMiddleware.APIKeyAuthProvider{ValidateFuncWithDatasources: validateFunc, Container: a.container}
-
 	a.addAuthMiddleware(middleware.APIKeyAuthMiddleware(middleware.APIKeyAuthProvider{
 		ValidateFuncWithDatasources: validateFunc,
 		Container:                   a.container,
-	}), grpcMiddleware.APIKeyAuthUnaryInterceptor(provider), grpcMiddleware.APIKeyAuthStreamInterceptor(provider))
+	}))
+	a.addGRPCAPIKeyAuth(nil, nil, validateFunc)
 }
 
 // EnableOAuth configures OAuth middleware for the application.
@@ -147,19 +136,14 @@ func (a *App) EnableOAuth(jwksEndpoint string,
 
 	publicKeyProvider := middleware.NewOAuth(oauthOption)
 
-	a.addAuthMiddleware(middleware.OAuth(publicKeyProvider, options...),
-		grpcMiddleware.OAuthUnaryInterceptor(publicKeyProvider, options...),
-		grpcMiddleware.OAuthStreamInterceptor(publicKeyProvider, options...))
+	a.addAuthMiddleware(middleware.OAuth(publicKeyProvider, options...))
+	a.addGRPCOAuth(publicKeyProvider, options...)
 }
 
-func (a *App) addAuthMiddleware(httpMW func(http.Handler) http.Handler,
-	grpcUnary grpc.UnaryServerInterceptor, grpcStream grpc.StreamServerInterceptor) {
+// addAuthMiddleware installs the HTTP half. The gRPC half is a separate call at
+// each site -- see grpc_auth_enabled.go for why it cannot be done here.
+func (a *App) addAuthMiddleware(httpMW func(http.Handler) http.Handler) {
 	if a.httpServer != nil {
 		a.httpServer.router.Use(httpMW)
-	}
-
-	if a.grpcServer != nil {
-		a.grpcServer.addUnaryInterceptors(grpcUnary)
-		a.grpcServer.addStreamInterceptors(grpcStream)
 	}
 }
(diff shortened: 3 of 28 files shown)


PiyushSingh-ZS (commented):
Reviewed this on top of #4167. The conjunctive-not-additive analysis is the most useful thing in the PR and I think it is correct. My main feedback is about packaging rather than correctness: there are four production bug fixes in here that are more urgent than the build tags and are currently blocked behind them.

### The four bug fixes should be their own PR, first

Listed under "Additional Information", all unrelated to build tags:

1. **`initTracer` built a `BatchSpanProcessor` over a nil exporter.** `getExporter` returns nil-and-no-error for an unsupported `TRACE_EXPORTER`, so a typo in config is enough, and the processor panics on its first flush on its own goroutine. That is a crash in production reachable from a one-character config mistake.
2. **`container.Error(err)` logged `{"message":{}}`** because the logger JSON-marshals the value and `errors.errorString` has no exported fields — every trace-exporter failure logged nothing usable. The note that the same pattern is likely elsewhere is right and deserves its own issue.
3. **Four exported gRPC methods dereferenced `App.grpcServer` without a nil check**, in a case `factory.go` explicitly treats as recoverable (it logs and continues when `newGRPCServer` fails, e.g. an out-of-range `GRPC_PORT`). Four panics in an already-handled error path.
4. **`TestUnifiedAuthenticationRegistration` could not fail** — `>= 2` when two are registered by default.

Each is small, independently reviewable and independently valuable. As things stand they can only land if maintainers also accept three build tags they may not have decided on yet, and #4168 is stacked on #4167 which is itself unresolved. (1) in particular looks like something worth getting in quickly on its own.

### `gofr_nodgraph` may not need to exist

The PR's own finding is that `pkg/gofr/migration` is the only dgo importer in the root module. Following that through, it is a single expression:

```go
// pkg/gofr/migration/dgraph.go
_, err = c.DGraph.Mutate(context.Background(), &api.Mutation{
	SetJson: jsonPayload,
})
```

One struct literal, in one place, is what pins `github.com/dgraph-io/dgo/v210` — and through it `google.golang.org/grpc` — into the root `go.mod` for **every GoFr user, including those who have never used Dgraph**. Meanwhile `pkg/gofr/datasource/dgraph` is already its own module and already depends on dgo.

Since `Mutate` takes `mu any` and type-asserts on the other side, the payload can cross that boundary without naming dgo's type at all — a `MutateJSON(ctx, []byte)` on the `DGraph` interface, or accepting `[]byte` in `Mutate` and constructing `api.Mutation` inside the datasource module where dgo already lives.

That removes the dependency for everyone with no build tag, no stub file, no CI matrix entry and no `migrator`-interface double-maintenance. It also seems strictly better than a tag worth 1.12 MB only to users who set it. Worth trying before committing to `gofr_nodgraph`.

(It touches the same call path as #4158, which is fixing `Mutate`'s commit semantics right now — those two want sequencing.)

### `gofr_nogrpc` is the one that needs a maintainer decision

I agree the API break is forced rather than a design choice: `RegisterService`, `AddGRPCServerOptions`, `AddGRPCUnaryInterceptors` and `AddGRPCServerStreamInterceptors` all name `grpc` types in their signatures, so a no-op stub is impossible and compiling them out is the only option. Scoping that CI step to `./pkg/...` because the gRPC examples are *supposed* to fail under the tag is the right call too.

So the honest framing is: this is a tag that changes GoFr's exported API, and on its own it is worth 0.67 MB. That is a real trade and I think it should be put to maintainers as its own question rather than sitting inside a 900-line diff with five other things.

### `grpcRunner` and the type-assertion guidance

CONTRIBUTING says, under take-interfaces-return-concrete-types:

> Be careful of type assertions in this context. If you take an interface and type assert to a type - then it's similar to taking a concrete type.

`App.grpcServer` becomes `grpcRunner`, and then several call sites assert straight back out:

```go
func (a *App) grpcSrv() *grpcServer {
	g, ok := a.grpcServer.(*grpcServer)
	...
}
```

`grpcRunner` genuinely abstracts only `Run` and `Shutdown`; everything else goes through the assertion. I think that is forced by the grpc types too — but a second, concrete, tag-guarded field declared only in `grpc.go` (`grpcConcrete *grpcServer`) would give the same linkage result with no assertions at all, and would sit better with the guideline.

### Smaller things

- The `auth.go` split is the right instinct — "HTTP auth is not gRPC's to take away" — and keeping `auth.go` untagged so HTTP auth works in every build is correct. But `addGRPCBasicAuth(users, validateFunc, validateFuncWithDatasources)` with exactly one of three non-nil at every call site is a triple-nil flag-argument shape. A small options struct in an untagged package would carry the same information without three `nil`s at each site.
- `addGRPCInterceptors` guarding with `!ok || g == nil` is correct and worth calling out: now that the field is an interface, `a.grpcServer != nil` no longer catches a typed nil. That is the same class of bug #4164 is fixing in the container, caught proactively here.
- `newGRPCRunner` returning the interface rather than `*grpcServer`, specifically so a nil result cannot reach `App` as a non-nil interface holding a nil pointer, with the nod to `container.isNil` — good.
- `disabledGRPC.Run` logs an error on a path the comment argues is unreachable (`grpcRegistered` is only set by `RegisterService`, which does not exist in this build). Defensive and harmless; no objection.

### CI

Noted that this targets `perf/optional-datasources-graphql` so the pipeline does not fire. Flagging it clearly in the description was the right thing to do — but it does mean nobody has a green signal on ~900 lines yet, which is another argument for pulling the bug fixes out where they can be tested on their own against `development`.

.github/workflows/go.yml
```diff
@@ -602,6 +602,40 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          go build -tags gofr_nootlp ./...
+          go vet   -tags gofr_nootlp ./pkg/gofr/ ./pkg/gofr/metrics/exporters/
+          go test  -tags gofr_nootlp -count=1 ./pkg/gofr/metrics/exporters/
+          go test  -tags gofr_nootlp -count=1 -run 'TestOTLPTracesOmitted|Test_initTracer' ./pkg/gofr/
+
       # The tags are opt-in, so the default build must stay exactly as it was.
       # This is the assertion that keeps that true.
       - name: Default build links the same packages as before
```
Umang01-hash, line 683: This step compiles but never compares the linked-package set to a baseline, so it doesn't assert what its name says. Suggest `go list -deps ./... | sort` diffed against a committed snapshot, or at least assert the grpc/otlp/dgo dep count is unchanged — that also catches an accidental default-build regression.

.github/workflows/go.yml
```diff
@@ -602,6 +602,40 @@ jobs:
           # The GraphQL stubs have their own tests, which only exist under the tag.
           go test -tags gofr_nographql -count=1 -run TestGraphQLDisabled ./pkg/gofr/
 
+      # gofr_nogrpc is scoped to ./pkg/... on purpose. Unlike the other tags it
+      # removes exported methods -- RegisterService and the AddGRPC* setters name
+      # grpc types and cannot exist without the import -- so the gRPC examples do
+      # not compile under it, which is the intended behaviour and not a failure.
+      - name: Build and test with the gRPC server omitted
```
Umang01-hash, line 613: Each new tag is built only in isolation here. A symbol that resolves under one tag but breaks under two would ship green. Worth adding a build of the all-3 combination (I verified it compiles today).

pkg/gofr/grpc.go
```diff
@@ -88,8 +102,42 @@ func (a *App) AddGRPCServerStreamInterceptors(interceptors ...grpc.StreamServerI
...
-	a.grpcServer.streamInterceptors = append(a.grpcServer.streamInterceptors, interceptors...)
+	g.addStreamInterceptors(interceptors...)
+}
+
+// grpcSrv returns the concrete server behind App's grpcRunner field, or nil when
+// this app has none -- newGRPCRunner failed, typically on an out-of-range
+// GRPC_PORT, and factory.go logs and continues. The exported AddGRPC* methods
+// dereferenced the field blind before it became an interface and panicked in
+// exactly that case; they now log and return.
+func (a *App) grpcSrv() *grpcServer {
```
Umang01-hash, line 120: The nil-guard here (and on RegisterService/AddGRPC*) is a genuine panic→log fix for the newGRPCRunner-failed path, but no test forces a nil runner and asserts the no-op. A ~10-line test would lock in the invariant so a future edit can't silently drop it.

Umang01-hash (changes requested):
Verified end-to-end locally at the head SHA in a clean worktree: the tag mechanism works, each tag drops its tree (dgo→0, OTLP exporters→0, grpc server −316 syms), the full tagset takes the binary 56.5MB→20.6MB with grpc fully gone, and the default build links the identical dependency trees as base (28=28, gofmt/vet/tests clean). Nice work — the typed-nil handling, fail-loud stubs, and boundary comments are excellent.

Requesting changes only on CI rigor — two gaps that would let a future regression ship green:

1. The "Default build links the same packages as before" step doesn't actually assert that; it only compiles. A `go list -deps` diff (or a grpc-count check) vs a baseline would make the step earn its name — and would guard the whole PR's value.
2. The three tags are only built in isolation. Combinations (esp. all-3) are never built. I confirmed they compile today, but nothing keeps that true.

Minor (non-blocking): the `grpcSrv()==nil` guard — a real panic→log fix — has no test; and the panic→log-and-return behavior change on `AddGRPC*`/`RegisterService` is worth a changelog line.

aryanmehrotra:
@Umang01-hash — all three threads addressed at `60d9207cc`, and the second one turned up a claim in this PR that was measurably wrong.

## 1. The all-tags combination is now built

You're right, and it's the failure mode tags are most prone to: a symbol that resolves under one tag and breaks under two ships green and fails only for the user who sets both. CI now builds, vets and tests **all six** together. Verified locally — clean, and 412 packages against the default 829.

Scoped to `./pkg/...` because `gofr_nogrpc` is: it removes exported methods, so the gRPC examples deliberately do not compile under it.

## 2. The package-diff — inherited from #4167

Same fix, same reasoning: `go list -deps ./... | sort` diffed against the merge base, failing on any change, with the baseline computed rather than committed so it cannot go stale.

## 3. The nil-runner no-op is tested

Agreed this needed locking in. `newGRPCRunner` fails on an out-of-range `GRPC_PORT` and `factory.go` logs and continues, so `App` runs on with no gRPC server — and these setters used to dereference the field blind, turning a config typo into a nil-pointer panic in the user's own setup code.

`TestGRPCSetters_NoServerIsANoOp` covers **all four** methods behind `grpcSrv()` — `AddGRPCServerOptions`, `AddGRPCUnaryInterceptors`, `AddGRPCServerStreamInterceptors`, `RegisterService` — since a later edit is as likely to reintroduce it in one of the others as in the one reported. Neutralising any guard makes it panic-fail.

## ⚠️ A claim in this PR was wrong, and I've corrected it

The description and the CI comment said `gofr_nootlp` "is the tag that actually releases `google.golang.org/grpc`". Measuring it against `gofr.dev/pkg/gofr`:

| Tags | `google.golang.org/grpc` packages |
|---|---|
| none | 82 |
| `gofr_nogrpc` | 82 |
| `gofr_nogrpc gofr_nootlp` | 81 |
| `gofr_nogrpc gofr_nootlp gofr_nodgraph` | 81 |
| **+ `gofr_nopubsub`** | **0** |

The OTLP exporters pin gRPC, but so does the Google Pub/Sub client through `cloud.google.com/go`. A shared dependency goes when its **last** importer does — so `gofr_nootlp` is necessary and not sufficient, and this PR's tags only pay off in combination with #4167's. That's a property of the tags worth documenting rather than a detail of this one, so it's in `docs/advanced-guide/slim-builds` with the table above.

## ⚠️ Worth knowing for whoever reviews this: CI barely runs here

This PR targets `perf/optional-datasources-graphql`, and `go.yml` triggers on `pull_request: branches: [main, development]` — so the 20-job matrix has **never run on it**. The only check is Snyk, which isn't branch-filtered. The green tick has been misleading since the PR opened.

Everything above was therefore verified locally: all six tags build, `go vet` and the tagged tests pass, and `golangci-lint` with all six tags reports 0 new issues. Real CI arrives the moment #4167 merges and this retargets to `development`; keeping it stacked until then keeps the diff reviewable.

.github/workflows/go.yml
aryanmehrotra: Done, and you were right that the step did not assert what its name said.

`.github/workflows/go.yml:684`, step **"Default build links the same packages as before"** now does the `go list -deps ./... | sort` diff you suggested:

```
go list -deps ./... | sort > /tmp/deps.head
git worktree add /tmp/base "$(git merge-base HEAD origin/${{ github.base_ref || 'development' }})"
(cd /tmp/base && go list -deps ./... | sort) > /tmp/deps.base
diff -u /tmp/deps.base /tmp/deps.head || { ...; exit 1; }
```

Baseline is the merge base rather than a committed snapshot, so it cannot go stale and needs no upkeep — that was the one part of your suggestion I deliberately did not take, and the reason is that a committed `deps.txt` has to be regenerated on every legitimate dependency change and silently rots when someone forgets. `fetch-depth: 0` was added to the job's checkout so the merge base resolves.

**⚠️ One thing you should know before you read the rest of this: no CI ran on this PR.** `.github/workflows/go.yml` triggers on `pull_request: branches: [main, development]` and this PR's base is `perf/optional-datasources-graphql`, so GitHub has run exactly **1 check** (Snyk) on it, ever. **The step above has never executed on a runner.** Everything below is a local run on darwin/arm64, go1.26.3.

I ran the step's logic by hand at `ac84f219c`:

- `go list -deps ./... | sort` at HEAD vs at the merge base `4e2121577` is **byte-identical, 861 packages**.
- Negative test, because a diff that never fires proves nothing: adding `_ "net/http/cgi"` to `pkg/gofr/gofr.go` makes the diff report `> net/http/cgi` and the step exits 1. So it catches the accidental default-build regression you were pointing at.

On the "assert the grpc/otlp/dgo dep count is unchanged" half — the full package diff is a superset of that, so I did not add a second count assertion. The counts themselves are in the PR description now, and measuring them turned up a real error in the docs this PR adds, fixed in `ac84f219c`: the composition table in `docs/advanced-guide/slim-builds/page.md` opened with **82** grpc packages linked by default, which is wrong — it is **86**. Against `gofr.dev/pkg/gofr`:

| tags | `google.golang.org/grpc` pkgs |
|---|---|
| none | 86 |
| `gofr_nogrpc` | 82 |
| `+ gofr_nootlp` | 81 |
| `+ gofr_nodgraph` | 81 |
| `+ gofr_nopubsub` | **0** |

With the old 82/82 first row the table read as `gofr_nogrpc` removing nothing at all, which overstated the very point it was making.

.github/workflows/go.yml
aryanmehrotra: Added, and taken further than the all-3 you asked for.

`.github/workflows/go.yml:653`, step **"Build with every optional subsystem omitted at once"**, builds all **six** tags together — the three from #4167 as well as the three here, since the same argument applies across the stack boundary:

```
ALL="gofr_nosqldrivers gofr_nopubsub gofr_nographql gofr_nogrpc gofr_nodgraph gofr_nootlp"
go build -tags "$ALL" ./pkg/...
go vet   -tags "$ALL" ./pkg/gofr/ ./pkg/gofr/container/ ./pkg/gofr/migration/
go test  -tags "$ALL" -count=1 ./pkg/gofr/container/ ./pkg/gofr/datasource/sql/ ./pkg/gofr/migration/
```

It is `./pkg/...` rather than `./...` for the same reason the `gofr_nogrpc` step is: that tag compiles out `RegisterService`, so the two gRPC example servers fail with `*gofr.App does not implement grpc.ServiceRegistrar`, which is the intended behaviour.

The tagged lint at `.github/workflows/go.yml:665` runs with the same six-tag set, so the combination is linted as well as built — the tagged halves are otherwise invisible to the default lint job.

**⚠️ Worth stating plainly: this step has never run on a runner.** `.github/workflows/go.yml` triggers on `pull_request: branches: [main, development]` and this PR's base is `perf/optional-datasources-graphql`, so only 1 check (Snyk) has ever run on this PR. Everything below is local, at `ac84f219c`, darwin/arm64, go1.26.3:

| configuration | build | vet | test |
|---|---|---|---|
| untagged | `./...` ✅ | `./...` ✅ | `./pkg/gofr/...` ✅ |
| `gofr_nosqldrivers` | `./...` ✅ | ✅ | ✅ |
| `gofr_nopubsub` | `./...` ✅ | ✅ | ✅ |
| `gofr_nographql` | `./...` ✅ | ✅ | ✅ |
| `gofr_nogrpc` | `./pkg/...` ✅ | ✅ | ✅ |
| `gofr_nodgraph` | `./...` ✅ | ✅ | ✅ |
| `gofr_nootlp` | `./...` ✅ | ✅ | ✅ |
| **all six** | `./pkg/...` ✅ | ✅ | ✅ |

`golangci-lint run ./pkg/gofr/...` reports **0 issues in the 28 Go files this PR touches**, both untagged and with all six `--build-tags`. (The repo-wide counts — 72 untagged, 67 tagged — are pre-existing, which is what `only-new-issues: true` is there for.)

pkg/gofr/grpc.go
aryanmehrotra: Test added: `TestGRPCSetters_NoServerIsANoOp`, `pkg/gofr/grpc_test.go:591`.

It forces exactly the state you named — `GRPC_PORT=99999` makes `newGRPCRunner` return `invalid port number: 99999`, `factory.go` logs and continues, and `App` is left with no server — then asserts the no-op on all four methods that share the guard (`grpc.go:64`, `:89`, `:105`, `:295`):

```
require.Nil(t, app.grpcSrv(), "the premise: this app has no gRPC server to configure")
// subtests: AddGRPCServerOptions, AddGRPCUnaryInterceptors,
//           AddGRPCServerStreamInterceptors, RegisterService
assert.NotPanics(t, call, "%s must log and return when there is no server", name)
```

All four rather than just the one reported, because a later edit is as likely to drop the guard in any of them.

**I checked that the assertion is load-bearing rather than just green.** Deleting the four `if g == nil { ...; return }` blocks from `grpc.go` and re-running:

```
--- FAIL: TestGRPCSetters_NoServerIsANoOp
    --- FAIL: TestGRPCSetters_NoServerIsANoOp/RegisterService
    --- FAIL: TestGRPCSetters_NoServerIsANoOp/AddGRPCServerOptions
    --- FAIL: TestGRPCSetters_NoServerIsANoOp/AddGRPCUnaryInterceptors
    --- FAIL: TestGRPCSetters_NoServerIsANoOp/AddGRPCServerStreamInterceptors
```

— each one a nil-pointer panic through `grpc.go:174`. So the invariant is actually pinned; a future edit that silently drops a guard fails here.

Passing run at `ac84f219c`:

```
$ go test ./pkg/gofr/ -run TestGRPCSetters_NoServerIsANoOp -v -count=1
ERROR  failed to create gRPC server: invalid port number: 99999
no gRPC server to register service test.Service on
no gRPC server to add options to
no gRPC server to add unary interceptors to
no gRPC server to add stream interceptors to
--- PASS: TestGRPCSetters_NoServerIsANoOp (0.00s)  [4/4 subtests PASS]
ok  gofr.dev/pkg/gofr 1.145s
```

Under `-tags gofr_nogrpc` the file is excluded and the run reports `no tests to run`, which is correct — the four methods do not exist under that tag.

**⚠️ For the record on all three threads: no CI ran on this PR.** `.github/workflows/go.yml` triggers on `pull_request: branches: [main, development]`, this PR's base is `perf/optional-datasources-graphql`, and GitHub has run exactly 1 check (Snyk) on it. Every result quoted in these replies is a local run on darwin/arm64, go1.26.3 — please do not read them as CI-covered. The full pipeline fires once this is retargeted to `development` after #4167 merges.

akshat-kumar-singhal (changes requested):
Request changes — mainly the rebase. Reviewed as the delta over #4167 at 4e21215 (this PR lacks #4167's last three commits). `./pkg/...` builds and vets for every tag combination; default `go list -deps ./...` matches its merge base exactly; moved tests (`TestUnifiedAuthenticationRegistration`, `TestStartGRPCServer_Registered`, OTLP `Test_initTracer`) are all still run. The grpc package table matches measurement exactly (86 / 82 / 81 / 81 / 0).

**Blocking: conflicts with development, and the OTLP-trace half targets code that moved.**
`git merge-tree origin/development pr/4168` conflicts in `pkg/gofr/otel.go`, `otel_test.go` and `grpc_test.go`. #4206/#4207/#4247/#4249 moved the OTLP trace exporter into `pkg/gofr/traces/exporters/otlp.go`, which imports `otlptracegrpc` and self-registers. `pkg/gofr/otlp_trace.go`/`otlp_trace_disabled.go` split code out of an `otel.go` that no longer holds it — resolved mechanically, `-tags gofr_nootlp` still links `otlptracegrpc` and its grpc tree via `traces/exporters`, silently removing only the metrics half. Put the `gofr_nootlp` constraint on `traces/exporters/otlp.go` with a stub that registers "otlp"/"jaeger" and returns the tag error, then re-measure.

**Should-fix: CI never checks the three new tags remove anything.** Without 8803000 the "Each tag removes the packages it claims to" step doesn't cover `gofr_nogrpc`/`gofr_nodgraph`/`gofr_nootlp`. After rebasing onto #4167, add checks for `go.opentelemetry.io/otel/exporters/otlp/`, `github.com/dgraph-io/dgo` and `google.golang.org/grpc` under the combined tags.

**Should-fix (doc): "The tags remove implementations, never API"** — [`page.md:70-72`](https://github.com/gofr-dev/gofr/blob/ac84f219c9621bafe5b433bd50fbd90944fc4b1e/docs/advanced-guide/slim-builds/page.md#L70-L72)
`gofr_nogrpc` removes `RegisterService` and `AddGRPC*`. `go build -tags gofr_nogrpc ./...` fails on `examples/grpc/grpc-{unary,streaming}-server/server/health_gofr.go:37` (`*gofr.App does not implement grpc.ServiceRegistrar (missing method RegisterService)`), as will any gofr-cli-generated `health_gofr.go` — so the doc's own all-six command at [L20](https://github.com/gofr-dev/gofr/blob/ac84f219c9621bafe5b433bd50fbd90944fc4b1e/docs/advanced-guide/slim-builds/page.md#L20) fails in this repo. Call out `gofr_nogrpc` as the exception.

**Should-fix (doc): wrong settings for `gofr_nootlp`** — [`page.md:34`](https://github.com/gofr-dev/gofr/blob/ac84f219c9621bafe5b433bd50fbd90944fc4b1e/docs/advanced-guide/slim-builds/page.md#L34): GoFr never reads `OTEL_EXPORTER_OTLP_ENDPOINT`, and `TRACER_URL` still works for zipkin/gofr. The affected settings are `TRACE_EXPORTER=otlp|jaeger` and `METRICS_EXPORTER=otlp`.

**Should-fix (doc): `gofr_nodgraph` is broader than "Dgraph migrations".** `migration.go` chains the Dgraph migrator whenever `c.DGraph` is set; the stub's `checkAndCreateMigrationTable` errors and [`migration.go:82`](https://github.com/gofr-dev/gofr/blob/ac84f219c9621bafe5b433bd50fbd90944fc4b1e/pkg/gofr/migration/migration.go#L82) calls `c.Fatalf`. A service with a Dgraph datasource and only SQL migrations exits at `app.Migrate` under this tag. Exiting loudly is defensible; the doc should say so.

**Nit** — [`otlp_trace_disabled.go:22`](https://github.com/gofr-dev/gofr/blob/ac84f219c9621bafe5b433bd50fbd90944fc4b1e/pkg/gofr/otlp_trace_disabled.go#L22): says "TRACE_EXPORTER=otlp is unavailable" even for `TRACE_EXPORTER=jaeger`, which takes the same path.

**Nit (PR description):** default-build behavior changes worth mentioning, both improvements: `AddGRPC*`/`RegisterService` now log and return when `newGRPCServer` failed (e.g. bad `GRPC_PORT`) instead of panicking on a nil `*grpcServer`; `initTracer` skips the BatchSpanProcessor for a nil exporter (fixes a panic on an unsupported `TRACE_EXPORTER`) and switches `Error(err)` to `Errorf`, changing the log format.

Total counts are ~20 above my linux/amd64 go1.26.0 measurements (e.g. none 809 vs 829; all six 415 vs 412) — the same platform offset as #4167. `gofr_nodgraph` alone removes 2 packages, but combined it matters: `nogrpc,nootlp,nopubsub` without it still links 64 grpc packages (dgo protos pin grpc); with it, 0.

aryanmehrotra:
Thanks @akshat-kumar-singhal — the blocking finding was correct and the measurement is unambiguous. Rebased onto the updated #4167 (`f1634f258`); head is `238cf65b1`, and the PR is `CLEAN` rather than conflicting.

### `gofr_nootlp` was removing nothing from the trace side

You called this exactly: #4206/#4207/#4247/#4249 moved the OTLP trace exporter to `pkg/gofr/traces/exporters/otlp.go`, where it self-registers and imports `otlptracegrpc`, and this PR's `gofr_nootlp` was splitting code out of an `otel.go` that no longer holds it.

```
otlptrace packages linked by gofr.dev/pkg/gofr

               before      after
  default        10          10
  gofr_nootlp    10  ←        0
```

The constraint now sits on `traces/exporters/otlp.go`, with `otlp_disabled.go` doing what you suggested: registering **"otlp" and "jaeger"** to a builder that fails naming the tag.

Registering the names rather than omitting them is the part worth arguing for. `Build`'s unknown-exporter path would report `TRACE_EXPORTER=otlp` as a name it does not recognize — which reads like a typo and sends an operator to check their spelling. The stub says the true thing: the name is right, this binary does not carry it. `Build`'s nil-exporter path then installs the `NeverSample` provider, so the service still starts and correlation IDs keep working.

The obsolete `pkg/gofr/otlp_trace*.go` are dropped. The metrics half was unaffected by the move and is unchanged.

Also, with this fixed, your composition point lands properly:

```
google.golang.org/grpc packages:  86 default  →  0 with nogrpc+nootlp+nodgraph+nopubsub
```

### CI now checks the three new tags

`gofr_nootlp` and `gofr_nodgraph` get their own patterns. **`gofr_nogrpc` is checked against the combined tag set, not alone** — grpc is pinned by the OTLP exporters, the Dgraph protos and `cloud.google.com/go` as well as by the server, so on its own it takes 4 of 86 and a bare "still linked" assertion would fail for reasons that have nothing to do with the tag working. That reasoning is in a comment next to the check.

All six run green locally:

| tag | packages |
|---|---|
| `gofr_nosqldrivers` | 828 → 784 |
| `gofr_nopubsub` | 828 → 616 |
| `gofr_nographql` | 828 → 808 |
| `gofr_nootlp` | 828 → 786 |
| `gofr_nodgraph` | 828 → 826 |
| `nogrpc,nootlp,nodgraph,nopubsub` | 828 → 475 |

### All three doc corrections

- **`gofr_nogrpc` removed from the all-six build command**, because `./...` genuinely cannot build with it in this repo — the doc's own command was failing. The exception is now stated where the page used to claim the tags never remove API.
- **`gofr_nootlp`'s settings** are `TRACE_EXPORTER=otlp|jaeger` and `METRICS_EXPORTER=otlp`. You were right that GoFr never reads `OTEL_EXPORTER_OTLP_ENDPOINT` and that `TRACER_URL` still works for zipkin and gofr.
- **`gofr_nodgraph` is broader than "Dgraph migrations"** — it has its own section now, saying plainly that a service with a Dgraph datasource and only SQL migrations exits at `app.Migrate`, and why failing loudly is the right call.

Counts re-measured on **darwin/arm64, Go 1.26.3** (stated in the doc, per your platform note): 828 default, 786 `nootlp`, 784 `nosqldrivers`, 808 `nographql`, 818 `nogrpc`, 826 `nodgraph`, **411 with all six**.

### One more tagged-build break the rebase surfaced

`redact_test.go`'s four OTLP/jaeger cases assert what the *real* builder logs, so they fail once the stub is what is registered. They take an `otlpTraceLinked` skip, mirroring `container.pubsubBackendsLinked`. That is the third instance in this stack of an untagged test not surviving a tag — worth noting as the ongoing maintenance cost @PiyushSingh-ZS predicted.

### Verification

Build and vet clean for default, each tag alone, and all six together. Tests pass untagged and under all six (`container`, `datasource/sql`, `migration`, `traces/exporters`, `metrics/exporters`). `typos`, `gofmt` and `golangci-lint` clean.

**Base left at `perf/optional-datasources-graphql`** deliberately. I briefly retargeted it to `development` to get CI on it and reverted immediately — stacked as it is, that makes the PR show #4167's diff as well. It still means this has had almost no CI, which is the one thing about it I cannot fix from here; everything above is local.

@PiyushSingh-ZS — your mechanism question and the "four bug fixes should ship first" point are both still open and both still stand. Nothing here answers them; this only makes the tag do what it claims.

PR #4208, opened 2026-09-14
fix(ci): pull MinIO from quay.io — docker.io/minio/minio is no longer public

akshat-kumar-singhal:
**Description:**

`Example Unit Testing` is currently failing on **every open PR**, on all three Go versions, at the `Start MinIO` step:

```
docker: Error response from daemon: pull access denied for minio/minio,
repository does not exist or may require 'docker login': denied: requested
access to the resource is denied
```

It fails on PRs that touch no example and no workflow, so it is not the PRs — Docker Hub's `minio/minio` repository is no longer publicly pullable.

Checked anonymously rather than inferred:

| image | result |
|---|---|
| `alpine:3.20` (control) | resolves |
| `docker.io/minio/minio:latest` | `denied: requested access to the resource is denied` |
| `docker.io/minio/minio@sha256:14cea493…` (the pin in `go.yml`) | `denied: …` |
| `quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z` | resolves |

The last green PR pipeline on this repo was 2026-09-11, so it broke after that date.

**The fix is a registry prefix and nothing else.** `quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z` carries the *identical* manifest list — `sha256:14cea493d9a34af32f524e538b8346cf79f3321eff8e708c1e2960462bd8936e`, byte-for-byte the digest `go.yml` already pins:

```
$ docker buildx imagetools inspect quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z
Name:      quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z
MediaType: application/vnd.docker.distribution.manifest.list.v2+json
Digest:    sha256:14cea493d9a34af32f524e538b8346cf79f3321eff8e708c1e2960462bd8936e
```

So the image CI runs is provably the one it ran on 2026-09-11, and the digest pin that protects the integration guard for #3804 stays intact. Bumping to a newer MinIO release instead would have changed two things at once and weakened that pin's purpose.

`examples/using-s3-filestore/README.md` told readers to `docker run … minio/minio server /data`, which is broken for the same reason, so it moves in the same change.

None. Same image, different registry; no MinIO version change, no test changes.

**Additional Information:**

- This is deliberately a standalone CI fix rather than part of any feature branch — it unblocks every open PR in the repo and should be reviewable and mergeable on its own.
- The comment above the `docker run` now records *why* the registry is quay.io, so a future reader doesn't "helpfully" move it back to Docker Hub.

**Checklist:**


aryanmehrotra (approved):
Approving. I re-verified the claim independently against the live registries rather than taking the description's word for it:

| check | result |
|---|---|
| `docker.io/minio/minio:RELEASE.2025-09-07T16-13-09Z` (anon) | **HTTP 401** — denied |
| `docker.io/library/alpine:3.20` (control) | HTTP 200 — Docker Hub itself is healthy |
| `quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z` (anon) | **HTTP 200** |
| quay.io `docker-content-digest` | `sha256:14cea493d9a34af32f524e538b8346cf79f3321eff8e708c1e2960462bd8936e` |
| digest already pinned in `go.yml` | `sha256:14cea493…` — **identical** |

So this is provably a registry change and not an image change, and the digest pin protecting the #3804 integration guard is unaffected. Keeping the pin rather than bumping to a newer MinIO release was the right call — a version bump would have changed two things at once and made the pin meaningless.

Also confirmed:
- `grep -rE '(^|[^./a-zA-Z0-9-])minio/minio'` over the tree returns exactly the two occurrences this PR changes — no bare reference is left behind.
- CI on this PR is green on all three Go versions, i.e. `Example Unit Testing` actually reaches `go test ./examples/...` again.

For context on why this is worth merging ahead of anything else: the green ticks on the other open PRs are **stale, not healthy**. Last `Example Unit Testing (v1.26)` run per PR:

| PR | last run | result |
|---|---|---|
| #4205 | 2026-09-14 04:40 | FAILURE |
| #4203 | 2026-09-11 15:37 | success (pre-break) |
| #4201 | 2026-09-11 11:11 | success (pre-break) |
| #4170 | 2026-09-10 18:29 | success (pre-break) |

Every one of those turns red on its next push. This unblocks all 31 open PRs, not the three currently showing red.

**Nit, non-blocking:** please strip the `🤖 Generated with [Claude Code]` footer and the `claude.ai/code/session_…` link from the PR description before merge.

---

**Follow-up, deliberately not in this PR.** #3868 already predicted this failure class — *"it works today; the day it stops working, every example test that touches pub/sub fails at once and the cause will not be obvious from the logs"* — and that is exactly what happened here, just with MinIO instead of Kafka. #3868 covers **pinning**. What no issue covers yet is **detection**: there is nothing that tells us a required external dependency has become unpullable, so this broke on 2026-09-11 and was found on 2026-09-14 only because a contributor pushed and read a red X as their own fault. I'll open a separate issue for the preflight-job + scheduled-canary side of it.

PR #4244, opened 2026-09-16
chore(deps): consolidate minor dependency updates (2026-09-16)

Umang01-hash:
## Consolidated Dependabot Updates (Minor/Patch)

Combines this week's 33 Dependabot minor/patch PRs into one, applied across the root module and the `go.work` datasource submodules (each has its own `go.mod`). Attribution reflects the **actual** `go.mod` changes after per-module `go mod tidy`, not the raw Dependabot targets.

### Updates included
| Package | Old | New | Modules changed |
|---------|-----|-----|-----------------|
| go.opentelemetry.io/otel + otel/trace | 1.45.0 | 1.46.0 | 17 submodules: arangodb, cassandra, clickhouse, couchbase, dbresolver, dgraph, elasticsearch, mongo, opentsdb, oracle, scylladb, solr, surrealdb, kv-store/badger, kv-store/dynamodb, kv-store/nats, pubsub/eventhub (root already on 1.46) |
| go.opentelemetry.io/otel/metric (indirect; no code change 1.45→1.46) | 1.45.0 | 1.46.0 | 13 submodules: arangodb, cassandra, couchbase, dbresolver, dgraph, elasticsearch, kv-store/badger, mongo, opentsdb, oracle, pubsub/eventhub, solr, surrealdb |
| github.com/aws/aws-sdk-go-v2/service/sqs (pulls aws-sdk-go-v2 1.46.0 → 1.47.0) | 1.51.0 | 1.52.0 | pubsub/sqs |
| github.com/jackc/pgx/v5 | 5.10.0 | 5.11.0 | cloudsql |
| github.com/surrealdb/surrealdb.go | 1.6.0 | 1.7.0 | surrealdb |
| golang.org/x/oauth2 | 0.36.0 | 0.37.0 | root, file/gcs, metrics/exporters/gcp |
| google.golang.org/api | 0.291.0 | 0.297.0 | root, file/gcs |
| github.com/vektah/gqlparser/v2 | 2.5.36 | 2.5.37 | root |
| golang.org/x/text | 0.41.0 | 0.42.0 | root |
| golang.org/x/term | 0.45.0 | 0.46.0 | root |
| golang.org/x/time | 0.15.0 | 0.16.0 | root |

Root indirect bumps pulled by the above: cloud.google.com/go/auth 0.22.0→0.23.2, googleapis/enterprise-certificate-proxy 0.3.19→0.3.20, googleapis/gax-go/v2 2.23.0→2.24.0, golang.org/x/sys 0.47.0→0.48.0.

### Example modules re-tidied (replace-directives)
`GOWORK=off go mod tidy` in the two example modules that `replace` a bumped submodule, so they build standalone (CI builds examples in workspace mode and cannot see this drift):
- `examples/using-cloudsql` — pgx/v5 5.10.0→5.11.0 indirect (regression this PR introduced; caught by @aryanmehrotra).
- `examples/using-gcp-metrics` — absorbs this PR's oauth2 propagation **and** clears pre-existing drift on `development` (detectors/gcp, otel sdk/metric, modernc).

### One behavior change to be aware of (not an API break, but user-visible)
`gqlparser/v2` 2.5.37 adds `validateInputObjectCircularRefs`, which runs inside `LoadSchema` (used at `pkg/gofr/graphql.go`). A schema with a **non-null input-object cycle** — always invalid per the GraphQL Oct-2021 spec but previously accepted by 2.5.36 — now fails at app startup. Upstream closing a spec gap; flagged here so it's findable if a GraphQL app stops booting after this bump.

### Verified locally before raising
- Per-module `go mod tidy` clean; no `go`/`toolchain` directive changes in any module.
- `go build ./...` green for the root module and all 21 changed submodules (incl. cgo `oracle`); `GOWORK=off go build ./...` green for both re-tidied example modules.
- Root `pkg/gofr/... -short` passes; key submodule `-short` suites pass.
- `otel` + `otel/trace` + `otel/metric` move together to 1.46.0 — no version split.
- Changeset is dependency-files-only; no source/binary files touched.

All library bumps are minor/patch — no major version changes, no import-path changes.

### Original Dependabot PRs (closed in favour of this one)
#4211 #4212 #4213 #4214 #4215 #4216 #4217 #4218 #4219 #4220 #4221 #4222 #4223 #4224 #4225 #4226 #4227 #4228 #4229 #4230 #4231 #4232 #4233 #4234 #4235 #4236 #4237 #4238 #4239 #4240 #4241 #4242 #4243


pkg/gofr/datasource/cloudsql/go.mod
```diff
@@ -5,7 +5,7 @@ go 1.26.0
 require (
 	cloud.google.com/go/cloudsqlconn v1.25.1
 	github.com/go-sql-driver/mysql v1.10.0
-	github.com/jackc/pgx/v5 v5.10.0
+	github.com/jackc/pgx/v5 v5.11.0
```
aryanmehrotra, line 8: This bump is correct, but it leaves `examples/using-cloudsql/go.mod` behind — that module `replace`s this one and still pins `pgx/v5 v5.10.0 // indirect`, so `GOWORK=off go build ./...` fails there. It builds clean on `development`. One `go mod tidy` in the example fixes it. Details in the review comment.

go.mod
```diff
@@ -27,7 +27,7 @@ require (
 	github.com/redis/go-redis/v9 v9.22.0
 	github.com/segmentio/kafka-go v0.4.51
 	github.com/stretchr/testify v1.12.1
-	github.com/vektah/gqlparser/v2 v2.5.36
+	github.com/vektah/gqlparser/v2 v2.5.37
```
aryanmehrotra, line 30: **`gqlparser` 2.5.37 adds a schema rule that can reject a schema 2.5.36 accepted** — worth calling out in the description, since the body currently says "no API breakage observed."

`validator/schema.go:192` adds `validateInputObjectCircularRefs`, which runs inside `gqlparser.LoadSchema` — the call at `pkg/gofr/graphql.go:131`. Verified against both pinned versions:

```
input Node { name: String!  child: Node! }

v2.5.36 -> LoadSchema OK, types: 15
v2.5.37 -> LoadSchema ERROR: Cannot reference Input Object "Node" within
           itself through a series of non-null fields: "child".
```

A GoFr app with a non-null input-object cycle in its schema now **fails at startup** instead of serving. The schema was always invalid per the GraphQL Oct2021 spec, so this is upstream closing a gap, not a bug — but it's a user-visible change and not something a reader would expect from a `.36 -> .37` patch bump.

Suggestion: soften "no API breakage observed" and add a line noting this, so it's findable when someone reports their GraphQL app stopped booting.

**Unrelated nit while here:** `go.opentelemetry.io/otel/metric` also moved 1.45.0 -> 1.46.0 in 13 submodules (arangodb, cassandra, couchbase, dbresolver, dgraph, elasticsearch, kv-store/badger, mongo, opentsdb, oracle, pubsub/eventhub, solr, surrealdb), but the table only names `otel` and `otel/trace`. Zero code changed in that module between the two versions, so it's purely a completeness fix to the table.

aryanmehrotra (commented):
**`examples/using-cloudsql` no longer builds outside the workspace**

This is a regression introduced by this PR, and CI structurally can't see it.

`examples/using-cloudsql` has `replace gofr.dev/pkg/gofr/datasource/cloudsql => ../../pkg/gofr/datasource/cloudsql`. That module now requires `pgx/v5 v5.11.0`, but the example's own `go.mod` still pins `v5.10.0 // indirect`.

```
$ GOWORK=off go build ./...   # in examples/using-cloudsql

development (851b78a77) -> clean
this PR     (b3227517e) -> go: updates to go.mod needed; to update it:
                               go mod tidy
```

CI builds examples in workspace mode, where MVS resolves `pgx` to `5.11.0` from the sibling module, so all 38 checks stay green. Anyone building the example directory on its own hits the error.

Fix:

```sh
cd examples/using-cloudsql && GOWORK=off go mod tidy
```

which produces exactly one `go.mod` line plus the matching `go.sum` entries:

```diff
-	github.com/jackc/pgx/v5 v5.10.0 // indirect
+	github.com/jackc/pgx/v5 v5.11.0 // indirect
```

**Separately — worth deciding on, not blocking this PR:** `examples/using-gcp-metrics` is *already* stale on `development` for the same reason (`cloud.google.com/go/auth`, `gax-go/v2`, `enterprise-certificate-proxy`, `detectors/gcp 1.33.0 -> 1.36.0`). So this isn't a one-off slip — the consolidation workflow has a blind spot for the example modules that carry `replace` directives. Either tidy them as part of the weekly pass, or add a CI job that fails on drift.

<details><summary>Everything else I checked — all clean</summary>

| Check | Result |
|---|---|
| `GOWORK=off go build ./...`, 32 modules | 30 OK, 2 fail (both examples; `using-gcp-metrics` pre-existing) |
| `GOWORK=off go test -short`, 27 suites | **27/27 pass** |
| Workspace `go build ./...` + `go vet ./pkg/gofr/...` | clean |
| `go mod tidy` drift, 32 modules | 30 tidy, 2 stale (above) |
| `go work sync` correctly skipped | confirmed — running it would raise every module to workspace max |
| `govulncheck ./...` | **0 module vulnerabilities** |
| Dependency-files-only changeset | confirmed, 0 source files |
| No `go`/`toolchain` directive drift | confirmed |
| otel API/SDK skew within each module | none — and moot, `otel/trace` + `otel/metric` are byte-identical across 1.45/1.46 |

Local failures that are **not** this PR: `file/ftp` (flaky, passes on rerun, `go.mod` untouched); `examples/using-migrations`, `http-server`, `using-subscriber`, `pkg/gofr` MQTT timeout — all need live infra, and `using-migrations` fails identically on `development`.

Bumps I verified against the pinned source and found genuinely clean, so I haven't commented on them: `otel/trace`, `otel/metric` and `x/time` are byte-identical; `otel` 1.46 is only the `attribute/set.go` nil-receiver `Filter` fix plus a `Hasher` refactor with the same digest; `google.golang.org/api` 0.291->0.297 has zero changes in `googleapi`/`iterator`/`option`, the only three packages GoFr imports; `surrealdb.go` 1.7.0 has a byte-identical `db.go` and only renames `spectron/` to `memory/`, neither imported; `pgx` 5.11 has a long changelog but all three risky paths miss GoFr (no `pgx.Rows` mocks, DSN is keyword/value not URL, `user=` always supplied); `x/oauth2` 0.37 is a one-character GCE metadata key fix; `aws-sdk-go-v2` 1.47 moves the `Amz-Sdk-Request` header between middlewares.
</details>

Umang01-hash:
Thanks @aryanmehrotra — reproduced and fixed in 29d932d.

- **using-cloudsql**: confirmed `GOWORK=off go build ./...` failed at head, clean on development. Tidied → single line `pgx/v5 5.10.0→5.11.0 // indirect`; builds standalone now.
- **using-gcp-metrics**: tidied too — it also `replace`s a module this PR bumps (oauth2 in the gcp exporter), and the full tidy additionally clears the pre-existing `development` drift you noted (detectors/gcp, otel sdk/metric, modernc). Both example diffs are indirect-only.
- **gqlparser 2.5.37**: good catch — softened "no API breakage" in the description and added a note about the `validateInputObjectCircularRefs` startup-reject for non-null input-object cycles.
- **otel/metric**: added to the table (13 submodules, no code change 1.45→1.46).

On the blind spot: the weekly consolidation now tidies example modules with `replace` directives via `GOWORK=off`. A CI job that fails on `GOWORK=off` drift across all modules would be the durable fix — happy to raise that separately.

pkg/gofr/datasource/cloudsql/go.mod
Umang01-hash: Fixed in 29d932d. Reproduced: `GOWORK=off go build ./...` failed at head, clean on development. Tidied → single `pgx/v5 5.10.0→5.11.0 // indirect` line, builds standalone now. Also did the same for `using-gcp-metrics` (it replaces the gcp exporter this PR bumps).

go.mod
Umang01-hash: Good catch — softened "no API breakage observed" in the description and added a note: gqlparser 2.5.37's `validateInputObjectCircularRefs` makes a non-null input-object cycle fail at `LoadSchema` (app startup) instead of serving. Also added the `otel/metric` 1.45→1.46 row (13 submodules, no code change between the two) to the table.

aryanmehrotra (approved):
All three points from my earlier review are addressed, re-verified from a clean worktree at `698a943c5` (base `851b78a77`, current `development` head).

| Check | Result |
|---|---|
| `GOWORK=off go build ./...`, 32 modules | 32/32 |
| `go mod tidy` drift | 32/32 tidy |
| Workspace `go build ./...` + `go vet ./pkg/gofr/...` | clean |
| Submodule `go test -short` | 27/27 pass |
| Root `./pkg/...` `go test -short` | 30/31 — `logging/remotelogger` failed once under full-parallel load (9.3s vs ~0.8s), then passed 11/11 alone on this head and 5/5 on `development`; pre-existing timing flake |
| `govulncheck` | 0 module vulnerabilities |
| CI | 38 pass |

- `examples/using-cloudsql` now builds standalone (`pgx` 5.11.0).
- `examples/using-gcp-metrics` tidy: all 9 extra indirect moves (`detectors/gcp`, `x/net`, `x/sync`, `grpc`, `modernc.*`, …) exactly match versions `development` already pins elsewhere, so no new module versions enter the tree.
- Description now calls out the gqlparser 2.5.37 input-object-cycle rule and lists `otel/metric`.

Commit 1 (`64a98e6fb`) is content-identical to the `b3227517e` I reviewed, so the upstream source checks (otel, pgx, gqlparser, surrealdb, google-api, oauth2, aws, x/*) carry forward. LGTM.

PR #116, opened 2023-12-11
Add logs to processErrors() for InternalServerErrors

vipul-rawat:

Issue #114, opened 2023-12-05
DB error does not get logged


vikash:
If any handler returns an error which results into 500 error from handler, the error is not getting logged at any level. 

For example, if handler has following as return statement: 
`return h.store.Create(ctx, cust)`

and the Create function has an error returned, it is expected to get it logged.

vipul-rawat:
@vikash do we need an error log or a debug? Error log prints the stack trace but that isn't all helpful for the user as the trace would consist entirely of gofr calls!

PR #4247, opened 2026-09-16
fix(tracing): redact tracer config in logs

aryanmehrotra:
**Description:**

The v1.61.0 release PR (#4246) fails its **CodeQL** check with 6 high-severity `go/clear-text-logging` alerts (#179–#184), all in `pkg/gofr/otel.go`. This PR clears them. The CodeQL workflow that scans PRs from forks is split out into #4248, because it needs a repo settings change before it can merge.

### Tracer config in logs

CodeQL traces the taint back to `dbresolver`'s `Get("DB_PASSWORD")`. That value comes back through the shared `config.Config` interface, so CodeQL treats **every** `Config.Get` result as a password. Most alerts of this kind in the repo have been dismissed as false positives, and 3 of these 6 are the same log lines as alerts #114–#116, which were dismissed before #4205 moved them.

**One of the risks is real, though.** Before #4205, `TRACER_URL` could only be `host:port`. Now `https://user:token@collector:4317` is valid, and `https://zipkin/api/v2/spans?api-key=…` always was. Every tracer startup log printed the URL as-is.

| Log line | Before | After |
|---|---|---|
| `Exporting traces to <exporter> at <url>` (otlp/jaeger) | raw URL + raw `TRACE_EXPORTER` | `redactURL(url)` + the matched constant (`otlp` / `jaeger`) |
| `Exporting traces to zipkin at <url>` | raw URL | `redactURL(url)` |
| `Exporting traces to GoFr at <url>` | raw URL | `redactURL(url)` |
| `TRACER_INSECURE is ignored for TRACER_URL=…` | raw URL | `redactURL(url)` |
| `traces are exported to … over plaintext with headers` | raw URL | `redactURL(url)` |
| `invalid TRACER_INSECURE …` | echoed the value | names the key and the accepted values |
| `unsupported TRACE_EXPORTER …` | echoed the value | lists `otlp, jaeger, zipkin or gofr` |

`redactURL` swaps userinfo and the query string for `REDACTED` instead of dropping them, so the log still shows that something was set there. It also drops the fragment. For a schemeless `user:pass@host:port`, it replaces everything before the last `@`. The value passed to the exporter itself is unchanged.

**What changes for users:** the two error messages no longer echo the bad value. The redacted URLs show `REDACTED` where credentials were. Nothing changes for an endpoint without credentials, which is what every existing test uses.

None to the API. Two log messages no longer include the configured value (see table).

**Additional Information:**

- `Test_redactURL`: 10 cases (userinfo, token-only userinfo, query, fragment, schemeless userinfo, unparsable input, and cases that stay unchanged).
- `Test_initTracer_doesNotLogCredentials` runs the real `initTracer` path for otlp, jaeger (with an ignored `TRACER_INSECURE`), an invalid `TRACER_INSECURE`, zipkin and gofr. It asserts the redacted line is in stdout/stderr and the secret is not.
- **Mutation-checked:** removing redaction from the zipkin log, from the otlp log, from the query branch, or from the schemeless branch fails 4–5 tests each time.
- `go test ./pkg/gofr/`: pass. `golangci-lint --new-from-rev=origin/development`: 0 issues. `actionlint` and `typos`: clean.
- **Not in scope:** alerts #172/#173 (`go/path-injection`, `datasource/file/local_fs.go`) are already open on `main` and predate this release.
- After merge, this commit needs to reach `release/v1.61.0` (cherry-pick, or re-cut the branch) for #4246's check to go green.

**Checklist:**


pkg/gofr/otel.go
```diff
@@ -331,11 +343,45 @@ func buildGoFrExporter(logger logging.Logger, url string) sdktrace.SpanExporter
...
+// schemeless host:port does not parse with a host, so everything before its
+// last '@' is replaced instead.
+func redactURL(raw string) string {
+	u, err := url.Parse(raw)
+	if err != nil || u.Host == "" {
+		if i := strings.LastIndex(raw, "@"); i >= 0 {
+			return redactedPlaceholder + raw[i:]
+		}
+
+		return raw
```
Umang01-hash: LOW / defense-in-depth: the schemeless branch redacts userinfo (via @) but not a query string. A schemeless URL with credentials in the query leaks — reproduced live: TRACE_EXPORTER=zipkin TRACER_URL='localhost:9411/api/v2/spans?api-key=SECRET' logs the api-key unredacted (url.Parse gives Host=='' so the RawQuery branch below is skipped). Same ?api-key= vector you already redact for scheme-bearing URLs. One-liner here: after the @ handling, if raw has a '?', replace everything after it with REDACTED (over-redaction is safe). Not blocking.

Umang01-hash (dismissed):
Verified independently at ddb19c13 — fuzzed redactURL and ran a live server, not just read the diff.

- All 5 URL log sites route through redactURL; the two value-echo lines now log static text. No breaking API change; gofmt/build/vet clean, golangci-lint --new-from-rev 0 issues.
- Live E2E: TRACER_URL=https://user:token@localhost:4317 logs https://REDACTED@localhost:4317 (secret absent). Mutation on the zipkin line fails Test_initTracer.../zipkin_query_key — tests are non-vacuous.

One LOW, non-blocking (inline). Also worth confirming in CI: otel.go:66 (unchanged) still echoes the raw TRACER_RATIO via the strconv error — not a credential, but check it isn't one of the 6 CodeQL alerts, else one survives.

Merge-ordering: this is on pre-#4206 otel.go; if the registry refactor lands first, redactURL + the log sites move into traces/exporters and this needs a rebase.

LGTM.

pkg/gofr/otel.go
aryanmehrotra: Fixed in 52952470f. Your repro was right: `localhost:9411/api/v2/spans?api-key=SECRET` parses with an empty `Host`, so it never reached the `RawQuery` branch.

When `url.Parse` gives no host, redaction now goes through `redactUnparsedURL`, in this order:
1. **userinfo** up to the last `@`
2. **fragment** dropped
3. **query** replaced with `REDACTED`

Userinfo goes first on purpose. If the query were cut first, a `?` inside a password (`user:s3?cret@host`) would cut the credential in half and leak the part before it. With userinfo first, the whole credential is replaced, and an `@` in a query value just over-redacts.

New cases: your exact repro through the real zipkin `initTracer` path, plus schemeless userinfo+query, a password containing `?`, `@` inside a query value, and a schemeless fragment. Reverting the query line fails 3 of them. Package tests pass; `golangci-lint --new-from-rev` reports 0 issues.

aryanmehrotra:
Thanks for checking it live. Replies to the two non-inline points:

**`TRACER_RATIO` (otel.go:66):** not one of the 6 alerts. From the CodeQL SARIF for `development`, the `go/clear-text-logging` results in `otel.go` are at lines 174, 207, 269, 290, 293 (×2), plus 319 and 334, which were already dismissed. None is at 66. This PR's CodeQL check passes, and none of #179–#184 has an instance on the PR merge ref. The ratio is a float, so I left that line alone.

**Merge order with #4206:** agreed. #4206 is still open against `development`. Whichever lands second rebases: if #4206 goes first, `redactURL`, `redactUnparsedURL` and the 5 log sites move into `traces/exporters` with their tests. This one is also needed on `release/v1.61.0` to unblock #4246, so merging it first keeps the release path simpler.

pkg/gofr/otel.go
```diff
@@ -331,11 +343,65 @@ func buildGoFrExporter(logger logging.Logger, url string) sdktrace.SpanExporter
...
+// redactedPlaceholder stands in for credentials when a tracer endpoint is logged.
+const redactedPlaceholder = "REDACTED"
+
+// redactURL returns a tracer endpoint that is safe to write to a log.
+//
+// A scheme-bearing TRACER_URL is a real URL, so it can carry credentials in its
+// userinfo (https://user:token@collector:4317) or its query
+// (https://collector/api/v2/spans?api-key=...). Both are replaced rather than
+// dropped, so the log still shows that something was configured there.
+func redactURL(raw string) string {
```
Umang01-hash, line 370: Newline/control char in the host/path of TRACER_URL survives redaction → forged log line in TTY pretty mode. Pre-existing (raw URL was logged before too), JSON-escaped in production, and TRACER_URL is operator config not attacker input — so out of scope here, just noting it.

pkg/gofr/otel.go
```diff
@@ -331,11 +343,65 @@ func buildGoFrExporter(logger logging.Logger, url string) sdktrace.SpanExporter
...
 
 	return NewExporter(url, logging.NewLogger(logging.INFO))
 }
 
+// redactedPlaceholder stands in for credentials when a tracer endpoint is logged.
+const redactedPlaceholder = "REDACTED"
+
+// redactURL returns a tracer endpoint that is safe to write to a log.
+//
+// A scheme-bearing TRACER_URL is a real URL, so it can carry credentials in its
```
Umang01-hash, line 358: Path-segment / matrix-param secrets (e.g. `https://host/v1/TOKEN/spans`) aren't redacted. Matches the documented userinfo+query contract and those aren't standard tracer credential spots — fine as-is.

pkg/gofr/otel.go
```diff
@@ -204,12 +204,24 @@ func (a *App) getExporter(name, host, port, url string) (sdktrace.SpanExporter,
 	case gofrTraceExporter:
 		exporter = buildGoFrExporter(a.Logger(), url)
 	default:
-		a.container.Errorf("unsupported TRACE_EXPORTER: %s", name)
+		a.container.Error("unsupported TRACE_EXPORTER: expected one of otlp, jaeger, zipkin or gofr")
```
Umang01-hash: Dropping the bad value from the error is safer, but a typo'd TRACE_EXPORTER is now harder to debug. Could keep the value out but hint the received length/first char, or leave it — your call.

Umang01-hash (approved):
Verified locally: redaction wraps only log args (connection endpoint untouched), all endpoint-log sites route through redactURL, auth headers never logged, no exported-API change. Ran an adversarial bypass sweep (userinfo/query/encoded/IPv6/scheme-relative/uppercase/opaque) — no credential leak. Live E2E with a credential-bearing TRACER_URL + TRACER_AUTH_KEY: neither leaks, log shows `https://REDACTED@…`. gofmt/vet/lint clean, tests fail-on-revert. Just a few optional nits.

Umang01-hash (dismissed):

pkg/gofr/otel.go
aryanmehrotra: Fixed in 176814361. Control characters make `url.Parse` fail, so they always reached the log through `redactUnparsedURL` unchanged. That path now escapes them (`\x0a`, `\x85`, …), so the value stays on one line in pretty output too. Tests cover \n with a forged JSON line, \r and \t, a C1 control, and a control character combined with userinfo. Removing the escape fails 5. The parsed path needs nothing: `url.Parse` rejects control characters, and `u.String()` percent-encodes the rest.

pkg/gofr/otel.go
aryanmehrotra: Agreed: kept out of scope, and `redactURL`'s doc comment now says so. No tracer backend GoFr supports takes credentials in a path segment, and redacting the path would hide the part of the endpoint an operator actually debugs.

pkg/gofr/otel.go
aryanmehrotra: Took the middle road in 176814361. The value is echoed as `unsupported TRACE_EXPORTER=otpl: …` only when it looks like a typo of an exporter name (≤16 characters, letters plus `-`/`_`). Anything else, such as digits, a URL, a control character or something long, logs as `TRACE_EXPORTER=REDACTED`, so a token pasted into the wrong variable still stays out. `Test_initTracer_doesNotLogCredentials` covers both directions through the real startup path, and dropping the character check fails 7 tests. I used `=` rather than `%q` because the JSON logger escapes the quotes.

Umang01-hash (approved):

PR #4258, opened 2026-09-20
fix(cmd): exit non-zero when a command handler returns an error (#4255)

### LINKED ISSUE #4255: CMD apps always exit 0 even when the handler returns an error
**Describe the bug**

For a CMD/CLI app built with GoFr, the process exits 0 even when the sub-command handler returns an error. Scripts, shell pipelines, and CI can't detect a failure.

`pkg/gofr/cmd/responder.go`

```go
func (*Responder) Respond(data any, err error) {
    // TODO - provide proper exit codes here. Using os.Exit directly is a problem for tests.
    if data != nil {
        fmt.Fprintln(os.Stdout, data)
    }

    if err != nil {
        fmt.Fprintln(os.Stderr, err)
    }
}
```

The error is printed to stderr but the exit code is never set, so `echo $?` is always 0. There's already a TODO noting this. The `Responder` type and `Respond` method also have no godoc.

**To Reproduce**

1. A CMD app whose handler returns an error:

   ```go
   app := gofr.NewCMD()
   app.SubCommand("fail", func(*gofr.Context) (any, error) {
       return nil, errors.New("something went wrong")
   })
   app.Run()
   ```

2. Run it:

   ```
   $ go run . fail
   something went wrong
   $ echo $?
   0        # expected non-zero
   ```

**Expected behavior**

When a handler returns an error, the process should exit with a non-zero status so callers can detect failure. The TODO notes that calling `os.Exit` directly breaks tests — a settable exit hook (e.g. an `exit func(int)` defaulting to `os.Exit`) keeps it testable.

I'd be happy to work on a fix.

NitinKumar004:
**Description:**

Fixes #4255.

CMD apps built with GoFr exited `0` even when the sub-command handler (or command
resolution) returned an error, so shells, pipelines, and CI could not detect the
failure.

Changes:
- `Responder` now records whether it responded with a non-nil error (`Errored()`).
- `cmd.Run` propagates that up to the caller.
- `runCMD` exits with a non-zero status **after** telemetry is flushed and the logger
  is closed, so a failed command still reports its final metrics/traces before the
  process ends. `os.Exit` is wrapped in a small testable seam (`osExit`) so the exit
  path can be unit-tested without terminating the test binary — this also addresses
  the existing TODO in `responder.go`.

Behavior (verified end-to-end with a real CMD app):
- successful command -> exit `0`
- handler returns an error -> exit `1`
- unknown command -> exit `1`

None to the public API. The only behavioral change is the intended one: a CMD app now
exits non-zero when a command fails (previously it always exited `0`).

**Additional Information:**

No new dependencies. `cmd.Run`'s unexported signature changed to return a `bool`
(internal only).

**Checklist:**


aryanmehrotra (changes requested):
Thanks for this — the behaviour is right and the exit-code matrix is correct. I built a CMD app with 8 subcommands and drove it out-of-process, reading `ExitError.ExitCode()`:

| case | `development` | this PR |
|---|---|---|
| handler returns nil | 0 | 0 |
| handler returns error | **0** (the bug) | **1** |
| data **and** error | 0 | **1** (stdout still written) |
| unknown / missing subcommand | 0 | **1** |
| `--help` | 0 | **0** |
| stderr written, nil error | 0 | **0** |

The last row is the one I most wanted to check and you got it right — keying off the returned `error` rather than stderr output is correct, and a naive implementation would have conflated them.

The flush ordering is careful work: extracting `flushCMDTelemetry` so its deferred `cancel` runs before the exit is a subtlety that's easy to miss, and I confirmed the final `ERROR` line reaches `CMD_LOGS_FILE` on the failure path.

Requesting changes on one thing.

### `testing.Testing()` makes the shipped behaviour untestable by construction

```go
if failed && !testing.Testing() {
    os.Exit(1)
}
```

I deleted `os.Exit(1)` from `run.go:41` entirely and the full `pkg/gofr` suite still passed (`ok gofr.dev/pkg/gofr 7.056s`). Coverage agrees — `run.go:39.34,42.3` has 0 hits. So the branch that defines this PR's whole contract is the one branch no test can reach.

Two consequences beyond that:

- A user who writes an in-process test asserting their CLI fails is told it succeeded.
- A CLI driven from a `go test -c` binary gets a different process contract than the same code shipped.

Your original `var osExit = os.Exit` (`a6665d3bc`) was the right shape, and `e02aca3da` removed it along with `run_test.go`. Could we bring the seam back and solve `examples/sample-cmd`'s two failure-path tests another way — either by having them set the seam, or by running the binary out-of-process for those two cases?

Worth knowing: #3942 adds an `exit func(int)` field on `App` plus `exitCodeStartupFailed = 1` and `abortStartup()`, for exactly this reason, about 30 lines away in the same file. I test-merged the two — no textual conflict, builds clean, tests pass — so this will land silently and leave `run.go` holding two different exit mechanisms. If #3942 goes first this becomes a two-line change and the `testing` import, the guard and the second `//nolint` all disappear.

### The description no longer matches the code

The body says *"`os.Exit` is wrapped in a small testable seam (`osExit`)"*, but `grep -rn osExit` on the branch returns nothing. Please update it — a reviewer reading the description would approve a design that isn't there.

### Smaller points

- **User `defer`s in `main()` are skipped on the failure path.** With `defer cleanup()` as the first line of `main`, `./app ok` runs it and `./app fail` does not. Anyone doing `defer db.Close()` or `defer pprof.StopCPUProfile()` loses it exactly when things went wrong. Worth a sentence on `runCMD`'s doc comment.
- **The `//nolint:revive` argues the wrong thing.** Its explanation ("exit status 1 signals the failed command to shells and CI") argues *why code 1*, not *why a deep exit is acceptable here*. I confirmed the lint genuinely fires without it (`deep-exit: calls to os.Exit only in main() or init() functions`). `abortStartup` in #3942 spends five lines on the exemption; same file, same rule.
- **Docs.** `docs/advanced-guide/building-cli-applications/page.md` never mentions exit codes. This is now a process contract that scripts and CI branch on — two lines there would help: nil → 0, error or unknown command → 1, `--help` → 0.
- **Nit:** `Errored()` is an addition to the exported `pkg/gofr/cmd` API, so "None to the public API" is slightly off. Additive and non-breaking, but worth stating accurately.
- **Nit:** `os.Exit(1)` is a bare literal here; #3942 introduces `exitCodeStartupFailed = 1`. Whichever lands second, these should share one constant.

Also checked and dismissed: importing `testing` adds no dependency — it's already a transitive import of `pkg/gofr` via `service`, `datasource/sql` and `container`. No binary-size concern there.

Verified: `gofmt` clean, `go vet` clean, `golangci-lint` has 0 issues in any changed file, all three new `cmd_test.go` tests kill their mutants, CI green on Go 1.24/1.25/1.26. The two local `-race`/`-count=5` failures (`websocket.go:47` race, port-2121 probe) reproduce identically on `development`.

NitinKumar004:
Thanks for driving it out-of-process. You're right that the `testing.Testing()` guard made the one branch that matters unreachable.

Pushed 2758da48:

- **The guard is gone.** The exit goes through an `exit func(int)` field on `App`, the same field #3942 adds with the same meaning: nil means `os.Exit`. `TestApp_runCMD_exitCode` sets it and covers nil error, error, data+error, unknown command, missing command and `--help`. Deleting the exit call fails it now, and so does `TestNewCMD`. That test was only passing before because of the guard; it now pins `os.Args` and records the exit.
- **examples/sample-cmd:** I went with running the binary out-of-process for the two failure paths. They re-run the test binary as the app and assert exit code 1 plus the exact stderr. One thing fell out of that: `command help` was in the help table, but `help` isn't a registered subcommand. It always printed `'help' is not a valid command.`, and the old test only checked stdout. It now exits 1 like any unknown command, so I moved it to the invalid-command table. `-h`/`--help` still exit 0.
- **#3942:** I test-merged #3942's head. There's a one-line conflict where both add the `exit` field at the end of `App`. Taking either side resolves it, and it builds and tests clean. So it's not silent, but it's trivial. The exit code is now a named `exitCodeCommandFailed = 1`. Whichever of the two lands second, I'll fold `exitProcess`/`abortStartup` and the two constants into one so `run.go` doesn't keep two exit mechanisms.
- **nolint:** it now argues why a deep exit is acceptable there: last step of the CMD app's `main`, after the telemetry flush and logger close, and the exit status is the only thing a shell or CI reads.
- **Defers:** called out in `runCMD`'s doc comment and in the docs.
- **Docs:** added an "Exit Codes" section to `building-cli-applications` with the table and the defer caveat.
- **Description:** rewritten. The old "osExit seam" text is gone, and it now says `Errored()` is an additive exported method rather than "none to the public API".

aryanmehrotra (changes requested):
The blocking item is genuinely fixed, and I verified it by mutation rather than by reading.

Deleting the exit decision — `if failed { a.exitProcess(exitCodeCommandFailed) }` — now fails **7 tests across 2 packages**:

```
--- FAIL: TestNewCMD
--- FAIL: TestApp_runCMD_exitCode/{handler_returns_error, handler_returns_data_and_error,
                                  unknown_command, missing_command}
--- FAIL: TestCMDRunWithNoArg
--- FAIL: TestCMDRunWithInvalidCommand/{invalid, help}
```

And deleting the real `os.Exit(code)` inside `exitProcess` fails the three `examples/sample-cmd` subprocess tests. Under the old head, deleting `os.Exit(1)` broke **nothing**. So the decision is covered in-process and the shipped exit is covered out-of-process by real child processes — that is the right split, and it is what the seam bought.

The exit-code matrix is **byte-identical** across all 12 rows between `e02aca3da` and `2758da48a`, measured with real binaries where `a.exit` is nil. The refactor moved no behaviour. `testing` is gone from `run.go`'s imports. The `defer`-in-`main()` doc matches reality — I ran it both ways and `MAIN-DEFER-RAN` appears on every exit-0 row and none of the exit-1 rows. The `//nolint:revive` now argues *why a deep exit is acceptable* and cites `logging/logger.go:183,195` as the in-repo precedent, which is exactly what I was asking for.

Adopting `App.exit` rather than reinventing a seam was the better call, and flagging the #3942 conflict in your own body before anyone raised it is appreciated.

Two things before I approve, one of them small but user-visible.

### 1. `mycli help` now exits 1, and the docs table doesn't say so

`parseArgs` (`cmd.go:79`) only treats `-h` and `--help` as help flags. A bare `help` falls through as a subcommand, isn't registered, and now exits **1** — measured, both at the old head and this one.

The Exit Codes table currently reads:

```
| The subcommand is missing or not registered | 1 |
| -h / --help on its own or after a registered subcommand | 0 |
```

Someone scanning that will reasonably read `help` as covered by the second row. It isn't — it's covered by the first. That section exists precisely so scripts and CI know what to expect, and `mycli help` in a smoke step is a realistic thing to have; it silently returned 0 before this PR and returns 1 after. One row fixes it:

```
| `help` as a bare word (it is not a help flag — see `-h` / `--help`) | `1` |
```

While you're there, a registered subcommand with a **nil handler** also exits 1, and the table only says "missing or not registered". Worth folding in, since the two print different things: an unknown command prints help to stdout, a nil handler doesn't.

### 2. The flush → close → exit ordering isn't pinned, and this PR is where it can be

I moved `if failed { a.exitProcess(...) }` to the **top** of `runCMD`, before `flushCMDTelemetry()` and `closer.Close()`:

```
ok  gofr.dev/pkg/gofr             14.927s
ok  gofr.dev/pkg/gofr/cmd          9.221s
ok  gofr.dev/examples/sample-cmd  15.044s
```

Nothing catches it. That ordering is the whole reason `flushCMDTelemetry` was split into its own function, and the commit message says so — but a future refactor can silently undo it.

That was true before this PR too, so I'm not calling it a regression. I'm asking because **this PR is what makes it cheap to close**: `runMainInSubprocess` already exists, so it's a failing command with `CMD_LOGS_FILE` set plus an assertion that the final ERROR line is on disk. Roughly ten lines, and it pins an invariant that currently has nothing holding it.

### Smaller notes, no action needed

- `runMainInSubprocess`'s doc says it "returns its stdout, stderr and exit code", but the `require.ErrorAs(t, c.Run(), &exitErr, …)` inside means it can only be used for non-zero exits — a success case fails with a confusing message. A name like `runMainInSubprocessExpectingFailure`, or returning the code without asserting, would read better.
- `(*cmd.Responder).Errored()` is permanent exported API on an exported type, and `Responder` gains mutable state. Additive and justified — just flagging that it's public surface you can't take back.
- `examples/` is excluded from golangci-lint (`.golangci.yml:166`), so the new 84-line subprocess harness gets no lint coverage. Not your doing.
- On coverage: `run.go:62.2,62.15` (the `os.Exit(code)` statement itself) shows 0 hits, which is unavoidable by design — it's unreachable in-process. If codecov gates on patch coverage that will read as one miss. `runCMD` is otherwise 100%, and `cmd/responder.go:31 Errored` went 0% → 100%.

### On the #3942 merge

I test-merged and confirmed your prediction: the only conflict is the `exit func(int)` field, both PRs adding it in the same spot after `readinessChecks`. After a four-line resolution the merged tree builds, vets and tests clean.

Two mechanisms do survive that merge rather than one — `exitProcess`/`exitCodeCommandFailed` here and `abortStartup`/`exitCodeStartupFailed` there, both doing nil-check-then-`os.Exit`, both `= 1`, with two copies of the same nolint justification. Your body already offers to fold them. #3942 is mine, so: whichever of us lands second does the fold, and I'll take it if #4258 goes first. Nothing for you to change now.

Gates all clean otherwise — `gofmt` clean on all 9 changed files, `go vet` clean, no new lint in any touched file, `pkg/gofr/cmd` coverage 97.8% → 97.9%. The `-race` and `-count=5` failures reproduce identically at the merge base (`websocket.go:47` ↔ `handler.go:190`, `terminal/spinner.go:78` vs `:62`, and the port-2121 `isPortAvailable` `os.Exit`).

NitinKumar004:
Thanks, and thanks for measuring the matrix with real binaries.

Pushed ecc06f1c:

1. **`help` and nil handlers in the Exit Codes table.** I added a row for bare `help` (not a help flag; only `-h`/`--help` are) → 1, and one for a subcommand registered with a nil handler → 1. There's also a note on the output difference: an unknown or missing command prints the command list to stdout, a nil-handler command prints only the error. I checked both with a throwaway app first. While doing that I noticed `nilh -h` also exits 1, so the `-h`/`--help` row now says "after a registered subcommand that has a handler".
2. **Pinning flush → close → exit.** I went a slightly different way from the log-file idea, because I don't think it would catch the reorder. `NewFileLogger` writes straight to an unbuffered `*os.File`, so the final ERROR line is on disk before `runCMD`'s cleanup runs, whatever the order. Instead, `TestApp_runCMD_cleanupOrder` records events through the seams this PR already has: `App.shutdownTracer` for the flush, the logger's `Close`, and `App.exit`. It asserts `flush traces → close logger → exit(1)` for a failed command and `flush → close` for a successful one. Moving the exit to the top of `runCMD` fails it (it's the only test that catches that one), and so does swapping flush and close.
3. **`runMainInSubprocess`** now returns the exit code instead of asserting failure, so it works for successful runs too. The existing callers still assert 1.

Noted on the rest, including `Errored()` being permanent API and the unreachable `os.Exit` line. Thanks for taking the #3942 fold if this lands first.

PR #4262, opened 2026-09-22
fix(cmd): preserve equals signs in request parameter values

ku2507u0309-create:
## Description

Fixes request parameter parsing when a value contains one or more `=` characters.

`NewRequest` previously split option arguments on every `=` character. As a result, values such as `--query=a=b` and `--token=abc==` were discarded because they produced more than two parts.

The parser now splits only at the first `=` and preserves the remainder of the value, including additional `=` characters and empty values.

## Breaking Changes

None. Existing flag parsing and `--key=value` / `-key=value` behavior are preserved.

## Testing

- Added table-driven regression tests for values containing additional `=` characters.
- Added coverage for empty values.
- Preserved flag behavior and ignored non-option arguments.
- Ran the relevant Go tests and formatting checks.

## Checklist

- [x] Formatted with gofmt/goimports.
- [x] New code is covered by tests.
- [x] No new dependencies were added.
- [x] No public API was changed.
- [x] PR targets the development branch.


NitinKumar004 (commented):
Thanks for this. It's a real bug: before this change, any option whose value contained more than one `=` (for example `--token=abc==` or `--url=http://h/p?x=1&y=2`) was silently dropped. The one-line switch to `SplitN` fixes it cleanly.

**What I checked locally**
- `go build`, `go vet`, `go test -race` on `pkg/gofr/cmd` and `pkg/gofr`: all pass. The only failure is `TestSpinner` in `pkg/gofr/cmd/terminal`, and it fails the same way on `development`, so it is unrelated.
- golangci-lint v2.12.2 on `pkg/gofr/cmd/...`: 0 issues on both base and this branch. Coverage is unchanged at 97.8%.
- I ran four mutations of the fix: reverting to `Split`, `SplitN(…, 3)`, trimming trailing `=`, and mapping an empty value to `true`. The new test catches all four.
- I probed edge cases against base. `-a=b=c`, `-a==`, unicode keys and values, `--q=x=y,z=w` through `Params`, and `Bind` with `-Token=abc==` now all work. `--a=`, `--a`, repeated flags (last wins), `--a=-b` and positional `k=v` args behave exactly as before. Subcommand routing in `parseArgs` is unaffected.
- The branch is 10 commits behind `development` but merges cleanly.

**Blocking**

None.

**Nits**
1. `go.work.sum:1796`: the last commit adds `golang.org/x/net v0.57.0/go.mod`, but nothing in the repo requires that version. It looks like a leftover from a local run. Could you drop it (`git checkout origin/development -- go.work.sum`)?
2. `pkg/gofr/cmd/request_test.go:111-119`: the third case mixes empty values, flags and non-option args. `--empty=` already worked before this change, so that case doesn't guard the fix. Consider splitting it into focused cases and adding `-a==` (expected value `=`), a unicode value such as `--name=héllo=wörld`, and a `Params` case like `--q=x=y,z=w` (expected `["x=y", "z=w"]`).
3. `pkg/gofr/cmd/request.go:56` (optional): `key, val, ok := strings.Cut(a, "=")` states "split at the first `=`" directly and would let you drop the `argsLen` constants and the switch.
4. Minor: the PR description says the parser now "preserves the remainder of the value, including additional `=` characters and empty values," which reads as if empty values were part of what this fixes. They weren't — `--a=` already worked before this change. Only values with more than one `=` were being dropped.

Unrelated, but I noticed it while testing: `docs/advanced-guide/building-cli-applications/page.md` shows `greet --name John`, and the parser doesn't support space-separated values (it yields `name=true`). That's a separate issue, not something for this PR.

Once the `go.work.sum` change is removed, this looks good to merge.

ku2507u0309-create:
Hi,

Thank you for the feedback. I’ve updated the request parsing tests to focus
on the specific edge cases raised and removed the unintended workspace
checksum change.

The parser behavior remains unchanged: it still splits at the first = and
preserves the remainder of the value. I also added regression coverage for:

   - -a==
   - --name=héllo=wörld
   - --q=x=y,z=w → ["x=y", "z=w"]

This PR is now ready for review. Please let me know if you’d like any
additional edge cases covered or if there are any other concerns.

Thanks again.

Issue #3428, opened 2026-05-14
Reduce per-request CPU and allocations


aryanmehrotra:
## Background

On a baseline run (Apple M4, GOMAXPROCS=4, loopback, hey -c 50) GoFr serves ~90k RPS. Per request that's ~44 µs of CPU and ~130 KB of allocations, with most of the allocation pressure in the OTel SDK path even when no TRACE_EXPORTER is configured.

We want the framework lighter and faster.

## Targets

| metric              | today    | target    |
|---------------------|---------:|----------:|
| RPS (local Apple)   |     90 k |  150 k+   |
| CPU µs per request  |       44 |    ≤ 25   |
| bytes alloc per req |  ~130 KB |   ≤ 10 KB |
| p99 latency         |   3.5 ms |  < 1.5 ms |

Zero breaking changes:
- response envelope `{"data": ...}` byte-equal
- Prometheus metric names and label keys unchanged
- env var names unchanged
- W3C `traceparent` propagation byte-for-byte preserved
- public method signatures unchanged

## Plan

This issue tracks the work as a series of PRs. Each checkbox is one PR.

### Setup
- [ ] Microbenchmarks for router, responder, span start, attr build, context, full chain
- [ ] Snapshot tests for response envelope, Prometheus output, log line schema, public API surface
- [ ] Trace propagation parity tests (inbound, outbound, baggage)
- [ ] Middleware chain-shape invariant test
- [ ] CI workflow that runs benchmarks + wrk on every PR and fails on RPS/CPU/B regression

### Fixes
- [ ] Use a noop TracerProvider when no TRACE_EXPORTER is set
- [ ] Cache `otel.Tracer("gofr")` once on the app
- [ ] Single Write + Content-Length in the responder
- [ ] Share one StatusResponseWriter between Logging and Metrics middleware
- [ ] Pool the shared StatusResponseWriter
- [ ] Pool RequestLog
- [ ] Cache traceID on `*gofr.Context`, omit zero-value field from logs
- [ ] Defer `time.Now()` past the probe filter in logger and metrics middleware
- [ ] Precompute `attribute.Set` per (method, route template) at registration
- [ ] CORS short-circuit when no policy is configured
- [ ] Skip auth middleware in the chain when no provider is wired
- [ ] Path-clean fast path in the router (only when URL contains `//` or `.`)
- [ ] Skip request-log build for 2xx when `LOG_LEVEL >= ERROR`
[...]

PR #4264, opened 2026-09-22
perf(http,observability): reduce per-request CPU and allocations in responder, tracer, and context

Barathwaj2006:
## Pull Request Template

**Description:**
This PR reduces per-request CPU and allocation overhead across GoFr's HTTP transport, logging, and OpenTelemetry observability layers, addressing #3428:

1. **HTTP Responder Content-Length & Zero-Alloc Cache** (`pkg/gofr/http/responder.go`):
   - Sets `Content-Length` on JSON, File, and XML responses before `w.WriteHeader()`. This prevents Go's `net/http` from falling back to chunked transfer encoding (`Transfer-Encoding: chunked`), enabling HTTP keep-alive connection reuse and predictable downstream streaming.
   - Pre-computes `smallContentLengths [128][]string` cache and canonical header key so lengths <128 bytes (covering health checks, common JSON envelopes, and API errors) allocate zero additional bytes.
2. **Tracer Middleware Status Code Attribute Precomputation** (`pkg/gofr/http/middleware/tracer.go`):
   - Pre-computes `attribute.KeyValue` for HTTP status codes [100, 600), eliminating 1 heap allocation per recording span (BenchmarkTracer_Recording: 18 -> 17 allocs/op).
3. **ContextLogger Trace ID Lazy Caching** (`pkg/gofr/logging/ctx_logger.go`):
   - Lazily formats and caches `traceID` string on first log invocation. Multiple logs emitted within the same request reuse the cached string, eliminating repeated 32-character hex formatting while keeping silent requests at 0 allocations.
4. **Context Tracer Contention & Sub-nanosecond Correlation ID** (`pkg/gofr/context.go`):
   - Caches `contextTracer = otel.Tracer("gofr-context")` at package level, eliminating TracerProvider read-lock contention and map lookup in `c.Trace()`.
   - Lazily caches `traceID` on `*Context`, reducing repeated `GetCorrelationID()` calls to sub-nanosecond 0-allocation lookups (BenchmarkContext_GetCorrelationID_Cached: 0.76 ns/op, 0 allocs), and returning static `zeroTraceID` when no span exists.

- None. All public signatures, return types, interfaces, and wire formats remain 100% backward compatible. Existing custom `Content-Length` headers are strictly preserved.

**Additional Information:**
- Zero `init()` functions used (deterministic package variables with helper constructors).
- Benchmarks:
  - `BenchmarkContext_GetCorrelationID_Cached`: 0.76 ns/op (0 B/op, 0 allocs/op)
  - `BenchmarkTracer_Recording`: 17 allocs/op (down from 18 allocs/op)
  - `BenchmarkRespond_String`: 259.2 ns/op (1 alloc, 16 B/op) — zero extra allocations with Content-Length added
  - `BenchmarkRespond_Map`: 491.3 ns/op (+3.0% speedup)

**Checklist:**


NitinKumar004 (changes requested):
Thanks for working on this, @Barathwaj2006. The hot path is exactly where we want these wins, so I spent some time checking it closely against the merge-base (2411b180).

**What I ran locally**
- `go build`, `go vet`, and `go test -race` on `pkg/gofr`, `pkg/gofr/http/...`, and `pkg/gofr/logging/...`. The only `-race` failures (`TestContext_WriteMessageToSocket`, `remotelogger`) also fail on base, so they're not from this PR.
- golangci-lint v2.12.2 on base and on the PR, then a diff of the two.
- Real `httptest` servers over HTTP/1.1 and HTTP/2 (TLS), comparing status, headers, and body for GET/HEAD/DELETE, error envelopes, the encode-failure fallback, File, and XML.
- Benchmarks with `-benchmem -count=6`, base and PR binaries interleaved, compared with benchstat.
- Mutation testing of the core changes, plus a few extra probe tests.
- `git merge-tree` against current `development`: it merges cleanly and the merged tree builds and passes tests.

### Blocking

1. **New data races on the request context** (`pkg/gofr/logging/ctx_logger.go:61-62`, `pkg/gofr/context.go:241/256`). `withTraceInfo` and `GetCorrelationID` used to be read-only; now they lazily write `traceID`, so a handler that logs or reads the correlation ID from more than one goroutine races. This isn't a stretch for GoFr specifically: the shipped `TraceHandler` example (`examples/http-server/main.go`, the `go func() { c.Redis.Ping(c) }` fan-out) already hands the same `*gofr.Context` to several goroutines, and it's one line away from one of them also calling `c.Errorf(...)` on failure. I confirmed the race directly: 8 goroutines calling `ctx.Error(...)` and `ctx.GetCorrelationID()` on one `Context` pass cleanly on base and produce two `DATA RACE` reports on the PR, at exactly those lines. I'd drop both caches. If we really want the logger one, it needs to be concurrency-safe (e.g. `sync.Once`), but I'm not convinced one saved alloc per log call is worth the complexity.

2. **Stale correlation ID after `ctx.Trace()`** (`pkg/gofr/context.go:238-256`). When no span exists yet, the zero ID gets cached. `Trace()` then swaps `c.Context` for a context holding a new root span, but the cache is never cleared. CMD contexts start with no span (`newCMDContext` just wraps `r.Context()`, and CMD never runs the Tracer middleware), so this isn't an edge case for that transport. I tested it: base returns the new span's trace ID after `Trace()`, while the PR keeps returning `000...0`.

3. **`contextTracer` is pinned to the first `TracerProvider`** (`pkg/gofr/context.go:53`). `otel.Tracer("gofr-context")` is evaluated once, at package init, before any `App` exists. otel's internal delegate (`internal/global/state.go`, `delegateTraceOnce sync.Once`) binds a pre-SDK tracer like this one to whichever `TracerProvider` wins the *first* `SetTracerProvider` call in the process, permanently — later calls only update `otel.GetTracerProvider()`, they don't rebind existing tracers. `App.initTracer` (`pkg/gofr/otel.go:78`) calls `otel.SetTracerProvider` unconditionally on every `App` construction, so this is reachable by nothing more exotic than building two `gofr.New()` apps in one process — a normal thing to do across test cases in the same package — where the second app's `ctx.Trace()` spans silently go to the first app's (possibly already-shut-down) provider. I reproduced it with two providers: base sends the span to the second provider, the PR sends it to the first. The `Tracer` middleware resolves its own tracer the same way but at chain-build time, after `initTracer` has already run, which is why it doesn't have this problem — see the comment above `middleware.Tracer`. Could `ctx.Trace` do the same (resolve after `initTracer`), or just call `otel.GetTracerProvider().Tracer(...)` per call like it used to?

4. **HTTP/2 204 responses now carry `Content-Length: 3`** (`pkg/gofr/http/responder.go:173-175`, and the File/XML branches at `:237` and `:260`). A DELETE success returns 204, and the envelope `{}\n` is 3 bytes (coincidentally, the File/XML test cases I used also produced a 3-byte body). On HTTP/1.1, net/http strips the header for statuses that can't have a body, but the HTTP/2 server doesn't. RFC 9110 §8.6 says a server MUST NOT send Content-Length on a 204, and GoFr serves HTTP/2 whenever `CERT_FILE`/`KEY_FILE` are set (`pkg/gofr/http_server.go:144`, `ListenAndServeTLS`). Every other case I compared was byte-identical to base. The only other wire change is that bodies over ~2KB now get a fixed Content-Length instead of chunked encoding, which is fine. Suggestion: skip setting Content-Length for statuses that can't have a body (1xx, 204, 304).

5. **Lint**: 6 new issues compared with base — 5 `wsl_v5` (`responder.go:237,240,260,263`, `context_test.go:407`) and 1 `testifylint` (`context_test.go:384`, float-compare; `assert.Zero(t, allocs)` or `assert.InDelta` fixes it). Separately, `gocyclo` on `responder.go`'s `handleSpecialResponseTypes` was already flagged on base and gets worse with this PR (12 → 16) — worth a look, but it's not a new finding, just a worse one.

### Nits

- `pkg/gofr/context.go:251`: `sc.IsValid()` also requires a span ID, so a `SpanContext` that has a trace ID but no span ID now returns zeros, where base returned the trace ID. `sc.HasTraceID()` keeps the old behavior.
- `pkg/gofr/http/middleware/tracer.go:135-161`: I couldn't reproduce the 18 → 17 alloc claim. `BenchmarkTracer_Recording` is already 17 allocs/op on base, and still 17 with this PR — `attribute.Int` doesn't allocate. I'd drop the 600-entry table.
- Responder: allocs and B/op are identical on all the `Respond` benchmarks, and the sec/op difference isn't significant on the interleaved runs, so I'd reword the numbers in the description. Chunked responses also keep the connection alive — chunked transfer encoding exists specifically to support keep-alive without a known length — so the keep-alive rationale in the description isn't accurate either.
- Every request's `Context` grows from 160B to 192B (`BenchmarkContext_New`) to hold two caches that most requests never read. That's a small step backwards for #3428.
- `pkg/gofr/logging/ctx_logger.go:17-27`: the doc comment still says repeated logging "pays per call"; it needs updating now that the cache makes repeats free.
- File/XML Content-Length has no test. A mutation to `len(v.Content)+1` still passes `pkg/gofr/http`. The `statusCodeKV` fallback outside [100, 600) isn't covered either.

The logger cache is a real saving (one alloc per log call after the first), so a race-free version of that part would be welcome. Happy to re-review once the above is addressed.

